PR #598, opened 2018-03-09
[WIP] Setup mypy type checker

smessmer:
- Setup mypy type checker
- Minimal set of code changes to make it pass
- Add it to Travis CI

PR #607, opened 2018-03-12
[Typing 1/3] Setup mypy type checker

smessmer:
* Setup mypy type checker
* Minimal set of code changes to make it pass
* Add it to Travis CI

This is the final version of https://github.com/onnx/onnx/pull/598, which was WIP. This one is for code review and landing.


onnx/backend/test/case/base.py
```diff
@@ -8,8 +8,9 @@
 import os
 import sys
 from textwrap import dedent
+from typing import Dict, Text, List, Tuple
 
-import numpy as np
+import numpy as np  # type: ignore
```
bddppq: Is there a way to avoid this? Need to mark every numpy import is pretty cumbersome.

.travis/install.sh
```diff
@@ -3,4 +3,4 @@
 script_path=$(python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" "${BASH_SOURCE[0]}")
 source "${script_path%/*}/setup.sh"
 
-time ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI pip install -v .
+time ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI pip install -v .[mypy]
```
bddppq: I prefer to keep here the same (because we want to test whether the "standard" installation works), and then after all the tests have passed, we install mypy and do type checking (like what we do for using flake8 to lint).

onnx/backend/test/case/base.py
smessmer: I could set `ignore_missing_imports = True` in the config file. However, that would mean that mypy won't report any invalid imports anymore (i.e. not just missing typing information, but it'll also not show any warning if the whole module is missing). And if for some reason a module exists but mypy doesn't find it, it'll also silently ignore it and not typecheck anything related to that module, hiding a lot of errors. Not sure if that is a good idea.

smessmer: An alternative might be to have a local onnx.numpy module that just does

    from numpy import *  # type: ignore

Then, other modules could do

    from onnx import numpy

without having to add `type: ignore`

smessmer: @bddppq Opinions?

bddppq: I'm a little scared of playing with python import mechanism :-), let's not add a local numpy import proxy.

setup.cfg
```diff
@@ -30,4 +30,29 @@ exclude =
   build/*,
   third_party/*
   *_pb2.py,
-  .cache/*
\ No newline at end of file
+  .cache/*
+
+[mypy]
+# follow-imports = silent  # TODO remove this
+mypy_path = stubs:third_party/pybind11
```
bddppq: What does this line do?

setup.py
```diff
@@ -242,9 +252,16 @@ def run(self):
                     PROTOC,
                     '--proto_path', SRC_DIR,
                     '--python_out', SRC_DIR,
+                    ] + protoc_mypy_args + [
                     '--cpp_out', SRC_DIR,
                     proto
                 ])
+                if gen_mypy:
+                    # Workaround for https://github.com/dropbox/mypy-protobuf/issues/8
+                    for file in os.listdir('onnx'):
```
bddppq: nit: don't use `file` as variable name, which is a builtin
use `SRC_DIR` instead of "onnx" (former is an absolute path)

setup.cfg
smessmer: It allows mypy to find typing information (`.pyi` files or `.py` files) in the `stubs/` directory and in the `third_party/pybind11` directory (`pybind11` for the corresponding module and `stubs` for some general local overrides)

onnx/backend/test/case/base.py
fumihwh: @smessmer 
BTW, what does it ignore? Imported but not used?

PR #608, opened 2018-03-13
add output shape as input for reshape

linkerzhang:
if we all agree, test case will be added :).


fumihwh:
Just a question, why do we need attribute `shape` and input `output_shape` both for defining the shape of output tensor?

linkerzhang:
@fumihwh we don't need both definitely. that's why I'm adding "output_shape" as optional for backward compatibility.  but as @bddppq said, if we all agree to make a breaking change here, it's a cleaner way. Let me do that and see whether there're any other concerns from other folks.

bddppq:
This change will pretty much break all models since I imagine Reshape is widely used.
One idea we have discussed before is for each breaking change we also provide a conversion/migration script to upgrade the existing models. Shall we use this as an opportunity to figure out such a mechanism? :-) cc @dzhulgakov @ezyang

linkerzhang:
Looks good! Thank you so much! @bddppq

PR #610, opened 2018-03-13
[Typing 2/3] Add python type hints for C++ code

smessmer:
This adds (manually written) type hints for our pybind11 C++ code, so the python/c++ interface can be type checked. Unfortunately, there isn't a good way to autogenerate them.


bddppq:
@onnxbot retest this please

PR #612, opened 2018-03-14
Minor code quality improvements in optimizer/

smessmer:
Minor improvements to readability, performance and compile-time checking. Also introduce make_unique<>, which allows better exception safety.


smessmer:
@houseroad It has advantages like exception safety (i.e. it's much harder to use it wrongly), and I don't see any disadvantages. Is there something I'm missing?

onnx/optimizer/optimize.h
```diff
@@ -14,25 +15,20 @@
 
 namespace ONNX_NAMESPACE { namespace optimization {
 
-void PrepareOutput(const ONNX_NAMESPACE::ModelProto& mp_in, ONNX_NAMESPACE::ModelProto& mp_out);
+ONNX_NAMESPACE::ModelProto PrepareOutput(const ONNX_NAMESPACE::ModelProto& mp_in);
 
-struct Optimizer {
+struct Optimizer final {
```
houseroad: If user want, they can have their own optimizer, which is a sub-class of our Optimizer class.

houseroad:
I prefer not introducing make_unique, since it won't bring many benefits in our case, and it may mislead developers by leaving them impression that they can use C++14 feature here.

onnx/optimizer/optimize.h
smessmer: Ok makes sense. I removed `final` and added a virtual destructor to avoid partial destruction when people subclass it.

smessmer:
I don't think we should avoid language features we can make work in c++11 just because they look like c++14. The reason we're sticking to c++11 is not that we like it better, but that we have to support older compilers. Whatever features bring us advantages and we can make work in our set of supported compilers (i.e. in c++11), we should. It's the job of CI to make sure nobody accidentally uses c++14 features we didn't backport to c++11 (yet).

onnx/optimizer/optimize.h
```diff
@@ -31,6 +31,8 @@ struct Optimizer final {
     _registerOptimizer<SplitPredict>();
   }
 
+  virtual ~Optimizer() = default;
```
rdzhabarov: curious, why there is a dup above
```
_registerOptimizer<FuseConsecutiveTransposes>();
_registerOptimizer<FuseConsecutiveTransposes>();
```
any specific reason?

onnx/common/stl_backports.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
```
bddppq: It's probably copied from caffe2? folly later added an additional android check.
Besides, I recommend don't flip the if-else from the original file, especially when there are several conditional checks.

smessmer: It was a slightly modified copy of the caffe2 implementation. I now switched to the one from folly.

onnx/optimizer/optimize.h
smessmer: nope. Thanks for catching.

onnx/common/stl_backports.h
```diff
@@ -2,8 +2,62 @@
...
+
+// std::make_unique is available
+
+#else
+
+namespace std {
+
+template <typename T, typename... Args>
+typename std::enable_if<!std::is_array<T>::value, std::unique_ptr<T>>::type
+make_unique(Args&&... args) {
```
bddppq: Let's put it inside `ONNX_NAMESPACE` (In case of either our detection macros is not robust enough or there is another library does the same trick to pollute std namespace, there will be multiple definitions of std::make_unique)

PR #613, opened 2018-03-14
Minor code quality improvements in defs/

smessmer:
Minor improvements to readability and performance, for example moving things instead of copying them.

I'm a bit uncertain about the move-initialized lambda captures. That's technically a C++14 feature. Do all our compilers support this? Would I notice if they don't, i.e. do we check old compilers in CI?


yinghai:
> That's technically a C++14 feature. Do all our compilers support this? 

We now include ONNX as part of Caffe2 and Caffe2 only supports up to C++11...

smessmer:
What guarantees do we give exactly? Is it "Caffe2 works on all compilers that support C++11"? Or is it "Caffe2 works on GCC since version x and Clang since version y"? In the latter case, we might be able to use some C++14 features since most old compiler versions support a subset of it.

bddppq:
ONNX has the same requirement/guarantee of using c++11

smessmer:
ok I re-established c++11 compatibility

PR #615, opened 2018-03-14
Add ReshapeDynamic

anderspapitto:


houseroad:
@anderspapitto could you also provide node test for ReshapeDynamic?

anderspapitto:
added tests

linkerzhang:
I'd suggest that we make a breaking change against existing Reshape OP (Versioning was designed for this purpose :) ).

PR #616, opened 2018-03-15
Add 8bit quantization support

hlu1:
Add scale/zero_point to ValueInfoProto
Add schemas for Quantize/Dequantize
Add tensor(uint8) to TypeConstraint to operators in math, nn, reduction, and tensor
Note the bias type is int32 for Conv and ConvTranspose when the input and weight are uint8

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -75,8 +75,8 @@ opset_import {
 #### Type Constraints
 
 <dl>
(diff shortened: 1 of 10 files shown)


Maratyszcza:
cc @ajtulloch, @jspark1105

linkerzhang:
Thank you very much @hlu1 for the PR. let's discuss and understand more about quantization cases here. https://gitter.im/onnx/quantization?utm_source=share-link&utm_medium=link&utm_campaign=share-link

Issue #595, opened 2018-03-08
AllowConsumed and EnforceConsumed would not be part of op definition?


linkerzhang:
AllowConsumed and EnforceConsumed are in the scope of implementation optimization. It would be better to have backend to decide how to handle it. Thoughts? @ezyang @dzhulgakov @ebarsoum @gramalingam

gramalingam:
I agree. These should be removed from the operator registration mechanism since we are not talking about any specific implementation of an operator.

bddppq:
Removing these sounds reasonable to me

houseroad:
https://github.com/onnx/models/issues/26

houseroad:
It's already removed. :-)

PR #617, opened 2018-03-15
Remove allowconsumed enforceconsumed from op schema.

linkerzhang:
This is for issue #595 .


bddppq:
Here also needs to be removed:
 https://github.com/linkerzhang/onnx/blob/e0ffbea9d46007ea75cba1371c03a5715207d9f1/onnx/defs/schema.cc#L161

bddppq:
@houseroad Models in model zoo also have `consumed_inputs` attributes

linkerzhang:
@bddppq Thank you for the comments! Removed the piece of codes. As you said, models in model zoo are using the consumed_inputs attribute, which is being broken.

houseroad:
Let's fix the model zoo first, will do it soon :-)

Tracking here: https://github.com/onnx/models/issues/26

linkerzhang:
onnx-fb-universe error:
E   AssertionError: 'ir_version: 2\nproducer_name: "pytorch"\nproducer_version: "0.3"\ngraph {\n  no [truncated]... != 'ir_version: 3\nproducer_name: "pytorch"\nproducer_version: "0.3"\ngraph {\n  no [truncated]...
00:43:49 E   - ir_version: 2
00:43:49 E   ?             ^
00:43:49 E   + ir_version: 3
00:43:49 E   ?  

@houseroad I think the error is not related with this change, right?

bddppq:
@onnxbot retest this please

onnx/defs/schema.h
```diff
@@ -506,7 +483,7 @@ class OpSchemaRegistry final {
       // Increase the highest version when you make BC-breaking changes to the
       // operator schema on specific domain. Update the lowest version when it's
       // determined to remove too old version history.
-      map_[ONNX_DOMAIN] = std::make_pair(1, 6);
+      map_[ONNX_DOMAIN] = std::make_pair(1, 7);
```
bddppq: It has been bumped from 5 to 6 for this purpose https://github.com/onnx/onnx/commit/5f69c37628002efa8c03d70d652db03c9d5ffca7

houseroad: Wait, as we discussed, we just keep using 6 right?

onnx/defs/math/defs.cc
```diff
@@ -297,22 +347,30 @@ produces one output data (Tensor<T>) where the function `f(x) = x^exponent`,
...
     .Output(0, "Z", "Output tensor (same size as X)", "T")
-    .TypeConstraint("T", { "tensor(float16)", "tensor(float)", "tensor(double)" },
+    .TypeConstraint(
+        "T",
+        {"tensor(float16)", "tensor(float)", "tensor(double)"},
         "Constrain input and output types to float tensors.");
 
 ONNX_OPERATOR_SCHEMA(PRelu)
-    .AllowConsumed({{0, 0}})
+    .SinceVersion(7)
```
houseroad: As we discussed, this should be .SinceVersion(6), (because in my previous PR, we already bump up the opset version, no need to do it again, otherwise, I have to prepare opset_7 model zoos, which does not exist yet... why our test case does not fail on this?

onnx/defs/schema.h
linkerzhang: technically, it's risky to reuse 6. as in the time in-between, there could be models using the other operators with consumed_input with version 6. :) I'm changing to use version 6 though.

linkerzhang:
Thank you very much @bddppq  and @houseroad  a lot for reviewing the PR and helping on make it happen. All comments are resolved and I'm merging it. Let me know if there're more comments please.

PR #623, opened 2018-03-17
Add preprocessor flag to strip docs from ONNX schema

jamesr66a:
Human-readable doc strings occupy binary space, and this can become problematic in memory-constrained cases where they are not needed, such as mobile deployments. This PR adds a preprocessor flag, `ONNX_STRIP_DOCS`, which will elide uses of documentation string literals, allowing a link-time optimization pass to eliminate them

cc @Maratyszcza please try this out and tell me if this has a considerable effect on binary size

== Preliminary results ==
Building ONNX with `-O3 -flto -DONNX_STRIP_DOCS` I look at the generated `onnx_cpp2py_export.cpython-36m-darwin.so` on my Mac. I extract the strings from that SO and compare sizes with and without having the strip flag set:

```
-rw-r--r--  1 jamesreed  1876110778    22K Mar 17 12:06 /Users/jamesreed/strings_stripped.txt
-rw-r--r--  1 jamesreed  1876110778    75K Mar 17 12:07 /Users/jamesreed/strings_unstripped.txt
```

Note that `-flto` is required for the toolchain to remove string literals from the compiled binary in this case

### DIFF
--- a/onnx/defs/logical/defs.cc
+++ b/onnx/defs/logical/defs.cc
@@ -9,6 +9,7 @@ namespace ONNX_NAMESPACE {
 
 std::function<void(OpSchema&)> BinaryLogicDocGenerator(const char* name) {
     return [=](OpSchema& schema) {
+#ifndef ONNX_STRIP_DOCS
         std::string doc = R"DOC(
 Returns the tensor resulted from performing the `{name}` logical operation
 elementwise on the input tensors `A` and `B`.
@@ -18,6 +19,9 @@ to match the shape of left-hand-side argument. See the doc of `Add` for a
 detailed description of the broadcasting rules.
 )DOC";
         ReplaceAll(doc, "{name}", name);
+#else
+        std::string doc = "";
+#endif
         schema.SetDoc(doc.c_str());
         schema.Attr("broadcast", "Enable broadcasting", AttributeProto::INT, static_cast<int64_t>(0));
         schema.Attr("axis", "If set, defines the broadcast dimensions.",
--- a/onnx/defs/logical/defs.cc
+++ b/onnx/defs/logical/defs.cc
@@ -18,7 +18,7 @@ to match the shape of left-hand-side argument. See the doc of `Add` for a
 detailed description of the broadcasting rules.
 )DOC";
         ReplaceAll(doc, "{name}", name);
-        schema.SetDoc(doc);
+        schema.SetDoc(doc.c_str());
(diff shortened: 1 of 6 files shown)


Maratyszcza:
@hlu1 did the size measurements on FB use-cases

smessmer:
I think it would be better if the macro doesn't have to be used on each operator definition. It could be hidden in the DSL instead.

bddppq:
I would prefer not to add these ifdefs since from our analysis doc strings only account for very small portion to the code size. @hlu1 has done a code size analysis for onnx:

```
                                           FILE SIZE
                                      --------------
 onnx/onnx/common/ir_pb_converter.cc   1.17Mi   6.3%
 onnx/onnx/defs/math/defs.cc           1.14Mi   6.1%
 onnx/onnx/common/interned_strings.cc  1.04Mi   5.6%
 onnx/onnx/checker.cc                  1.01Mi   5.4%
 [section .debug_loc]                   813Ki   4.3%
 onnx/onnx/defs/nn/defs.cc              797Ki   4.2%
 onnx/onnx/optimizer/optimize.cc        735Ki   3.9%
 onnx/onnx/defs/experiments/defs.cc     718Ki   3.8%
 onnx/onnx/defs/schema.cc               682Ki   3.6%
 onnx/onnx/defs/tensor/defs.cc          645Ki   3.4%
...
```

Considering we have <150 ops and assuming each op has <1000 chars doc strings, total number of them would be <150k, looks to me striping them won't help too much.

Maratyszcza:
@bddppq For mobile, 150K is a lot!

bddppq:
@Maratyszcza 150k is uncompressed, compare it with other entries in the size table

Maratyszcza:
Something isn't right with the table, it probably includes debug symbols

dzhulgakov:
I wonder if we can move setDoc as an inline function to the header file and then C++ would be smart enough to eliminate string literals even without flto

smessmer:
@dzhulgakov Yes, that's what I meant above with "hiding it in the DSL". The C++ stripper should be smart enough to do that, especially if we make the strings `constexpr` and keep them declare-only (i.e. don't provide a definition).

bddppq:
@jamesr66a @Maratyszcza Is this PR still needed?

bddppq:
@Maratyszcza For Caffe2 it has been changed to only link with onnx pb.{h,cc} files in mobile builds

Maratyszcza:
Ok, if schema is not included in mobile builds, it is safe do drop this PR

PR #628, opened 2018-03-19
Allow backend tests to set test tolerance on a per-case basis

smessmer:


smessmer:
Will add a PR to onnx-coreml shortly. The general syntax is:

    backend_test = onnx.backend.test.BackendTest(CoreMLBackend, __name__)
    backend_test.include('test_densenet121')
    backend_test.tolerance('test_densenet121', rtol=1e-3, atol=1e-7)

houseroad:
Could you add the information in data.json for each model? such as
https://github.com/onnx/onnx/blob/master/onnx/backend/test/data/model/real/test_bvlc_alexnet/data.json

And also the logic to parse this info

smessmer:
You mean the model should specify the tolerance? I thought it might be different for different backends, that's why I added it to the test definition instead of the model definition.

houseroad:
Yeah, I mean each model can provide a default tolerance. Of course, we can change that in ONNX backend test, just like your code did.

onnx/backend/test/runner/__init__.py
```diff
@@ -141,14 +146,14 @@ def tests(self):
         return tests
 
     @staticmethod
-    def _assert_similar_outputs(ref_outputs, outputs):
+    def _assert_similar_outputs(ref_outputs, outputs, rtol, atol):
```
houseroad: you probably want to set the default values, in case you have a lot node tests, do you really want to manually set them one by one. Check line 296

smessmer: The default values are still there, they just moved to the `_tolerance` function.

smessmer:
Having a default value per model makes sense. It's probably a separate stream of work though and can be done independently from this PR.

pk-g:
Thanks for the PR! this would be really helpful. LG in general, with a minor suggestion. Let's consider the example you provided for setting tolerance on densenet:

"backend_test = onnx.backend.test.BackendTest(CoreMLBackend, __name__)
 backend_test.include('test_densenet121')
 backend_test.tolerance('test_densenet121', rtol=1e-3, atol=1e-7)"

Assuming that we include a batch tests including above, wouldn't this setup override the global tolerance for other tests as well? would it be possible to add the tolerance upon inclusion only for the given case? (i.e. something like, backend_test.include('text_densenet121', rtol=1e-3, atol= 1e-7)?

smessmer:
@pk-g : Calling this `tolerance()` function will only set the tolerance for the one test you specify as its parameter.

onnx/backend/test/runner/__init__.py
```diff
@@ -257,10 +263,18 @@ def run(test_self, device):
...
-                self._assert_similar_outputs(ref_outputs, outputs)
+                self._assert_similar_outputs(ref_outputs, outputs, rtol, atol)
 
         self._add_test(kind + 'Model', model_test.name, run, model_marker)
 
+    def _tolerance(self, name):
+        for pattern, rtol, atol in self._tolerance_patterns:
+            if pattern.search(name):
+                return (rtol, atol)
+        rtol = 1e-3
```
bddppq: make these two constants configurable (and expose them in the class level).

onnx/backend/test/runner/__init__.py
bddppq: Maybe do `rtol=None` and `atol=None` and set them to the default value if not specified? So people don't have to override both of them.

onnx/backend/test/runner/__init__.py
```diff
@@ -257,10 +263,18 @@ def run(test_self, device):
                         tensor.ParseFromString(f.read())
                     ref_outputs.append(numpy_helper.to_array(tensor))
                 outputs = list(prepared_model.run(inputs))
-                self._assert_similar_outputs(ref_outputs, outputs)
+                self._assert_similar_outputs(ref_outputs, outputs, rtol, atol)
 
         self._add_test(kind + 'Model', model_test.name, run, model_marker)
 
+    def _tolerance(self, name):
+        for pattern, rtol, atol in self._tolerance_patterns:
```
bddppq: You can finalize these configurations in `tests_cases`, `test_suites` and `tests` functions, to avoid doing this search repeatedly inside the real tests (extend `Testitem` store the corresponding `rtol` and `atol` as metadata).

bddppq:
The current diff only allows configuring these tolerances for model tests, not node tests.

houseroad:
@smessmer like @bddppq and I mentioned, check line 296, you don't provide the rtol and atol for node_test. Please fix it. Here is the log: https://ci.pytorch.org/jenkins/job/caffe2-builds/job/py2-cuda8.0-cudnn6-ubuntu16.04-test/1701/console

dzhulgakov:
Didn't we want to do all node tests behave as model tests (doesn't really make sense to have this difference). And we certainly want to specify tolerance on test level in the onnx repo instead of making each backend copy-paste the same regex + tolerance override. If it's blocking - let's land this one but please follow up.

onnx/backend/test/runner/__init__.py
bddppq: I tag the wrong line, I meant in `def tolerance`, not `def _assert_similar_outputs`. But looks like you have done the right change :-)

bddppq:
@smessmer Are you still working on this one?

houseroad:
@Ac2zoom would you like to pick up this one?

PR #648, opened 2018-03-23
struct InferenceContext needs a virtual destructor

jamesr66a:


onnx/defs/schema.h
```diff
@@ -41,6 +41,7 @@ struct InferenceContext {
   virtual const TypeProto_Tensor* getInputType(size_t index) const = 0;
   virtual size_t getNumOutputTypes() const = 0;
   virtual TypeProto_Tensor* getOutputType(size_t index) = 0;
+  virtual ~InferenceContext() {}
```
houseroad: yeah, we need it :-)

PR #652, opened 2018-03-23
Fix optimizer split pass for models with constant output

bddppq:
e.g. A dummy graph that only has one Constant node


onnx/test/optimizer_test.py
```diff
@@ -117,6 +117,32 @@ def test_preserve_value_info(self):
...
+        )
+        graph = helper.make_graph(
+            [node],
+            'test-optimize-split',
+            [],
+            [helper.make_tensor_value_info('X', TensorProto.FLOAT, (1,))])
+
+        init_model = self._optimized(graph, ['split_init'])
+        self.assertEqual(len(init_model.graph.node), 1)
+        self.assertEqual(init_model.graph.node[0].op_type, 'Constant')
```
houseroad: Check output == 1?

Issue #632, opened 2018-03-20
Enhance shape inference


anderspapitto:
This task tracks improvements to shape inference which I intend to defer out of https://github.com/onnx/onnx/pull/564

```
I wonder whether we can have a simple wrapper that typecasts the attribute values to the right type. It'd make implementing functions much easier. Those helpers might be more isolated that ir.h and be just getAttr<vector<int>>("perm", optional_default) or something
- dmytro
```

```
shall we have some helper function for creating TensorProto_Tensor too? otherwise this mechanics of protobuf are too cumbersome to operate.
- dmytro
```

```
should it be merging of data? what if only type but not shape is specified in the valueInfoProto? in this case we should merge. Also after inference details might be more specified that they were before. Ideally we should check compatibility and override if necessary
- dmytro
```

```
In terms of testing, @houseroad is changing the backend tests to pass output value info to the backend in #588, this can be used to test the inference functions (create a dummy backend like here

onnx/onnx/test/test_backend_test.py

Line 24 in e1d3155

 class DummyBackend(onnx.backend.base.Backend): 
, in run_node do shape inference and assert they match the output value info and then skip the real test).
- junjie
```

```
 how the partial shape inference would work, and also what is going to be arithmetic on those values (e.g. the simplest case is that if any operation is done with symbolic value the new symbolic value is created)
- dmytro
```

- add many more shape inference functions

PR #655, opened 2018-03-23
Enhancements to shape inference

anderspapitto:
#632


anderspapitto:
closing temporarily so that my other PR runs sooner in appveyor

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include "data_type_utils.h"
+
+namespace ONNX_NAMESPACE {
+
+typedef TensorShapeProto_Dimension InferenceDimension;
+
+struct InferenceContext {
+  virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
+  virtual size_t getNumInputTypes() const = 0;
```
dzhulgakov: this is just getNumInputs?

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,48 @@
...
+      values.push_back(value);
+    }
+    return true;
+  } else {
+    return false;
+  }
+
+}
+
+inline void copyElemType(InferenceContext& ctx, size_t inputIndex, size_t outputIndex) {
```
dzhulgakov: put it into InferenceContext itself?

and maybe something like 'typeSameAsInput' or 'propagateTypeFromInputToOutput'

onnx/defs/tensor/defs.cc
```diff
@@ -242,37 +242,22 @@ will be (2, 1, 3).
...
         }
-      }
 
-      TypeProto_Tensor tt;
-      {
-        tt.set_elem_type(ctx.getInputType(0)->elem_type());
-        auto shape = tt.mutable_shape();
+        std::vector<int64_t> perm;
+        if (!getRepeatedAttribute(ctx, "perm", perm)) {
+          for (int i = ctx.getInputType(0)->shape().dim_size() - 1; i >= 0; i--) {
```
dzhulgakov: so what would be the shape if we don't even know the rank of some tensor? would you expect that shape won't be set at all in this case? Then we need to test for it somehow

onnx/defs/tensor/defs.cc
```diff
@@ -242,37 +242,22 @@ will be (2, 1, 3).
...
+            perm.push_back(i);
+          }
+        }
 
+        copyElemType(ctx, 0, 0);
         for (size_t i = 0; i < perm.size(); i++) {
-          auto dim = shape->add_dim();
-          *dim = ctx.getInputType(0)->shape().dim(perm[i]);
+          copySingleDimFrom(ctx, 0, 0, perm[i]);
```
dzhulgakov: it's not obvious that it appends to the shape tensor at all

onnx/defs/tensor/defs.cc
```diff
@@ -242,37 +242,22 @@ will be (2, 1, 3).
...
-        if (perm_attr) {
-          for (auto dim : perm_attr->ints()) {
-            perm.push_back(dim);
-          }
-        } else {
-          int ndims = ctx.getInputType(0)->shape().dim_size();
-          for (int i = 0; i < ndims; i++) {
-            perm.push_back(ndims - i - 1);
-          }
+        if (ctx.getNumInputTypes() != 1 || !ctx.getInputType(0)) {
```
dzhulgakov: should we just enforce that it has only 1 input?

onnx/shape_inference/shape_inference.h
```diff
@@ -77,22 +77,60 @@ void InferShapes(ModelProto& m) {
...
-
       for (int i = 0; i < n.output_size(); i++) {
         const auto& output = n.output(i);
-        if (ctx.getOutputType(i)->elem_type() == TensorProto::UNDEFINED) {
+        const auto& inferredType = *ctx.getOutputType(i);
+
+        // In this case, we have no new information, so don't bother
+        // to add a contentless ValueInfo.
+        if (inferredType.elem_type() == TensorProto::UNDEFINED &&
+            inferredType.shape().dim_size() == 0) {
```
dzhulgakov: you need to check that shape() is not set. Size zero shape basically means scalar

onnx/shape_inference/shape_inference.h
```diff
@@ -77,22 +77,60 @@ void InferShapes(ModelProto& m) {
...
           vi->set_name(output);
-          *vi->mutable_type()->mutable_tensor_type() = *ctx.getOutputType(i);
+          existingType = vi->mutable_type()->mutable_tensor_type();
+        }
+
+        // Incorporate the inferred information.
+        if (inferredType.elem_type() != TensorProto::UNDEFINED) {
+          if (existingType->elem_type() != TensorProto::UNDEFINED &&
+              existingType->elem_type() != inferredType.elem_type()) {
+            std::cerr << "Warning: inferred type differs from existing type" << std::endl;
```
dzhulgakov: hm, I'm curious whether we should throw in this case

onnx/shape_inference/shape_inference.h
```diff
@@ -77,22 +77,60 @@ void InferShapes(ModelProto& m) {
...
+          existingType = iter->second;
+        } else {
+          auto vi = g->add_value_info();
           vi->set_name(output);
-          *vi->mutable_type()->mutable_tensor_type() = *ctx.getOutputType(i);
+          existingType = vi->mutable_type()->mutable_tensor_type();
+        }
+
+        // Incorporate the inferred information.
+        if (inferredType.elem_type() != TensorProto::UNDEFINED) {
```
dzhulgakov: so elem_type actually defaults to FLOAT in protobuf. We should verify that the field is set

onnx/shape_inference/shape_inference.h
```diff
@@ -77,22 +77,60 @@ void InferShapes(ModelProto& m) {
...
+        if (inferredType.shape().dim_size() != 0) {
+          if (existingType->shape().dim_size() == 0) {
+            for (int j = 0; j < inferredType.shape().dim_size(); j++) {
+              existingType->mutable_shape()->add_dim();
+            }
+          }
+          for (int j = 0; j < inferredType.shape().dim_size(); j++) {
+            const auto& inferredDim = inferredType.shape().dim(j);
+            auto* existingDim = existingType->mutable_shape()->mutable_dim(j);
+            if (inferredDim.has_dim_value()) {
```
dzhulgakov: you don't handle dim_param here

onnx/defs/shape_inference.h
anderspapitto: @linkerzhang are you ok with adding helpers as methods of InferenceContext, or do you prefer that I leave them separate?

onnx/defs/tensor/defs.cc
anderspapitto: this is implementing "By default, reverse the dimensions" from the ONNX spec for transpose. I can add a comment

onnx/shape_inference/shape_inference.h
anderspapitto: ++

onnx/shape_inference/shape_inference.h
anderspapitto: why/where does it default to FLOAT?

onnx/defs/shape_inference.h
anderspapitto: yes

onnx/defs/tensor/defs.cc
anderspapitto: seems like more the role of the checker

onnx/defs/tensor/defs.cc
anderspapitto: sure, can rename

onnx/shape_inference/shape_inference.h
anderspapitto: yeah, I guess we might as well throw here, and convert it to a warning message in the main loop

onnx/shape_inference/shape_inference.h
anderspapitto: yes, I intentionally left that out for now. If I infer a shape of `(3, x, 10)`, it's not quite clear what to do with the `x`, so I just leave the second dimension alone

onnx/defs/tensor/defs.cc
anderspapitto: ah I see what you mean. I guess there are two cases

- there is no type information at all (handled at the top of the function). This represents unknown rank

- dim_size() == 0. This would represent a scalar tensor, and this code should handle that fine (it puts nothing in the output type dimensions). I'll add a test for this case

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include "data_type_utils.h"
+
+namespace ONNX_NAMESPACE {
+
+typedef TensorShapeProto_Dimension InferenceDimension;
+
+struct InferenceContext {
```
linkerzhang: Please kindly keep this interface in the schema.h, so that it's the only thing that other partners (say, me) will need to depend on.

onnx/shape_inference/shape_inference.h
```diff
@@ -34,15 +34,15 @@ struct InferenceContextImpl : public InferenceContext {
 void InferShapes(ModelProto& m) {
   auto* g = m.mutable_graph();
 
-  std::unordered_map<std::string, TypeProto_Tensor> valueTypesByName;
```
linkerzhang: can we rename this file please? I'm seeing two files with name "shape_inference.h".

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,139 @@
...
+  for (auto& vi : *g->mutable_input()) {
+    valueTypesByName[vi.name()] = vi.mutable_type()->mutable_tensor_type();
+  }
+  for (auto& vi : *g->mutable_output()) {
+    valueTypesByName[vi.name()] = vi.mutable_type()->mutable_tensor_type();
+  }
+
+  for (const auto& n : g->node()) {
+    auto schema = OpSchemaRegistry::Schema(n.op_type());
+    if (schema) {
```
bddppq: nit: Could you change it to "if (!schema) continue;" to avoid one extra level of indentation?

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,139 @@
...
+      InferenceContextImpl ctx;
+
+      {
+        std::unordered_map<std::string, const AttributeProto *> attributesByName;
```
bddppq: Doesn't look like we need this extra tmp variable.

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,139 @@
...
+            allInputTypes.push_back(iter->second);
+          } else {
+            allInputTypes.push_back(nullptr);
+          }
+        }
+
+        ctx.allInputTypes_ = std::move(allInputTypes);
+      }
+
+      ctx.allOutputTypes_.resize(n.output_size());
```
bddppq: From line 51 to here is to construct a context from a graph, maybe put these into the constructor InferenceContext?

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,139 @@
...
+        if (inferredType.elem_type() != TensorProto::UNDEFINED) {
+          if (existingType->elem_type() != TensorProto::UNDEFINED &&
+              existingType->elem_type() != inferredType.elem_type()) {
+            throw std::runtime_error("inferred type differs from existing type");
+          } else {
+            existingType->set_elem_type(inferredType.elem_type());
+          }
+        }
+        if (inferredType.shape().dim_size() != 0) {
+          if (existingType->shape().dim_size() == 0) {
```
bddppq: scalar's dim_size is 0, so we can not use it as the null check

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,139 @@
...
+        const auto& inferredType = *ctx.getOutputType(i);
+
+        // In this case, we have no new information, so don't bother
+        // to add a contentless ValueInfo.
+        if (inferredType.elem_type() == TensorProto::UNDEFINED &&
+            !inferredType.has_shape()) {
+          continue;
+        }
+
+        // If there is already a ValueInfo associated with this
```
bddppq: How about let's change the code starts from here to the end into:
```
auto iter = valueTypesByName.find(output);
if (iter != valueTypesByName.end()) {
   assert_equal_value_info(iter->second, inferredType);
} else {
   valueTypesByName[output] = inferredType;
}
```
(of course will need to add a separate function `assert_equal_value_info` dedicated for doing the equality check).

onnx/defs/schema.h
```diff
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include "data_type_utils.h"
+#include "shape_inference.h"
```
bddppq: nit: change this to absolute include

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include "data_type_utils.h"
+
+namespace ONNX_NAMESPACE {
+
+typedef TensorShapeProto_Dimension InferenceDimension;
```
bddppq: Do we still need this typedef?

onnx/defs/tensor/defs.cc
```diff
@@ -242,37 +242,25 @@ will be (2, 1, 3).
...
-        if (perm_attr) {
-          for (auto dim : perm_attr->ints()) {
-            perm.push_back(dim);
-          }
-        } else {
-          int ndims = ctx.getInputType(0)->shape().dim_size();
-          for (int i = 0; i < ndims; i++) {
-            perm.push_back(ndims - i - 1);
-          }
+        if (ctx.getNumInputTypes() != 1) {
```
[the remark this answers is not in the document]
anderspapitto: even if it has 1 input, that input might not have a ValueInfo

onnx/shape_inference/implementation.h
anderspapitto: sure

onnx/shape_inference/implementation.h
anderspapitto: ++

onnx/shape_inference/implementation.h
anderspapitto: I already have these changes in a followup diff https://github.com/onnx/onnx/pull/702/files

no reason to pull them into this one

onnx/defs/schema.h
anderspapitto: ++

onnx/shape_inference/implementation.h
anderspapitto: actually this is also in https://github.com/onnx/onnx/pull/702/files, don't want to pull it in here

onnx/defs/schema.h
anderspapitto: ++

onnx/defs/shape_inference.h
anderspapitto: nope, will remove

onnx/defs/tensor/defs.cc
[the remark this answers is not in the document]
anderspapitto: actually, yeah can take it out

onnx/shape_inference/implementation.h
anderspapitto: ++

onnx/shape_inference/implementation.h
anderspapitto: it's not an equality check though. There can be an existing value info with partial information, which needs to be merged in (e.g. one of the dims is "x3", and we infer it to be 5

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,46 @@
...
+  virtual ~InferenceContext() {}
+};
+
+typedef void (*InferenceFunction)(InferenceContext&);
+
+template<typename T> inline google::protobuf::RepeatedField<T> retrieveValues(const AttributeProto* attr);
+template<> inline google::protobuf::RepeatedField<int64_t> retrieveValues(const AttributeProto* attr) { return attr->ints(); }
+
+template<typename T>
+inline bool getRepeatedAttribute(InferenceContext& ctx,
```
dzhulgakov: maybe make it part of the InferenceContext for consistency?

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,46 @@
...
+      values.push_back(value);
+    }
+    return true;
+  } else {
+    return false;
+  }
+
+}
+
+inline void propagateElemTypeFromInputToOutput(InferenceContext& ctx, size_t inputIndex, size_t outputIndex) {
```
dzhulgakov: same?

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
+
+#include "data_type_utils.h"
+
+namespace ONNX_NAMESPACE {
+
+struct InferenceContext {
+  virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
+  virtual size_t getNumInputTypes() const = 0;
```
dzhulgakov: actually rename it to numInputs/getNumInputs? "Types" part is confusing

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,135 @@
...
+          auto vi = g->add_value_info();
+          vi->set_name(output);
+          existingType = vi->mutable_type()->mutable_tensor_type();
+        }
+
+        // Incorporate the inferred information.
+        if (inferredType.elem_type() != TensorProto::UNDEFINED) {
+          if (existingType->elem_type() != TensorProto::UNDEFINED &&
+              existingType->elem_type() != inferredType.elem_type()) {
+            throw std::runtime_error("inferred type differs from existing type");
```
dzhulgakov: more details in the error message?

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,135 @@
...
+          } else {
+            existingType->set_elem_type(inferredType.elem_type());
+          }
+        }
+        for (int j = 0; j < inferredType.shape().dim_size(); j++) {
```
dzhulgakov: hm, I think you actually need to error out if the sizes are different. I.e. if ranks conflict

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,135 @@
...
+        TypeProto_Tensor* existingType = nullptr;
+        if (iter != valueTypesByName.end()) {
+          existingType = iter->second;
+        } else {
+          auto vi = g->add_value_info();
+          vi->set_name(output);
+          existingType = vi->mutable_type()->mutable_tensor_type();
+        }
+
+        // Incorporate the inferred information.
```
dzhulgakov: maybe extract all this comparison semantics into a separate function?

onnx/defs/shape_inference.h
anderspapitto: I will do this iff @linkerzhang is ok with it

onnx/defs/shape_inference.h
anderspapitto: sure

onnx/shape_inference/implementation.h
anderspapitto: ++

onnx/shape_inference/implementation.h
anderspapitto: hmm

onnx/shape_inference/implementation.h
anderspapitto: yeah, i guess I am working around some bugs in make_tensor_value_info() on the python side - it still uses empty dimensions to represent a missing shape. I'll fix that code and then I can just use has_shape() here

anderspapitto:
@linkerzhang are you ok with me adding the additional utility methods to InferenceContext as requested by @dzhulgakov, or would you prefer for me to keep them separate?

onnx/shape_inference/implementation.h
bddppq: oh right, then how about
```
auto iter = valueTypesByName.find(output);
if (iter != valueTypesByName.end()) {
   assert_loose_equal_value_info(iter->second, inferredType);
}
iter->second = inferredType;
```

anderspapitto: well that also doesn't quite work - the inferredType may also lack information which the existing ValueInfo already contains.

I'll pull the logic out in some fashion though

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,46 @@
...
+  virtual const TypeProto_Tensor* getInputType(size_t index) const = 0;
+  virtual size_t getNumOutputs() const = 0;
+  virtual TypeProto_Tensor* getOutputType(size_t index) = 0;
+  virtual ~InferenceContext() {}
+};
+
+typedef void (*InferenceFunction)(InferenceContext&);
+
+template<typename T> inline google::protobuf::RepeatedField<T> retrieveValues(const AttributeProto* attr);
+template<> inline google::protobuf::RepeatedField<int64_t> retrieveValues(const AttributeProto* attr) { return attr->ints(); }
```
bddppq: const AttributeProto&

onnx/test/shape_inference_test.py
```diff
@@ -42,57 +42,73 @@ def test_transpose_preexisting(self):
...
             "test",
             [helper.make_tensor_value_info("X", TensorProto.FLOAT, (2, 3, 4))],
             [])
         graph.value_info.extend([
-            helper.make_tensor_value_info("Y", TensorProto.STRING, (3, 2, 5))
+            helper.make_tensor_value_info("Y", TensorProto.FLOAT, (5, 5, 5))
         ])
 
-        inferred_model = self._inferred(graph)
+        try:
```
bddppq: `self.assertRaises`

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,138 @@
+#pragma once
+
+#include "onnx/proto_utils.h"
+#include "onnx/defs/schema.h"
+
+namespace ONNX_NAMESPACE { namespace shape_inference {
+
+struct InferenceContextImpl : public InferenceContext {
+  InferenceContextImpl(const NodeProto & n,
+                       std::unordered_map<std::string, TypeProto_Tensor *>& valueTypesByName) {
```
bddppq: const

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,138 @@
...
+    valueTypesByName[vi.name()] = vi.mutable_type()->mutable_tensor_type();
+  }
+  for (auto& vi : *g->mutable_input()) {
+    valueTypesByName[vi.name()] = vi.mutable_type()->mutable_tensor_type();
+  }
+  for (auto& vi : *g->mutable_output()) {
+    valueTypesByName[vi.name()] = vi.mutable_type()->mutable_tensor_type();
+  }
+
+  std::unordered_map<std::string, int> dimParamValues;
```
bddppq: this doesn't seem to be needed

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,138 @@
...
+
+  for (const auto& n : g->node()) {
+    auto schema = OpSchemaRegistry::Schema(n.op_type());
```
bddppq: You need to pass the opset_version to get the correct schema.

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,138 @@
...
+    if (!schema) {
+      continue;
+    }
+
+    InferenceContextImpl ctx(n, valueTypesByName);
+
+    schema->GetShapeInferenceFunction()(ctx);
+
+    for (int i = 0; i < n.output_size(); i++) {
```
bddppq: ++i

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,138 @@
...
+      if (inferredType.elem_type() != TensorProto::UNDEFINED) {
+        if (existingType->elem_type() != TensorProto::UNDEFINED &&
+            existingType->elem_type() != inferredType.elem_type()) {
+          throw std::runtime_error("inferred type differs from existing type");
+        } else {
+          existingType->set_elem_type(inferredType.elem_type());
+        }
+      }
+
+      for (int j = 0; j < inferredType.shape().dim_size(); j++) {
```
bddppq: ++j

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,138 @@
...
+        const auto& inferredDim = inferredType.shape().dim(j);
+        if (existingType->mutable_shape()->dim_size() <= j) {
```
bddppq: Here should error out if `existingType` is existing ValueInfo in the graph

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,138 @@
...
+          existingType->mutable_shape()->add_dim();
+        }
+        auto* existingDim = existingType->mutable_shape()->mutable_dim(j);
+        if (inferredDim.has_dim_value()) {
+          auto inferredDimValue = inferredDim.dim_value();
+          if (existingDim->has_dim_value() && existingDim->dim_value() != inferredDimValue) {
+            throw std::runtime_error("inferred dimension differs from existing dimension");
+          }
+          if (existingDim->has_dim_param()) {
```
bddppq: remove this clause

onnx/defs/shape_inference.h
linkerzhang: it seems to me that they're helper functions, let's keep them outside of the interface for now. Thank you very much!

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,46 @@
...
+  virtual const TypeProto_Tensor* getInputType(size_t index) const = 0;
+  virtual size_t getNumOutputs() const = 0;
+  virtual TypeProto_Tensor* getOutputType(size_t index) = 0;
+  virtual ~InferenceContext() {}
+};
+
+typedef void (*InferenceFunction)(InferenceContext&);
+
+template<typename T> inline google::protobuf::RepeatedField<T> retrieveValues(const AttributeProto& attr);
+template<> inline google::protobuf::RepeatedField<int64_t> retrieveValues(const AttributeProto& attr) { return attr.ints(); }
```
bddppq: These should be moved to proto_utils.{h,cc}. (Can be done is separate PR).

onnx/defs/shape_inference.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
+
+#include "data_type_utils.h"
```
bddppq: nit: absolute include

onnx/defs/shape_inference.h
bddppq: Why not putting these into the InferenceContext class? They are tightly tied to InferenceContext and there is no way to use them.
@linkerzhang

onnx/defs/tensor/defs.cc
```diff
@@ -242,37 +242,22 @@ will be (2, 1, 3).
...
+            perm.push_back(i);
+          }
         }
-      }
 
-      *ctx.getOutputType(0) = tt;
-    });
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+        for (size_t i = 0; i < perm.size(); ++i) {
+          appendSingleDimCopiedFromInputTypeToOutputType(ctx, 0, 0, perm[i]);
```
bddppq: So in case `perm` attribute is provided, and shape is not set, this will cause segfault.

onnx/defs/shape_inference.h
anderspapitto: ++

onnx/defs/shape_inference.h
anderspapitto: ++

onnx/defs/shape_inference.h
anderspapitto: I think either way is acceptable. @linkerzhang has asked that we keep them out of InferenceContext though (I guess so the API exposed to 3rd parties is smaller), so I am keeping it separate

onnx/defs/tensor/defs.cc

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,150 @@
...
+}
+
+void InferShapes(ModelProto& m) {
+  checker::CheckerContext cctx;
+  cctx.set_ir_version(static_cast<int>(m.ir_version()));
+  std::unordered_map<std::string, int> opset_imports;
+  for (const auto& opset_import : m.opset_import()) {
+    opset_imports[opset_import.domain()] = static_cast<int>(opset_import.version());
+  }
+  cctx.set_opset_imports(opset_imports);
```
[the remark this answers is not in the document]
anderspapitto: ++ I'll extract the relevant bits of CheckerContext

onnx/proto_utils.h
```diff
@@ -32,4 +32,8 @@ bool ParseProtoFromBytes(Proto* proto, const char* buffer, size_t length) {
   coded_stream.SetTotalBytesLimit((2048LL << 20) - 1, 512LL << 20);
   return proto->ParseFromCodedStream(&coded_stream);
 }
+
+template<typename T> inline google::protobuf::RepeatedField<T> retrieveValues(const AttributeProto& attr);
```
bddppq: Upper case the function name so it's consistent with other functions in this file.

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,145 @@
...
+
+  for (const auto& n : g->node()) {
+    // Resolve domain for node
+    auto dit = opset_imports.find(n.domain());
+    if (dit == opset_imports.end()) {
+      continue;
+    }
+    auto domain_version = dit->second;
+
+    auto schema = OpSchemaRegistry::Schema(n.op_type(), domain_version, n.domain());
```
bddppq: nit: const

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,145 @@
...
+      if (inferredType.elem_type() != TensorProto::UNDEFINED) {
+        if (existingType->elem_type() != TensorProto::UNDEFINED &&
+            existingType->elem_type() != inferredType.elem_type()) {
+          throw std::runtime_error("inferred type differs from existing type");
+        } else {
+          existingType->set_elem_type(inferredType.elem_type());
+        }
+      }
+
+      for (int j = 0; j < inferredType.shape().dim_size(); ++j) {
```
bddppq: If existingType has multiple dims but the inferredType has zero dim, then this for loop will be skipped and so we won't be able to detect the mismatch.

onnx/shape_inference/implementation.h
```diff
@@ -0,0 +1,145 @@
...
+        const auto& inferredDim = inferredType.shape().dim(j);
+        if (existingType->mutable_shape()->dim_size() <= j) {
```
bddppq: This doesn't need to call "mutable"_shape(), in general we should be careful about calling it because it can change the semantics of whether the shape is not set or it's a scalar.

PR #662, opened 2018-03-26
[for discussion] Add UnspecifiedDimension and clarify meaning of symbolic dims

anderspapitto:
This makes it clear that if there are two places in the graph that
both have a symbolic dimension of "x", during any given graph
execution, they must take on the same concrete value. For sizes not
intended to correlate in this way, we introduce UnspecifiedDimension.


onnx/onnx.in.proto
```diff
@@ -370,14 +374,19 @@ message TensorProto {
...
-// Defines a tensor shape. A dimension can be either an integer value
-// or a symbolic variable. A symbolic variable represents an unknown
-// dimension.
+// Defines a tensor shape. A dimension can be
+//   - an integer value
+//   - a symbolic variable. All symbolic dimensions in the graph with
+//     the same name are constrained to take on the same value.
+//   - unspecified. Unspecified dimensions have no constraints on
+//     them. They are equivalent to symbolic dimensions with unique,
+//     autogenerated names.
```
linkerzhang: I'd keep current design to use unique symbolic name in this case as implied here, other than having another flag (enum).

dzhulgakov:
Maybe instead of having a flag we can have a `dim_param` with empty string? Basically meaning "not filled in"

anderspapitto:
@linkerzhang would you be ok with using the empty string name as @dzhulgakov suggests?

All unique names works, but it is an extra burden on anyone generating onnx models (have to have a supply of unique names), and also might make them less readable (you have to look at the whole model to find if a name is unique, instead of just seeing that it is UNSPECIFIED or "")

houseroad:
I agree with @linkerzhang and @dzhulgakov, adding another flag to the Dimension is too much. And we should avoid changing the proto files if not necessary. We should use a special symbolic variable (name) to represent the unspecified dimension, "" sounds good to me, or you can choose another one.

PR #666, opened 2018-03-27
Refine a little bit about op spec.

linkerzhang:


linkerzhang:
@bddppq In this PR, I changed "axes" for Reduce ops to be required. It seems that it breaks some model verification in onnx-fb-universe. What's the default value in caffe2 given it's not a required attribute please? Thank you!

bddppq:
@linkerzhang it's not about default values but our pytorch exporter doesn't add the `axes` attribute when exporting `prod`, and this is because this PR has changed the schema but not bumped the version.

houseroad:
In Caffe2, if axes is empty, we reduce along all the dimensions. This behavior is consist with other ops in onnx spec, such as slice.

I also checked numpy and tensorflow, they have the same definition on empty axes  list. So we should not give the 0 as default value.

linkerzhang:
@houseroad / @bddppq , ok, I'm keeping axes as OPTIONAL as it was before.

onnx/defs/reduction/defs.cc
```diff
@@ -1,99 +1,101 @@
...
+        "keepdims",
+        "Keep the reduced dimension or not, default 1 mean keep reduced dimension.",
+        AttributeProto::INT,
+        static_cast<int64_t>(1));
+    schema.Input(0, "data", "An input tensor.", "T");
+    schema.Output(
+        0,
+        "reduced",
+        "Reduced output tensor with integer data type.",
+        "tensor(uint64)");
```
houseroad: shall we use int64 here to be consistent with topk?

linkerzhang: I think, we probably should change topK's indices (2nd output) with type "uint64" too, since there's no negative index. :)

onnx/defs/schema.h
```diff
@@ -348,19 +348,20 @@ class OpSchema final {
...
-                                                                "int8",
-                                                                "uint16",
-                                                                "int16",
-                                                                "int64",
-                                                                "float16",
-                                                                "double"};
-    return all_integral_types;
+  static const std::vector<std::string>& all_numeric_types() {
+    static const std::vector<std::string> all_numeric_types = {
+        "tensor(float)",
```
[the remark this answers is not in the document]
linkerzhang: sure.

onnx/defs/reduction/defs.cc
bddppq: I had some offline discussions with @dzhulgakov, we think it's better to use int64 to represent indices:

1. some system doesn't support uint64 (e.g. Java)
2. in case error happened, say -1 is returned, it's easier to discover when it's -1 instead of wraparound

houseroad: Another data point is from tensorflow, they also only support signed int for the indices.

For example, in their argmax/argmin, the return values (indices) are either int or int64. I think signed int is more expressive when we need to return some special values/errors.

onnx/defs/math/defs.cc
```diff
@@ -138,8 +138,14 @@ the tensor elementwise.
     .Output(0, "Y", "Output tensor", "T")
     .TypeConstraint(
         "T",
-        {"tensor(float16)", "tensor(float)", "tensor(double)"},
-        "Constrain input and output types to float tensors.");
+        {"tensor(float)",
```
bddppq: nit: sort these types :-)

linkerzhang:
@bddppq ,I'd suggest doing that. @ebarsoum will talk with folks in next week's sync meeting on this proposal.

PR #668, opened 2018-03-27
Liqun/reduceop

liqunfu:


houseroad:
@liqunfu could you keep only one pr for the same content?

PR #676, opened 2018-03-28
[WIP] Setup mypy type checker

smessmer:
**This is WIP, please ignore this pull request.**
It's just here because I couldn't repro the CI issue of #607 locally ;)


CMakeLists.txt
```diff
@@ -154,6 +154,10 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP SRCS HDRS ROOT_DIR)
       VERBATIM )
 
     set(PROTOC_ARGS ${GENERATED_PROTO} -I ${OUTPUT_PROTO_DIR} --cpp_out ${ONNX_DLLEXPORT_STR}${OUTPUT_PROTO_DIR})
+    if(NOT WIN32)
```
bddppq: Could you put it in the BUILD_PYTHON block below? Also the output files should be put in `${ONNX_DLLEXPORT_STR}${OUTPUT_PROTO_DIR}` like other autogenerated files and then setup.py copy them to the corresponding path.

PR #677, opened 2018-03-29
Fix Average pool test cases converted from PyTorch

houseroad:
Fix the wrong padding attributes in AveragePool cases.


bddppq:
Do you have a script to generate these files? Maybe also put it into this repo?

houseroad:
It's here: https://github.com/onnxbot/onnx-fb-universe/blob/master/test/convert_pytorch_test.py

Since we have dependencies on pytorch and caffe2... probably not appropriate to add to onnx repo directly. :-)

houseroad:
@spidyDev I just updated the files, could you check again?

spidyDev:
@houseroad  AvgPool3d tests LGTM.
For: AvgPool1d (onnx/backend/test/data/pytorch-converted/test_AvgPool1d/model.onnx)
This test seems to be testing avgpool2d instead of avgpool1d.
I  think we dont need  (unsqueeze-squeeze) , also the avgpool op can have kernel_shape , pad and stride corresponding to 1d tensor i.e. (2) , (0,0) , (2) respectively.

houseroad:
@spidyDev thanks for the check. Actually those tests are automatically converted from PyTorch tests. In PyTorch, the AvgPool1d is implemented as unsqueeze + avgpool2d + squeeze. That's why we have such models. I think such models are useful since they also cover other ops, such as Unsqueeze and Squeeze. We should also include them in the test data. :-)

spidyDev:
@houseroad  Thanks for clarifying. PR LGTM.

PR #678, opened 2018-03-29
external_data: Store large tensor values in separate files

postrational:
**Updated after ONNX Partners workshop November 2018.**

### Motivation

ONNX is designed to support the exchange of machine learning models which can grow quite large. Model data can easily exceed 2 GB, the maximum size of a Protobuf message. We need a method for optionally storing tensor data in separate files, which can be loaded on demand.

### `data_location` and `external_data` fields

We propose to add two fields to the `TensorProto` message type.

    repeated StringStringEntryProto external_data = 13;
    optional DataLocation data_location = 14;

Where `DataLocation` is a new `enum`:

      enum DataLocation {
        MESSAGE = 0;
        RAW = 1;
        EXTERNAL = 2;
      }

#### `data_location` field

`data_location` field stores the location of data for this tensor. Value MUST be one of:
- `MESSAGE` - data stored in type-specific fields inside the protobuf message.
- `RAW` - data stored in `raw_data` field.
- `EXTERNAL` - data stored in an external location as described by `external_data` field.
- value not set - legacy value. Assume data is stored in `raw_data` (if set) otherwise in message.

#### `external_data` field

`external_data` field stores key-value pairs of strings describing data location. 

Recognized keys are:
- `"location"` (required) - file path relative to the filesystem directory where the ONNX protobuf model was stored. Up-directory path components such as `..` are disallowed and should be stripped when parsing.
- `"offset"` (optional) - position of byte at which stored data begins. Integer stored as string. Offset values SHOULD be multiples 4096 (page size) to enable mmap support.
- `"length"` (optional) - number of bytes containing data. Integer stored as string.
- `"checksum"` (optional) - SHA1 digest of file specified in under 'location' key.

After an ONNX file is loaded, all `external_data` fields may be updated with an additional key (`"basepath"`), which stores the path to the directory from which he ONNX model file was loaded.

#### External data files

Data stored in external data files will be in the same binary bytes string format as is used by the `raw_data` field in current ONNX implementations.


jamesr66a:
Hi @postrational! 

Thanks for the PR and proposal. We've actually been working on a similar proposal this week that treats the `zip` file format as a first-class archive format for ONNX models: https://github.com/onnx/onnx/wiki/%5Bdraft%5D-Proposal-for-ONNX-archive-file-format. I think we can incorporate the ideas from both of our proposals. Let me read over them and see what we should do

postrational:
Sounds good, we can work together to combine these ideas.

NiklasGustafsson:
Looking at both proposals, it seems to me that some advancement on this front will be a great plus.

Here are some questions the proposals:

1. If the tensors are stored in a Zip file in the proposal by @jamesr66a, but you want to be able to memory-map the data directly, I presume you would not be able to compress anything. Is that right?

2. The ONNX spec specifically forbids strings from the 'raw_data' field or TensorProto. How will string tensors be serialized in these schemes?

3. There's another use case for serializing tensors outside the model file -- incremental updates to a model after retraining. This can be useful in situations where weights are adjusted on a frequent basis and there are bandwidth concerns. The Zip-based proposal would complicate this, it seems to me, while the proposal from @postrational would not.

Thanks,

Niklas
Microsoft

lutzroeder:
Could something similar to the ZIP approach be accomplished in a less invasive way? For example, building on the `external_data` field? Attach the tensor data at the end of a regular ONNX protobuf file and use a special URI scheme to index into that data? This might even be possible without having to change the ONNX protobuf format sans the URI support.

Taking a dependency on ZIP means all tools need to take a dependency on ZIP. Might make sense as a transport optimization but should the standard enforce a ZIP dependency or is that limiting options down the road? I'm thinking about this in terms of how opening those file in a browser would require extra hoops but decompression time vs. memory mapping might cause issues for mobile as well.

jamesr66a:
@NiklasGustafsson just spoke with @Yangqing and he pointed out the constraints on raw strings in the `raw_data` field from his experience:

>The tricky thing is that if you specify a string delimiter, one has to make sure that binary strings do not contain that delimiter. That makes two constraints, either (1) we only allow certain encodings, or (2) we need to do a base64-type encoding to explicitly reserve some delimiter

These issues are probably not insurmountable, but I'd like to discover 1) How important is this for you and 2) What approach is the best for the common use cases in ONNX?

NiklasGustafsson:
I share the view on compression not being particularly likely for floats, given the typical bit patterns. My question was about the ability to memory-map, which seems to preclude compression, and we're on the same page there.

I have no particular interest in string tensors, I just wondered whether it wasn't the case that the Zip-based design would exclude string tensors. The multi-file design also seems to have the same limitation, by the way. I don't know how common it will be to have string tensors as anything except graph inputs and outputs, but it's a limitation we should be cognizant of.

Beyond the ability to have really large models, one thing that is <b>really</b> important for some of our scenarios is the ability to incrementally update models, especially those really large models that this is being designed for. Therefore, I want to make sure that if we consider a multi-part model design, that this incremental-update ability be part of any solution.

linkerzhang:
I had a concern about having one file to store one tensor, which will create too many files potentially. Offline discussed Michal, we will extend the external_data to be able to carry index information so that one file can store more than one tensor as needed. The external_data will look like <file_name>?offset=<>&length=<>.

shinh:
@postrational As locally chatted at ONNX workshop, it'd be worth considering to have some recommendation and/or restriction on alignments of `offset` field to allow backends to use mmap efficiently (e.g., SIMD alignment). I'm guessing just enforcing 4k boundary would be the simplest choice, but I'm not sure.

postrational:
I added updates to the PR to reflect changes based on discussions from the last ONNX workshop.

onnx/onnx-ml.proto
```diff
@@ -380,6 +380,29 @@ message TensorProto {
...
+  // - "checksum" (optional) - SHA1 digest of file specified in under 'location' key.
+  repeated StringStringEntryProto external_data = 13;
+
+  // Location of the data for this tensor. MUST be one of:
+  // - MESSAGE - data stored in type-specific fields inside the protobuf message.
+  // - RAW - data stored in raw_data field.
+  // - EXTERNAL - data stored in an external location as described by external_data field.
+  // - value not set - legacy value. Assume data is stored in raw_data (if set) otherwise in message.
+  enum DataLocation {
+    MESSAGE = 0;
```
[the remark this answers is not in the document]
postrational: I would think we should default to `UNKNOWN = 0` or `NOTSET = 0` and treat that as the legacy value.

yufenglee:
@houseroad I added the checker. Could you help to take a look at it?

houseroad:
@yufenglee I will find some time tomorrow and push it forward, no worries.

onnx/__init__.py
```diff
@@ -98,7 +112,14 @@ def load_model(f, format=None):  # type: (Union[IO[bytes], Text], Optional[Any])
...
     s = _load_bytes(f)
-    return load_model_from_string(s, format=format)
+    model = load_model_from_string(s, format=format)
+
+    model_filepath = _get_file_path(f)
+    if model_filepath:
+        basepath = os.path.dirname(model_filepath)
+        model = add_basepath_to_external_data_tensors(model, basepath)
+
+    return model
```
houseroad: for `load_model` we should have another parameter `load_external_data` (default should be true) to let user decide whether loading the external data or not. 

Also we should have another helper function `load_external_data_for_model(model, base_dir)`. I think we need to make it clear that it's users' responsibility to keep the base dir info. It's kind of ugly to save extra info and strip it out after loading.

onnx/external_data_helper.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    for tensor in _get_all_tensors(model):
+        if len(tensor.external_data):
+            tensor.external_data.add()
+            tensor.external_data[-1].key = 'basepath'
+            tensor.external_data[-1].value = filepath
+    return model
+
+
+def write_external_data_tensors(model, filepath):  # type: (ModelProto, Text) -> ModelProto
```
houseroad: I think it's to ask user to provide the base_dir.

onnx/external_data_helper.py
```diff
@@ -0,0 +1,185 @@
...
+    model: Model object to modify.
+    filepath: System path to the directory which should be treated as base path for external data.
+
+    @return
+    The modified model object.
+    """
+    for tensor in _get_all_tensors(model):
+        if len(tensor.external_data):
+            tensor.external_data.add()
+            tensor.external_data[-1].key = 'basepath'
```
houseroad: Let's don't add basepath in the external_data, instead, ask user to provide it as an argument.

onnx/__init__.py
```diff
@@ -153,6 +174,14 @@ def save_model(proto, f, format=None):  # type: (Union[ModelProto, bytes], Union
...
     format is for future use
     '''
+    if isinstance(proto, bytes):
+        proto = _deserialize(proto, ModelProto())
+
+    model_filepath = _get_file_path(f)
+    if model_filepath:
+        basepath = os.path.dirname(model_filepath)
+        proto = write_external_data_tensors(proto, basepath)
+
```
houseroad: We can add a parameter in save_model, all_data_as_external=False, so all tensors' data will be save to external_data. We probably need to modify the proto, and revert it back after saving.

houseroad: how about having three options for users: 
1) embedded weights (default)
2) one file for all external data
3) separate files for different external data

onnx/__init__.py
postrational: Agreed. We can also make it more flexible by adding more methods and parameters.

We should just make sure we keep the user interface simple and backwards compatible. A simple call to `onnx.load_model(path)` and `onnx.save_model(proto, path)` should work whether data is internally or externally stored.

onnx/onnx.in.proto
```diff
@@ -378,6 +378,28 @@ message TensorProto {
   // When this field is present, the data_type field MUST NOT be STRING or UNDEFINED
   optional bytes raw_data = 9;
 
+  // Data can be stored inside the protobuf file using type-specific fields or raw_data.
+  // Alternatively, raw bytes data can be stored in an external file, using the external_data field.
+  // external_data stores key-value pairs describing data location. Recognized keys are:
+  // - "location" (required) - file path relative to the filesystem directory where the ONNX
```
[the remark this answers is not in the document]
postrational: Done.

onnx/external_data_helper.py
```diff
@@ -0,0 +1,185 @@
...
+
+from typing import Iterable, Text
+
+from .onnx_pb import TensorProto, ModelProto
+
+
+class ExternalDataInfo(object):
+
+    def __init__(self, tensor):  # type: (TensorProto) -> None
+        self.location = ''
```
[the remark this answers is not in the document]
postrational: `ExternalDataInfo` acts as just a struct for holding information about the external data. I don't think we should prevent the construction of this object if the `location` field is not set.

onnx/__init__.py
```diff
@@ -153,6 +174,14 @@ def save_model(proto, f, format=None):  # type: (Union[ModelProto, bytes], Union
     f can be a file-like object (has "write" function) or a string containing a file name
     format is for future use
     '''
+    if isinstance(proto, bytes):
+        proto = _deserialize(proto, ModelProto())
```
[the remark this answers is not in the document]
postrational: What should be warning message in this case?

onnx/external_data_helper.py
linkerzhang: That's not good. It means the model file itself is not self-contained.

houseroad: this is not a valid key in our spec.

linkerzhang:
@houseroad @postrational  guys, thanks a lot for pushing this forward. I'm seeing that we all agreed the proto changes but having some different opinions on the helper functions. 

I'm suggesting to merge this PR for 1.4 release, as the standard itself cares most about proto changes, while helper functions are nice to have (no impact on the standard).  Thoughts?

houseroad:
Yes, we should push it in. Let's polish more on the API and merge it.

onnx/external_data_helper.py
```diff
@@ -0,0 +1,185 @@
...
+from __future__ import unicode_literals
+
+import os
+from itertools import chain
+
+from typing import Iterable, Text
+
+from .onnx_pb import TensorProto, ModelProto
+
+
```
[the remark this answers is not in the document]
yufenglee: Added.

onnx/__init__.py
yufenglee: Added a helper convert_model_to_external_data. Users can call this function first and then save tensor data externally.

PR #681, opened 2018-03-30
Clarify behavior of Shape names

anderspapitto:


docs/IR.md
```diff
@@ -55,12 +55,13 @@ The following describes the metadata properties of a model graph:
...
-Names are organized into separate namespaces, and must be unique within a namespace. The namespaces include the following:
- - `Node`: names that identify specific nodes in the graph, but not necessarily any particular input or output of the node.
- - `Graph`: names that identify graphs in the protobuf.
+Names are organized into separate namespaces. The namespaces include the following:
+ - `Node`: names that identify specific nodes in the graph, but not necessarily any particular input or output of the node. Each node name must be unique.
+ - `Graph`: names that identify graphs in the protobuf. Each graph name must be unique.
  - `Attribute`: names that identify attribute names for extra attributes that are passed to operators.
- - `Operator`: names that identify particular operators.
- - `Tensor`: names that identify intermediate tensor values flowing through the computation of a graph.
+ - `Operator`: names that identify particular operators. Each operator name must be unique.
```
linkerzhang: operator name should be unique in its domain.

docs/IR.md
```diff
@@ -55,12 +55,13 @@ The following describes the metadata properties of a model graph:
...
+ - `Tensor`: names that identify intermediate tensor values flowing through the computation of a graph. Each tensor name must be unique.
+ - `Shape`: names that identify symbolic dimensions. Shape names that occur more than once are constrained to take on the same value at each usage, except for the empty string name.
```
houseroad: Let's also provide the meaning of the empty string here.

onnx/onnx.in.proto
```diff
@@ -372,7 +372,9 @@ message TensorProto {
 
 // Defines a tensor shape. A dimension can be either an integer value
 // or a symbolic variable. A symbolic variable represents an unknown
-// dimension.
+// dimension. In general, any two dimensions in the graph with the
+// same value of dim_param are constrained to be equal. The only
+// exception to this is the emtpy string.
```
houseroad: Also here, the meaning of the empty string.

docs/IR.md
```diff
@@ -55,12 +55,13 @@ The following describes the metadata properties of a model graph:
 
 #### Names Within a Graph
 
-Names are organized into separate namespaces, and must be unique within a namespace. The namespaces include the following:
- - `Node`: names that identify specific nodes in the graph, but not necessarily any particular input or output of the node.
- - `Graph`: names that identify graphs in the protobuf.
+Names are organized into separate namespaces. The namespaces include the following:
+ - `Node`: names that identify specific nodes in the graph, but not necessarily any particular input or output of the node. Each node name must be unique.
```
houseroad: We should add scope for each type of names. For example, Node name must be unique in the whole model.

docs/IR.md
```diff
@@ -55,12 +55,13 @@ The following describes the metadata properties of a model graph:
...
+ - `Graph`: names that identify graphs in the protobuf. Each graph name must be unique.
```
houseroad: Model scope unique.

docs/IR.md
```diff
@@ -55,12 +55,13 @@ The following describes the metadata properties of a model graph:
...
  - `Attribute`: names that identify attribute names for extra attributes that are passed to operators.
```
houseroad: Node scope unique?

docs/IR.md
houseroad: Domain scope unique.

docs/IR.md
dzhulgakov: Actually - do we require graph names to be present? If that's an inlined for loop - it probably doesn't matter that much

docs/IR.md
dzhulgakov: Yep, we should discuss it - or maybe make it graph-level unique? @jamesr66a probably has a view on it given the control flow considerations

Btw, shall it be "optional"?

docs/IR.md
dzhulgakov: Yeah, operators don't fit it as they're more part of the spec than the model

docs/IR.md
```diff
@@ -55,12 +55,13 @@ The following describes the metadata properties of a model graph:
...
- - `Operator`: names that identify particular operators.
- - `Tensor`: names that identify intermediate tensor values flowing through the computation of a graph.
+ - `Operator`: names that identify particular operators. Each operator name must be unique.
+ - `Tensor`: names that identify intermediate tensor values flowing through the computation of a graph. Each tensor name must be unique.
```
dzhulgakov: Global scope?

onnx/onnx.in.proto
```diff
@@ -368,7 +368,10 @@ message TensorProto {
 
 // Defines a tensor shape. A dimension can be either an integer value
 // or a symbolic variable. A symbolic variable represents an unknown
-// dimension.
+// dimension. In general, any two dimensions in the graph with the
+// same value of dim_param are constrained to be equal. The only
+// exception to this is the emtpy string, which is a special value
+// that is treated as always being unique.
```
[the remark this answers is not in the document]
gramalingam: So, how should we handle worst-case scenarios where shape-inference cannot determine anything about the output shape? Is the plan to use a missing shape in a TypeProto::Tensor to indicate a completely unknown shape? (That seems the obvious choice, but want to double-check.)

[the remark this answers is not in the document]
anderspapitto: yes. has_shape() will return false for a completely unknown shape. has_shape() = true and dim_size() = 0 indicates a scalar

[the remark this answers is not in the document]
anderspapitto: note that you can also just construct your protobuf in this state directly, so those semantics are already set even before introducing shape inference

anderspapitto:
actually, I feel this is not needed, because a `oneof` is actually a one-or-zero-of. That means that we can use a dim which is present but which is neither a dim_value nor a dim_param to represent uncorrelated, unknown dimensions.

(Open to discussing further, but I'm satisfied with that approach, which already works, and so I'll close this PR)

Issue #323, opened 2017-11-27
onnx build from source issue on mac


Roshrini:
I want to build onnx from source on macos sierra 10.12 and facing the following issue:
Can anyone take a look at this please?

:~ roshanin$ python
Python 2.7.12 |Anaconda custom (x86_64)| (default, Jul  2 2016, 17:43:17) 
[GCC 4.2.1 (Based on Apple Inc. build 5658) (LLVM build 2336.11.00)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import onnx
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/roshanin/anaconda2/lib/python2.7/site-packages/onnx-0.2.1-py2.7-macosx-10.6-x86_64.egg/onnx/__init__.py", line 10, in <module>
    import onnx.helper  # noqa
  File "/Users/roshanin/anaconda2/lib/python2.7/site-packages/onnx-0.2.1-py2.7-macosx-10.6-x86_64.egg/onnx/helper.py", line 15, in <module>
    import onnx.defs as defs
  File "/Users/roshanin/anaconda2/lib/python2.7/site-packages/onnx-0.2.1-py2.7-macosx-10.6-x86_64.egg/onnx/defs/__init__.py", line 6, in <module>
    import onnx.onnx_cpp2py_export.defs as C
ImportError: dlopen(/Users/roshanin/anaconda2/lib/python2.7/site-packages/onnx-0.2.1-py2.7-macosx-10.6-x86_64.egg/onnx/onnx_cpp2py_export.so, 2): Symbol not found: __ZN6google8protobuf14MessageFactory29InternalRegisterGeneratedFileEPKcPFvRKSsE
  Referenced from: /Users/roshanin/anaconda2/lib/python2.7/site-packages/onnx-0.2.1-py2.7-macosx-10.6-x86_64.egg/onnx/onnx_cpp2py_export.so
  Expected in: flat namespace
 in /Users/roshanin/anaconda2/lib/python2.7/site-packages/onnx-0.2.1-py2.7-macosx-10.6-x86_64.egg/onnx/onnx_cpp2py_export.so

Tested that "conda install -c ezyang onnx" works fine.
 
roshanin$ protoc --version
libprotoc 3.4.0

python installed through conda and has 2.7.12 version

[...]

Issue #263, opened 2017-11-15
Remove pip install dependency on protoc


postrational:
When the ONNX Python package becomes a dependency of a project, then currently the Protocol Buffers compiler becomes a dependency.

When installing the ONNX Python package many users get the following error:

```
Could not find "protoc" executable!
```

This is caused by the missing Protocol Buffers compiler. On some systems it's easy to install Protocol Buffers, but on Centos or Ubuntu 14.04 this is not so simple. Packaged versions of Protocol Buffers are older than 2.6.1, so users get these errors:

```
onnx.proto:308:7: Expected "required", "optional", or "repeated".
onnx.proto:308:19: Missing field number.
onnx.proto:330:3: Expected "required", "optional", or "repeated".
onnx.proto:330:15: Missing field number.
```

Would it be possible to remove the dependency on `protoc` from the `pip install` process?
Could the PyPI package be distributed with a precompiled form of the proto?

bddppq:
Even with precompiled pb.cc files, you will still need the protobuf headers and libraries to build onnx.

PR #684, opened 2018-04-01
Clarifies installation error message when external protobuf dependencies are missing

daniel-j-h:
At the moment on a system without protobuf dependencies the installation fails with

    pip install onnx

```
    Could not find "protoc" executable!
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-r45lhxcn/onnx/
```

There seem to be a couplf of folks running into this
- https://github.com/onnx/onnx/issues/263
- https://github.com/onnx/onnx/issues/323

Solution on my Ubuntu 16.04 was to install

    sudo apt install protobuf-compiler libprotobuf-dev

Let's clarify the error message so folks know how to proceed when running into this error.


bddppq:
Thanks for the fix

Issue #511, opened 2018-02-06
Determine if a NodeProto.input is a constant or a graph input 


MrGeva:
Is there a way to determine if a NodeProto input is a constant (e.g. weights, biases) or a graph input (e.g. image)?
It seems that in both cases they may appear in the model in GraphProto.input and in GraphProto.initializer, so I can't distinguish between them.

bddppq:
initializers are all weights. GraphProto.input includes both weights and graph input, so it’s a super set of initilaizers. Technically initializers are list of TensorProto (which has the values), while GraphProto.input is a list of ValueInfoProto, which only has the type & shape information.

MrGeva:
The documentation in onnx.proto is somewhat misleading because the comment above GraphProto.initializer implies inputs to the graph can be initialized with default values:

> // A list of named tensor values (constants), used to specify default
  // values for some of the inputs of the graph...
  repeated TensorProto initializer = 5;

PR #690, opened 2018-04-02
Fix doc for initializer

bddppq:
https://github.com/onnx/onnx/issues/511

### DIFF
--- a/onnx/onnx-ml.proto
+++ b/onnx/onnx-ml.proto
@@ -236,13 +236,10 @@ message GraphProto {
   // The name of the graph.
   optional string name = 2;   // namespace Graph
 
-  // A list of named tensor values (constants), used to specify default
+  // A list of named tensor values (constants), used to specify constant
   // values for some of the inputs of the graph.
   // Each TensorProto entry must have a distinct name (within the list) that
   // also appears in the input list.
-  // In an evaluation, the default value specified here is used if and only if
-  // user specifies no value for the corresponding input parameter.
-  // May be used to pass serialized parameters for networks.
   repeated TensorProto initializer = 5;
 
   // A human-readable documentation for this graph. Markdown is allowed.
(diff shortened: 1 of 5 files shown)


onnx/onnx-ml.proto
```diff
@@ -236,13 +236,10 @@ message GraphProto {
   // The name of the graph.
   optional string name = 2;   // namespace Graph
 
-  // A list of named tensor values (constants), used to specify default
+  // A list of named tensor values (constants), used to specify constant
```
houseroad: delete the first `constants` in the parenthesis?

onnx/onnx-ml.proto
```diff
@@ -236,13 +236,10 @@ message GraphProto {
...
   // values for some of the inputs of the graph.
   // Each TensorProto entry must have a distinct name (within the list) that
   // also appears in the input list.
-  // In an evaluation, the default value specified here is used if and only if
```
houseroad: I think this paragraph says that even the graph contains some values in the initializer, we can still ignore it by passing the value in the input list.

bddppq: @houseroad Yeah that was in the old days we were trying to provide the convenience to users to override the constant values in the `run_model`, `run_node` python interface, but it doesn't work anymore as we have more and more stuffs relying on the static graph analysis.

onnx/onnx-ml.proto
houseroad: Also shall we explicitly say these are weights of the models. :-)

linkerzhang:
Should we allow "GraphProto.input includes both weights and graph input," please? I feel graphproto.input should only contain graph input. make sense please?

bddppq:
@linkerzhang the purpose of putting initializer into graph.input is if (someday) onnx supports training,  then existing models can simply drop the initializers and its graph is still a valid graph. TBH I don't feel there are too much differences between including or not including initilaizers' ValueInfoProto in graph.input :-) But since we have started this way, it's probably not worth the effort of introducing a breaking change?

gramalingam:
@bddppq : but the change in the documentation seems to have dropped the lines that say that the caller-specified value overrides the weight specified in the model ... so, this is a bit confusing. If we want to retain the original behavior, it would be nice to have a line in the documentation indicating this.

PR #692, opened 2018-04-02
Add model save function

houseroad:
Save model function and related test.

### DIFF
--- a/onnx/test/basic_test.py
+++ b/onnx/test/basic_test.py
@@ -34,29 +34,33 @@ class TestProtobufExists(unittest.TestCase):
         self.assertTrue(model == loaded_model)
 
         # Test if input is a file name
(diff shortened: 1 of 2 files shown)


onnx/test/basic_test.py
```diff
@@ -37,6 +41,23 @@ def test_load(self):
...
+        onnx.save(model, f)
+        f.close()
+        loaded_model = onnx.load(f.name)
+        self.assertTrue(model == loaded_model)
+
+        # Test path string paramter
+        onnx.save(model, f.name)
+        loaded_model = onnx.load(f.name)
+        self.assertTrue(model == loaded_model)
+        os.remove(f.name)
```
dzhulgakov: nit: put it into try/finally?

houseroad: Good point!

PR #706, opened 2018-04-04
CMake fixes and setup.py cleanup

bddppq:


setup.py
```diff
@@ -291,61 +167,46 @@ def run(self):
...
     def run(self):
         self.run_command('create_version')
-        self.run_command('build_proto')
         self.run_command('cmake_build')
+
+        for src in glob.glob(
+                os.path.join(CMAKE_BUILD_DIR, 'onnx', '*.py')):
+            dst = os.path.join(
+                TOP_DIR, os.path.relpath(src, CMAKE_BUILD_DIR))
+            self.copy_file(src, dst)
```
anderspapitto: is this part new?

CMakeLists.txt
```diff
@@ -112,20 +112,29 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP SRCS HDRS ROOT_DIR)
...
     set(OUTPUT_PROTO_DIR "${CMAKE_CURRENT_BINARY_DIR}/${REL_DIR}")
 
-    if(ONNX_NAMESPACE STREQUAL "onnx")
-      set(OUTPUT_PB_HEADER "${OUTPUT_PROTO_DIR}/${FILE_WE}.pb.h")
-      set(OUTPUT_PB_SRC "${OUTPUT_PROTO_DIR}/${FILE_WE}.pb.cc")
-      set(GENERATED_PROTO "${OUTPUT_PROTO_DIR}/${FILE_WE}.proto")
-    else()
-      set(OUTPUT_PB_HEADER "${OUTPUT_PROTO_DIR}/${FILE_WE}_${ONNX_NAMESPACE}.pb.h")
-      set(OUTPUT_PB_SRC "${OUTPUT_PROTO_DIR}/${FILE_WE}_${ONNX_NAMESPACE}.pb.cc")
+    set(OUTPUT_PB_HEADER "${OUTPUT_PROTO_DIR}/${GENERATED_FILE_WE}.pb.h")
```
anderspapitto: is this bit new behavior or pure refactor

bddppq: new, this is the cmake fixes that needed for supporting ONNX_ML build

setup.py
bddppq: yep, it's needed to copy the cmake generated python files to setup.py's directory. This bug was not discovered because previously setup.py does its own protobuf autogen work (in addition to cmake).

bddppq:
@anderspapitto mixing is needed because there are bugs discovered after removing redundant code in setup.py.

Issue #535, opened 2018-02-16
Simple optimization pass: fuse Conv+Add into Conv with Bias


dzhulgakov:
Conv in onnx spec supports embedded bias as it's often represented this way in accelerated backends: https://github.com/onnx/onnx/blob/master/docs/Operators.md#conv

However, some exporters produce separate Conv and Add operators. We should fuse them if bias is 1-d in this case and have it available as optimization pass (https://github.com/onnx/onnx/tree/master/onnx/optimizer/passes)

fumihwh:
@dzhulgakov 
I would love to take this. But I am new to C++ and have no idea how to debug C++ (especially using CLion).

Following is my progress:
- build onnx from source
- do cmake (cmake-build-debug has been created)
- try test_fuse_transpose

But I can't step into `onnx.optimizer.optimize(orig_model_str, opts)`.
I googled and find there is no executable C application of optimizer under onnx root folder.
Anyone help?

bddppq:
@fumihwh Thanks for working on this! Yeah right now all of our tests are driven by python so there is no C++ executable targets. I'm not familiar with CLion. Normally I use gdb/lldb to step into the c++ code. e.g. you can do 
1. `gdb --args python onnx/tests/optimizer_test.py` in the shell to put a python process into gdb
2. `b onnx::optimization::Optimizer::Optimizer` inside gdb to set a breakpoint at the [Optimizer's constructor](https://github.com/onnx/onnx/blob/8dc7369bb9c3c1b63056211339987b0ad015faba/onnx/optimizer/optimize.h#L22)

fumihwh:
@bddppq 
Great! Thanks! I will try this later.

Currently, I am using another way to do this. Make an onnx-test.cpp to parse pb first and call optimizer.
Succeed in parse pb to ModelProto by using https://github.com/lcskrishna/onnx-parser . (It lets me to know a little bit more about C++ build, compile...)
And when I try to call (**add**, strictly) `onnx::optimization::Optimize`, failed on build. `ld: symbol(s) not found for architecture x86_64` occurs.

So if you have time, could you tell me what should I add to CMakeLists.txt?
I tried followed CMakeLists.txt but not work. (Just add them in the last of current CMakeLists.txt in onnx)
```
include_directories(onnx)
include_directories(onnx/third_party/pybind11/include)
include_directories(/usr/include/python2.7)
list(APPEND SOURCES
        main.cpp
[...]

PR #707, opened 2018-04-04
add fuse_conv_add_into_bias optimizer

fumihwh:
#535


onnx/optimizer/passes/fuse_conv_add_into_bias.h
```diff
@@ -0,0 +1,43 @@
...
+#include "onnx/optimizer/passes/optimize_pass.h"
+
+namespace ONNX_NAMESPACE { namespace optimization {
+
+struct FuseConvAddIntoBias final : public OptimizePass {
+  explicit FuseConvAddIntoBias()
+    : OptimizePass("fuse_conv_add_into_bias", API_TYPE::IR) {
+  }
+
+  static int idx_of_conv(const onnx::ArrayRef<onnx::Value *> & values) {
```
houseroad: We use ONNX_NAMESPACE

onnx/optimizer/passes/fuse_conv_add_into_bias.h
```diff
@@ -0,0 +1,43 @@
...
+      if (values[i]->node()->kind() == kConv)
+        return i;
+    return -1;
+  }
+
+  void fuse_conv_add_into_bias(Graph& graph) {
+    for (auto it = graph.begin(); it != graph.end(); ++it) {
+      auto* n = *it;
+      if (n->kind() == kAdd) {
+        int idx = idx_of_conv(n->inputs());
```
houseroad: Add only has and only has two inputs, I think it's fine to have this assumption.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
```diff
@@ -0,0 +1,43 @@
...
+        if (idx != -1 && n->inputs()[idx]->node()->inputs().size() <= 2) {
```
houseroad: We need to check 1) Add's broadcast flag, 2) the shape of bias. (The first dimension of bias is the same as the Conv's 2nd input's 1st dimension.)

onnx/optimizer/passes/fuse_conv_add_into_bias.h
```diff
@@ -0,0 +1,43 @@
...
+          auto origConv = n->inputs()[idx];
+          auto origBias = n->inputs()[1 - idx];
```
houseroad: You already have two input assumption here. :-)

onnx/test/optimizer_test.py
```diff
@@ -99,6 +99,20 @@ def test_fuse_transpose_into_gemm(self):
         assert len(list(optimized_model.graph.node)) == 1
         assert optimized_model.graph.node[0].op_type == "Gemm"
 
+    def test_fuse_conv_add_into_bias(self):
+        conv = helper.make_node("Conv", ["X", "Y"], ["Z"])
```
houseroad: I would like to have another case, with broadcast enabled, our optimizer should not have any optimization

onnx/optimizer/passes/fuse_conv_add_into_bias.h
fumihwh: @houseroad 
Yes. And I don't know which input comes from conv op, so I use `idx_of_conv` to get the index.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
fumihwh: I use `namespace ONNX_NAMESPACE { namespace optimization {` at first. What should I do with this method?

onnx/test/optimizer_test.py
fumihwh: I think our optimizer should only work with broadcast enabled and axis=1.
So why with broadcast enabled is unavailable?

houseroad: I mean we have some constraints on Conv's bias, it has to be a 1D tensor with size M (which is the feature number of W). So in case broadcasting case, this fusion cannot be conducted.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
houseroad: I will check whether there is a bug in our IR constructor or just we don't have enough information in this case.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
houseroad: I think `onnx::` is not needed, ArrayRef should be visible. I will double check.

onnx/test/optimizer_test.py
fumihwh: Depends on above comment mentioned that we need conv shape information to check this.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
fumihwh: Ok. I’ll also check the constructor.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
fumihwh: checked and removed.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
fumihwh: @houseroad 
For priori,
```
NodeProto doesn't have dim information. And field type of input and output is string.
TensorProto and ValueInfoProto has dim information.
```

In my practice,
I call `def make_graph(nodes, name, inputs, outputs, initializer=None, doc_string=None)` to make graph.

nodes: list of NodeProto
inputs: list of ValueInfoProto
outputs: list of ValueInfoProto
initializer: list of TensorProto

I think it should be a problem that each node does not keep the output shape.
onnx knows shape information of inputs and outputs and initializer.
But after one operator, shape information disappears. We can not find it, because it doesn't exist in graph.

anderspapitto: yes, there is no guarantee to have size information for intermediate values. We are separately working on adding shape inference, to fill it in in more cases.

anderspapitto:
so, if the optimization is unsound when the dimensions don't match, then we should make the optimization only fire when the shapes are known.

There are two ways that we can make sure the shapes are known

- producers of ONNX graphs can manually provide ValueInfoProtos for the intermediate values

OR

- we can add enough shape inference that it's not necessary. This work is ongoing.

I think it would also be fine to emit some warning that says "we want to fuse this operation, but we cannot because size information is missing. Please annotate your graph for extra performance"

fumihwh:
@anderspapitto @houseroad 
I made following changes
- check if size information exists
- check if bias is 1d
- check if bias dim is equal to second dim of conv
- add 2 test cases

onnx/optimizer.py
```diff
@@ -22,6 +22,7 @@
     -- nop
     -- eliminate_nop_transpose
     -- fuse_consecutive_transposes
+    -- fuse_conv_add_into_bias
```
houseroad: I would like to call it fuse_add_bias_into_conv.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
```diff
@@ -0,0 +1,62 @@
...
+            orig_conv->node()->addInput(orig_bias);
+            n->replaceAllUsesWith(orig_conv->node());
+            it.destroyCurrent();
+          }
+        }
+      }
+    }
+    if (size_lack_count != 0) {
+      std::cout <<
+                "We can't fuse some operations due to lack of size information."
```
anderspapitto: can we be a little more specific? e.g. "Warning: failed to fuse Add into Conv bias due to ..."

onnx/test/optimizer_test.py
```diff
@@ -99,6 +99,57 @@ def test_fuse_transpose_into_gemm(self):
...
+
+    def test_fuse_conv_add_into_bias_no_fuse_3d_bias(self):
+        conv = helper.make_node("Conv", ["X", "Y"], ["Z"])
+        add = helper.make_node("Add", ["Z", "A"], ["B"], broadcast=1, axis=1)
+        graph = helper.make_graph(
+            [conv, add],
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (1, 5, 3, 3)),
+             helper.make_tensor_value_info("Y", TensorProto.FLOAT, (5, 16, 3, 3)),
+             helper.make_tensor_value_info("A", TensorProto.FLOAT, (16, 1, 1))],
```
houseroad: If you want, you can try to handle such cases.

Hints: add reshape to change the shape of bias. So it can be fused into conv.
This case is a little bit challenging. :-)

onnx/optimizer/passes/fuse_conv_add_into_bias.h
```diff
@@ -0,0 +1,62 @@
...
+        if (idx != -1 && n->inputs()[idx]->node()->inputs().size() <= 2) {
+          auto orig_conv = n->inputs()[idx];
+          auto orig_bias = n->inputs()[1 - idx];
+          auto conv_shape = orig_conv->sizes();
+          auto bias_shape = orig_bias->sizes();
+          if (bias_shape.size() == 0 || conv_shape.size() == 0) {
+            size_lack_count += 1;
+            continue;
+          }
+          if (bias_shape.size() != 1 || bias_shape[0].dim != conv_shape[1].dim) {
```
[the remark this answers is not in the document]
fumihwh: I am not sure if we need to also check weight dim M same to bias dim.
Why check output from conv is not enough?

onnx/test/optimizer_test.py
```diff
@@ -99,6 +99,57 @@ def test_fuse_transpose_into_gemm(self):
...
+
+    def test_fuse_conv_add_into_bias_no_fuse_no_size(self):
+        conv = helper.make_node("Conv", ["X", "Y"], ["Z"])
+        add = helper.make_node("Add", ["Z", "A"], ["B"], broadcast=1, axis=1)
+        graph = helper.make_graph(
+            [conv, add],
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (1, 5, 3, 3)),
+             helper.make_tensor_value_info("Y", TensorProto.FLOAT, (5, 16, 3, 3)),
+             helper.make_tensor_value_info("A", TensorProto.FLOAT, (16,))],
```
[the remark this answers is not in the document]
fumihwh: Aha, I know why do you want to check weight dim.
I think checking weight dim should not be included here. It's a Conv specification unconformity. Should check in checker.py? maybe.

onnx/optimizer/passes/fuse_conv_add_into_bias.h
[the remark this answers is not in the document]
houseroad: I didn't say checking output is not enough. I mean you can enhance the optimization by checking the weights (2nd input) of Conv.

As long as you get either weight's shape or output's shape, you can do the optimization. (Of course, Bias's shape is also necessary.)

onnx/test/optimizer_test.py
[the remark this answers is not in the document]
houseroad: I think we can depends on weight's shape information. Because in the operator doc, we explicitly say both weight_shape[0].dim = bias_shape[0].dim = M.

onnx/optimizer/passes/fuse_add_bias_into_conv.h
```diff
@@ -30,11 +30,16 @@ struct FuseAddBiasIntoConv final : public OptimizePass {
           auto orig_bias = n->inputs()[1 - idx];
           auto conv_shape = orig_conv->sizes();
           auto bias_shape = orig_bias->sizes();
-          if (bias_shape.size() == 0 || conv_shape.size() == 0) {
+          auto weight_shape = orig_conv->node()->inputs()[1]->sizes();
+          if (bias_shape.size() == 0
+              || conv_shape.size() == 0
```
houseroad: Here should be (conv_shape.size() == 0 && weight_shape.size() == 0).

Either we have conv_shape[1] or weight_shape[1], we can proceed. Because they are equivalent.

fumihwh:
@houseroad 
No idea with appveyor stops at optimizer_test.py.....

onnx/common/ir.h
```diff
@@ -671,6 +671,23 @@ struct Node : public Attributes<Node> {
...
 
+  // Check whether this node is before node n in the graph.
+  bool isBefore(Node* n) {
+    if (n == nullptr || this == n) {
+      // Bail out early.
+      return false;
+    }
+    ONNX_ASSERT(n->inGraphList());
+    Node* p = next();
+    while (p != nullptr) {
```
anderspapitto: you're going to get featured in https://accidentallyquadratic.tumblr.com/

but it's probably fine in this case

onnx/common/ir.h
```diff
@@ -991,7 +1008,8 @@ inline Value::Value(Node * node_, size_t offset_)
   offset_(offset_),
   unique_(node_->graph_->next_unique_++),
   stage_(node_->graph_->new_node_stage_),
-  has_unique_name_(false) {
+  has_unique_name_(false),
+  elem_type_(ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED) {
```
anderspapitto: ++ thanks for fixing

onnx/common/ir_pb_converter.cc
```diff
@@ -20,6 +20,7 @@ Tensor tensorProtoToTensor(const ONNX_NAMESPACE::TensorProto & tp) {
   switch(tp.data_type()) {
   case ONNX_NAMESPACE::TensorProto_DataType_FLOAT:
   case ONNX_NAMESPACE::TensorProto_DataType_COMPLEX64: {
+    // TODO: fix this handling
```
anderspapitto: what is the issue here? should we fix it now, or maybe open a task to fix it later?

onnx/optimizer/passes/fuse_add_bias_into_conv.h
```diff
@@ -0,0 +1,128 @@
+// ATTENTION: The code in this file is highly EXPERIMENTAL.
+// Adventurous users should note that the APIs will probably change.
+
+#pragma once
+
+// Z = Conv(X, Y)
+// B = Z + A
```
anderspapitto: can you add a line that describes what this is transformed into? I guess it would be `B = Conv(X, Y, A)` or something

houseroad: Good point :-)

onnx/common/ir_pb_converter.cc
houseroad: I think this is fine. Never mind. :-)

onnx/optimizer/passes/fuse_add_bias_into_conv.h
```diff
@@ -0,0 +1,128 @@
...
+  void fuse_add_bias_into_conv(Graph& graph) {
+    int size_lack_count = 0;
+    for (auto it = graph.begin(); it != graph.end(); ++it) {
+      auto* n = *it;
+      if (n->kind() == kAdd && n->inputs()[0]->node()->kind() == kConv
+          && n->inputs()[0]->node()->inputs().size() == 2) {
+        // due to current broadcasting's constraint, Conv has to be the first oprand
+        auto orig_conv = n->inputs()[0];
+        auto orig_bias = n->inputs()[1];
+        // check if bias is Const or in graph's initializers
```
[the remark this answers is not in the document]
houseroad: If bias is not constant or initializer, then we have to check all of the bias's dependencies has no dependencies on conv, so we can move bias before conv.

We don't have any dependency analysis here yet, so better to give this assumption.

Issue #708, opened 2018-04-04
CMake setup fail on local dev build


raymondxyang:
See error from cmake when build using `pip install -e .` . Looking into it now

```
    ERROR Failed to remove indentation from:
    """

          from distutils import sysconfig
          print(sysconfig.get_python_lib(prefix=''))

    """
        Python dedent failed with error code: The system cannot find the file specified
    CMake Error at cmake/Utils.cmake:31 (message):
       Python dedent failed with error code: The system cannot find the file specified

```

raymondxyang:
Fixed

PR #709, opened 2018-04-04
Fix cmake windows local build

raymondxyang:
resolve issue <https://github.com/onnx/onnx/issues/708>


bddppq:
We should even remove this py_cmd stuff from cmake, instead do all the necessary stuffs in setup.py and pass it to cmake as flag.

PR #711, opened 2018-04-04
Input test data of concat op should be float

snnn:


bddppq:
Also need to re-generate the doc

snnn:
Where should I put the data?

houseroad:
The script will automatically generate the data and put them in the right place `onnx/backend/test/data/...`

houseroad:
And @bddppq is right, you also need to regenerate the doc too :-)

snnn:
Just curious, who knows I didn't put any malicious data inside this pr?

houseroad:
@snnn that's a good question. We use checker/CI to verify the model. But in theory, they are not exhaustive. If interested, your contribution on security check is welcome.

bddppq:
Ideally we should in the CI check that the new data files you are submitting are exactly the same as what our tool generates. But data files generated on different platforms/environments can have small precision differences, so the comparison is not that straightforward. This is solvable, but not implemented.

houseroad:
As @bddppq mentioned, we could and should check the generated data. Even with such checks, it may still be possible to add some unnecessary (maybe malicious) data in the PR to exploit vulnerabilities.

houseroad:
Just created https://github.com/onnx/onnx/issues/715 to track the data verification stuff.

snnn:
Hi @houseroad   Document is updated.

snnn:
Anything should I do?

PR #713, opened 2018-04-04
Improve optimizer's API and docs

houseroad:
I will do shape inference one in a different diff


onnx/optimizer.py
```diff
@@ -10,18 +10,36 @@
...
 Supported pass names:
     -- nop
     -- eliminate_nop_transpose
     -- fuse_consecutive_transposes
     -- fuse_transpose_into_gemm
 """
-optimize = C.optimize
+def optimize(model, pass_list=None):
+    if pass_list is None or len(pass_list) == 0:
+        pass_list = ['eliminate_nop_transpose',
```
bddppq: put this list of pass names in a tuple and set it as the default value of pass_list function argument?

onnx/optimizer.py
```diff
@@ -10,18 +10,36 @@
...
+    return (string or ModelProto depends on input) (serialized) optimized model
+
 Supported pass names:
     -- nop
     -- eliminate_nop_transpose
     -- fuse_consecutive_transposes
     -- fuse_transpose_into_gemm
 """
-optimize = C.optimize
+def optimize(model, pass_list=None):
```
bddppq: nit: call it `passes`?

onnx/optimizer.py
```diff
@@ -10,18 +10,36 @@
...
+    if pass_list is None or len(pass_list) == 0:
+        pass_list = ['eliminate_nop_transpose',
+                     'fuse_consecutive_transposes',
+                     'fuse_transpose_into_gemm']
+
+    if isinstance(model, ModelProto):
+        model_str = model.SerializeToString()
+        optimized_model_str = C.optimize(model_str, pass_list)
+        return onnx.load_from_string(optimized_model_str)
+    elif isinstance(model, str):
```
bddppq: I prefer to have one fixed input & output type.

onnx/optimizer.py
```diff
@@ -10,18 +10,35 @@
...
 
+import onnx
 import onnx.onnx_cpp2py_export.optimizer as C  # type: ignore
+from onnx import ModelProto
 
 """Apply the optimization on the serialized ModelProto.
 
 Arguments:
-    input (string): serialized ModelProto
+    input (string or ModelProto): (serialized) model
```
bddppq: The signature should not allow string as input anymore. and the output type should remain as ModelProto

onnx/optimizer.py
```diff
@@ -10,18 +10,35 @@
...
     names (list of string): list of optimization names
 
+Return:
+    return (string or ModelProto depends on input) (serialized) optimized model
```
bddppq: this description doesn't look correct. I think it always return ModelProto now

onnx/optimizer.py
houseroad: good catch

PR #716, opened 2018-04-05
Fix the shape inference python API

houseroad:
Take ModelProto instead of str


bddppq:
There are lint errors:

```
./onnx/shape_inference.py:25:1: E302 expected 2 blank lines, found 0
./onnx/test/shape_inference_test.py:6:1: F401 'onnx.ModelProto' imported but unused
./onnx/test/backend_shape_inference_test.py:12:1: F401 'onnx.ModelProto' imported but unused
```

houseroad:
@anderspapitto when I wrote the doc, I had to convert between str and ModelProto. This is not user-friendly. Dima also mentioned it. So let's change it before it is used widely.

PR #717, opened 2018-04-05
Proposal: Nested attributes

postrational:
There are scenarios where a custom `Node` can contain a complex set of Attributes. We propose an extension of ONNX `AttributeProto` to allow attribute nesting.

For example, if we wanted to implement a `Node`, which fuses convolution and pooling, we would need to store attributes such as `kernel_shape` for both operations. Using nested attributes we could store these attributes in nested groups as follows:

```
op_type: "com.acme.ConvPool"
attribute {
  name: "conv"
  attributes {
    name: "strides"
    ints: 1
    ints: 1
    type: INTS
  }
  attributes {
    name: "kernel_shape"
    ints: 3
    ints: 3
    type: INTS
  }
  type: ATTRIBUTES
}
attribute {
  name: "pool"
  attributes {
    name: "strides"
    ints: 1
    ints: 1
    type: INTS
  }
  attributes {
    name: "kernel_shape"
    ints: 8
    ints: 8
    type: INTS
  }
  type: ATTRIBUTES
}
```

### DIFF
--- a/onnx/checker.cc
+++ b/onnx/checker.cc
@@ -172,11 +172,13 @@ void check_attribute(
   check_singular_field(s, AttributeProto::STRING);
   check_singular_field(t, AttributeProto::TENSOR);
   check_singular_field(g, AttributeProto::GRAPH);
+  check_singular_field(a, AttributeProto::ATTRIBUTE);
   check_repeated_field(floats, AttributeProto::FLOATS);
   check_repeated_field(ints, AttributeProto::INTS);
   check_repeated_field(strings, AttributeProto::STRINGS);
   check_repeated_field(tensors, AttributeProto::TENSORS);
   check_repeated_field(graphs, AttributeProto::GRAPHS);
+  check_repeated_field(attributes, AttributeProto::ATTRIBUTES);
 
 #undef check_type
 #undef check_singular_field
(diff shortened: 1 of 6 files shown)


jspisak:
Thanks for the proposal! Adding Ke as well for review and comment.. @bddppq @linkerzhang

linkerzhang:
Thank you @postrational for the proposal! Node fusion (of graph optimization) is being discussed. This is the wiki of the proposal being discussed, https://github.com/onnx/onnx/wiki/Proposal----Adding-Function-into-ONNX.

Back to this attribute nesting, I personally would suggest not doing it since having attribute flatten in node level does not introduce any inconvenience. In the proposal I attached, attribute reference is a way designed to allow fusion sub nodes inside to refer to attribute data outside (in parent node). We may discuss more on gitter or messenger.

RanACohen:
Maybe putting node fusion as an example is not the best example as it can be solved in other means as suggested.

But for our needs, we have defined a TensorIterator layer (which resembles the caffe2 iterator layer). This layers has a list of attributes where each attribute is a structure of several different attributes.

I know this can also be flatten but it is not convenient to flatten such complex structures and I do image that someone else might have a need for this also.

askhade:
@postrational : Can we close this PR?

PR #719, opened 2018-04-05
Fix gen_proto in cmake

bddppq:


bddppq:
"default" CI failure is irrelevant

bddppq:
@onnxbot retest this please

snnn:
Should it be Python_EXECUTABLE instead of PYTHON_EXECUTABLE ?

https://cmake.org/cmake/help/v3.15/module/FindPython.html

PR #720, opened 2018-04-05
Spec clarity: IR.md modifications.

NiklasGustafsson:
This is part of a series of pull requests related to achieving better clarity in the ONNX specification, a task Microsoft took on driving a couple of weeks ago.

This PR mostly consists of significant changes to the IR.md document, an attempt to reach better clarity and precision in defining the semantics of the ONNX IR and type system. The changes are so substantial that there is little point in reviewing the changes as a 'diff'.

In attempting to reach improved clarity, a few things may have been misunderstood, a few clarificiations may be controversial, and a few things have been left open.

Left Open:

1. There is no description of the optional 'value_info' field in GraphProto.
2. I left out a proposed convention for versioning to distinguish semantic versioning from simple monotonically increasing numbers. The convention would have implied a change to IR versions from semver to simple numbers. Conventions aren't technically normative, but they are strongly encouraged, and it is worth more discussion first.
3. A systematic scrubbing of the .proto files, looking for explanations of semantics that potentially remain uncovered in this document.
4. An explanation of name scoping rules for nested graphs (graph attributes).
5. Version numbers are declared as 64-bit integers in the .proto files, but the 'Versioning.md' document clearly states that version numbers are packed in 32 bits. This needs to be resolved.

Potentially misunderstood:

1. The discussion in this recent PR https://github.com/onnx/onnx/pull/690 suggested that initializers were usable not just as default values for graph inputs. I formalized those rules in the specification based on my understanding of the change in PR #690. 
2. Various explanations of fields. I took a lot of the text from the .proto files, but edited many, and errors may have been introduced in the process.
3. Stuff I'm blissfully unaware of.

Potentially "controversial":

1. The document establishes itself as the normative authority on the semantics of the IR and type system, meaning that comments on semantics found in the .proto files take a back seat and are not normative.
2. It raises the importance of explicitly defining operator sets to a point that no framework actually implements the specification, as far as I know.
3. Explicit mention of the existence of two variants of ONNX.

I probably forgot a number of things that belong on these lists.

In reviewing this, please pay attention not only to whether the text accurately reflects the intended semantics, but also on word choice details. A normative specification requires precision and clarity, so proper language is essential.

Thanks,

Niklas


docs/IR.md
```diff
@@ -1,127 +1,353 @@
...
 
-Extensibility
--------------
+ONNX currently does not define a sparse tensor type.
 
-ONNX is expected to evolve over time and provides features that enable users to experiment and implement additions before they are added to the specifications.
+#### Tensor shapes
 
-### Metadata
+In addition to element type, tensors have shape. A tensor shape is a list of records that define whether the tensor is a vector, a matrix, or a higher-dimensional value. For example, a 100x100 matrix has the shape [100,100].
```
gramalingam: How about "A tensor shape is a list of dimensions"?

docs/IR.md
```diff
@@ -1,127 +1,353 @@
...
+op_type|string|The symbolic identifier of the operator to invoke.
+domain|string|The domain of the operator set that contains the operator named by the op_type.
+attribute|Attribute[]|Named attributes, another form of operator parameterization, used for constant values rather than propagated values.
+doc_string|string|A human-readable documentation for this value. Markdown is allowed.
 
 Edges in the computation graph are established by outputs of one node being referenced by name in the inputs of a subsequent node.
 
-Names of tensor values flowing through the graph are unique - a particular name can reference either a graph input or an output of a single node. Reuse of names for several outputs is not allowed.
+Node inputs MAY also refer to graph inputs and initializers. The outputs of a given node MAY introduce new names into the graph, name a graph output, or coincide with the outputs of other nodes. Thus, using overlapping output names, two nodes MAY compute the same output value. Graph outputs MAY NOT be used to establish data dependency edges in the graph by being named as node inputs.
```
gramalingam: I don't  understand this ... two different node-outpus cannot have the same name, right?

docs/IR.md
NiklasGustafsson: Yes, I agree that it's better.

docs/IR.md
gramalingam: The last sentence "Graph outputs MAY NOT ..." is unclear. Every graph output X must also appear as the output of some node (or must refer to a graph input). This X could be legally used as an input in some other node as well. I guess the sentence is trying to clarify that no dependence edge is created from the graph-output to such an input (in this case)? But it seems to imply the use of X as an input of some node is not valid. Since dependency edges are created only between nodes, perhaps this clarification is not necessary?

NiklasGustafsson: I think this is clumsy wording on my part. 

The motivation was the If() operator, where you would expect the same outputs, or some overlap of them, to be computed by each of the graph attributes, but it's really the operator outputs that are computed by two graphs, not by two nodes, and graph outputs would be associated with If() outputs positionally, not by name, as far as I understand.

NiklasGustafsson: Regarding your second comment -- we need to specify that node inputs cannot depend on a graph output that hasn't already been computed by another node. So, I think some clarification is necessary, but you are right in that the current one seems imprecise.

docs/IR.md
```diff
@@ -1,127 +1,353 @@
...
+The types of the inputs and outputs of a graph must be specified.
 
 #### Optional Inputs
 
-Some Operators have inputs that are marked as optional. There are two ways to leave an optional input unspecified. The first is to simply not provide that input. However, this is not always possible - for example, if you wish to leave the fourth input unspecified, but still provide a value for the fifth input. Therefore, any input with a name of the empty string is treated as an unspecified optional input.
+Some operators have inputs that are marked as optional. There are two ways to leave an optional input unspecified. The first is to simply not provide that input. However, this is not always possible - for example, if you wish to leave the fourth input unspecified, but still provide a value for the fifth input. Therefore, any input with a name of the empty string is treated as an unspecified optional input.
```
gramalingam: Both inputs and outputs can be optional.

docs/IR.md
```diff
@@ -247,10 +249,16 @@ The representation distinguishes between two kinds of values: attribute values,
...
+ 
+The last input or output of an operator MAY be marked as variadic. For each variadic operator input, one or more node inputs must be specified. For each variadic operator output, one or more node outputs must be specified. For example, the operator 'Max()' can be used to compute the maximum of a varying number of input values.
+
+#### Optional Inputs and Outputs
 
 Some operators have inputs that are marked as optional. There are two ways to leave an optional input unspecified. The first is to simply not provide that input. However, this is not always possible - for example, if you wish to leave the fourth input unspecified, but still provide a value for the fifth input. Therefore, any input with a name of the empty string is treated as an unspecified optional input.
 
+Some operators have outputs that are optional. This means that an operator, depending on its input parameters and/or attributes, may not compute the corresponding value. Each node referring to such an operator must have knowledge of the circumstances under which output is computed or not, and provide a name for the output when it is computed. Failure to do so will result in the output not being captured.
```
wschin: Consider LSTM with output_sequence=False as an example. Does "provide a name for the output when it is computed" mean that we can assign an empty string to LSTM's first output in this case?

NiklasGustafsson: Optional outputs is perhaps the murkiest part of the semantics, hardest to explain, I think. 

Optional inputs make sense -- they don't have to be provided? 

Optional outputs, however -- who is it that has an option, the caller or the operator? In other words, is it that the operator has the option of not computing the value, or the model graph that has the option of not picking it up? I believe the former is what's I've heard described, the operator may not compute it. 

If that is correct, then looking at my wording above, the text is focusing on the wrong issue -- it's not about missing to pick up a value that is computed, the issue is giving a name to an output that is not actually computed, and then using it as input to something else.

wschin: For LSTM/RNN/GRU, I think the existence of optional outputs (or inputs) means that an operator can have multiple static configurations because whether the first output is computed or not is determined by a user-specified attribute. Therefore, the user of such an operator should be in charge of choosing the right outputs. If she turns output_sequence on, the first output name would be a meaningful string. If not, she can use an empty string. My point is that as long as an output configuration is (implicitly) determined by an user, using an empty string is better than setting up a flag+provide an uncomputed tensor name.

gramalingam: Yes, the optional outputs part is somewhat murky. My interpretation is that

(a) Generic (operator-independent) constraints: The model may omit an output (using an empty string or completely omitting it in the case of trailing outputs) only if that output is marked as optional in the operator specification.

(b) There will typically be further operator-specific constraints on when omitting an output is valid, which is currently captured only in the informal operator description.

gramalingam: Regarding @NiklasGustafsson 's question as to who has the option: I believe it is a contract between both the caller and the operator, which unfortunately is only informally captured in the operator description. Since this part is not formally captured, "model validation" cannot check these aspects of the contract (e.g., such as "an output is computed iff a specific attribute has a specific value"). This, in turn, means that the semantics of what happens during a model's execution if this contract is violated is undefined.

NiklasGustafsson: This raises a question -- if a node fails to give a name to an output that is computed (optional or not), what is the failure mode? It seems to me that giving a name to an output that is __not__ computed is definitely an error, while the reverse isn't as obviously wrong. I guess that some implementations could have trouble with both.

wschin: @NiklasGustafsson , what do you mean by failure mode? Is it "missing output tensor name"? Also, why assigning an output name to an uncomputed output is wrong? I think it's fine if no other operator refers to that undefined output. Do you mean "forgetting assigning a name to a computed output is wrong"?

gramalingam: Giving a name to an output that is not computed: this could happen even with non-optional outputs. When this named output is subsequently used in a different node, it would pretty much behave like the use of an uninitialized variable (the corresponding tensor memory could have random values). (Obviously, this depends on the runtime implementation.) Unfortunately, we cannot always catch this statically.

Failing to give a name for an output that is not optional is easier to handle: model validation can catch this, and a runtime can avoid initiating execution. Even when it can't be caught in model validation, it can be caught at runtime. How this exception is reported to the user is implementation-dependent. I guess ONNX could try to standardize the exception behavior in such cases.

NiklasGustafsson: By "failure mode," I mean how does an error manifest itself at runtime? 

If an output (optional or not) is not named, it cannot be used as an input or graph output. What downstream error can arise from that? (I'm not saying there can't be one, it's just not clear to me what it could be.)

By assigning a name to an output that is __not__ computed, errors can arise, depending on the runtime implementation. Since such a name can be used as input to other operators, or as a graph output, uninitialized values may be propagated, which leads to further errors and, in the worst case, crashes. Failing to name an output that is computed could be an indication that the model is incorrectly constructed, which is worth at least a warning during validation, but it doesn't seem to be able to cause corruption of data, and if a model doesn't need the output, then why make it assign a name?

Just asking questions...

gramalingam: Yes, for the second part, we can naturally extend the language to allow callers to indicate that they don't care about a particular output, e.g., perhaps by using a special name like "_". (But my understanding is that this scenario is not what the current optional-output mechanism targets.)

NiklasGustafsson: OK, I'll change the text to reflect this.

wschin: I think there are two different questions. Below I summarize our thread.

1. How to specify an operator's optional output when it's optional? I assume we all agree that it should be an empty string. If my assumption is correct, why do we have output_sequence flag in LSTM/GRU/RNN? It looks a bit redundant to me because it's implicitly determined by the first output's name (empty string --> output_sequence=False, meaningful string --> output_sequence=True). 

2. Is it legal if one doesn't provide a name (to the operator) for a computed output? Yes, if it's optional (equivalent to assigning an empty string). No, if it's not optional. @NiklasGustafsson

wschin: Let me add a table for all possible output configurations. We can use it for further discussion.

| Optional  | Named (non-empty name) |  Computed  |  Status  |
| ------------- | ------------- | ------------- | ------------- |
|  Y  |  Y  |  Y  |  OK  |
|  Y  |  Y  |  N  |  Error  |
|  Y  |  N  |  Y  |  Error  |
|  Y  |  N  |  N  |  OK  |
|  N  |  Y  |  Y  |  OK  |
|  N  |  Y  |  N  |  Error  |
|  N  |  N  |  Y  |  Error  |
|  N  |  N  |  N  |  Error  |

NiklasGustafsson: Thanks for the summary. I believe that the corrected text reflects the rules of that table.

docs/IR.md
```diff
@@ -324,7 +331,11 @@ For example, a NxM matrix would have the shape list [N,M].
 
 The name of each dimension variable MUST adhere to C identifier syntax.
 
-Dimension variables are scoped to the declaration (graph signature, node signature, or single operator declaration) that they appear in. Thus, any given name denotes the same value within a declaration, allowing a declaration to describe how the shapes of inputs and outputs are related. For example, a graph that performs matrix cross-product may be defined as taking two inputs of shape [K,M] and [M,N], and produce an output of shape [K,N].
+Dimension variables are scoped to the declaration that they appear in. For graph inputs and outputs, the graph itself is the declaration. Dimension variables appearing in a graph's 'value_info' record are scoped to the value. Consequently, any name that is repeated denotes the same value within a declaration, allowing a declaration to describe how the shapes of inputs and outputs are related.
```
wschin: Does it mean that the dimensions defined by a 'value_info"s are mutually independent even if some of them are identical strings?

docs/IR.md
```diff
@@ -1,127 +1,372 @@
...
+since_version|int64|The version of the operator set when this operator was introduced.
+status|OperatorStatus|One of ‘EXPERIMENTAL’ or ‘STABLE.’
+doc_string|string|A human-readable documentation string for this operator. Markdown is allowed.
+
+The version value MUST be the same value as the operator set version when the operator was first published. Subsequent versions of the operator set MUST NOT alter the signature or semantics of the operator once published as STABLE.
+
+The ‘status’ property indicates whether the syntax, semantics, or presence of the operator is in an experimental or stable stage. Once an operator is published as STABLE, it’s syntax and semantics MUST NOT change in subsequent versions of the operator set.
+
+There are two distinct ways to pass information to operators – inputs and attributes. The latter are used for values that are literal constants in the graph, while the former represent graph inputs or values computed elsewhere in the graph. This distinction may be highly relevant to achieving good performance for some implementations, while completely irrelevant to others.
```
wschin: I think attributes should be non-trainable parameters? What do you think?

docs/IR.md
```diff
@@ -1,127 +1,372 @@
...
+strings|byte[][]|List of strings
+tensors|Tensor[]|List of tensor values
+graphs|Graph[]|List of graphs
+
+The properties ‘name’ and ‘type’ are required on all attributes, and ‘doc_string’ SHOULD be used on all attributes. An attribute MUST have only one of the value-carrying properties.
 
 #### Values
 
-The representation distinguishes between two kinds of values: attribute values, which are statically known, and runtime values. The type of values permitted in the two cases are different. The permitted types of attribute values are indicated by the enumeration `AttributeType`, while the permitted types of runtime values are described by `TypeProto`. 
+The representation distinguishes between two kinds of values: attribute values, which are statically known, and runtime values. The types of values permitted in the two cases are different.
```
wschin: It'd be better to give some examples here. attribute values ---> attribute values (e.g., an operator's attributes); runtime values ---> runtime values (e.g., inputs and outputs of an operator). It's not extremely clear to me what runtime values are. As a standard, you might want this to be more specific.

docs/IR.md
```diff
@@ -1,127 +1,372 @@
...
+domain|string|The domain of the operator set that contains the operator named by the op_type.
+attribute|Attribute[]|Named attributes, another form of operator parameterization, used for constant values rather than propagated values.
+doc_string|string|A human-readable documentation for this value. Markdown is allowed.
 
 Edges in the computation graph are established by outputs of one node being referenced by name in the inputs of a subsequent node.
 
-Names of tensor values flowing through the graph are unique - a particular name can reference either a graph input or an output of a single node. Reuse of names for several outputs is not allowed.
+The outputs of a given node introduce new names into the graph. The values of node outputs are computed by the node's operator. Node inputs MAY refer to node outputs, graph inputs, and graph initializers. When the name of a node output coincides with the name of a graph output, the graph output's value is the corresponding output value computed by that node.
+
+The graph must use single-static assignment for all node outputs. All node output namess MUST be unique within the graph.
```
wschin: Can one variable be used as an output for an operator and an input for another operator? If yes, you might want to add  a tiny note.

docs/IR.md
```diff
@@ -1,127 +1,372 @@
...
 
-Some Operators have inputs that are marked as optional. There are two ways to leave an optional input unspecified. The first is to simply not provide that input. However, this is not always possible - for example, if you wish to leave the fourth input unspecified, but still provide a value for the fifth input. Therefore, any input with a name of the empty string is treated as an unspecified optional input.
+For each variadic operator input, one or more node inputs must be specified. For each variadic operator output, one or more node outputs must be specified. 
 
-Built-in Operators and Standard Data Types
-------------------------------------------
+#### Optional Inputs and Outputs
 
-### Operators
+Some operators have inputs that are marked as optional, which means that a referring node MAY choose not to provide a value for that input.
```
wschin: "a referring node MAY choose not to provide a value for that input" or "a user of such an operator MAY choose not to provide values for some of optional inputs"?

docs/IR.md
```diff
@@ -1,127 +1,372 @@
...
+Some operators have outputs that are optional, which means that an operator, depending on its input parameters and/or attributes, MAY choose not to compute all its declared outputs. 
+
+There are two ways to leave an optional input or output unspecified: the first, available only for trailing inputs and outputs, is to simply not provide that input; the second method is to use an empty string in place of an input or output name.
+
+Each node referring to an operator with optional outputs MUST provide a name for each output that is computed and MUST NOT provide names for outputs that are not computed.
+
+
+## Standard data types
+
+There are two official ONNX variants; the main distinction between the two is found in the supported types. As mentioned earlier, the __ONNX__ definition recognizes only tensors as input and output types, while the Classical Machine Learning extension. __ONNX-ML__, also recognizes sequences and maps.
```
wschin: I believe they also support different operator sets. For example, "LinearClassifier" is not a part of ONNX but belongs to ONNX-ML. Please correct me if I am wrong. Thanks!

docs/IR.md
```diff
@@ -1,127 +1,372 @@
...
+Which is referenced by the Tensor type message:
+
+```
+  message Tensor {
+    optional TensorProto.DataType elem_type = 1;
+    optional TensorShapeProto shape = 2;
+  }
+```
+
+The empty list of dimension sizes, [], is a valid tensor shape, denoting a zero-dimension (scalar) value. A zero-dimension tensor is distinct from a tensor of unknown dimensionality, which is indicated by an absent 'shape' property in the Tensor record.
```
wschin: Would you want to add a note here to explain that symbolic dimension variable is not unknown but variable-length? It's kind of confusing to my tiny brain.

docs/IR.md
NiklasGustafsson: Wei-Sheng,

I don't quite understand what you are proposing. Are you saying we shouldn't call them "attributes"? If so, I disagree -- we've been using that name since the beginning, and it has little to do with trainability.

docs/IR.md
```diff
@@ -1,127 +1,372 @@
...
+For each variadic operator input, one or more node inputs must be specified. For each variadic operator output, one or more node outputs must be specified. 
 
-Built-in Operators and Standard Data Types
-------------------------------------------
+#### Optional Inputs and Outputs
 
-### Operators
+Some operators have inputs that are marked as optional, which means that a referring node MAY choose not to provide a value for that input.
+
+Some operators have outputs that are optional, which means that an operator, depending on its input parameters and/or attributes, MAY choose not to compute all its declared outputs.
```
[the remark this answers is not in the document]
NiklasGustafsson: Fixed

docs/IR.md
NiklasGustafsson: Yes, that is an important point. Thanks!

docs/IR.md
NiklasGustafsson: What I'm trying to say here, perhaps not precisely enough, is that for graph inputs and outputs, dimension variables are scoped to the collection of inputs and outputs. This is so that we can associate dimension information between inputs and outputs. For value_info records that are not inputs or outputs, I believe we don't force association across values, but maybe that's a misunderstanding on my part.

docs/IR.md
wschin: Sorry for not making it clear. My original opinion is to have a specific rule to distinguish "inputs" and "attributes" from each other. A "constant" is not ultimately clear to me because for example, LSTM's recursion matrices are constant under test mode but they are "inputs." Not sure if changing "literal constants" to "literal constants at inference stage and training stage" is helpful?

docs/IR.md
wschin: I believe you're right but this might need help from all praters to confirm. To make this concept clearer, you might want to change **"Dimension variables appearing in a graph's 'value_info' record are scoped to the value"** to **"Dimension variables appearing in a graph's 'value_info' record are scoped to the value so they don't force any association across values which are not graph's inputs or outputs"**

docs/IR.md
wschin: Sorry for causing ambiguation. When looking at BatchNormalization where "mean" appears in its input and output lists, I was considering if using a tensor for both of an operator's input and output is legal. If not legal, I would like to add a sentence to explicitly tell users not to do so.

NiklasGustafsson: Doing so would, it seems to me, create a cycle in the graph from the node's output to its input, which is explicitly disallowed in the spec.

wschin: I know it's not allowed because it's a loop, but I would like to emphasize this point.

NiklasGustafsson: In-place computation violates the single static assignment rule, so that is already covered.

wschin: "single static assignment" is a magic word to some poor users like me. My background is applied math.

docs/IR.md
```diff
@@ -57,6 +57,7 @@ The valid IR versions is defined by an enumeration, which currently has the foll
 
 Operator sets use a simple number as the version number. This specification does not provide guidance on what versioning scheme model producers should be using.
 
+More details on conventions and best practices for versioning of IR, pperator sets, and models can be found in [Versioning](Versioning.md).
```
wschin: pperator --> operator

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+domain|string|The domain of the operator set that contains the operator named by the op_type.
+attribute|Attribute[]|Named attributes, another form of operator parameterization, used for constant values rather than propagated values.
+doc_string|string|A human-readable documentation for this value. Markdown is allowed.
 
 Edges in the computation graph are established by outputs of one node being referenced by name in the inputs of a subsequent node.
 
-Names of tensor values flowing through the graph are unique - a particular name can reference either a graph input or an output of a single node. Reuse of names for several outputs is not allowed.
+The outputs of a given node introduce new names into the graph. The values of node outputs are computed by the node's operator. Node inputs MAY refer to node outputs, graph inputs, and graph initializers. When the name of a node output coincides with the name of a graph output, the graph output's value is the corresponding output value computed by that node.
+
+The graph MUST use single static assignment for all node outputs, which means that all node output names MUST be unique within a graph.
```
wschin: According to some discussions [here](https://github.com/onnx/onnx/pull/755), a graph can have multiple identical output names. Could we clarify this point somewhere, saying that node outputs are different from graph outputs?

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+
+Input and outputs are positionally associated with operator inputs and outputs. Attributes are associated with operator attributes by name.
+
+They have the following properties:
+
+Name|Type|Description
+|---|---|---|
+name|string|An optional name of the node, used for diagnostic purposes only.
+input|string[]|Names of the values used by the node to propagate input values to the node operator. It must refer to either a graph input or a node output.
+output|string[]|Names of the outputs used by the node to capture data from the operator invoked by the node. It either introduces a  value in the graph or refers to a graph output.
```
wschin: introduces a value ---> remove an extra space between "a" and "value"

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+#### Attributes
+
+Attributes, which represent literal constants passed to operators from nodes, have the following properties:
+
+Name|Type|Description
+|---|---|---|
+name|string|The name of the attribute. Must be unique among attributes, inputs, and outputs for any given operator and node.
+doc_string|string|A human-readable documentation for this value. Markdown is allowed.
+type|AttributeType|The type of the attribute, determining which of the remaining fields is used to hold the value of the attribute.
+f|float|A floating-point value.
```
houseroad: More specific, a 32 bit floating point value?

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+i|int64|An integer value
```
houseroad: A 64 bit signed integer value?

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
-ONNX is an open specification that consists of the following components:
+__Purpose__
+
+This document contains the normative specification of the semantics of ONNX. The .proto and .proto3 files found under the ‘onnx’ folder form the normative specification of its syntax. Commentary found in the .proto and .proto3 files are intended to improve readability of those files, but are not normative if they conflict with this document. Such conflicts should be reported as documentation bugs.
 
-1)  A definition of an extensible computation graph model
+__Notes on model validation__
 
-2)  Definitions of built-in operators and standard data types
+A [Python-based tool](../onnx/checker.py) is available to perform general validation of models against this specification.
```
houseroad: checker.py is a python wrapper, actually the real implementation is in C++.

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+
+ONNX does not pre-suppose or imply any particular method of runtime implementation. 
+
+For example, an implementation may consist of a rich runtime which interprets the model; it may be a code generator that translates the model in its entirety to executable code for some target programming language; it may be a hardware implementation; it may be a combination of two or three of those. 
+
+Nothing in this specification should be construed as advocating one implementation approach over any other; any comments on the inner workings of concrete implementations are to be interpreted as examples.
+
+## ONNX Versioning
+
+Versioning features in several places in ONNX -- the IR specification itself, the version of a model, and the version of an operator set. Futhermore, each individual operator indicates which version of its containing operator set it was introduced or stabilized in.
```
houseroad: Shall we also talk about release version here?

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+### Models
 
-ONNX specifies the portable, serialized format of the computation graph. It may not be the form a framework chooses to use and manipulate. For example, a framework may keep the graph in memory in a format that it finds more efficient to manipulate for optimization passes. We make use of protobuf2 (with oneof, added in protobuf 2.6.1) for the serialized format.
+The top-level ONNX construct is a ‘Model.’
 
-### Model
+The main purpose of the model structure is to associate metadata with a graph, which is what contains all the executable elements. The metadata is used when first reading the model file, giving an implementation the information that it needs in order to determine whether it will be able to execute the model, generate logging messages, error reports, etc. Further, the metadata is useful to tools, such as IDEs and model galleries, which need it for informing humans about a given model’s purpose and characteristics.
 
-The top-level container object in an ONNX model file is a ModelProto object that combines a computation graph with the following metadata properties.
+Each model has has the following components:
```
houseroad: two `has` here

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+
+Each operator set SHALL be defined in a separate document, also using protobuf as the serialization format. How operator set documents are found at runtime is implementation-dependent.
+
+__Note: As of the publication of this document, no ONNX implementation is known to process operator set documents.__
+
+The properties of an operator set are:
+
+Name|Type|Description
+|---|---|---|
+magic|string|The value ‘ONXXOPSET’
```
houseroad: ONNXOPSET

docs/IR.md
NiklasGustafsson: OK, will explain.

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
 
-|Name|Type|Format|Description|
-|----|----|------|-----------|
-|ir_version|int64||The version of the IR format specification|
-|opset_import|||Custom operator sets imported by the model|
-|domain|string|Valid DNS name|A namespace for the model, following the style of package names, that is, reverse DNS domain name.|
-|model_version|int64||The version of the model|
-|producer_name|string||Name of framework/tool that generated this model|
-|producer_version|string||Version of the framework/tool that generated this model|
+ Models MUST specify a domain and use reverse domain names based on the responsible organization's identity, the same convention that is traditionally used for naming Java packages.
```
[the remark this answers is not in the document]
NiklasGustafsson: So, what should be changed in the text? That isn't clear from your comment. Should I insert a 'SHOULD' before 'use reverse domain names'?

docs/IR.md
NiklasGustafsson: Thanks

docs/IR.md
houseroad: Attributes are statically determined, and input can be dynamically computed. This is orthogonal to training and reference.

houseroad: Although calling a graph attribute constant literal sounds a little bit weird to me.

NiklasGustafsson: Yes, I agree that "literal constants" is redundant, I'll fix that.
I don't understand your comment about 'training and reference,' where does the text mention that with regard to attributes or inputs?

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+
+There are two distinct ways to pass information to operators – inputs and attributes. The latter are used for values that are literal constants in the graph, while the former represent graph inputs or values computed elsewhere in the graph. This distinction may be highly relevant to achieving good performance for some implementations, while completely irrelevant to others.
 
 ### Graphs
 
-Each computation dataflow graph is structured as a list of nodes that form a graph. Each node is a call to an operator. Nodes have zero or more inputs, one or more outputs, and zero or more attribute-value pairs. Additionally, a valid ONNX graph must fulfill the following requirements:
+A serialized graph is comprised of a set of metadata fields, a list of model parameters, and a list of computation nodes.
+
+Each computation dataflow graph is structured as a topologically sorted list of nodes that form a graph, which MUST be free of cycles. Each node represents a call to an operator. Each node has zero or more inputs and one or more outputs.
```
houseroad: SSA is an important property, why it's missing?

NiklasGustafsson: It's mentioned under 'Nodes' farther down in the document. Does it make the graph explanation clearer if it's included here, too?

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+|---|---|---|
+name|string|The name of the value/parameter.
+type|Type|The type of the value.
+doc_string|string|A human-readable documentation for this value. Markdown is allowed.
 
-|Name|Type|Format|Description|
-|----|----|------|-----------|
-|name|string|Valid C identifier|A name for the model.|
-|doc_string|string|Free form|A human-readable documentation string intended to summarize the purpose of the model. Markdown is allowed.|
+Each graph MUST specify a name.
```
houseroad: Why do we make graph's name required?

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+strings|byte[][]|A list of strings
+tensors|Tensor[]|A list of tensor values
+graphs|Graph[]|A list of graphs
+
+The properties ‘name’ and ‘type’ are required on all attributes, and ‘doc_string’ SHOULD be used on all attributes. An attribute MUST have only one of the value-carrying properties.
 
 #### Values
 
-The representation distinguishes between two kinds of values: attribute values, which are statically known, and runtime values. The type of values permitted in the two cases are different. The permitted types of attribute values are indicated by the enumeration `AttributeType`, while the permitted types of runtime values are described by `TypeProto`. 
+The representation distinguishes between two kinds of values: attribute values, which are statically known, and runtime values. The types of values permitted in the two cases are different.
```
houseroad: Shall we use a different name? Here is quite ambiguous... We describe the values namespace in the previous section, here we keep using values, but we are talking about the attribute values and input/output values.

docs/IR.md
NiklasGustafsson: Runtimes are not the only consumers of models; some tools will need something to identify the model. It's also important to identify models in logging, error messages, etc. In the future, if and when we get to supporting model composition, names will be essential for that ability, too.

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
+The following data types are supported by ONNX for inputs and outputs of graphs and nodes as well as the the initializers of a graph.
+
+Primitive numeric, string, and Boolean types MUST be used as elements of tensors. Maps and sequences MUST contain tensors as values.
+
+### Tensor Element Types
 
-See the [operator documentation](Operators.md) for details.
+|Group|Types|Description| 
+|---|---|---|
+Floating Point Types|float16, float32, float64|Values adhering to the IEEE 754-2008 standard representation of floating-point data.
```
houseroad: Here, we probably should explicitly say float is float32.

docs/IR.md
NiklasGustafsson: Not sure where the ambiguity comes from. The text is referring to 'attribute values' and 'runtime values,' and elaborates on the two. But clarity is essential, so if there's a better formulation, please suggest one. It seems unambiguous to me, but I wrote it, so that's not the bar.

docs/IR.md
```diff
@@ -1,127 +1,382 @@
...
-- The nodes list MUST be in topologically sorted order, meaning if input of node `N2` is the output of node `N1`, `N2` must appear after `N1` in the nodes list.
+|Name|Type|Description|
+|---|---|---|
+name|string|The name of the model graph.
+node|Node[]|A list of nodes, forming a partially ordered computation graph based on input/output data dependencies.
+initializer|Tensor[]|A list of named tensor values, used to specify default values for some of the inputs of the graph. Each initializer value is associated with an input by name matching.
+doc_string|string|A human-readable documentation for this model. Markdown is allowed.
+input|ValueInfo[]|The input “parameters” of the graph, possibly initialized by a default value found in ‘initializer.’
+output|ValueInfo[]|The output parameters of the graph. Once all output parameters have been written to by a graph execution, the execution is complete.
+value_info|ValueInfo|__TODO: Write this description__
```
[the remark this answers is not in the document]
NiklasGustafsson: Thanks.

NiklasGustafsson:
It must have been the merge after the last 'Update Branch' that resulted in those edits.

NiklasGustafsson:
Lu,
Thank you for all your review comments. My latest commit should have addressed most of them, excluding those where I had a follow-up question for you.

Thanks,
Niklas

docs/IR.md
houseroad: I guess we can explain more in Versioning.md

docs/IR.md
houseroad: But we already have the model name, I am totally confused why we need both graph name and model name. Especially, each model can only contain one graph...

docs/IR.md
houseroad: I think you can move the attribute values to the previous subsection attributes, and change the names of this subsection to Operator Input and Output

docs/IR.md
NiklasGustafsson: I don't find a name property in ModelProto. The model name is stored in the graph message.

From the current onnx.in.proto, commenting on the domain name of a model:

  // Domain name of the model.
  // We use reverse domain names as name space indicators. For example:
  // `com.facebook.fair` or `com.microsoft.cognitiveservices`
  //
  // Together with `model_version` and GraphProto.name, this forms the unique identity of
  // the graph.
  optional string domain = 4;

houseroad:
@NiklasGustafsson thanks for the prompt update. I have added more comments. :-)

docs/IR.md
NiklasGustafsson: That makes sense. I intend to tackle Versioning.md next, by the way.

PR #725, opened 2018-04-05
Shape inference for Cast

anderspapitto:


anderspapitto:
built on top of https://github.com/onnx/onnx/pull/655

onnx/defs/schema.cc
```diff
@@ -584,4 +584,54 @@ size_t ReplaceAll(std::string& s, const char* from, const char* to) {
   }
   return numReplaced;
 }
+
+TensorProto::DataType datatypeFromString(const std::string& type) {
+  if (type == "FLOAT") {
+    return TensorProto::FLOAT;
+  }
+  if (type == "FLOAT") {
```
houseroad: 2 FLOAT? C++ may miss some float, so we need a second check to catch them?

anderspapitto: float is a very important datatype and needs extra attention

onnx/defs/shape_inference.h
```diff
@@ -32,6 +32,18 @@ inline bool getRepeatedAttribute(InferenceContext& ctx,
 
 }
 
+inline bool hasExactlyNInputTypes(InferenceContext& ctx, int n, const std::string& opname) {
+  if (ctx.getNumInputs() != n) {
+    throw std::runtime_error(opname + " has wrong number of inputs");
```
gramalingam: Why not return false, instead of throwing an exception? E.g., I see this being invoked for Concat, where this may not be an error.

onnx/defs/tensor/defs.cc
```diff
@@ -136,7 +171,70 @@ ONNX_OPERATOR_SCHEMA(Concat)
...
-        "Constrain output types to float tensors.");
+        "Constrain output types to float tensors.")
+    .ShapeInferenceFunction([](InferenceContext& ctx) {
+        if (!hasExactlyNInputTypes(ctx, 1, "Concat")) {
+          return;
+        }
+
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+
+        if (ctx.getNumInputs() == 0) {
```
gramalingam: I don't understand: there's a check earlier for exactly 1 input type.

onnx/defs/tensor/defs.cc
```diff
@@ -99,7 +118,19 @@ Takes a tensor as input and outputs an 1D int64 tensor containing the shape of t
     .TypeConstraint(
         "T1",
         {"tensor(int64)"},
-        "Constrains output to int64 tensor.");
+        "Constrains output to int64 tensor.")
+    .ShapeInferenceFunction([](InferenceContext& ctx) {
+        if (!hasExactlyNInputTypes(ctx, 1, "Cast")) {
```
gramalingam: "Cast" -> "Shape"

onnx/defs/shape_inference.h
anderspapitto: well, this is just a helper that throws an exception. the error is with concat - i'll remove that usage

onnx/defs/tensor/defs.cc

onnx/defs/tensor/defs.cc

onnx/defs/tensor/defs.cc
```diff
@@ -156,7 +250,38 @@ ONNX_OPERATOR_SCHEMA(Split)
...
+          }
+          const auto& splitDim = ctx.getInputType(0)->shape().dim(axis);
+          if (!splitDim.has_dim_value()) {
+            return;
+          }
+          int splitDimValue = splitDim.dim_value();
+          int chunkSize = splitDimValue / ctx.getNumOutputs();
+          int leftOver = splitDimValue - (chunkSize * ctx.getNumOutputs());
+          for (int i = 0; i < ctx.getNumOutputs(); i++) {
+            split.push_back(i < leftOver ? chunkSize + 1 : chunkSize);
```
gramalingam: The spec above doesn't actually say what the behavior is when leftover is non-zero. May be a good idea to extend the spec to do that (though that should probably be a separate PR since it affects the spec).

onnx/defs/tensor/defs.cc
```diff
@@ -327,7 +452,31 @@ Example 2:
...
+        int r = ctx.getInputType(0)->shape()->dim_size();
+        int q = ctx.getInputType(1)->shape()->dim_size();
+
+        int out_rank = q + r - 1;
+
+        if (out_rank == 0) {
+          ctx.getOutputType(0)->mutable_shape();
+        }
+        for (int i = 0; i < out_rank; ++i) {
+          ctx.getOutputType(0)->mutable_shape()->add_dim();
```
gramalingam: Is this supposed to be initialized to some value from the input dims?

onnx/defs/tensor/defs.cc
anderspapitto: ++ will make a PR for the spec

onnx/defs/tensor/defs.cc
anderspapitto: not at the moment - this is more "rank inference" than "shape inference". It can be enhanced in a future diff

gramalingam: It would be safer to initialize the dim with the empty string param than leave it uninitialized (neither value nor param set).

anderspapitto: I guess representing this is still a little up in air - but see the comment I just wrote at https://github.com/onnx/onnx/pull/681#issuecomment-381221996

gramalingam: Okay. As long as the meaning/representation is standardized and well-documented, it's fine.

PR #727, opened 2018-04-06
Change the "to" attribute of Cast operator to of type int

bddppq:


onnx/defs/tensor/old.cc
```diff
@@ -117,4 +117,56 @@ from the input tensor).)DOC")
...
+the converted type. The 'to' argument must be one of the data types specified
+in the 'DataType' enum field in the TensorProto message.
+
+NOTE: Casting to and from strings is not supported yet.
+)DOC")
+    .Attr(
+        "to",
+        "The data type to which the elements of the input tensor are cast."
+        "Strictly must be one of the types from DataType enum in TensorProto",
+        AttributeProto::STRING)
```
houseroad: Here is still string?

onnx/examples/make_model.ipynb
```diff
@@ -71,7 +71,7 @@
       "  }\n",
       "}\n",
       "opset_import {\n",
-      "  version: 6\n",
+      "  version: 7\n",
```
houseroad: Why do we only change this file?

houseroad: I was wondering why only this notebook needs changes. I think I know why now. Because, in other cases, we only load the existing model, not making. Only in this case, new model is created. :-)

linkerzhang:
@ebarsoum is going to send a proposal of bumping op_set version together with ONNX release, which makes lots of sense, I think. For now, between 2 ONNX releases, say, 1.0 and 1.1 or 1.1 and 1.2 in future, there may be many op_set version created due to breaking change introduced in between. This makes backend tough to follow and to ensure backward compatibility (too many version needs to be supported).

bddppq:
@linkerzhang @ebarsoum Please review this schema change at your convenience

bddppq:
I think reducing the number of op_set bump makes senses. But I think we should freeze a version together with a release and bump the version number in master right **after** making a release.
Adding @ezyang to

linkerzhang:
YES. basically, after an ONNX release, if there's an op breaking change, bump the op_set version once and keep the version till next ONNX release.

houseroad:
I would like to have two types of opset_import version:
1) develop (all odd)
2) release/stable (all even)

For example, we use opset 7 as a develop opset version. It will be bumped until next release, the next release will use opset version 8, and after release cut, we use opset version 9 for development.

The benefit is obvious, we never need to promise any long term support for develop opset version.

ezyang:
This is OK, but I also prefer the even-odd stable-unstable numbering scheme, and to also have the checker warn when a model is tracking an unstable opset version.

Also, I'd like to know what the pain points of opset number bumps are, because they are supposed to be cheap (or, more specifically, it should be no more expensive to have five bumps for five changes, than one bump for five changes), so if they are weirdly expensive then it's important to know, because that affects compliance for real releases as well.

houseroad:
One point is the model zoo. Right now, we maintain the models for each opset. :-)

bddppq:
@ezyang I feel two pain points:
1. compliant frontend and backend need to follow each opset and provide backward compatibility
2. it makes onnx afraid of changing any existing schema, delaying the bump to each release gives us a very good staging period.

On the other hand, if we change to tie opset to release, then we should do more frequent (and regular) releases so people know when they can expect the backward compatibility guarantee.

ezyang:
OK, let's do it.

onnx/defs/tensor/defs.cc
```diff
@@ -18,7 +19,7 @@ NOTE: Casting to and from strings is not supported yet.
         "to",
         "The data type to which the elements of the input tensor are cast."
         "Strictly must be one of the types from DataType enum in TensorProto",
-        AttributeProto::STRING)
+        AttributeProto::INT)
```
anderspapitto: as we discussed in person, I believe that the correct way to do this is to add a case to the attribute type enum, which corresponds to TensorProto::DataType, rather than using an integer that happens to line up with the integer values of the DataType enum.

The advantages are that (1) it's clearer, and (2) tooling which understands protobuf will automatically work - for example the `str()` of a `ModelProto` in python will automatically say `TensorProto.FLOAT` instead of `4`.

But if i'm the lone voice here then I won't hold it up

bddppq: Having the visualization tooling support is indeed more ideal, however the cost is not small:
1. each backend needs to be updated to acknowledge a new AttributeProto data type
2. existing tools (checker, our c++ IR, schema declarations) needs to be updated to acknowledge this as well

And use cases seem to be small, so far I can only think of Cast and GivenTensorFill (experimental).

houseroad:
We also need to change here https://github.com/onnx/onnx/blob/master/onnx/test/shape_inference_test.py#L87

If not bumping up the opset_version, all the node tests have to be updated, too

PR #729, opened 2018-04-06
Add backend test for upsample

smessmer:


onnx/backend/test/case/node/upsample.py
```diff
@@ -0,0 +1,39 @@
...
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+class Upsample(Base):
+
+    @staticmethod
+    def export():
```
houseroad: Nit: export ==> export_nearest

onnx/backend/test/case/node/upsample.py
```diff
@@ -0,0 +1,39 @@
...
+
+        output = np.array([[[
+            [1, 1, 1, 2, 2, 2],
+            [1, 1, 1, 2, 2, 2],
+            [3, 3, 3, 4, 4, 4],
+            [3, 3, 3, 4, 4, 4],
+        ]]], dtype=np.float32)
+
+        expect(node, inputs=[data], outputs=[output],
+               name='test_upsample')
```
houseroad: Nit: 'test_upsample' ==> 'test_upsample_nearest'

PR #730, opened 2018-04-06
Fix incorrect package output paths

bddppq:
The bug can cause `pip uninstall` to remove the user's entire site-packages directory


anderspapitto:
oops

PR #738, opened 2018-04-10
Update PythonAPIOverview.md

dzhulgakov:


docs/PythonAPIOverview.md
```diff
@@ -4,7 +4,8 @@
 ```python
 import onnx
 
-onnx_model = onnx.load('path/to/the/model')
+onnx_model = onnx.load('path/to/the/model.onnx')
+# `onnx_model` is a ModelProto struct
```
bddppq: `struct` -- so c++ ;)

houseroad: or pure c :-P

PR #739, opened 2018-04-10
RNN tests: LSTM, GRU, SimpleRNN

pk-g:
This PR contains an initial round of tests for LSTM, GRU, and SimpleRNN


onnx/backend/test/case/node/lstm.py
```diff
@@ -0,0 +1,94 @@
...
+        W = np.ones((1, 4 * hidden_size, input_size)).astype(np.float32)
+        R = np.ones((1, 4 * hidden_size, hidden_size)).astype(np.float32)
+
+        output = np.array([[0.70377535, 0.70377535, 0.70377535, 0.70377535],
+                           [0.96009213, 0.96009213, 0.96009213, 0.96009213],
+                           [0.99451119, 0.99451119, 0.99451119, 0.99451119]]).astype(np.float32)
+
+        expect(node, inputs=[input, W, R], outputs=[output], name='test_lstm_defaults')
+
+    @staticmethod
```
spandantiwari: Can we add tests for bidirectional LSTMs please?

onnx/backend/test/case/node/lstm.py
```diff
@@ -0,0 +1,94 @@
...
+                           [0.96049958, 0.96049958, 0.96049958, 0.96049958],
+                           [0.99456745, 0.99456745, 0.99456745, 0.99456745]]).astype(np.float32)
+
+        expect(node, inputs=[input, W, R, B], outputs=[output], name='test_lstm_with_initial_bias')
+
+    @staticmethod
+    def export_peepholes():
+            input = np.array([[[1., 2.], [3., 4.], [5., 6.]]]).astype(np.float32)
+
+            input_size = 2
```
spandantiwari: A (few) different combination of input_size and hidden_size would be good to have.

onnx/backend/test/data/node/test_gru_with_initial_bias/model.onnx
```diff
@@ -0,0 +1,30 @@
+backend-test:�
```
spandantiwari: Do we need to check in model.onnx files?

docs/Operators.md
```diff
@@ -2371,6 +2371,72 @@ Other versions of this operator: <a href="Changelog.md#GRU-1">GRU-1</a>
...
+```python
+input = np.array([[[1., 2.], [3., 4.], [5., 6.]]]).astype(np.float32)
+
+input_size = 2
+hidden_size = 4
+custom_bias = 0.1
+
+node = onnx.helper.make_node(
+    'GRU',
+    inputs=['X', 'W', 'R', 'B'],
```
spandantiwari: Are we planning to add more coverage for different activation functions?

onnx/backend/test/data/node/test_gru_with_initial_bias/model.onnx
pk-g: yes, that will be needed by backends that will try to run the tests.

onnx/backend/test/case/node/gru.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import absolute_import
```
snnn: Where do the numbers come from? What if someone's code failed this test? How can the author debug his code? He would wondering, why the result should be such numbers, not the others?

onnx/backend/test/case/node/lstm.py
pk-g: Sure, will add.

onnx/backend/test/case/node/lstm.py
pk-g: The scope of this PR is to provide basic test coverage for forward case, as well as reference numpy implmentation for it. Bidirectional LSTMs shall be added seperately as part of other PRs.

docs/Operators.md
pk-g: The scope of this PR is to provide basic test coverage for forward case, as well as reference numpy implmentation for it. Different variations shall be added seperately as part of other PRs.

onnx/backend/test/case/node/gru.py
pk-g: Please take a look at the updated PR. I removed the hardcoded numbers and instead included the numpy implementations based on formulas provided in ONNX documentation.

docs/Operators.md
spandantiwari: OK.

docs/Operators.md
```diff
@@ -2368,6 +2368,72 @@ Other versions of this operator: <a href="Changelog.md#GRU-1">GRU-1</a>
...
+weight_scale = 0.1
+
+node = onnx.helper.make_node(
+    'GRU',
+    inputs=['X', 'W', 'R'],
+    outputs=['Y'],
+    hidden_size=hidden_size
+)
+
+W = weight_scale * np.ones((1, 3 * hidden_size, input_size)).astype(np.float32)
```
spandantiwari: Can we give capture the 3 as a variable with a suitable name instead of a magic number for the benefit of the reader? Similarly for LSTM tests below we can capture 4 as a variable.

onnx/backend/test/case/node/lstm.py
```diff
@@ -0,0 +1,157 @@
...
+
+            self.X = params[X]
+            self.W = params[W]
+            self.R = params[R]
+            self.B = b
+            self.P = p
+            self.H_0 = h_0
+            self.C_0 = c_0
+        else:
+            raise NotImplementedError()
```
spandantiwari: Bidirectional LSTM?

docs/Operators.md
pk-g: Sure, updated the PR.

onnx/backend/test/case/node/lstm.py
pk-g: I agree that we need tests for bidirectional LSTM, as well as different variations of activations and etc. However, The scope for this PR is to add basic test coverage for forward case and also provide a base for numpy implantations that shall be extended later. Additional tests, which would also required extension of base numpy implementation, shall be added later.

spandantiwari: OK.

PR #741, opened 2018-04-10
Guard type checking numpy imports

bddppq:


houseroad:
I didn't notice the mypy had been merged. Thanks for fixing this.

PR #743, opened 2018-04-10
merge the dummy backend back into the main one

anderspapitto:


onnx/test/test_backend_test.py
```diff
@@ -7,9 +7,13 @@
 import unittest
 import onnx.backend.base
 import onnx.backend.test
+
+import onnx
+from onnx import helper
+from onnx.mapping import NP_TYPE_TO_TENSOR_TYPE
```
bddppq: these imports are not needed anymore

bddppq:
@houseroad model tests are run via `prepare` + `run`, so `run_model` is not needed.

Issue #749, opened 2018-04-12
auto_pad formula (MaxPool)


newling:
From the documentation for MaxPool:

```
SAME_UPPER or SAME_LOWER: 
output_spatial_shape[i] = floor(input_spatial_shape[i] / strides_spatial_shape[i] + 1)
```

surely this should be
```
output_spatial_shape[i] = floor(input_spatial_shape[i] / strides_spatial_shape[i])
```

as when stride is  1, output and input shapes are the same.

fumihwh:
@newling 
nice catch~~~thanks~
I will fix it.

newling:
Thanks!

houseroad:
Fixed

PR #751, opened 2018-04-12
fix docs of pool op

fumihwh:
An alternative fix is change old one's + 1 to + 0.99999999 which is ugly.
```
VALID: output_spatial_shape[i] = floor((input_spatial_shape[i] - kernel_spatial_shape[i]) / strides_spatial_shape[i] + 0.9999999)
SAME_UPPER or SAME_LOWER: output_spatial_shape[i] = floor(input_spatial_shape[i] / strides_spatial_shape[i] + 0.9999999)
```
I think using ceil for SAME and VALID is acceptable. 
#749


jspisak:
@houseroad can you take a look here?

onnx/defs/nn/defs.cc
```diff
@@ -42,8 +42,8 @@ std::function<void(OpSchema&)> PoolOpSchemaGenerator(
 
  `auto_pad` is a DEPRECATED attribute. If you are using them currently, the output spatial shape will be following:
  ```
- VALID: output_spatial_shape[i] = floor((input_spatial_shape[i] - kernel_spatial_shape[i]) / strides_spatial_shape[i] + 1)
- SAME_UPPER or SAME_LOWER: output_spatial_shape[i] = floor(input_spatial_shape[i] / strides_spatial_shape[i] + 1)
+ VALID: output_spatial_shape[i] = ceil((input_spatial_shape[i] - kernel_spatial_shape[i] + 1) / strides_spatial_shape[i]))
```
houseroad: Looks like we have one unnecessary trailing ')'

PR #755, opened 2018-04-13
add eliminate identity optimizer

fumihwh:
@houseroad 
Add eliminate identity optimizer.


bddppq:
Since ONNX requires SSA and we have removed the in-place annotation "consumed_inputs", I don't think we need Identity operator anymore. Shall we remove it? @linkerzhang @SherlockNoMad

houseroad:
@fumihwh can you check why the model converted from tensorflow contains `Identity` operator?

onnx/test/optimizer_test.py
```diff
@@ -19,6 +19,33 @@ def _optimized(self, graph, opts):
         checker.check_model(optimized_model)
         return optimized_model
 
+    def test_identity(self):
```
houseroad: test_eliminate_identity_single_use

onnx/test/optimizer_test.py
```diff
@@ -19,6 +19,33 @@ def _optimized(self, graph, opts):
...
+        identity = helper.make_node("Identity", ["X"], ["Y"])
+        graph = helper.make_graph(
+            [identity],
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (5,))],
+            [helper.make_tensor_value_info("Y", TensorProto.FLOAT, (5,))])
+        optimized_model = self._optimized(graph, ["eliminate_identity"])
+
```
houseroad: check len(node) == 0 and also check the output is "Y"

onnx/test/optimizer_test.py
```diff
@@ -19,6 +19,33 @@ def _optimized(self, graph, opts):
...
+        add = helper.make_node("Add", ["Z", "Y"], ["A"])
+        mul = helper.make_node("Mul", ["A", "Y"], ["B"])
+        graph = helper.make_graph(
+            [identity, add, mul],
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (5,)),
+             helper.make_tensor_value_info("Z", TensorProto.FLOAT, (5,))],
+            [helper.make_tensor_value_info("B", TensorProto.FLOAT, (5,))])
+        optimized_model = self._optimized(graph, ["eliminate_identity"])
+
```
houseroad: Please also check len(node) == 2

houseroad:
Identity is not useful in SSA format. That's we need to eliminate them from the graph. :-)
I guess it will make the conversion work much easier for model exporters, since all the framework support Identity-ish operator. If not, we should remove it. @tjingrant @ebarsoum @linkerzhang please confirm.

onnx/test/optimizer_test.py
fumihwh: @houseroad 
The output is "X" because Identity is eliminated also "Y".

fumihwh:
@houseroad 
BTW, in tf, when initialize variables, following code will be run, which create an Identity op in graph.
https://github.com/tensorflow/tensorflow/blob/r1.7/tensorflow/python/ops/variables.py#L386-L391

bddppq:
@wschin Let's leave it if it's useful.

onnx/optimizer/passes/eliminate_identity.h
```diff
@@ -0,0 +1,30 @@
...
+struct EliminateIdentity final : public OptimizePass {
+  explicit EliminateIdentity()
+    : OptimizePass("eliminate_identity", API_TYPE::IR) {
+  }
+
+  void eliminate_identity(Graph& graph) {
+    for (auto it = graph.begin(); it != graph.end(); ++it) {
+      auto* n = *it;
+      if (n->kind() == kIdentity) {
+        n->output()->replaceAllUsesWith(n->input());
```
wschin: If an Identity is connected with a graph output, we should keep the output name?

onnx/test/optimizer_test.py
wschin: We probably need to keep output name if the considered Identity is a leaf node.

fumihwh: @wschin You mean case that Identity is an output node right? Agree.

fumihwh:
@wschin 
If there is no effect to the computation graph, we could optimize graph by removing them.
I am concerned about 2), what do two identical outputs do? Why we need that?

onnx/test/optimizer_test.py
houseroad: @fumihwh yeah, I mean, check the output is "X" in the graph.

onnx/optimizer/passes/eliminate_identity.h
houseroad: For now, I think no need, because we didn't keep the output name in other passes either.

But it may be worth to create some helper functions to rename the inputs and outputs

wschin:
In my opinion, we can remove Identity operators only if it's not going to change the "interface" of a computational graph. Users don't care about what a black box is doing internally. However, if the interface (i.e., graph I/O) connecting user application and the real computation is changed, it's not a black box anymore (it probably becomes a red box) because the poor user needs to modify his application. For (2), it happens when output_sequence=False and output_state=True. Again, as RNNs naturally allow this case, ONNX should have it.

houseroad:
@wschin the property you mentioned is good to have. But we didn't require this at the beginning and we already break it. In some case, it's really hard to keep this property, for example, if the whole graph only contains one identity op or non-op transpose, I think we should definitely remove this op. In general, the optimizer should preprocess the model offline. (Yes, some frameworks will use the optimizer on the fly, but it's easy to update the names of outputs.)

Further more, it's more appropriate to add this feature in a different PR.

wschin:
@houseroad , if the only op, an Identity, is removed, what is the expected behavior of this graph? There is no op, so semantically it shouldn't do anything. Is it a defined behavior of a 0-op graph in ONNX? For RNNs, CoreML can output both of the two identical outputs as a part their graph outputs. If optimization happens, we have two duplicated output names. I don't know if ONNX allows this situation. Therefore, an alternative solution is (1) to explain what users should do for empty graphs (2) clarify that duplicated graph outputs are allowed (it's confusing because operator should also produce unique output names).

houseroad:
The inputs and outputs are defined here: https://github.com/onnx/onnx/blob/master/onnx/onnx.proto#L248--L249. The inputs and outputs of an ONNX model are well defined, even if we don't have any node in the graph. ONNX models can have two identical outputs without any problem, because any tensor/var is only defined in initializer or operator. We can still keep SSA property.

I am not trying to remove Identity operator, actually, I also would like to keep it in the spec, if it will greatly ease the exporting work. :-)

PR #757, opened 2018-04-13
Tile - align with numpy

houseroad:
~~Change the type and description of 2nd and 3rd inputs.~~

With more investigation, most existing frameworks aligned with numpy. So I aligned our spec with numpy, too. Numpy strongly suggests not using broadcast in tile. In our spec, broadcast is not supported.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -690,16 +690,32 @@ and width dimensions.
         "Constrain input types to float tensors.");
 
 ONNX_OPERATOR_SCHEMA(Tile)
-    .SetDoc(R"DOC(Repeat the elements of a tensor along an axis.)DOC")
-    .Input(0, "input", "Input tensor of any shape.", "T")
+    .SetDoc(R"DOC(Constructs a tensor by tiling a given tensor.
+This is the same as function `tile` in Numpy, but no broadcast.
+For example A = [[1, 2], [3, 4]], B = [1, 2], tile(A, B) = [[1, 2, 1, 2], [3, 4, 3, 4]]
+)DOC")
+    .Input(
+        0,
+        "input",
+        "Input tensor of any shape.",
+        "T")
     .Input(
         1,
-        "tiles",
-        "Number of repeated copies to make of the input tensor.",
+        "repeats",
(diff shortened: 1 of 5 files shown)


onnx/defs/tensor/defs.cc
```diff
@@ -472,15 +472,27 @@ and width dimensions.
...
-    .Input(2, "axis", "Axis along which to repeat.", "T")
-    .Output(0, "output", "Output tensor of same shape and type as input.", "T")
+        "1D int64 tensor, include numbers of repeated copies to make of the input tensor.",
+        "T1")
+    .Input(
+        2,
+        "axes",
+        "1D int64 tensor, has the same shape as input tiles, indicates along which axes to repeat.",
+        "T1")
+    .Output(0, "output", "Output tensor of same shape and type as input.", "T1")
```
bddppq: this is of type T

houseroad: good catch

onnx/defs/tensor/defs.cc
```diff
@@ -472,15 +472,31 @@ and width dimensions.
...
+        "T1")
+    .Input(
+        2,
+        "axes",
+        "1D int64 tensor, has the same shape as input tiles, indicates along which axes to repeat.",
+        "T1")
+    .Output(
+        0,
+        "output",
+        "Output tensor of same shape and type as input.",
```
bddppq: output should not have same shape as input

linkerzhang:
@houseroad , I noticed that the "tiles" and "axis" were designed as input before, and we're using "repeats" for doing tile in multiple dimension. However, question please. do we really need this kind of dynamic "repeats", which introduces performance hits in run-time, I believe.

houseroad:
@linkerzhang Yeah, making `repeats` as inputs can bring us the following benefits:
- Most existing frameworks (such as TensorFlow, Caffe2, PyTorch) take repeats as inputs. We should align with them, so the exporting will be much easier.
- Even in the static cases, we can have an optimization pass to preprocess the onnx graph while importing, so performance won't be a concern.
- We indeed have some dynamic `repeats` cases internally, which means args are not enough.

linkerzhang:
@houseroad  Thank you for the detail clarification. Looks good to me.

PR #758, opened 2018-04-13
more shape inference implementations

anderspapitto:


onnx/test/shape_inference_test.py
```diff
@@ -126,6 +126,20 @@ def test_gather_into_scalar(self):
...
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, ())])
 
+    def test_squeeze(self):
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, (1, 3, 1, 1, 2, 1))],
+            [make_node('Squeeze', 'x', 'y', axes=[0, 2, 3, 5])],
+            [])
+        self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (3, 2))])
+
+    def test_squeeze(self):
```
bddppq: unsqueeze

onnx/defs/tensor/defs.cc
```diff
@@ -492,7 +492,33 @@ Takes a  parameter `axes` with a list of axes to squeeze.
...
+        }
+
+        if (!ctx.getInputType(0)->has_shape()) {
+          return;
+        }
+
+        ctx.getOutputType(0)->mutable_shape();
+
+        for (int i = 0, j = 0; i < ctx.getInputType(0)->shape().dim_size(); ++i) {
+          if (j < axes.size() && axes[j] == i) {
```
bddppq: This requires `axes` to be sorted. (which is unspecified in the spec)

onnx/defs/tensor/defs.cc
```diff
@@ -492,7 +492,33 @@ Takes a  parameter `axes` with a list of axes to squeeze.
...
+        "Constrain input and output types to float tensors.")
+    .ShapeInferenceFunction([](InferenceContext& ctx) {
+        if (!hasExactlyNInputTypes(ctx, 1, "Squeeze")) {
+          return;
+        }
+
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+
+        std::vector<int64_t> axes;
+        if (!getRepeatedAttribute(ctx, "axes", axes)) {
```
bddppq: `axes` is required attribute

onnx/test/shape_inference_test.py
anderspapitto: ++

onnx/defs/tensor/defs.cc

onnx/defs/tensor/defs.cc
anderspapitto: so - would you like me to throw an exception, or just assume it succeeds and remove the check? or leave it as is?

houseroad: we should have assertion here, but I know assertion will cause Appveyor hanging...

houseroad: throw an exception should be fine, because model is incorrect.

houseroad: yes, please throw an exception, or use this code: https://github.com/onnx/onnx/blob/master/onnx/common/assertions.h

ONNX_ASSERT is actually an exception, not a real assert

gramalingam: IMHO, just returning is fine too. I think checking the model for correctness is not the job of shape-inference. It is intended for optimization, and it is fine if doesn't infer the output shape under some conditions.

bddppq: exactly because "checking the model for correctness is not the job of shape-inference", here should throw instead of silently return.

PR #763, opened 2018-04-16
Add unit test framework for the project C++ APIs

raymondxyang:
* Add unit test framework for the project
* Add initial test case for op registration schema
* Update CI pipeline for enabling unit tests

Changes from last PR:
* Avoid redundant build of ONNX
* Disable c++ test by default

I am gonna separate the projects (in CMake) later. We may need the tests running first for other developers to contribute test cases first


.travis/install.sh
```diff
@@ -3,4 +3,8 @@
 script_path=$(python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" "${BASH_SOURCE[0]}")
 source "${script_path%/*}/setup.sh"
 
+time ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI pip install -v -e .
+find . -name "onnx_gtests" -ls -exec chmod +x {} \; -exec {} \;
+pip uninstall -y onnx || python setup.py develop --uninstall
+rm -rf .setuptools-cmake-build || :
```
bddppq: what's the reason to suppress the exit code?

CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ set(ONNX_ROOT ${PROJECT_SOURCE_DIR})
 set(CMAKE_MODULE_PATH "")
 list(APPEND CMAKE_MODULE_PATH ${ONNX_ROOT}/cmake/Modules)
 
+if(ENABLE_GTESTS)
```
bddppq: Add "ONNX" to the flag. Otherwise if onnx's cmake files get included by other frameworks's cmake build system, the parent scope variables will leak to onnx's.

external/googletest.cmake
```diff
@@ -0,0 +1,30 @@
+include (ExternalProject)
```
bddppq: IMO it's better to simply add googletest as a submodule in onnx

setup.py
```diff
@@ -149,6 +149,7 @@ def run(self):
                 '-DCMAKE_EXPORT_COMPILE_COMMANDS=ON',
                 '-DONNX_NAMESPACE={}'.format(ONNX_NAMESPACE),
                 '-DPY_EXT_SUFFIX={}'.format(sysconfig.get_config_var('EXT_SUFFIX') or ''),
+                '-DENABLE_GTESTS=ON'
```
bddppq: python build should not build anything that python side doesn't need (in this case cpp tests are not usable from python side).

.travis/install.sh
```diff
@@ -3,4 +3,8 @@
 script_path=$(python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" "${BASH_SOURCE[0]}")
 source "${script_path%/*}/setup.sh"
 
+time ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI pip install -v -e .
```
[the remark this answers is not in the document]
raymondxyang: So we need to do a insource develop install for building and run the unittest, then a regular install for other tests defined in script.sh

.travis/install.sh
raymondxyang: This one is unnecessary. Will remove it

setup.py
raymondxyang: Good point.. I ll change this to enabling UT in CI instead

.travis/install.sh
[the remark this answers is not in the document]
bddppq: pip doesn't seem to be the right way to trigger the build of c++ test binaries. How about we do it completely separated? I mean we first pip install python part, run python tests as before and then add new steps to directly invoke cmake to build the test binaries and run the tests.

raymondxyang:
I am gonna separate the projects (in CMake) later. We may need the tests running first for other developers to contribute test cases first

cmake/unittest.cmake
```diff
@@ -0,0 +1,106 @@
...
+    # Assume everything else is like gcc
+    set(${output_var} "-Wl,--whole-archive $<TARGET_FILE:${lib}> -Wl,--no-whole-archive" PARENT_SCOPE)
+  endif()
+endfunction()
+
+set(${UT_NAME}_libs
+    ${googletest_STATIC_LIBRARIES}
+)
+
+if(NOT TARGET protobuf::libprotobuf)
```
bddppq: why do we need to find protobuf here again? I think we have already done this in the top level cmake file?

CMakeLists.txt
```diff
@@ -186,12 +191,17 @@ RELATIVE_PROTOBUF_GENERATE_CPP(PROTO_SRCS PROTO_HDRS ${ONNX_ROOT}
     onnx/onnx-operators.in.proto
 )
 
+file(GLOB_RECURSE onnx_INCLUDE_DIRS "${ONNX_ROOT}/onnx/*.h")
```
bddppq: wait I think this is finding a list of header files, not "dirs". We have set ONNX_INCLUDE_DIRS, will it be enough by just adding it to the include dirs when compiling the test cc files?

CMakeLists.txt
```diff
@@ -333,3 +343,14 @@ install(DIRECTORY ${ONNX_ROOT}/onnx DESTINATION include
 install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/onnx DESTINATION include
     FILES_MATCHING
     PATTERN "*.h")
+if(ONNX_ENABLE_GTESTS)
+    include_directories(
```
bddppq: I don't think this is needed (and also it's too broad, e.g. it's adding gtest include dir even when compiling non-test cc files). In unittest.cmake it already adds the include dirs by `target_include_directories(${_UT_TARGET} PUBLIC ${googletest_INCLUDE_DIRS} ${onnx_INCLUDE_DIRS} ${PROTOBUF_INCLUDE_DIRS})`

onnx/test/gtests/op_reg_test.cc
```diff
@@ -0,0 +1,24 @@
...
+		TEST(OpRegistrationTest, AffineOp)
+		{
+			auto opSchema = OpSchemaRegistry::Schema("Affine");
+			EXPECT_TRUE(nullptr != opSchema);
+			size_t input_size = opSchema->inputs().size();
+			EXPECT_EQ(input_size, 1);
+			EXPECT_EQ(opSchema->inputs()[0].GetTypes(), opSchema->outputs()[0].GetTypes());
+			size_t attr_size = opSchema->attributes().size();
+			EXPECT_EQ(attr_size, 2);
+			EXPECT_NE(opSchema->attributes().find("alpha"), opSchema->attributes().end());
```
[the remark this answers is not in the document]
raymondxyang: if use count, gtest does support EXPECT_NE(NULL, ptr) see https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-does-google-test-support-expect_eqnull-ptr-and-assert_eqnull-ptr-but-not-expect_nenull-ptr-and-assert_nenull-ptr

cmake/unittest.cmake
bddppq: I checked the last failure, it should be irrelevant. I think the `if` should be removed.

cmake/unittest.cmake
```diff
@@ -0,0 +1,87 @@
...
+  list(APPEND ${UT_NAME}_libs ${PROTOBUF_LIBRARY})
```
bddppq: `PROTOBUF_LIBRARY` and `PROTOBUF_LIBRARIES` are the same. It's just normally in cmake libs maintainer export all these similar names to the same values to protect typos when other people use it.

cmake/external/googletest.cmake
```diff
@@ -0,0 +1,30 @@
+include (ExternalProject)
+
+set(googletest_INCLUDE_DIRS ${CMAKE_CURRENT_BINARY_DIR}/googletest/src/googletest/googletest/include)
```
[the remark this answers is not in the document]
bddppq: @raymondxyang Alright, pinky swear :-)

.travis/install.sh
```diff
@@ -3,4 +3,7 @@
 script_path=$(python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" "${BASH_SOURCE[0]}")
 source "${script_path%/*}/setup.sh"
 
-time ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI pip install -v .
+export ONNX_BUILD_TESTS=1
+pip install protobuf numpy
+time ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI python  setup.py bdist_wheel --universal --dist-dir .
+find . -maxdepth 1 -name "*.whl" -ls -exec pip install {} \;
```
bddppq: I guess you are doing this so the test binaries are generated in the source dir?
You can add `-b` to the `pip install` command in `script.sh` to point it to use a specified build path, so that later you can find and execute the test binaries.

.travis/script.sh
```diff
@@ -3,7 +3,10 @@
 script_path=$(python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" "${BASH_SOURCE[0]}")
 source "${script_path%/*}/setup.sh"
 
-# onnx tests
+# onnx c++ API tests
+find .setuptools-cmake-build/ -name "onnx_gtests" -ls -exec chmod +x {} \; -exec {} \;
```
bddppq: I don't think this is needed. These binaries should be executable already.

CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ option(ONNX_BUILD_BENCHMARKS "Build ONNX micro-benchmarks" OFF)
 
 option(BUILD_ONNX_PYTHON "Build Python binaries" OFF)
 
+option(ONNX_BUILD_TESTS "Build ONNX micro-benchmarks" OFF)
```
bddppq: The description needs to be adjusted.

CMakeLists.txt
```diff
@@ -333,3 +345,10 @@ install(DIRECTORY ${ONNX_ROOT}/onnx DESTINATION include
 install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/onnx DESTINATION include
     FILES_MATCHING
     PATTERN "*.h")
+if(ONNX_BUILD_TESTS)
+    include_directories(
```
bddppq: Do we still need to add these include dirs at the top level? You are already doing `target_include_directories` in unittest.cmake.

cmake/unittest.cmake
```diff
@@ -0,0 +1,87 @@
...
+function(AddTest)
+  cmake_parse_arguments(_UT "" "TARGET" "LIBS;SOURCES" ${ARGN})
+
+  list(REMOVE_DUPLICATES _UT_LIBS)
+  list(REMOVE_DUPLICATES _UT_SOURCES)
+
+  add_executable(${_UT_TARGET} ${_UT_SOURCES})
+  add_dependencies(${_UT_TARGET} onnx onnx_proto googletest)
+
+  target_include_directories(${_UT_TARGET} PUBLIC ${googletest_INCLUDE_DIRS} ${onnx_INCLUDE_DIRS} ${PROTOBUF_INCLUDE_DIRS})
```
bddppq: it should be ONNX_INCLUDE_DIRS.

.travis/install.sh
raymondxyang: I ll re-test it later.. I tried set TMPDIR and --noclean somehow its not working.. I didnot remmeber if I used -b before

.travis/script.sh
raymondxyang: for chmod?

CMakeLists.txt
raymondxyang: yes.. that's the minimal set working on windows

.travis/script.sh
raymondxyang: haha not so familiar with the linux rule so always 777

cmake/external/googletest.cmake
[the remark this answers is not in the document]
raymondxyang: It just came to me that if we treat it as external project, we only download and build it on the fly when user activate the option. If its a sub-module we'll always have it around..

CMakeLists.txt
raymondxyang: sorry i misunderstand ur point.. moved the include dirs from top level to target_include_directories

.travis/install.sh
raymondxyang: Just tested the `-b` is not working.. only protobuf is retained in the given directory.. dont know the reason but log can be found https://api.travis-ci.org/v3/job/374997390/log.txt, command used is `pip install -b ${PWD}/.setuptools-cmake-build/ -v .`

cmake/external/googletest.cmake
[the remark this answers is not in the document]
snnn: Another solution maybe: checkin googletests directly at here. The whole project can be packed into only 2 files: one header file and one cc file. It's simpler to manage.

cmake/unittest.cmake
```diff
@@ -0,0 +1,83 @@
...
+  endif()
+endfunction()
+
+set(${UT_NAME}_libs
+    ${googletest_STATIC_LIBRARIES}
+)
+
+list(APPEND ${UT_NAME}_libs ${PROTOBUF_LIBRARIES})
+
+add_whole_archive_flag(onnx_proto tmp)
```
snnn: not needed?

cmake/unittest.cmake
```diff
@@ -0,0 +1,83 @@
...
+  target_link_libraries(${_UT_TARGET} ${_UT_LIBS} ${CMAKE_THREAD_LIBS_INIT})
+  if(TARGET protobuf::libprotobuf)
+    target_link_libraries(${_UT_TARGET} protobuf::libprotobuf)
+  endif()
+
+  if (WIN32)
+    target_compile_options(${_UT_TARGET} PRIVATE
+        /EHsc   # exception handling - C++ may throw, extern "C" will not
+    )
+    target_compile_options(${_UT_TARGET} PRIVATE /MT)
```
snnn: Why /MT?

cmake/unittest.cmake
```diff
@@ -0,0 +1,83 @@
...
+  list(REMOVE_DUPLICATES _UT_SOURCES)
+
+  add_executable(${_UT_TARGET} ${_UT_SOURCES})
+  add_dependencies(${_UT_TARGET} onnx onnx_proto googletest)
+
+  target_include_directories(${_UT_TARGET} PUBLIC ${googletest_INCLUDE_DIRS} ${ONNX_INCLUDE_DIRS} ${PROTOBUF_INCLUDE_DIRS} ${ONNX_ROOT} ${CMAKE_CURRENT_BINARY_DIR})
+  target_link_libraries(${_UT_TARGET} ${_UT_LIBS} ${CMAKE_THREAD_LIBS_INIT})
+  if(TARGET protobuf::libprotobuf)
+    target_link_libraries(${_UT_TARGET} protobuf::libprotobuf)
+  endif()
```
snnn: else?

linkerzhang:
@bddppq this looks good to you now?

bddppq:
@linkerzhang I will take a look tomorrow

linkerzhang:
@bddppq  I'm merging this in. Please add more comments if any and will fix them separately. Thank you very much!

Issue #754, opened 2018-04-13
Python schema API not take domain attributes


raymondxyang:
Find the python `has_schema` and `get_shcema` api was not implemented consistently with the backend implementation. Fixing this to support ONNX-ML.

+ @linkerzhang

PR #764, opened 2018-04-16
Update Python schema API to take domain

raymondxyang:
#754


onnx/cpp2py_export.cc
```diff
@@ -91,21 +91,32 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
       .value("COMMON", OpSchema::SupportType::COMMON)
       .value("EXPERIMENTAL", OpSchema::SupportType::EXPERIMENTAL);
 
-  defs.def("has_schema", [](const std::string& op_type) -> bool {
-    return OpSchemaRegistry::Schema(op_type) != nullptr;
+  defs.def("has_schema", [](const std::string& op_type,
```
bddppq: Need to change the python interface to expose this new function signature https://github.com/onnx/onnx/blob/master/onnx/defs/__init__.py#L13.
Actually I think it's better to just do

```
has_schema = C.has_schema
has = has_schema  # for backward compatibility
```

onnx/cpp2py_export.cc
```diff
@@ -91,21 +91,32 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+			    const std::string& domain) -> bool {
+    return OpSchemaRegistry::Schema(op_type, domain) != nullptr;
+  }).def("has_schema", [](const std::string& op_type) -> bool {
```
bddppq: pybind11 supports specifying default value for function argument, so no need to define twice for different function signatures.

raymondxyang: I tried default value but pytest gave me errors 
```   
def test_typecheck(self):
>       defs.get_schema("Conv")
E       TypeError: get_schema(): incompatible function arguments. The following argument types are supported:
E           1. (arg0: str, arg1: int, arg2: str) -> onnx.onnx_cpp2py_export.defs.OpSchema
E
E       Invoked with: 'Conv'

onnx\test\schema_test.py:12: TypeError
```

onnx/cpp2py_export.cc
raymondxyang: updated

onnx/cpp2py_export.cc
bddppq: Have you annotated the default arguments like this: http://pybind11.readthedocs.io/en/stable/basics.html#default-arguments ?

onnx/defs/__init__.py
```diff
@@ -6,38 +6,21 @@
...
-def has(op_type):
-    return C.has_schema(op_type)
-
-
-def onnx_opset_version():
-    return C.schema_version_map()[ONNX_DOMAIN][1]
-
-
-def get_schema(op_type, max_inclusive_version=onnx_opset_version()):
-    return C.get_schema(op_type, max_inclusive_version)
```
fumihwh: @raymondxyang 
You've removed default value of max_inclusive_version and get that error.

bddppq: let's add the default argument in the pybind11 bindings

onnx/cpp2py_export.cc
raymondxyang: yes... I also tried none(true), not working as well... also I feel a little bit uncertain want should the default value be, because we already have a default one defined in schema.h

onnx/defs/__init__.py
raymondxyang: Gonna revert this first until we figure out on which step(s) we need to define a default value.. currently we have: 1) py apis(some with default value) 2) pybind(no default value) 3) c apis(default value defined)

fumihwh: Both 2 and 3 don't have default value I think.
And we should not add default value to schema.h because of there is another constructor which is using maxInclusiveVersion.
What we need to do I think is just add default value to pybind11 bindings as @bddppq said.
Maybe we can also add default value of domain.

onnx/cpp2py_export.cc
bddppq: the default value should match the one in c++
btw the error you got seems to be from get_schema, not has_schema

raymondxyang: they have the same error so I copied one of it.

bddppq: ok as this is just implementation details that doesn't affect user interface, let's get this in first and we can revisit it later.

PR #765, opened 2018-04-16
The content of a string is doubled if it's a string tensor

wschin:
Fixed the problem described in the title.

### DIFF
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -117,14 +117,12 @@ def make_tensor(name, data_type, dims, vals, raw=False):
     tensor.data_type = data_type
     tensor.name = name
 
-    if data_type == TensorProto.STRING:
-        assert not raw, "Can not use raw_data to store string type"
-        tensor.string_data.extend(vals)
-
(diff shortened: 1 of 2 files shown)


onnx/helper.py
```diff
@@ -117,14 +117,12 @@ def make_tensor(name, data_type, dims, vals, raw=False):
...
-    if data_type == TensorProto.STRING:
-        assert not raw, "Can not use raw_data to store string type"
-        tensor.string_data.extend(vals)
-
     if (data_type == TensorProto.COMPLEX64 or
             data_type == TensorProto.COMPLEX128):
         vals = split_complex_to_pairs(vals)
     if raw:
+        if data_type == TensorProto.STRING:
+            assert not raw, "Can not use raw_data to store string type"
```
houseroad: This does not look right to me
In `raw` branch, assert `not raw`?

bddppq:
should simply remove line 122.
Could yu also add a test case to verify this?

wschin:
@bddppq , Sure. I will fix them. Thanks.

PR #770, opened 2018-04-17
Add functions to load and save TensorProto easily

houseroad:
Load and save TensorProto is often used. So add this function.


onnx/__init__.py
```diff
@@ -14,6 +14,25 @@
...
+        with open(obj, 'rb') as f:
+            s = f.read()
+    return s
+
+
+def _save(obj, writable):
+    if hasattr(obj, 'SerializeToString') and callable(obj.SerializeToString):
+        writable.write(obj.SerializeToString())
+    elif isinstance(obj, str):
+        writable.write(model)
```
anderspapitto: `model` is an unbound variable?

houseroad: ++

onnx/__init__.py
```diff
@@ -14,6 +14,60 @@
...
+        s = obj.read()
+    else:
+        with open(obj, 'rb') as f:
+            s = f.read()
+    return s
+
+
+def _save_string(str, f):
+    if hasattr(f, 'write') and callable(f.write):
+        f.write(obj, f)
```
bddppq: str

onnx/__init__.py
```diff
@@ -14,6 +14,60 @@
...
+    else:
+        with open(f, 'wb') as writable:
+            writable.write(obj, writable)
+
+
+def serialize(proto):
```
bddppq: Doesn't seem more useful than `proto.SerializeToString`?

bddppq:
I feel it's probably better to just add a new "type" argument to load and save (default to `ModelProto` for backward compatibility).

onnx/__init__.py
houseroad: err... i ran the wrong scripts to test this ;-)

houseroad:
Yep, adding another parameter sounds better...

load and load_tensor are really confusing, let me refactor.

onnx/__init__.py
```diff
@@ -14,59 +14,110 @@
...
+    If cls is not specified, loads obj as ModelProto
+
+    @return
+    Loaded in-memory proto object
+    '''
+    s = _load_bytes(obj)
+    return load_from_string(s, cls)
+
 
+def load_from_string(s, cls=None):
```
bddppq: why not default to ModelProto?

onnx/__init__.py
```diff
@@ -14,59 +14,110 @@
...
 
 
-def _save(model, writable):
-    if isinstance(model, ModelProto):
-        writable.write(model.SerializeToString())
-    elif isinstance(model, str):
-        writable.write(model)
-    else:
-        raise ValueError('Model is neither ModelProto nor str.\n{}'.format(model))
+def load(obj, cls=None):
```
bddppq: nit: use "f" instead of `obj` to match `save`?

bddppq:
I think it's better to not expose serialize and deserialize.
For serialize because proto objects already have a "serializeToString" method, and also for the new model format (that offloads tensors to separate files), we actually don't know what should be return.
For deserialize, I think same as load.

onnx/__init__.py
houseroad: I feel bad to use any mutable object as default values

onnx/__init__.py
```diff
@@ -14,59 +14,110 @@
...
+    The proto instance filled in by s
     '''
-    model = ModelProto()
-    decoded = model.ParseFromString(s)
-    # in python implementation ParseFromString returns None
+    if not isinstance(s, bytes):
+        raise ValueError('Parameter s must be bytes, but got type: {}'.format(type(s)))
+
+    if callable(obj):
+        obj = obj()
```
bddppq: it's better not to smash these two cases (i.e. passing the proto class vs. passing the proto instance) into one function, they are pretty different (e.g one will modify the proto instance in-place, while in the other case it returns a new instance).

houseroad: Yep, I was thinking whether we should support in-place deserialization. But it won't bring any benefits in this case. I will stick to class instead of instance.

PR #771, opened 2018-04-17
Clean shape inference

anderspapitto:


anderspapitto:
[ do not review till https://github.com/onnx/onnx/pull/758 is pushed ]

onnx/defs/shape_inference.h
```diff
@@ -44,8 +44,22 @@ inline bool hasExactlyNInputTypes(InferenceContext& ctx, int n, const std::strin
   return true;
 }
 
+inline bool hasNInputShapes(InferenceContext& ctx, int n, const std::string& opname) {
```
bddppq: probably not worth adding a parameter "opname" just for error message (if you really want to show the opname, you can store the opname in the InferenceContext itself).

onnx/defs/shape_inference.h
```diff
@@ -44,8 +44,22 @@ inline bool hasExactlyNInputTypes(InferenceContext& ctx, int n, const std::strin
...
+    if (!ctx.getInputType(i) || !ctx.getInputType(i)->has_shape()) {
+      return false;
+    }
+  }
+  return true;
+}
+
 inline void propagateElemTypeFromInputToOutput(InferenceContext& ctx, size_t inputIndex, size_t outputIndex) {
-  ctx.getOutputType(outputIndex)->set_elem_type(ctx.getInputType(inputIndex)->elem_type());
+  if (ctx.getInputType(inputIndex)) {
```
bddppq: maybe assert inputIndex and outputIndex are not out of bound?

bddppq: actually it might be better to assert inside getInputType and getOutputType

anderspapitto: done, see implementation.h

onnx/shape_inference/implementation.h
```diff
@@ -36,12 +36,18 @@ struct InferenceContextImpl : public InferenceContext {
     return allInputTypes_.size();
   }
   const TypeProto_Tensor* getInputType(size_t index) const override {
+    if (index >= allInputTypes_.size()) {
+      throw std::runtime_error("input " + std::to_string(index) + " is out of bounds");
```
bddppq: use ONNX_NAMESPACE::to_string

PR #772, opened 2018-04-17
Inference impls

anderspapitto:


anderspapitto:
[ do not review till https://github.com/onnx/onnx/pull/758 is pushed ]

anderspapitto:
onnx-fb-universe failure is a spurious network error

onnx/defs/tensor/defs.cc
```diff
@@ -632,7 +673,29 @@ Given `data` tensor, pads, mode, and value.
...
+        }
+
+        ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+
+        for (size_t i = 0; i < ctx.getInputType(0)->tensor_type().shape().dim_size(); ++i) {
+          auto* newdim = ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape()->add_dim();
+          if (ctx.getInputType(0)->tensor_type().shape().dim(i).has_dim_value()) {
+            newdim->set_dim_value(ctx.getInputType(0)->tensor_type().shape().dim(i).dim_value() + pads[2 * i] + pads[2 * i + 1]);
+          } else {
+            *newdim = ctx.getInputType(0)->tensor_type().shape().dim(i);
```
gramalingam: This doesn't seem correct unless the pads value is zero. In particular, if input dim is "N" and pads is non-zero, we shouldn't set output dim to be "N"

gramalingam:
I have a high-level question/comment: how do we represent a completely unknown shape (of unknown rank)? If we restrict ourselves to the current protobuf definition, then one solution is to have the shape field of TypeProto not be set (so that has_shape() is false). Is there any other alternative?

If we go with the above convention, then this PR needs to be updated accordingly ... (a) don't assume that an input has_shape() ... check if it has_shape() and treat it as an unknown rank otherwise, (b) don't set the output's shape unless it's rank is known.

onnx/defs/tensor/defs.cc
anderspapitto: good catch

anderspapitto:
@gramalingam yes, you're correct - it's fully determined by has_shape(). We in fact do check has_shape() and treat it as unknown rank if there is no shape - see the implementation and usage of `hasNInputShapes()`

anderspapitto:
It was hasNInputTypes() a few commits ago, I updated it at some point to make this handling better :)

onnx/defs/nn/defs.cc
```diff
@@ -246,8 +246,8 @@ computes the output.)DOC";
         "is the number of channels, and kH and kW are the "
         "height and width of the kernel, and M is the number "
         "of feature maps. For more than 2 dimensions, the "
-        "kernel shape will be (M x C x k1 x k2 x ... x kn), "
-        "where is the dimension of the kernel",
+        "weight shape will be (M x C x k1 x k2 x ... x kn), "
```
houseroad: Thanks for fixing this.

onnx/defs/nn/defs.cc
```diff
@@ -292,6 +292,75 @@ computes the output.)DOC";
...
         AttributeProto::INT,
         static_cast<int64_t>(1));
+    schema.TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+
+        if (!hasNInputShapes(ctx, 1)) {
+          return;
+        }
+
+        // don't bother with legacy auto_pad for now
```
houseroad: TODO?

onnx/defs/nn/defs.cc
```diff
@@ -292,6 +292,75 @@ computes the output.)DOC";
...
+            dilations.push_back(1);
+          }
+        }
+
+        std::vector<int64_t> kernel_shape;
+        if (!getRepeatedAttribute(ctx, "kernel_shape", kernel_shape)) {
+          for (int i = 2; i < ctx.getInputType(1)->tensor_type().shape().dim_size(); ++i) {
+            kernel_shape.push_back(ctx.getInputType(1)->tensor_type().shape().dim(i).dim_value());
+          }
+        }
```
houseroad: else add some assertion here?

onnx/defs/nn/defs.cc
```diff
@@ -292,6 +292,75 @@ computes the output.)DOC";
...
+
+        std::vector<int64_t> strides;
+        if (!getRepeatedAttribute(ctx, "strides", strides)) {
+          for (int i = 0; i < ctx.getInputType(0)->tensor_type().shape().dim_size(); ++i) {
+            strides.push_back(1);
+          }
+        }
+
+        *ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape()->add_dim() =
```
houseroad: Dumb question: if ctx.getInputType(0) does not have shape info, what will happen?

onnx/defs/nn/defs.cc
```diff
@@ -292,6 +292,75 @@ computes the output.)DOC";
...
+          }
+        }
+
+        std::vector<int64_t> strides;
+        if (!getRepeatedAttribute(ctx, "strides", strides)) {
+          for (int i = 0; i < ctx.getInputType(0)->tensor_type().shape().dim_size(); ++i) {
+            strides.push_back(1);
+          }
+        }
+
```
houseroad: Could you add the formula here: how do you calculate the output shape...

onnx/defs/nn/defs.cc
```diff
@@ -292,6 +292,75 @@ computes the output.)DOC";
...
+          // accounting for dilation, how big is the kernel in this dimension
+          int64_t effective_kernel_size = kernel_shape[i];
+          effective_kernel_size = (effective_kernel_size - 1) * dilations[i] + 1;
+
+          // output size, no accounting for strides
+          int64_t kernel_positions = effective_input_size - effective_kernel_size + 1;
+
+          // output size, accounting for strides
+          int64_t effective_kernel_positions = kernel_positions / strides[i];
```
houseroad: The formula here seems incorrect:

For example: pad=0, dilation=1, input=5, kernel=5, stride=2, you will get output=0

I think right formula should be:
Output[i+2] = (Input[i+2] + padding[i] + padding[i+len(kernel_shape)] - dilations[i] *( kernel_shape[i] - 1) - 1) / stride[i] + 1

onnx/defs/nn/defs.cc
houseroad: To answer my own question: I think we depends on hasNInputShapes.

onnx/defs/nn/defs.cc
```diff
@@ -292,6 +292,75 @@ computes the output.)DOC";
         "number of groups input channels and output channels are divided into, default is 1.",
         AttributeProto::INT,
         static_cast<int64_t>(1));
+    schema.TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+
+        if (!hasNInputShapes(ctx, 1)) {
```
houseroad: ctx, 2, we also depends on weights' shape

onnx/defs/nn/defs.cc
anderspapitto: it's in the inline comments below

onnx/defs/nn/defs.cc
anderspapitto: yes

onnx/defs/nn/defs.cc
houseroad: Assert kernel_shape == dim_weight

onnx/defs/nn/defs.cc
```diff
@@ -292,6 +292,75 @@ computes the output.)DOC";
...
+          return;
+        }
+
+        // don't bother with legacy auto_pad for now
+        if (ctx.getAttribute("auto_pad")) {
+          return;
+        }
+
+        std::vector<int64_t> dilations;
+        if (!getRepeatedAttribute(ctx, "dilations", dilations)) {
```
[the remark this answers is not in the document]
anderspapitto: ++ I'll add a check here

onnx/defs/nn/defs.cc
```diff
@@ -292,6 +292,75 @@ computes the output.)DOC";
...
+          ctx.getInputType(1)->tensor_type().shape().dim(0);
+
+        for (size_t i = 0; i < kernel_shape.size(); ++i) {
+          if (!ctx.getInputType(0)->tensor_type().shape().dim(2 + (int)i).has_dim_value()) {
+            continue;
+          }
+          // how big is the input, including padding
+          int64_t effective_input_size = ctx.getInputType(0)->tensor_type().shape().dim(2 + (int)i).dim_value();
+          effective_input_size += pads[i];
+          effective_input_size += pads[i + kernel_shape.size()];
```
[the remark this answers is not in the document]
anderspapitto: ++ will also add a length check for `pads`

onnx/defs/tensor/defs.cc
```diff
@@ -351,7 +351,48 @@ Example 2:
...
+        std::vector<int64_t> starts;
+        std::vector<int64_t> ends;
+        if (!getRepeatedAttribute(ctx, "starts", starts) ||
+            !getRepeatedAttribute(ctx, "ends", ends) ||
+            starts.size() != ends.size()) {
+          return;
+        }
+        std::vector<int64_t> axes;
+        if (!getRepeatedAttribute(ctx, "axes", axes)) {
+          for (int i = 0; (size_t) i < starts.size(); ++i) {
```
houseroad: Do you want to sort the axes?

onnx/defs/tensor/defs.cc
```diff
@@ -632,7 +673,29 @@ Given `data` tensor, pads, mode, and value.
...
-        "Constrain input and output types to float tensors.");
+        "Constrain input and output types to float tensors.")
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+        if (!hasNInputShapes(ctx, 1)) {
+          return;
+        }
+
+        std::vector<int64_t> pads;
+        if (!getRepeatedAttribute(ctx, "pads", pads)) {
```
houseroad: Similar here, we need to check `pads`'s sizes

houseroad:
@anderspapitto In general, it looks good to me, but some shape_inference test failed. Could you take a look?

PR #775, opened 2018-04-18
Align schema function signatures in python and c++

bddppq:
@fumihwh


bddppq:
@linkerzhang @prasanthpul Seems like @raymondxyang doesn't write permission to the repo?

PR #777, opened 2018-04-18
Run pytest against installed version of onnx, not against source directory

smessmer:


bddppq:
Doesn't seem to work, closing. Feel free to reopen if we still need this.

PR #779, opened 2018-04-18
change the inference context api to use TypeProto

linkerzhang:


onnx/defs/shape_inference.h
```diff
@@ -44,17 +45,43 @@ inline bool hasExactlyNInputTypes(InferenceContext& ctx, int n, const std::strin
   return true;
 }
 
-inline void propagateElemTypeFromInputToOutput(InferenceContext& ctx, size_t inputIndex, size_t outputIndex) {
-  ctx.getOutputType(outputIndex)->set_elem_type(ctx.getInputType(inputIndex)->elem_type());
+inline void propagateElemTypeFromInputToOutput(
```
gramalingam: Change name to "propagateTypeFromInputToOutput"?

bddppq:
Could you only clang-format the changed parts? :-) It's pretty hard to review.

onnx/shape_inference/implementation.h
```diff
@@ -140,18 +148,21 @@ void InferShapes(ModelProto& m) {
...
+                  "inferred dimension differs from existing dimension");
             }
             *existingDim = inferredDim;
           }
         }
       }
 
       // Make it available to futher inference.
-      valueTypesByName[output] = existingType;
+      valueTypesByName[output] = iter->second;
```
bddppq: what if `iter == valueTypesByName.end()`?

onnx/defs/shape_inference.h
linkerzhang: I found that what it was supposed to do is really coping element type instead of whole type proto.

bddppq:
@anderspapitto Could you take a look?

PR #781, opened 2018-04-19
Fix some checks not ideal to onnx-ml

wschin:
1. Need to consider cases where no default ONNX operator set exists
2. Cannot assume that value_info's are all tensors (CoreML's neural
network classifiers contain ZipMap, so they may produce Maps)
3. Make some error messages actionable


onnx/checker.cc
```diff
@@ -45,8 +45,6 @@ void check_value_info(const ValueInfoProto& value_info, const CheckerContext&) {
       enforce_has_field(type, elem_type);
       enforce_has_field(type, shape);
     } break;
-    default:
```
linkerzhang: we should add #ifdef ONNX_ML and add cases of map/sequence in-place.

onnx/checker.cc
```diff
@@ -336,8 +334,7 @@ void check_model(const ModelProto& model) {
   for (const auto& opset_import : model.opset_import()) {
     opset_imports[opset_import.domain()] = static_cast<int>(opset_import.version());
   }
-  auto dit = opset_imports.find(ONNX_DOMAIN);
-  if (dit == opset_imports.end()) {
+  if (opset_imports.size() < 1) {
```
bddppq: `.empty()`

onnx/checker.cc
```diff
@@ -336,8 +347,7 @@ void check_model(const ModelProto& model) {
   for (const auto& opset_import : model.opset_import()) {
     opset_imports[opset_import.domain()] = static_cast<int>(opset_import.version());
   }
-  auto dit = opset_imports.find(ONNX_DOMAIN);
-  if (dit == opset_imports.end()) {
+  if (opset_imports.empty()) {
```
houseroad: This doesn't seem right to me. Do you want to avoid the check when you have other domain?

wschin: The following checks are not necessary if at least one domain exists because (1) opset_import is good with IR>=3 (2) default opset_imports should be set only if opset_import is empty.

houseroad: I think the right check should be:
```
if (model.ir_version() >= 3) {
  // enforce opset_imports is not empty
} else {
  // enforce opset_imports is empty,
  // and set the default value for ONNX_DOMAIN
}
```

wschin: Looks better.

onnx/checker.cc
```diff
@@ -336,14 +347,13 @@ void check_model(const ModelProto& model) {
...
-    if (model.ir_version() >= 3) {
+  if (model.ir_version() >= 3) {
+    if (opset_imports.empty())
       fail_check(
           "model with IR version >= 3 must specify opset_import for ONNX");
-    } else {
+  } else {
+    if (opset_imports.empty())
       opset_imports[ONNX_DOMAIN] = 1;
-    }
```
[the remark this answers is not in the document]
wschin: Done. Thanks.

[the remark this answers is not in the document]
wschin: It breaks a lot of torch tests.

houseroad:
@linkerzhang Some test cases were not updated. Checking the reason

PR #782, opened 2018-04-19
Try using pep518 to install the protobuf build dependency

smessmer:
This PR is to try out if this would work with pip on CI.

An alternative is: https://github.com/onnx/onnx/pull/783


smessmer:
CI says this works. No protoc-gen-mypy error message anymore.

I think we should land this. However, it means that we should encourage people to use pip. People not using pip will have to pre-install protobuf.

PR #783, opened 2018-04-19
Try using setup_requires to install protobuf build time dependency

smessmer:
This PR is to try out if this would work with pip on CI.

An alternative is: https://github.com/onnx/onnx/pull/782


smessmer:
doesn't work, CI still shows the protoc-gen-mypy error message

PR #798, opened 2018-04-20
Remove debug string

hlu1:


Yangqing:
Only 1 place used this string :)

https://github.com/onnx/onnx/search?utf8=%E2%9C%93&q=zhangke&type=

PR #799, opened 2018-04-21
Kezhan/update ml op spec

linkerzhang:


buehrer:
Looks good to me. Thanks!

PR #802, opened 2018-04-21
function proto for composite op.

linkerzhang:
This is the first step of registering common functions into onnx by using primitive onnx ops.


NiklasGustafsson:
This seems to be a big-enough change to warrant bumping the IR version, right?

onnx/onnx-operators.in.proto
```diff
@@ -26,6 +26,55 @@ import "onnx.proto";
...
+  // Functions "become" stable by deprecating the experimental version and
+  // introducing a new stable function with the same name.
+  optional OperatorStatus status = 3;
+
+  // The inputs and outputs of the function.
+  repeated string input = 4;
+  repeated string output = 5;
+
+  // The attributes declaration of the function.
+  repeated AttrDeclProto attr_decl = 6;
```
NiklasGustafsson: If we can derive the types of inputs and outputs through inference, why can't we do the same for attributes?

linkerzhang: Good question. We can. so that the attribute here will also be just "names", which will be referred by node inside the body.

linkerzhang:
@NiklasGustafsson , we should bump the IR version when release. Same idea as we're now bumping op_set version together with onnx release.

linkerzhang:
@ezyang please kindly review this change and add your comments if any. Thank you very much!

onnx/onnx.in.proto
```diff
@@ -99,6 +99,11 @@ message AttributeProto {
 
   // The name field MUST be present for this version of the IR.
   optional string name = 1;           // namespace Attribute
+ 
+  // if ref_attr_name is not empty, ref_attr_name is the attribute name in parent function.
+  // In this case, this AttributeProto does not contain data, and it's a reference of attribute
+  // in parent scope.
+  optional string ref_attr_name = 21;
```
ezyang: Need to mention that this is not currently valid in ModelProto

linkerzhang: Sure.

linkerzhang:
@ezyang Thank you very much for reviewing this! For checker logic, yes, they're not substantive, for the function cases. will add that with function registration code in-place.

PR #804, opened 2018-04-22
Pass to lift captured values as inputs to control nodes

jamesr66a:
Lift lexically-scoped references within control operators to be inputs of the
ops themselves. This transformation yields a graph that does not conform to
the ONNX spec.
The purpose of this pass is to expose the data dependencies within control
blocks for frameworks that use those dependencies to schedule parallel
execution. e.g. caffe2 graph execution.

There are a few edits here and there to allow for this transformation, but nothing earth-shattering.

### DIFF
--- a/onnx/common/ir_pb_converter.cc
+++ b/onnx/common/ir_pb_converter.cc
@@ -6,7 +6,7 @@
 namespace ONNX_NAMESPACE {
 
 // Part 1: convert ONNX Protobuf to IR
-std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp);
+std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp, bool nested=false);
 
 Tensor tensorProtoToTensor(const ONNX_NAMESPACE::TensorProto & tp) {
   Tensor ret;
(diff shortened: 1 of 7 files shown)


onnx/common/ir_pb_converter.cc
```diff
@@ -6,7 +6,7 @@
 namespace ONNX_NAMESPACE {
 
 // Part 1: convert ONNX Protobuf to IR
-std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp);
+std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp, bool nested=false);
```
anderspapitto: since this isn't a public api and has like 1 or 2 callsites, i'd slightly prefer to not have the parameter defaulted.

anderspapitto:
yeah, overall I'm fine with this. We're not really committing hard to anything; if we come up with a cleaner place to put this logic down the line (e.g. directly in the caffe2 backend.py) we can switch pretty easily.

anderspapitto:
I think it would be slightly better to put the unresolved references directly inside your __new_attribute, rather than as extra inputs - it would upset the checker less

onnx/test/optimizer_test.py
```diff
@@ -13,10 +13,11 @@
 
 class TestOptimizer(unittest.TestCase):
 
-    def _optimized(self, graph, opts):
+    def _optimized(self, graph, opts, check=True):
```
houseroad: Make checking unhappy does not look good. Let's store the input names in the attributes then :-)

Issue #805, opened 2018-04-23
attr value of ConstantFill should be required or has default value


fumihwh:
As title, I think attr `value` of ConstantFill should be required or has default value.
I found this problem in `test_operator_lstm`'s model.onnx, it has ConstantFill node with attrs {'input_as_shape': 1}. 
How could I fill output without value provided?

houseroad:
Good catch, the default value is 0.

Thanks!

PR #808, opened 2018-04-23
Fix the ConstantFill spec

houseroad:
https://github.com/onnx/onnx/issues/805


fumihwh:
@houseroad 
I am not sure if we should also fix caffe2.
https://github.com/caffe2/caffe2/blob/master/caffe2/operators/filler_op.cc#L43

houseroad:
@fumihwh yeah, feel free to update the doc there :-)

PR #809, opened 2018-04-24
low-precision support in onnx for quantization

linkerzhang:
To move this forward, this is the PR per the proposal we discussed (although more discussion may be needed). it contains,
1. proto change for handling <8 bits quantization in future (more meta-data may be needed in packing information).
2.  Conv_Integer and Matmul_Integer operator defined for supporting low-precision data types.
3. More ops to support low-precision data types will be added in separate PRs later.

@liwchang


fumihwh:
@linkerzhang 
Great works!
So this PR defines conv and matmul for quantization data types as you said in 2.
input and weight are assumed to int, right?
If true, will the op, called A_Quantizer that processes them from float to int be defined in the future?

linkerzhang:
@fumihwh Great question! 

"If true, will the op, called A_Quantizer that processes them from float to int be defined in the future?"

Yes, it's true. that's why in this PR the op called as "*_Integer".  The op, called A_Quantizer that quantize a float to int, may or may not be defined as part of ONNX. Common quantizer ops may be defined in onnx as a Function #802 . Cusotmized quantizer ops should be owned by customers as custom function (sub-graph).

sergeigofman:
@linkerzhang Hi Ke,
Thanks for pushing this.
A couple of questions/opens:
1.	You have defined Conv_Integer and MatMul_integer operators that accept input and weight. In order to carry out the operation, need to know that zero_points of these tensors. Where is this information contained? Is it a meta-data of the tensor itself? If yes, can you point me where is it defined in the schema?
2.	It would be beneficial that the Bias for those operators would be int32, because we would like to merely add it to the accumulator.

liwchang:
@sergeigofman 
There is a "Z" in Conv_integer containing zero_point information of X for padding.
Z is a tensor, not an attribute, since it MIGHT be computed during runtime.
For MatMul_integer itself, I don't think zero_point is necessary. 
(zero_points are needed for Dequanizing back float later, but MatMul itself just doesn't need it.)

For the Bias, we were thinking users can simply use Add after Conv_integer. 
Since ONNX only captures semantics, it is up to platforms to fuse them to their own implementation or not.

liwchang:
@sergeigofman 
That doesn't counteract what I said. 
As you said there are different zero_point setting, 
that is the exact reason we exclude zero_points out of MatMul_Integer.
Since MatMul is linear, we should be used MatMul_Integer plus with other ONNX operators to define a quantized MalMul.

1. Regardless what zero_point definition you use, you can also transform it back to a linear equation. 
x_float = x_int*x_scale + x_base. 
w_float = w_int*w_scale + w_base.
Sum(x_float*x_float) = Sum(x_int*w_int*x_scale*w_scale ) + Sum(x_int*x_scale*w_base)
                                     +Sum ( w_int*w_scale *x_base) +Sum(x_base*w_base).
                                 = x_scale*w_scale*Sum(x_int*w_int) + x_scale*w_base*Sum(x_int)
                                   + w_scale *x_base*Sum ( w_int) + x_base*w_base*Sum(1).

MatMul_Integer is used for the first term, which is Sum(x_int*w_int).
The rest should be covered by the existing ONNX operators.

2. If you use the following formula.
x_float = (x_int - x_zero)*x_scale 
2a. if x_zero is the same type as x_int.
You can  simply do Sub op before MatMul_Integer.

2b. If x_zero is the float type, 
you can linearize it to
x_float =  (x_int - x_zero)*x_scale  = x_int*x_scale - x_zero*x_scale = x_int*x_scale + x_base

I think using generic MalMul_Integer should be enough to cover all semantics of linear quantization in MatMul.

sergeigofman:
@liwchang Thanks for th clarifications. I better understand your point now. To your points.
1. That would work, but it will make a simple self-contained operator a sequence of multiple operators which might be not that easy identifiable by graph pattern matching for the HW that support these in a fused fashion. Android NN API has it all fused, with zero_point information. But in general what you propose will work.
2a. I cannot really do what you propose and stay within the same bit width, because,e.g. if the x is int8 and x_zero is int8, their difference will overflow 8 bits and then I won't be able to use int8 flavor of MatMul_Integer.
2b. I am not sure I understand your point here. In any case, my assumption is that both x and x_zero are of the same data type. Is this a wrong assumption? I think it has to be.

liwchang:
@sergeigofman 
1. We understand self-contained operator might lead to better performance in some specific architectures. However, I think quantization itself is not stable enough to be standardized at this moment. Hope you can understand it. If users really care about performance, as @Ke mentioned, there is another proposed op called Function #802, which provides a hint to contain a sequence of multiple operators. That might help for some vendors.

2a. I will be surprised that it will cause overflow. 
When x_int is int8 and x_zero is also int8, (x_int - x_zero) using Sub op should stay in int8 range, if 0.0f is in the dynamic range of data or near it. If not, that simply means your dynamic range is far to 0.0f. In those cases, don't you think you pick a wrong quantization formula. Switch to the formula in 1, it should work well. If you really want to use this formula in this case, you can add a Cast op to promote it from int8 to int32 to avoid overflow. Then you can add Function to force self-containing if you have a concern of API mapping in your platform.

2b is the case that x_zero is float. There are tons of quantization methods. You never know what users can come up with. :)

sergeigofman:
@liwchang @linkerzhang 
1. I think I get the “Functions” proposal (#802). It lets users define sequences of nodes that will operate on given inputs, an return resulting outputs. This indeed would allow wrapping the sequence that will be equivalent to a compound node that would need zero points I had in mind. Appreciate your confirmation.
Let’s look at the required sequence however:
(A - zpA) * (B – zpB) = A*B – zpA * SumColumns(B) – zpB * SumRows(A) + zpA * zpB * CommonDimension(A,B)
The first term would map to MatMul_Integer. The last term would map to broadcasted Add (Tensor,Scalar).
The mid terms should map to  broadcasted Mul(zpA, ReduceSum(B, axis=1) and Mul(zpB, ReduceSum(A, axis=0).
In order to be able to execute mid terms, we need to have the ReduceSum() operator to receive inputs of (int/uint)(8/16) and return int32 result. Can we add this extension as a part of this proposal?

2a. If you subtract zero_point from the quantized tensor, it will overflow for sure. Let me give you a concrete example. Let’s assume we have a tensor with real values from 0.0 to 63.75. When we quantize it to int8 with zero_point, the scale will be (63.75-0)/ (127 - -128) = 0.25 and zero_point will be -128.
0.0f is clearly in the dynamic range. Now if I take a positive int8 number from the quantized tensor, say, 3, and subtract the zero_point from it, I get 3 – (-128) = 131 which is not representable by int8. The only case this math will not overflow is when the zero_point is 0, which defeats the purpose of the zero_point. The suggested method of casting to int32 will not let me do GEMM in int8, which is where the efficiency comes from. Therefore, I am actually forced to use the decomposition in my (1).
2b. Your proposal already stipulates that zero_point must be of the same type as input tensor for the reasons of padding. This makes this kind of hypothetical, amn’t I wrong here?

Now to a different point. In order to be able to implement LinearQuantize “function” from existing operators, we need to be able to perform rounding. Currently, ONNX spec contains only the following rounding-related operators: Floor, Ceil. Using these operators, it is only possible to implement the following rounding policies: round-ties-up and round-ties-down (via Floor(Tensor + 0.5), Ceil(Tensor -0.5)). Are we OK to limit ourselves to those rounding methods, or do we want to introduce Round operator which will round ties away from zero, or NearbyInt that lets user chose their preferred rounding method?

liwchang:
@sergeigofman  @linkerzhang 
1. ReduceSum was actually discussed. 
We tend to use ReduceSum(Cast(A, int32)), instead of introducing a new one.
There is a debate whether ReduceSum(Cast(A, int32)) can capture the semantics of loading A in int8, casting it to int32, then doing ReduceSum on it.

2a. If you reserve 1 bit, then it won't. :)
Or you saturate it?
Let's put it this way. 
There are three (or more?) ways to dodge overflow: promoting, reserving, saturating.
2ai. In promoting, you need to perform multiply in int16.
So what you want is loading in int8, multiply in int16, accumulate in int32.
You can use Cast to int16 to keep accuracy. 
2aii. In reserving, you can simply use Sub.
2aiii. In saturating, the question will become whether it can be easily supported in ONNX and that is disscusable. ( Cast(Clip(Sub(Cast))) vs introducing saturation-supported ops)
Let's know if you really use saturating and have a trouble to do that. 
BTW, @linkerzhang, this reminds us to check/add overflow behavior description in Add and Sub, and others.

2b. Yes, it is hypothetical. It is kind of weird using 2b rather than 1.
But we didn't limit users from doing it. MatMul_Integer also didn't have padding to limit users.
There is also no limit to only one quantization schema in one model. 

3. For the rounding, it is a good point.
C++ does have those functions. 
Though ONNX might have "If" op in future, it might be easier to introduce rounding.
Whether there is a need or we have to introduce it in this phase is discussable. 
Let's know if you really need it.

docs/Changelog.md
```diff
@@ -790,6 +790,58 @@ This version of the operator has been available since version 1 of the default O
 <dd>Constrain input and output types to float tensors.</dd>
 </dl>
 
+### <a name="Conv_Integer-1"></a>**Conv_Integer-1**</a>
+
+  The integer convolution operator consumes an input tensor, a filter and a padding value, and computes the output. The production MUST never overflow. The accumulation may overflow if and only if in 32 bits.
```
Maratyszcza: What do you mean by `The accumulation may overflow if and only if in 32 bits.`?

docs/Changelog.md
```diff
@@ -790,6 +790,58 @@ This version of the operator has been available since version 1 of the default O
...
+
+#### Version
+
+This version of the operator has been available since version 1 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>auto_pad</tt> : string</dt>
```
Maratyszcza: I'm not sure we need to support it for quantized convolution

docs/Changelog.md
```diff
@@ -790,6 +790,58 @@ This version of the operator has been available since version 1 of the default O
...
+</dl>
+
+#### Inputs (2 - 3)
+
+<dl>
+<dt><tt>X</tt> : T1</dt>
+<dd>Input data tensor from previous layer; has size (N x C x H x W), where N is the batch size, C is the number of channels, and H and W are the height and width. Note that this is for the 2D image. Otherwise the size is (N x C x D1 x D2 ... x Dn)</dd>
+<dt><tt>W</tt> : T2</dt>
+<dd>The weight tensor that will be used in the convolutions; has size (M x C x kH x kW), where C is the number of channels, and kH and kW are the height and width of the kernel, and M is the number of feature maps. For more than 2 dimensions, the kernel shape will be (M x C x k1 x k2 x ... x kn), where is the dimension of the kernel</dd>
+<dt><tt>Z</tt> (optional) : T1</dt>
```
Maratyszcza: Is `Z` the bias?

onnx/defs/nn/defs.cc
```diff
@@ -308,6 +308,84 @@ computes the output.)DOC";
...
+        0,
+        "Y",
+        "Output data tensor that contains the result of the "
+        "convolution. The output dimensions are functions "
+        "of the kernel size, stride size, and pad lengths.",
+        "tensor(int32)")
+    .TypeConstraint(
+        "T1",
+        {"tensor(int8)", "tensor(uint8)", "tensor(int16)", "tensor(uint16)"},
+        "Constrain input and output types to float tensors.")
```
Maratyszcza: Typo: `float` -> `fixed-point` tensors

Maratyszcza:
@linkerzhang Thanks for putting up the proposal.

Is the expectation that `Conv_Integer` and `MatMul_Integer` produce exact result?

liwchang:
@Maratyszcza 
Do you mean when both matrices' values are all integer, MatMat_Integer will generate the same number you calculate manually?
If so, I think the answer is "yes" when the accumulator doesn't overflow.

I think @linkerzhang also put some description mentioning whether multiplication or accumulation can overflow in what condition to make sure every platform will generate identical results.
(Though I saw you have some comment on it.)

docs/Changelog.md
liwchang: It probably means if it overflows, it only can overflow in 32 bit.

docs/Changelog.md
liwchang: No, Z is the padding value for Tensor X. 
You might need it for padding.

docs/Changelog.md
liwchang: Hmm... I saw "DEPRECATION NOTE: auto_pad is only intended to support legacy uses, and for framework authors, one is explicitly encouraged to use explicit padding specified in the pads attribute." in Conv.

If we drop it in Conv, then we should also drop it here.
Two should be consistent.

onnx/defs/nn/defs.cc
linkerzhang: Good catch! Will fix it.

docs/Changelog.md
linkerzhang: Yes, it means it may overflow only in 32 bits, normally it should never overflow.

Maratyszcza:
Producing exact 32-bit output feels too restrictive:
- Models using `Conv_Integer` and `MatMul_Integer` wouldn't be able to run on FP16/FP32 hardware, due to inexact nature of results. Being able to run a quantized model on conventional floating-point hardware is a very desirable property.
- I know of an upcoming NN accelerator that operates on 8-bit inputs, accumulates to 32 bits, but doesn't support writing out full 32-bit results. Only quantized outputs can be stored.

liwchang:
@Maratyszcza 
Good points. I actually thought about them before.
- For the first one, I think some rounding error (in the final model output) might be acceptable (though it might be questionable that a hw only has f16/32 without int). Even in regular models, people often change algorithms, or execution order to gain performance. Those can also impact rounding error. Let's call it simulated MatMul_Integer. How large rounding error you think we should put? e-7 for f32? e-4 for f16?

- For the second one, you can think it is a MatMul_Integer with following  (Saturate + or some ops +) Cast. Don't forget ONNX is just capturing semantics. It doesn't tell you when to read from/store to dram. Output of an op can be intermediately stored in scratchpad/register/sram. You can use compiler analysis or Function #802 to achieve it. 

The key design here is that some hw/people use 32bit as output, some hw/people use 8bit after some requantization or saturation tricks, but the common ground is using 32bit as accumulator. 
That will make op interface to dirty. So we pick 32bit as "output" (which not implying store back to dram), and let 8bit output conversion tricks as subgraph followed by MatMul_Integer.

onnx/defs/nn/defs.cc
```diff
@@ -1229,4 +1225,91 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "convolution. The output dimensions are functions "
+            "of the kernel size, stride size, and pad lengths.",
+            "tensor(int32)")
+        .TypeConstraint(
+            "T1",
+            {"tensor(int8)",
+             "tensor(uint8)",
+             "tensor(int16)",
+             "tensor(uint16)"},
+            "Constrain input and output types to float tensors.")
```
fdwr: "float tensors" -> "integer tensors"

onnx/defs/nn/defs.cc
```diff
@@ -1229,4 +1225,91 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .TypeConstraint(
+            "T2",
+            {"tensor(int8)",
+             "tensor(uint8)",
+             "tensor(int16)",
+             "tensor(uint16)"},
+            "Constrain input and output types to float tensors.")
```
fdwr: Although allowing the weights to be distinct types from the input does allow flexibility, it also yields 16 (4x4) different combinations to handle :/.

fdwr:
Should redirect readers to PR #1219 now.

prasanthpul:
superseded by https://github.com/onnx/onnx/pull/1908

PR #810, opened 2018-04-24
Correct Tile operator inputs to use int64 instead of T

RyanUnderhill:
For axis and tiles, only integer values make any sense.


linkerzhang:
#757  also fixed this issue. Let's merge this two.

houseroad:
Fix in https://github.com/onnx/onnx/pull/757

PR #812, opened 2018-04-24
Check if CMAKE_BUILD_TYPE set before project().

kit1980:
Checking for CMAKE_BUILD_TYPE must be done before project(), otherwise it will be set to Debug.
See https://cmake.org/pipermail/cmake/2008-September/023808.html for details.
Because of this, the build currently defaults to Debug despite the "defaulting to Release" intent.


linkerzhang:
@kit1980 please sign the CLA accordingly. Thank you!

houseroad:
@onnxbot retest this please

snnn:
Ideally, we should not use CMAKE_BUILD_TYPE in cmake files. We shouldn't force user choose a CMAKE_BUILD_TYPE while invoking cmake command to generate VC projects files on Windows

PR #818, opened 2018-04-24
Change the return type for the zipmap operator to match the description in the spec.

pranavsharma:
Change the return type for the zipmap operator to match the description in the spec.


pranavsharma:
@linkerzhang signed the CLA. Looks like it takes sometime to get updated. Here's what I see when I click on the link to sign the CLA: You have signed the CLA for onnx/onnx.

pranavsharma:
How much time does it take generally for the CLA to recognize that I've already signed the agreement?

wschin:
Do we need to increate the version number of this op?

PR #820, opened 2018-04-24
specify defaults for attributes of Affine op

gramalingam:


linkerzhang:
@gramalingam please also run \onnx\defs\gen_doc.py to update the operator.md and Changelog.md

houseroad:
@gramalingam could you add defaults into the descriptions of these two attributes? And update the doc using gen_doc.py?

PR #821, opened 2018-04-24
Rename real model test case from VGG-16 to ZFNet

houseroad:
To align with our model zoo.

### DIFF
--- a/onnx/backend/test/case/model/__init__.py
+++ b/onnx/backend/test/case/model/__init__.py
@@ -44,8 +44,8 @@ def collect_testcases():
         ('test_resnet50', 'resnet50'),
         ('test_shufflenet', 'shufflenet'),
         ('test_squeezenet', 'squeezenet'),
-        ('test_vgg16', 'vgg16'),
         ('test_vgg19', 'vgg19'),
+        ('test_zfnet', 'zfnet'),
     ]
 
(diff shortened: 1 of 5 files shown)


onnx/backend/test/case/model/__init__.py
```diff
@@ -44,8 +44,8 @@ def collect_testcases():
         ('test_resnet50', 'resnet50'),
         ('test_shufflenet', 'shufflenet'),
         ('test_squeezenet', 'squeezenet'),
-        ('test_vgg16', 'vgg16'),
         ('test_vgg19', 'vgg19'),
+        ('test_zfnet', 'zfnet'),
```
yinghai: what's zfnet? :)

houseroad: Another model for image classification, won ImageNet 2013

houseroad: More details here: https://github.com/onnx/models/tree/master/zfnet

:-)

jspisak: I've never heard of it either. But LGTM.. :)

houseroad: The paper has high citation: 3000+ :-)
Let's read it.

PR #830, opened 2018-04-26
disable shape inference test to unbreak ci

houseroad:
Let's make ci happy


houseroad:
Windows CI was never happy...

houseroad:
@anderspapitto btw, I think we should also delete `onnx/test/backend_shape_inference_test.py`, since it was merged into `test_backend_test.py`

linkerzhang:
it was not a good idea to check in change that does not pass windows ci :).

PR #831, opened 2018-04-26
add pytorch-operator test for tile

fumihwh:
dim overflow means len of repeats is bigger than number of dims in input


bddppq:
Do these test cases already exist in pytorch?

fumihwh:
@bddppq 
You mean https://github.com/onnxbot/onnx-fb-universe/pull/1780?

houseroad:
yeah, I think they are in onnx-fb-universe already. :-)

PR #832, opened 2018-04-26
add install step

ffk0716:


raymondxyang:
@houseroad the changes lgtm but I dont know if its what we want to put libs and headers under environment paths?

ffk0716:
I am using onnx in cmake ExternalProject. Currently I have to locate the libraries in binary_dir. I think it will be connivence if we have offical install step in cmake.

CMakeLists.txt
```diff
@@ -257,6 +257,8 @@ if(BUILD_ONNX_PYTHON)
       )
     target_compile_options(onnx_cpp2py_export PRIVATE /MT)
   endif()
+  install(TARGETS onnx_cpp2py_export
+      LIBRARY DESTINATION lib)
```
yinghai: Why do we need to install this to lib?

CMakeLists.txt
```diff
@@ -320,3 +320,13 @@ endif()
 # ---[ ONNX Interface for Framework Integratin (ONNXIFI)
 add_library(onnxifi INTERFACE)
 target_include_directories(onnxifi INTERFACE onnx)
+
+
+install(TARGETS onnx onnx_proto
+    ARCHIVE DESTINATION lib)
```
bddppq: These can be shared libs as well. Simply remove `ARCHIVE`.

CMakeLists.txt
```diff
@@ -320,3 +320,13 @@ endif()
...
+install(DIRECTORY onnx DESTINATION include
```
bddppq: `onnx` -> `${ONNX_ROOT}/onnx`

ffk0716: Should I squash to one commit or commit a new one?

PR #839, opened 2018-04-27
Fix SELU attributes' default values

houseroad:
After some research, all the existing frameworks use the same default values:
1) alpha=1.6732632423543772848170429916717
2) gamma=1.0507009873554804934193349852946

In most of the frameworks, they are fixed. So we should make them as accurate as possible.


linkerzhang:
Per discussion, let's define it with fixed value for float or double, instead of having it to be truncated. Thank you very much!

PR #845, opened 2018-04-27
Add some math function shape inference

anderspapitto:
these are all pretty trivial

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -61,6 +61,7 @@ Performs element-wise binary {name} (with limited broadcast support).
         "T",
         OpSchema::high_precision_numeric_types(),
         "Constrain input and output types to high-precision numeric tensors.");
+    schema.TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput);
   };
 }
 
@@ -229,7 +230,8 @@ the tensor elementwise.
     .TypeConstraint(
         "T",
         {"tensor(float16)", "tensor(float)", "tensor(double)"},
-        "Constrain input and output types to float tensors.");
+        "Constrain input and output types to float tensors.")
+    .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput);
 
 ONNX_OPERATOR_SCHEMA(LeakyRelu)
     .SinceVersion(6)
@@ -466,7 +468,8 @@ have the same shape and data type.
     .TypeConstraint(
         "T",
         {"tensor(float16)", "tensor(float)", "tensor(double)"},
-        "Constrain input and output types to float tensors.");
+        "Constrain input and output types to float tensors.")
+    .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput);
 
 ONNX_OPERATOR_SCHEMA(Mean)
     .SinceVersion(6)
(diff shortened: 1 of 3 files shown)


onnx/defs/math/defs.cc
```diff
@@ -229,7 +237,14 @@ the tensor elementwise.
...
         {"tensor(float16)", "tensor(float)", "tensor(double)"},
-        "Constrain input and output types to float tensors.");
+        "Constrain input and output types to float tensors.")
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+        if (!hasNInputShapes(ctx, 1)) {
+          return;
+        }
+        propagateShapeFromInputToOutput(ctx, 0, 0);
+      });
```
gramalingam: It would be useful to define this function (unaryElementWiseOpInferenceFn?) in one place and refer to it in the several functions that can use this.

gramalingam: I mean "several ops"

houseroad: You mean extracting this part into one function `propageteShapeTypeFromFirstInput`?

onnx/defs/math/defs.cc
```diff
@@ -584,7 +606,23 @@ if attribute transA is non-zero, same for B and transB.
         "beta",
         "Scalar multiplier for input tensor C",
         AttributeProto::FLOAT,
-        1.0f);
+        1.0f)
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+        if (!hasNInputShapes(ctx, 2)) {
+          return;
```
gramalingam: One simple extension would be: in the non-broad-cast case, use shape of C, if that is available.

onnx/defs/math/defs.cc
gramalingam: Yes

houseroad:
Can we output the original shape information and the computed one when mismatch happens? It's easier for debug.

onnx/defs/math/defs.cc
```diff
@@ -584,7 +606,23 @@ if attribute transA is non-zero, same for B and transB.
...
+
+        auto transAAttr = ctx.getAttribute("transA");
+        bool transA = transAAttr ? static_cast<int>(transAAttr->i()) != 0 : false;
+        auto transBAttr = ctx.getAttribute("transB");
+        bool transB = transBAttr ? static_cast<int>(transBAttr->i()) != 0 : false;
+
+        *ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape()->add_dim() =
+          ctx.getInputType(0)->tensor_type().shape().dim(transA ? 1 : 0);
+        *ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape()->add_dim() =
+          ctx.getInputType(1)->tensor_type().shape().dim(transB ? 1 : 0);
```
houseroad: I think the problem is here. should be `transB ? 0 : 1`

According to the spec, without any transpose, we should have shapes like A = M x K, B = K x N,

anderspapitto:
I've addressed all comments, except for improving the error messages, which I've done in a separate PR (https://github.com/onnx/onnx/pull/855)

PR #846, opened 2018-04-27
Clarify random like operators

skottmckay:
Update RandomUniformLike and RandomNormalLike to clarify the behaviour that the shape is copied from the input tensor rather than read from it. Adjust input tensor type constraints accordingly. 

Fix issue with optional dtype parameter being defaulted. By setting a default you don't know if it was specified or not.


docs/Changelog.md
```diff
@@ -3461,13 +3461,13 @@ This version of the operator has been available since version 1 of the default O
 
 ### <a name="RandomNormalLike-1"></a>**RandomNormalLike-1**</a>
 
-  Generate a tensor with random values drawn from a normal distribution. The shape
-  of the tensor is computed from the input argument and the parameter of the normal distribution
-  specified by `mean` and `scale`.
+  Generate a tensor with random values drawn from a normal distribution. 
+  The shape of the output tensor is inferred from the input tensor,
```
ebarsoum: Shouldn't be using the shape of the inputs and not inferring?

ebarsoum:
I like the change, but this is a breaking change. The OP version need to increase.

linkerzhang:
@ebarsoum We'll bump the version when doing ONNX 1.2 release.

onnx/defs/generator/defs.cc
```diff
@@ -169,11 +170,11 @@ TensorProto message.
           "(Optional) The data type for the elements of the output tensor, if not specified, we will use"
           "the data type of the input tensor.",
           AttributeProto::INT,
-          static_cast<int64_t>(TensorProto::FLOAT))
+          OPTIONAL)
     .Input(
            0,
            "input",
-           "Input tensor to provide shape information.", "tensor(int32)")
+          "Input tensor to infer shape and optionally type information from.", "T")
```
houseroad: Nit: One more space?

onnx/defs/generator/defs.cc
```diff
@@ -131,7 +132,7 @@ TensorProto message.
     .Input(
            0,
            "input",
-           "Input tensor to provide shape information.", "tensor(int32)")
+           "Input tensor to copy shape and optionally type information from.", "T")
```
bddppq: Input and output types should not use the same parameter, they can be different.

linkerzhang: so the out type inference will be, if the "dtype" specified, then output type is "dtype", otherwise, it will use the same type as input?

gramalingam: yes, that's my understanding too. so, we should use T1 and T2 for input and output type.

skottmckay: What should the input type constraints be?

gramalingam: Good question! I guess we have two options:
(1) Allow only the same types permitted for output, or
(2) Allow any type, and have the type-checker separately check that the input type is one of the permitted output types if "dtype" is missing.
(2) seems the right solution since it is less restrictive. (But it is a bit extra work.)

.gitignore
```diff
@@ -78,6 +78,7 @@ compile_commands.json
 
 # generated files
 onnx/version.py
+onnx/onnx_cpp2py_export.cp36-win_amd64.pyd
```
bddppq: nit: Should add `*.pyd` in the "Compiled Dynamic libraries" instead.

onnx/defs/generator/defs.cc
bddppq: Vote for (2)

PR #847, opened 2018-04-28
update defs for reduce, rnn, and tensor depth-space ops

liqunfu:


onnx/defs/reduction/defs.cc
```diff
@@ -73,7 +73,7 @@ The type of the output tensor is integer.)DOC";
     schema.SetDoc(doc);
     schema.Attr(
         "axis",
-        "The axis in which to compute the arg indices",
+        "The axis in which to compute the arg indices. By default it shall compute along the last axis.",
```
linkerzhang: There's already a default "0" specified. Are you changing the spec to use last axis anyway? if yes, you'll need to remove the default value specified below.

onnx/defs/tensor/defs.cc
```diff
@@ -700,7 +700,7 @@ Given `data` tensor, pads, mode, and value.
 ONNX_OPERATOR_SCHEMA(SpaceToDepth)
     .Attr(
         "blocksize",
-        "Blocks of [blocksize, blocksize] are moved.",
+        "Blocks of [blocksize, blocksize] are moved. Default blocksize is 1.",
         AttributeProto::INT,
         OPTIONAL)
```
linkerzhang: replace "OPTIONAL" with static_cast<int64_t>(1) then.

onnx/defs/reduction/defs.cc
```diff
@@ -19,7 +19,7 @@ False instead of True.)DOC";
     schema.SetDoc(doc);
     schema.Attr(
         "axes",
-        "A list of integers, along which to reduce.",
+        "A list of integers, along which to reduce. If not specified, reduce operation will be along the last axis.",
```
houseroad: We should align with numpy, if no axes are specified, we should do the reduction along all the dimensions.

onnx/defs/reduction/defs.cc
houseroad: Let's keep 0 as default value, unless we have good reason for that.

onnx/defs/tensor/defs.cc
houseroad: In my opinion, this should be a required field

onnx/defs/tensor/defs.cc
```diff
@@ -701,8 +701,7 @@ ONNX_OPERATOR_SCHEMA(SpaceToDepth)
     .Attr(
         "blocksize",
         "Blocks of [blocksize, blocksize] are moved.",
-        AttributeProto::INT,
-        OPTIONAL)
+        AttributeProto::INT)
```
linkerzhang: @houseroad / @bddppq / @ebarsoum , are we all ok to change to the "blocksize" from "optional" to "required" please? or actually this should be "required" and was a spec bug?

bddppq: @linkerzhang I agree this should be required and should be considered as a bug (instead of a breaking change).

onnx/defs/reduction/defs.cc
liqunfu: resolved

onnx/defs/reduction/defs.cc
liqunfu: resolved

onnx/defs/tensor/defs.cc
liqunfu: changed required. this is resolved.

onnx/defs/tensor/defs.cc
houseroad: I think it's spec bug. Usually, users should aware what he is doing, :-)

PR #848, opened 2018-04-28
function registration c++ API

linkerzhang:
Adding function registration c++ API, so that common onnx functions will  be able to registered later.


onnx/checker.cc
```diff
@@ -360,6 +360,98 @@ void check_graph(
   }
 }
 
+Status check_function(
```
bddppq: All other checker functions are using exceptions, this new `check_function` should be consistent with the others.

onnx/defs/schema.h
```diff
@@ -12,16 +12,18 @@
 #include <ostream>
 #include <set>
 #include <string>
-#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
 
 #include "data_type_utils.h"
+#include "onnx/checker.h"
```
bddppq: what do you need from checker in schema? this will introduce circular dependencies.

onnx/onnx-operators_pb.h
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Facebook Inc. and Microsoft Corporation.
+// Licensed under the MIT license.
+
+#ifndef ONNX_ONNX_OPERATOR_PB_H
+#define ONNX_ONNX_OPERATOR_PB_H
+
```
bddppq: This file also needs [these boilerplates](https://github.com/onnx/onnx/blob/master/onnx/onnx_pb.h#L7-L59)

onnx/defs/schema.h
```diff
@@ -667,4 +669,39 @@ class OpSchemaRegistry final {
...
+  void Register(const FunctionBuilder& function_builder);
+
+  // Get functions for specific domain.
+  Status GetFunctions(
+      const std::string& domain,
+      /*out*/
+      std::multimap<std::string, std::shared_ptr<FunctionProto>>* function_set)
+      const;
+
+  static FunctionBuilderRegistry& OnnxInstance();
```
bddppq: nit: it's already inside the onnx namespace, so IMO no need to have "Onnx" in its name

bddppq:
Please provide an example usage of this API.

linkerzhang:
@bddppq Thank you very much for the comments!

1. This API is used to register a function, as we designed functionproto and are going to use it for composite ops. There has to be a way of register such composite ops/functions, similar with the current op schema registration.
2. Having check* function to return Status instead of throwing exception is something we have ever discussed, but I didn't get chance to introduce the "Status" class into ONNX.
3. Having "OnnxInstance" is giving the upper layer a chance to also have its own local instance of FunctionBuilderRegistry instance. OnnxInstance is always immutable.

Make sense?

onnx/defs/schema.h
linkerzhang: Fixed.

onnx/checker.cc
linkerzhang: I explained below. a while ago, I talked to Dima somehow to introduce "Status" class into onnx for error handling instead of throwing exceptions. Now, I'm adding it, it will take some time to change existing exception usage though.

linkerzhang: btw, I can move this function into function.h/cc for now, if you feel that's better. :)

bddppq: Oh I didn't know you had discussions with @dzhulgakov already. My main concern is having **inconsistent** styles of error handling in our api. What was the problem of using exceptions?

bddppq:
@linkerzhang I know at a high level this is a set of new api for registering functions. What I wanted to see is a concrete (toy) example usage of this API, this helps exercising from a user's perspective.

bddppq:
re. separate instances of registries, are you saying for custom domains? so we are handling it differently with the current op schema registry (which uses a single instance)?

linkerzhang:
@bddppq ,toy cases will be added as test cases later in separate PR after https://github.com/onnx/onnx/pull/763 merged in (Raymond is actively working on it).

"re. separate instances of registries, are you saying for custom domains? so we are handling it differently with the current op schema registry (which uses a single instance)?"

Yes, there're cases that upper layer wants to dynamically load/unload customized  function registration. So, we'd allow more than one function registratry instance here.

Make sense?

onnx/checker.cc
linkerzhang: Whether using exception for error handling, I believe it could be a long-debate question. As a core ONNX lib, I personally think it's better to not throw exceptions, especially in some of our static registration codes, throwing exceptions or "abort" process is not a good idea. I agree with you that we should have consistent error handling. I'm targeting to move all error handling to use "Status", instead of throwing exceptions. Thoughts? Again, if there's different opinion here, to avoid long debate on this PR, I can move this function into function.h instead of putting it here. Which way do you prefer please?

bddppq: I don't think moving into a separate file will help the issue of inconsistency.
If we are doing this at the beginning, I probably won't be against using status code (as you said it's a long-debate question that I don't want us get into it if possible). However, considering this api has been established for a while (so there could be many users already), I prefer to not change it.

bddppq:
Re. toy use case I was not requesting to add it into the code, just a code snippet as a comment in this thread would be enough. TBH I don't see how is this going to be used. e.g. I don't see any registration macros like in operator schemas so I don't know how are you going to register functions in the onnx core domain.
Re. dynamically registering functions with custom domains, operator schemas also supports that with single registry instance, right?

onnx/checker.cc
linkerzhang: We can't allow "exception" and "abort" in our static registration code, in windows ml scenario. That's why we should avoid using exception as much as possible. To me, it's acheiveable to not use exception in onnx. Make sense?

onnx/onnx-operators_pb.h
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Facebook Inc. and Microsoft Corporation.
```
bddppq: This file also needs [these boilerplates](https://github.com/onnx/onnx/blob/master/onnx/onnx_pb.h#L11).

onnx/defs/function.cc
```diff
@@ -0,0 +1,91 @@
...
+  return Status::OK();
+}
+
+// Get functions for specific domain.
+Status FunctionBuilderRegistry::GetFunctions(
+    const std::string& domain,
+    /*out*/
+    std::multimap<std::string, std::shared_ptr<FunctionProto>>* function_set)
+    const {
+  if (nullptr == function_set) {
```
bddppq: this seems to be worth to create a macro/function in status.{h,cc} because there will be quite often such check.

onnx/defs/function.cc
```diff
@@ -0,0 +1,91 @@
...
+    auto range = function_set->equal_range(func_name);
+    for (auto i = range.first; i != range.second; ++i) {
+      auto version = i->second->since_version();
+      if (function_proto->since_version() == version) {
+        // There's already a function with same name/since_version registered.
+        return Common::Status(
+            Common::OPSCHEMA,
+            Common::FAIL,
+            "A function (" + func_name + ") with version (" +
+                std::to_string(version) + ") has already been registered.");
```
bddppq: ONNX_NAMESPACE::to_string

onnx/common/status.h
```diff
@@ -0,0 +1,92 @@
...
+
+#include <memory>
+#include <string>
+
+namespace ONNX_NAMESPACE {
+namespace Common {
+
+enum StatusCategory {
+  NONE = 0,
+  OPSCHEMA = 1,
```
bddppq: CHECKER is more appropriate IMO, there are some checks don't belong to schema validation.

onnx/defs/function.h
```diff
@@ -0,0 +1,83 @@
...
+#include <string>
+#include <vector>
+
+#include "onnx/common/status.h"
+#include "onnx/onnx-operators_pb.h"
+
+namespace ONNX_NAMESPACE {
+using namespace Common;
+
+typedef Common::Status (*BuildFunction)(std::shared_ptr<FunctionProto>*);
```
bddppq: why does it take shared_ptr?

onnx/defs/function.cc
```diff
@@ -0,0 +1,91 @@
...
+    // Check no op version conflicts.
+    auto range = function_set->equal_range(func_name);
+    for (auto i = range.first; i != range.second; ++i) {
+      auto version = i->second->since_version();
+      if (function_proto->since_version() == version) {
+        // There's already a function with same name/since_version registered.
+        return Common::Status(
+            Common::OPSCHEMA,
+            Common::FAIL,
+            "A function (" + func_name + ") with version (" +
```
bddppq: string_utils.h has MakeString to simplify this

onnx/defs/function.cc
```diff
@@ -0,0 +1,91 @@
...
+
+Status FunctionBuilderRegistry::Register(
+    const FunctionBuilder& function_builder) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  function_builders.push_back(function_builder);
+  return Status::OK();
+}
+
+// Get functions for specific domain.
+Status FunctionBuilderRegistry::GetFunctions(
```
bddppq: Should there be another version of GetFunction that return the function of specific name and specific version (default to latest)?

onnx/checker.cc
```diff
@@ -360,6 +360,81 @@ void check_graph(
...
 }
 
+void check_function(
+    const FunctionProto& function,
+    const CheckerContext& ctx,
+    const LexicalScopeContext& parent_lex) {
+  enforce_non_empty_field(function, name);
+  enforce_has_field(function, since_version);
+
+  std::unordered_set<std::string> output_names;
```
bddppq: lol it's indeed very very similar to checking graph

linkerzhang: Yes. they're having similar structure.

onnx/defs/function.cc
linkerzhang: I'd not do it but having upper layer to do that, because this get function is indeed calling BuildFunctions to create FunctionProto, I don't want to expose any API for doing this creation again and again.

onnx/defs/function.cc
linkerzhang: You mean macro to create status, right? given I'm using the MakeString in string_utils, I feel it's not needed. the macro will look similar as the current lines of codes.

onnx/onnx-operators_pb.h
linkerzhang: I'm adding #include "onnx/onnx_pb.h" in this file.

onnx/defs/function.cc
linkerzhang: the builder function is not exposed. I don't want to do caching here. Upper layer callers can do that.

onnx/defs/function.cc
```diff
@@ -0,0 +1,96 @@
...
+    const std::string& domain,
+    /*out*/
+    std::multimap<std::string, std::shared_ptr<FunctionProto>>* function_set)
```
bddppq: Why shared_ptr here?

linkerzhang: it should be unique_ptr. Good catch!

onnx/common/status.cc
```diff
@@ -0,0 +1,82 @@
...
+
+  return result;
+}
+
+const Status& Status::OK() noexcept {
+  static Status s_ok;
+  return s_ok;
+}
+
+const std::string& Status::EmptyString() {
```
bddppq: hmm what's the benefit of this (comparing to directly use `""`)?

onnx/common/status.cc
```diff
@@ -0,0 +1,82 @@
...
+  } else if (StatusCategory::OPTIMIZER == state_->category) {
+    result += "[OptimizerError]";
+  }
+
+  result += " : ";
+  result += ONNX_NAMESPACE::to_string(Code());
+  std::string msg;
+
+  switch (static_cast<StatusCode>(Code())) {
+    case INVALID_ARGUMENT:
```
bddppq: Maybe macro each case?

onnx/common/status.h
```diff
@@ -0,0 +1,92 @@
...
+  int Code() const noexcept;
+
+  StatusCategory Category() const noexcept;
+
+  const std::string& ErrorMessage() const;
+
+  std::string ToString() const;
+
+  bool operator==(const Status& other) const {
+    return (this->state_ == other.state_) || (ToString() == other.ToString());
```
bddppq: Better to implement `==` for class `State` so we don't need to construct the string representation of status.

onnx/defs/function.cc
```diff
@@ -0,0 +1,96 @@
...
+  std::lock_guard<std::mutex> lock(mutex_);
+  function_builders.push_back(function_builder);
+  return Status::OK();
+}
+
+// Get functions for specific domain.
+Status FunctionBuilderRegistry::GetFunctions(
+    const std::string& domain,
+    /*out*/
+    std::multimap<std::string, std::unique_ptr<FunctionProto>>* function_set)
```
bddppq: Could you use `std::multimap<std::string, FunctionProto>` here?

onnx/defs/function.h
```diff
@@ -0,0 +1,83 @@
...
+#include <string>
+#include <vector>
+
+#include "onnx/common/status.h"
+#include "onnx/onnx-operators_pb.h"
+
+namespace ONNX_NAMESPACE {
+using namespace Common;
+
+typedef Common::Status (*BuildFunction)(std::unique_ptr<FunctionProto>*);
```
bddppq: Could you use `FunctionProto*` here?

linkerzhang:
@houseroad is the travis ci experiencing some issues please?

onnx/common/status.cc
linkerzhang: I'm seeing it only having a explicit name :).

onnx/defs/function.h
linkerzhang: I'd keep the unique_ptr here to 1) avoid a copy, 2) clarify its unique ownership (folks should not copy it over around).

linkerzhang:
Finally, travis ci passed, I'm merging it.

onnx/common/status.cc
linkerzhang: will check this comment later.

onnx/common/status.h
linkerzhang: will check this comment later.

Issue #819, opened 2018-04-24
onnx/backend/test/case/node/pow.py: y should be a scalar


snnn:
In the unitest of pow_broadcast
```python
x = np.array([1, 2, 3]).astype(np.float32)
y = np.array([2]).astype(np.float32)
z = np.power(x, y)  # expected output [1., 4., 9.]
```

x has shape of [3] ,y has shape of [1], which should be []. Scalars should have shape of [] instead of [1].

PR #852, opened 2018-04-30
fix y in pow test case to scalar

fumihwh:
#819


houseroad:
Thanks for fixing the problem!!!

snnn:
Unfortunately, this fix doesn't really work.

After the fix, the ValueInfoProto of graph input y became:
```
name: "y"
type {
  tensor_type {
    elem_type: FLOAT
    shape {
    }
  }
}
```
It's perfect. But,

in onnx\defs\data_type_utils.cc, function  DataTypeUtils::ToString, we have such code:
```
 switch (type_proto.value_case()) {
    case TypeProto::ValueCase::kTensorType: {
      if (type_proto.tensor_type().has_shape() &&
          type_proto.tensor_type().shape().dim_size() == 0) {
        // Scalar case.
        return left + ToDataTypeString(type_proto.tensor_type().elem_type()) +
            right;
      } else {
        return left + "tensor(" +
            ToDataTypeString(type_proto.tensor_type().elem_type()) + ")" +
            right;
      }
    }
```

So the type string of this input would be 'float' , not 'tensor(float)'. 
It doesn't match the spec of 'Pow' .

houseroad:
@snnn I think this change works perfectly. I manually checked the tensor proto and model proto the script generated. Their dims' len are both 0. I think what you want is here: https://github.com/onnx/onnx/pull/858. Or you want to add `floa16`, `float`, `double` to the type of second input.

snnn:
Thanks.

PR #856, opened 2018-04-30
More shape inference impls

anderspapitto:


onnx/defs/math/defs.cc
```diff
@@ -112,6 +112,7 @@ will throw errors.
         "T",
         {"tensor(float16)", "tensor(float)", "tensor(double)"},
         "Constrain input and output types to float tensors.");
+    schema.TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput);
```
houseroad: This is wrong for when dimension is higher than 2.

onnx/defs/nn/defs.cc
```diff
@@ -23,6 +23,94 @@ static std::string auto_pad_doc =
...
+    }
+  } else {
+    for (size_t i = 0; i < n_input_dims; ++i) {
+      dilations.push_back(1);
+    }
+  }
+
+  std::vector<int64_t> kernel_shape;
+  if (getRepeatedAttribute(ctx, "kernel_shape", kernel_shape)) {
+    if (kernel_shape.size() != ctx.getInputType(1)->tensor_type().shape().dim_size() - 2) {
```
gramalingam: getInputType(0)?

onnx/defs/nn/defs.cc
```diff
@@ -23,6 +23,94 @@ static std::string auto_pad_doc =
...
+      return;
+    }
+  } else {
+    for (int i = 2; i < ctx.getInputType(1)->tensor_type().shape().dim_size(); ++i) {
```
gramalingam: getInputType(0)?

onnx/defs/nn/defs.cc
```diff
@@ -23,6 +23,94 @@ static std::string auto_pad_doc =
...
+  }
+
+  *ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape()->add_dim() =
+    ctx.getInputType(0)->tensor_type().shape().dim(0);
+  *ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape()->add_dim() =
+    ctx.getInputType(1)->tensor_type().shape().dim(0);
+
+  for (size_t i = 0; i < kernel_shape.size(); ++i) {
+    if (!ctx.getInputType(0)->tensor_type().shape().dim(2 + (int)i).has_dim_value()) {
+      continue;
```
gramalingam: Add an (unknown) dim to output shape in this case?

onnx/defs/nn/defs.cc
```diff
@@ -23,6 +23,94 @@ static std::string auto_pad_doc =
...
+
+  // don't bother with legacy auto_pad for now
+  if (ctx.getAttribute("auto_pad")) {
+    return;
+  }
+
+  size_t n_input_dims = (size_t) (ctx.getInputType(0)->tensor_type().shape().dim_size() - 2);
+
+  std::vector<int64_t> dilations;
+  if (getRepeatedAttribute(ctx, "dilations", dilations)) {
```
gramalingam: I don't see attribute "dilations" mentioned in these ops ... should it be added to these ops in spec?

anderspapitto: the pooling ops don't have dilations, but Conv does, and it's harmless to share the code since it is a no-op when unspecified.

However I will add a comment to make this clear in the code

gramalingam: It's not harmless for models where the pooling ops specify a dilations attribute. The pooling ops implementation will ignore the dilations if specified (unless some attribute-checking throws out the model as an invalid model), but the shape inference will take them into account and produce a wrong shape, leading to possible memory corruption if memory is allocation on this basis.

gramalingam: May be we could add a Boolean parameter to the inference function to indicate if dilations should be considered or not?

anderspapitto: do we allow arbitrary extra attributes (e.g. putting a dilation on something that doesn't support it?) If so then yes this is a problem.

In any case yes a flag sounds like a good idea

onnx/defs/math/defs.cc
anderspapitto: ++ so it is

onnx/defs/nn/defs.cc
anderspapitto: yeah, that will work. I guess there's a little more difference between Conv and Pool than I was thinking

onnx/defs/nn/defs.cc
anderspapitto: ++

onnx/defs/nn/defs.cc
anderspapitto: ++

onnx/defs/nn/defs.cc
```diff
@@ -23,6 +23,103 @@ static std::string auto_pad_doc =
...
+  // Pooling operations don't support dilation, only Conv. For
+  // simplicity of the code, we just treat them as having all-1s
+  // dilation.
+  std::vector<int64_t> dilations;
+  if (use_dilation && getRepeatedAttribute(ctx, "dilations", dilations)) {
+    if (dilations.size() != n_input_dims) {
+      return;
+    }
+  } else {
+    for (size_t i = 0; i < n_input_dims; ++i) {
```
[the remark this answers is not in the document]
anderspapitto: sure ++

onnx/defs/nn/defs.cc
```diff
@@ -685,7 +690,15 @@ Output case #2: Y (test mode)
     .TypeConstraint(
         "T",
         {"tensor(float16)", "tensor(float)", "tensor(double)"},
-        "Constrain input and output types to float tensors.");
+        "Constrain input and output types to float tensors.")
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        if (!ctx.getAttribute("is_test") ||
```
houseroad: Nit: Whether we have `is_test` or not, calling `propagateShapeAndTypeFromFirstInput` should be fine.

PR #857, opened 2018-04-30
add type inference for traditional ml ops except classifier ops.

linkerzhang:


onnx/defs/traditionalml/defs.cc
```diff
@@ -74,7 +74,18 @@ ONNX_OPERATOR_SCHEMA(CastMap)
         "max_map",
         "if map_form packing is SPARSE, what is the total length of each output in N (max index value)",
         AttributeProto::INT,
-        static_cast<int64_t>(1));
+        static_cast<int64_t>(1))
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+      auto& cast_to = ctx.getAttribute("cast_to")->s();
```
gramalingam: getAttribute() can return null if the attribute is not specified. check for null (and do nothing if it is null). (In the longer run, we should probably add methods to InferenceContext so that we can return an error-code to indicate the exact typing-errors for better error messages.)

onnx/defs/traditionalml/defs.cc
```diff
@@ -121,7 +132,16 @@ ONNX_OPERATOR_SCHEMA(CategoryMapper)
         "default_int64",
         "int value to use if the string is not in the map",
         AttributeProto::INT,
-        static_cast<int64_t>(-1));
+        static_cast<int64_t>(-1))
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+      auto input_elem_type = ctx.getInputType(0)->tensor_type().elem_type();
```
gramalingam: Add a hasNInputTypes(1) check?

onnx/defs/traditionalml/defs.cc
linkerzhang: fixed.

PR #858, opened 2018-04-30
fix string representation of scalar types

gramalingam:
Change the treatment of scalars in DataTypeUtils::ToType(TypeProto). We need a unique representation of types as strings here for type-checking to work. Since all runtime scalar values are represented as tensors, removing the special case treatment of tensors with an empty-sequence shape.


onnx/defs/data_type_utils.cc
```diff
@@ -119,16 +119,11 @@ std::string DataTypeUtils::ToString(
     const std::string& right) {
   switch (type_proto.value_case()) {
     case TypeProto::ValueCase::kTensorType: {
-      if (type_proto.tensor_type().has_shape() &&
-          type_proto.tensor_type().shape().dim_size() == 0) {
-        // Scalar case.
-        return left + ToDataTypeString(type_proto.tensor_type().elem_type()) +
-            right;
-      } else {
+        // Note: We do not distinguish tensors with zero rank (a shape consisting of
```
houseroad: This method is only called in the schema. Do we really need to have special handling for the scalar tensor in the type system? If not, this change looks good to me. Otherwise, we should change the spec instead.

snnn:
Good!

PR #860, opened 2018-04-30
Eliminate unused initializer

fumihwh:
Add optimizer for eliminating unused initializer.


fumihwh:
@houseroad 
So in that case we also need to remove it, right?

houseroad:
@fumihwh correct, we should also remove unused input, too. That's why I ask for this case.

onnx/optimizer/passes/eliminate_unused_initializer.h
```diff
@@ -0,0 +1,80 @@
...
+      // put any initializer used as input of any node to new vector
+      for (auto* input : n->inputs()) {
+        if (std::find(
+                graph.initializer_names().begin(),
+                graph.initializer_names().end(),
+                input->uniqueName()) != graph.initializer_names().end()) {
+          Tensor initializer;
+          for (auto _init : graph.initializers()) {
+            if (_init.name() == input->uniqueName()) {
+              initializer = _init;
```
dzhulgakov: this is going to copy potentially big struct. Feel free to change Graph's methods to manipulate initializers if necessary

onnx/optimizer/passes/eliminate_unused_initializer.h
```diff
@@ -0,0 +1,80 @@
...
+            }
+          }
+          new_initializers.push_back(initializer);
+          new_initializer_names.push_back(input->uniqueName());
+        }
+      }
+    }
+    // get diff between old and new initializers vector
+    std::vector<std::string> diff_initializer_names, removed_initializer_names;
+    std::set_difference(
```
dzhulgakov: it assumes sorted ranges which is not the case. Do you want to use std::unordered_set instead?

onnx/optimizer/passes/eliminate_unused_initializer.h
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  void eliminate_unused_initializer(Graph& graph) {
+    std::vector<Tensor> new_initializers;
+    std::vector<std::string> new_initializer_names;
+    for (auto it = graph.begin(); it != graph.end(); ++it) {
+      auto* n = *it;
+      DescendOnGraphAttributes(n, [this](Graph& g){eliminate_unused_initializer(g);});
+      // put any initializer used as input of any node to new vector
+      for (auto* input : n->inputs()) {
```
dzhulgakov: it will break if the same initializer is used twice in different nodes

fumihwh: @dzhulgakov 
Thanks. 
By using `std::unordered_set`, it will not be a problem anymore.

houseroad:
Here is more context. Since PyTorch is not smart enough, it may export some unused parameters as initializers. These data should be removed. And in general, since the removed inputs are initializers, users still provide the same input and get the same output, this pass does not really change the semantics. To be safe, we can only remove the input when it’s a useless initializer.

onnx/common/ir.h
```diff
@@ -837,6 +837,22 @@ friend struct Value;
     initializers_.push_back(std::move(initializer));
     initializer_names_.push_back(std::move(name));
   }
+  void eraseInitializer(std::string name) {
+    initializers_.erase(
+        std::remove_if(
+            initializers_.begin(),
+            initializers_.end(),
+            [&name](Tensor initializer) {
```
houseroad: Tensor& initializer? otherwise the cost may be very high if initializer is large.

onnx/optimizer/passes/eliminate_unused_initializer.h
```diff
@@ -0,0 +1,73 @@
+// ATTENTION: The code in this file is highly EXPERIMENTAL.
+// Adventurous users should note that the APIs will probably change.
+
+#pragma once
+
```
houseroad: Nit: add some example to explain the behavior of this pass, something like https://github.com/onnx/onnx/blob/master/onnx/optimizer/passes/fuse_add_bias_into_conv.h#L6

onnx/optimizer/passes/eliminate_unused_initializer.h
```diff
@@ -0,0 +1,73 @@
...
+    // get outputs unique names
+    std::vector<std::string> outputs_unique_names;
+    std::transform(
+        graph.outputs().begin(),
+        graph.outputs().end(),
+        std::back_inserter(outputs_unique_names),
+        [](Value* output) { return output->uniqueName(); });
+    for (auto it = graph.begin(); it != graph.end(); ++it) {
+      auto* n = *it;
+      DescendOnGraphAttributes(n, [this](Graph& g){eliminate_unused_initializer(g);});
```
houseroad: Here, we cannot simply call DescendOnGraphAttributes, because we also need record all the inputs of the nodes in the subgraph.

onnx/test/optimizer_test.py
```diff
@@ -153,6 +153,74 @@ def test_nop_transpose_default(self):
         assert len(list(optimized_model.graph.node)) == 1
         assert optimized_model.graph.node[0].op_type == "Transpose"
 
+    def test_eliminate_unused_initializer(self):
```
houseroad: Nit: add one test case with control flow graph (while/if operator, so you can test subgraph case).

fumihwh:
@houseroad Done~~~;)

onnx/optimizer/passes/eliminate_unused_initializer.h
```diff
@@ -0,0 +1,66 @@
+// ATTENTION: The code in this file is highly EXPERIMENTAL.
+// Adventurous users should note that the APIs will probably change.
+
+#pragma once
+
+// Before:
+//   A = Const
```
houseroad: We are not able to remove unused Const :-)
To be more precise,  you can say `A, B, C are in the initializer list`

fumihwh: @houseroad initializer could not be a const?

houseroad: kConstant means you have a `Constant` op, see https://github.com/onnx/onnx/blob/master/docs/Operators.md#constant

kParam means the initializers, I think.

onnx/test/optimizer_test.py
```diff
@@ -162,6 +162,104 @@ def test_nop_transpose_default(self):  # type: () -> None
...
+             helper.make_tensor_value_info("Y", TensorProto.FLOAT, (1, 2)),
+             helper.make_tensor_value_info("A", TensorProto.FLOAT, (2, 3))],
+            [helper.make_tensor_value_info("Z", TensorProto.FLOAT, (1, 2))],
+            [helper.make_tensor("A", TensorProto.FLOAT,
+                                dims=(2, 3),
+                                vals=np.random.randn(2, 3).astype(np.float32).tobytes(),
+                                raw=True)])
+        optimized_model = self._optimized(graph, ["eliminate_unused_initializer"])
+
+        assert len(list(optimized_model.graph.initializer)) == 0
```
houseroad: Nit: assert len(input) == 2

onnx/optimizer/passes/eliminate_unused_initializer.h
fumihwh: @houseroad Maybe we should clarify them in future.

PR #867, opened 2018-05-01
Werror sign compare

anderspapitto:
note: review top commit only, the bottom commit is https://github.com/onnx/onnx/pull/856


anderspapitto:
@dzhulgakov does it look good to you now?

PR #868, opened 2018-05-01
add shape inference for reduce ops

souptc:
This pr is to add shape inference function for reduce ops.


onnx/defs/reduction/defs.cc
```diff
@@ -33,6 +34,47 @@ False instead of True.)DOC";
...
         "Constrain input and output types to high-precision numeric tensors.");
+    schema.TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+      propagateElemTypeFromInputToOutput(ctx, 0, 0);
+      if (!hasNInputShapes(ctx, 1)) {
+        return;
+      }
+
+      int64_t keep_dims = 1;
+      auto attr_proto = ctx.getAttribute("keepdims");
+      if (attr_proto) {
```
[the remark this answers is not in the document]
bddppq: We should basically rip off ArgumentHelper from https://github.com/pytorch/pytorch/blob/master/caffe2/utils/proto_utils.h

onnx/defs/reduction/defs.cc
```diff
@@ -33,6 +34,47 @@ False instead of True.)DOC";
...
+        if (!attr_proto->has_i()) // incorrect attribtue
```
bddppq: I would really prefer not to silently swallow illegal error here. IMO once shape inference needs to handle error checking, the code will become very verbose, the rule should be shape inference will assume the whole model file has already passed the checks in checker.

souptc: I agree, if onnx guarantee the model already resolved correctly, I am happy to remove those checks here.

onnx/defs/reduction/defs.cc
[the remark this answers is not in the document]
souptc: yes, but I didn't find such a helper in onnx, add such a help seems out of this pr's scope, so I prefer to address it later.

onnx/defs/reduction/defs.cc
linkerzhang: We'll need to return sth in these cases. I'm going to add "Status" as the return value for the inference function. Discussed with Anders before on this too. It's not added at the beginning is because there's no Status class in ONNX repo at that time.

gramalingam:
Looks good to merge

Issue #866, opened 2018-05-01
Trigonometric functions support?


zasdfgbnm:
Is there any plan, or is it contribute welcome, to support trigonometric functions?
sin, cos, tan, asin, acos, atan

These functions are widely supported, for example the cosine function can be found in:
[pytorch](http://pytorch.org/docs/master/tensors.html#torch.Tensor.cos), [tensorflow](https://www.tensorflow.org/api_docs/python/tf/cos), [numpy](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.cos.html), [mxnet](https://mxnet.incubator.apache.org/api/python/symbol/symbol.html#mxnet.symbol.cos), [cntk](https://cntk.ai/pythondocs/cntk.ops.html#cntk.ops.cos)

bddppq:
I agree these should be added to onnx. Are you interested in writing up a PR? :-)

bddppq:
@zasdfgbnm thanks. here is a previous PR that adds a new operator into onnx: https://github.com/onnx/onnx/pull/496 you can take a look to use it as a reference.

zasdfgbnm:
Fixed.

PR #869, opened 2018-05-02
add trigonometric functions

zasdfgbnm:
https://github.com/onnx/onnx/issues/866


zasdfgbnm:
There is something I'm not sure what to do:
One is whether I need to put a `.SinceVersion(6)` in `def.cc` and if needed, which version to put.
Another is I don't know how to generate the pb files in `onnx/backend/test/data/node/`

fumihwh:
As new ops, since version 1 is OK. As default SinceVersion is 1, so you could ignore it.
Use cmd_tools.py to generate test data.

zasdfgbnm:
@fumihwh Data added, thanks!

linkerzhang:
The since version should be a new version bumped against the latest/max version of a domain. It should be 7 (6+1). (odd version is used as developer version, while even version is used for release). We'll bump the version to be 8 when doing next ONNX release.

btw, the since version (8) in coming ONNX release is telling customers that onnx domain operator set has these new operators in version 8. 

@ebarsoum  is going to improve the documentation about the versioning of operator set, btw.

Thank you!

zasdfgbnm:
@linkerzhang I'm trying to add `.SinceVersion(7)` to operators, but get the following error:
```
Trying to register schema with name Sin (domain:  version: 7) from file /home/gaoxiang/onnx/onnx/defs/math/defs.cc line 668, but it its version is notin the inclusive range [1, 6] (usually, this means you bumped the operator version but forgot to update the version range in DomainToVersionRange in onnx/defs/schema.h).
fish: “python onnx/defs/gen_doc.py” terminated by signal SIGABRT (Abort)
```

Is this an expected behavior which means that I should go ahead and update `onnx/defs/schema.h`?

zasdfgbnm:
@linkerzhang Fixed

bddppq:
Test data will need to be regenerated as well.

houseroad:
Hold on, the model zoo is not ready yet, since you bump up the opset version

bddppq:
@zasdfgbnm @linkerzhang @houseroad I agree we should do the version bump in a separate PR (this PR will need to wait for that to be merged first).

linkerzhang:
Should we only bump the version for all test data when doing ONNX release?

linkerzhang:
Agreed on bumping the version for test data separately. Thank you very much!

linkerzhang:
We may again need to do it when doing ONNX release :(.

houseroad:
as @ebarsoum said, if there is no change on op spec, we should not bump up the opset version. Sorry, I would like to retract my odd-even version proposal. Instead, I think @bddppq's original proposal is better, we only bump up the opset version when the following two conditions are both true:
1) we are about to introduce some breaking changes,
2) and we haven't bumped up the opset version since last release.

@linkerzhang, @ebarsoum, @bddppq what do you think? :-)

linkerzhang:
It's OK. The odd/even thing makes the 2nd condition check easier although.

zasdfgbnm:
I'm a bit confused. What is the relationship between version like "1.1.1" and like "7"? Are they independent?

Also, there seems to be dependency cycle between version bump and new op. If we increase the version, then we need new model zoo to pass tests. But without new op set, we can not have a model zoo on this set... So at some point we have to fail tests... 

I would suggest to do the following:
1. The rule is: new op must always be in a new version
1. postpone the version bump until release
1. Add a macro indicating whether the build is develop or product
1. If the build is develop, allow the version to be 1 larger than the upper bound.

This can be done by changing the following line:
https://github.com/onnx/onnx/blob/7d1e102e73ffd0479a87d62c8aa950e03cfdda95/onnx/defs/schema.h#L560
```C++
#ifdef DEVELOP
if (!(lower_bound_incl <= ver && upper_bound_incl + 1 >= ver)) { 
#else
if (!(lower_bound_incl <= ver && upper_bound_incl >= ver)) { 
#endif
```

linkerzhang:
@zasdfgbnm do you want to resolve the conflicts please? Thank you!

zasdfgbnm:
@linkerzhang Done

zasdfgbnm:
Can anyone reviewing this give a quick comment on if theses operators in https://github.com/onnx/onnx/issues/870 are contribute welcome? I want to start working on these, but before that I just need to make sure if they are welcome.

bddppq:
CI failure is irrelevant. @zasdfgbnm Could you resolve the merge conflict?

zasdfgbnm:
@bddppq Done.
Looks like op set 7 is having many new ops and ONNX is becoming more complete :)

zasdfgbnm:
@bddppq Do you mean files in `onnx/backend/test/data/`?
After doing a `python ./onnx/backend/test/cmd_tools.py generate-data`, I get the following changes:
```
On branch trigonometric
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   onnx/backend/test/data/node/test_exp/test_data_set_0/output_0.pb
	modified:   onnx/backend/test/data/node/test_identity/model.onnx
	modified:   onnx/backend/test/data/node/test_log/test_data_set_0/input_0.pb
	modified:   onnx/backend/test/data/node/test_log/test_data_set_0/output_0.pb
	modified:   onnx/backend/test/data/node/test_lstm_defaults/test_data_set_0/output_0.pb
	modified:   onnx/backend/test/data/node/test_lstm_with_initial_bias/test_data_set_0/output_0.pb
	modified:   onnx/backend/test/data/node/test_lstm_with_peepholes/test_data_set_0/output_0.pb
	modified:   onnx/backend/test/data/node/test_matmul_2d/test_data_set_0/output_0.pb
	modified:   onnx/backend/test/data/node/test_reduce_log_sum_asc_axes/model.onnx
	modified:   onnx/backend/test/data/node/test_reduce_log_sum_default/model.onnx
	modified:   onnx/backend/test/data/node/test_reduce_log_sum_desc_axes/model.onnx
	modified:   onnx/backend/test/data/node/test_softplus/test_data_set_0/output_0.pb
	modified:   onnx/backend/test/data/node/test_tanh/test_data_set_0/output_0.pb

no changes added to commit (use "git add" and/or "git commit -a")
```
All those seems to be unrelated. Do I need to generate something else?

bddppq:
@zasdfgbnm Hmm these are unrelated, they should be updated in the PR that bumped up the opset version. I will create a separate PR to fix these.

PR #871, opened 2018-05-02
Type and shape inference for generator ops

gramalingam:
RandomUniform, RandomNormal. RandomUniformLike, RandomNormalLike, and Constant ops.


anderspapitto:
do you have tests? I'd like to have at least a couple simple cases in shape_inference_test.py

implementation looks great

anderspapitto:
either way is fine with me. Feel free to add `# gramalingam new tests here` somewhere in the middle of the file, so we don't all conflict appending to the end :)

linkerzhang:
I just merged without checking all comments. @gramalingam feel free add tests in separate PR then. :)

PR #874, opened 2018-05-02
shape inference for matmul

anderspapitto:
this one is relatively involved, since it requires implementing numpy
broadcasting semantics

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -621,7 +621,105 @@ ONNX_OPERATOR_SCHEMA(MatMul)
         "Constrain input and output types to float tensors.")
     .SetDoc(R"DOC(
 Matrix product that behaves like numpy.matmul: https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.matmul.html
-)DOC");
+)DOC")
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        propagateElemTypeFromInputToOutput(ctx, 0, 0);
+        if (!hasNInputShapes(ctx, 2)) {
+          return;
+        }
+
+        auto shape0 = ctx.getInputType(0)->tensor_type().shape();
+        auto shape1 = ctx.getInputType(1)->tensor_type().shape();
+
+        if (shape0.dim_size() == 0 || shape1.dim_size() == 0) {
+          return;
+        }
+
+        TensorShapeProto paddedShapeL;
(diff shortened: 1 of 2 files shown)


onnx/defs/math/defs.cc
```diff
@@ -621,7 +621,101 @@ ONNX_OPERATOR_SCHEMA(MatMul)
...
+            } else if (dimR == 1) {
+              newdim->set_dim_value(dimL);
+            } else {
+              return;
+            }
+          } else if (paddedShapeL.dim(i).has_dim_value()) {
+            auto dimL = paddedShapeL.dim(i).dim_value();
+            if (dimL != 1) {
+              newdim->set_dim_value(dimL);
+            }
```
gramalingam: else set it to paddedShapeR.dim(...) ? (Autocorrect insists of converting lowercase I to uppercase!)

onnx/defs/math/defs.cc
```diff
@@ -621,7 +621,101 @@ ONNX_OPERATOR_SCHEMA(MatMul)
...
+          } else if (paddedShapeR.dim(i).has_dim_value()) {
+            auto dimR = paddedShapeR.dim(i).dim_value();
+            if (dimR != 1) {
+              newdim->set_dim_value(dimR);
+            }
```
gramalingam: else use param from paddedShapeL?

onnx/defs/math/defs.cc
```diff
@@ -621,7 +621,101 @@ ONNX_OPERATOR_SCHEMA(MatMul)
...
+          auto dimR = paddedShapeR.dim(dimSize - 2);
+          if (dimL.has_dim_value() && dimR.has_dim_value() &&
+              dimL.dim_value() != dimR.dim_value()) {
+            return;
+          }
+        }
+
+        TensorShapeProto resultShape;
+
+        for (int i = 0; i < dimSize - 2; ++i) {
```
gramalingam: Looks good to me ... I guess the broadcasting-logic alone may be reusable across multiple broadcasting ops, if we can factor it out. (I haven't checked the other broadcasting ops yet.)

onnx/defs/math/defs.cc
anderspapitto: yes, I suppose that's a bit better, (merge 1 and 'a' -> 'a', rather than to unknown)

onnx/defs/math/defs.cc
anderspapitto: yes, same here

PR #443, opened 2018-01-23
dimension denotation

tjingrant:
As per https://github.com/onnx/onnx/issues/406

I settled on the word standard_denotation b/c denotation seems to imply "direct meaning". This helps to convey the idea that such denotation is for the algorithms not for some programmer's high level perception.

I hope that people don't confuse these two as this could happen if someone wants to do channel pooling on data tensor of size [N, C, W] by transpose the data tensor into [N, W, C] and then do spatial pool on [N, W, C]; in such case one should still annotate the tensor as [data_batch, data_channel, data_spatial]. But, people doing such hacky implementation should be aware of the potential consequence anyways...

PR #879, opened 2018-05-03
Added TensorDenotation and metadata_props for images

walrusmcd:
This proposal introduces Tensor Denotation 

Tensor Denotation is an experimental attempt along with Dimension Denotation to give tensor semantic description and enable model authors to describe what the input and output tensors are. 

The motivation of such a mechanism can be illustrated via a simple example. In the the neural network SqueezeNet, it takes in an NCHW image input float[1,2,244,244] and produces a output float[1,1000,1,1]: 
input_in_NCHW -> data_0 -> SqueezeNet() -> output_softmaxout_1 
In order to run this model the user needs a lot of information.    In this case the user needs to know: 
* the input is an image 
* the image is in the format of NCHW 
* the color channels are in the order of bgr 
* the pixel data is 8 bit 
* the pixel data is stored as values 0-255 

This proposal consists of three key components to provide all of this information: Tensor Denotation, Dimension Denotation, and model metadata.  Each of which will be discussed in detail.


onnx/onnx.in.proto
```diff
@@ -422,6 +422,10 @@ message TypeProto {
     // This field MUST be present for this version of the IR.
     optional TensorProto.DataType elem_type = 1;
     optional TensorShapeProto shape = 2;
+    // An optional denotation can be used to denote the whole 
+    // tensor with a standard semantic description as to what is 
+    // stored inside
+    optional string denotation = 3;
```
linkerzhang: We'd put this denotation in TypeProto instead of Tensor so that other types (say Map, Sequence) will also be able to use it?

docs/DimensionDenotation.md
```diff
@@ -43,3 +43,7 @@ for i, j in enumerate(perm):
 ## Denotation Verification
 
 Denotation Verification happens when an operation expects its input to arrive in a particular format. An example operation where denotation verification happens is AveragePool operation where the input, if annotated with dimension denotation, in the 2D case should have the denotation [`DATA_BATCH`, `DATA_CHANNEL`, `DATA_FEATURE`, `DATA_FEATURE`]. If there is a mismatch between the expected dimension denotation and the actual dimension denotation, an error should be reported.
+
+## Tensor Denotation
+
+See the [tensor denotation documentation](TensorDenotation.md) for more details on denotation applied at the tensor level.
```
linkerzhang: Merge all denotation documents into one file?

docs/TensorDenotation.md
```diff
@@ -0,0 +1,48 @@
...
+In order to run this model the user needs a lot of information.    In this case the user needs to know:
+* the input is an image
+* the image is in the format of NCHW
+* the color channels are in the order of bgr
+* the pixel data is 8 bit
+* the pixel data is normalized as values 0-255
+
+This proposal consists of three key components to provide all of this information: Tensor Denotation, [Dimension Denotation](DimensionDenotation.md), and [model metadata](MetadataProps.md).  Each of which will be discussed in detail.
+
+## Tensor Denotatio Definition
```
prasanthpul: n is missing -> Denotation

docs/MetadataProps.md
```diff
@@ -0,0 +1,31 @@
...
+## Image Category Definition
+
+For every tensor in this model that uses [Tensor Denotation](TensorDenotation.md) to declare itself an `IMAGE`, you can provide more information to the model consumer.  Note that any metadata provided using this mechanism is global to ALL tensors
+with the accompanying denotation.
+
+Specifically, we define here the following set image metadata:
+
+|Key|Value|Description|
+|-----|----|-----------|
+|`Image.BitmapPixelFormat`|__string__|Specifies the format of pixel data. Each enumeration value defines a channel ordering and bit depth. Possible values: <ul><li>`Gray8`: 1 channel image, the pixel data is 8 bpp grayscale.</li><li>`Rgb8`: 3 channel image, channel order is RGB, pixel data is 8bpp (No alpha)</li><li>`Bgr8`: 3 channel image, channel order is BGR, pixel data is 8bpp (No alpha)</li><li>`Rgba8`: 4 channel image, channel order is RGBA, pixel data is 8bpp (Straight alpha)</li><li>`Bgra8`: 4 channel image, channel order is BGRA, pixel data is 8bpp (Straight alpha)</li></ul>|
```
prasanthpul: can you call out that keys and values are case insensitive?

docs/TensorDenotation.md
```diff
@@ -0,0 +1,48 @@
...
+
+For tensors, Denotation Propagation does not automatically occur.   It is used to describe the initial input or final outputs, but as data flows through the graph no inference is made to propogate if the data still holds an image (for example).   A model builder or conversion tool MAY apply propagation manually in the model if they knows that subsequent tensors share the same semantic denotation.
+
+## Denotation Verification
+
+Denotation Verification is not enforced.   It is simply a method for model authors to indicate to model consumers what they should be passing in and what they should be expecting out.  No error is reported if you do not actually pass in an image (for example).
+
+## Dimension Denotation
+
+Tensor denotation can be combined with the new experimental feature of Dimension Denotation.  For example if the Tensor Denotation is `IMAGE`, then that tensor SHOULD also have [Dimension Denotation](DimensionDenotation.md) stating the NCHW channel layout.  (described as [`DATA_BATCH`, `DATA_CHANNEL`, `DATA_FEATURE`, `DATA_FEATURE`]).
```
tjingrant: NCHW format on a model level is not easily enforceable and thus brings about nightmares in terms of compliance checking. ONNX, as it stands today, only enforces node-level input format, which does not translate to any restrictions on model level format. C.f. my comment copied from the thread discussing Dimension Denotation:


1. > ONNX as a standard will keep as it is to support only NCHW.

    We need to be careful about what you mean. There are no operations supporting NHWC format in ONNX; however, that does not preclude models from supporting NHWC format input. In fact, you cannot preclude such use case as long as you have a **transpose** in the operator set because you can always transpose the input once and enter the NCHW world thereafter. Thus, unless you explicitly pattern match and prohibit these cases in the checker, there is no way to prevent a user from creating  ONNX **models** that accept NHWC input but uses NCHW **operators**. Thus it is my opinion that 
    > This means all ONNX models should be in NCHW format,

    does not hold true as an implication from your previous statement.

docs/MetadataProps.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Motivation
+
+The motivation of such a mechanism is to allow model authors to convey to model consumers enough information for them to consume the model.  
+
+In the case of images there are many option for providing valid image data.  However a model which consumes images was trained with a particular set of these options which must 
+be used during inferencing.
+
+The goal is this proposal is to provide enough metadata that the model consumer can perform their own featurization prior to running the model and provide a compatible input 
+or retrive an output and know what it's format is.
```
linkerzhang: it's --> its ?

onnx/onnx.in.proto
walrusmcd: Yes, I had thought of that option as well.   I think the real challenge here is that we have Tensor as our base …. and all of these other things are sub classes of tensor.   Map and Sequence are first class citizens right now.   I imagine after this experiment, we would promote Image/Audio/Text to first class citezens as well.    Thus my though was to keep it on tensor to allow us to experiment for candidates to promote to first class types.    
Also, maps and sequences don't have the denotations we define in this first iteration.   A map cannot be an image/audio/text.   Nor can a sequence.   I think if we find candidates for denotation on map & sequence (Things that are common to all TypeProto's) ; we could add one then.
 thoughts?

docs/DimensionDenotation.md
walrusmcd: I thought about that , but then it becomes on larger doc.     Preference / thoughts?

if you really like the idea I can take the time to merge them and see how it looks.   lmk.

docs/TensorDenotation.md
walrusmcd: good catch.   fixed in the the next iteration.

docs/MetadataProps.md
walrusmcd: nice !   added to the next iteration.

docs/MetadataProps.md
walrusmcd: fixed in next iteration

docs/TensorDenotation.md
walrusmcd: agreed.   with this we are not proposing enforcing or being strict about NCHW, but instead allowing enough metadata and denotation for a model builder to let a model consumer know what to do.    This PR is less about NCHW (dimension denotation talks about that).  This PR is more about the other things metadata talks about (bitmap pixel format, etc..etc..) .

thoughts?   Would you propose changing anything in the docs or strings for this PR ?  thx !

tjingrant:
@walrusmcd In addition to the wording comments, I would suggest you don't try to draw a parallel between other types of denotations and dimension denotations. And I will outline my reason:

Dimension denotation is a unique subject whose complexity goes very much beyond TensorDenotation or TypeDenotation. The primary goal of dimension denotation is **not** the storage of meta data, but rather the **propagation and verification of the meta data**, which is no where near the goal of other types of denotations, you are wasting spaces trying to explain that the other meta datas are not propagated and verified. Furthermore, you are distracting readers who need to realize the complexity and purpose of dimension denotation. What you are interested in is a very specific sub feature of Dimension Denotation, so the best way to refer to Dimension Denotation is via a link to the original proposal.

Not to mention that, dimension denotation was the result of lengthy discussion, debate and deliberations (c.f., https://github.com/onnx/onnx/pull/443). Even the word `denotation` was carefully chosen to express the connotation of "literal meaning" or "meaning at its lowest level of abstraction", which is BTW a totally unnecessary word choice in this context and may make your proposal a bit less understandable to people whose mother tone is not English. I would kindly ask you please do not move the original proposal around unless it is absolutely necessary.

linkerzhang:
@bddppq @houseroad  please help to review and share your comments if any.

docs/TypeDenotation.md
```diff
@@ -0,0 +1,55 @@
...
+
+There are 2 types of denotation : [Type Denotation](TypeDenotation.md) and [Dimension Denotation](DimensionDenotation.md).
+
+### Type Denotation
+
+Type Denotation is used to describe semantic information around what the inputs and outputs are.    It is stored on the TypeProto message.
+
+#### Motivation
+
+The motivation of such a mechanism can be illustrated via a simple example. In the the neural network SqueezeNet, it takes in an NCHW image input float[1,2,244,244] and produces a output float[1,1000,1,1]:
```
houseroad: A typical SqueezeNet takes [1, 3, 224, 224] as input.

docs/MetadataProps.md
```diff
@@ -0,0 +1,34 @@
+# Metadata
+# Metadata
+
+In addition to the core metadata recommendations listed in the [extensibility documentation](IR.md#metadata) there is additional experimental metadata to help provide information for model inputs and outputs.
```
houseroad: This doc is called metadata propagation, so we should not introduce the additional experimental metadata here.
Let's move the introduction of the experimental metadata to metadata section in IR doc.

docs/TypeDenotation.md
```diff
@@ -0,0 +1,55 @@
+Using Denotation For Semantic Description
+------------------------------------------
+
+Denotation is an experiment to give semantic description to models.  This enables model authors to describe the parts of their model application developers need to know in order to consume them.
+
+There are 2 types of denotation : [Type Denotation](TypeDenotation.md) and [Dimension Denotation](DimensionDenotation.md).
```
houseroad: Why do we mention Dimension Denotation here?

docs/TypeDenotation.md
```diff
@@ -0,0 +1,55 @@
...
+
+#### Motivation
+
+The motivation of such a mechanism can be illustrated via a simple example. In the the neural network SqueezeNet, it takes in an NCHW image input float[1,2,244,244] and produces a output float[1,1000,1,1]:
+
+```
+input_in_NCHW -> data_0 -> SqueezeNet() -> output_softmaxout_1
+```
+
+In order to run this model the user needs a lot of information.    In this case the user needs to know:
```
houseroad: I think instead of using type denotation, saving the following information in the metadata is more straightforward.

docs/TypeDenotation.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. `IMAGE` describes that a type holds an image.  You can use dimension denotation to learn more about the layout of the image, and also the optional model metadata_props.
+2. `AUDIO` describes that a type holds an audio clip.   
+3. `TEXT` describes that a type holds a block of text.
+
+Model authors SHOULD add type denotation to inputs and outputs for the model.
+
+#### Denotation Propagation
+
+Type Denotation propagation does not automatically occur.   It is used to describe the initial input or final outputs, but as data flows through the graph no inference is made to propogate if the data still holds an image (for example).   A model builder or conversion tool MAY apply propagation manually in the model if they knows that subsequent types share the same semantic denotation.
```
houseroad: How does this work?

docs/TypeDenotation.md
walrusmcd: Good catch, I meant 3.  fixed.

docs/TypeDenotation.md
walrusmcd: great catch !!  I think an example would be perfect here.   I'm adding one.

docs/MetadataProps.md
walrusmcd: The example should help to show that we have 3 separate pieces here, and how to tie them all together.

docs/MetadataProps.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Motivation
+
+The motivation of such a mechanism is to allow model authors to convey to model consumers enough information for them to consume the model.  
+
+In the case of images there are many option for providing valid image data.  However a model which consumes images was trained with a particular set of these options which must 
+be used during inferencing.
+
+The goal is this proposal is to provide enough metadata that the model consumer can perform their own featurization prior to running the model and provide a compatible input 
+or retrive an output and know what its format is.
```
lupesko: nit: no need for a new line here

docs/TypeDenotation.md
lupesko: +1 multiple inputs is a good call

docs/TypeDenotation.md
```diff
@@ -0,0 +1,53 @@
...
+## Type Denotation Definition
+
+To begin with, we define a set of semantic types that define what models generally consume as inputs and produce as outputs.
+
+Specifically, in our first proposal we define the following set of standard denotations:
+
+1. `IMAGE` describes that a type holds an image.  You can use dimension denotation to learn more about the layout of the image, and also the optional model metadata_props.
+2. `AUDIO` describes that a type holds an audio clip.   
+3. `TEXT` describes that a type holds a block of text.
+
```
lupesko: What about good old numerical tensors, for other tasks such as recommendations, forecasting, anomaly detection, etc?
Should we add another type TENSOR for those?

onnx/onnx-ml.proto
```diff
@@ -476,6 +476,23 @@ message TypeProto {
...
+  // An optional denotation can be used to denote the whole 
+  // type with a standard semantic description as to what is 
+  // stored inside
+  optional string denotation = 6;
+}
+
+// A set of pre-defined constants to be used as values for
+// the standard denotation field in Tensor for semantic 
+// description of the tensor.
+message TypeDenotationConstProto {
```
lupesko: Please see my comment above about a generic tensor

docs/MetadataProps.md
```diff
@@ -0,0 +1,34 @@
...
+|-----|----|-----------|
+|`Image.BitmapPixelFormat`|__string__|Specifies the format of pixel data. Each enumeration value defines a channel ordering and bit depth. Possible values: <ul><li>`Gray8`: 1 channel image, the pixel data is 8 bpp grayscale.</li><li>`Rgb8`: 3 channel image, channel order is RGB, pixel data is 8bpp (No alpha)</li><li>`Bgr8`: 3 channel image, channel order is BGR, pixel data is 8bpp (No alpha)</li><li>`Rgba8`: 4 channel image, channel order is RGBA, pixel data is 8bpp (Straight alpha)</li><li>`Bgra8`: 4 channel image, channel order is BGRA, pixel data is 8bpp (Straight alpha)</li></ul>|
+|`Image.ColorSpaceGamma`|__string__|Specifies the gamma color space used. Possible values:<ul><li>`Linear`: Linear color space, gamma == 1.0</li><li>`SRGB`: sRGB color space, gamma == 2.2</li></ul>|
+|`Image.NominalPixelRange`|__string__|Specifies the range that pixel values are stored. Possible values: <ul><li>`NominalRange_0_255`:  [0...255] for 8bpp samples</li><li>`Normalized_0_1`: [0...1] pixel data is stored normalized</li><li>`Normalized_1_1`: [-1...1] pixel data is stored normalized</li><li>`NominalRange_16_235`: [16...235] for 8bpp samples</li></ul>|
```
houseroad: In some case, a mean for each channel is also needed to preprocess the input.

docs/MetadataProps.md
walrusmcd: fixed !

docs/TypeDenotation.md
walrusmcd: Makes sense !    I added the 3 we had been using the most in our models (image/audio/text) and added explicit metadata for image.   I assumed that there would be follow up proposals and PR's as we add more "types" here.    What you thinking ?     Also, we do have the normal case where there is no denotation .  In that case the tensor is also a good old numerical tensor.     since denotation is optional, it still has all the fun stuff, like shape and type.

onnx/onnx-ml.proto
walrusmcd: I was thinking we make denotation optional, and "generic" would be that there is no denotation at all.    Are you saying we should add a TENSOR to the constproto as the default/0 case so that you could provide a tensor denotation and say it is just a "tensor" ?

docs/MetadataProps.md
walrusmcd: How do you think that would look ?    Are you OK if we add this as a follow up PR later ?   I think this follows this model of adding more and more metadata as we find it to be useful.   love it !

docs/TypeDenotation.md
```diff
@@ -0,0 +1,53 @@
...
+Let's use the same SqueezeNet example from above and show everything to properly annotate the model:
+
+* First set the TypeProto.denotation =`IMAGE` for the ValueInfoProto `data_0`
+* Because it's an image, the model consumer now knows to go look for image metadata on the model
+* Then include 3 metadata strings on ModelProto.metadata_props
+	* `Image.BitmapPixelFormat` = `Bgr8`
+	* `Image.ColorSpaceGamma` = `SRGB`
+	* `Image.NominalPixelRange` = `NominalRange_0_255`
+* For that same ValueInfoProto, make sure to also use Dimension Denotations to denote NCHW
```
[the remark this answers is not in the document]
walrusmcd: Yes.   We considered this and did not go that route in that the goal was not to do the transformation for you, but instead provide the final values that the model expects.   This allows people using the model to perform their own transformation outside the model.

onnx/onnx-ml.proto
lupesko: Agreed.

docs/TypeDenotation.md
```diff
@@ -0,0 +1,54 @@
...
+* [Dimension Denotation](DimensionDenotation.md), 
+* [Model Metadata](MetadataProps.md).
+
+## Type Denotation Definition
+
+To begin with, we define a set of semantic types that define what models generally consume as inputs and produce as outputs.
+
+Specifically, in our first proposal we define the following set of standard denotations:
+
+0. `TENSOR` describes that a type holds an generic tensor using the standard TypeProto message.
```
lupesko: nit: "a" not "an"

docs/TypeDenotation.md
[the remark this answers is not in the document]
rednoah91: I mean if the metadata contains the transformation information, the model user can know how to perform the transform that the model expects. For example, specify the mean value [123.68, 116.78, 103.94] for VGG net in the metadata in order to hint model user to do the image pre-processing.

docs/TypeDenotation.md
walrusmcd: fixed :)

Issue #426, opened 2018-01-17
Call for Contributions


houseroad:
# Call for Contributions
ONNX is open source, and contributions from the community is welcome. Many tasks, (some are low-hanging fruits), are listed here. If you would like to contribute to the project, but don't know where to start, this list is worth to check.

## Adding ONNX Backend Test Case
Ideally, we should have sample cases for every operator, which can generate [backend test data](https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/) and also serve as [examples](https://github.com/onnx/onnx/blob/master/docs/Operators.md#examples-1). However, many operators have no sample cases, and even not covered by ONNX backend test cases. So please contribute your useful cases to ONNX repository.

[How to contribute ONNX backend test?](https://github.com/onnx/onnx/blob/master/docs/OnnxBackendTest.md)

The operators which don't have examples:
- [x] Abs
- [x] Add
- [x] And
- [x] ArgMax
- [x] ArgMin
- [x] AveragePool
- [x] BatchNormalization
- [x] Cast
- [x] Ceil
- [x] Clip
- [x] Concat
- [x] Conv
- [x] ConvTranspose
- [x] DepthToSpace
- [x] Div
- [x] Dropout
- [x] Elu
- [x] Equal
- [x] Exp
- [x] Flatten
- [x] Floor
- [x] GRU
- [x] Gather
- [x] Gemm
- [x] GlobalAveragePool
- [ ] GlobalLpPool
- [x] GlobalMaxPool
- [x] Greater
- [x] HardSigmoid
- [x] Hardmax
- [x] InstanceNormalization
- [x] LRN
- [x] LSTM
- [x] LeakyRelu
- [x] Less
- [x] Log
- [x] LogSoftmax
- [x] LpNormalization
- [ ] LpPool
- [x] Max
- [x] MaxPool
- [ ] MaxRoiPool
- [x] Mean
- [x] Min
- [x] Mul
- [x] Neg
- [x] Not
- [x] Or
- [x] PRelu
- [x] Pow
- [x] RNN
- [x] RandomNormal
- [x] RandomNormalLike
- [x] RandomUniform
- [x] RandomUniformLike
- [x] Reciprocal
- [x] ReduceL1
- [x] ReduceL2
- [x] ReduceLogSum
- [x] ReduceLogSumExp
- [x] ReduceMax
- [x] ReduceMean
- [x] ReduceMin
- [x] ReduceProd
- [x] ReduceSum
[...]

PR #881, opened 2018-05-03
add matmul shape inference

souptc:
Add shape inference function for matmul.


souptc:
Oh, not aware there is already a pr on this. If we already plan to merge that, I am ok to close this one.

bddppq:
@souptc Since #874 has earlier PR creation date, I'm going to close this one, sorry!
@houseroad @anderspapitto Probably create a list like in #426 for shape inference?

PR #876, opened 2018-05-03
version bump to 7

zasdfgbnm:
As mentioned in: https://github.com/onnx/onnx/pull/869#issuecomment-386167530

PR #884, opened 2018-05-04
Extend AveragePool to support average count include padding

bddppq:
counterpart for `count_include_pad` in PyTorch and `CUDNN_POOLING_AVERAGE_COUNT_INCLUDE_PADDING` cudnn.


bddppq:
This PR needs to wait for https://github.com/onnx/onnx/pull/876 which bumps the opset_version since last release

docs/Operators.md
```diff
@@ -564,18 +564,22 @@ This version of the operator has been available since version 1 of the default O
...
+This version of the operator has been available since version 7 of the default ONNX operator set.
+
+Other versions of this operator: <a href="Changelog.md#AveragePool-1">AveragePool-1</a>
 
 #### Attributes
 
 <dl>
 <dt><tt>auto_pad</tt> : string</dt>
 <dd>auto_pad must be either SAME_UPPER, SAME_LOWER or VALID. Where SAME_UPPER or SAME_LOWER mean pad the input so that the output size match the input.In case of odd number add the extra padding at the end for SAME_UPPER and at the beginning for SAME_LOWER. VALID mean no padding. DEPRECATION NOTE: auto_pad is only intended to support legacy uses, and for framework authors, one is explicitly encouraged to use explicit padding specified in the pads attribute.</dd>
+<dt><tt>count_include_pad</tt> : int</dt>
```
anirudhacharya: Shouldn't this specify that the default value = 0.

bddppq: @anirudhacharya It has been specified to default to zero in the schema, but our doc doesn't show that (we probably should, need someone working on correctly rendering AttributeProto in the markdown).

bddppq:
ping @linkerzhang @ebarsoum for reviewing schema changes

onnx/defs/nn/defs.cc
```diff
@@ -197,6 +197,18 @@ ONNX_OPERATOR_SCHEMA(AveragePool)
...
         "The output of each pooling window is divided by the number of elements exclude pad."));
 
+ONNX_OPERATOR_SCHEMA(AveragePool)
+    .SinceVersion(7)
+    .FillUsing(PoolOpSchemaGenerator(
+        "AveragePool",
+        "average",
+        "The output of each pooling window is divided by the number of elements (exclude pad when attribute count_include_pad is zero)."))
+    .Attr(
+        "count_include_pad",
```
ebarsoum: Should we add something similar to MaxPool? So if all elements are negatives out zeros.

bddppq: Yeah for the sake of leaving no ambiguity, let's add that as well (in a separate PR).

wschin:
Without this PR, it's still possible to compose this attribute. See [here](https://github.com/onnx/onnxmltools/blob/ccf3acb629d90d4a5634252951c2ea530302cc03/onnxmltools/convert/coreml/operator_converters/neural_network/Pool.py#L83). Anyway, I am glad to see that ONNX accepts composed operators (for optimization reasons).

PR #885, opened 2018-05-04
Tests for Gemm operator

anirudhacharya:


onnx/backend/test/case/node/gemm.py
```diff
@@ -0,0 +1,47 @@
...
+    @staticmethod
+    def export_transpose():
+        node = onnx.helper.make_node(
+            'Gemm',
+            inputs=['a', 'b', 'c'],
+            outputs=['y'],
+            alpha=0.5,
+            beta=0.5,
+            transA=1,
+            transB=1
```
bddppq: needs to set `broadcast=1`

onnx/backend/test/case/node/gemm.py
```diff
@@ -0,0 +1,48 @@
...
+            outputs=['y'],
+            alpha=0.5,
+            beta=0.5,
+            broadcast=1,
+            transA=1,
+            transB=1
+        )
+        a = np.random.ranf([6, 3]).astype(np.float32)
+        b = np.random.ranf([4, 6]).astype(np.float32)
+        c = np.random.ranf([1, 1]).astype(np.float32)
```
houseroad: This case is a little bit tricky on broadcasting... which is not aligned with our current broadcasting rule in other ops...

And in Gemm's doc there is no description on how to do the broadcast.

anirudhacharya: Yes, I just noticed that. I went by the mxnet standards on broadcasting -
```python
import numpy as np
import mxnet as mx
e=mx.nd.array(np.random.rand(4,5))
f=mx.nd.array(np.random.rand(1,1))
mx.nd.broadcast_add(e,f)
```

would it be possible to expand the definition of broadcast in ONNX. It would be useful.

houseroad: Yeah, I know, we are also about to move to numpy style broadcast. Let's first move the 1-element case aligned with numpy. https://github.com/onnx/onnx/pull/902

bddppq: The broadcast clarification/standardization in the schema probably should not block adding these tests?

houseroad: The only concern is that, in our current broadcast solution, this case is not allowed. We have to relax the 1-element case a little bit to support this case.

PR #886, opened 2018-05-04
Add conv transpose test cases

fumihwh:


houseroad:
@fumihwh could you add type info to the new cases?

bddppq:
@fumihwh @houseroad Just wanna double check, are the test data of opset 6 or opset 7?

fumihwh:
@bddppq I didn't get your point. I think ConvTranspose has only opset 1..

bddppq:
@fumihwh When exporting the node test cases, they are wrapped as a ModelProto (with its graph contains only one single node), I was asking the what's the opset version encoded in the new ModelProto added in this PR. I guess this doesn't matter as you said ConvTranspose hasn't been changed since version 1.

PR #887, opened 2018-05-04
Not outputing the opset_version in the notebook example

houseroad:
Two benefits:
1) No need to update the notebook when bump up the opset version
2) Show an example to create and set opsetid in model


houseroad:
@bddppq now i just don't output opset_version, the model is still the newest one.

zasdfgbnm:
I think check_model, load_model, and optimize_onnx also print opset version.

houseroad:
@zasdfgbnm but their model's opset version won't change when bumping up opset in the master.

bddppq:
CI failure unrelated

PR #888, opened 2018-05-04
add an ONNX_NO_WERROR flag to disable Werror

anderspapitto:


CMakeLists.txt
```diff
@@ -316,7 +316,9 @@ if (MSVC)
         STATIC_LIBRARY_FLAGS "${onnx_static_library_flags}")
 elseif(APPLE)
 else()
-  target_compile_options(onnx PRIVATE -Werror=sign-compare -Werror=conversion)
+  if (NOT ${ONNX_NO_WERROR})
```
bddppq: This is should declared as an option at the top of the cmake file as well (see ONNX_BUILD_PYTHON).

PR #890, opened 2018-05-04
type and shape inference for experimental ops

gramalingam:
inference for experimental ops


onnx/defs/experiments/defs.cc
```diff
@@ -162,7 +186,29 @@ ONNX_OPERATOR_SCHEMA(GivenTensorFill)
     .Attr("values", "", AttributeProto::FLOATS, OPTIONAL)
     .Attr("shape", "", AttributeProto::INTS, OPTIONAL)
     .Attr("input_as_shape", "", AttributeProto::INT, OPTIONAL)
-    .Attr("extra_shape", "", AttributeProto::INTS, OPTIONAL);
+    .Attr("extra_shape", "", AttributeProto::INTS, OPTIONAL)
```
anderspapitto: if this pattern shows up a third time it should probably be pulled out into a function

PR #892, opened 2018-05-04
promote identity op 

linkerzhang:
@wschin @guschmue


onnx/defs/tensor/defs.cc
```diff
@@ -780,3 +780,14 @@ For example A = [[1, 2], [3, 4]], B = [1, 2], tile(A, B) = [[1, 2, 1, 2], [3, 4,
...
+ONNX_OPERATOR_SCHEMA(Identity)
+.SetDoc("Identity operator")
+.Input(0, "input", "Input tensor", "T")
+.Output(
+	0,
+	"output",
+	"Tensor to copy input into.",
+	"T")
+	.TypeConstraint("T", OpSchema::all_tensor_types(),
+		"Constrain input and output types to all tensor types.");
```
gramalingam: add ".TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput)" (anyway we are probably going to see this in a merge conflict!)

linkerzhang: fixed.

PR #894, opened 2018-05-04
Enable non-static schema registration

jeffbloo:
This changes how operator schema are registered and used by ONNX. 

The motivation for this change is two-fold.  First, it transfers policy to the consumer for which operator schema and versions of those schema will be registered.  Second, it avoids a pattern of static object usage which can cause downstream compilation and performance issues.

Before this change, schema are registered using a macro (ONNX_OPERATOR_SCHEMA) which populates a static map.  This PR changes these macros to implement methods constructing the schema.  It also exposes functions that iterate across schema in different operator sets.

By default, the static registration map is populated with the operator sets on first use.  Callers can also disable this and populate the map manually.  Finally they can implement ISchemaRegistry, and request this be  which is used instead of the static map.

So that consumers may generate code with calls to the new registration functions, the macros are now always called from the ONNNX namespace.  The codebase was previously inconsistent on using either this or the global namespace.  The objects defined in this macro also use a naming convention, which includes the name, domain, and version which reflects the operator set within which it was first defined.  This affects the macro syntax and removes the need to use __COUNTER__ for uniqueness.

### DIFF
--- a/onnx/checker.cc
+++ b/onnx/checker.cc
@@ -1,5 +1,5 @@
-#include "onnx/checker.h"
 #include "onnx/defs/schema.h"
+#include "onnx/checker.h"
 #include "onnx/proto_utils.h"
 #include "onnx/string_utils.h"
 
@@ -271,8 +271,7 @@ void check_node(
     check_attribute(attr, ctx, lex_ctx);
   }
 
-  const auto* schema =
-      OpSchemaRegistry::Schema(node.op_type(), domain_version, node.domain());
+  const auto* schema = ctx.get_schema_registry()->GetSchema(node.op_type(), domain_version, node.domain());
   if (!schema) {
     fail_check(
         "No Schema registered for " + node.op_type() +
--- a/onnx/checker.cc
+++ b/onnx/checker.cc
@@ -1,5 +1,5 @@
-#include "onnx/defs/schema.h"
 #include "onnx/checker.h"
+#include "onnx/defs/schema.h"
 #include "onnx/proto_utils.h"
 #include "onnx/string_utils.h"
 
@@ -271,7 +271,8 @@ void check_node(
     check_attribute(attr, ctx, lex_ctx);
   }
 
-  const auto* schema = ctx.get_schema_registry()->GetSchema(node.op_type(), domain_version, node.domain());
+  const auto* schema = ctx.get_schema_registry()->GetSchema(
+      node.op_type(), domain_version, node.domain());
   if (!schema) {
     fail_check(
         "No Schema registered for " + node.op_type() +
--- a/onnx/common/ir_pb_converter.cc
+++ b/onnx/common/ir_pb_converter.cc
@@ -70,7 +70,7 @@ Tensor tensorProtoToTensor(const ONNX_NAMESPACE::TensorProto & tp) {
     break;
   }
   case ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED:
-    abort();
+    fail_convert("Unknown tensor data type");
   }
 
   // The only way to know if we should be using raw_data or
@@ -152,7 +152,7 @@ void convertAttribute(const ONNX_NAMESPACE::AttributeProto & ap, Node * n) {
     break;
   }
   case ONNX_NAMESPACE::AttributeProto_AttributeType_UNDEFINED:
-    abort();
+    fail_convert("Unknown tensor data type");
     break;
   }
 }
@@ -371,7 +371,7 @@ void encodeTensor(ONNX_NAMESPACE::TensorProto * p, const Tensor & tensor) {
     break;
   }
   case ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED:
-    abort();
+    fail_convert("Unknown tensor data type");
   }
   if (!tensor.raw().empty()) {
     p->set_raw_data(tensor.raw());
--- a/onnx/defs/experiments/defs.cc
+++ b/onnx/defs/experiments/defs.cc
@@ -2,33 +2,38 @@
 // Licensed under the MIT license.
 
 #include "onnx/defs/schema.h"
-using namespace ONNX_NAMESPACE;
+namespace ONNX_NAMESPACE
+{
+  using SupportType = OpSchema::SupportType;
 
 using SupportType = ONNX_NAMESPACE::OpSchema::SupportType;
 
-ONNX_OPERATOR_SCHEMA(Affine)
-    .SetSupportLevel(SupportType::EXPERIMENTAL)
-    .SetDoc(R"DOC(
+static const char * Affine_ver1_doc = R"DOC(
 Affine takes one input data (Tensor<T>) and produces one output data
 (Tensor<T>) where the affine function, y = alpha * x + beta,
 is applied to the tensor elementwise.
-)DOC")
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(Affine, 1, OpSchema()
+    .SetSupportLevel(SupportType::EXPERIMENTAL)
+    .SetDoc(Affine_ver1_doc)
     .Attr("alpha", "Value of alpha", AttributeProto::FLOAT, 1.0f)
     .Attr("beta" , "Value of beta", AttributeProto::FLOAT, 0.0f)
     .Input(0, "X", "1D input tensor", "T")
     .Output(0, "Y", "1D output tensor", "T")
     .TypeConstraint("T", { "tensor(float16)", "tensor(float)", "tensor(double)" },
         "Constrain input and output types to float tensors.")
-    .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput);
+    .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
-
-ONNX_OPERATOR_SCHEMA(ThresholdedRelu)
-    .SetSupportLevel(SupportType::EXPERIMENTAL)
-    .SetDoc(R"DOC(
+static const char * ThresholdedRelu_ver1_doc = R"DOC(
 ThresholdedRelu takes one input data (Tensor<T>) and produces one output data
 (Tensor<T>) where the rectified linear function, y = x for x > alpha, y = 0 otherwise,
 is applied to the tensor elementwise.
-)DOC")
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(ThresholdedRelu, 1, OpSchema()
+    .SetSupportLevel(SupportType::EXPERIMENTAL)
+    .SetDoc(ThresholdedRelu_ver1_doc)
     .Attr("alpha",
           "Threshold value",
           AttributeProto::FLOAT,
@@ -37,15 +42,17 @@ is applied to the tensor elementwise.
     .Output(0, "Y", "Output tensor", "T")
     .TypeConstraint("T", { "tensor(float16)", "tensor(float)", "tensor(double)" },
         "Constrain input and output types to float tensors.")
-    .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput);
+    .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
-ONNX_OPERATOR_SCHEMA(ScaledTanh)
-    .SetSupportLevel(SupportType::EXPERIMENTAL)
-    .SetDoc(R"DOC(
+static const char * ScaledTanh_ver1_doc = R"DOC(
 Calculates the scaled hyperbolic tangent of the given input tensor element-wise,
 alpha * tanh(beta * x). This operation can be done in an in-place fashion too,
 by providing the same input and output blobs.
-    )DOC")
+    )DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(ScaledTanh, 1, OpSchema()
+    .SetSupportLevel(SupportType::EXPERIMENTAL)
+    .SetDoc(ScaledTanh_ver1_doc)
     .Attr("alpha", "Scaling value", AttributeProto::FLOAT, OPTIONAL)
     .Attr("beta", "Scaling value", AttributeProto::FLOAT, OPTIONAL)
     .Input(0, "input", "Input tensor", "T")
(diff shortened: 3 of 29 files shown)


bddppq:
Please resolve merge conflict (and make sure the CI passing).

bddppq:
>Second, it avoids a pattern of static object usage which can cause downstream compilation and performance issues.

Could you elaborate?

onnx/defs/schema.h
```diff
@@ -655,14 +662,21 @@ class OpSchemaRegistry final {
...
 };
 
-#define ONNX_OPERATOR_SCHEMA(name) \
-  ONNX_OPERATOR_SCHEMA_UNIQ_HELPER(__COUNTER__, name)
-#define ONNX_OPERATOR_SCHEMA_UNIQ_HELPER(Counter, name) \
-  ONNX_OPERATOR_SCHEMA_UNIQ(Counter, name)
-#define ONNX_OPERATOR_SCHEMA_UNIQ(Counter, name)                 \
+// ONNX_OPERATOR_SCHEMA is implemented either with static registration or by exposing
+// a factory method.  
+#ifndef ONNX_NON_STATIC_SCHEMA_INITIALIZATION
```
linkerzhang: I'm nervous about adding this kind of option, similar with the one we added for ONNX_ML. it makes the code diverge and different partners focus on each option. It's generally hard to maintain. For such divergence, either all partners agree on same direction, or put the specific stuff in local code base.

shschaefer: I agree that having different patterns is not ideal.  However, not all vendors may be able to leverage the code in its current state, but the code contains a lot of necessary information.  To answer the question above @bddppq, control of the timing of initialization is important to manage potential startup cost for the compilation unit.  As we are looking at this, the code can be brought into many different processes, some where deferring this cost is important.

shschaefer: For non-framework scenarios, when this code is loaded as part of an ONNX library, it can be consumed by numerous processes.  We do a lot of work to eliminate this overhead, delay initialization until the library is used, etc...  The dynamic initializer makes this code hard to consume.  Not sure if providing benchmarks changes this.

bddppq: @shschaefer I'm asking benchmark numbers mainly because I think the overhead of these static initialization should be negligible.

onnx/defs/schema.h
```diff
@@ -655,14 +662,21 @@ class OpSchemaRegistry final {
...
   static ONNX_NAMESPACE::OpSchemaRegistry::OpSchemaRegisterOnce( \
-      op_schema_register_once##name##Counter) =                  \
-      OpSchema(#name, __FILE__, __LINE__)
+      op_schema_register_once_##name##_##domain##_ver##ver) = impl\
+      .SetName(#name).SetDomain(domain##_NAME).SinceVersion(ver).SetLocation(__FILE__, __LINE__)
+#else
+template <typename T> OpSchema GetOpSchema();
+#define ONNX_OPERATOR_SCHEMA(name, domain, ver, impl) \
+    class name##_##domain##_ver##ver##_OpSchemaId{}; \
+    template<> OpSchema ONNX_NAMESPACE::GetOpSchema< name##_##domain##_ver##ver##_OpSchemaId>(){ \
```
bddppq: Actually could you elaborate how does it work? Where is this `GetOpSchema` declared and how is user code supposed to call it?

onnx/defs/schema.h
jeffbloo: I measured that registering schema consumes 4 milliseconds and increases memory consumption by 1.1 MB.  This is very non-negligible for a single component, especially if it may be loaded into many processes.

onnx/defs/schema.h
jeffbloo: GetOpSchema is declared slightly above this.  Macro call sites define its specializations.  

Based on the feedback here, I'm preparing a change to remove the original mode of registration during static initialization.  This will include a new file with related declarations for each schema, and functions which retrieve and/or register the schema of specific operator sets.

bddppq:
@jeffbloo @linkerzhang @shschaefer @dzhulgakov How about we move the schema instantiation from global static variables to a function static variable in the `map()` function? i.e. something like:
```
class SchemasRegisterer {
  SchemasRegisterer() {
   // AUTOGENERATED BEGIN
    _map.insert(std::make_pair("Aaa", OperatorSchema(Aaa)....));
    ...
   _map.insert(std::make_pair("Zzz", OperatorSchema(Zzz)....));
   // AUTOGENERATED END
  }
}

Map* map() {
  static auto SchemasRegisterer{};
  return _map
}
```
This way we only do the initialization at the first usage time, and there is no need for an Init and it's thread safe.

To achieve this we will need to move the source of truth of the schemas to say yaml files (instead .cc files) and then have a script to auto-generate the chunk between `AUTOGENERATED BEGIN` and `AUTOGENERATED END` above.

jeffbloo:
@bddppq - It's possible to populate the map by using the same couple code lines as added to cpp2py_export.cc.  Is it your goal to make this change non-breaking, so that updates are not required in any callers?

Another downside to that change would be coupling the policy of whether to use OpSchemaRegistry with the policy of which operator sets and versions will be used, in which registry.  

This could also make it less clear whether other code in the ONNX library should rely on this map.  For a similar reason, I was considering removing OpSchemaRegistry's internal statics altogether, and making callers create their own static instances of OpSchemaRegistry and share them as needed.  This is more disruptive to calling code though.

I'd prefer to keep this simple and define operator sets directly for now.  Code generation could be a more involved change if designed holistically (considering the most efficient format to generate, how to best handle inferencing callbacks, etc.).   The iteration routines in the existing PR also provide more flexibility for the library consumer, without requiring them to use a code-gen step.

jeffbloo:
The latest update works around a bug in older GCC compilers that the build hit.  Raw strings inside macros were not handled correctly.

I also removed usages of std::abort to avoid affecting terminating a whole process, and added a conditional to allow large doc strings to be compiled out to save memory.

onnx/defs/experiments/defs.cc
```diff
@@ -212,6 +207,11 @@ ONNX_OPERATOR_SCHEMA(GivenTensorFill, ONNX_DOMAIN, 1, OpSchema()
         }
     }));
 
+const char * Scale_ver1_doc = R"DOC(
```
bddppq: What's the reason of moving this string outside? This makes the schema (in c++ files) less readable.

onnx/defs/schema.h
```diff
@@ -659,14 +704,27 @@ class OpSchemaRegistry final {
...
+};
+
+// Forward declaration for the non-specialized GetOpSchema method.  This enforces a consistent
+// signature on functions that query individual schema, which are defined as specializations
+// of this function.
+template <typename T> OpSchema GetOpSchema();
+
+// Defines specialization of GetOpSchema for a class whose name is determined based on 
+// a convention using name, domain, and version.
+#define ONNX_OPERATOR_SCHEMA(name, domain, ver, impl) \
```
bddppq: most schemas are of the onnx core domain. can you change this to a helper macro and leave `ONNX_OPERATOR_SCHEMA` dedicated to the onnx core domain? For ONNX_ML I think `ONNX_ML_OPERATOR_SCHEMA` can be used.

onnx/defs/schema.cc
```diff
@@ -312,62 +331,59 @@ OpSchema& OpSchema::Attr(
...
-      std::cerr << "Attribute specification type mismatch.";        \
-      abort();                                                      \
-    }                                                               \
-    AttributeProto a;                                               \
-    a.set_name(name);                                               \
-    *(a.mutable_##field()) = default_value;                         \
-    a.set_type(attr_type);                                          \
-    Attr(Attribute(std::move(name), std::move(description), a));    \
-    return *this;                                                   \
+#define ATTR_SETTER_WITH_SINGLE_COMPLEXVALUE(type, field, attrtype)       \
```
bddppq: Could you not clang-format the unchanged part? It's really hard to review. :-)

onnx/defs/schema.cc
```diff
@@ -5,6 +5,8 @@
 #include <stdexcept>
 #include <unordered_set>
 #include "onnx/checker.h"
+#include "onnx/defs/operator_sets.h"
+#include "onnx/defs/operator_sets-ml.h"
```
bddppq: Need to guard this include with ONNX_ML for non ONNX_ML builds

onnx/defs/operator_sets.h
```diff
@@ -0,0 +1,385 @@
+// Copyright (c) Facebook Inc. and Microsoft Corporation.
```
bddppq: So this file should be autogenerated, it's so easy to forget to update when adding new schemas and also unfriendly for new contributors.

onnx/defs/schema.cc
```diff
@@ -575,11 +591,32 @@ std::ostream& operator<<(std::ostream& out, const OpSchema& schema) {
...
+    public:
+      SchemasRegisterer() {
+        RegisterOnnxOperatorSetSchema();
+#ifdef ONNX_ML
+        RegisterOnnxMLOperatorSetSchema();
+#endif    
+      }
+  };
+
+  #ifndef ONNX_DISABLE_STATIC_REGISTRATION
```
bddppq: Can we remove this variant? It's not happening at initialization time already, what's the concern?

onnx/defs/schema.h
```diff
@@ -209,10 +210,25 @@ class OpSchema final {
   OpSchema& SetSupportLevel(SupportType supportType);
 
   // Functions to do documentation for the operator schema.
-  OpSchema& SetDoc(std::string doc);
+  // This may be disabled to save memory.
+  OpSchema& SetDoc(const char* doc) {
+#ifndef ONNX_NO_DOC_STRINGS
```
bddppq: Can you separate the doc string stripping change to anther PR? I prefer not to let these two discussions don't slow down each other. Also according to our previous discussion in #623, removing doc strings really doesn't save you too much (~50kb **uncompressed** size, <10kb **compressed**)

bddppq:
Also what's the strategy for users adding their own schemas? As mentioned before, moving the schemas instantiation into function static variable makes the registration closed.

onnx/defs/schema.h
jeffbloo: I can do that

onnx/defs/schema.cc
jeffbloo: OK

onnx/defs/schema.cc
jeffbloo: The headers had internal conditionals, but I can add them here.

onnx/defs/schema.h
jeffbloo: I see a similar gain - 56 KB.  I also have to make the strings static and avoid consuming onnx.lib with whole-archive (which is now possible with this change).  I think this is worthwhile for such an isolated change.  Savings like this one add up.

onnx/defs/operator_sets.h
jeffbloo: In either case, the operators won't be available or get generated into documentation, so I don't think this will be missed.  

Not requiring code-gen steps is simplifying in itself.  While I can imagine other benefits if we combine code-gen with other refactoring, I don't want to block this PR on that larger project.

onnx/defs/schema.cc
jeffbloo: This was intended to allow the static registry to be used by a caller which wants to constrain it to specific operator sets.  For example, to deprecate an older operator set or withhold others.

onnx/defs/schema.h
jeffbloo: This applies to your question below about custom schema.  They may also implement ISchemaRegistry.

jeffbloo:
@bddppq - As to binary size, this was indeed a slight increase.  Measured on our compiler, about 85K larger while optimizing for speed, and 55K for size.  This is without using the new macro to remove doc strings.

A lot of bloat is related to constructing STL objects before each registration call.  I'm fixing this by simply wrapping the OpSchema methods consuming STL types.  This will make the size 175K smaller than the master branch, and another 56K smaller if removing doc strings.

onnx/defs/schema.cc
jeffbloo: This applies to your question below about custom schema. They may also implement ISchemaRegistry.

onnx/defs/operator_sets.h
bddppq: This will be missed. We can't rely on reviewers to remember how many docs changes they need to see so the changes are legit, the only viable way to guard such boilerplate changes is adding a check in the CI. Also think about new coming/occasional contributors, it's impossible for them to remember.

onnx/defs/schema.h
bddppq: What's the total size in your measurement? If it's the same as in the previous PR, the total size is around 8MB, which means 50 KB is really not worth adding this complexity into the code base (note as you have found this not only needs to add those macros, but also require moving couple strings and changing some function signature in the OpSchema class, and in some code you will need to add guards preventing accessing the doc strings).
In any case, I feel this really should be a separate discussion, could you split it out into a separate PR.

onnx/defs/schema.cc
bddppq: Could you elaborate what's the purpose of constrain the registration to a specific operator sets? If it's for overriding an operator definition in the onnx core I'm pretty against it.

onnx/defs/experiments/defs.cc
jeffbloo: This is the workaround I mentioned for a GCC bug.  Raw strings in macros were busted.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57824

onnx/defs/schema.h
jeffbloo: The baseline size was around 2.5 MB.  In the context of a larger product this isn't as relevant.  Components of different sizes sharing a common budget should be equally judicious with size deltas, and this is an easy win.   It also reduces heap usage, which is multiplied across every single process loading the binary.

Doc strings are optional, so guards aren't necessary.  The string movements were necessary for different reasons - the GCC bug workaround.  The signature changes also also needed for other reasons - minimizing static STL string initialization and avoiding instruction bloat in the macro-defined functions.

onnx/defs/schema.cc
jeffbloo: No problem.  It's not actually intended for overriding ones that already exist.  One goal is for a caller to have more control over the contract it exposes to its own customers.  This naturally includes the operators that it supports. 

Another goal is to enable callers to handle certain operator sets outside the core library, based in part on its strategy for deploying updates.

onnx/defs/operator_sets.h
jeffbloo: Our thinking was that exposing operators which were never tested was not as important, but I understand wanting to make this easier to discover.

I'll see if I can cleanly add check in debug mode that the number of registration macros usages matches the total of schema exposed by the iteration methods.  Does that sound reasonable?

jeffbloo: I've made that change.  This also fixes a pre-existing bug in assert implementations which caused them to always pass.

onnx/checker.cc
```diff
@@ -465,7 +464,7 @@ void check_model(const ModelProto& model) {
           "model with IR version >= 3 must specify opset_import for ONNX");
   } else {
     if (opset_imports.empty())
-      opset_imports[ONNX_DOMAIN] = 1;
+      opset_imports[ONNX_DOMAIN_NAME] = 1;
```
bddppq: Could you preserve ONNX_DOMAIN as well?

onnx/common/ir_pb_converter.cc
```diff
@@ -70,7 +70,7 @@ Tensor tensorProtoToTensor(const ONNX_NAMESPACE::TensorProto & tp) {
     break;
   }
   case ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED:
-    abort();
+    throw std::runtime_error("Unknown tensor data type");
```
bddppq: In other places (e.g. checker and shape inference) we define a custom exception class inherit from std::runtime_error and macros (e.g. `fail_check`, `fail_shape_inference`).

PR #897, opened 2018-05-06
add multinomial op to ONNX

linkerzhang:


onnx/defs/generator/defs.cc
```diff
@@ -225,3 +225,56 @@ TensorProto message, and be valid as an output type.
...
+        {"tensor(float16)", "tensor(float)", "tensor(double)"},
+        "Constrain input types to float tensors.")
+    .TypeConstraint(
+        "T2",
+        {"tensor(int32)", "tensor(int64)"},
+        "Constrain output types to integral tensors.")
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+      auto output_type = ctx.getOutputType(0);
+      auto input_type = ctx.getInputType(0);
+      if (TypeProto::kTensorType != output_type->value_case() ||
```
bddppq: input_type and output_type can be `nullptr`

onnx/defs/generator/defs.cc
```diff
@@ -225,3 +225,56 @@ TensorProto message, and be valid as an output type.
...
+          TypeProto::kTensorType != input_type->value_case()) {
+        return;
+      }
+      auto sample_size = ctx.getAttribute("sample_size");
+      if (nullptr == sample_size || // attribute not present
```
bddppq: This can't be true for a valid Operator. I suggest shape inference code don't do validation and always operate on the assumption that the model file has passed the checker. (At least here don't silently return, it should raise some errors or similar) @anderspapitto @gramalingam What's the error reporting mechanism in shape inference?

onnx/defs/generator/defs.cc
```diff
@@ -225,3 +225,56 @@ TensorProto message, and be valid as an output type.
...
+          !sample_size->has_type() || // invalid attribute : has no type
+          sample_size->type() !=
+              AttributeProto_AttributeType_INT) { // invalid attribute type
+        return;
+      }
+      auto shape = output_type->mutable_tensor_type()->mutable_shape();
+      auto dim = shape->add_dim();
+      *dim = input_type->tensor_type().shape().dim(static_cast<int>(0));
```
bddppq: I'm curious does compiler complain if you don't static_cast<int>?

onnx/defs/generator/defs.cc
```diff
@@ -225,3 +225,56 @@ TensorProto message, and be valid as an output type.
...
+        "output",
+        "Output tensor with shape [batch_size, sample_size], where sample_size is the number of times to sample. ",
+        "T2")
+    .TypeConstraint(
+        "T1",
+        {"tensor(float16)", "tensor(float)", "tensor(double)"},
+        "Constrain input types to float tensors.")
+    .TypeConstraint(
+        "T2",
+        {"tensor(int32)", "tensor(int64)"},
```
bddppq: How is the backend supposed to know which output type it should take?

onnx/defs/generator/defs.cc
```diff
@@ -225,3 +225,56 @@ TensorProto message, and be valid as an output type.
...
+        static_cast<int64_t>(0))
+    .Attr(
+        "seed",
+        "(Optional) Seed to the random generator, if not specified we will auto generate one.",
+        AttributeProto::FLOAT,
+        OPTIONAL)
+    .Input(
+        0,
+        "input",
+        "Input tensor with shape [batch_size, class_size], where class_size is the number of all possible outcomes.",
```
bddppq: Please clarify the meaning the values in this input tensor.

onnx/defs/generator/defs.cc
linkerzhang: About error handling in shape inference, I'd suggest to add "Status" as return value. @anderspapitto  and @gramalingam  let me know whether you're OK on it or not please. The "Status" class was added recently. I had ever given @anderspapitto this proposal (using status as return value) when the interface was checked in.

onnx/defs/generator/defs.cc
linkerzhang: will check that :).

onnx/defs/generator/defs.cc
gramalingam: I think returning a Status code is reasonable. While I agree that it is reasonable to assume that the model file has passed the checker before inference is invoked, it is still a good idea to safeguard the inference code against null-dereference etc. (E.g., this ensures that even if the op signature is changed to make an attribute optional, we don't fail with a null dereference.) Shape inference is (currently) viewed as optional (for optimization): there will be models where shape inference won't successfully infer shapes. For this reason, returning silently is a reasonable compromise for shape inference (in such edge cases). At least, that's my personal opinion.

gramalingam: You can use getAttribute(ctx, "sample_size", -1) and return if the value is negative.

onnx/defs/generator/defs.cc
gramalingam: Yes, this is not clear to me either. If we want both, we probably should add a "dtype" attribute that specifies the output type

onnx/defs/generator/defs.cc
anderspapitto: @bddppq do we have a list or documentation somewhere of what guarantees the checker provides about the model?

onnx/defs/generator/defs.cc
anderspapitto: add some test cases?

onnx/defs/generator/defs.cc
bddppq: @anderspapitto Nope, but it's a good idea to document this.

onnx/defs/generator/defs.cc
houseroad: Shape inference test will also be helpful here.

onnx/defs/generator/defs.cc
wchao1115: Fixed.

onnx/defs/generator/defs.cc
wchao1115: This is fixed in the new revision.

onnx/defs/generator/defs.cc
wchao1115: As for the backend test case, I've looked at it and realized that this is a generator op. Although I think one could implement an AB testing comparing the result of a random function against another (say std::mt19937), it would be too stringent to assume that all backend implementation must conform to a specific algorithm.

docs/Operators.md
```diff
@@ -5058,6 +5059,50 @@ expect(node, inputs=[x, y], outputs=[z],
...
 
 
+### <a name="Multinomial"></a><a name="multinomial">**Multinomial**</a>
+
+  Generate a tensor of samples from a multinomial distribution according to the probabilities
+  of each of the possible outcomes.
+
+#### Version
+
+This version of the operator has been available since version 1 of the default ONNX operator set.
```
anirudhacharya: this is a new operator, It has not been available since op_set version 1.

onnx/defs/generator/defs.cc
```diff
@@ -1,227 +1,282 @@
...
+          propagateElemTypeFromAttributeToOutput(ctx, "dtype", 0);
+        else
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+        if (!hasNInputShapes(ctx, 1)) {
+          return;
+        }
+        propagateShapeFromInputToOutput(ctx, 0, 0);
+    });
+
+ONNX_OPERATOR_SCHEMA(Multinomial)
```
anirudhacharya: why isn't this experimental?

onnx/defs/generator/defs.cc
```diff
@@ -1,227 +1,282 @@
...
+    .SetDoc(R"DOC(
+Generate a tensor of samples from a multinomial distribution according to the probabilities
+of each of the possible outcomes.
+)DOC")
+    .Attr(
+        "sample_size",
```
anirudhacharya: sample size should default to 1

wchao1115: Fixed default sample size to 1.

bddppq:
@wchao1115 There are windows line ending in onnx/defs/generator/defs.cc.

onnx/defs/generator/defs.cc
```diff
@@ -1,227 +1,292 @@
...
+    .Attr(
+        "seed",
+        "(Optional) Seed to the random generator, if not specified we will auto generate one.",
+        AttributeProto::FLOAT,
+        OPTIONAL)
+    .Attr(
+        "dtype",
+        "(Optional) The data type for the elements of the output tensor, if not specified, we will use int32.",
+        AttributeProto::INT,
+        OPTIONAL)
```
bddppq: The default value can be specified in the schema here.

onnx/defs/generator/defs.cc
```diff
@@ -1,227 +1,292 @@
...
+        "Constrain output types to integral tensors.")
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        auto output_type = ctx.getOutputType(0);
+        auto input_type = ctx.getInputType(0);
+        if (input_type == nullptr || output_type == nullptr ||
+            TypeProto::kTensorType != output_type->value_case() ||
+            TypeProto::kTensorType != input_type->value_case()) {
+            return;
+        }
+        auto sample_size = ctx.getAttribute("sample_size");
```
bddppq: `sample_size` is an optional attribute and so here you can possibly get a `nullptr`

onnx/defs/generator/defs.cc
```diff
@@ -1,227 +1,292 @@
...
+        OPTIONAL)
+    .Attr(
+        "dtype",
+        "(Optional) The data type for the elements of the output tensor, if not specified, we will use int32.",
+        AttributeProto::INT,
+        OPTIONAL)
+    .Input(
+        0,
+        "input",
+        "Input tensor with shape [batch_size, class_size], where class_size is the number of all possible outcomes. Each value along the axis zero represents the unnormalized log-probability of each corresponding outcome in a batch.", "T1")
```
bddppq: Why log-probability here (instead of just the plain probability)?

docs/Operators.md
wchao1115: @linkerzhang produces this. I think simply by running gen_doc. Ke, can you look into this feedback? I somehow had problem generating docs in my machine with gen_doc.py. It decimated the doc as a result but without an error.

onnx/defs/generator/defs.cc
wchao1115: Why shouldn't it be real if there is a clear use case for it in a model? This operator is requested by Unity. @prasanthpul can you help facilitate this conversation?

prasanthpul: @anirudhacharya generally ops are added as experimental because their use or value is uncertain. for example it could be an op that is only used in 1 framework or an op that potentially could be decomposed into other ops without a perf hit. In this case, the op is supported by multiple frameworks such as TF, PyTorch, mxnet, etc. and can be considered a primitive. So the proper course is to add it as a core op - similar to how the trig ops were added recently.

onnx/defs/generator/defs.cc
linkerzhang: fixed.

onnx/defs/generator/defs.cc
linkerzhang: fixed.

docs/Operators.md
linkerzhang: good catch. fixed.

onnx/defs/generator/defs.cc
bddppq: @wchao1115 It is optional, you currently has set a default value for it, which defines how should the backend should interpret when it's missing. My previous comment was for required attribute.

Issue #696, opened 2018-04-03
Is it ok and better to get rid of using static variables in onnx?


linkerzhang:
Currently, op registration happens during static initialization, whose order is undefined. To support adding operators for customized domain, dynamically adding a domain should happen before customized op registration. There may be two options,
1. Move all operator registration into a static function to ensure domain customization will happen before op registration.
2. Not use static variables in onnx so that backends may control the initialization in right order.

Thoughts? @bddppq @houseroad @gramalingam  @ebarsoum

bddppq:
I think we just need to change the [OpSchemaRegisterOnce ctor](https://github.com/onnx/onnx/blob/2af45df08182f26c2279c8177e3f736fc66e77ae/onnx/defs/schema.h#L533) to allow dynamically register domain.

A related question: why do we need to hardcode min&max version for each domain in `DomainToVersionRange`? I think we can instead programmatically figure out these versions in the registry?

houseroad:
@linkerzhang @bddppq 
I vote for deterministic initialization order. This is quite important. Without it, we may have unexpected errors when using different compilers.

linkerzhang:
Let's use this issue to track this work item to get rid of using static variables to initialize the op schema map.

bddppq:
@linkerzhang sorry I don't get why we need get rid of static variables?

bddppq:
@houseroad what's the error you are referring to?

houseroad:
@linkerzhang I have synced with @anderspapitto and @bddppq offline. We would like to make sure the initialization order, and at the same, also please keep the interface user-friendly.

bddppq:
@houseroad  @linkerzhang  Sorry why do we need to make sure the order of the schemas registration?

houseroad:
What I can think of is that at the compile time we can make sure all the schemas are registered in the right order. So when register some different ops (at compilation time), we can have some legit assumption.

On the other hand, because schema is so important, little change may break a lot of things, @linkerzhang I think @bddppq's concern makes sense. We don't want to go too far and introduce unnecessary changes/complexity.

linkerzhang:
Great! Let me see whether I may come out of a solution of not impacting the current API (or at least not big change) as well as supporting non-static registration.

PR #898, opened 2018-05-06
function registration refactor.

linkerzhang:
More APIs may be added into FunctionBuilderRegistry for different accessing patterns. I'll change OpSchemaRegistry following this pattern to remove those "abort()" function call.


bddppq:
Can we simply change from calling abort to throwing exception? This change will require users to call "Init" for each registry instance which looks to me very cumbersome (especially right now it's not thread safe)

bddppq:
@onnxbot retest this please

bddppq:
@linkerzhang No #894 will make the code super hard to maintain (not to mention playing with macros is really not fun).
So errors in schema registration phase should rarely happens (for op schema it only happens when we accidentally declared duplicated schemas), and when error happens it's an undefined behavior in onnx. If we can not afford abort during this phase, how about let's just log to stderr and keep going?

linkerzhang:
#894 could be discussed in-place in the PR.

@bddppq  having an init() function makes the order of doing initialization clearly, which was kind of discussed and agreed in #696 . logging the error and keep going is good for other apps except windows, I think :), although I agree with you that rare error cases actually, just in case there's one.

bddppq:
Another choice would be having separate registries for onnx core schemas and user defined schemas, this can hugely improve the guarantee of no duplications.

linkerzhang:
@shschaefer may share more thoughts on having a deterministic order of initialization please. Thank you!

linkerzhang:
btw, @bddppq  to have all the function proto cached, it's also good to have an init function :). Otherwise, the initialzation logic will be put in the getter covered by a "if" statement, which is not good.

bddppq:
@linkerzhang Well FunctionProto instantiation can be done the same way as op schema registration, in that case there will no need of caching or Init. Adding Init is because the users code will end up doing onnx::init everywhere it imports onnx. And as I have mentioned above, this is an api breaking change, all existing onnx integrations need to be updated to insert the init call.
If you really insist to make the schema's initialization order predictable, we can archive it by putting the schemas' source of truth into yaml files and then write a script to generate one single .cc file that has the c++ part of all the op schemas in desired order, the instantiation order of static variables in one single compilation unit is deterministic.

PR #902, opened 2018-05-07
Enhance the 1-element broadcast case

houseroad:
Since we are about to align to numpy broadcasting, let's move forward. This case will enable 1-element case aligned with numpy broadcasting.


houseroad:
ongoing PR is here for the record: https://github.com/onnx/onnx/pull/907

houseroad:
@anirudhacharya @ebarsoum @linkerzhang ping for review :-)

docs/Operators.md
```diff
@@ -177,14 +177,16 @@ expect(node, inputs=[x], outputs=[y],
...
+  tensor with rank equal to or smaller than the first tensor), or having its
+  shape as a contiguous subset of the first tensor's shape. The starting of the
+  mutually equal shape is specified by the argument "axis", and if it is not set,
+  suffix matching is assumed. 1-dim expansion doesn't work yet.
   
   For example, the following tensor shapes are supported (with broadcast=1):
   
-    shape(A) = (2, 3, 4, 5), shape(B) = (,), i.e. B is a scalar
+    shape(A) = (2, 3, 4, 5), shape(B) = (,), i.e. B is a scalar tensor
+    shape(A) = (2, 3, 4, 5), shape(B) = (1, 1), i.e. B is an 1-element tensor
```
anirudhacharya: This change should be shown in the documentation of the following operators too, which also have broadcast attribute - AND, Equal, Xor, Greater, Lesser, and GEMM.

bddppq: Repeatedly putting this chunk in each relevant op seems cumbersome. Shall we put the description of our broadcasting rule as a separate doc, and then refer to it in each op? This can also help keeping the broadcast semantics consistent across all relevant ops.

anirudhacharya: sounds good.

houseroad: Yeah, I did so in the other PR.

houseroad: @anirudhacharya do you want to approve this PR? We will solve the repeated broadcasting text in the other PR #907, otherwise we cannot check in the Gemm test cases.

anirudhacharya: @houseroad approved.

PR #903, opened 2018-05-07
Tests for LRN operator

anirudhacharya:
- [x] Need to add a test case where the window size is greater than 1.
- [x] Clarify the definition of the operator. Is the normalization happing within the channel or across the channel. How to ascertain the window of the local region?


houseroad:
@anirudhacharya friendly ping, we have updated LRN's spec, do you mind updating the pr accordingly?

anirudhacharya:
@houseroad will do.

anirudhacharya:
@houseroad ping for review

onnx/backend/test/case/node/lrn.py
```diff
@@ -0,0 +1,65 @@
...
+            'LRN',
+            inputs=['x'],
+            outputs=['y'],
+            alpha=alpha,
+            beta=beta,
+            bias=bias,
+            size=nsize
+        )
+        x = np.random.rand(5, 5, 5, 5).astype(np.float32) * 1000
+        square_sum = np.zeros((5,5,5,5))
```
houseroad: default is float64, please convert it to float32.

onnx/backend/test/case/node/lrn.py
```diff
@@ -0,0 +1,65 @@
...
+        bias = 1.0
+        nsize = 3
+        node = onnx.helper.make_node(
+            'LRN',
+            inputs=['x'],
+            outputs=['y'],
+            size=3
+        )
+        x = np.random.rand(5, 5, 5, 5).astype(np.float32) * 1000
+        square_sum = np.zeros((5,5,5,5))
```
houseroad: default is float64, please convert it to float32.

houseroad:
Lint the code please: https://travis-ci.org/onnx/onnx/jobs/385386568

anirudhacharya:
@houseroad ping

gramalingam:
@anirudhacharya : a question: the ONNX spec for LRN mentions the upper-limit of the local region as min(C - 1, c + ceil((size - 1) / 2) - 1), while the test case at https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/lrn.py uses min(4, c + int(math.ceil((nsize - 1) / 2)) + 1) … that is, + 1 instead of - 1. Can you please clarify?

gramalingam:
Yes, the spec seems to have an erroneous "-1" … otherwise, even the description of "size" as "The number of channels to sum over" does not match. It would be great if you can update the spec. Thanks.

Issue #905, opened 2018-05-08
Bump up the opset_version to 7


houseroad:
- [x] https://github.com/onnx/onnx/pull/876
- [x] update the pytorch converted models
- [x] update the model zoo files

houseroad:
with https://github.com/onnx/onnx/pull/907/, we will use numpy style broadcasting

PR #907, opened 2018-05-08
Align to Numpy broadcasting

houseroad:
# Change the specs
- [x] Add
- [x] Div
- [x] Mul
- [x] Pow
- [x] Sub
- [x] And
- [x] Or
- [x] Xor
- [x] Equal
- [x] Greater
- [x] Less
- [x] Gemm
- [x] Prelu

# Update the Node test cases
- [x] Add
- [x] Div
- [x] Mul
- [x] Pow
- [x] Sub
- [x] And
- [x] Or
- [x] Xor
- [x] Equal
- [x] Greater
- [x] Less
- [x] Gemm
- [x] Prelu

# Fix ONNX optimizer
- [x] Fix Fuse Add into Conv pass

# Update the converted test cases (Optional)
Will do it later. Keep tracking in https://github.com/onnx/onnx/issues/905

# Test
- [x] Make sure Caffe2 CI is green


houseroad:
WIP ==> work in progress, don't worry, we will cover all these cases :-)

snnn:
Please also add node test cases to reflect this change.

houseroad:
@snnn sure, will and have to do that, otherwise, CI won't be green. :-)

docs/Broadcasting.md
```diff
@@ -0,0 +1,66 @@
...
+- [Xor](Operators.md#Xor)
+- [Equal](Operators.md#Equal)
+- [Greater](Operators.md#Greater)
+- [Less](Operators.md#Less)
+
+
+## Unidirectional Broadcasting
+
+In ONNX, tensor B is unidirectional broadcastable to tensor A
+if one of the followig is true:
```
linkerzhang: following

docs/Broadcasting.md
```diff
@@ -0,0 +1,66 @@
...
+respecitively in the following sections.
+
+
+## Multidirectional Broadcasting
+
+In ONNX, a set of tensors are multidirectional broadcastable to the same shape
+if one of the followig is true:
+- The tensors all have exactly the same shape.
+- The tensors all have the same number of dimensions and the length of
+each dimensions is either a common length or 1.
```
linkerzhang: In numpy, if the dimension value is not 1, then they should be the same, right? (common length means same length here?).

onnx/defs/math/defs.cc
```diff
@@ -8,62 +8,31 @@ using namespace ONNX_NAMESPACE;
...
-        "If broadcasting is disabled it should be of the same size.",
+        "Second operand.",
         "T");
-    schema.Output(0, "C", "Result, has same dimensions and type as A", "T");
+    schema.Output(0, "C", "Result, has same element type as two inputs", "T");
     schema.TypeConstraint(
         "T",
         OpSchema::high_precision_numeric_types(),
         "Constrain input and output types to high-precision numeric tensors.");
-    schema.TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput);
```
linkerzhang: @gramalingam @anderspapitto  please be aware of that type and shape inference function for math ops need to be updated. it's removed in this PR :).

onnx/defs/math/defs.cc
```diff
@@ -403,13 +362,13 @@ PRelu takes input data (Tensor<T>) and slope tensor as input, and produces one
...
-)DOC")
+)DOC" + GenerateBroadcastingDocUni("tensor slope", "input tensor X"))
     .Input(0, "X", "Input tensor", "T")
     .Input(
         1,
         "slope",
-        "Slope tensor. If `Slope` is of size 1, the value is shared"
-        "across different channels",
+        "Slope tensor. The shape of slope can be smaller then first input X; "
+        "if so, its shape must be broadcastable to X",
```
linkerzhang: unidirectional broadcastable?

onnx/defs/schema.h
```diff
@@ -669,4 +669,16 @@ class OpSchemaRegistry final {
 // Helper function
 size_t ReplaceAll(std::string& s, const char* from, const char* to);
 
+inline std::string GenerateBroadcastingDocMul() {
+  return "This operator supports **multidirectional (i.e., Numpy-style) broadcasting**;"
+         " for more details please check [the doc](Broadcasting.md).";
+}
+
+inline std::string GenerateBroadcastingDocUni(const char* from, const char* to) {
+  std::string ret = "This operator supports **multidirectional broadcasting** (";
```
linkerzhang: unidirectional broadcasting, I think.

linkerzhang:
Thank you very much! @houseroad for pushing this forward! Overall it looks good to me. Some minor editing comments was given.

docs/Broadcasting.md
```diff
@@ -0,0 +1,66 @@
...
+
+In ONNX, tensor B is unidirectional broadcastable to tensor A
+if one of the followig is true:
+- Tensor A and B both have exactly the same shape.
+- Tensor A and B all have the same number of dimensions and the length of
+each dimensions is either a common length or B's length is 1.
+- Tensor B has too few dimensions, and B can have its shapes prepended
+with a dimension of length 1 to satisfy property 2.
+
+When unidirectional broadcasting happens, the output's shape is the same as the larger
```
anderspapitto: specifically, the output shape is the same as the shape of A, right? If B is bigger than you can't do unidirectional broadcasting

onnx/defs/math/defs.cc
anderspapitto: @gramalingam are you interested in re-implementing these to account for the new broadcasting or shall I?

gramalingam: @anderspapitto if you can, that would be great!

docs/Broadcasting.md
linkerzhang: yes, I think.

docs/Broadcasting.md
houseroad: yes

houseroad:
@wchao1115 the unidirectional broadcasting rule is slightly different from numpy broadcasting. So we cannot simply refer to numpy rules.

linkerzhang:
Thank you very much! @houseroad

anderspapitto:
@houseroad did you re-enable all the shape inference tests?

kit1980:
What about Min and Max? Currently, ONNX Min and Max don't support broadcasting at all, while numpy elementwise minimum/maximum have broadcasting.

PR #908, opened 2018-05-08
make werror opt-in

anderspapitto:


.travis/install.sh
```diff
@@ -3,4 +3,4 @@
 script_path=$(python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" "${BASH_SOURCE[0]}")
 source "${script_path%/*}/setup.sh"
 
-time ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI pip install -v .
+time ONNX_WERROR=ON ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI pip install -v .
```
bddppq: I believe environment variable doesn't work, needs to be a cmake variable

CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ project(onnx C CXX)
 option(ONNX_BUILD_BENCHMARKS "Build ONNX micro-benchmarks" OFF)
 
 option(BUILD_ONNX_PYTHON "Build Python binaries" OFF)
-option(ONNX_NO_WERROR "Disable use of Werror" OFF)
+option(ONNX_WERROR "Build with Werror" OFF)
```
snnn: Would better be: "Treat warning as error"

I hope this option also works for Windows.

PR #909, opened 2018-05-08
Changes done internally at Facebook

houseroad:
Fix minor syntax errors/warnings.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -391,7 +391,7 @@ Example 2:
                   .has_dim_value() &&
               starts[j] >= 0 && ends[j] >= 0) {
             auto newval = std::min(
-                              ctx.getInputType(0)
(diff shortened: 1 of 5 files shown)


onnx/defs/tensor/defs.cc
```diff
@@ -381,7 +381,7 @@ Example 2:
             // handle some simple cases.
             if (ctx.getInputType(0)->tensor_type().shape().dim((int)i).has_dim_value() &&
                 starts[j] >= 0 && ends[j] >= 0) {
-              auto newval = std::min(ctx.getInputType(0)->tensor_type().shape().dim((int)i).dim_value(), ends[j]) - starts[j];
+              auto newval = std::min((int64_t)ctx.getInputType(0)->tensor_type().shape().dim((int)i).dim_value(), ends[j]) - starts[j];
```
snnn: Why not add a template argument to std::min? It would be safer is someone changed the type of "ends" in the latter.

PR #911, opened 2018-05-08
[Typing 2/5] Add type hints to onnx/defs

smessmer:
Stacked diff. First commit is reviewed in https://github.com/onnx/onnx/pull/910


onnx/defs/__init__.py
```diff
@@ -5,6 +5,7 @@
 
 import onnx
 import onnx.onnx_cpp2py_export.defs as C
+from typing import Sequence, Any, Text
```
bddppq: Are `Sequence` and `Any` not needed?

onnx/onnx_cpp2py_export/checker.pyi
```diff
@@ -6,7 +6,7 @@ class CheckerContext(object):
     opset_imports: Dict[Text, int] = ...
 
 
-class ValidationError(object):
+class ValidationError(BaseException):
```
bddppq: it actually inherits from `Exception`

PR #912, opened 2018-05-08
[Typing 3/5] Add type hints to some top level files

smessmer:
Stacked diff. First two commits are reviewed in https://github.com/onnx/onnx/pull/911

### DIFF
--- a/onnx/__init__.py
+++ b/onnx/__init__.py
@@ -12,29 +12,31 @@ import onnx.checker  # noqa
 import onnx.defs  # noqa
 
 import google.protobuf.message  # type: ignore
+from typing import Union, Optional, Text, IO
 
+from typing import Union, Text, IO, Optional, cast, TypeVar, Any
 
 # f should be either readable or a file path
-def _load_bytes(f):
-    if hasattr(f, 'read') and callable(f.read):
-        s = f.read()
+def _load_bytes(f):  # type: (Union[IO[bytes], Text]) -> bytes
+    if hasattr(f, 'read') and callable(f.read):  # type: ignore
+        s = f.read()  # type: ignore
     else:
-        with open(f, 'rb') as readable:
+        with open(f, 'rb') as readable:  # type: ignore
             s = readable.read()
     return s
 
 
 # str should be bytes,
 # f should be either writable or a file path
-def _save_bytes(str, f):
-    if hasattr(f, 'write') and callable(f.write):
-        f.write(str)
+def _save_bytes(str, f):  # type: (bytes, Union[IO[bytes], Text]) -> None
+    if hasattr(f, 'write') and callable(f.write):  # type: ignore
+        f.write(str)  # type: ignore
     else:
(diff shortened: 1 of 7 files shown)


onnx/__init__.py
```diff
@@ -12,29 +12,31 @@
 import onnx.defs  # noqa
 
 import google.protobuf.message  # type: ignore
+from typing import Union, Optional, Text, IO
 
+from typing import Union, Text, IO, Optional, cast, TypeVar, Any
```
bddppq: duplicated with line 15

onnx/__init__.py
```diff
@@ -12,29 +12,31 @@
...
 
 # f should be either readable or a file path
-def _load_bytes(f):
-    if hasattr(f, 'read') and callable(f.read):
-        s = f.read()
+def _load_bytes(f):  # type: (Union[IO[bytes], Text]) -> bytes
+    if hasattr(f, 'read') and callable(f.read):  # type: ignore
```
bddppq: why do we need to mark ignore here?

onnx/__init__.py
```diff
@@ -12,29 +12,31 @@
...
+    if hasattr(f, 'write') and callable(f.write):  # type: ignore
+        f.write(str)  # type: ignore
     else:
-        with open(f, 'wb') as writable:
+        with open(f, 'wb') as writable:  # type: ignore
             writable.write(str)
 
 
-def _serialize(proto):
+def _serialize(proto):  # type: (google.protobuf.message.Message) -> bytes
```
bddppq: proto could be of type `bytes`

onnx/__init__.py
```diff
@@ -47,13 +49,16 @@ def _serialize(proto):
...
-        return proto.SerializeToString()
+        result = proto.SerializeToString()  # type: bytes
+        return result
     else:
         raise ValueError('No SerializeToString method is detected. '
                          'neither proto is a str.\ntype is {}'.format(type(proto)))
 
 
-def _deserialize(s, proto):
+_Proto = TypeVar('_Proto', bound=google.protobuf.message.Message)
```
bddppq: Move this above so `_serialize` can also use it?

onnx/__init__.py
smessmer: because mypy doesn't support hasattr() checks yet and complains about the `f.read` (since Text doesn't have a .read member). But I removed it now and replaced it with a type cast, which is a bit less intrusive.

onnx/__init__.py
smessmer: Not needed above. The reason for this is that _Proto makes _deserialize() a generic function. If you call it with ModelProto, it returns a ModelProto. If you call it with NodeProto, it returns a NodeProto. _serialize() above only uses the proto parmeter once, so default variance is doing the job.

onnx/__init__.py
smessmer: fixed

PR #917, opened 2018-05-09
Add type and shape inference for instance normalization.

pranavsharma:
Add type and shape inference for instance normalization.


pranavsharma:
@linkerzhang @gramalingam can one of you approve? thanks.

PR #920, opened 2018-05-09
fix gru, rnn, lstm test cases to match the specification and add some cases

fumihwh:
`lstm`
- cast zeros to `np.float32`. default `np.float64` causes inconsistency of input's dtype.

`lstm`, `rnn` and `gru`
- expand Y dim to match `[seq_length, num_directions, batch_size, hidden_size]`
- add input seq_length = 2 's test case


onnx/backend/test/case/node/rnn.py
```diff
@@ -47,10 +48,17 @@ def f(self, x):
         return np.tanh(x)
 
     def step(self):
-        [w_b, r_b] = np.split(self.B, 2)
-
-        H = self.f(np.dot(self.X, np.transpose(self.W)) + np.dot(self.H_0, self.R) + w_b + r_b)
-        return np.reshape(H, (self.num_directions, self.batch_size, self.hidden_size))
+        h_list = []
```
linkerzhang: instead of having this step function/implementation, shall we have the output result explicitly set in the test case please? otherwise, I'm seeing this case is testing "test code" (this step function).

onnx/backend/test/case/node/lstm.py
```diff
@@ -128,36 +137,128 @@ def export_initial_bias():
...
+        b_r = np.array([[-0.03111535, -0.33301952, -0.17032543, -0.05586442, -0.32023185,
+                         -0.3442515, 0.44716895, -0.2340269, 0.27279443, -0.32590097,
+                         -0.07209095, 0.19784379, 0.27797413, 0.00461715, -0.36984265,
+                         -0.07255584, -0.12558547, -0.11320454, -0.35173178, -0.13761121]]).astype(np.float32)
+        b = np.concatenate((b_w, b_r), axis=1)
+
+        lstm = LSTM_Helper(X=x, W=w, R=r, B=b)
+        output = lstm.step()
+
+        output_precomputed = np.array([[[[-0.09183919, -0.20181324, 0.04475278, -0.12022446,
```
pk-g: Where are these numbers coming from and what is the need to have them here instead of relying on reference implementation? Using precomputed numbers may create unnecessary challenges in future. For instance, If a particular framework generates numbers that do not match these numbers in future, how can we resolve conflicts?

onnx/backend/test/case/node/gru.py
```diff
@@ -27,15 +27,16 @@ def __init__(self, **params):
...
-                params[k] = np.squeeze(params[k], axis=0)
+                if k != X:
+                    params[k] = np.squeeze(params[k], axis=0)
 
-            self.hidden_size = params[R].shape[-1]
-            self.batch_size = params[X].shape[0]
+            hidden_size = params[R].shape[-1]
+            batch_size = params[X].shape[1]
 
-            b = params[B] if B in params else np.zeros(2 * number_of_gates * self.hidden_size)
```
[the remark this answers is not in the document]
fumihwh: I think we don't need to reuse `hidden_size`. All things (init b, p, h, c) need `hidden_size` is done when initialize class.
Even until last version doesn't have it.
The author of this version wanted to use `hidden_size` in step (to reshape the result) and put it in to class.

onnx/backend/test/case/node/lstm.py
fumihwh: It comes from pytorch.
Yes, make sense. I will remove this.

onnx/backend/test/case/node/rnn.py
linkerzhang: haha,  I was seeing this test case is testing the "step" function. I understood that it's used to generate the data. ignore this.

onnx/backend/test/case/node/gru.py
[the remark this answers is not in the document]
pk-g: Thanks, sounds good.

PR #923, opened 2018-05-09
update output shape of RNN ops according to ONNX spec

liqunfu:


liqunfu:
@houseroad @bddppq, @ebarsoum kindly ping. This is to fix the expected output shape of RNN ops.

gramalingam:
Hi, can you please also check if the output type is correctly specified as "float" (when the inputs are "float")? The model that is currently in ONNX seems to have the type wrong (as "double" instead of "float"). I can't view the model produced by this PR, so haven't checked this yet.

liqunfu:
Hi @gramalingam  the mode produced have inputs/output being in float format. I also passed the test with the generated test model and data. Thanks.
Here is the model dump:
Name
test_simple_rnn_defaults
Inputs
X: float[1,3,2]
W: float[1,4,2]
R: float[1,4,4]
Outputs
Y: float[1,3,4]

fumihwh:
`np.reshape(H, (self.num_directions, self.batch_size, self.hidden_size))`
This is not cool.
Luckily current test case has only one direction. Bidirectional will calculate separately and `concat` them finally.
And `H` already has shape `(batch_size, hidden_size)`, we don't need pass them to `step` by putting in `self`.
Anyway, now `reshape` is no problem in calculation but not correct.

PR #926, opened 2018-05-09
update some model files

anderspapitto:


anderspapitto:
so, the process for generating and testing these things is kind of involved, but I have confirmed locally that this fixes the failure in the linked caffe2 diff

PR #928, opened 2018-05-10
Add shape inference for LpPool, RoiPool, and fix MaxPool, AveragePool, and Conv

jaliyae:


onnx/defs/nn/defs.cc
```diff
@@ -24,19 +24,31 @@ static std::string auto_pad_doc =
...
-void convPoolTypeAndShapeInference(InferenceContext& ctx, bool use_dilation, bool require_kernel_shape) {
+void convPoolTypeAndShapeInference(
+    InferenceContext& ctx,
+    bool use_dilation,
+    bool require_kernel_shape) {
   propagateElemTypeFromInputToOutput(ctx, 0, 0);
 
-  if (!hasNInputShapes(ctx, 2)) {
-    return;
+  if (require_kernel_shape) {
```
gramalingam: I think this logic should be combined with lines 83-99 below: check if kernel_shape is available via attribute, and then determine whether we need 1 or 2 input shapes. (I think that the check on NumInputs below probably should be has-1-InputShape, right?)

gramalingam: Sorry, it looks like you need the second input shape for the conv case even if kernel_shape attribute is specified. So, just choose 1 or 2 as required shapes based on require_kernel_shape

onnx/defs/nn/defs.cc
```diff
@@ -259,13 +284,50 @@ std::function<void(OpSchema&)> LpPoolOpSchemaGenerator(const char* name) {
...
 ONNX_OPERATOR_SCHEMA(LpPool).FillUsing(LpPoolOpSchemaGenerator("LpPool"));
 
 } // namespace ONNX_NAMESPACE
 
+// For ROI pool operations.
+void roiPoolTypeShapeInference(InferenceContext& ctx) {
+  propagateElemTypeFromInputToOutput(ctx, 0, 0);
+
+  // expect only one input.
+  if (!hasNInputShapes(ctx, 2)) {
```
gramalingam: doesn't match preceding comment

onnx/defs/nn/defs.cc
```diff
@@ -259,13 +284,50 @@ std::function<void(OpSchema&)> LpPoolOpSchemaGenerator(const char* name) {
...
+  std::vector<int64_t> pooled_shape;
+  if (getRepeatedAttribute(ctx, "pooled_shape", pooled_shape)) {
+    if (pooled_shape.size() != n_input_dims) {
+      return;
+    }
+  } else {
+    return; // cannot produce output shape.
+  }
+
+  int num_rios = static_cast<int>(rios_shape.dim(0).dim_value());
```
gramalingam: Check if dim(0).has_value() first.

onnx/defs/nn/defs.cc
jaliyae: Fixed.

onnx/defs/nn/defs.cc
jaliyae: Change to   *output_shape->add_dim() = rios_shape.dim(0); it could be an inferred dim right?

onnx/defs/nn/defs.cc
```diff
@@ -36,34 +45,31 @@ void convPoolTypeAndShapeInference(InferenceContext& ctx, bool use_dilation, boo
     return;
   }
 
-  size_t n_input_dims = (size_t) (ctx.getInputType(0)->tensor_type().shape().dim_size() - 2);
+  auto input_shape = ctx.getInputType(0)->tensor_type().shape();
```
bddppq: Isn't that getInputType can return `nullptr`? @anderspapitto @gramalingam

gramalingam: The preceding hasNInputShapes check verifies that the inputType is nonnull.

PR #930, opened 2018-05-10
Shape and type inference for Flatten, SpaceToDepth, DepthToSpace

gramalingam:


onnx/defs/shape_inference.h
```diff
@@ -38,6 +38,47 @@ inline int64_t getAttribute(InferenceContext& ctx, const std::string& attributeN
   return defaultValue;
 }
 
+inline TensorShapeProto::Dimension operator*(TensorShapeProto::Dimension dim1, TensorShapeProto::Dimension dim2) {
+  TensorShapeProto::Dimension result;
+  if (dim1.has_dim_value() && dim2.has_dim_value()) {
+    result.set_dim_value(dim1.dim_value() * dim2.dim_value());
+  }	else if (dim1.has_dim_value() && (dim1.dim_value() == 1)) {
```
anderspapitto: is this the default indentation of your editor? I've never seen this style of `}    else if` before

onnx/defs/shape_inference.h
```diff
@@ -38,6 +38,47 @@ inline int64_t getAttribute(InferenceContext& ctx, const std::string& attributeN
...
+  TensorShapeProto::Dimension result;
+  if (dim1.has_dim_value()) {
+	  result.set_dim_value(dim1.dim_value() * dim2);
+  }	else if (dim2 == 1) {
+    return dim1;
+  }
+  return result;
+}
+
+inline TensorShapeProto::Dimension operator/(TensorShapeProto::Dimension dim1, int64_t dim2) {
```
anderspapitto: I feel that dividing dimensions is pretty rare, and what to do when they don't divide evenly is not necessarily so clear and universal that it should necessarily be wrapped up into an operator.

onnx/defs/shape_inference.h
gramalingam: Probably not, but I was sticking to manual formatting to try to keep it similar to the code around. (If I were choosing, I would stick to this style, but I am happy to follow any specified style guidelines.)

Issue #747, opened 2018-04-11
Spec clarity: Semantic versioning vs. simple incremented numbers


NiklasGustafsson:
Reading the document 'Versioning.md,' one is generally left with the impression that ONNX expects version numbers to follow the conventions of semantic versioning, and the text describes best practices in some detail. 

For IR and model versions, it explicitly mentions semantic versioning conventions, but remains silent for operator sets. In fact, the 'prose' example of writing operator ids suggests that versions are simple monotonically increasing numbers:

``
A given operator is identified by a three-tuple: (domain, op_type, and op_version). This is written as domain.op_type:op_version in prose (e.g., com.acme.FastConv:3). Nodes in graphs always refer to operators by their three-part identifier.
``

Other content also suggests that operator sets use simple number increments, as does the Operators.md and Operators-ml.md files, which refer to operator set versions using a simple number scheme.

I note that using semver for IR versions means that we're now on 0.0.3. Based on the conventions for semantic versioning listed in Versioning.md, we should at least be on version 0.3, if not 3.0, depending on whether the changes we introduced were truly BC-breaking (I don't recall). My observation is that the IR versioning to date seems to effectively be using simple incrementing numbers rather than semantic versioning.

I propose that we recognize that the version numbers being integers easily facilitates supporting both semver and simple increasing numbers. We could add a rule that says that leaving the MAJOR/MINOR tuple (0,0) indicates simple numbers, anything else indicates semver.

I further propose that we recognize that the IR version scheme is, effectively, already simple incrementing numbers, and that we change the language in the specification to reflect that.

I also propose that the ONNX specification remain silent on whether models should use semantic versioning or increasing numbers, but that models are expected to follow the outlined conventions if they choose to use semver.

Issue #746, opened 2018-04-11
Spec clarity: Versioning -- 64 bits vs 32 bits.


NiklasGustafsson:
In all the places where version numbers are used in the protobuf specification, they are represented using 64 bits. 

However, the 'Versioning.md' document unambiguously states that version numbers are packed in 32 bits:

``
For efficiency, ONNX serializes the MAJOR, MINOR, and PATCH values as a bit-packed 32-bit integer; the most significant byte is the MAJOR component, the second most significant byte is the MINOR component, the least significant two bytes are the PATCH component.
``

This difference leads to confusion around what the correct encoding is, and needs to be straightened out. 

I see three options:

1. Using 64 bits to store the integer, but only using 32 bits. This is odd, but would require only that the documentation clearly explains the oddity.
2. We simply change the documentation to reflect packing of version numbers into 64 bits. Instead of one byte for each of the major/minor numbers, we would use two for each, and instead of two bytes for the PATCH component, we would use four.
3. Change the proto files to hold the version numbers in 32 bits. This would be a breaking change.

Option no. 1 would be the least impactful, but would perpetuate something rather odd.

So far, neither the IR versions or the operator set versions actually depends on the stated 32-bit encoding, so as long as no published models rely on the encoding described in 'Versioning.md,' going with option no. 2 would not impact anything, but I'm not sure whether we know that.

No. 3 seems to me the least desirable, since it's a breaking change to fix something rather trivial.

linkerzhang:
vote for 2.

prasanthpul:
spec was clarified

PR #931, opened 2018-05-10
Spec clarity: Versioning

NiklasGustafsson:
The current specification text leaves the reader a little bit confused about how versioning is implemented in ONNX. 

As discussed in #746, 64-bit integers are allocated in the IR, but the Versioning.md says that semver numbers are packed in 32 bits. 

As discussed in #747, comments in the protobuf files state that the IR versions are based on semantic versioning. Based on the history of ONNX IR versions, that would mean that we're still just on v0.0, and that the rules for updating major and minor numbers outlined in Versioning.md would have been violated. Therefore, I would argue that the reality is that the IR versioning scheme is based on simple increasing numbers rather than semver, and my suggested changes to the docs reflect that reality.

Niklas
Microsoft


prasanthpul:
@ezyang does this look good to you?

Issue #933, opened 2018-05-11
onnx build fail, onnx.proto: File not found.


ffk0716:
There should be a build dependency between onnx-operators.proto and onnx.proto.
But the CMakeLists.txt does not describe this.

PR #934, opened 2018-05-11
fix build dependency between onnx-operators.proto and onnx.proto

ffk0716:
related to issue 933

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,7 +91,7 @@ endif()
 # function(RELATIVE_PROTOBUF_GENERATE_CPP SRCS HDRS ROOT_DIR)
 # from https://github.com/tensorflow/tensorflow/blob/d2c3b873c6f8ff999a2e4ee707a84ff00d9c15a5/tensorflow/contrib/cmake/tf_core_framework.cmake
 # to solve the problem that customized dir can't be specified when calling PROTOBUF_GENERATE_CPP.
-function(RELATIVE_PROTOBUF_GENERATE_CPP SRCS HDRS ROOT_DIR)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -172,18 +172,22 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP SRCS HDRS ROOT_DIR)
...
   set_source_files_properties(${${SRCS}} ${${HDRS}} PROPERTIES GENERATED TRUE)
   set(${SRCS} ${${SRCS}} PARENT_SCOPE)
   set(${HDRS} ${${HDRS}} PARENT_SCOPE)
 endfunction()
 
-RELATIVE_PROTOBUF_GENERATE_CPP(PROTO_SRCS PROTO_HDRS ${ONNX_ROOT}
+RELATIVE_PROTOBUF_GENERATE_CPP(gen_onnx_proto PROTO_SRCS PROTO_HDRS ${ONNX_ROOT} ""
     onnx/onnx.in.proto
+)
+RELATIVE_PROTOBUF_GENERATE_CPP(gen_onnx_operators_proto PROTO_SRCS PROTO_HDRS ${ONNX_ROOT} gen_onnx_proto
```
bddppq: PROTO_SRCS got overridden to only have the onnx_operators.pb.cc now, this is the root cause in the current CI linkage error

raymondxyang:
+ #933 for reference

CMakeLists.txt
ffk0716: oh!! my mistake, sorry!
I will fix this!

Issue #921, opened 2018-05-09
lstm - pytorch exported pb doesn't match specification


fumihwh:
Use Netron open onnx/backend/test/data/pytorch-operator/test_operator_lstm/model.onnx can get:
![image](https://user-images.githubusercontent.com/6838753/39824565-a4bf96d2-53ea-11e8-87c5-edb132a839e7.png)

Here, output from first LSTM should have shape [seq_length, num_directions, batch_size, hidden_size] following the docs. (In pytorch, should have  (seq_len, batch, hidden_size * num_directions))
And then pass to next nodes [Shape, Gather, Unsqueeze ....] to get second LSTM's initial_h and initial_c which has shape [num_directions, batch_size, hidden_size].

Here is the problem,
normally, Gather should gather batch_size from output, and Concat with num_directions and hidden_size to obtain [num_directions, batch_size, hidden_size].
But pytorch export pb with index 1 of Gather and get num_directions because output shape changed from (seq_len, batch, hidden_size * num_directions) to [seq_length, num_directions, batch_size, hidden_size]. batch_size is no longer at index 1.
Finally get [num_directions, num_directions, hidden_size] to make a wrong initial_h and initial_c for second LSTM.

houseroad:
@anderspapitto would you mind to take a look?

anderspapitto:
@fumihwh is this the same issue as https://github.com/pytorch/pytorch/issues/6879?

fumihwh:
@anderspapitto 
Nearly, but that issue claims weight tensors `W`, `R`, `B`. 
However this issue focuses on output tensors `Y`, `Y_h`, `Y_c`.

jekbradbury:
It also needs to be fixed in onnx-tensorflow, which assumes yet another distinct format for outputs, but we're on that :+1:. Thanks Anders!

PR #937, opened 2018-05-11
Type and shape inference for RNN, LSTM, GRU

gramalingam:


anderspapitto:
looks great! One complication is that there's a bug in the pytorch->onnx->caffe2 conversions that is causing the tests to fail. The bug is https://github.com/onnx/onnx/issues/921 and I'm actively working on it, but in the meantime we need to keep the tests green somehow

gramalingam:
Any suggestions what to do? If the bug is only LSTM (and not RNN/GRU), we could register the inference function for RNN/GRU and not for LSTM.

anderspapitto:
it's kind of annoying, because the tests are scattered across repos, and also anyone running pytorch export will hit this, even outside the tests.

I have two potential suggestions, does either work for you?

1) hold of on pushing this for a day or two until I fix the linked bug
2) add a try-catch around the invocation of checkShapesAndTypes in implementation.h, so that if an exception is thrown for RNN/LSTM/GRU, we simply continue. I will then remove this when I fix the bug.

and yes, it affects all three of RNN/LSTM/GRU.

anderspapitto:
@gramalingam I've merged https://github.com/pytorch/pytorch/pull/7511, so as far as I'm aware the incorrect shapes are fixed, and everything should agree with shape inference now. If you notice anything still being off please let me know.

gramalingam:
Hi @anderspapitto , I noticed that you updated the tests in test/data/pytorch_operator. But there seem to be similar problems with test/data/node (the lstm/rnn/gru cases) ...

anderspapitto:
@gramalingam well all tests are passing in master. It looks like in this PR there are still some errors saying "rank does not match" - could you figure out which inputs/outputs are failing to match, and whether it's to do with this PR or with some existing code?

gramalingam:
@anderspapitto if I look at the ONNX model in the folder https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/node/test_simple_rnn_defaults .and view it, the output shape in the model seems incorrect. (Similarly for the other ops like LSTM.) I don't know where these models come from yet.

gramalingam:
I guess perhaps the model is generated from https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/rnn.py ?

gramalingam:
@pk-g can you please take a look whether the output shape of https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/rnn.py is consistent with the ONNX spec?

gramalingam:
Ok, I also notice an ambiguity in the spec description "It is optional if `output_sequence` is 0." for the first output. I guess this means that: if output_sequence is non-zero, the first output is Y (and the second output, if present, is Y_h). if output_sequence is zero, the first output is Y_h. At least, this is how I am going to interpret it ... please correct me if this is wrong.

anderspapitto:
I think we decided that optional outputs are still present but just contain empty tensors or something like that (so the outputs wouldn't shift) - not sure if we have it written down somewhere.

regarding the test, yes seems like @pk-g is the right person to comment on that - it's unrelated to the fixes I was making

anderspapitto:
"There are two ways to leave an optional input or output unspecified: the first, available only for trailing inputs and outputs, is to simply not provide that input; the second method is to use an empty string in place of an input or output name." - from https://github.com/onnx/onnx/blob/master/docs/IR.md

gramalingam:
@anderspapitto , your interpretation is the one I made. However, I checked our internal implementation and noticed it to be different. It would be good to double check. I don't want to rely too much on IR.md because that was Niklas' attempt at a rational reconstruction in the general case and we need to understand what was intended for this specific operation. Furthermore, with our original interpretation, the attribute "output_sequence" is redundant. @ebarsoum @bddppq can you please confirm which interpretation is right?

skottmckay:
For the three operators all outputs are optional, but only the first output explicitly says "It is optional if `output_sequence` is 0.". Should that statement be made for outputs, resulting in setting output_sequence to non-zero returning valid values in them all? That seems like more deterministic behaviour to me vs. maybe getting an empty string in some places.

gramalingam:
If we are making a new choice, I would suggest dropping the "output_sequence" attribute, and the different outputs appear in fixed positions (0, 1 or 2) (using empty strings as in other ops). However, I am not so sure about breaking compatibility: if existing implementations all agree on some behavior, we could go with that behavior for compatibility.

skottmckay:
Y_h is redundant if Y is returned, so is output_sequence a flag that chooses which of these should be returned?

Not clear to me how to determine if Y_c in LSTM should be returned though.

anderspapitto:
well, I would say that in the scope of this diff, let's just go with what's written in the spec - i.e. the outputs do not shift position. In the particular case of shape inference, when output_sequence == 0, that would mean simply don't infer anything for the first output, and infer as usual for the second and third.

If any one of us really wants to change this representation, let's have a separate discussion.

If you are really averse to committing to an interpretation here, then I would suggest to just bail out if `output_sequence == 0` for now so that we can get this merged, and then come back to it later.

anderspapitto:
@skottmckay the way the spec is written currently, Y_h must always be returned, even if it happens to be redundant with Y. 

Y_c is in a bit of a strange place - I believe there has separately been discussion about adding an additional output to LSTM which is the concatenation of cell states (i.e. <new output> is to Y_c as Y is to Y_h). But it doesn't affect this PR

pk-g:
@gramalingam thanks for brining this up. With regards to the output shape mistmatch, as you correctly mentioned, it appears that with [Pull Request 923](https://github.com/onnx/onnx/pull/923) @liqunfu is providing the required fix.

PR #946, opened 2018-05-13
shape inference for TopK and trigonometric functions

gramalingam:


onnx/defs/math/defs.cc
```diff
@@ -784,7 +784,28 @@ Given two equivalent values, this operator uses the indices along the axis  as
         "Dimension on which to do the sort. Default -1, which indicates the last"
         " axis",
         AttributeProto::INT,
-        static_cast<int64_t>(-1));
+        static_cast<int64_t>(-1))
+	.TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+	    // Type inference:
+	    propagateElemTypeFromInputToOutput(ctx, 0, 0);
```
bddppq: The spaces of these lines aren't aligned up. Could you clang-format them?

PR #951, opened 2018-05-14
Add type hints to numpy_helper_test.py

smessmer:


smessmer:
@onnxbot retest this please

onnx/test/numpy_helper_test.py
```diff
@@ -6,20 +6,21 @@
...
 from onnx import numpy_helper
+from typing import Text
 
 import unittest
 
 
 class TestNumpyHelper(unittest.TestCase):
 
-    def _test_numpy_helper_float_type(self, dtype):
+    def _test_numpy_helper_float_type(self, dtype):  # type: (Text) -> None
```
bddppq: dtype is of numpy.number type. How does this type annotation even pass the CI? :-)

smessmer: Because we `type: ignore` the numpy include, so (1) when calling `_test_numpy_helper_float_type` with `numpy.float32` as argument, it doesn't know anything about `numpy.float32` and (for all it knows) it might be `Text`. (2) inside the function, `dtype` is passed back to numpy again in `.astype(dtype)`. Again, because of the `type: ignore` it doesn't know what argument `astype` is supposed to take and lets it through.

Fixed it.

onnx/test/numpy_helper_test.py
```diff
@@ -6,20 +6,21 @@
 import numpy as np  # type: ignore
 
 from onnx import numpy_helper
+from typing import Text
```
bddppq: no need to import Text anymore

PR #955, opened 2018-05-14
Add type hints to onnx/test/optimizer_test.py

smessmer:


smessmer:
@onnxbot retest this please

PR #958, opened 2018-05-14
shape inference return status

linkerzhang:


onnx/defs/experiments/defs.cc
```diff
@@ -175,25 +181,31 @@ ONNX_OPERATOR_SCHEMA(GivenTensorFill)
...
+		if (!hasInputShape(ctx, 0)) {
+			return Status::OK();
+		}
+
+		std::vector<int64_t> extra_shape;
+		getRepeatedAttribute(ctx, "extra_shape", extra_shape);
+		TensorShapeProto shape = ctx.getInputType(0)->tensor_type().shape();
+		for (auto extra_dim_val : extra_shape) {
+			if (extra_dim_val < 0) {
+				return Status(INFERENCE, INVALID_PROTOBUF, MakeString("extra_shape specified: ", extra_dim_val, " should not be negtive."));
```
gramalingam: When should we use "INVALID_PROTOBUF" and when "INVALID_ARGUMENT"?

onnx/defs/shape_inference.h
```diff
@@ -172,94 +176,108 @@ inline void updateOutputElemType(
...
-inline void updateOutputShape(
+inline Status updateOutputShape(
     InferenceContext& ctx,
     size_t outputIndex,
     const TensorShapeProto& shape) {
   auto* output_shape = getOutputShape(ctx, outputIndex);
-  if (output_shape != nullptr) {
-    *output_shape = shape;
+  if (output_shape == nullptr) {
+	return Status(INFERENCE, INVALID_PROTOBUF, MakeString("Output: ", outputIndex, " should be tensor-type or no-set."));
```
gramalingam: I think this won't work. We need to allow optional outputs (which may be missing, and that is valid). Unfortunately, we can distinguish between the two cases only inside "getOutputShape".

anderspapitto:
@gramalingam the implementation we provide (in implementation.h) has the behavior of

- getOutputType(n) is always a valid, fresh pointer (as long as n < numOutputs)

I think we shouldn't have to really deal with optional inputs/outputs - by the time shape inference runs, there already is some complete graph, and each node we see has a fixed number of inputs/outputs. The shape inference function should simply infer shapes for as many outputs as it can.

onnx/defs/experiments/defs.cc
linkerzhang: I may remove INVALID_ARGUMENT if it's confusing. It's not used in this PR.

onnx/common/status.cc
```diff
@@ -40,6 +40,8 @@ std::string Status::ToString() const {
     result += "[CheckerError]";
   } else if (StatusCategory::OPTIMIZER == state_->category) {
     result += "[OptimizerError]";
+  } else if (StatusCategory::INFERENCE == state_->category) {
+	result += "[InferenceError]";
```
anderspapitto: I would suggest "ShapeInferenceError" in a user-facing message, since Inference is an overloaded term

gramalingam:
@anderspapitto : ok (but this just hides optional non-trailing outputs; for optional trailing outputs, the inference code still needs to check "numOutputs" to figure out if that is present or not)

onnx/defs/shape_inference.h
gramalingam: Ok, please ignore my comment above, as per discussion elsewhere

gramalingam: But, essentially, this condition will never arise as per our discussion.

gramalingam:
@anderspapitto : is there a potential problem in implementation.h? if an optional output indicated by an empty string name has an inferred type, then it gets saved as the type of the empty string, and may be used as the type of another optional input/output elsewhere ... at least implementation.h needs to handle input/outputs with an empty string name, right?

onnx/common/status.cc
linkerzhang: But type inference also covered in the function :). So, TypeShapeInference? if you want.

onnx/defs/tensor/defs.cc
```diff
@@ -251,27 +254,27 @@ Otherwise, the tensor is split to equal sized parts.
...
     .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
       for (int i = 0; i < static_cast<int>(ctx.getNumOutputs()); ++i) {
-        propagateElemTypeFromInputToOutput(ctx, 0, i);
+        ONNX_RETURN_IF_ERROR(propagateElemTypeFromInputToOutput(ctx, 0, i));
       }
 
-      if (!hasNInputShapes(ctx, 1)) {
-        return;
-      }
+	  if (hasNInputShapes(ctx, 1)) {
```
gramalingam: Missing !

linkerzhang: Thank you so much for the catch and fix!

bddppq:
@linkerzhang The CI reported there are core dump in the tests, probably some nullptr dereferencing?

anderspapitto:
@gramalingam yes, there might be an issue like that. Do we currently have any operators with optional, non-trailing outputs? We may actually want to also just come up with a cleaner representation for optional inputs/outputs than "empty string name".

dzhulgakov:
Hey folks, maybe it was discussed before - but why status codes instead of exceptions? It's much easier to propagate errors that way and we already do use exceptions in onnx codebase

linkerzhang:
@dzhulgakov I thought we discussed this before that we agreed to have "Status" introduced into ONNX for error handling. Not all upper callers would like to follow using exception for error handling. For shape inference specifically, it's not an exception if it fails. Runtimes may still be able to handle the case without shape inference.

bddppq:
@linkerzhang If shape inference throws exception, can your code try && catch? I'm really not trying to say exception is better than status code in general, it will be a endless debate. My point is onnx started with using exceptions, there is no good reason to do such an api breaking change, and we definitely need to have a consistent style of error checking.

PR #960, opened 2018-05-14
update onnx model tests for rnn/lstm/gru

anderspapitto:


bddppq:
Could you attach the diff of the proto text? Currently this is really un-reviewable :-)

anderspapitto:
@bddppq well, mostly you can just review the associated code diff. but yes i can pretty-print the proto files here

bddppq:
@anderspapitto ok thanks. the change look reasonable to me @houseroad

PR #961, opened 2018-05-14
[docs] Error case for squeeze operator.

rdzhabarov:
I was browsing through the supported operators and it was not super clear what happens in a certain case for the Squeeze operator.


bddppq:
@linkerzhang @ebarsoum Could you look at this simple schema change (technically not a change)?

houseroad:
Shall we have a more general solution? I mean clarifying the error out mechanism about invalid axis.

bddppq:
@houseroad I don't think it's wise to force all backends to use the same error reporting mechanism. Just clarifying this is an invalid case is already enough.

houseroad:
@bddppq sorry for the confusion, I mean we should do similar things to other operators.

rdzhabarov:
@houseroad @bddppq It's somewhat tough to get the branch not out of date with the frequent master branch updates. I do not see any conflicts with the master branch though. Can you merge the PR?

PR #962, opened 2018-05-15
Clarified description of Pad attribute

RyanUnderhill:
It was not clear that negative values were possible in the 'pads' attribute values.


bddppq:
Hmm I don't think we intend to support negative pads. Have you seen this in our backend test cases/models in the model zoo?

RyanUnderhill:
Yep, it's in a bunch of tests: ConstantPad2d ReflectionPad2D ReplicationPad2D ZeroPad2d and more.

We discovered it due to the public test cases failing due to negative padding values.

prasanthpul:
@RyanUnderhill the documentation is auto generated so you need to update the def files and then run the document generator, rather than updating operators.md directly

linkerzhang:
@houseroad @bddppq  if we think negative padding is not needed, let's fix the tests please. :). Thank you very much!

houseroad:
@linkerzhang Yeah, I agree with @bddppq, it's better to support this semantics since now. Since this is  just a clarification on the doc, I just merged it. :-)

PR #967, opened 2018-05-15
add script to count shape inference implementations

anderspapitto:
at the moment, this outputs

```
103 operators have a type/shape inference function.
12 do not. These are:
 ATen
 ConvTranspose
 Crop
 FC
 GRU
 GRUUnit
 If
 LSTM
 Loop
 LoopIndexTensor
 RNN
 Upsample
```

### DIFF
--- a/onnx/defs/gen_shape_inference_information.py
+++ b/onnx/defs/gen_shape_inference_information.py
@@ -1,0 +1,24 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from onnx import defs
+
+def main():
+    # domain -> support level -> name -> [schema]
+    with_inference = []
+    without_inference = []
+    for schema in defs.get_all_schemas():
+        domain, name, has_inference = schema.domain, schema.name, schema.has_type_and_shape_inference_function
+        if has_inference:
+            with_inference.append((domain, name))
+        else:
+            without_inference.append((domain, name))
+    print(len(with_inference), 'operators have a type/shape inference function.')
+    print(len(without_inference), 'do not. These are:')
+    for domain, name in sorted(without_inference):
+        print(domain, name)
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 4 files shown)


anderspapitto:
We could commit this and perhaps add the output into the docs somewhere, so that we stay up-to-date as more operators are added, or I can just keep it around as a one-off patch to run manually

onnx/defs/gen_shape_inference_information.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import print_function
+
+from collections import defaultdict
```
bddppq: not used

onnx/defs/gen_shape_inference_information.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import print_function
```
bddppq: In general all python files should have these boilerplate lines at the top

```
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
```

onnx/defs/gen_shape_inference_information.py
```diff
@@ -0,0 +1,20 @@
...
+
+from onnx import defs
+
+def main():
+    # domain -> support level -> name -> [schema]
+    with_inference = []
+    without_inference = []
+    for schema in defs.get_all_schemas():
+        domain, name, has_inference = schema.domain, schema.name, schema.has_type_and_shape_inference_function
+        (with_inference if has_inference else without_inference).append((domain, name))
```
bddppq: nit: too magic

onnx/defs/shape_inference.h
```diff
@@ -17,6 +17,8 @@ struct InferenceContext {
 
 typedef void (*InferenceFunction)(InferenceContext&);
 
+inline void dummyInferenceFunction(InferenceContext&) { };
```
bddppq: Maybe add a comment saying this is the default one that intend to do nothing.

PR #968, opened 2018-05-15
Update mypy version

smessmer:


bddppq:
looks like all this breaks all python3 CI

smessmer:
This PR depends on https://github.com/onnx/onnx/pull/982. That one is landed now and I rebased this one - in short, it should be fixed now.

PR #969, opened 2018-05-15
Add onnx/test/__init__.py (otherwise mypy won't check files in it)

smessmer:
There's also #972, which just fixes the errors without enabling mypy for the onnx/test/ folder.


smessmer:
superseded by https://github.com/onnx/onnx/pull/972

PR #970, opened 2018-05-15
mypy info

anderspapitto:


onnx/defs/gen_shape_inference_information.py
```diff
@@ -5,7 +5,8 @@
 
 from onnx import defs
 
-def main():
+
+def main():  # type: () -> None
     # domain -> support level -> name -> [schema]
     with_inference = []
```
smessmer: mypy might not be able to infer the type of these two arrays. If CI doesn't complain, you're fine. If it does, just add a `# type: ` comment with the concrete type to these lines.

PR #971, opened 2018-05-15
add more shape inference tests

anderspapitto:


onnx/test/test_backend_test.py
```diff
@@ -37,7 +38,20 @@ def prepare(cls,
         super(DummyBackend, cls).prepare(model, device, **kwargs)
 
         # test shape inference
-        onnx.shape_inference.infer_shapes(model)
+        model = onnx.shape_inference.infer_shapes(model)
+        value_infos = {vi.name: vi for vi in itertools.chain(model.graph.value_info, model.graph.output)}
+
+        for node in model.graph.node:
+            if node.op_type in set(['RNN', 'LSTM', 'GRU', 'Reshape']):
```
bddppq: let's put this blacklist/whitelist logic into a separate function

onnx/test/test_backend_test.py
```diff
@@ -37,7 +38,20 @@ def prepare(cls,
...
+                break
+            for i, output in enumerate(node.output):
+                if node.op_type == 'Dropout' and i != 0:
```
bddppq: what's special about dropout?

onnx/test/test_backend_test.py
```diff
@@ -62,6 +75,20 @@ def supports_device(cls, device):
...
+shape_coverage_whitelist = set(
+    ['bvlc_alexnet', 'densenet121', 'inception_v1', 'inception_v2',
+     'resnet50', 'shufflenet', 'SingleRelu', 'squeezenet_old', 'vgg19', 'zfnet'])
+
+
+def do_enforce_shape_inference_coverage(model):
+    if model.graph.name not in shape_coverage_whitelist:
+        return False
+    for node in model.graph.node:
+        if node.op_type in set(['RNN', 'LSTM', 'GRU']):
```
bddppq: nit: there is syntax sugar `{...}` for `set([...])`

onnx/test/test_backend_test.py
```diff
@@ -37,7 +38,19 @@ def prepare(cls,
...
+        model = onnx.shape_inference.infer_shapes(model)
+        value_infos = {vi.name: vi for vi in itertools.chain(model.graph.value_info, model.graph.output)}
+
+        if do_enforce_shape_inference_coverage(model):
+            for node in model.graph.node:
+                for i, output in enumerate(node.output):
+                    if node.op_type == 'Dropout' and i != 0:
+                        continue
+                    assert output in value_infos
+                    tt = value_infos[output].type.tensor_type
```
linkerzhang: what will happen in onnx_ml case?

anderspapitto: we may have to extend this logic if we add models using the onnx_ml types to the whitelist

PR #972, opened 2018-05-15
Fix mypy errors in onnx/test/

smessmer:
See https://github.com/onnx/onnx/pull/969 for why these errors weren't found on CI


smessmer:
note: https://github.com/onnx/onnx/pull/982 needs to land first and this needs to be adapted to it. Otherwise things will break.

setup.py
```diff
@@ -242,12 +242,29 @@ def build_extensions(self):
             self.copy_file(src, dst)
 
 
+class mypy_type_check(setuptools.Command):
```
bddppq: Inherit from `ONNXCommand` to saves you some boilerplate.

setup.py
```diff
@@ -242,12 +242,29 @@ def build_extensions(self):
...
+
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        """Run command."""
+        returncode = subprocess.call(["python", "./tools/mypy-onnx.py"])
```
bddppq: 1. Use `sys.executable` instead of `"python"` (this is not nit, because e.g. virtualenv can be used without activation, and so the correct python is actually not in `$PATH`, and also as we recently discovered, many users can have `python3`, `python3.6` as their main python installation).
2. Use absolute path for the `mypy-onnx.py`, otherwise it requires user to invoke from the onnx top directory.

setup.py
```diff
@@ -242,12 +242,29 @@ def build_extensions(self):
...
+        sys.exit(returncode)
```
bddppq: this can be combined with last line as `subprocess.check_call`.

tools/mypy-onnx.py
```diff
@@ -0,0 +1,33 @@
...
+        # Enumerate py and pyi files, they're listed without file extension
+        py_files = [os.path.relpath(os.path.join(dirpath, f), root_folder)
+                   for dirpath, dirnames, files in os.walk(source_folder)
+                   for f in files if f.endswith('.py')]
+
+        subprocess.check_call(["mypy"] + py_files)
+        subprocess.check_call(["mypy", "--py2"] + py_files)
+
+        exit(0)
+    except subprocess.CalledProcessError:
```
bddppq: what's the purpose of this try-catch?

tools/mypy-onnx.py
```diff
@@ -0,0 +1,33 @@
...
+
+def main():
+    try:
+        subprocess.check_call(["mypy", "."])
+        subprocess.check_call(["mypy", "--py2", "."])
+
+        # Since test cases aren't a python package (missing __init__.py),
+        # mypy ignores them. Explicitly call mypy with these files.
+
+        root_folder = os.path.realpath(os.path.join(os.path.dirname(__file__), ".."))
```
bddppq: nit: Do `os.path.abspath` before doing `os.path.dirname` to avoid symlink confusion.
Use `os.pardir` instead of `..` , or you can just add another `os.path.dirname`.

tools/mypy-onnx.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
+
+import subprocess
+import os
+
+
+def main():
+    try:
+        subprocess.check_call(["mypy", "."])
+        subprocess.check_call(["mypy", "--py2", "."])
```
bddppq: use __file__ and some path resolution functions to get the top level directory path

setup.py
smessmer: Yes, but check_call throws an exception, which then outputs a backtrace, cluttering the type errors listed by mypy. So it's either check_call and try/catch or call and sys.exit.

tools/mypy-onnx.py
smessmer: See explanation above. I added an explanatory comment.

tools/mypy-onnx.py
smessmer: I moved it below the `chdir`, that should fix it

PR #973, opened 2018-05-16
Shape inference for ConvTranspose

jaliyae:
Shape inference for ConvTranspose


onnx/defs/nn/defs.cc
```diff
@@ -487,6 +487,125 @@ ONNX_OPERATOR_SCHEMA(Conv).FillUsing(ConvOpSchemaGenerator("a filter"));
...
+  }
+
+  // don't bother with legacy auto_pad for now
+  if (ctx.getAttribute("auto_pad")) {
+    return;
+  }
+
+  auto input_shape = ctx.getInputType(0)->tensor_type().shape();
+  if (input_shape.dim_size() < 2) {
+    return; // The input shape is not properly set.
```
gramalingam: Please change comment to "Input tensor should have at least two dimensions" ... This will be important if we convert these into error messages (as we plan to do over next few days).

onnx/test/shape_inference_test.py
```diff
@@ -819,6 +819,38 @@ def test_global_lppool(self):
             [])
         self._assert_inferred(graph, [make_tensor_value_info("Y", TensorProto.FLOAT, (5, 3, 1, 1))])
 
+    def test_conv_trasnpose(self):
```
gramalingam: transpose => transpose in all these function names?

onnx/defs/nn/defs.cc
```diff
@@ -487,6 +487,125 @@ ONNX_OPERATOR_SCHEMA(Conv).FillUsing(ConvOpSchemaGenerator("a filter"));
...
+      return;
+    }
+  }
+
+  std::vector<int64_t> output_padding;
+  if (getRepeatedAttribute(ctx, "output_padding", output_padding)) {
+    if (output_padding.size() != n_input_dims) { // Added only to one side.
+      return;
+    }
+  } else {
```
gramalingam: Build failure seems to be complaining about a comparison of signed and unsigned integers somewhere here ...

jaliyae: This comparison is both size_t, I think the issue is few lines above. I pushed a new change.

PR #974, opened 2018-05-16
Do not generate protobuf python type stubs if protobuf python package is not installed

bddppq:
type stubs should not be a hard requirements for users

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ project(onnx C CXX)
 option(ONNX_BUILD_BENCHMARKS "Build ONNX micro-benchmarks" OFF)
 
 option(BUILD_ONNX_PYTHON "Build Python binaries" OFF)
+option(ONNX_GEN_PB_TYPE_STUBS "Generate protobuf python type stubs" ON)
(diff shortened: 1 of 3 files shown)


bddppq:
This is actually more error-prone than we thought. On some systems that doesn't by default ship python3, users likely have multiple python installations (python, python2, python3.6), but we can't pass `PYTHON_EXECUTABLE` from cmake to `protoc-gen-mypy.py` and so it always invokes invokes `python`, it can easily run into situation that the default python installation doesn't have protobuf installed.

smessmer:
Let's output a warning in setup.py if protobuf isn't found so users who want type stubs get a clue as to why they don't work.

What do you mean with it being fragile? I assume protoc-gen-mypy is called with the python runmokg setup.py, or not?

bddppq:
@smessmer print a warning sounds good to me.
Users can call setup.py with e.g. `python3.6 setup.py install` (or even when they do `pip install` the pip might belong to a python installation that is different with the `python` on the system.

bddppq:
@smessmer actually, I'm going to remove the check at setup.py, because it might not be checking the correct python installation.

PR #978, opened 2018-05-16
Update avgpool test cases

fumihwh:
add count_include_pad test cases


onnx/backend/test/case/node/averagepool.py
```diff
@@ -192,6 +222,37 @@ def export_averagepool_2d_same_upper():
...
+        iutput_shape: [1, 3, 32, 32]
+        output_shape: [1, 3, 32, 32]
+        pad_shape: [1, 1] -> [0, 1, 0, 1] by axis
+        """
+        node = onnx.helper.make_node(
+            'AveragePool',
+            inputs=['x'],
+            outputs=['y'],
+            kernel_shape=[2, 2],
+            auto_pad='SAME_UPPER',
```
anirudhacharya: seeing that autopad is now deprecated? why is this test being added?

fumihwh: @anirudhacharya 
Tensorflow uses something like `auto_pad` and I am working on backend of tf. 
At least for test case coverage for tf, I added this. 
Of course I am ok with removing it.

anirudhacharya: It would be good to remove this and then evaluate a plan when the deprecated attributes/operators will be EOL'ed.

onnx/backend/test/case/node/averagepool.py
```diff
@@ -252,6 +282,37 @@ def export_averagepool_2d_pads():  # type: () -> None
 
         expect(node, inputs=[x], outputs=[y], name='test_averagepool_2d_pads')
 
+    @staticmethod
+    def export_averagepool_2d_pads_count_include_pad():  # type: () -> None
+        """
+        iutput_shape: [1, 3, 28, 28]
```
anirudhacharya: nit: input_shape

houseroad:
@fumihwh could you use https://github.com/onnx/onnx/blob/master/onnx/backend/test/stat_coverage.py to update the TestCoverage page?

PR #986, opened 2018-05-17
Update shape inference methods to throw exception

gramalingam:


onnx/defs/nn/defs.cc
```diff
@@ -47,7 +47,7 @@ void convPoolTypeAndShapeInference(
 
   auto input_shape = ctx.getInputType(0)->tensor_type().shape();
   if (input_shape.dim_size() < 2) {
-    return; // The input shape is not properly set.
+    fail_shape_inference("Input tensor must have atleast 2 dimensions");
```
linkerzhang: at least

PR #989, opened 2018-05-18
Enable checking for functions that don't have a type hint

smessmer:
Only look at last commit "Enable checking for functions that don't have a type hint", rest is from dependend PR https://github.com/onnx/onnx/pull/988

This also fixes a typo bug found by enabling this setting.


onnx/gen_proto.py
```diff
@@ -181,7 +181,7 @@ def main():
     args = parser.parse_args()
 
     if not os.path.exists(args.output):
-        os.mkdirs(args.output)
+        os.makedirs(args.output)
```
bddppq: good catch

PR #994, opened 2018-05-19
Fix comma in Gemm description

lutzroeder:


houseroad:
I think I already fix this in https://github.com/onnx/onnx/pull/907 :-)

lutzroeder:
@houseroad This is fixed in `Gemm` v7 but the older versions still have the comma on the wrong line.

houseroad:
@lutzroeder yes, you are right, I forgot to fix the problem in the old.cc.

Could you create another PR to solve it?

Thanks!

lutzroeder:
@houseroad #1035

PR #1004, opened 2018-05-21
Better error message if protoc isn't found

smessmer:
Protobuf compiler wasn't found on my system, which caused the cmake ONNX_PROTOC_EXECUTABLE to be not set, and what should have been a command like `protoc onnx.proto ...` became `onnx.proto ...`. The error message was then `onnx.proto: Permission denied` which was quite unhelpful and took me some time to track down.

This PR improves the error message for this case.


bddppq:
I blame cmake on this :-)

PR #1005, opened 2018-05-21
Fix .gitignore and add missing files

smessmer:
`.gitignore` was set up in a way that prevented some of the stubs/google/protobuf files being uploaded when I added these. Re-adding them and fixing `.gitignore`.


bddppq:
Just to double check, this is only temporary and once your PRs to the upstream google protobuf repository are merged we can change to refer the upstream instead of having these files copied into onnx repo?

smessmer:
yes

PR #1006, opened 2018-05-21
Simplify and clarify RNN spec by omitting output_sequence attribute

gramalingam:
* Drop the output_sequence attribute
* Extend backend test case node to handle optional inputs/outputs
* Fix test cases accordingly


linkerzhang:
Thank you very much! @gramalingam 

Do you want to document "optional" input/output semantics somewhere please? including if an optional input or output does not exist for a graph, then an empty string needs to be specified as place holder.

gramalingam:
@linkerzhang : but isn't that the same treatment of optional inputs/outputs for all ops? does it require special mention?

wschin:
RNNs use optional inputs and outputs a lot. It'd be very helpful if you can add a small note. You can also point out where to find doc to specify unused I/O.

anderspapitto:
I (weakly) vote against explicitly talking about optional outputs in the documentation of RNN. After this change, the outputs are no more special than the outputs of e.g. MatMul. In particular, it is not necessary to provide an empty string - any name can be provided as an output, and then you can simply choose not to provide that name as an input to any further ops.

anderspapitto:
@wschin I'm fine with adding a pointer to docs related to optional inputs, because they are a real special case that needs special handling. I'm just saying that outputs don't need any special handling (and this diff happens to be about outputs)

gramalingam:
@NiklasGustafsson I updated a line describing optional outputs in IR.md. Hope this looks fine.

PR #1008, opened 2018-05-21
Add Cast history back

houseroad:
Add the old version of Cast back.


onnx/defs/tensor/defs.cc
```diff
@@ -8,6 +8,7 @@
 using namespace ONNX_NAMESPACE;
 
 ONNX_OPERATOR_SCHEMA(Cast)
+    .SinceVersion(6)
```
linkerzhang: 6 or 7?

linkerzhang:
@bddppq this is not related to this PR. I realized that "shape" was also not since version 1, right? :)

onnx/defs/tensor/defs.cc
houseroad: We introduced this change in opset 6, I think it's better to keep the truth in our code.

houseroad: Any particular reason to use 7? If there is, we can use 7.

bddppq:
@linkerzhang yeah that was not :(

onnx/defs/tensor/defs.cc
bddppq: This should be 6, which was the time we introduced this op.

PR #1010, opened 2018-05-22
[WIP] Introducing Some Operators Required by Object Detection Models

houseroad:
Still in progress, only RoIAlign's definition is ready for review; other parts are still under development.

Operator definitions:
- [ ] BBoxTransform
- [ ] BoxWithNMSLimit
- [ ] GenerateProposals
- [x] RoIAlign

Node tests:
- [ ] BBoxTransform
- [ ] BoxWithNMSLimit
- [ ] GenerateProposals
- [ ] RoIAlign

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -431,6 +431,71 @@ This version of the operator has been available since version 1 of the default O
 <dd>Constrain input and output types to float tensors.</dd>
 </dl>
 
+### <a name="BoxWithNMSLimit-1"></a>**BoxWithNMSLimit-1**</a>
+
+  Apply NMS to each class (except background) and limit the number of
+  returned boxes.
+
+#### Version
+
+This version of the operator has been available since version 1 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>detections_per_im</tt> : int</dt>
+<dd>(int) TEST.DEECTIONS_PER_IM</dd>
+<dt><tt>nms</tt> : float</dt>
+<dd>(float) TEST.NMS</dd>
+<dt><tt>score_thresh</tt> : float</dt>
+<dd>(float) TEST.SCORE_THRESH</dd>
+<dt><tt>soft_nms_enabled</tt> : int</dt>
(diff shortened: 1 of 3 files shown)


wschin:
For those very specific operators, in addition to high-level descriptions, users may also need mathematical documentation for implementing them. Otherwise, it might lead to some reverse engineering and hard to sync with all partners.

houseroad:
@wschin yeah, I will add more description, and make sure people can understand the logic behind.

Maratyszcza:
@houseroad Would be great to use this opportunity to find better names for operators.

houseroad:
@Maratyszcza i am totally open to it. Let's have some offline discussion next week :-)

houseroad:
Some design decision needs to be made here. I think RoIAlign is fine. For the other three, we probably need to reconsider how to define the API.

houseroad:
@linkerzhang if you guys want to propose other design, feel free to create new PR(s). I can just drop this one.

zhangxiang1993:
Is this PR still in Progress? Thanks! @houseroad

daquexian:
@houseroad What's the current states of detection ops? :) Is there something I can help with (like tests)?

houseroad:
@daquexian thanks for offering help. I think RoIAlign is good to go. Could you create a separate PR for it and adding some tests for RoIAlign? Other ops, not sure they are general enough to be pushed into ONNX spec, since most of frameworks don't have those ops.

Maratyszcza:
These operators bear a lot of legacy issues. In particular, storing batch index in a float tensor is a 100% bad idea. There are refactored versions of Mask R-CNN ops in pytorch/pytorch#15889, they would be a better starting point for ONNX standartization.

houseroad:
@daquexian before working on RoIAlign, checking @Maratyszcza's proposal should be a good idea. :-)

daquexian:
@houseroad Hi, sorry for the delay :) I have read @Maratyszcza's implementation. Does it mean it is better to add a separate `batch_split` input to `RoIAlign`, instead of storing `batch_id` in `rois`?

prasanthpul:
@houseroad should this be closed since there are now PRs for the individual ops?

PR #1012, opened 2018-05-22
Updating Doc of ConvTranspose to be more clear

nickfeeney:
This PR updates the ConvTranspose description to more accurately describe how padding parameters interact. Also lifts the restriction on output_padding which is unnecessary.


houseroad:
@nickfeeney could you sign the CLA?

prasanthpul:
Nick has signed the CLA (verified in the CLA logs). Might be an email mismatch. In any case, let's not block on that.

PR #1013, opened 2018-05-22
refactor matmul shape inference for generality

anderspapitto:
I've factored out a function that I will use to implement shape inference after https://github.com/onnx/onnx/pull/907 lands.

However this diff is standalone and can be pushed separately

### DIFF
--- a/onnx/defs/shape_inference.h
+++ b/onnx/defs/shape_inference.h
@@ -322,4 +322,50 @@ inline void propagateShapeFromAttributeToOutput(
   updateOutputShape(ctx, outputIndex, shape);
 }
 
+inline void bidirectionalBroadcastShapeInference(
+    const TensorShapeProto& shapeL,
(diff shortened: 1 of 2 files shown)


onnx/defs/shape_inference.h
```diff
@@ -322,4 +322,50 @@ inline void propagateShapeFromAttributeToOutput(
   updateOutputShape(ctx, outputIndex, shape);
 }
 
+inline void bidirectionalBroadcastShapeInference(
```
bddppq: Call it numpyBroadcast?

houseroad: I think we won't have any function called `unidirectionalBroadcastShapeInference`, so calling it `numpyBroadcasting` is fine.

houseroad:
Yes, there are many test cases in shape_inference_test.py for matmul.

PR #1014, opened 2018-05-22
correct op spec to follow same optional output semantics.

linkerzhang:
This is a following-up PR of https://github.com/onnx/onnx/pull/1006 to clean all op specs with same 'optional" output semantics.

Will refer to the same piece of documentation once PR #1006 merged.


docs/Operators.md
```diff
@@ -1161,13 +1159,13 @@ Other versions of this operator: <a href="Changelog.md#BatchNormalization-1">Bat
 <dt><tt>Y</tt> : T</dt>
 <dd>The output tensor of the same shape as X.</dd>
 <dt><tt>mean</tt> (optional) : T</dt>
-<dd>The running mean after the BatchNormalization operator. Must be in-place with the input mean. Should not be used for testing.</dd>
+<dd>The running mean after the BatchNormalization operator. Must be in-place with the input mean.</dd>
```
gramalingam: Omit "Must be in-place with the input mean"

docs/Operators.md
```diff
@@ -1161,13 +1159,13 @@ Other versions of this operator: <a href="Changelog.md#BatchNormalization-1">Bat
...
 <dt><tt>var</tt> (optional) : T</dt>
-<dd>The running variance after the BatchNormalization operator. Must be in-place with the input var. Should not be used for testing.</dd>
+<dd>The running variance after the BatchNormalization operator. Must be in-place with the input var.</dd>
```
gramalingam: Omit reference to in-place update

docs/Operators.md
linkerzhang: will fix.

docs/Operators.md
linkerzhang: will fix.

PR #1015, opened 2018-05-22
Use std::function for shape and type inference

jeffbloo:
This switches InferenceFunction to use std::function to allow capturing of state at registration sites.


anderspapitto:
What's an example of state that you want to close over? It's possible the need for this indicates that we should expand InferenceContext

jeffbloo:
@anderspapitto - Enabling indirection to an inference function of a different signature which is only made available at runtime

PR #1017, opened 2018-05-23
change travis ci as onnx_ml=1 covers onnx_ml=0.

linkerzhang:
…educe the ci latency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,22 +4,6 @@ matrix:
   include:
     - os: linux
       sudo: required
-      env: PYTHON_VERSION=python2
-      language: python
-      python: "2.7"
-    - os: linux
-      sudo: required
-      env: PYTHON_VERSION=python3
-      language: python
-      python: "3.6"
-    - os: osx
-      osx_image: xcode9.3beta
-      env: PYTHON_VERSION=python2
-    - os: osx
-      osx_image: xcode9.3beta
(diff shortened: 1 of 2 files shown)


linkerzhang:
it is covered, right?

bddppq:
@linkerzhang 
> "the code should be always be built with ONNX_ML =0 once it could be built when ONNX_ML=1" 

Not really, we have spotted couple times in PRs that people forgot adding ifdefs to guard access to onnxml stuffs.

wschin:
@bddppq , does it mean it's a good reason to add ONNX_ML=1 for test? Those situation can be avoided if we have tests.

bddppq:
@wschin Sorry I don't quit follow your suggestion. Could you elaborate how is that going to help?

wschin:
If one doesn't consider ONNX_ML in their PR, the unit test related that PR may throw so that the author can fix it before checking in the changes.

bddppq:
@wschin code example of such tests?

wschin:
That test makes sense! If we have ONNX_ML=0 and ONNX_ML=1, we need to test both of them. I have another question; couldn't everyone use ONNX_ML=1? We can still separate traditional ML ops from default onnx domain.

bddppq:
Closing. We can revisit this if someday ONNX_ML={0,1} are unified.

PR #1031, opened 2018-05-25
remove mypy which blocks release.

linkerzhang:
I'd remove it to unblock release and we can discuss later to understand it for both linux and windows.


bddppq:
@smessmer If the CI breaks, could you help adjust it to skip the type checking stuffs to unblock release?

linkerzhang:
@smessmer  please kindly fix the CI failure and we may discuss whether we need type stubs and also come out of a solution working well for both windows and linux. Thank you very much!

PR #1046, opened 2018-05-29
fuse_arithmetic_into_batch_norm optimizer

fumihwh:
Fuse arithmetic into BatchNorm. Cases are following.
```
Before:
  case 1:
    Y = BatchNorm(X, scale, bias, mean, var)
    Z = Add(Y, A)
  case 2:
    Y = BatchNorm(X, scale, bias, mean, var)
    Z = Sub(Y, A)
  case 1:
    Y = BatchNorm(X, scale, bias, mean, var)
    Z = Mul(Y, A)
  case 2:
    Y = BatchNorm(X, scale, bias, mean, var)
    Z = Div(Y, A)
After:
  case 1:
    B = Add(bias, A)
    Y = BatchNorm(X, scale, B, mean, var)
  case 2:
    B = Sub(bias, A)
    Y = BatchNorm(X, scale, B, mean, var)
  case 3:
    S = Mul(scale, A)
    B = Mul(bias, A)
    Y = BatchNorm(X, S, B, mean, var)
  case 4:
    S = Div(scale, A)
    B = Div(bias, A)
    Y = BatchNorm(X, S, B, mean, var)
  and there is no Z in graph.
```


onnx/optimizer/passes/fuse_arithmetic_into_batch_norm.h
```diff
@@ -0,0 +1,129 @@
...
+  void fuse_arithmetic_into_batch_norm(Graph& graph) {
+    for (auto it = graph.begin(); it != graph.end(); ++it) {
+      auto* n = *it;
+      DescendOnGraphAttributes(
+          n, [this](Graph& g) { fuse_arithmetic_into_batch_norm(g); });
+
+      // target kinds of arithmetic node
+      std::vector<NodeKind> kind_group = {kAdd, kSub, kMul, kDiv};
+      NodeKind node_kind = n->kind();
+      if (std::find(kind_group.begin(), kind_group.end(), n->kind()) !=
```
anderspapitto: I have a slight preference to change from 

```
if (kind matches) {
  do_stuff();
}
```

to

```
if (! kind matches) {
  continue;
}
do_stuff();
```

onnx/optimizer/passes/fuse_arithmetic_into_batch_norm.h
```diff
@@ -0,0 +1,129 @@
...
+        if (!orig_batch_norm || !orig_const) {
+          continue;
+        }
+
+        // check if BatchNorm is only used by one node
+        bool bn_multi_use = false;
+        for (auto output : orig_batch_norm->node()->outputs()) {
+          if (output->uses().size() > 1) {
+            bn_multi_use = true;
+            continue;
```
anderspapitto: this `continue` is dead code - change it to either `break` or simply remove it

onnx/optimizer/passes/fuse_arithmetic_into_batch_norm.h
```diff
@@ -0,0 +1,129 @@
...
+          }
+        }
+        if (bn_multi_use) {
+          continue;
+        }
+
+        // check if Arithmetic is only used by one node
```
anderspapitto: why is this necessary?

anderspapitto:
The math operators support broadcasting, but I don't believe that scale/bias parameters to BatchNorm do. You need to disable this optimization when broadcasting is used (and then please add a test for this).

fumihwh:
@anderspapitto from opset 7, there is no more broadcast attribute.
I think add a constraint of `arithmetic param dim should not be greater than batch norm dim` is good enough.
What do you think?

onnx/test/optimizer_test.py
```diff
@@ -618,6 +618,37 @@ def test_fuse_arithmetic_into_batch_norm_multi_broadcast(self):  # type: () -> N
...
+        arith = helper.make_node("Add", ["Y", "A"], ["Z"])
+        graph = helper.make_graph(
+            [batch_norm, arith],
+            "test",
+            [helper.make_tensor_value_info("X1", TensorProto.FLOAT, (1, 5, 3, 3)),
+             helper.make_tensor_value_info("X2", TensorProto.FLOAT, (5,)),
+             helper.make_tensor_value_info("X3", TensorProto.FLOAT, (5,)),
+             helper.make_tensor_value_info("X4", TensorProto.FLOAT, (5,)),
+             helper.make_tensor_value_info("X5", TensorProto.FLOAT, (5,)),
+             helper.make_tensor_value_info("A", TensorProto.FLOAT, (1, 1, 5, 1, 1))],
```
anderspapitto: what if the size of A were

- (5, 1, 1)
- (1, 5, 1, 1)
- (4, 5, 3, 3)
- (4, 5, 1, 1)

these are all valid graphs, which should disable this optimization, right? Because when they use an explicit Add operator, they use generalized numpy broadcasting, but when they are converted to use scale/bias, there is no broadcasting allowed

fumihwh: @anderspapitto I got your point. Yes, it's right. Let me fix it.

zakk0610:
Could it support below case?
Before:
Y = BatchNorm(X, scale, bias, mean, var)
Z = MUL(Y, A, broadcast=1, axis=1)
After
Y = BatchNorm(X, scale, bias, mean, var)

like https://github.com/onnx/onnx/pull/1139?

BTW, this optimization cannot handle tensor operand, right?

fumihwh:
@zakk0610 Sorry for unexplaining. From opset version 7, we introduced a new strategy for broadcasting which is almost same to numpy.
According to that, we don't need to support case you mentioned above.

zakk0610:
Hi @fumihwh, 
Sorry i don't understand why this optimization need to insert arithmetic and squeeze node in graph.
Could you explain why this optimizer does not like fuse_bn_into_conv optimizer, which only need to update tensor?

fumihwh:
@zakk0610 
For example,
```
X = BN(input, scale, bias)    with shape_scale=(C), shape_bias=(C)
Y = Add(X, A)    with shape_A=(1, C, 1, 1)
↓
Z = Squeeze(Y)    shape=(C)
B = Add(bias, Z)    shape=(C)
C = BN(input, scale, B)
```
`bias` always has shape (C) and if `Y` has shape (1, C, 1, 1), we should do squeeze first and add them. Then pass new_bias as an input to BN.

zakk0610:
@fumihwh Got it, thanks.

If A is tensor (weight), can we fuse A into bias and remove Add operator?
For example, 
```
X = BN(input, scale, bias)    with shape_scale=(C), shape_bias=(C)
Y = Mul(X, A)    with shape_A=(1, C, 1, 1)  ; A is Tensor (weight) 
Z = Add(Y, B)    with shape_B=(1, C, 1, 1)  ; B is Tensor (weight) 
↓
C = BN(input, new_scale, new_bias)    with new_scale = scale * A ; 
                                                            with new_bias = bias * A + B
 ```
If the model is converted from Caffe->Caffe2->ONNX, the above case will be happened, because Caffe uses two layer (BatchNorm+Scale) to implements BatchNorm operator.

or this fuse optimization should be done in converter tool?

Thanks

**Comment by fumihwh** [comment]
@zakk0610 
This optimizer can handle case you mentioned.

**Comment by zakk0610** [comment]
@fumihwh 
Why not to replace old tensor by new tensor?
like [this]( https://github.com/onnx/onnx/pull/1139/files#diff-3c66244f77c2f73f1041ba1e989aeda3R98).
Then we don't need to add Add and Mul op.
Thanks.

```
 Tensor scale = *scale;
 Tensor bias = *bias;
 Tensor A = *A;
 Tensor B = *B;
 scale.multiply(A);
 bias.multiply(A);
 bias.add(B);
// update bn's inputs
```

**Comment by fumihwh** [comment]
@zakk0610 
Yes, we could do this.

**Comment by zakk0610** [comment]
@fumihwh 
Yes, you are right. it's my mistake, I use the wrong example.
This opt will fuse two tensors if they have the same shape.

for example, I get the following case when the model is converted form Caffe
```
X = Conv ...              with shape_X = (N, C, H, W)
Y = BN(X, scale, bias)    with shape_scale=(1, C, 1, 1), shape_bias=(1, C, 1, 1)
Z = Mul(Y, A)             with shape_A=(1, C, 1, 1)  ; A is Tensor (weight) 
W = Add(Z, B)             with shape_B=(1, C, 1, 1)  ; B is Tensor (weight) 
```
If we can fuse tensor(BN+Mul-->BN, BN+Add-->BN) and apply this optimization https://github.com/onnx/onnx/pull/1106, 
The result will be very simple.
```
X = Conv with new_weight, new_bias
```
It's why I post the PR https://github.com/onnx/onnx/pull/1139.
Thanks.

**Comment by fumihwh** [comment]
@zakk0610 
In your example, that model does not meet the spec.
`Y = BN(X, scale, bias)    with shape_scale=(1, C, 1, 1), shape_bias=(1, C, 1, 1)`
scale and bias MUST have shape `(C,)`, otherwise, the model is wrong.

So, with correct model,
```
X = Conv ...              with shape_X = (N, C, H, W)
Y = BN(X, scale, bias)    with shape_scale=(C,), shape_bias=(C,)
Z = Mul(Y, A)             with shape_A=(1, C, 1, 1)  ; A is Tensor (weight) 
W = Add(Z, B)             with shape_B=(1, C, 1, 1)  ; B is Tensor (weight) 
```
We can't just update scale and bias in-place by using A and B now.
So I add squeeze and arithmetic node before BN.
```
X = Conv ...              with shape_X = (N, C, H, W)
NEW_A = Squeeze(A)    with shape_A=(1, C, 1, 1)  ; A is Tensor (weight), and shape_NEW_A=(C,)
NEW_B = Squeeze(B)    with shape_B=(1, C, 1, 1)  ; B is Tensor (weight), and shape_NEW_B=(C,)
NEW_SCALE = Mul(scale, NEW_A)      scale is original scale, shape_scale=(C,), shape_NEW_SCALE=(C,)
NEW_BIAS = Mul(bias, NEW_A)
NEW_BIAS = Add(B, NEW_BIAS)    bias is original bias, shape_bias=(C,), shape_NEW_BIAS=(C,)
Y = BN(X, NEW_SCALE, NEW_BIAS)
```

**Comment by zakk0610** [comment]
@fumihwh 
Got it, thanks for your patience. 
I made a wrong example again , so sorry... 

The root cause i want to optimize is the model converted from Caffe. (caffe->caffe2->onnx)
Caffe implement BatchNorm by BN + Scale layer.
```
BN = ( X - mean ) / (sqrt(var))
Scale = BN  * gamma + beta   //  (channel-wise)
```
in the opset version 6, the converter will generate below case
```
X = Conv ...              with shape_X = (N, C, H, W)
Y = BN (X, scale(1), bias(0), mean, var)
Z = Mul(Y, gamma ,axis=1, broadcast=1)
W = Add(Z, beta, axis=1, broadcast=1)
```
as you mentioned before, in the opset version 7, there is no more broadcast attribute, 
But in naturally, I can use only one BN Operator to instead of BN + Mul + Add, Right?
or maybe this opt should be done in converter tool?

fumihwh:
> But in naturally, I can use only one BN Operator to instead of BN + Mul + Add, Right?
Yes, we are aiming to achieve this also. But currently, I can't update tensor inplace as I said before. (Else if there is a way to squeeze tensor, for example, improving `tensor.h` to support more operators.)
What I can do is using Squeeze + Mul + Add + BN instead of BN + Mul + Add.
BTW, optimization does not have version compatibility. We will not consider opset v6.

zakk0610:
@fumihwh  Got it, Thanks for the explanation.
BTW, currently is there any benefit of this fuse optimization?

fumihwh:
@zakk0610 
I think although number of nodes does not reduce by this optimization, the dependencies of node BN be more concise than before.
In another word, all arithmetic node after BN are removed and operands are changed to constant, it will be easier to deal with in edge side (complier or something else).

zakk0610:
@fumihwh Thanks for your opinion.

askhade:
Closing this PR since the optimizers are now moved to a separate repo. @fumihwh if this is still need please open this PR in optimizers

PR #1047, opened 2018-05-29
Generate protoc type hints on Windows

smessmer:
This PR adds type hint generation for our .proto files on Windows clients.

Also adds type checks to AppVeyor CI to make sure we don't break the Windows platform in the future.


CMakeLists.txt
```diff
@@ -185,6 +185,8 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP NAME SRCS HDRS ROOT_DIR DEPEND)
           # If onnx was packaged to pypi from Windows, protoc-gen-mypy.py is missing the +x flag. Add it.
           execute_process(COMMAND chmod +x ${ROOT_DIR}/tools/protoc-gen-mypy.py)
           list(APPEND PROTOC_ARGS --plugin protoc-gen-mypy=${ROOT_DIR}/tools/protoc-gen-mypy.py --mypy_out ${ONNX_DLLEXPORT_STR}${OUTPUT_PROTO_DIR})
+        else(NOT WIN32)
```
bddppq: nit: maybe only use the if-else to set a variable and then put this long `list(APPEND ...)` in common code path?

PR #1048, opened 2018-05-29
[WIP] Add mypy to Windows CI

smessmer:
[work in progress]

Since mypy now works on Windows (see https://github.com/onnx/onnx/pull/1047), this adds it to Windows CI to make sure it stays this way.


smessmer:
merged into #1047

PR #1049, opened 2018-05-30
add arg test plus a few minor fixes

liqunfu:


onnx/backend/test/case/node/upsample.py
```diff
@@ -18,7 +18,7 @@ def export_nearest():  # type: () -> None
             'Upsample',
             inputs=['x'],
             outputs=['y'],
-            scales=[1.0, 1.0, 2.0, 3.0],
+            scales=[2.0, 3.0],
             mode='nearest',
```
linkerzhang: no. it was discussed and agreed to keep the element number of scales the same as the rank of inputs. In this test case, the input's rank is 4, and we don't want to upsample the 1st and 2nd dimension, so it's 1.0. Make sense?

linkerzhang: As the upsample op now support multiple-D input tensor now (instead of 4-D tensor previously).

PR #1050, opened 2018-05-30
Extract constant to initializer

zrphercule:
I am still writing the testcases, so I submitted the code for review first. This commit contains a new optimization pass that will extract all Constant nodes in the graph to initializer. Testcases will be submitted later.


onnx/optimizer/passes/extract_constant_to_initializer.h
```diff
@@ -0,0 +1,47 @@
...
+                    Tensor t = n->t(sym);
+                    auto name = n->output()->uniqueName();
+               
+                    graph.addInitializer(t, name);
+                     
+                    Node* param = graph.create(kParam, 1);
+                    std::vector<Dimension> s = {1};
+                    param->output()->setUniqueName(name);
+                    param->output()->setSizes(s);
+                    param->output()->setElemType(TensorProto_DataType_INT64);
```
bddppq: You can inspect the shape and elem type from the Tensor attribute as well

onnx/optimizer/passes/extract_constant_to_initializer.h
```diff
@@ -0,0 +1,51 @@
...
+            : OptimizePass("extract_constant_to_initializer", API_TYPE::IR){
+            }
+
+        void extract_constant_to_initializer(Graph& graph){
+            for (auto it = graph.begin(); it != graph.end(); ++it){
+                auto* n = *it;
+                DescendOnGraphAttributes(n, [this](Graph& g){extract_constant_to_initializer(g);});
+                if (n->kind() == kConstant){
+                    
+                    auto name = n->output()->uniqueName();
```
bddppq: nit: const

anderspapitto:
the windows build failure is legit - if you click through, you'll see that it's complaining because you're implicitly converting between integer types somewhere.

Once you fix that and all tests are passing, looks great!

zrphercule:
Yes, there is an implicity converting of int_64 and int, and maybe more. Let me try to find some way out.

onnx/optimizer/passes/extract_constant_to_initializer.h
```diff
@@ -0,0 +1,51 @@
...
+                    
+                    const auto name = n->output()->uniqueName();
+                    
+                    Symbol sym = Symbol("value");
+                    Tensor t = n->t(sym);
+                    t.setName(name); 
+                    graph.addInitializer(t, name);
+
+                    std::vector<Dimension> tsizes;
+                    for (auto v : t.sizes())
```
dzhulgakov: style guide: add {} over all loops and if statements even if they're one line

onnx/optimizer/passes/extract_constant_to_initializer.h
```diff
@@ -0,0 +1,51 @@
...
+                auto* n = *it;
+                DescendOnGraphAttributes(n, [this](Graph& g){extract_constant_to_initializer(g);});
+                if (n->kind() == kConstant){
+                    
+                    const auto name = n->output()->uniqueName();
+                    
+                    Symbol sym = Symbol("value");
+                    Tensor t = n->t(sym);
+                    t.setName(name); 
+                    graph.addInitializer(t, name);
```
dzhulgakov: std::move(t) ?

onnx/optimizer/passes/extract_constant_to_initializer.h
```diff
@@ -0,0 +1,51 @@
...
+
+        void extract_constant_to_initializer(Graph& graph){
+            for (auto it = graph.begin(); it != graph.end(); ++it){
+                auto* n = *it;
+                DescendOnGraphAttributes(n, [this](Graph& g){extract_constant_to_initializer(g);});
+                if (n->kind() == kConstant){
+                    
+                    const auto name = n->output()->uniqueName();
+                    
+                    Symbol sym = Symbol("value");
```
dzhulgakov: It's interned - I guess you can just use `kValue`

zrphercule: I guess you mean kvalue?

onnx/common/ir.h
```diff
@@ -73,6 +73,7 @@ struct Dimension final {
   Dimension(std::string param)
     : is_int(false), dim(-1), param(std::move(param)) {
   }
+  Dimension(int64_t dim) : is_int(true), dim(dim) {}
```
bddppq: @zrphercule @anderspapitto Should we change the original version to take `int64_t`? I think then the compiler won't complain for neither passing in `int` nor `int64_t`.

zrphercule: I guess you are right. I see no other use of this "int" version, but I dunno if there is any.

houseroad: Yeah, I also vote for `int64_t`. We use `int64_t` for dimensions in onnx proto, so it would be nice to keep consistency.

onnx/optimizer/passes/extract_constant_to_initializer.h
```diff
@@ -0,0 +1,46 @@
+// ATTENTION: The code in this file is highly EXPERIMENTAL.
+// Adventurous users should note that the APIs will probably change.
+
+#pragma once
+
+#include "onnx/optimizer/passes/optimize_pass.h"
+
```
houseroad: Nit: add pseudo code to explain the effect of the optimization pass.

Something like the following code:
https://github.com/onnx/onnx/blob/master/onnx/optimizer/passes/eliminate_unused_initializer.h#L7

bddppq:
CI failures are unrelated, it's the model download from the model zoo being flaky.

dzhulgakov:
Actually @houseroad - has a point: do we have across the board support for nested initializers?

e.g. nothing in https://github.com/pytorch/pytorch/blob/master/caffe2/python/onnx/backend.py handles it properly. We should decide one way or another. We might even consider banning nested initializers in checker if we don't handle them yet.

bddppq:
IMO subgraph should be just a normal graph, and it's better to not make it to be different with the top level graph. Extracting blobs from subgraph to parent level makes the global graph looks messy, and as mentioned above it can potentionally hit name conflict. Re. support in c2, I feel it's lacking proper support of nested graph in general, a lot of places don't descend into subgraphs to do the same processing as for the top-level graph, so I don't want to use it  to judge whether a feature should be supported in onnx or not.

onnx/common/ir.h
anderspapitto: sounds good to me

onnx/optimizer/passes/extract_constant_to_initializer.h
```diff
@@ -0,0 +1,54 @@
...
+        t.setName(name);
+        std::vector<Dimension> tsizes;
+        for (auto v : t.sizes()) {
+          tsizes.push_back(v);
+        }
+        graph.addInitializer(std::move(t), name);
+        Node* param = graph.create(kParam, 1);
+        param->output()->setUniqueName(name);
+        param->output()->setSizes(tsizes);
+        param->output()->setElemType(t.elem_type());
```
linkerzhang: this is not ok after std::move(t), right?

bddppq: Yep need to move the initializer after this

wschin:
If constants are moved to initializer list, does it mean that constant can be changed anyway? For example, to store labels of a classifier, I prefer to use a constant than an initializer.

wschin:
Right. But user could potentially provide an input to overwrite constants like labels. I am not sure if it is ok.

bddppq:
@wschin Nope users are not supposed to provide an input to override initialzier either. https://github.com/onnx/onnx/commit/3e560dd7cb76006c15591c9180eedfd8d4c68587

PR #1052, opened 2018-05-30
Remove the usage of default value to fix invalid proto3 files.

linkerzhang:
This PR contains some format changes, as it's a very easy change so that I keep all format changes (suppose we should always run clang-format on each file before checking in so that format is always good).

Adding CI to cover this check will be done later. I'm figuring out how to do that, our linux ci protoc can only recognize proto2 and protoc can't be recognized in windows ci env :(.


onnx/common/constants.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <string>
+
+namespace ONNX_NAMESPACE {
+namespace Common {
+// For ONNX op/function registration.
+
+// ONNX domains.
+constexpr const char* AI_ONNX_ML_DOMAIN = "ai.onnx.ml";
+constexpr const char* ONNX_DOMAIN = "";
```
bddppq: Hmm, such move is a breaking change

onnx/defs/math/old.cc
```diff
@@ -5,7 +5,7 @@
 #include "onnx/defs/schema.h"
 
 namespace ONNX_NAMESPACE {
-
+using namespace Common;
```
houseroad: I think `using Common::ONNX_DOMAIN;` should be enough in most of the cases.

onnx/common/constants.h
linkerzhang: :). OK, removed the namespace Common.

onnx/defs/math/old.cc
linkerzhang: Common namespace is not used for now.

linkerzhang:
Comments resolved. Merging it.

PR #1053, opened 2018-05-31
Add index check for Transpose's type inference function

snnn:
@gramalingam Before passing an index to protobuf's repeated field, you must check if the index is valid first.  Otherwise, it's a security issue.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -296,8 +296,11 @@ ONNX_OPERATOR_SET_SCHEMA(
             if (!ctx.getInputType(0)->tensor_type().has_shape()) {
               return;
             }
-            const auto& splitDim =
-                ctx.getInputType(0)->tensor_type().shape().dim(axis);
+            const auto& shape = ctx.getInputType(0)->tensor_type().shape();
+            if (axis >= shape.dim_size()) {
+              fail_type_inference("Invalid value of attribute 'axis'");
+            }
+            const auto& splitDim = shape.dim(axis);
             if (!splitDim.has_dim_value()) {
               return;
             }
@@ -312,7 +315,7 @@ ONNX_OPERATOR_SET_SCHEMA(
 
             for (size_t i = 0; i < ctx.getNumOutputs(); i++) {
               *ctx.getOutputType(i)->mutable_tensor_type()->mutable_shape() =
-                  ctx.getInputType(0)->tensor_type().shape();
+                  shape;
               ctx.getOutputType(i)
                   ->mutable_tensor_type()
                   ->mutable_shape()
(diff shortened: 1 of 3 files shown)


onnx/defs/tensor/defs.cc
```diff
@@ -474,16 +477,37 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-                 --i) {
+          bool has_perm_attr = getRepeatedAttribute(ctx, "perm", perm);
+          if (!has_perm_attr) {
+            for (int i = shape.dim_size() - 1; i >= 0; --i) {
               perm.push_back(i);
             }
           }
+          else if(!perm.empty()){
+              for (int64_t fromDimIndex : perm)
+                  if (shape.dim_size() <= fromDimIndex || fromDimIndex >= std::numeric_limits<int>::max()) {
```
gramalingam: if (! ((0 <= fromDimIndex) && (fromDimIndex < shape.dim_size()))

snnn: if (fromDimIndex < 0 || shape.dim_size() <= fromDimIndex || fromDimIndex >= std::numeric_limits<int>::max())  ?

snnn: Thanks. Will change it.

onnx/defs/tensor/defs.cc
```diff
@@ -296,8 +296,11 @@ ONNX_OPERATOR_SET_SCHEMA(
             if (!ctx.getInputType(0)->tensor_type().has_shape()) {
               return;
             }
-            const auto& splitDim =
-                ctx.getInputType(0)->tensor_type().shape().dim(axis);
+            const auto& shape = ctx.getInputType(0)->tensor_type().shape();
+            if (axis >= shape.dim_size()) {
+                return;
```
bddppq: you should throw error here instead of silently return

onnx/defs/tensor/defs.cc
```diff
@@ -474,16 +477,38 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-                     ctx.getInputType(0)->tensor_type().shape().dim_size() - 1;
-                 i >= 0;
-                 --i) {
+          bool has_perm_attr = getRepeatedAttribute(ctx, "perm", perm);
+          if (!has_perm_attr) {
+            for (int i = shape.dim_size() - 1; i >= 0; --i) {
               perm.push_back(i);
             }
           }
+          else if(!perm.empty()){
```
bddppq: please run clang-format

onnx/defs/tensor/defs.cc
```diff
@@ -474,16 +477,38 @@ ONNX_OPERATOR_SET_SCHEMA(
           if (!hasNInputShapes(ctx, 1)) {
             return;
           }
-
+          auto input_type = ctx.getInputType(0);
+          if (!input_type->has_tensor_type()) {
```
bddppq: `hasNInputShapes` already checks `has_tensor_type`

onnx/defs/tensor/defs.cc
```diff
@@ -474,16 +477,38 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              throw InferenceError("Input 0 expected to have tensor type");
```
bddppq: use the macro `fail_shape_inference`

onnx/defs/tensor/defs.cc
```diff
@@ -474,16 +477,38 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                      std::ostringstream oss;
+                      oss << "Invalid attribute perm {" << perm[0];
+                      for (size_t i=1;i!=perm.size();++i) {
+                          oss << ", " << perm[i];
+                      }
+                      oss << "}, input shape = {";
+                      if (shape.dim_size() > 0) {
+                          oss<< shape.dim(0).dim_value();
+                          for (int i = 1; i != shape.dim_size(); ++i) {
+                              oss << ", " << shape.dim(i).dim_value();
```
bddppq: Why do you need to print the dim values? what user should know is `shape.dim_size()` right?

onnx/defs/tensor/defs.cc
snnn: Fixed.

onnx/defs/tensor/defs.cc
snnn: Fixed.

onnx/defs/tensor/defs.cc
snnn: Fixed.

onnx/defs/tensor/defs.cc
snnn: Fixed.

onnx/defs/tensor/defs.cc
snnn: It's not so obvious. Because they may be inferred from another node.  Currently, there is no way to view it.

gramalingam:
Hmmm … the Travis CI build keeps timing out … any idea what we can do?

PR #1054, opened 2018-05-31
Remove pytorch data

snnn:
They are not compatible with onnx 1.2. So they must be removed from onnx 1.2 release.


bddppq:
How are they "not compatible"?

snnn:
Hi @bddppq 
See #1055

test_PReLU_1d_multiparam\model.onnx is one example of that.
test_PReLU_1d_multiparam\model.onnx is using opset 6.

bddppq:
So it's the opset not being the latest, but they are still valid test data right? A backend is recommended to inspect the opset version, and then decide what's the corresponding semantics of this model at the specific opset version (of course a backend can raise not supported error when the opset is not supported). Even in the backend test suite you can skip the tests once it's an opset version that this backend doesn't support.

snnn:
It depends on how you judge it's valid or not.
From an end user's perspective, I'll read onnx 1.2's document, to get know how this OP(PRelu) works.  If it doesn’t work like the document says, it’s wrong. The document says it’s from OPSet 6, and the test data is in opset 6, and my backend supports opset 6.  Therefore, they should be compatible. This is how version check works. I wouldn’t bother to read all the history version of “Operators.md”.

Otherwise, this thing could be very complicated. Caffe2’s document would say: Oh, PRelu has two major versions: PRelu(new) and PRelu-1(old). For each major version, it has multiple subversions. We only support PRelu(new) for opset 6 and PRelu-1 for opset 1 but we don’t support PRelu(new) for opset 7 or opset 5. If so, why we need to distinguish them as “PRelu” and “PRelu-1”? Just “PRelu” is enough. Because each op can have different behavior in each opset.

bddppq:
Ok I think I have misunderstood your issue, I thought you were claiming that **all** opset 6 models are invalid because I saw you are deleting all pytorch generated tests.
PRelu's opset indeed should be bumped to opset 7. But I don't issues for other ops so far, should we instead only delete PRelu test data?

snnn:
Most of these data are fine, except 6: PReLU_1d_multiparam, PReLU_2d_multiparam, PReLU_3d_multiparam, operator_transpose, operator_type_as, operator_view

For PRelu related, I suggest upgrading PRelu version to 7
For operator_transpose and operator_type_as, please delete them
operator_view, I'm investigating

PR #1057, opened 2018-05-31
Convert schema assert for duplicate type names to exception

jeffbloo:
This converts a schema validation error from an assert to a normal SchemaError exception, which is more appropriate for callers consuming external schema.


bddppq:
@onnxbot retest this please

bddppq:
@onnxbot retest this please

PR #1060, opened 2018-06-01
add missing hasNInputShapes check for bidirectionalBroadcastShapeInference

gramalingam:


linkerzhang:
we may add test case to cover it.

PR #1063, opened 2018-06-01
bump version to 7 for prelu.

linkerzhang:


onnx/defs/operator_sets.h
```diff
@@ -443,6 +444,7 @@ class OpSet_Onnx_ver7 {
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(
            Onnx, 7, Multinomial)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 7, Xor)>());
+	fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 7, PRelu)>());
```
bddppq: nit: alignment

onnx/defs/math/defs.cc
```diff
@@ -424,7 +424,7 @@ output data (Tensor<T>) where the function `f(x) = slope * x for x < 0`,
 
 ONNX_OPERATOR_SET_SCHEMA(
     PRelu,
-    6,
+    7,
```
bddppq: nit: change  `PRelu_ver6_doc` to `PRelu_ver7_doc`

houseroad:
Thanks for fixing the problem.

onnx/defs/math/defs.cc
linkerzhang: fixed. thanks!

PR #1070, opened 2018-06-02
add input tensor dimension checks to shape inference

gramalingam:


onnx/defs/nn/defs.cc
```diff
@@ -334,6 +338,13 @@ void roiPoolTypeShapeInference(InferenceContext& ctx) {
   auto input_shape = ctx.getInputType(0)->tensor_type().shape();
   auto rios_shape = ctx.getInputType(1)->tensor_type().shape();
 
+  if (input_shape.dim_size() < 2) {
+	  fail_shape_inference("Input tensor must have atleast 2 dimensions");
```
linkerzhang: atleast -> at least. please also run clang-format for the file. Thanks!

PR #1075, opened 2018-06-04
fix Node::isBefore

fumihwh:


onnx/common/ir.h
```diff
@@ -1065,6 +1051,26 @@ inline void Node::eraseOutput(size_t i) {
   }
 }
 
+inline bool Node::isBefore(Node* n) {
```
anderspapitto: this seems to be getting longer/more complicated than it needs to be. Can't we just use

```
  // Check whether this node is before node n in the graph.
  bool isBefore(Node* n) {
    ONNX_ASSERT(n != nullptr);
    ONNX_ASSERT(n->inGraphList());
    for (Node * p = next(); p != nullptr; p = p->next()) {
      if (p == n) {
        return true;
      }
    }
    return false;
  }
```

fumihwh: @anderspapitto 
No..
https://github.com/onnx/onnx/blob/master/onnx/common/ir.h#L389
`this circular is a doubly-linked list`
Means `p->next()` will never get `nullptr`, it will go to the first node of graph after reach the last one.
So I use `*graph_->end()` instead.

If `kind_ == kParam || kind_ == kConstant`, we need return true because self is in initializer, it is before any nodes in graph.
As same reason, we need to return false if `n->kind() == kParam || n->kind() == kConstant`.

houseroad:
@onnxbot retest this please

fumihwh:
@houseroad 
Current node's Param will be triggered first. 
Return true means current Param is before target Param.
Although no order between them actually.

houseroad:
@fumihwh If they are both kParams, just return their order in the graph. Does this sound good to you? :-)

houseroad:
@fumihwh ok, then it does not make sense to return their order in the graph... thanks!

PR #1076, opened 2018-06-04
Spec clarity: Clarifying some of the ONNX-ML operators.

NiklasGustafsson:
This PR addresses the documentation of the ONNX-ML operators, mostly changing the descriptions to add little bit more detail and clarity.


onnx/defs/traditionalml/defs.cc
```diff
@@ -18,7 +18,7 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
         .Input(
             1,
             "Y",
-            "The index values to select as a int64 tensor",
+            "The indices, which must be at least one and no more than the number of dimensions of 'X.'",
```
linkerzhang: 'X.'  --> 'X'.

onnx/defs/traditionalml/defs.cc
```diff
@@ -28,10 +28,10 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
              "tensor(int64)",
              "tensor(int32)",
              "tensor(string)"},
-            "allowed types."));
```
linkerzhang: adding some description here is better.

onnx/defs/traditionalml/defs.cc
```diff
@@ -47,16 +47,16 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
              "tensor(double)",
              "tensor(int64)",
              "tensor(int32)"},
-            "allowed types.")
+            ".")
         .Attr(
             "threshold",
-            "Values greater than this are set to 1, else set to 0",
+            "Values greater than this are replaced by 1, others by 0.<br>The default threshold is 0.0.",
```
linkerzhang: 'Set’ is better than "Replace". The input tensor is NOT replaced. it's 0 or 1 in the output tensor.

onnx/defs/traditionalml/defs.cc
```diff
@@ -65,29 +65,29 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
     OpSchema()
         .SetDoc(CastMap_ver1_doc)
-        .Input(0, "X", "Data to be encoded", "T1")
-        .Output(0, "Y", "encoded output data", "T2")
+        .Input(0, "X", "The input map that is to be cast to a tensor", "T1")
+        .Output(0, "Y", "A tensor representing the same data as the input map, ordered by their keys", "T2")
         .TypeConstraint(
             "T1",
             {"map(int64, string)", "map(int64, float)"},
-            " allowed input types.")
```
linkerzhang: same as above.

onnx/defs/traditionalml/defs.cc
```diff
@@ -814,108 +854,106 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
 
 static const char* ZipMap_ver1_doc = R"DOC(
-    Makes a map from the input and the attributes.
-    Assumes input 0 are the values, and the keys are specified by the attributes.
-    Must provide keys in either classlabels_strings or classlabels_int64s (but not both).
-    Input 0 may have a batch size larger than 1,
-    but each input in the batch must be the size of the keys specified by the attributes.
-    The order of the input and attributes determines the key-value mapping.
+    Creates a map from the input and the attributes.<br>
+    The values are provides by the input tensor, while the keys are specified by the attributes.
```
linkerzhang: provides -> provided.

onnx/defs/traditionalml/defs.cc
```diff
@@ -814,108 +854,106 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
         .Attr(
             "nodes_treeids",
-            "tree id for this node",
+            "Tree id for each node.",
             AttributeProto::INTS,
             OPTIONAL)
         .Attr(
             "nodes_nodeids",
-            "node id for this node, node ids must restart at zero for each tree and increase sequentially.",
+            "Node id for each node. Ids ids must restart at zero for each tree and increase sequentially.",
```
linkerzhang: Ids ids -> node ids

onnx/defs/traditionalml/defs.cc
```diff
@@ -814,108 +854,106 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
         .Attr(
             "nodes_hitrates",
-            "popularity of the node, used for performance and may be omitted",
+            "Popularity of each node, used for performance and may be omitted.", 
             AttributeProto::FLOATS,
             OPTIONAL)
         .Attr(
             "nodes_modes",
-            "enum of behavior for this node as enum of BRANCH_LEQ, BRANCH_LT, BRANCH_GTE, BRANCH_GT, BRANCH_EQ, BRANCH_NEQ, LEAF",
+            "The node kind, that is, the comparison to make at the node. There is no comparison to make at a leaf node.<br>One of 'BRANCH_LEQ', 'BRANCH_LT', 'BRANCH_GTE', 'BRANCH_GT', 'BRANCH_EQ', 'BRANCH_NEQ', 'LEAF'",
```
linkerzhang: original description contains the enum strings. Keep them please so that readers will know what are the potential string values for the attribute.

onnx/defs/traditionalml/defs.cc
shauheen: >more than the number of dimensions of [](start = 60, length = 37)

Can this be further improved? no more than dimension of 'X'.?

onnx/defs/traditionalml/defs.cc
```diff
@@ -28,10 +28,10 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
              "tensor(int64)",
              "tensor(int32)",
              "tensor(string)"},
-            "allowed types."));
+            ""));
 
 static const char* Binarizer_ver1_doc = R"DOC(
-    Makes values 1 or 0 based on a single threshold.
+    Replaces the values of the input tensor by either 0 or 1, element-wise, based on the outcome of a comparison against a threshold value.
```
shauheen: >Replaces [](start = 4, length = 8)

I think @  is correct, in which case this should also be set.

onnx/defs/traditionalml/defs.cc
gramalingam: How about "are mapped to 1, others to 0" ?

onnx/defs/traditionalml/defs.cc
gramalingam: How about "Maps the values of the input tensor to either 0 or 1 …"?

onnx/defs/traditionalml/defs.cc
NiklasGustafsson: I agree. The one I found there was way to generic and didn't add any information. I'll go through and write something useful for each type constraint.

onnx/defs/traditionalml/defs.cc
```diff
@@ -65,29 +65,29 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+        .Input(0, "X", "The input map that is to be cast to a tensor", "T1")
+        .Output(0, "Y", "A tensor representing the same data as the input map, ordered by their keys", "T2")
         .TypeConstraint(
             "T1",
             {"map(int64, string)", "map(int64, float)"},
-            " allowed input types.")
+            "")
         .TypeConstraint(
             "T2",
             {"tensor(string)", "tensor(float)", "tensor(int64)"},
```
NiklasGustafsson: Is this correct? Where does the int64 tensor come from?

onnx/defs/traditionalml/defs.cc
NiklasGustafsson: @shauheen, I'm not sure I understand your comment. What are you referring to?

onnx/defs/traditionalml/defs.cc
NiklasGustafsson: I went with 'mapped'

onnx/defs/traditionalml/defs.cc
NiklasGustafsson: Thanks

onnx/defs/traditionalml/defs.cc
NiklasGustafsson: They are still there. The line got long, but the values are there.

onnx/defs/traditionalml/defs.cc
gramalingam: I assumed he meant change "Replace" to "set" …

onnx/defs/traditionalml/defs.cc
```diff
@@ -814,108 +854,106 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
         .Attr(
             "nodes_falsenodeids",
-            "child node if expression is false",
+            "Child node if expression is false",
             AttributeProto::INTS,
             OPTIONAL)
         .Attr(
             "nodes_missing_value_tracks_true",
-            "for each node, decide if the value is missing (nan) then use true branch, this field can be left unset and will assume false for all nodes",
+            "For each node, define what to do in the presence of a missing value: if a value is missing (NaN), use the 'true' or 'false' branch based on the value in this array.<br>This attribute may be left undefined, and the defalt value is false (0) for all nodes.",
```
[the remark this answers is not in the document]
NiklasGustafsson: I tried to clarify. Which branch to use actually depends on whether the array has a 0 or 1 for a give node.

onnx/defs/traditionalml/defs.cc
NiklasGustafsson: Fixed.

docs/Changelog-ml.md
```diff
@@ -33,12 +33,12 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
 <dl>
 <dt><tt>T</tt> : tensor(float), tensor(double), tensor(int64), tensor(int32), tensor(string)</dt>
-<dd></dd>
+<dd>The input must be a tensor of a numeric type or string. The output will be of the same tensor type.</dd>
 </dl>
 
 ### <a name="ai.onnx.ml.Binarizer-1"></a>**ai.onnx.ml.Binarizer-1**</a>
 
-  Replaces the values of the input tensor by either 0 or 1, element-wise, based on the outcome of a comparison against a threshold value.
+  Maps the values of the input tensor by either 0 or 1, element-wise, based on the outcome of a comparison against a threshold value.
```
gramalingam: change "by either" to "to either"?

docs/Operators-ml.md
```diff
@@ -537,38 +545,38 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
 <dt><tt>T</tt> : tensor(string), tensor(int64), tensor(int32), tensor(float), tensor(double)</dt>
-<dd> allowed types.</dd>
+<dd>The input must be a tensor of a numeric type.</dd>
 </dl>
 
 
 ### <a name="ai.onnx.ml.SVMClassifier"></a><a name="ai.onnx.ml.svmclassifier">**ai.onnx.ml.SVMClassifier**</a>
 
-  SVM classifier prediction
+  Support Vector Machine classifier
```
wschin: What are its input and output restrictions? I feel like we need to have a input/output format for all traditional classifiers and regressor somewhere. The format document may include something like:
- Input shape can be [C] or [N, C]
- For classifiers, if the input shape is [C], the output label/probabilities shape will be [1]/[L]. If the input shape becomes [N, C] the output labels/probabilities shape will be [N]/[N, L], where L is the number of total labels.
- For regressors with a scalar output, if the input shape is [C], the output shape will be [1]. If the input shape becomes [N, C] the output shape will be [N].
- For regressors with a vector output, if the input shape is [C], the output shape will be [L]. If the input shape becomes [N, C] the output shape will be [N, L], where L is the length of an example's output vector.

docs/Operators-ml.md
```diff
@@ -690,46 +698,44 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
 
 <dl>
 <dt><tt>offset</tt> : list of floats</dt>
-<dd>first, offset by this, must be same length as scale</dd>
+<dd>First, offset by this. Can be length of features or length 1, in which case it applies to all features</dd>
```
wschin: Does it mean that the input shape must be [N, C] (where C is the feature length)? If I have an input shape [N, C, H, W], this statement looks not very clear.

docs/Operators-ml.md
```diff
@@ -26,7 +26,7 @@
 ## ai.onnx.ml
 ### <a name="ai.onnx.ml.ArrayFeatureExtractor"></a><a name="ai.onnx.ml.arrayfeatureextractor">**ai.onnx.ml.ArrayFeatureExtractor**</a>
 
-  Select a subset of the data X based on the indices provided Y.
+  Select elements of the input tensor based on the indices passed.
```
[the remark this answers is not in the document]
NiklasGustafsson: Is it correct that AFE() only accepts 1- and 2-D tensors? I had not caught on to that. Do you have any reference to it that you can forward to me?

docs/Operators-ml.md
```diff
@@ -52,13 +52,13 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
 <dt><tt>T</tt> : tensor(float), tensor(double), tensor(int64), tensor(int32), tensor(string)</dt>
-<dd>allowed types.</dd>
+<dd>The input must be a tensor of a numeric type or string. The output will be of the same tensor type.</dd>
 </dl>
 
 
 ### <a name="ai.onnx.ml.Binarizer"></a><a name="ai.onnx.ml.binarizer">**ai.onnx.ml.Binarizer**</a>
 
-  Makes values 1 or 0 based on a single threshold.
+  Maps the values of the input tensor by either 0 or 1, element-wise, based on the outcome of a comparison against a threshold value.
```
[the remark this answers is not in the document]
NiklasGustafsson: It should accept any shape.

docs/Operators-ml.md
```diff
@@ -107,47 +107,47 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
-<dd>Data to be encoded</dd>
+<dd>The input map that is to be cast to a tensor</dd>
 </dl>
 
 #### Outputs
 
 <dl>
 <dt><tt>Y</tt> : T2</dt>
-<dd>encoded output data</dd>
+<dd>A tensor representing the same data as the input map, ordered by their keys</dd>
```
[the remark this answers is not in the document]
NiklasGustafsson: I believe it's [1,C].

docs/Operators-ml.md
```diff
@@ -177,28 +177,31 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
-<dd> allowed types.</dd>
+<dd>The output is a tensor of strings or integers.</dd>
 </dl>
 
 
 ### <a name="ai.onnx.ml.DictVectorizer"></a><a name="ai.onnx.ml.dictvectorizer">**ai.onnx.ml.DictVectorizer**</a>
 
-  Uses an index mapping to convert a dictionary to an array.
-      The output array will be equal in length to the index mapping vector parameter.
+  Uses an index mapping to convert a dictionary to an array.<br>
```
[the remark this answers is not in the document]
NiklasGustafsson: Done

docs/Operators-ml.md
```diff
@@ -387,52 +396,52 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
-<dd>enum NONE, SOFTMAX, LOGISTIC, SOFTMAX_ZERO, PROBIT</dd>
+<dd>Indicates the transform to apply to the scores vector.<br>One of 'NONE,' 'SOFTMAX,' 'LOGISTIC,' 'SOFTMAX_ZERO,' or 'PROBIT'</dd>
 </dl>
 
 #### Inputs
 
 <dl>
 <dt><tt>X</tt> : T1</dt>
-<dd>Data to be classified</dd>
+<dd>Data to be classified.</dd>
```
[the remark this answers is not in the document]
NiklasGustafsson: The implementation I have looked at source code for accepts 1-D tensors and that N defaults to 1 if there's just one dimension. Do you believe that is an error?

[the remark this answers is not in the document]
NiklasGustafsson: The implementation I have looked at source code for accepts 1-D tensors and that N defaults to 1 if there's just one dimension. Do you believe that is an error?

docs/Operators-ml.md
```diff
@@ -443,44 +452,46 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
 
 
 ### <a name="ai.onnx.ml.Normalizer"></a><a name="ai.onnx.ml.normalizer">**ai.onnx.ml.Normalizer**</a>
 
-  Normalize the input.  There are three normalization modes,
-      which have the corresponding formulas:
-      Max .. math::     max(x_i)
-      L1  .. math::  z = ||x||_1 = \sum_{i=1}^{n} |x_i|
-      L2  .. math::  z = ||x||_2 = \sqrt{\sum_{i=1}^{n} x_i^2}
+  Normalize the input.  There are three normalization modes, which have the corresponding formulas,
```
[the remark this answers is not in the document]
NiklasGustafsson: The implementation I have looked at allows [C]. Do you feel that doesn't make sense?

docs/Operators-ml.md
```diff
@@ -490,44 +501,41 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
-      This operator assumes every input in X is of the same category set
-      (meaning there is only one category count).
-  
-      If the input is a tensor of float, int32, or double, the data will be cast
-      to int64s and the cats_int64s category list will be used for the lookups.
+  Replace each input element with an array of ones and zeros, where a single
+      one is placed at the index of the category that was passed in. The total category count 
+      will determine the sized of the extra dimension of the output array Y.<br>
+      For example, if we pass a tensor with a single value of 4, and a category count of 8, 
+      the output will be a tensor with ``[0,0,0,0,1,0,0,0]``.<br>
```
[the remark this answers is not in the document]
NiklasGustafsson: This operator always adds another dimension to the end of the shape, with the size of the new dimension equal to the number of categories.

docs/Operators-ml.md
NiklasGustafsson: Right. Clarified to indicate special case for [N,F] situation.

docs/Operators-ml.md
wschin: I believe it'd be better to repeat it for each operator because the definitions of different operators are essentially independent. A shortcut is to define on an operator (for example, only specify those restrictions for linear classifier and regressor) and the refer it at other places.

docs/Operators-ml.md
wschin: Do you mean that Scaler only accepts [N,C]-inputs? If so, we need to say something like "This operator only consumes 2-D inputs" in its document string. If it can handle both of 1-D (aka [C]) and 2-D (aka [N, C]) inputs, we can add "This operator only consumes 1-D and 2-D inputs. The right-most axis is considered as the feature dimension."

docs/Operators-ml.md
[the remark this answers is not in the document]
wschin: Thanks. Could you add your illustration into the doc string?

docs/Operators-ml.md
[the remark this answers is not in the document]
wschin: It's ok to accept [C]. My point was that the following formulation cannot be applied to [C] and [N, C] at the same time. More specifically, the new formulations here are perfect for inputs with shape [C], but they are not mathematically meaningful when input becomes [N, C]. You probably want to say something like
  1. This operator accepts only [C] and [N, C]
  2. The following formulations are applied along C-axis. That is, different examples in one batch are normalized independently.

docs/Operators-ml.md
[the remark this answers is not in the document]
wschin: It's fine to accept 1-D input, but the spec needs to explicitly indicate what shapes can be consumed by this operator and how the corresponding output shapes are calculated.

docs/Operators-ml.md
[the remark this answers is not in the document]
wschin: If so, it's not consistent with DictVectorizer's output shape [C]. Anyway, it's not problem to have different behaviors for different operators. We just need to document the desired I/O shape down.

docs/Operators-ml.md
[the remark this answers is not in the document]
wschin: Yeah. Would you mind to add "This operator may accept a tensors with an arbitrary shape" to the end of this operator's introduction?

docs/Operators-ml.md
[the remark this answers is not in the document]
wschin: If it is not 1-D (i.e. shape [C]) or 2-D (i.e., shape [N, C]), which axis we should extract values along? Considering a similar ONNX operator, Slice, in ONNX, we can see that for arbitrary shape's extraction, an attribute `axes` is required. Also, you can take a look at our latest runtime code.

onnx/defs/traditionalml/defs.cc
```diff
@@ -18,7 +18,7 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
         .Input(
             1,
             "Y",
-            "The index values to select as a int64 tensor",
+            "The indices, which must be at least one and no more than the number of dimensions of 'X'.",
```
linkerzhang: no more than number of dimensions of 'X' is misleading. number of dimensions means "rank" of 'X".

The meaning here is, no more than the size of 'X'.

onnx/defs/traditionalml/defs.cc
```diff
@@ -168,13 +168,16 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
         }));
 
 static const char* DictVectorizer_ver1_doc = R"DOC(
-    Uses an index mapping to convert a dictionary to an array.
-    The output array will be equal in length to the index mapping vector parameter.
+    Uses an index mapping to convert a dictionary to an array.<br>
+    Given a dictionary, each key is looked up in the vocabulary attribute corresponding to
+    the key type. The index into the vocabulary array at which the key is found is then
+    used to index the output 1-D tensor 'Y' and insert into it the
+    the value found in the dictionary 'X'.<br>
```
linkerzhang: two “the”s, remove one.

docs/Operators-ml.md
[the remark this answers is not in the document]
NiklasGustafsson: I'll address it in an update.

docs/Operators-ml.md
```diff
@@ -326,15 +328,22 @@ This version of the operator has been available since version 1 of the 'ai.onnx.
...
+<dd>The input type must be a tensor of a numeric type. The output type will be of the same tensor type.</dd>
 </dl>
 
 
 ### <a name="ai.onnx.ml.LabelEncoder"></a><a name="ai.onnx.ml.labelencoder">**ai.onnx.ml.LabelEncoder**</a>
 
-  Convert class label to their integral type and vice versa.
-      In both cases the operator is instantiated with the list of class strings.
-      The integral value of the string is the index position in the list.
+  Converts strings to integers and vice versa.<br>
```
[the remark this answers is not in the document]
NiklasGustafsson: LabelEncode accepts tensors of all shapes.

PR #1077, opened 2018-06-04
Add retry logic to model downloading

zrphercule:
T29961396
Now model downloading will try 3 more times if download is failed.


onnx/backend/test/runner/__init__.py
```diff
@@ -155,6 +155,36 @@ def _assert_similar_outputs(ref_outputs, outputs):  # type: (Sequence[Any], Sequ
                 rtol=1e-3,
                 atol=1e-7)
 
+    def _retry_excute(times=3, func, *args, **kwargs):  # type: (int, function(), *args, **kwargs) -> None
+
+        for i in range(times):
+            try:
+                func(*args)
```
bddppq: kwargs

bddppq: better to return the return value of `func` to caller

```
try:
  return func(*args, **kwargs)
...
```

onnx/backend/test/runner/__init__.py
```diff
@@ -155,6 +155,36 @@ def _assert_similar_outputs(ref_outputs, outputs):  # type: (Sequence[Any], Sequ
...
+                return
+            except Exception as e:
+                remain = times-i-1
+                print('{} retry times remain'.format(remain))
```
bddppq: nit: better to print how many times you are retrying (instead of remaining)

onnx/backend/test/runner/__init__.py
```diff
@@ -155,6 +155,36 @@ def _assert_similar_outputs(ref_outputs, outputs):  # type: (Sequence[Any], Sequ
...
+                if remain == 0:
+                    raise
+
+    def _download_model(*args):
```
bddppq: no need to use args, better to use explicit arguments

onnx/backend/test/runner/__init__.py
```diff
@@ -175,17 +205,15 @@ def _prepare_model_data(self, model_test):  # type: (TestCase) -> Text
...
-                print('Start downloading model {} from {}'.format(
-                    model_test.model_name, model_test.url))
-                urlretrieve(model_test.url, download_file.name)
-                print('Done')
-                with tarfile.open(download_file.name) as t:
-                    t.extractall(models_dir)
-            except Exception as e:
-                print('Failed to prepare data for model {}: {}'.format(
-                    model_test.model_name, e))
+                _retry_excute(retry_times,
```
bddppq: self._retry_execute

onnx/backend/test/runner/__init__.py
```diff
@@ -175,17 +205,15 @@ def _prepare_model_data(self, model_test):  # type: (TestCase) -> Text
             # On Windows, NamedTemporaryFile can not be opened for a
             # second time
             download_file = tempfile.NamedTemporaryFile(delete=False)
```
bddppq: Let's put the creation and deletion of the tmp file into the _download_model? (I think it's ok to create tmp files couple times)

zrphercule: Sure, and move the "os.remove" into it as well

zrphercule:
@bddppq Looks like finally it passed all checks, could you please review it for me? Thanks!

onnx/backend/test/runner/__init__.py
```diff
@@ -172,23 +202,15 @@ def _prepare_model_data(self, model_test):  # type: (TestCase) -> Text
...
-                    t.extractall(models_dir)
-            except Exception as e:
-                print('Failed to prepare data for model {}: {}'.format(
-                    model_test.model_name, e))
+                self._retry_excute(self._download_model,
+                                retry_times,
+                                model_test=model_test,
+                                model_dir=model_dir,
+                                models_dir=models_dir)
+            except:
```
bddppq: don't need to try except here anymore

onnx/backend/test/runner/__init__.py
```diff
@@ -155,6 +155,36 @@ def _assert_similar_outputs(ref_outputs, outputs):  # type: (Sequence[Any], Sequ
                 rtol=1e-3,
                 atol=1e-7)
 
+    def _retry_excute(self, func, times=3, **kwargs):  # type: (Callable[..., Any], int, **Any) -> Any
+        for i in range(times):
+            try:
+                return func(**kwargs)
+            except:
```
bddppq: Do not except everything (this will catch unexpected events, e.g. user pressing ctrl-c). `except Exception` is more suitable.

onnx/backend/test/runner/__init__.py
```diff
@@ -155,6 +155,36 @@ def _assert_similar_outputs(ref_outputs, outputs):  # type: (Sequence[Any], Sequ
...
+                print('{} times tried'.format(i))
+                if i == times - 1:
+                    raise
+
+    def _download_model(self, model_test, model_dir, models_dir):  # type: (TestCase, Text, Text) -> None
+        try:
+            # On Windows, NamedTemporaryFile can not be opened for a
+            # second time
+            download_file = tempfile.NamedTemporaryFile(delete=False)
```
bddppq: in case `NamedTemoraryFile` throws, download_file will be not yet defined, and so in your finally cause, download_file will be an undefined variable.

onnx/backend/test/runner/__init__.py
```diff
@@ -155,6 +155,36 @@ def _assert_similar_outputs(ref_outputs, outputs):  # type: (Sequence[Any], Sequ
                 rtol=1e-3,
                 atol=1e-7)
 
+    def _retry_excute(self, func, times=3, **kwargs):  # type: (Callable[..., Any], int, **Any) -> Any
```
bddppq: you can still support both `*args` and `**kwargs`:
```python
def _retry_execute(self, func, *args, **kwargs):
   times = kwargs.pop("times", 3)
   ...
```
What I normally would do is to separate the special arguments in a decorator:
```python
def retry(times):
    assert times >= 1
    def wrapper(func):
        @functools.wraps(func)
        def wrapped(*args, **kwargs):
            ...
        return wrapped
    return wrapper
```
and of course then the call side becomes:

```python
@retry(3)
def download_model(...):
    ...

def _prepare_model_data(...):
    ...
    self.download_model(...)
    ...
```

zrphercule:
Just ignore these stupid commits I pushed hahaha... Finally learnt how to run Travis locally.
Thanks for your review and help Junjie!

bddppq:
lol travis just provided you a good chance to test this change: https://travis-ci.org/onnx/onnx/jobs/389062433

```
Start downloading model inception_v2 from https://s3.amazonaws.com/download.onnx/models/opset_7/inception_v2.tar.gz
Failed to prepare data for model inception_v2: [Errno socket error] [Errno 110] Connection timed out
0 times tried
Start downloading model inception_v2 from https://s3.amazonaws.com/download.onnx/models/opset_7/inception_v2.tar.gz
Failed to prepare data for model inception_v2: [Errno socket error] [Errno 110] Connection timed out
1 times tried
Start downloading model inception_v2 from https://s3.amazonaws.com/download.onnx/models/opset_7/inception_v2.tar.gz
Failed to prepare data for model inception_v2: [Errno socket error] [Errno 110] Connection timed out
2 times tried
```

This means:

1. this change is working correctly (retried three times)
2. we probably should add `time.sleep` between different tries

PR #1086, opened 2018-06-06
add string tensor type mapping

fumihwh:


bddppq:
hmm I intentionally didn't add the type mapping for string because it's not very clear when dealing with encodings https://github.com/onnx/onnx/blob/6bd04dd59ad7c39394df2a47023edeea4216bf72/onnx/numpy_helper.py#L37-L38

PR #1089, opened 2018-06-06
Add new operator "expand" to onnx 

zrphercule:
This PR will add a new operator "expand" to onnx.


onnx/defs/math/defs.cc
```diff
@@ -1046,4 +1046,28 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain input and output types to float tensors.")
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
+static const char* Expand_ver8_doc = R"DOC(
+Expand the input tensor following the given shape.
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+	Expand,
+	8,
```
bddppq: Need to bump the opset here https://github.com/onnx/onnx/blob/033f956f41c55fd409e1c4a0d09795ae5411447f/onnx/defs/schema.h#L588, since 7 has been frozen in release 1.2 already.

onnx/backend/test/case/node/expand.py
```diff
@@ -0,0 +1,36 @@
...
+    @staticmethod
+    def export():  # type: () -> None
+        node = onnx.helper.make_node(
+            'Expand',
+            inputs=['data', 'new_shape'],
+            outputs=['expanded'],
+        )
+
+        shape = [3, 1]
+        new_shape = [3, 4]
```
houseroad: In this case, we can also use `Tile`.

bddppq:
@houseroad We aim to use it as materialized broadcast.

houseroad:
@bddppq how about directly enhancing `Tile`?

bddppq:
@houseroad that requires some calculation to get the corresponding `repeats`, and this calculation is not easy to express with our existing operators (for dynamic shape). I think broadcast is common enough to worth it's own operator. This will help avoiding error-prone conversion logic based on static shapes when exporting from a dynamic frontend to onnx.

bddppq:
I would like this operator to mimic the full numpy broadcast semantics, e.g.
- It should support expanding to higher rank: `Expand([1.0], (2, 2)) => [[1.0, 1.0], [1.0, 1.0]]`
- It should support expanding in both directions: `Expand(ones(1, 2, 1), (2, 1, 2))` => ones(2, 2, 2)`

Basically I think it should be roughly equivalent to `Expand(lhs, shape) => lhs + zeros(shape)` with `+` being the broadcast add.

zrphercule:
@bddppq I wonder if you mean "Expand(ones(1, 2, 1), (2, -1, 2)) => ones(2, 2, 2)"? Or we need to ignore the second dim of shape (2, 1, 2) if it is smaller than the original shape?

bddppq:
@zrphercule No I meant `shape=(2, 1, 2)`. onnx has now fully aligned our broadcast semantics with numpy https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html

zrphercule:
@bddppq Then what about -1? Should we support -1 like pytorch?
https://pytorch.org/docs/master/tensors.html#torch.Tensor.expand

bddppq:
@zrphercule I think we shouldn't support that semantics (and we should be able to simply change all -1 to 1 when exporting from pytorch to onnx).

linkerzhang:
I don't quite understand this op right now. For some cases, Tile can handle, for some other cases, reshape + pad could handle. No?

bddppq:
@linkerzhang For **static shape**, and if the dims of one side is greater than the other side, Tile and Expand are convertible. However, it's super hard to express the numpy broadcast logic for dynamic shapes (because that means we can only use Operators to express it). Besides, this will help backends to adopt out new braodcast semantics because they can split a broadcast-add to first do materialize broadcast (aka. expand here) and then do non-broadcast-add.

houseroad:
@zrphercule could you also add some node tests for the new operator? Please follow [the doc](https://github.com/onnx/onnx/blob/master/docs/OnnxBackendTest.md)

zrphercule:
@houseroad Sure, I will submit some node test data and docs after I finished the caffe2 side of this operator.

onnx/backend/test/case/node/expand.py
```diff
@@ -0,0 +1,47 @@
...
+        #[[1.], [2.], [3.]]
+        expanded = np.tile(data, 4)
+        #print(expanded)
+        #[[1., 1., 1., 1.],
+        # [2., 2., 2., 2.],
+        # [3., 3., 3., 3.]]
+        new_shape = np.array(new_shape)
+        expect(node, inputs=[data, new_shape], outputs=[expanded],
+               name='test_expand_dim_unchanged')
+        new_shape = [2, 1, 6]
```
bddppq: move it to a second `export_*` function so it will appear as a separate code snippet in the doc.

onnx/defs/math/defs.cc
```diff
@@ -1046,4 +1046,27 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain input and output types to float tensors.")
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
+static const char* Expand_ver8_doc = R"DOC(
+Expand the input tensor following the given shape.
```
bddppq: This description is too ambiguous. User might think the `given shape` will be exactly the `output shape` (which is not).

onnx/defs/math/defs.cc
```diff
@@ -1046,4 +1046,27 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    Expand,
+    8,
+    OpSchema()
+        .SetDoc(Expand_ver8_doc)
+        .Input(0, "input", "Input tensor", "T")
+        .Input(1, "shape", "Shape of output tensor", "T")
```
bddppq: the description is incorrect

onnx/defs/math/defs.cc
```diff
@@ -1046,4 +1046,27 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Output(0, "output", "Output tensor", "T")
+        .TypeConstraint(
```
bddppq: all types should be supported (there is a variable defined in schema.h for representing the whole set of types).

onnx/defs/math/defs.cc
houseroad: Agree with @bddppq. This operator is not very popular (For example, in tensorflow and numpy, people use expand_dim + tile instead). I think that only with the vague description users cannot get the precise definition of the op. Please put more details and make sure the backend developers will know how to implement it after reading our description.

linkerzhang:
Per the two test cases, I'm seeing they should be exactly using "tile", No? if yes, may you guys give an example that tile can't be used to represent please?

bddppq:
@linkerzhang @zrphercule  Yes this is because NodeTests in onnx backend tests are more focus on isolated single node and therefore there is no way to express dynamic shape in it. Let's use simple model tests (like [here](https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/model/single-relu.py)) to add tests for dynamic shape:

```
Graph(
  input=['X', 'shape'],
  output=['Y'],
  initializers=[
    Tensor(
      name='X',
      shape=(1, 3, 1),
      floats=(...),
    )
  ],
  nodes=[
    Node(
      op_type='Expand',
      input=['X', 'shape'],
      output=['Y'],
    ),
  ]
)
```

and multiple `expect(...)` with different input `shape` to demo the different cases that expand supports (e.g. `[1, 3]`, `[3, 1, 3]`, `[3, 3, 2]`).

onnx/defs/math/defs.cc
```diff
@@ -1046,4 +1046,30 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain input and output types to float tensors.")
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
+static const char* Expand_ver8_doc = R"DOC(
+Broadcast the input tensor following the given shape and the broadcast rule.
+The broadcast rule is similar to numpy.array(input) * numpy.ones(shape):
```
houseroad: You can explicitly say that this operator is similar to numpy.broadcast_to, and describe the difference here.

zrphercule: sure :)

linkerzhang:
@bddppq Tile now supports the "repeats" as an input, which means it could be dynamic. Am I missing anything? Looks to me the two operators this "expand" and "tile", we only need one.

bddppq:
@linkerzhang Yes `repeats` is input and it can express dynamic values. But if onnx only has `tile`, this means when a frontend needs to export a model that has `expand` to onnx, it needs to do conversion from the expand's `shape` input to tile's `repeats` input, and when the shape is dynamic, the frontends needs to express all the conversion logic using existing onnx operators, instead of some python/c++ code, and this is pretty hard (or even impossible right now) for the `shape` => `repeats` conversion.
AFAIK both tf and pytorch have `expand` (also numpy).

linkerzhang:
@bddppq so I'm seeing that adding "expand" is for model conversion convenience. If this is the case, then we should either remove "tile" or have "expand" as a function using "tile". 

As ONNX is going to cover more models from other frameworks, I don't think that we'll have an operator union from all frameworks. This is also why we agree that several experimental ops were kept and not checked-in to ONNX standard.

Make sense please?

bddppq:
1. it's not just for convenience, as I have explained countless times in this thread, there are fundamental troubles when expressing the conversion logic between tile and expand via onnx operators
2. function is an experimental feature, and support (both tooling wise and framework wise) for it is just half done or even less
3. there needs to be a boundary on what should go through function and not. should we express BatchNormalization as a function? apparently not. Same reason for expand, it represents broadcast which is a fundamental operation in any DL framework, and this can help us avoid adding broadcast support to all of our operators.

zrphercule:
@bddppq @houseroad @linkerzhang So do you think we should land it?

zrphercule:
@linkerzhang Hi Ke, I wonder if you have anything else to add regard this diff? Or we can land it now?

bddppq:
Final ping @linkerzhang

linkerzhang:
I am sorry that I forgot to get back to this PR.

To me, "expand" and "tile" can be converted to each other, as "shape" is able to be converted to "repeats" (an issue there is "repeats" is int64, the "shape" may not be n times of input's dim value), or "repeats" is able to be converted to "shape" too (this is more straight forward to be converted without any issue). 

@bddppq  @zrphercule  do you want to retire "tile" or replace it with a function if folks still need it please?

Thank you!

btw, a composed operator "MeanVarianceNormalization" was changed to use Function as an example for your reference here, https://github.com/onnx/onnx/blob/master/onnx/defs/experiments/experiments_functions.cc

onnx/defs/math/defs.cc
```diff
@@ -1046,4 +1046,32 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(0, "input", "Input tensor", "T")
+        .Input(
+            1,
+            "shape",
+            "A 1-D tensor indicates the shape you want to expand to, following the broadcast rule",
+            "T")
+        .Output(0, "output", "Output tensor", "T")
+        .TypeConstraint(
+            "T",
+            OpSchema::all_tensor_types(),
```
linkerzhang: It should only work for numeric types, right? all_tensor_types contains "string" and "bool".

onnx/defs/math/defs.cc
```diff
@@ -1046,4 +1046,32 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    Expand,
+    8,
+    OpSchema()
+        .SetDoc(Expand_ver8_doc)
+        .Input(0, "input", "Input tensor", "T")
+        .Input(
+            1,
+            "shape",
+            "A 1-D tensor indicates the shape you want to expand to, following the broadcast rule",
+            "T")
```
bddppq: @zrphercule the second input is shape, so its type should be int64, and it doesn't need to be the same as input(0). use a different type variable here, and add a typeconstraint says this type variable can only be int64.

onnx/defs/math/defs.cc
```diff
@@ -1029,4 +1029,35 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    Expand,
+    8,
+    OpSchema()
+        .SetDoc(Expand_ver8_doc)
+        .Input(0, "input", "Input tensor", "T")
+        .Input(
+            1,
+            "shape",
+            "A 1-D tensor indicates the shape you want to expand to, following the broadcast rule",
+            "S")
```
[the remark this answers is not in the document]
bddppq: @linkerzhang fixed the type description. re. format, this chunk is actually already clang-formatted.

bddppq:
ping @linkerzhang

springishere:
@bddppq Looks like this commits added Expand operator. Can you add it to the export as well? Currently we can't export pytorch to onnx. From code here expand is not handled: https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/export.cpp#L47

onnx/backend/test/case/model/expand.py
```diff
@@ -0,0 +1,65 @@
...
+                nodes=[node],
+                name='Expand',
+                inputs=[onnx.helper.make_tensor_value_info('X',
+                                                             onnx.TensorProto.FLOAT,
+                                                             input_shape),
+                          onnx.helper.make_tensor_value_info('shape',
+                                                             onnx.TensorProto.INT64,
+                                                             shape_shape)],
+                outputs=[onnx.helper.make_tensor_value_info('Y',
+                                                              onnx.TensorProto.FLOAT,
```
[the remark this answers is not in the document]
zrphercule: You are right, I have submitted a pr for this:
https://github.com/onnx/onnx/pull/1283

Issue #1043, opened 2018-05-27
LpNormalization operator is missing epsilon as an attribute


anirudhacharya:
[LpNormalization](https://github.com/onnx/onnx/blob/master/docs/Operators.md#lpnormalization) operator, when ``p=2``, is defined as 
```
for i in 0...N
  out[i,:,:,...,:] = data[i,:,:,...,:] / sqrt(sum(data[i,:,:,...,:] ** 2) + eps)
```

the eps is added to prevent dividing by zero in edge cases. But the ONNX specification does not have this attribute.

houseroad:
I would propose to remove the op from ONNX spec. https://github.com/onnx/onnx/pull/1120

ShaunZheng:
hello,when i use the shufflenet,there is some erros,there is a "Attribute not found: epsilon",i hope to know how can i deel with it.

```
source_path: s3://jarvis-output-jarvis-public/zhengshuang_test_models/ONNX/shufflenet.tar.gz
target_path: /workspace/model/shufflenet.tar.gz

Converter::ctor 

Converter::ctor input graph contains 203 nodes, 282 inputs, 1 outputs, 281 initializers 

Converter:: inserting node gpu_0/conv3_0_1

Converter:: inserting node gpu_0/conv3_0_bn_1

Converter:: inserting node gpu_0/conv3_0_bn_2

Converter:: inserting node gpu_0/pool_0_1

Converter:: inserting node gpu_0/gconv1_0_1

Converter:: inserting node gpu_0/gconv1_0_bn_1

Converter:: inserting node gpu_0/gconv1_0_bn_2

Converter:: inserting node OC2_DUMMY_0

Converter:: inserting node OC2_DUMMY_2

Converter:: inserting node gpu_0/shuffle_0_1

Converter:: inserting node gpu_0/gconv3_0_1

Converter:: inserting node gpu_0/gconv3_0_bn_1

Converter:: inserting node gpu_0/gconv1_1_1

Converter:: inserting node gpu_0/gconv1_1_bn_1

Converter:: inserting node gpu_0/pool_1_1

Converter:: inserting node gpu_0/block0_1

Converter:: inserting node gpu_0/block0_2

Converter:: inserting node gpu_0/gconv1_2_1

Converter:: inserting node gpu_0/gconv1_2_bn_1

Converter:: inserting node gpu_0/gconv1_2_bn_2

Converter:: inserting node OC2_DUMMY_4

[...]

PR #1091, opened 2018-06-06
Add epsilon attribute to LpNormalization operator

anirudhacharya:
Issue - https://github.com/onnx/onnx/issues/1043

- [x] OP_SET version bump.
- [x] Test cases.


anirudhacharya:
sure I will add both.

linkerzhang:
Hmmm, given it's an optional attribute with default value specified, looks to me version needs not to be bumped. However, the default value is not 0, so models with using current version lpnormalization will have a slightly different result when running with backend using this updated op spec. Do we agree that this is still backward compatible? :) @ebarsoum @bddppq  @houseroad @ezyang @gramalingam

bddppq:
> However, the default value is not 0, so models with using current version lpnormalization will have a slightly different result when running with backend using this updated op spec.

This means backward incompatible :-)

onnx/defs/nn/defs.cc
```diff
@@ -1035,7 +1035,12 @@ ONNX_OPERATOR_SET_SCHEMA(
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to float tensors.")
         .SetDoc(LpNormalization_ver1_doc)
-        .Attr(
+		.Attr(
```
linkerzhang: run clang-format please

houseroad:
@anirudhacharya friendly ping for update :-)

houseroad:
Turn out the operator is a general version of `L2Normalization` in [TF](https://www.tensorflow.org/api_docs/python/tf/nn/l2_normalize). @anirudhacharya could you put more details to the description, so users can easily understand the semantics of this operator. Thanks!

anirudhacharya:
@houseroad sure I can make the changes. So do I take it that ReduceL1, ReduceL2 and LpNormalization are all going to stay?

If LpNormalization stays, do we restrict p value to 1 and 2? Should we restrict the axes along which L2Norm can be performed? I am inclined to yes for both these questions.

With regards to your question in your own PR, most frameworks have some form of both ReduceL1/2 and L2Normalization. MXNet, for example, has [L2Normalization](http://mxnet.incubator.apache.org/api/python/symbol/symbol.html#mxnet.symbol.L2Normalization) and [norm](http://mxnet.incubator.apache.org/api/python/symbol/symbol.html#mxnet.symbol.norm) which is very similar to ReduceL1/2.

houseroad:
@anirudhacharya yes, they will stay all, unless you want to make LpNormalization as a function op. :-)

For p, yes, restricting to 1 and 2 sounds good to me.
For axis, I think we don't need to set any restriction.

anirudhacharya:
The reason to restrict is that we can have a concrete definition for the operator and we can have test cases to verify the behavior of the operator. For instance how would we define the operator for a 6 dimensional tensor with p=3 and axis=4. Will we have tests for such cases?

anirudhacharya:
@houseroad ping for review. I still need to update the data files and generate the docs(currently running into errors while generating them. will fix it). but can you please take a look at the definition and the tests I have added?

onnx/backend/test/case/node/lpnormalization.py
```diff
@@ -0,0 +1,123 @@
...
+    def export_l2normalization_axis_0():  # type: () -> None
+        node = onnx.helper.make_node(
+            'LpNormalization',
+            inputs=['x'],
+            outputs=['y'],
+            epsilon = 1e-5,
+            axis = 0,
+            p = 2
+        )
+        x = np.random.randn(3, 4, 5)
```
houseroad: please set the type of x using astype(np.float32)

onnx/backend/test/case/node/lpnormalization.py
```diff
@@ -0,0 +1,123 @@
...
+    def export_l2normalization_axis_1():  # type: () -> None
+        node = onnx.helper.make_node(
+            'LpNormalization',
+            inputs=['x'],
+            outputs=['y'],
+            epsilon = 1e-5,
+            axis = 1,
+            p = 2
+        )
+        x = np.random.randn(3, 4, 5)
```
houseroad: please set the type of x using astype(np.float32)

onnx/backend/test/case/node/lpnormalization.py
```diff
@@ -0,0 +1,123 @@
...
+    def export_l2normalization_axis_last():  # type: () -> None
+        node = onnx.helper.make_node(
+            'LpNormalization',
+            inputs=['x'],
+            outputs=['y'],
+            epsilon = 1e-5,
+            axis = -1,
+            p = 2
+        )
+        x = np.random.randn(3, 4, 5)
```
houseroad: please set the type of x using astype(np.float32)

onnx/backend/test/case/node/lpnormalization.py
```diff
@@ -0,0 +1,123 @@
...
+    def export_l1normalization_axis_0():  # type: () -> None
+        node = onnx.helper.make_node(
+            'LpNormalization',
+            inputs=['x'],
+            outputs=['y'],
+            epsilon = 1e-5,
+            axis = 0,
+            p = 1
+        )
+        x = np.random.rand(3, 4, 5)
```
houseroad: please set the type of x using astype(np.float32)

onnx/defs/nn/defs.cc
```diff
@@ -1020,22 +1020,36 @@ ONNX_OPERATOR_SET_SCHEMA(
           propagateShapeAndTypeFromFirstInput(ctx);
         }));
 
-static const char* LpNormalization_ver1_doc = R"DOC(
+static const char* LpNormalization_ver7_doc = R"DOC(
 Given a matrix, apply Lp-normalization along the provided axis.
+
+For p=2,
+
+y = x  / (sqrt(sum(x ** 2)) + epsilon),
```
houseroad: Tensorflow defines L2Normalization like: `output = x / sqrt(max(sum(x**2), epsilon))`
@ebarsoum what's your thought?

houseroad:
@anirudhacharya friendly ping for update :-)

anirudhacharya:
@houseroad I addressed your comments. pls take a look. Regarding the op_set version bump I think @bddppq has a point. Why did you suggest that this does not require an op_set bump?

houseroad:
It's more like a bug fix instead of adding new features, and I don't think we have any existing models using this ops. Let's why I would suggest not bumping up the opset version.

houseroad:
@anirudhacharya could you rebase to master and pass the CI?

anirudhacharya:
@houseroad okay, will do.

anirudhacharya:
@houseroad sorry for the delay on this PR. Instead of resolving merge conflicts in this PR, I have created another PR with these changes here - https://github.com/onnx/onnx/pull/1330

The comments raised in this PR have been addressed there.

PR #1095, opened 2018-06-07
fix upper-bound for local-region in lrn test case

gramalingam:
@anirudhacharya there seems to be an off-by-one error in the LRN test-case also (or parenthesization error).


onnx/backend/test/case/node/lrn.py
```diff
@@ -54,7 +54,7 @@ def export_default():  # type: () -> None
         square_sum = np.zeros((5, 5, 5, 5)).astype(np.float32)
         for n, c, h, w in np.ndindex(x.shape):
             square_sum[n, c, h, w] = sum(x[n,
-                                           max(0, c - int(math.floor((nsize - 1) / 2))):min(4, c + int(math.ceil((nsize - 1) / 2)) + 1),
+                                           max(0, c - int(math.floor((nsize - 1) / 2))):min(5, c + int(math.ceil((nsize - 1) / 2)) + 1),
```
anirudhacharya: no I tested this on both numpy and with mxnet as backend. here 5 is the size of the array and 4 is the index. infact having 5 here will give ArrayIndexOutOfBoundsException.

gramalingam: As I understand, numpy indexing is upper-bound exclusive. You mentioned previously that the "test case has a +1 because numpy indexing is upper-bound exclusive." The +1 appears only in the second argument of "min". Shouldn't you either move the +1 outside the min or, equivalently, have a +1 for both arguments of "min"?

gramalingam: When I run this modified version, it works fine for me. If I use the original value of 4, I get incorrect results. Is it possible that you have different parenthesization in your mxnet version (with +1 outside min) than in the above file?

houseroad: I agree with @gramalingam, since the upper bound is exclusive, we should use C (i.e., 5) instead of C-1 in this case. I locally tested the script, it worked.

anirudhacharya: @gramalingam yes in my test example, I had the +1 outside the parentheses.

anirudhacharya:
the window size parameter for LRN is supposed to take only odd values. Could you please update the spec?

houseroad:
@anirudhacharya I think there is no need to forbid even values.

PR #1102, opened 2018-06-08
add float16 support back for math and reduction ops

linkerzhang:
float16 was supported for math and reduction ops before, and now add it back.

cublas support half op half -> half as https://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemm

### DIFF
--- a/onnx/defs/schema.h
+++ b/onnx/defs/schema.h
@@ -423,15 +423,16 @@ class OpSchema final {
   // Convenience members for types
 
   // All high-precision numeric types.
-  static const std::vector<std::string>& high_precision_numeric_types() {
-    static const std::vector<std::string> high_precision_numeric_types = {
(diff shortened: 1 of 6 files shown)


onnx/defs/schema.h
```diff
@@ -423,15 +423,16 @@ class OpSchema final {
...
   // All high-precision numeric types.
-  static const std::vector<std::string>& high_precision_numeric_types() {
-    static const std::vector<std::string> high_precision_numeric_types = {
+  static const std::vector<std::string>& numeric_types_for_math_reduction() {
+    static const std::vector<std::string> numeric_types_for_math_reduction = {
         "tensor(uint32)",
         "tensor(uint64)",
         "tensor(int32)",
         "tensor(int64)",
+		"tensor(float16)",
```
rdzhabarov: missing clang-format

Issue #1019, opened 2018-05-23
GRU equations/python


skottmckay:
Should R[zrh] be treated the same as W[zrh] values and be transposed before use? That would match the CNTK implementation and be a more consistent treatment of the W and R values. 

Current equations are
```
    zt = f(Xt*(Wz^T) + Ht-1*Rz + Wbz + Rbz)
    rt = f(Xt*(Wr^T) + Ht-1*Rr + Wbr + Rbr)
    ht = g(Xt*(Wh^T) + (rt (.) Ht-1)*Rh + Rbh + Wbh) # default, when linear_before_reset = 0
    ht = g(Xt*(Wh^T) + (rt (.) (Ht-1*Rh + Rbh) + Wbh) # when linear_before_reset != 0
    Ht = (1 - zt) (.) ht + zt (.) Ht-1
```

current python from GRU_helper.py
```
  z = self.f(np.dot(self.X, np.transpose(w_z)) + np.dot(self.H_0, r_z) + w_bz + r_bz)
  r = self.f(np.dot(self.X, np.transpose(w_r)) + np.dot(self.H_0, r_r) + w_br + r_br)
  h_default = self.g(np.dot(self.X, np.transpose(w_h)) + np.dot(r * self.H_0, r_h) + w_bh + r_bh)
  h_linear = self.g(np.dot(self.X, np.transpose(w_h)) + r * (np.dot(self.H_0, r_h) + r_bh) + w_bh)
```

python with transpose of R[zrh] that produces same results as CNTK
```
  z = self.f(np.dot(row, np.transpose(w_z)) + np.dot(self.H_0, np.transpose(r_z)) + w_bz + r_bz)
  r = self.f(np.dot(row, np.transpose(w_r)) + np.dot(self.H_0, np.transpose(r_r)) + w_br + r_br)
  h_default = self.g(np.dot(row, np.transpose(w_h)) + np.dot(r * self.H_0, np.transpose(r_h)) + w_bh + r_bh)
  h_linear = self.g(np.dot(row, np.transpose(w_h)) + r * (np.dot(self.H_0, np.transpose(r_h)) + r_bh) + w_bh)
```

skottmckay:
FWIW The LSTM python transposes the R weights before calling np.dot with Ht-1:

https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/lstm.py#L71
[...]

PR #1103, opened 2018-06-09
Make RNN/LSTM/GRU treatment of recurrent weights consistent

skottmckay:
After this recent change https://github.com/onnx/onnx/commit/d2a46da13b0b0e9d3ed7a1da34e015b4bc8fe0de#diff-98b0f6b804ec8aebff60069760608d05

we are in a state where the python for the RNN operators does the following with recurrent weights:

  GRU: Rz and Rr are transposed but not Rh 
  LSTM: All R weights are transposed
  RNN: R weights are not transposed

I believe the R weights should all be treated consistently and transposed in all cases, which makes them consistent with how the W weights are treated.

This PR updates the equations in the spec to transpose the R weights, the GRU test to transpose Rh, and the RNN test to transpose R. 

See also: https://github.com/onnx/onnx/issues/1019


linkerzhang:
@bddppq @houseroad @lupesko please help to review this op spec clarification.

linkerzhang:
@skottmckay  for your reference, keep op history example as https://github.com/onnx/onnx/blob/master/onnx/defs/nn/old.cc.

skottmckay:
Could you clarify the request for me? Does this require a version number increment in order to keep the old information (the 'output_sequence' attribute was removed from RNN ops in V7 - https://github.com/onnx/onnx/commit/4642745765188bde07de434fa5413dc98d7f8c2c#diff-b5717083f7a2a8796673400e589552af), and if so where would that need to be done?

gramalingam:
If the specification of the op is changed, the old definition/specification is moved to the corresponding old.cc file (in the same folder), and the new definition/specification is given an incremented version-number. I am guessing that is what Ke is referring to.

skottmckay:
Just wanted to check that updating the version number was the desired update. 

This change just clarifies how the R weights are interpreted so things are consistent, and there are no changes to the attributes/inputs/outputs. Additionally we didn't change the version number when the test case python for some of the operators changed recently. Happy to make whatever change is needed.

gramalingam:
Ok … I hope @houseroad can answer that question. Changing the equations changes the specification of the operator,  which would suggest increasing the version number ...

houseroad:
@skottmckay even it is just clarification of R, it already changes the semantics of the operator's input. The RNN/GRU/LSTM support may already depend on this. @anderspapitto could you confirm this?
If so, we should bump up the opset_version, and add the old def in `old.cc`.

linkerzhang:
@houseroad  I just realized that this PR is clarifying semantics and fixing the documentation issue, so it should not be taken as a breaking change. Yep. please confirm whether fb's implementation is following the clarified documentation in this PR. Thank you!

linkerzhang:
@houseroad  @anderspapitto please confirm whether fb's implementation is following the clarified documentation in this PR. Thanks a lot!

houseroad:
Keep tracking https://github.com/pytorch/pytorch/issues/8544

PR #1106, opened 2018-06-11
Optimization pass to fuse batch normalization operator with convolution operator

varunjain99:
If relevant inputs are initializers and we are in testing (not training) mode, we can fuse these operators together


onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,206 @@
...
+
+namespace ONNX_NAMESPACE { namespace optimization {
+//TODO: Currently broken for complex values and float16
+struct FuseBNIntoConv final : public OptimizePass {
+  explicit FuseBNIntoConv()
+    : OptimizePass("fuse_bn_into_conv", API_TYPE::IR) {
+  }
+
+  template<typename T>
+  static void add_nums(void* x, const void* y) {
```
bddppq: move these to tensor.h, and define add, sub, mul, div on the Tensor class that applies the corresponding operation to the whole tensor.

onnx/common/ir.h
```diff
@@ -847,6 +848,15 @@ friend struct Value;
   const std::vector<std::string>& initializer_names() {
     return initializer_names_;
   }
+  const int64_t get_initializer_index(Value* v) {
```
bddppq: use `std::find`, which will returns you an iterator (similar to a pointer) to the corresponding initializer (or end() which indicates not found).

bddppq:
You have modified the `third_party/benchmark` submodule, which should be intentional.

onnx/common/ir.h
```diff
@@ -953,6 +963,43 @@ friend struct Value;
     return n;
   }
 
+  //Adds to graph initializer list, initializer names list, and as a graph input
+  //Also syncs the initializer name, tensor name, and value name
+  Value* addInitializerAndInput(Tensor &initializer) {
+    std::vector<Dimension> dim_sizes;
+    for (auto v : initializer.sizes()) {
```
yinghai: `const auto& v`?

onnx/common/tensor.h
```diff
@@ -139,6 +139,221 @@ struct Tensor final {
...
+    if (is_raw_data_)  {
+      T_ptr = (T*) malloc(raw_data_.size());
+      memcpy(T_ptr, raw_data_.c_str(), raw_data_.size());
+    } else {
+      T_ptr = (T*) &T_data_[0];
+    }
+    if (a.is_raw_data())  {
+      a_ptr = (const T*) a.raw().c_str();
+    } else {
+      a_ptr = (const T*) &a_T_data_[0];
```
yinghai: `a_T_data_.data()`?

onnx/common/tensor.h
```diff
@@ -139,6 +139,221 @@ struct Tensor final {
     has_name_ = true;
     name_ = std::move(name);
   }
+
+  bool is_raw_data() const {
+    return is_raw_data_;
+  }
+
+  template<typename T, int64_t block_size>
+  void bin_func(void (*f)(void*, const void*), Tensor& a, std::vector<T>& T_data_, std::vector<T>& a_T_data_) {
```
yinghai: nit: the naming of the parameters are not very intuitive.

onnx/common/tensor.h
```diff
@@ -139,6 +139,221 @@ struct Tensor final {
...
+    int64_t num_elements = 1;
+    for (auto i : sizes_) {
```
yinghai: `std::accumulate`?

onnx/common/tensor.h
yinghai: Also I suggest we use template class F instead of function pointer.

onnx/common/tensor.h
```diff
@@ -139,6 +139,221 @@ struct Tensor final {
...
+      num_elements *= i;
+    }
+    T* T_ptr;
+    const T* a_ptr;
+    if (is_raw_data_)  {
+      T_ptr = (T*) malloc(raw_data_.size());
```
yinghai: Explicit memory allocation here is not very safe. We can just we vector<T>.

onnx/common/tensor.h
```diff
@@ -139,6 +139,221 @@ struct Tensor final {
...
+      case ONNX_NAMESPACE::TensorProto_DataType_DOUBLE: {
+        scale_dim<double, 1>(s, double_data_);
+        break;
+      }
+      case ONNX_NAMESPACE::TensorProto_DataType_COMPLEX128: {
+        scale_dim<double, 2>(s, double_data_);
+        break;
+      }
+      default:
+        throw("Incompatible data type: FLOAT, COMPLEX64, FLOAT16, DOUBLE and COMPLEX128 supported");
```
yinghai: What exception type are we throwing here?

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,206 @@
...
+        graph.eraseInitializerAndInput(old_b_value);
+      }
+    } else {
+      Value* new_b_value = graph.addInitializerAndInput(b);
+      conv->addInput(new_b_value);
+    }
+  }
+
+  bool modify_conv(Node* conv, Node* bn, Graph& graph)  {
+    auto bn_inputs = bn->inputs();
```
yinghai: Shouldn't it be `const auto&`? We are incurring a copy here.

onnx/optimizer/passes/fuse_bn_into_conv.h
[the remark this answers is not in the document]
yinghai: Can't we just use something like `std::multiply<T>`?

onnx/common/ir.h
varunjain99: included this function because I need the tensor from a name; however initializers and initializers_name are separate lists.

varunjain99: perhaps i should instead define a function getInitializerByName?

onnx/common/tensor.h
varunjain99: fixed

onnx/common/tensor.h
varunjain99: changed void to T and moved template to private as the only purpose is to serve apply_binary_function

onnx/common/tensor.h
varunjain99: changed

onnx/common/ir.h
bddppq: Yeah you can just name it as getInitializer with `const std::string&` as input type

onnx/common/tensor.h
bddppq: Agree

varunjain99: changed to vector

onnx/common/tensor.h
```diff
@@ -57,7 +57,7 @@ struct Tensor final {
 
   template<typename T, int64_t block_size>
   void bin_func(void (*f)(T*, const T*), Tensor& a, std::vector<T>& T_data_, std::vector<T>& a_T_data_) {
-    int64_t num_elements = std::accumulate(sizes_.begin(), sizes_.end(), 1, std::multiplies<int64_t>());
+    int64_t num_elements = std::accumulate(sizes_.begin(), sizes_.end(), (int64_t) 1, std::multiplies<int64_t>());
```
yinghai: `1L` also works. :)

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,174 @@
...
+          s.set_raw_data(string_rep);
+        }
+        W.scale_by_first_dim(s);
+        bc.subtract(m);
+        bc.multiply(s);
+        bc.add(bbn);
+        break;
+      }
+      default:
+        throw("Incompatible data type");
```
yinghai: @bddppq Do we have Exception types in ONNX?

onnx/common/tensor.h
```diff
@@ -27,6 +29,115 @@ struct Tensor final {
...
+  }
+
+  template<typename T>
+  static void sqrt_num(T* x) {
+    *x = (T) std::sqrt((double) *x);
+  }
+
+
+  template<typename T, int64_t block_size>
+  void bin_func(void (*f)(T*, const T*), Tensor& a, std::vector<T>& T_data_, std::vector<T>& a_T_data_) {
```
yinghai: Can we have
```
template<typename T, typename F, int64_t block_size>
void BinFunc(F&& f, Tensor* a, std::vector<T>* T_data, std::vector<T>* a_T_data)
```
as signature? First, it conforms to the Google code style better. Second, it avoids passing by function pointers. Simple functions like binary ops should really be inlined.

varunjain99: This would get rid of type checks on f right? Is there a need for that?

onnx/test/optimizer_test.py
```diff
@@ -707,6 +707,97 @@ def test_lift_lex_if(self):  # type: () -> None
         assert optimized_model.graph.node[2].attribute[2].strings[0] == b"X"
         assert optimized_model.graph.node[2].attribute[2].strings[1] == b"Y"
 
+    def test_fuse_bn_into_conv_simple_float(self):  # type: () -> None
```
houseroad: Two test cases are so similar, why not merge them together.

onnx/common/tensor.h
yinghai: Compiler will do type check and throw error if substitution doesn't work. Making a function call to each of the element in a tensor is potentially costly compared to inlined.

onnx/optimizer/passes/fuse_bn_into_conv.h
bddppq: Here shouldn't throw exceptions, because it's just data type that this optimization pass doesn't support (yet), not that the model is invalid.

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,174 @@
...
+    return true;
+  }
+
+  void fuse_bn_into_conv(Graph& graph) {
+
+    for (auto it = graph.begin(); it != graph.end(); ++it) {
+      auto* n = *it;
+      DescendOnGraphAttributes(n, [this](Graph& g){fuse_bn_into_conv(g);});
+      if (n->kind() == kBatchNormalization && n->inputs()[0]->node()->kind() == kConv) {
+        auto origInput = n->inputs()[0];
```
bddppq: Create two local variables `conv` and `bn` to refer to the corresponding nodes (aka. `n->inputs()[0]->node()` and n) will make the rest of this function more readable.

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,174 @@
...
+        if (origInput->uses().size() > 1 ||
+            n->outputs().size() > 1 ||
+            !modify_conv(origInput->node(), n, graph)) {
+          continue;
+        }
+        for (int i = 1; i <= 4; i++)  {
+          if (n->inputs()[i]->uses().size() == 0) {
+            graph.eraseInitializerAndInput(n->inputs()[i]);
```
bddppq: You need to check whether these inputs are really in initializers.

onnx/common/tensor.h
```diff
@@ -139,6 +250,382 @@ struct Tensor final {
...
+  //TODO: Support for FLOAT16
+  void sqrt();
+
+  //Element wise scaling of tensor s
+  //s is one dimensional, has size M, where M is size of first dimension of tensor
+  //s must have data as raw_data and has data type corresponding to this
+  void scale_by_first_dim(Tensor& s);
 };
 
+inline void Tensor::add(Tensor& a) {
```
bddppq: a should be const

onnx/common/tensor.h
bddppq: I think you only need `a_T_data` (or `a`) here? Also they should be const

onnx/common/tensor.h
```diff
@@ -27,6 +29,115 @@ struct Tensor final {
...
+    std::vector<T> vals;
+    if (is_raw_data_)  {
+      for (int64_t i = 0; i < (int64_t) raw_data_.size(); i += (int64_t) sizeof(T))  {
+        vals.push_back(*((const T*)(raw_data_.c_str() + i)));
+      }
+      T_ptr = (T*) vals.data();
+    } else {
+      T_ptr = (T*) T_data_.data();
+    }
+    for (int64_t i = 0; i < num_elements; i++) {
```
bddppq: ++i

onnx/common/tensor.h
```diff
@@ -27,6 +29,115 @@ struct Tensor final {
   bool is_raw_data_;
   std::string raw_data_;
 
+  template<typename T>
+  static void add_nums(T* x, const T* y) {
```
bddppq: There are std::plus, std::minus, std::multiplies and std::divides

onnx/common/tensor.h
```diff
@@ -139,6 +250,382 @@ struct Tensor final {
...
   }
+
+  bool is_raw_data() const {
+    return is_raw_data_;
+  }
+
+  //applies function f element-wise to this, storing result in this
+  //type T must correspond to appropriate tensor type
+
+  void apply_unary_function(void (*f)(float*))  {
```
bddppq: Looks to me these functions are quite similar, maybe better to use macros to reduce copy-paste like here https://github.com/onnx/onnx/blob/master/onnx/defs/schema.cc#L353-L379?

onnx/common/tensor.h
varunjain99: changed

onnx/common/tensor.h
varunjain99: made const, but you need both because you don't know where to look for the data if the tensor is stored as raw data - unless you do another switch on the types

onnx/common/tensor.h
varunjain99: changed all to const

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: this is checked in modify_conv - if not initializers then continue; so this function is not run

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: renamed

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: changed handling of different data types so that inputs have the same tensor type and optimization is not run if not float or double

onnx/common/tensor.h
varunjain99: Thought about it, but the functions may differ slightly in implementation for FLOAT16, COMPLEX64, COMPLEX128. I'm not too familiar with schema macros though so feel free to suggest if refactoring with macros is still plausible

onnx/common/tensor.h
varunjain99: changed to use

onnx/common/tensor.h
varunjain99: changed template structure to reflect above

onnx/common/tensor.h
```diff
@@ -139,6 +251,417 @@ struct Tensor final {
...
+      bin_func<F, double, 2>(f, a, double_data_, a.doubles());
+      break;
+    }
+    default:
+      throw("Operation not supported for this data type");
+  }
+}
+
+
+inline void Tensor::add(const Tensor& a) {
```
bddppq: The type switch in Tensor::add, Tensor::subtract, Tensor::multiply, Tensor::divide are pretty similar, use macros to reduce duplicate.

onnx/common/tensor.h
bddppq: I guess these `*_nums` are not used anymore so they can be deleted?

onnx/common/tensor.h
```diff
@@ -27,6 +30,115 @@ struct Tensor final {
...
+    }
+  }
+
+  template<typename F, typename T, int64_t block_size>
+  void un_func(F f, std::vector<T>& T_data_)  {
+    int64_t num_elements = std::accumulate(sizes_.begin(), sizes_.end(), (int64_t) 1, std::multiplies<int64_t>());
+    T* T_ptr;
+    std::vector<T> vals;
+    if (is_raw_data_)  {
+      for (int64_t i = 0; i < (int64_t) raw_data_.size(); i += (int64_t) sizeof(T))  {
```
bddppq: no need to cast `.size()` and `sizeof`

onnx/common/ir.h
```diff
@@ -847,6 +848,14 @@ friend struct Value;
   const std::vector<std::string>& initializer_names() {
     return initializer_names_;
   }
+  std::vector<Tensor>::iterator getInitializer(const std::string& name) {
+    for (auto it = initializers_.begin(); it != initializers_.end(); it++) {
+      if (name.compare(it->name()) == 0) {
```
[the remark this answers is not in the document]
varunjain99: changed

onnx/common/ir.h
```diff
@@ -953,6 +962,43 @@ friend struct Value;
     return n;
   }
 
+  //Adds to graph initializer list, initializer names list, and as a graph input
+  //Also syncs the initializer name, tensor name, and value name
+  Value* addInitializerAndInput(Tensor &initializer) {
+    std::vector<Dimension> dim_sizes;
```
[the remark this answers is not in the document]
varunjain99: changed

onnx/common/ir.h
```diff
@@ -953,6 +962,43 @@ friend struct Value;
     return n;
   }
 
+  //Adds to graph initializer list, initializer names list, and as a graph input
+  //Also syncs the initializer name, tensor name, and value name
+  Value* addInitializerAndInput(Tensor &initializer) {
```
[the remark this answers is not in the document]
varunjain99: modified

onnx/common/ir.h
```diff
@@ -953,6 +962,43 @@ friend struct Value;
...
+    }
+    Value* new_init = addInput();
+    initializer.setName(new_init->uniqueName());
+    new_init->setSizes(dim_sizes);
+    new_init->setElemType(initializer.elem_type());
+    addInitializer(initializer, new_init->uniqueName());
+    return new_init;
+  }
+
+  Value* addInitializerAndInput(Tensor &initializer, std::string name) {
```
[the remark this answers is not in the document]
varunjain99: changed

onnx/common/ir.h
```diff
@@ -847,6 +848,14 @@ friend struct Value;
   const std::vector<std::string>& initializer_names() {
     return initializer_names_;
   }
+  std::vector<Tensor>::iterator getInitializer(const std::string& name) {
+    for (auto it = initializers_.begin(); it != initializers_.end(); it++) {
```
bddppq: `const auto`, `cbegin`, `cend`

onnx/test/optimizer_test.py
bddppq: Agree, should merge to once test function and use a for loop like:

```
for dtype in [TensorProto.FLOAT, TensorProto.DOUBLE]:
  ...
```

onnx/test/optimizer_test.py
```diff
@@ -707,6 +707,97 @@ def test_lift_lex_if(self):  # type: () -> None
...
+                                vals=np.random.randn(3,).astype(np.float32).tobytes(),
+                                raw=True),
+            helper.make_tensor("var", TensorProto.FLOAT,
+                                dims=(3,),
+                                vals=np.random.randn(3,).astype(np.float32).tobytes(),
+                                raw=True)],
+        )
+        optimized_model = self._optimized(graph, ["fuse_bn_into_conv"])
+
+        assert len(list(optimized_model.graph.node)) == 1
```
bddppq: use `self.assertEqual`, which will print the mismatched values in case the test failed.

onnx/common/tensor.h
```diff
@@ -139,6 +258,295 @@ struct Tensor final {
...
+    case ONNX_NAMESPACE::TensorProto_DataType_COMPLEX128:  {
+      bin_func<F, double, 2>(f, a, double_data_, a.doubles());
+      break;
+    }
+    default:
+      throw("Operation not supported for this data type");
+  }
+}
+
+inline void Tensor::add(const Tensor& a) {
```
bddppq: put the function signature into the macro as well, right now the macro is assuming the input variable is named `a`.

bddppq: put the macro right above these functions, also one good convention to avoid macros name clashing is undef right after all usage:
```
#define my_macro
  ...

my_macro(add, std::plus)
my_macro(subtract, std::minus)

#undef my_macro

```

onnx/common/tensor.h
```diff
@@ -4,6 +4,42 @@
 #pragma once
 
 #include "onnx/onnx_pb.h"
+#include <math.h>
+#include <numeric>
+#include <functional>
+
+#define APPLY_FUNCTION(f)  {                                  \
```
bddppq: looks to me `apply_binary_function_*` functions are doing similar type switch again to determine the underlying field and forward to `bin_func`, can we collapse those into here and remove another level of indirection?

onnx/common/ir.h
varunjain99: changed

onnx/common/tensor.h
varunjain99: compressed

onnx/common/tensor.h
varunjain99: organized

varunjain99: combined signature

onnx/test/optimizer_test.py
varunjain99: added loop and assertEqual

bddppq:
@onnxbot retest this please

bddppq:
@onnxbot retest this please

onnx/common/tensor.h
```diff
@@ -139,6 +151,219 @@ struct Tensor final {
...
+      case ONNX_NAMESPACE::TensorProto_DataType_UINT64:  {                     \
+        CALL_BIN_FUNC(uint64_t, uint64s, f)                                    \
+        break;                                                                 \
+      }                                                                        \
+      case ONNX_NAMESPACE::TensorProto_DataType_DOUBLE: {                      \
+        CALL_BIN_FUNC(double, doubles, f)                                      \
+        break;                                                                 \
+      }                                                                        \
+      default:                                                                 \
+        throw("Operation not supported for this data type");                   \
```
bddppq: put elem_type_ into the error message to make it clear

onnx/common/tensor.h
```diff
@@ -139,6 +151,219 @@ struct Tensor final {
...
+      un_func<float (*)(float), float>(static_cast<float (*)(float)>(std::sqrt), this_data_ptr);
+      break;
+    }
+    case ONNX_NAMESPACE::TensorProto_DataType_DOUBLE: {
+      DATA(this, double, doubles)
+      un_func<double (*)(double), double>(static_cast<double (*)(double)>(std::sqrt), this_data_ptr);
+      break;
+    }
+    default:
+      throw("Operation not supported for this data type");
```
bddppq: ditto

onnx/common/tensor.h
```diff
@@ -139,6 +151,219 @@ struct Tensor final {
...
+  CONST_DATA(a, type, vec)                                                     \
+  bin_func<f<type>, type>(f<type>(), this_data_ptr, a_const_data_ptr);         \
+
+
+
+#define APPLY_BINARY_FUNCTION(op_name, f)                                      \
+  inline void Tensor::op_name(const Tensor& a_tensor) {                        \
+    const Tensor* a = &a_tensor;                                               \
+    if (a->elem_type() != elem_type_) {                                        \
+      throw("Type of tensors do not match");                                   \
```
bddppq: create a dedicated exception subclass for optimizer like here https://github.com/onnx/onnx/blob/master/onnx/checker.h#L13-L33 and throw that instead of general exception

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,212 @@
...
+          std::string string_rep(ptr, ptr + sizeof(float) * s.floats().size());
+          s.set_raw_data(string_rep);
+        }
+        W.scale_by_first_dim(s);
+        bc.subtract(m);
+        bc.multiply(s);
+        bc.add(bbn);
+        break;
+      }
+      case ONNX_NAMESPACE::TensorProto_DataType_DOUBLE:
```
bddppq: these two cases are almost identical, use another macro to avoid duplication?

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,212 @@
...
+          ONNX_ASSERT(bc.sizes().size() == 1 && bc.sizes()[0] == s.sizes()[0]);
+        }
+        var.add(eps);
+        var.sqrt();
+        s.divide(var);
+
+        if (!s.is_raw_data()) {
+          const char * ptr = reinterpret_cast<const char *>(s.floats().data());
+          std::string string_rep(ptr, ptr + sizeof(float) * s.floats().size());
+          s.set_raw_data(string_rep);
```
bddppq: here why setting raw_data? s should have stored the values in floats field already?

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,212 @@
...
+    auto epsilon = bn->hasAttribute(kepsilon) ? bn->f(kepsilon) : 1e-5;
+    Tensor eps;
+    Tensor bc;
+
+    switch(s.elem_type()) {
+      case ONNX_NAMESPACE::TensorProto_DataType_FLOAT: {
+        eps.sizes().push_back(s.sizes()[0]);
+        eps.elem_type() = ONNX_NAMESPACE::TensorProto_DataType_FLOAT;
+        for (int64_t i = 0; i < eps.sizes()[0]; ++i)  {
+          eps.floats().push_back((float) epsilon);
```
bddppq: no need to explicit cast

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,212 @@
...
+        bc.add(bbn);
+        break;
+      }
+      case ONNX_NAMESPACE::TensorProto_DataType_DOUBLE:
+      {
+
+        eps.sizes().push_back(s.sizes()[0]);
+        eps.elem_type() = ONNX_NAMESPACE::TensorProto_DataType_DOUBLE;
+        for (int64_t i = 0; i < eps.sizes()[0]; ++i)  {
+          eps.doubles().push_back((double)epsilon);
```
bddppq: no need to explicit cast

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,212 @@
...
+        bc.sizes().push_back(s.sizes()[0]);
+        bc.elem_type() = ONNX_NAMESPACE::TensorProto_DataType_FLOAT;
+        for (int64_t i = 0; i < eps.sizes()[0]; ++i)  {
+          bc.floats().push_back(0.f);
+        }
+
+        if (conv_inputs.size() == 3) {
+          auto bc_iter = graph.getInitializer(conv_inputs[2]->uniqueName());
+          if (bc_iter == end_iter) {
+            return false;
```
bddppq: it's possible the original conv doesn't bias but the fused one has, so here should add a new bias to the fused conv instead of stop doing optimization

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,212 @@
...
+          eps.floats().push_back((float) epsilon);
+        }
+        bc.sizes().push_back(s.sizes()[0]);
+        bc.elem_type() = ONNX_NAMESPACE::TensorProto_DataType_FLOAT;
+        for (int64_t i = 0; i < eps.sizes()[0]; ++i)  {
+          bc.floats().push_back(0.f);
+        }
+
+        if (conv_inputs.size() == 3) {
+          auto bc_iter = graph.getInitializer(conv_inputs[2]->uniqueName());
```
bddppq: orig conv can have no bias, in that case here will be out of bound access

onnx/common/ir.h
```diff
@@ -847,6 +848,14 @@ friend struct Value;
   const std::vector<std::string>& initializer_names() {
     return initializer_names_;
   }
+  std::vector<Tensor>::iterator getInitializer(const std::string& name) {
+    for (auto it = initializers_.begin(); it != initializers_.end(); ++it) {
```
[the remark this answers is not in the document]
varunjain99: changed

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: residue of previous implementation - removed

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: This code is not stopping the optimization if the original conv doesn't have bias. It is checking if there are 3 inputs (i.e. input of a bias). If so, we get the initializer; if that input is not an initializer, then we cannot do the optimization as we do not know the value. If there aren't 3 inputs, we still have a bc tensor - defined above - and we replace it in the replace_inputs function

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: Same idea as above - in the case of no bias, this branch will not run as conv_inputs will have size 2 not 3

bddppq: oops you are right sorry I missed the line above

onnx/optimizer/passes/fuse_bn_into_conv.h
bddppq: In that case could you move this bias check to above (where you checked whether other inputs are initializers).

onnx/common/tensor.h
varunjain99: fixed

onnx/common/tensor.h
varunjain99: fixed

onnx/common/tensor.h
varunjain99: added

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: used macro

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: moved

onnx/common/tensor.h
```diff
@@ -139,6 +151,219 @@ struct Tensor final {
...
+  }                                                                            \
+
+
+
+template<typename F, typename T>
+inline void Tensor::bin_func(F f, T* ptr, const T* a_ptr) {
+  int64_t num_elements = std::accumulate(sizes_.begin(), sizes_.end(),
+                                        (int64_t) 1, std::multiplies<int64_t>());
+  for (int64_t i = 0; i < num_elements; ++i) {
+    ptr[i] = f((T) ptr[i], (T) a_ptr[i]);
```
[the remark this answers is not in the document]
varunjain99: get error on CI without cast

onnx/optimizer/passes/fuse_bn_into_conv.h
varunjain99: error on CI without cast

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -0,0 +1,179 @@
...
+      if (conv_inputs.size() != 3) {                                           \
+        bc.sizes().push_back(s.sizes()[0]);                                    \
+        bc.elem_type() = ONNX_NAMESPACE::TensorProto_DataType_##TENSOR_TYPE;   \
+        for (int64_t i = 0; i < eps.sizes()[0]; ++i)  {                        \
+          bc.vec().push_back(0.f);                                             \
+        }                                                                      \
+      }                                                                        \
+      var.add(eps);                                                            \
+      var.sqrt();                                                              \
+      s.divide(var);                                                           \
```
bddppq: You need to check `s` and `var` don't have other uses in rest of the graph before modifying them inplace

varunjain99: they don't because I'm copying the tensors into s and var (L99, L102) 
auto s = *s_iter;
auto var = *var_iter;

Issue #1096, opened 2018-06-07
Outdated installation instructions in Installation.md


kit1980:
Probably https://github.com/onnx/onnx/blob/master/docs/Installation.md should be removed, because it's outdated and the README.md already has installation instructions.

linkerzhang:
Agree. Do you want to try to send a PR to remove it please? :)

houseroad:
The PR is merged.

PR #1108, opened 2018-06-12
Remove obsolete installation doc.

kit1980:
#1096


houseroad:
Actually, we can keep the page, and put the new link `https://github.com/onnx/onnx/#installation` in the page.

PR #1110, opened 2018-06-12
Add stats on ONNX node tests

houseroad:
- [x] Add the scripts
- [x] Generate the results
- [x] Add the script to CI

Although right now the generated page is a little bit similar to the operator page, later more information will be added. The purpose of two pages are also quite different. So it's better to have a separate page for test coverage information instead of adding it into the existing page.

No need to run the test cases.

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -35,6 +35,7 @@ git diff --exit-code
 # check auto-gen files up-to-date
 python onnx/defs/gen_doc.py
 python onnx/gen_proto.py
+python onnx/backend/test/stat_coverage.py
 backend-test-tools generate-data
 git status
 git diff --exit-code
--- a/docs/TestCoverage-ml.md
+++ b/docs/TestCoverage-ml.md
@@ -19,23 +19,59 @@ Node tests have covered 0/0 (N/A) experimental operators.
 
 ## &#x1F494;No Cover Common Operators
 ### ArrayFeatureExtractor (call for test cases)
+
+
 ### Binarizer (call for test cases)
+
+
 ### CastMap (call for test cases)
+
+
 ### CategoryMapper (call for test cases)
+
+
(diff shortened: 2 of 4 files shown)


houseroad:
@anirudhacharya the purpose of this page is to automatically generate the stats on the test cases. So we can know what the quality of our test cases is. We discussed the idea long time ago (probably in Feb.), now I kick off the work. Currently, the stat page only shows the information based on the node test, later I plan to add more information to the page, such as the stats on model test, and the coverage on the attributes, etc (I have created an issue to track the progress). Adding the script to CI is to make sure that the page is always up-to-date, just like `Operators.md`. If a contributor forgot to update the page, the CI will yell.

I hope my explanation has addressed all your concerns. If not, please let me know.

anirudhacharya:
@houseroad so the script will eventually generate % of coverage for each operator based on all the node, model and pytorch model converted tests?

houseroad:
Yes, eventually we will get there. This is the first step. :-)

houseroad:
After merging this PR, we can refer to the generated doc in https://github.com/onnx/onnx/issues/426, instead of manually updating the progress.

houseroad:
@anirudhacharya please let me know if you have any concern or question. Thanks.

houseroad:
@bddppq yes, current model tests have already covered the model test and attributes. But it's not very straightforward to ask users to check the coverage report every time. Also, we can break the coverage to model level, for example, what kind of operators and attributes are covered by one specific model, this is not covered by the current coverage report. Again, this is only the first step. (I will leave the major part to my intern. :-)) Later, we will have more powerful model analyzer and we can add more useful information to the test coverage page without a real backend.

houseroad:
@bddppq I will figure out how to reuse the existing coverage report logic when dealing with the model coverage and overall coverage parts (Not reinventing the wheel).

PR #1112, opened 2018-06-12
Implement function retrieval APIs; Add documentation for functions

raymondxyang:
* Implement function retrieval python APIs on existing ABIs
* Add documentation generation scripts for functions
* Minor fixes on dependencies and configurations in CMake files


onnx/defs/function.h
```diff
@@ -57,27 +57,4 @@ class FunctionBuilderRegistry {
   static Common::Status function_builder_##counter##_status = \
       FunctionBuilderRegistry::OnnxInstance().Register(function_builder);
 
-// Example to register a function.
```
linkerzhang: please keep it here as an example.

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Facebook Inc. and Microsoft Corporation.
+// Licensed under the MIT license.
+
+#include "onnx/defs/function.h"
+using namespace ONNX_NAMESPACE;
+
+static Common::Status BuildFc(std::unique_ptr<FunctionProto>* func_proto) {
```
linkerzhang: Let's switch MeanVarianceNormalization to be a function. FC was removed/deprecated.

onnx/defs/schema.cc
```diff
@@ -191,6 +190,14 @@ void OpSchema::Verify(const NodeProto& node) const {
       fail_check("Unrecognized attribute: ", name);
     }
 
+    if (attr_proto.has_ref_attr_name()) {
```
linkerzhang: this if should be removed?

onnx/defs/function.cc
```diff
@@ -57,6 +57,16 @@ Status FunctionBuilderRegistry::GetFunctions(
     }
 
     CheckerContext ctx;
+    std::unordered_map<std::string, int> op_set;
+    auto version_range =
+        OpSchemaRegistry::DomainToVersionRange::Instance().Map().at(
+            func_builder.GetDomain());
+    op_set.insert({func_builder.GetDomain(),
+                   function_proto->since_version() < version_range.second
```
linkerzhang: if function's since version is greater than the version_range.second, it should be an error, I think.

onnx/defs/schema.cc
raymondxyang: The if logic here determine whether this is a solid attribute or a placeholder attribute. I extract the if- logic here for checks on placeholder attr_proto going under this condition. IMO it is a valid condition checker here.

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,98 @@
...
+  node0->add_input("X");
+  node0->add_output("X_RM");
+  AttributeProto* attr0 = node0->add_attribute();
+  attr0->set_ref_attr_name("axes");
+  attr0->set_name("axes");
+  attr0->set_type(AttributeProto_AttributeType_INTS);
+  NodeProto* node1 = func.add_node();
+  node1->set_name("Pow_0");
+  node1->set_domain("");
+  node1->set_doc_string("Caculating (EX)^2");
```
wschin: Calculating --> Calculate

docs/Functions.md
```diff
@@ -0,0 +1,58 @@
...
+            [def files](/onnx/defs) via [this script](/onnx/defs/gen_doc.py).
+            Do not modify directly and instead edit function definitions.*
+## ai.onnx (default)
+  * <sub>experimental</sub><a href="#MeanVarianceNormalization">MeanVarianceNormalization</a>
+
+
+
+### <sub>experimental</sub> <a name="MeanVarianceNormalization"></a><a name="meanvariancenormalization">**MeanVarianceNormalization**</a>
+
+  A MeanVarianceNormalization Function: Perform mean variance normalization on the input tensor X
```
wschin: Need detailed description and mathematical formulations. As a standard, a clear document is very important. The supported input shape is likely any dimension. Please describe at least [N, C, H, W] and N-D cases.

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  func_proto->reset(new FunctionProto);
+  auto& func = **func_proto;
+  func.set_name("MeanVarianceNormalization");
+  func.set_doc_string(
+      "A MeanVarianceNormalization Function: Perform mean variance normalization on the input tensor X");
+  func.set_since_version(8);
+  func.add_input("X");
+  func.add_input("Pow_exponent");
```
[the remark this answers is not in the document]
linkerzhang: Yes, function does not have initializer right now.

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,98 @@
...
+  NodeProto* node6 = func.add_node();
+  node6->set_name("SUB_1");
+  node6->set_domain("");
+  node6->set_doc_string("Caculating X-EX");
+  node6->set_op_type("Sub");
+  node6->add_input("X");
+  node6->add_input("X_RM");
+  node6->add_output("X_VAR");
+  NodeProto* node7 = func.add_node();
+  node7->set_name("DIV_0");
```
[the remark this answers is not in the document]
raymondxyang: I noticed this by looking at Caffe's code. However per https://github.com/onnx/onnx/blob/master/docs/Operators.md#div there is no attribute/input for Div op in ONNX. Do we need to manually add an add op before div?

linkerzhang:
Let's also add test case for this function, like test cases for other operators.

onnx/defs/function.cc
```diff
@@ -94,4 +104,32 @@ FunctionBuilderRegistry& FunctionBuilderRegistry::OnnxInstance() {
   return func_builder_registry;
 }
 
+Common::Status FunctionNewNodeHelper(
```
linkerzhang: This is a common nodeproto construction. You may change it with a general name, say, BuildNode or... Make sense?

onnx/defs/schema.cc
```diff
@@ -191,6 +190,14 @@ void OpSchema::Verify(const NodeProto& node) const {
       fail_check("Unrecognized attribute: ", name, " for operator ", node.op_type());
     }
 
+    if (attr_proto.has_ref_attr_name()) {
```
linkerzhang: Move this block to the "default” section in following switch block?

docs/Functions.md
```diff
@@ -0,0 +1,59 @@
...
+<dt>X_MVN; </dt>
+<br/></dl>
+
+#### Attributes
+
+<dl>
+<dt>axes;<br/></dt>
+</dl>
+
+#### Nodes
```
linkerzhang: Let's figure out a better way to show function body. 

We may add a netron picture or link in function description (we may ask Netron to be able to visualize a function).

onnx/backend/test/case/node/mvn.py
```diff
@@ -0,0 +1,45 @@
...
+from . import expect
+
+
+class MVN(Base):
+
+    @staticmethod
+    def export():  # type: () -> None
+        node = onnx.helper.make_node(
+            'FuncMeanVarianceNormalization',
+            inputs=['X'],
```
linkerzhang: no attributes needed?

docs/Functions.md
raymondxyang: I guess this can be next step as we prob need to explore the ways to do that

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,132 @@
...
+
+  func_proto->reset(new FunctionProto);
+  auto& func = **func_proto;
+  func.set_name("FuncMeanVarianceNormalization");
+  func.set_doc_string(
+      "A MeanVarianceNormalization Function: Perform mean variance normalization "
+      "on the input tensor X using formula: <br/> ``` (X-EX)/sqrt(E(X-EX)^2) ``` <br/>"
+      "<b>INPUT: </b>X(float/float16/double) with Shape [N,C,W,H] <br/>"
+      "<b>ATTRIBUTE: </b><br/>&nbsp;&nbsp;&nbsp;&nbsp;axes: will be passed to ReducedMean Ops. "
+      "Use [0,2,3] for no across channel, [0,1,2,3] for across channel caculation.<br/>"
```
wschin: for no across channel ---> for calculating means and variances along channels. Two variables with the same C-coordinate are associated with the same mean and variance.
for across channel ---> to calculate global mean and  global variance. In this case, all variables share the same mean/variance.

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,132 @@
...
+      "Pow_exponent_0",
+      "",
+      "Initialize a Constant tensor to caculate squared products",
+      "Constant",
+      std::vector<std::string>{},
+      std::vector<std::string>{});
+  AttributeProto* value_attr = initial_node0->add_attribute();
+  value_attr->set_name("value");
+  value_attr->set_doc_string(
+      "Value for the constant tensor as power exponent to caculate MVN");
```
wschin: A clearer description: Exponent (i.e., 2) to element-wisely calculate the square of a tensor

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,132 @@
...
+  NodeProto* initial_node0 = func.add_node();
+  BuildNode(
+      initial_node0,
+      "Pow_exponent_0",
+      "",
+      "Initialize a Constant tensor to caculate squared products",
+      "Constant",
+      std::vector<std::string>{},
+      std::vector<std::string>{});
+  AttributeProto* value_attr = initial_node0->add_attribute();
```
wschin: Not sure if creating attribute should be a part of BuildNode(…) like the python helper for making NodeProto.

onnx/defs/function.h
```diff
@@ -57,6 +57,18 @@ class FunctionBuilderRegistry {
   static Common::Status function_builder_##counter##_status = \
       FunctionBuilderRegistry::OnnxInstance().Register(function_builder);
 
+// Helper function for register nodes in
+// a FunctionProto. Attributes need to be
+// registered separately.
+Common::Status BuildNode(
```
wschin: This is similar to a function in helper.py. If C++ doesn't have such a helper module, we should create one. Otherwise, it'd be better to reuse it.

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,132 @@
...
+      "on the input tensor X using formula: <br/> ``` (X-EX)/sqrt(E(X-EX)^2) ``` <br/>"
+      "<b>INPUT: </b>X(float/float16/double) with Shape [N,C,W,H] <br/>"
+      "<b>ATTRIBUTE: </b><br/>&nbsp;&nbsp;&nbsp;&nbsp;axes: will be passed to ReducedMean Ops. "
+      "Use [0,2,3] for no across channel, [0,1,2,3] for across channel caculation.<br/>"
+      "&nbsp;&nbsp;&nbsp;&nbsp;(The KeepDims attribute in ReducedMean ops is set to true for caculation)<br/>"
+      "<b>OUTPUT: </b>X_MVN(float/float16/double) with Shape [N,C,W,H] <br/>");
+  func.set_since_version(8);
+  func.add_input("X");
+  func.add_output("X_MVN");
+  func.add_attribute("axes");
```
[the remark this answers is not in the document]
raymondxyang: I saw this in caffe's code but Div op in onnx does not take epsilon as an attribute..

onnx/defs/function.h
raymondxyang: Double checked and there is no node constructing functions in C++

onnx/defs/experiments/functions.cc
raymondxyang: I thought about it before but it is complicated as there can be two type of attributes: full attributeprotos and/or referencing attributeprotos. For the full attrs I cant find an universal set of parameters with varied required fields..

onnx/defs/experiments/functions.cc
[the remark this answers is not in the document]
wschin: You need to create an Add operator for this.

onnx/defs/function.h
linkerzhang: In this way, can we move this function to a new common file please? say, ModelHelper.h/cc in common folder.

onnx/defs/experiments/functions.cc
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) Facebook Inc. and Microsoft Corporation.
```
linkerzhang: let's rename this file to experiments_functions.cc. Otherwise, there will be many functions.cc (like defs.cc, which is not a good idea). It makes hard to find some definition codes. :) Thank you!

docs/Functions.md
```diff
@@ -0,0 +1,36 @@
...
+  A MeanVarianceNormalization Function: Perform mean variance normalization on the input tensor X using formula: <br/> ``` (X-EX)/sqrt(E(X-EX)^2) ``` <br/><br/><b>INPUT: </b>X(float/float16/double) with shape [N,C,W,H] or N-D shape <br/><br/><b>ATTRIBUTE: </b><br/>&nbsp;&nbsp;&nbsp;&nbsp;<tt>axes: </tt>will be passed to ReducedMean Ops. Use [0,2,3] (without C axis for N-D cases) for for calculating means and variances along channels. Two variables with the same C-coordinate are associated with the same mean and variance. Use [0,1,2,3] (with C axis) to calculate global mean and global variance with all variables sharing the same mean/variance.<br/>&nbsp;&nbsp;&nbsp;&nbsp;(The KeepDims attribute in ReducedMean is set to true for caculation)<br/><br/><b>OUTPUT: </b>X_MVN(float/float16/double) with shape [N,C,W,H] or the input N-D shape <br/>
+
+#### Version
+
+This version of the function has been available since version 8 of the default ONNX operator set.
+
+#### Inputs
+
+<dl>
+<dt>X; </dt>
```
linkerzhang: the type constraint info should also be inferred from its body and printed here. You may add it in separate PR and push this in firstly.

onnx/defs/experiments/functions.cc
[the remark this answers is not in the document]
raymondxyang: I tested with Eigen if we use the eps value in caffe, 1e-9, in some cases it showed no differences (0 for determinant and inf for inversed tensors).. what value shall we have in order to avoid it?

onnx/defs/model_helpers.h
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Facebook Inc. and Microsoft Corporation.
```
linkerzhang: put it into common folder.

docs/Functions.md
```diff
@@ -0,0 +1,36 @@
...
+## ai.onnx (default)
+  * <sub>experimental</sub><a href="#FuncMeanVarianceNormalization">FuncMeanVarianceNormalization</a>
+
+
+
+### <sub>experimental</sub> <a name="FuncMeanVarianceNormalization"></a><a name="funcmeanvariancenormalization">**FuncMeanVarianceNormalization**</a>
+
+  A MeanVarianceNormalization Function: Perform mean variance normalization on the input tensor X using formula: <br/> ``` (X-EX)/sqrt(E(X-EX)^2) ``` <br/><br/><b>INPUT: </b>X(float/float16/double) with shape [N,C,W,H] or N-D shape <br/><br/><b>ATTRIBUTE: </b><br/>&nbsp;&nbsp;&nbsp;&nbsp;<tt>axes: </tt>will be passed to ReducedMean Ops. Use [0,2,3] (without C axis for N-D cases) for for calculating means and variances along channels. Two variables with the same C-coordinate are associated with the same mean and variance. Use [0,1,2,3] (with C axis) to calculate global mean and global variance with all variables sharing the same mean/variance.<br/>&nbsp;&nbsp;&nbsp;&nbsp;(The KeepDims attribute in ReducedMean is set to true for caculation)<br/><br/><b>OUTPUT: </b>X_MVN(float/float16/double) with shape [N,C,W,H] or the input N-D shape <br/>
```
linkerzhang: caculation --> calculation.

onnx/defs/function.cc
```diff
@@ -57,6 +57,16 @@ Status FunctionBuilderRegistry::GetFunctions(
     }
 
     CheckerContext ctx;
+    std::unordered_map<std::string, int> op_set;
+    auto version_range =
+        OpSchemaRegistry::DomainToVersionRange::Instance().Map().at(
+            func_builder.GetDomain());
+    if (function_proto->since_version() > version_range.second) {
```
raymondxyang: also check the lower version.

CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ list(APPEND CMAKE_MODULE_PATH ${ONNX_ROOT}/cmake/Modules)
 if(NOT MSVC)
   set(CMAKE_C_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
   set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")
```
bddppq: why is this needed?

onnx/defs/schema.cc
```diff
@@ -191,6 +190,14 @@ void OpSchema::Verify(const NodeProto& node) const {
...
+      if (!attr_proto.has_type() || attr_proto.type() != expected_type) {
```
bddppq: looks to me this check is also applicable when the attr doesn't have ref_attr_name (however you will need to guard the has_type check by ir_version, IIRC type was added in ir_version 2

CMakeLists.txt
```diff
@@ -361,6 +363,7 @@ if (MSVC)
         /wd4800 # disable warning type' : forcing value to bool 'true' or 'false' (performance warning)
         /wd4503 # identifier' : decorated name length exceeded, name was truncated
         /wd4146 # unary minus operator applied to unsigned type, result still unsigned
+        /wd4244 # conversion from 'protobuf::int64' to 'int'
```
bddppq: where does such cast happen? do you also need to ignore this warning in other targets?

raymondxyang: This happens when I convert fields in functionproto in cpp2py.. I only notice the warning in onnx target..

onnx/defs/schema.cc
raymondxyang: Actually if it is a solid attr_proto if the check will be covered under checker.cc L199-208.. This validation logic is aimed to be only conducted on the referencing attr_proto.

linkerzhang:
Thank you!

CMakeLists.txt
bddppq: at least you would also need to ignore it in other platforms as well. maybe what's better is to find the place in pybind11 binding and fix it?

onnx/cpp2py_export.cc
```diff
@@ -96,6 +97,57 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
       .value("COMMON", OpSchema::SupportType::COMMON)
       .value("EXPERIMENTAL", OpSchema::SupportType::EXPERIMENTAL);
 
+  py::class_<FunctionProto> function_proto(defs, "FunctionProto");
```
bddppq: This is not good, because python land already has a `FunctionProto` class from the auto-generated protobuf python files (`onnx.FunctionProto`), and its underlying memory layout is totally different from the c++ protobuf structs, and so python land users will see two classes with the same name while they are completely different. what we have been doing when passing protobuf structs across c++ and python is serialize on one side and then deserialize on the other side (look at check_model)

raymondxyang: Currently the solution is blocked by #1194 (and seems the import issue is in discussion in protobuf repo as well). I could rename the generated pybind class (to function_schema probably) and explore a way to resolve #1194 (probably need to manually run text replacement scripts), switch to use function_proto later. This class is only used for the display of function information now so there is use case now for user to really create one using function_proto class in _pb2.py

raymondxyang: Found a way to manually correct the import in build. Pushed.

raymondxyang: Comment resolved

onnx/test/c++/function_get_test.cc
```diff
@@ -0,0 +1,19 @@
...
+#include "gtest/gtest.h"
+#include "onnx/defs/function.h"
+
+namespace ONNX_NAMESPACE {
+namespace Test {
+TEST(FunctionAPITest, Get_All_Functions) {
+  std::multimap<std::string, std::unique_ptr<FunctionProto>> temp_map;
+  FunctionBuilderRegistry& function_registry =
+      FunctionBuilderRegistry::OnnxInstance();
+  Common::Status status = function_registry.GetFunctions("", &temp_map);
```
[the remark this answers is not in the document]
raymondxyang: fixed. Thanks for pointing out 👍

onnx/defs/gen_doc.py
```diff
@@ -9,13 +9,16 @@
...
-from onnx.defs import OpSchema
+from onnx.defs import OpSchema, FunctionProto
 from onnx.backend.test.case import collect_snippets
 from typing import Text, Sequence, Dict, List, Type, Set, Tuple
 
 
 SNIPPETS = collect_snippets()
 ONNX_ML = bool(os.getenv('ONNX_ML') == '1')
+ONNX_DOMAIN = ""
+ONNX_ML_DOMAIN = 'ai.onnx.ml'
```
bddppq: onnx.defs has `ONNX_DOMAIN` defined, maybe also move ONNX_ML_DOMAIN to there and here will just refer them through onnx.defs?

onnx/defs/gen_doc.py
```diff
@@ -9,13 +9,16 @@
 from collections import defaultdict
 
 from onnx import defs
-from onnx.defs import OpSchema
+from onnx.defs import OpSchema, FunctionProto
```
bddppq: FunctionProto should be imported from onnx

onnx/defs/function.h
```diff
@@ -7,6 +7,7 @@
 #include <string>
 #include <vector>
 
+#include "onnx/common/constants.h"
```
bddppq: nit: move this include to function.cc? (since it's used there)

onnx/cpp2py_export.cc
```diff
@@ -146,6 +147,38 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+  defs.def(
+      "get_all_functions",
+      [](const std::string& domain)
+          -> std::unordered_map<std::string, std::vector<py::bytes>> {
+        std::multimap<std::string, std::unique_ptr<FunctionProto>> temp_ptr_map;
+        std::unordered_map<std::string, std::vector<py::bytes>> temp_map;
+        FunctionBuilderRegistry& function_registry =
+            FunctionBuilderRegistry::OnnxInstance();
+
+        Common::Status status =
```
bddppq: check status?

onnx/cpp2py_export.cc
```diff
@@ -146,6 +147,38 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+                "'!");
+          }
+          if (!temp_map.count(iter->first)) {
+            std::vector<py::bytes> tmp_vec;
+            tmp_vec.emplace_back(py::bytes(bytes));
+            temp_map.insert(
+                std::unordered_map<std::string, std::vector<py::bytes>>::
+                    value_type(iter->first, tmp_vec));
+          } else {
+            temp_map.at(iter->first).emplace_back(py::bytes(bytes));
```
bddppq: std::move bytes?

onnx/cpp2py_export.cc
```diff
@@ -146,6 +147,38 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+            function_registry.GetFunctions(domain, &temp_ptr_map);
+        for (auto iter = temp_ptr_map.begin(); iter != temp_ptr_map.end();
+             ++iter) {
+          std::string bytes;
+          if (!iter->second->SerializeToString(&bytes)) {
+            throw std::runtime_error(
+                "Failed to serilize registered function for '" + iter->first +
+                "'!");
+          }
+          if (!temp_map.count(iter->first)) {
```
bddppq: you can use `[]` of unordered_map to implicitly create an entry if the key doesn't already exist.

onnx/defs/gen_doc.py
```diff
@@ -275,12 +375,65 @@ def main(args):  # type: (Type[Args]) -> None
...
+                fout.write(s)
+
 
 if __name__ == '__main__':
     base_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
     docs_dir = os.path.join(base_dir, 'docs')
 
     class Args(object):
         output = os.path.join(docs_dir, 'Operators' + ext)
+        function_output = os.path.join(docs_dir, 'Functions' + ext)
```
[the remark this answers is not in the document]
raymondxyang: IMO the formatting and visualization could be a future task. Currently I am trying to have type/shape inference of function working and may change the visualization accordingly. And we also talked with Netron people and try to have image generated for the functions.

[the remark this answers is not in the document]
raymondxyang: all comments except this resolved

raymondxyang:
Thanks for the detailed review :)

PR #1115, opened 2018-06-14
Add Node Tests for Dropout

houseroad:
Pretty straightforward.


houseroad:
@anirudhacharya since we just removed `is_test` tag, very likely we won't add it back soon. Even we change the schema later, we will adjust the test cases accordingly. It does not make sense to block the merge here.

I have test the cases end to end.

PR #1117, opened 2018-06-14
Add Node Tests for BatchNormalization

houseroad:
Node test for batchnorm.


houseroad:
Very likely, we won't add `is_test` back soon. I think it's good to check this PR in at the moment. The cases I introduce only contain 1 output. So even we change the output number in the schema, it won't affect the cases. Bottom line is adjusting the test case while changing the schema.

I have tested the cases end to end.

houseroad:
@anirudhacharya opset version is designed for bc changes in operator spec. Each time, we introduce bc changes, we should bump up the opset. Those models in `onnx/model` will be updated soon (est, next week).

fumihwh:
@houseroad `momentum` is not considered in test cases. Did I miss something?

PR #1120, opened 2018-06-15
Remove duplicate op LpNormalization

houseroad:
LpNormalization is applying the Lp Normalization along specified axis (axes). The definition should be quite similar to the `norm` op in [tensorflow](https://www.tensorflow.org/api_docs/python/tf/norm) and `lpnorm` in [pytorch](https://github.com/pytorch/pytorch/blob/master/caffe2/operators/lpnorm_op.cc#L6).

After some research, I found we already have equivalent operators in ONNX, i.e., ReduceL1 and ReduceL2. So, let's remove unused and duplicate operator `LpNormalization`.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -2513,44 +2513,6 @@ This version of the operator has been available since version 1 of the default O
 <dd>Indices</dd>
 </dl>
 
-### <a name="LpNormalization-1"></a>**LpNormalization-1**</a>
-
-  Given a matrix, apply Lp-normalization along the provided axis.
-
-#### Version
-
-This version of the operator has been available since version 1 of the default ONNX operator set.
-
-#### Attributes
-
-<dl>
-<dt><tt>axis</tt> : int</dt>
-<dd>(int64, default -1) the axis on which to apply normalization, -1 mean last axis.</dd>
-<dt><tt>p</tt> : int</dt>
-<dd>(int64, default 2) the order of the normalization, only 1 or 2 are supported.</dd>
-</dl>
-
-#### Inputs
-
-<dl>
-<dt><tt>input</tt> : T</dt>
-<dd>Input matrix</dd>
-</dl>
-
-#### Outputs
-
-<dl>
-<dt><tt>output</tt> : T</dt>
-<dd>Matrix after normalization</dd>
-</dl>
-
-#### Type Constraints
-
-<dl>
-<dt><tt>T</tt> : tensor(float16), tensor(float), tensor(double)</dt>
-<dd>Constrain input and output types to float tensors.</dd>
-</dl>
-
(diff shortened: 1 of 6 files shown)


bddppq:
This is a breaking change and I think our current opset mechanism doesn't support removing operator. I think we need to add "UntilVersion()" (similar to "SinceVersion()").

anirudhacharya:
I am okay with removing LpNormalization but there is/can be a difference between ReduceL2 and L2Normalization. 

ReduceL2 is ``sqrt{sum(x_i**2)}`` where ``0<= i <=n``

and L2Normalization is  ``x_i / sqrt{sum(x_i**2)}`` where ``0<= i <=n``

The above definition is for 1-d arrays, but similar logic can be extended to high dimension arrays with different axes.

anirudhacharya:
For higher dimension tensors -
```python
import numpy as np
import math
norm_eps=1e-10

#Reduce L2
in_data = np.random.rand(3,4,5)
reduceL2 = np.linalg.norm(in_data, axis=1, keepdims=True)
print(reduceL2)
```

```python
# L2Normalization
np_norm = np.linalg.norm(in_data, axis=1) + norm_eps
np_norm = np.repeat(1. / np.expand_dims(np_norm, axis=1), in_data.shape[1], axis=1)
np_out = np.multiply(in_data, np_norm)
print(np_out)
```

houseroad:
Shall we consider this as a bug? I think no one really use this operator due to its vague description.

@bddppq @anirudhacharya @ebarsoum @linkerzhang @dzhulgakov

houseroad:
@anirudhacharya could you list the frameworks which implement `LpNormalization ops`?

ebarsoum:
This isn't norm OP in TF, it is l2_norm: https://www.tensorflow.org/api_docs/python/tf/nn/l2_normalize. Initially it was called L2Normalization, then we change it to LpNormalization to be more generic. So as @anirudhacharya said they aren't the same.

houseroad:
Close the PR, since the fix is not right.

MrGeva:
The definition of LpNormalization is unclear.  
TF implements them as ReduceL1 or ReduceL2 depends on value of p. It is evident in https://github.com/onnx/onnx-tensorflow/blob/master/onnx_tf/handlers/backend/lp_normalization.py , LpNormalization is implemented using tf.norm, which is same as Reduce . Moreover, when running TF on a network with LpNormalization layers, we get results identical to Reduce.
If the layer is identical to ReduceL1 or ReduceL2, it is duplicate and better be removed.

On the other hand, in this thread LpNormalization was described differently. For p=2, result is x_i/( sqrt( sum(x_j) ) ).

If the latter definition is correct, pls confirm and document it clearly.

PR #1129, opened 2018-06-21
Use cmath instead of math.h

bstriner:
Should fix issues with std I just saw in the pytorch builds. For example:

https://ci.pytorch.org/jenkins/job/pytorch-builds/job/pytorch-linux-trusty-py3.6-gcc5.4-build/10549/console
```
/var/lib/jenkins/workspace/third_party/onnx/onnx/common/tensor.h:342:40: error: 'sqrt' is not a member of 'std'
```

### DIFF
--- a/onnx/common/tensor.h
+++ b/onnx/common/tensor.h
@@ -4,7 +4,7 @@
 #pragma once
 
 #include "onnx/onnx_pb.h"
-#include <math.h>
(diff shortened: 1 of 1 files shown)


bstriner:
@jamesr66a thanks for the quick turnaround! pytorch build back to working on latest onnx.

PR #1132, opened 2018-06-21
Add is_compatible method in python backend

houseroad:


onnx/backend/base.py
```diff
@@ -55,6 +55,15 @@ def run(self, inputs, **kwargs):  # type: (Any, **Any) -> Tuple[Any, ...]
 
 
 class Backend(object):
+    @classmethod
+    def is_compatible(cls,
+                      model,  # type: ModelProto
+                      device='CPU',  # type: Text
+                      **kwargs  # type: Any
+                      ):  # type: (...) -> Optional[bool]
+        # Return whether the model is comptabile with the backend.
```
smessmer: nit: typo

onnx/backend/base.py
```diff
@@ -55,6 +55,15 @@ def run(self, inputs, **kwargs):  # type: (Any, **Any) -> Tuple[Any, ...]
 
 
 class Backend(object):
+    @classmethod
+    def is_compatible(cls,
+                      model,  # type: ModelProto
+                      device='CPU',  # type: Text
+                      **kwargs  # type: Any
+                      ):  # type: (...) -> Optional[bool]
```
smessmer: Why `Optional`?

smessmer:
Can you also add code to call this from the backend tests and skip a test case if it returns false?

PR #1138, opened 2018-06-23
Some changes to enable Clang-cl build for Windows

peterjc123:
Disable some warnings that brought about by the use of clang-cl on Windows and change the `/MP` flag.
It is an essential part of the downstream pull pytorch/pytorch [#8784](https://github.com/pytorch/pytorch/pull/8784).


setup.py
```diff
@@ -175,6 +175,9 @@ def run(self):
                     # Temp fix for CI
                     # TODO: need a better way to determine generator
                     cmake_args.append('-DCMAKE_GENERATOR_PLATFORM=x64')
+                cmd_exists = lambda x: any(os.access(os.path.join(path, x), os.X_OK) for path in os.environ["PATH"].split(os.pathsep))
```
bddppq: Does `distutils.spawn.find_executable` work on windows?

bddppq:
@peterjc123 Thanks for working on this!

setup.py
peterjc123: Yes, I'll try!

peterjc123:
@bddppq Hi, the problem is that although `clang-cl.exe` is properly installed on Appveyor, it cannot be correctly detected by MSBuild without some extra file copying work.
There're two ways to solve this problem:
1. Copy the files in the using the environment scripts according to this [post](https://gitlab.kitware.com/cmake/cmake/issues/17930).
2. Use ninja-build and specify environmental variables like `CC` and `CXX`. It will also make the build faster.
What do you think?

bddppq:
@peterjc123 2. sounds less intrusive to me (to the user's environment), because it only requires users to install one more thing, instead of moving files around.
@raymondxyang @linkerzhang

setup.py
bddppq: Actually I would prefer to not auto opt in to use clang once it's available, instead, user can choose to use it by passing this extra flag via CMAKE_ARGS environment variable. Same for ninja. What do you think?

peterjc123: @bddppq Okay then, I'll revert the changes and add the additional tests into the build matrix.

raymondxyang:
+@snnn into the thread

peterjc123:
Test item `onnx/test/shape_inference_test.py::TestShapeInference::test_transpose_preexisting_incorrect_shape` seems to be flaky when compiled using clang-cl. I don't know the exact reason.

peterjc123:
I guess the problem is that clang-cl could not work happily with Python libraries, like pybind11. That may be the reason for the random crashes.

peterjc123:
@bddppq Since the problem is on the compiling of the python extension, what about just skipping them now? Let's consider the tests for clang-cl later.

CMakeLists.txt
```diff
@@ -18,7 +18,9 @@ option(ONNX_COVERAGE "Build with coverage instrumentation" OFF)
 option(ONNX_BUILD_TESTS "Build ONNX C++ APIs Tests" OFF)
 
 # Set C++11 as standard for the whole project
-set(CMAKE_CXX_STANDARD 11)
+IF(NOT MSVC)
```
bddppq: Doesn't MSVC support c++11?
@raymondxyang @linkerzhang

peterjc123: @bddppq The flag is useless in MSVC. Starting from C++14, it can be controlled by some flags like `/c++:14`. But for c++ 11, it can be just determined by MSVC version and is enabled by default.

snnn: @peterjc123   Then by default it would be?

snnn: I don't think onnx can support VC verions earlier than 2015. According to https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version, after your this change, it will default to C++14. Am I right?

PR #1148, opened 2018-06-27
Opset Version Converter

Ac2zoom:
A tool to convert an ingested model to any other opset version in any other domain, making both backwards compatibility and version updating easier.


onnx/common/ir.h
```diff
@@ -767,11 +767,19 @@ struct Node : public Attributes<Node> {
...
+  int64_t version;
+};
+
 struct Graph final {
 ONNX_DISALLOW_COPY_AND_ASSIGN(Graph);
 friend struct Node;
 friend struct Value;
 
+public:
+  OpSetID opset_version;
```
bddppq: ModelProto can have a list of opset_versions: https://github.com/onnx/onnx/blob/33e9cd4182fe468675241fba4ae8a16c2f0bd82f/onnx/onnx.proto#L198

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,165 @@
...
+            op_opset_version = version_pair.first;
+            current_opschemas[op] = op_domain_map[initial_domain][op_opset_version];
+          }
+        }
+      }
+    }
+
+    // Iterate over all versions to target_version for specified
+    int64_t curr_version = initial_version.version;
+    int64_t next_version;
```
houseroad: instead of having next_version, you can use step ==> -1 or 1

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,165 @@
...
+    }
+    // Identify index of this domain in g.opset_versions
+    unsigned int domain_index = 0;
+    for (unsigned int i = 0; i < g->opset_versions.size(); i++) {
+      if (g->opset_versions[i].domain == "") {
+        domain_index = i;
+      }
+    }
+    while (curr_version != target_version.version) {
+      std::cerr << "curr_version: " << curr_version << ", next_version: " << next_version << std::endl;
```
houseroad: Probably you don't want to always output this, adding a debug switch?

houseroad:
jenkins CI failure is unrelated.

bddppq:
@onnxbot retest this please

onnx/version_converter.py
```diff
@@ -0,0 +1,44 @@
...
+
+Return:
+    return (ModelProto) converted model
+
+Supported adapters:
+    --
+"""
+
+
+def convert_version(model, initial_version, target_version):  # type: (ModelProto, OperatorSetIdProto, OperatorSetIdProto) -> ModelProto
```
bddppq: Shouldn't the `initial_version` be taken from the model?

houseroad: The model may contain several opset_imports. I agree with that if no initial_version is specified, we just use the only opset_import in the model.

houseroad: After the discussion, let's stick to the opset_imports in model.

The new API should only contain model and target_version.

onnx/common/ir_pb_converter.cc
```diff
@@ -530,6 +537,58 @@ void encodeGraph(ONNX_NAMESPACE::GraphProto * p_g, const std::shared_ptr<Graph>
...
+  // Add new opset_versions
+  p_m->clear_opset_import();
+  for (OpSetID opset : g->opset_versions) {
+    OperatorSetIdProto *opset_version_output = p_m->add_opset_import();
+    opset_version_output->set_domain(opset.domain);
+    opset_version_output->set_version(opset.version);
+  }
+}
+
+ONNX_NAMESPACE::ModelProto PrepareOutput(const ONNX_NAMESPACE::ModelProto& mp_in) {
```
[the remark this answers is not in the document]
Ac2zoom: Given that the bulk of a ModelProto is really just a GraphProto, does this have much use outside of providing a separate container for the OpsetVersion?

onnx/test/version_converter_test.py
```diff
@@ -0,0 +1,67 @@
...
+    def _converted(self, graph, initial_version, target_version):  # type: (GraphProto, OperatorSetIdProto, OperatorSetIdProto) -> ModelProto
+        orig_model = helper.make_model(graph, producer_name='onnx-test')
+        # print(type(orig_model))
+        converted_model = onnx.version_converter.convert_version(orig_model,
+                initial_version, target_version)
+        checker.check_model(converted_model)
+        return converted_model
+
+    # fn is a function that takes a single node as argument
+    def _visit_all_nodes_recursive(self, graph, fn):  # type: (GraphProto, Callable[[NodeProto], None]) -> None
```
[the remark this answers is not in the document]
Ac2zoom: Is this in reference to the output type or the output type of the Callable parameter?  I believe it should be appropriate for both: this is the syntax currently used in the optimizer.

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,117 @@
...
+  virtual ~BaseVersionConverter() = default;
+
+  Adapter* adapter_lookup(Node* op,
+      const OpSetID& initial_version,
+      const OpSetID& target_version) {
+    std::string op_name = op->name();
+    std::string initial = stringify_opsetid(initial_version);
+    std::string target = stringify_opsetid(target_version);
+    // TODO: Find appropriate adapter in adapters map for provided initial and target versions
+    if (adapters.find(op_name) != adapters.end()) {
```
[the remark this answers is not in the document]
Ac2zoom: Beyond the consideration of concerns specific to the default domain, where else should we focus?  I've written a brief new proposal regarding modularizing the downwards vs. cross-domain vs. upwards adapters such that they can be invoked by the BaseConverter and each separate module can be invoked by the DefaultConverter.

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,117 @@
...
+    std::stringstream ss(target);
+    std::string segment;
+    std::vector<std::string> seglist;
+    while (std::getline(ss, segment, '$')) {
+      seglist.push_back(segment);
+    }
+    return seglist;
+  }
+
+  OpSetID operatorsetidproto_to_opsetid(ONNX_NAMESPACE::OperatorSetIdProto proto) {
```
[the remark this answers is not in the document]
Ac2zoom: Thank you!

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,117 @@
...
+    OpSetID iv = a_ptr->initial_version;
+    OpSetID tv = a_ptr->target_version;
+    adapters[a_ptr->name][stringify_opsetid(iv)][stringify_opsetid(tv)] = a_ptr;
+  }
+
+  std::string stringify_opsetid(OpSetID target) {
+    return target.domain + "$" + std::to_string(target.version);
+  }
+
+  std::vector<std::string> destringify_opsetid(std::string target) {
```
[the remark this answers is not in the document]
Ac2zoom: This could actually just be a constructor of OpSetID, and I could even make better use of the object in the existing calls to destringify.

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,59 @@
...
+  ONNX_NAMESPACE::ModelProto convert_version(
+      const ONNX_NAMESPACE::ModelProto& mp_in,
+      const OpSetID initial_version,
+      const OpSetID target_version) {
+    barf("BaseConverter does not include an implementation of convert_version.  "
+        "Please use a more specific converter, such as DefaultConverter.");
+  };
+
+  void registerAdapter(Adapter* a_ptr, std::string domain) {
+    OpSetID iv = a_ptr->initial_version;
```
smessmer: Taking these as const references doesn't need to copy it

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
...
+      // adapter implemented for initial_version. If we're adapting upwards, we
+      // want to actually use the SinceVersion value for the given op.
+      if (target_version.version < initial_version.version) {
+        // Downwards adapter
+        if (adapters[op_name].find(initial) != adapters[op_name].end()) {
+          // Either an upwards or a downwards adapter exists
+          // Check if downwards adapter exists (only one should)
+          const auto target_map = adapters[op_name][initial];
+          for (auto it = target_map.begin(); it != target_map.end(); ++it) {
+            int64_t new_target = (new OpSetID(it->first))->version;
```
smessmer: This is a memory leak

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
...
+    // DefaultConverter to separate methods here and maintain the procedure in Base Converter
+    if (adapters.find(op_name) != adapters.end()) {
+      // If we're adapting downwards, we just want to find the one downwards
+      // adapter implemented for initial_version. If we're adapting upwards, we
+      // want to actually use the SinceVersion value for the given op.
+      if (target_version.version < initial_version.version) {
+        // Downwards adapter
+        if (adapters[op_name].find(initial) != adapters[op_name].end()) {
+          // Either an upwards or a downwards adapter exists
+          // Check if downwards adapter exists (only one should)
```
smessmer: Is this always true, that there will only be exactly one up- or downwards adapter? If yes, we can encode it into the data structure and make it faster by having a nesting level less. You don't need a map if it always has only one entry.

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
...
+    } else {
+      // No adapters exist for the given op
+      // TODO: Instead return NoAdapterForOp
+      return NULL;
+    }
+  }
+
+  ONNX_NAMESPACE::ModelProto convert_version(
+      const ONNX_NAMESPACE::ModelProto& mp_in,
+      const OpSetID initial_version,
```
smessmer: pass by reference

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
...
+      version_range.first, version_range.second);
+    // Compile list of all ops used in the model
+    graph_node_list nodes = g->nodes();
+
+    std::vector<OpSchema> all_opschemas = ONNX_NAMESPACE::OpSchemaRegistry::get_all_schemas_with_history();
+
+    // Create Map for All Versions of format {op_name: {domain: {version: schema}}}
+    std::unordered_map<std::basic_string<char>, std::unordered_map<std::basic_string<char>, std::map<int64_t, ONNX_NAMESPACE::OpSchema*>>>  all_schemas;
+
+    for (OpSchema schema : all_opschemas) {
```
smessmer: take by reference

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
...
+      all_schemas[schema.Name()][schema.domain()][(int64_t) schema.since_version()] = &schema;
+    }
+
+    const char* initial_domain = initial_version.domain.c_str();
+    const char* target_domain = target_version.domain.c_str();
+
+    // Create Map for Current Version
+    for (Node* op : nodes) {
```
smessmer: `const Node*` instead of `Node*` is a bit safer

docs/OpsetVersionConverter.md
```diff
@@ -0,0 +1,50 @@
...
+The version converter may be invoked either via C++ or Python. The Python API
+is described, with example,
+[here](docs/PythonAPIOverview.md#converting-opset-version-of-an-onnx-model).
+
+The C++ API consists of a single function
+
+```
+ModelProto ConvertVersion(
+    const ModelProto& mp_in,
+    const OpSetID& initial_version,
```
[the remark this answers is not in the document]
Ac2zoom: @houseroad thoughts on providing another layer of indirection here by primarily providing a method without initial_version to the C++ client?

onnx/common/ir_pb_converter.cc
```diff
@@ -530,6 +537,58 @@ void encodeGraph(ONNX_NAMESPACE::GraphProto * p_g, const std::shared_ptr<Graph>
...
+  ONNX_NAMESPACE::ModelProto mp_out{};
+
+  if (mp_in.has_ir_version()) {
+    mp_out.set_ir_version(mp_in.ir_version());
+  }
+  if (mp_in.has_producer_name()) {
+    mp_out.set_producer_name(mp_in.producer_name());
+  }
+  if (mp_in.has_producer_version()) {
```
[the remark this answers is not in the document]
Ac2zoom: Yeah, I think that makes sense!  @houseroad who would be the best person from the Optimizer project to verify this with?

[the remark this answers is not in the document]
Ac2zoom: Fwiw, we can entirely bypass using PrepareOutput in the VersionConverter and instead just copy mp_in directly in convert_version

onnx/test/version_converter_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+import numpy as np  # type: ignore
+
+import onnx.version_converter
+import unittest
+
+
+class TestVersionConverter(unittest.TestCase):
+
+    def _converted(self, graph, initial_version, target_version):  # type: (GraphProto, OperatorSetIdProto, int) -> ModelProto
```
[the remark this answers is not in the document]
Ac2zoom: Should I change this in optimizer_test, too?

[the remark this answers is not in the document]
Ac2zoom: In fact, should I abstract these to a test_helper class @houseroad?

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,59 @@
...
+struct BaseVersionConverter {
+  // Schema for adapters: {<op_name>:{<from_domain>$<from_version>:{<to_domain>
+  // <to_version>: adapter}}}
+  std::unordered_map<std::string, std::unordered_map<std::string, std::unordered_map<std::string, Adapter*>>> adapters;
+
+  std::unordered_map<Node*, OpSchema*> current_opschemas;
+
+  BaseVersionConverter() {}
+
+  ~BaseVersionConverter() {
```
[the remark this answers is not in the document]
Ac2zoom: Is it best practice to make constructors virtual, as well, so as to prevent variables instantiated as BaseConverter from solely invoking BaseConverter's methods?

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,59 @@
...
+
+  ONNX_NAMESPACE::ModelProto convert_version(
+      const ONNX_NAMESPACE::ModelProto& mp_in,
+      const OpSetID initial_version,
+      const OpSetID target_version) {
+    barf("BaseConverter does not include an implementation of convert_version.  "
+        "Please use a more specific converter, such as DefaultConverter.");
+  };
+
+  void registerAdapter(Adapter* a_ptr, std::string domain) {
```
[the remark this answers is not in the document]
Ac2zoom: domain parameter removed

onnx/version_converter/adapters/adapter.h
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include "onnx/common/ir.h"
+#include "onnx/onnx_pb.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+struct Adapter {
+
+  virtual ~Adapter() noexcept = 0;
```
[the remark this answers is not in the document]
Ac2zoom: Another carryover from Optimizer.  I wonder why they did it there.

onnx/common/ir_pb_converter.cc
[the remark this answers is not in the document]
Ac2zoom: Ok so here's the design issue right now: If we attempt to refill the properties based on the new graph, we face the problem of re-adding existing inputs/outputs/attributes.  As such, to appropriately implement this, we would need to write a new version of ExportModelProto which replaces elements/nodes rather than adding them.  If @houseroad thinks this could be useful, I'd be happy to do so.

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
+// Converter for ONNX models between different opset versions
+// in the same domain.
+
+#pragma once
+
+#include "onnx/version_converter/BaseConverter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+struct IntraDomainVersionConverter : BaseVersionConverter {
```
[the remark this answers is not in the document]
Ac2zoom: DefaultVersionConverter extends this, so it can't be final.  However, that one might be.  I haven't done so in case we want to be able to extend it (for example to implement more restrictions on the ops we convert/the process of doing so).

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
...
+  bool DEBUG = false;
+
+  IntraDomainVersionConverter() {
```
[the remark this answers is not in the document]
Ac2zoom: We might ultimately add adapters here, so I figured I'd leave it be.

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
...
+    // Find appropriate adapter in adapters map for provided initial and target versions
+    // TODO: Consider abstracting elements of this that are specific to
+    // DefaultConverter to separate methods here and maintain the procedure in Base Converter
+    if (adapters.find(op_name) != adapters.end()) {
+      // If we're adapting downwards, we just want to find the one downwards
+      // adapter implemented for initial_version. If we're adapting upwards, we
+      // want to actually use the SinceVersion value for the given op.
+      if (target_version.version < initial_version.version) {
+        // Downwards adapter
+        if (adapters[op_name].find(initial) != adapters[op_name].end()) {
```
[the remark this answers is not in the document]
Ac2zoom: To properly take advantage of this, I may need to unfold some of the block at lines 63-70.

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,212 @@
...
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+struct IntraDomainVersionConverter : BaseVersionConverter {
+  bool DEBUG = false;
+
+  IntraDomainVersionConverter() {
+    // TODO: Register adapters to the version converter
+  }
+
+  Adapter* adapter_lookup(Node* op,
```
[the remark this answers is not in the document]
Ac2zoom: Yes, absolutely.  Should we do the same for convert_version @houseroad?

onnx/version_converter/BaseConverter.h
Ac2zoom: Even the pointer?

onnx/version_converter/IntraDomainConverter.h
Ac2zoom: For this purpose, as we continue to evaluate the current version of adapter_lookup, it may make sense to move it back to the BaseConverter.

docs/PythonAPIOverview.md
```diff
@@ -177,6 +177,28 @@ print('After shape inference, the shape info of Y is:\n{}'.format(inferred_model
...
+
+# Preprocessing: load the model to be converted.
+model_path = 'path/to/the/model.onnx'
+original_model = onnx.load(model_path)
+
+print('The model before conversion:\n{}'.format(original_model))
+
+# A full list of supported adapters can be found here:
+# https://github.com/onnx/onnx/blob/master/onnx/version_converter.py#L21
+initial_version = helper.make_operatorsetid("<domain>", <version>)
```
houseroad: We need to update the doc for python API.

onnx/common/ir.h
```diff
@@ -767,6 +767,56 @@ struct Node : public Attributes<Node> {
...
+      :domain_(proto.domain()), version_(proto.version()) {}
+
+    // Default Domain Constructor
+    explicit OpSetID(const int64_t& version)
+      :domain_(""), version_(version) {}
+
+    explicit OpSetID(const std::string& domain, int version)
+      :domain_(domain), version_(version) {}
+
+    explicit OpSetID()
```
houseroad: I think this one is not necessary.

Ac2zoom: Perfect.  I was originally using it somewhere, but I seem to have removed that usage

docs/PythonAPIOverview.md
```diff
@@ -177,6 +177,25 @@ print('After shape inference, the shape info of Y is:\n{}'.format(inferred_model
 Runnable IPython notebooks:
 - [shape_inference.ipynb](https://github.com/onnx/onnx/tree/master/onnx/examples/shape_inference.ipynb)
 
+## Converting Version of an ONNX Model
```
houseroad: within default domain

onnx/common/ir.h
```diff
@@ -767,6 +767,53 @@ struct Node : public Attributes<Node> {
...
+    explicit OpSetID(const std::string& target) {
+      std::stringstream ss(target);
+      std::string segment;
+      std::vector<std::string> seglist;
+      while (std::getline(ss, segment, '$')) {
+        seglist.push_back(segment);
+      }
+      ONNX_ASSERTM(seglist.size() == 2, "string must be of format $<domain>$<version>")
+      domain_ = seglist[0];
+      version_ = atoi(seglist[1].c_str());
```
houseroad: just release, `stoi` is more robust.

onnx/common/ir.h
```diff
@@ -767,6 +767,53 @@ struct Node : public Attributes<Node> {
...
+
+    std::string toString() const {
+      return "$" + domain_ + "$" + std::to_string(version_);
+    }
+
+    const std::string& domain() const {
+      return domain_;
+    }
+
+    const int64_t& version() const {
```
houseroad: For primitive type, no need to use/return reference, instead, return value is even faster.

onnx/version_converter/BaseConverter.h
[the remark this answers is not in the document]
smessmer: Constructors can't be virtual. Virtual means that it's deciding which function to call based on what the object actually is - but for a constructor call you don't have an object yet to base that decision on.

onnx/version_converter/IntraDomainConverter.h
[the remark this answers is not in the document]
smessmer: ok makes sense, didn't see that DefaultVersionConverter inherits from it.

onnx/common/ir_pb_converter.cc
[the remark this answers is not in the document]
houseroad: This is a legacy problem. Let's fix it in a different pr.

onnx/common/ir_pb_converter.h
```diff
@@ -37,4 +37,6 @@ void ExportModelProto(ONNX_NAMESPACE::ModelProto* p_m, const std::shared_ptr<Gra
 
 std::unique_ptr<Graph> ImportModelProto(const ONNX_NAMESPACE::ModelProto& mp);
 
+ONNX_NAMESPACE::ModelProto PrepareOutput(const ONNX_NAMESPACE::ModelProto& mp_in);
```
houseroad: It's inside ONNX_NAMESPACE. It should be ok to just use ModelProto.

onnx/version_converter/BaseConverter.h
smessmer: I didn't mean the function arguments but the `OpSetId iv` and below. Function arguments might also be made const references if not retained, but as you noticed that does not make sense for the pointer.

onnx/version_converter/convert.cc
```diff
@@ -0,0 +1,24 @@
...
+static DefaultVersionConverter _version_converter;
+
+ONNX_NAMESPACE::ModelProto ConvertVersion(
+    const ONNX_NAMESPACE::ModelProto& mp_in,
+    const int target_version) {
+  // Get initial_opsetid from mp_in
+  OpSetID* initial_struct;
+  for (auto it = mp_in.opset_import().begin(); it != mp_in.opset_import().end(); ++it) {
+    if (it->domain() == "" || it->domain() == "ai.onnx") {
+      initial_struct = new OpSetID(*it);
```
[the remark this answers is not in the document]
smessmer: yes

onnx/test/version_converter_test.py
```diff
@@ -0,0 +1,77 @@
...
+    ):  # type: (...) -> ModelProto
+        orig_model = helper.make_model(graph, producer_name='onnx-test', opset_imports=[initial_version])
+        # print(type(orig_model))
+        converted_model = onnx.version_converter.convert_version(orig_model,
+                target_version)
+        checker.check_model(converted_model)
+        return converted_model
+
+    # fn is a function that takes a single node as argument
+    def _visit_all_nodes_recursive(
```
houseroad: I guess this node is not necessary.

onnx/test/version_converter_test.py
```diff
@@ -0,0 +1,77 @@
...
+        for node in graph.node:
+            fn(node)
+            for attr in node.attribute:
+                if attr.g is not None:
+                    self._visit_all_nodes_recursive(attr.g, fn)
+                if len(attr.graphs):
+                    for gr in attr.graphs:
+                        self._visit_all_nodes_recursive(gr, fn)
+
+    # Test 1: Backwards Incompatible Conversion: Add: 8 -> 2
```
houseroad: You probably want to make this case more complicated. Add has broadcast and axis attribute.

Ac2zoom: This is all covered in the actual adapter tests on the adapters branch: https://github.com/onnx/onnx/pull/1184/files#diff-fe3eada8cd6b3b792277582934f39198

houseroad: My point is that let's have a case, in which the adapter is not visible.

onnx/version_converter.py
```diff
@@ -0,0 +1,39 @@
+"""onnx version converter
+
+This enables users to convert their models between different opsets.
```
houseroad: Let's mention this is a opset version converter only for default domain.

onnx/version_converter/convert.cc
```diff
@@ -0,0 +1,25 @@
...
+  OpSetID initial_struct(0);
+  for (auto it = mp_in.opset_import().begin(); it != mp_in.opset_import().end(); ++it) {
+    if (it->domain() == "" || it->domain() == "ai.onnx") {
+      initial_struct = OpSetID(*it);
+    }
+  }
+  OpSetID target_struct = OpSetID(target_version);
+  return _version_converter.convert_version(mp_in, initial_struct, target_struct);
+}
+}}
```
houseroad: Add comment to tell it's the end of which namespace.

onnx/version_converter/convert.cc
```diff
@@ -0,0 +1,25 @@
+// ATTENTION: The code in this file is highly EXPERIMENTAL.
```
houseroad: Not experimental :-)

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,41 @@
...
+
+#pragma once
+
+#include "onnx/version_converter/IntraDomainConverter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+class DefaultVersionConverter : IntraDomainVersionConverter {
+  private:
+    bool DEBUG = false;
```
houseroad: how to use this flag?

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,55 @@
...
+    ONNX_NAMESPACE::ModelProto convert_version(
+        const ONNX_NAMESPACE::ModelProto& mp_in,
+        const OpSetID& initial_version,
+        const OpSetID& target_version) const {
+      ONNX_ASSERTM(false, "BaseConverter does not include an implementation of convert_version.  "
+          "Please use a more specific converter, such as DefaultConverter.");
+      throw "BaseVersionConverter Exception";
+    };
+
+    void registerAdapter(Adapter* a_ptr) {
```
houseroad: Take unique_ptr as parameter?

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,55 @@
...
+  // <to_version>: adapter}}}
+  protected:
+    std::unordered_map<std::string, std::unordered_map<std::string, std::unordered_map<std::string, std::unique_ptr<Adapter>>>> adapters;
+
+  public:
+    BaseVersionConverter() = default;
+
+    virtual ~BaseVersionConverter() = default;
+
+    const Adapter& adapter_lookup(Node* op,
```
houseroad: Add some comments talking about when `adapter_lookup` should be called?

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,207 @@
...
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+class IntraDomainVersionConverter : BaseVersionConverter {
+  private:
+    bool DEBUG = false;
+
+  public:
+    IntraDomainVersionConverter() {
+      // TODO: Register adapters to the version converter
```
houseroad: This is not a TODO, just a note :-)

houseroad: Also do we really want to register any adapter directly in intra one? probably just delete this comment.

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,205 @@
...
+          std::string since = target_version.domain() + std::to_string(
+              current_opschemas.at(op)->since_version());
+          const auto& op_adapters = adapters.at(op_name);
+          if (op_adapters.find(since) != op_adapters.end() &&
+            op_adapters.at(since).find(target) != op_adapters
+            .at(since).end()) {
+            return *(op_adapters.at(since).at(target));
+          } else {
+            // TODO: Instead return NoUpwardsAdapter
+            throw "NoUpwardsAdapter";
```
houseroad: follow the tensorerror in https://github.com/onnx/onnx/blob/master/onnx/common/assertions.h#L16.
Create a new type of error, or just use ONNX_ASSERTM

Please don't directly throw in .h file.

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,205 @@
...
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+class IntraDomainVersionConverter : BaseVersionConverter {
+  private:
+    bool DEBUG = false;
+
+  public:
+    IntraDomainVersionConverter() = default;
+
+    const Adapter& adapter_lookup(const Node* op,
```
houseroad: add `override` specifier.

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,59 @@
...
+  public:
+    BaseVersionConverter() = default;
+
+    virtual ~BaseVersionConverter() = default;
+
+    // adapter_lookup should be called in convert_version when the user would
+    // like to identify the proper registered adapter in the adapters map for
+    // a given Node from a certain version to another. It should only be called
+    // when the user knows that an adapter should exist for the given context.
+    const Adapter& adapter_lookup(Node* op,
```
houseroad: virtual = 0

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,59 @@
...
+        const OpSetID& initial_version,
+        const OpSetID& target_version) const {
+      // TODO: Abstract to helper?
+      ONNX_ASSERTM(false, "BaseConverter does not include an implementation of adapter_lookup.  "
+          "Please use a more specific converter, such as DefaultConverter.");
+      throw "BaseVersionConverter Exception";
+    }
+
+    ONNX_NAMESPACE::ModelProto convert_version(
```
houseroad: virtual = 0

onnx/version_converter/IntraDomainConverter.h
houseroad: Why not make this as a general look up? What's the optimization here?

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,205 @@
...
+          }
+        }
+      } else {
+        // No adapters exist for the given op
+        // TODO: Instead return NoAdapterForOp
+        throw "NoAdapterForOp";
+      }
+    }
+
+    ONNX_NAMESPACE::ModelProto convert_version(
```
houseroad: override specifier

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,205 @@
...
+        const ONNX_NAMESPACE::ModelProto& mp_in,
+        const OpSetID& initial_version,
+        const OpSetID& target_version) const {
+      ONNX_ASSERTM(strcmp(initial_version.domain().c_str(), target_version.domain()
```
houseroad: just directly using `==` to compare two std::string should be fine

houseroad: You can overload `==` for OpSetID.

onnx/version_converter/IntraDomainConverter.h
```diff
@@ -0,0 +1,205 @@
...
+      const std::string& search_domain = target_version.domain() == "ai.onnx" ? "" : target_version.domain();
+      const std::pair<int, int>& version_range = versions_map.at(search_domain);
+      ONNX_ASSERTM(target_version.version() >= version_range.first && target_version
+          .version() <= version_range.second,
+        "Warning: invalid target_version (must be between %s and %s",
+        version_range.first, version_range.second);
+      // Compile list of all ops used in the model
+      graph_node_list nodes = g->nodes();
+
+      const std::vector<OpSchema>& all_opschemas = ONNX_NAMESPACE::OpSchemaRegistry::get_all_schemas_with_history();
```
houseroad: probably you can remove all the ONNX_NAMESPACE in this file, because at the beginning, we already specify that we are using ONNX_NAMESPACE.

onnx/version_converter/IntraDomainConverter.h
houseroad: This whole logic should be moved to default converter. Because other intra-domain converters may not follow our logic.

onnx/version_converter/convert.cc
```diff
@@ -0,0 +1,23 @@
...
+}
+
+ONNX_NAMESPACE::ModelProto ConvertVersion(
+    const ONNX_NAMESPACE::ModelProto& mp_in,
+    const int target_version) {
+  // Get initial_opsetid from mp_in
+  OpSetID initial_struct(0);
+  for (auto it = mp_in.opset_import().begin(); it != mp_in.opset_import().end(); ++it) {
+    if (it->domain() == "" || it->domain() == "ai.onnx") {
+      initial_struct = OpSetID(*it);
```
houseroad: No need to create a new OpSetID. Just set the opset field should be enough, and then break the loop, since you already find the default domain one

onnx/version_converter/IntraDomainConverter.h
Ac2zoom: Unfortunately didn't work: compilation error with strings not being possible to pass as an expression

Ac2zoom: We're really only concerned with the domain; we don't really check equality elsewhere.  If it'd still be good protocol, though, I can definitely overload ==.

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,38 @@
...
+      // Register adapters to the version converter
+    }
+
+    ONNX_NAMESPACE::ModelProto convert_version(
+        const ONNX_NAMESPACE::ModelProto& mp_in,
+        const OpSetID& initial_version,
+        const OpSetID& target_version) const {
+      const char* initial_domain = initial_version.domain().c_str();
+      const char* target_domain = target_version.domain().c_str();
+      ONNX_ASSERTM((strcmp(initial_domain, "") == 0 || strcmp(initial_domain,
```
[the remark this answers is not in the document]
Ac2zoom: Exclusively?

onnx/common/ir.h
```diff
@@ -767,6 +767,56 @@ struct Node : public Attributes<Node> {
...
+  private:
+    std::string domain_;
+    int64_t version_;
+
+  public:
+    explicit OpSetID(const ONNX_NAMESPACE::OperatorSetIdProto& proto)
+      :domain_(proto.domain()), version_(proto.version()) {}
+
+    // Default Domain Constructor
+    explicit OpSetID(const int64_t& version)
```
smessmer: no need for reference because reference is 64bit and int64_t is also just 64bit - i.e. copying an int64_t is not more expensive than copying the reference. Then, the indirection of a reference here introduces a small performance hit.

onnx/common/ir.h
```diff
@@ -767,6 +767,56 @@ struct Node : public Attributes<Node> {
...
+      :domain_(""), version_(version) {}
+
+    explicit OpSetID(const std::string& domain, int64_t version)
```
smessmer: You can move in the domain, i.e. have it by-value in the argument and assign it to the member with `std::move`.

onnx/common/ir.h
```diff
@@ -767,6 +767,56 @@ struct Node : public Attributes<Node> {
...
+
+    std::string toString() const {
+      return "$" + domain_ + "$" + std::to_string(version_);
+    }
+
+    static OpSetID fromString(const std::string& target) {
+      try {
+        std::string new_domain = target.substr(0, target.find("$"));
+        int new_version = std::stoi(target.substr(target.find("$") + 1, target.length()).c_str());
+        return OpSetID(new_domain, new_version);
```
smessmer: you can move `new_domain` into the `OpSetID`

onnx/common/ir.h
```diff
@@ -875,6 +927,11 @@ friend struct Value;
   const_graph_node_list nodes() const {
     return const_graph_node_list(output_, kNextDirection);
   }
+
+  std::vector<OpSetID>& opset_versions() {
```
smessmer: Can you return a const reference to make sure people don't modify it and break class invariants?
If people need to be able to change it but only one kind of change (say they only need to be able to add stuff), add a member `addOpSetVersion()` specifically only allowing this one kind of change. That makes it easier to uphold class invariants.

onnx/common/ir_pb_converter.cc
```diff
@@ -299,7 +299,12 @@ std::unique_ptr<Graph> ImportModelProto(const ONNX_NAMESPACE::ModelProto& mp) {
     return nullptr;
   }
 
-  return graphProtoToGraph(mp.graph(), false);
+  std::unique_ptr<Graph> g(graphProtoToGraph(mp.graph(), false));
+  for (int i = 0; i < mp.opset_import_size(); i++) {
+    OpSetID new_opset_version(mp.opset_import(i).domain(), mp.opset_import(i).version());
+    g->opset_versions().push_back(new_opset_version);
```
smessmer: Use `emplace_back` and `std::move`

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,78 @@
...
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+// TODO: Consider creating interface for this class.
+class BaseVersionConverter {
+  // Schema for adapters: {<op_name>:{<from_domain>$<from_version>:{<to_domain>
+  // <to_version>: adapter}}}
+  protected:
+    std::unordered_map<std::string, std::unordered_map<std::string, std::unordered_map<std::string, std::unique_ptr<Adapter>>>> adapters;
```
smessmer: Why do we still have the three-level map? Didn't we say a table (i.e. two-level) is enough?

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,78 @@
...
+    virtual ~BaseVersionConverter() = default;
+
+    // adapter_lookup should be called in convert_version when the user would
+    // like to identify the proper registered adapter in the adapters map for
+    // a given Node from a certain version to another. It should only be called
+    // when the user knows that an adapter should exist for the given context.
+    const Adapter& adapter_lookup(const Node* op,
+        const OpSetID& initial_version,
+        const OpSetID& target_version) const {
+      const std::string& op_name = op->kind().toString();
```
smessmer: `.toString()` likely returns a temporary object, not a reference to some object stored inside `op`. So if you just keep a reference (i.e. pointer) to it, the object itself will not live anymore when you try to access it later through the reference. You have to use non-references here.

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,78 @@
...
+      const std::string& initial = initial_version.toString();
+      const std::string& target = target_version.toString();
+      // Find appropriate adapter in adapters map for provided initial and target versions
+      // TODO: Consider abstracting elements of this that are specific to
+      // DefaultConverter to separate methods here and maintain the procedure in Base Converter
+      const auto& op_adapters = adapters.find(op_name);
```
smessmer: `adapters.find()` returns an iterator object. This is basically just a pointer and cheap to copy, so you don't need a reference here. Also, the reference here actually breaks your code, because (as in the previous comment), the iterator object gets destroyed if you don't store it and your reference will point to something dead.

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,78 @@
...
+      if (op_adapters != adapters.end()) {
+        // If we're adapting downwards, we just want to find the one downwards
+        // adapter implemented for initial_version. If we're adapting upwards, we
+        // want to actually use the SinceVersion value for the given op.
+        const auto& target_map = op_adapters->second.find(initial);
```
smessmer: same here

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,78 @@
...
+        if (target_map != op_adapters->second.end()) {
+          // Either adapt from SinceVersion or Incompatible Breaking Change
+          const auto& adapter_ptr = target_map->second.find(target);
```
smessmer: and here

onnx/version_converter/BaseConverter.h
```diff
@@ -0,0 +1,78 @@
...
+        ONNX_ASSERTM(false, "NoAdapterForOp");
+      }
+  }
+
+  virtual ModelProto convert_version(
+      const ONNX_NAMESPACE::ModelProto& mp_in,
+      const OpSetID& initial_version,
+      const OpSetID& target_version) const = 0;
+
+  void registerAdapter(std::unique_ptr<Adapter>&& a_ptr) {
```
smessmer: There's no need to use `&&` here. Just pass in `(unique_ptr<Adapter> a_ptr)`

onnx/version_converter/adapters/NoPreviousVersionAdapter.h
```diff
@@ -0,0 +1,19 @@
...
+// opset version.
+
+#pragma once
+
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+struct NoPreviousVersionAdapter final : public Adapter {
+  explicit NoPreviousVersionAdapter(const std::string& op_name, const OpSetID&
```
smessmer: You can probably move in `op_name`, `initial` and `target`.

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,162 @@
...
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+class DefaultVersionConverter : public BaseVersionConverter {
+  private:
+    bool DEBUG = false;
+
+    bool searchOpDomainMap(const std::unordered_map<std::string, std::map<
+      int64_t, const OpSchema*>>& op_domain_map, int64_t curr_version) const {
+      const auto& version_it = op_domain_map.find("");
```
smessmer: no reference

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,162 @@
...
+      return version_it != op_domain_map.end() &&
+          version_it->second.find(curr_version) !=
+          version_it->second.end();
+    }
+
+  public:
+    DefaultVersionConverter() {
+      // Register adapters to the version converter
+      registerAdapter(std::unique_ptr<Adapter>(new NoPreviousVersionAdapter("Cos",
```
smessmer: use `make_unique` instead, never directly call `new` or `delete`.

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,162 @@
...
+      // No conversion necessary if Model has single, equivalent opset version
+      // if (initial_versions.size() == 1 && initial_versions[0].version ==
+      //    target_version.version && initial_versions[0].domain ==
+      //    target_version.domain) {
+      //  return mp_in;
+      // }
+
+      // Check if target_version is valid
+      const std::unordered_map<std::string, std::pair<int, int>>& versions_map = OpSchemaRegistry::DomainToVersionRange::Instance().Map();
+      const std::string& search_domain = target_version.domain() == "ai.onnx" ? "" : target_version.domain();
```
smessmer: Here you're also storing a reference to a dead object.

onnx/common/ir.h
Ac2zoom: So I could do this, but it would require removing the appropriate OpSetVersion at each iteration of my converter (or, I guess, at the end).  By contrast, if I just edit the vector, it's simpler there.

onnx/version_converter/BaseConverter.h
Ac2zoom: I think we decided that we wanted to maintain 3-level for the cross-domain case, since it made sense to maintain BaseConverter for that purpose.  That may have changed due to the abstraction of adapter_lookup to BaseConverter and the removal of IntraDomainConverter, though.  What do you think, @houseroad?

onnx/version_converter/BaseConverter.h
Ac2zoom: @houseroad mentioned it as an indication to users that a_ptr will be unusable

onnx/version_converter/adapters/NoPreviousVersionAdapter.h
Ac2zoom: I also need to do this for BackwardsCompatibleAdapter

onnx/version_converter/convert.h
Ac2zoom: Doesn't compile due to Adapter being an abstract class.  @houseroad?

onnx/version_converter/convert.h
Ac2zoom: search_domain, right? (not versions_map)

onnx/common/ir.h
Ac2zoom: Will change name of method

onnx/version_converter/convert.h
Ac2zoom: need to change this for adapters, too

Ac2zoom:
Are all the test failures on caffe2 my fault?

Ac2zoom:
Quite strange that it seems like all the CI is passing otherwise

houseroad:
@Ac2zoom ignore the default CI, the image is broken.

onnx/common/ir.h
```diff
@@ -767,6 +767,56 @@ struct Node : public Attributes<Node> {
...
 
+// A class with the same properties as OperatorSetIdProto, but without protobuf
+// overhead, resulting in a simpler and more readable workflow.
+class OpSetID final {
+  private:
+    std::string domain_;
+    int64_t version_;
+
+  public:
+    explicit OpSetID(const ONNX_NAMESPACE::OperatorSetIdProto& proto)
```
houseroad: remove ONNX_NAMESPACE?

onnx/common/ir.h
```diff
@@ -767,6 +767,56 @@ struct Node : public Attributes<Node> {
...
+    explicit OpSetID(const std::string domain, int64_t version)
+      :domain_(std::move(domain)), version_(version) {}
+
+    std::string toString() const {
+      return "$" + domain_ + "$" + std::to_string(version_);
+    }
+
+    static OpSetID fromString(const std::string& target) {
+      try {
+        std::string new_domain = target.substr(0, target.find("$"));
```
houseroad: what is the format of target? Shall we add it to the comment?
If target starts with "$", this logic does not seem right.

onnx/common/ir.h
houseroad: change the name?

onnx/test/version_converter_test.py
```diff
@@ -0,0 +1,73 @@
...
+            target_version  # type: int
+    ):  # type: (...) -> ModelProto
+        orig_model = helper.make_model(graph, producer_name='onnx-test', opset_imports=[initial_version])
+        # print(type(orig_model))
+        converted_model = onnx.version_converter.convert_version(orig_model,
+                target_version)
+        checker.check_model(converted_model)
+        return converted_model
+
+    # Test 1: Backwards Incompatible Conversion: Add: 8 -> 2
```
houseroad: Update the comment please

onnx/version_converter/BaseConverter.h
houseroad: 2 level domain only works for intra/default domain case. The map and its lookup won't be a performance bottleneck. I would suggest to stick to this design, since later, we may have inter-domain converter.

onnx/common/ir.h
Ac2zoom: More legacy relative to adapters branch

Ac2zoom:
Seems like some issue from autograd is failing the build

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,160 @@
...
+      // No conversion necessary if Model has single, equivalent opset version
+      // if (initial_versions.size() == 1 && initial_versions[0].version ==
+      //    target_version.version && initial_versions[0].domain ==
+      //    target_version.domain) {
+      //  return mp_in;
+      // }
+
+      // Check if target_version is valid
+      const std::unordered_map<std::string, std::pair<int, int>>& versions_map = OpSchemaRegistry::DomainToVersionRange::Instance().Map();
+      const std::string search_domain = target_version.domain() == "ai.onnx" ? "" : target_version.domain();
```
houseroad: directly set ""?

houseroad: actually, please confirm what string ("" or "ai.onnx") we use in `versions_map`

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,160 @@
...
+        "Warning: invalid target_version (must be between %s and %s",
+        version_range.first, version_range.second);
+      // Compile list of all ops used in the model
+      graph_node_list nodes = g->nodes();
+
+      const std::vector<OpSchema>& all_opschemas = OpSchemaRegistry::get_all_schemas_with_history();
+
+      // Create Map for All Versions of format {op_name: {domain: {version: schema}}}
+      std::unordered_map<std::string, std::unordered_map<std::string, std::map<int64_t, const OpSchema*>>>  all_schemas;
+
```
houseroad: You can iterate the schemas here, and maintain a map [opname==> first_version]. With this, you can automatically create and register `NoPreviousVersionAdapter`.

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,160 @@
...
+      for (const Node* op : nodes) {
+        // Iterate through all OperatorSetVersions, select highest that is leq initial_version
+        int64_t op_opset_version = -1;
+        auto& op_domain_map = all_schemas[op->kind().toString()];
+        if (op_domain_map.find(initial_domain) != op_domain_map.end()) {
+          // If op isn't defined for initial domain, we won't convert it
+          for (const auto& version_pair : op_domain_map[initial_domain]) {
+            if (version_pair.first > op_opset_version && version_pair.first <= initial_version.version()) {
+              op_opset_version = version_pair.first;
+              current_opschemas[op] = op_domain_map.at(initial_domain).at(op_opset_version);
```
houseroad: This current_opschemas seems not quite useful.

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+  public:
+    DefaultVersionConverter() {
+      // Register adapters to the version converter
+      registerAdapter(make_unique<NoPreviousVersionAdapter>("Cos",
+        OpSetID(7), OpSetID(6)));
+    }
+
+    ModelProto convert_version(
```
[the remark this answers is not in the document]
Ac2zoom: This seems to introduce an issue with overriding

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,160 @@
...
+      const auto version_it = op_domain_map.find("");
+      return version_it != op_domain_map.end() &&
+          version_it->second.find(curr_version) !=
+          version_it->second.end();
+    }
+
+  public:
+    DefaultVersionConverter() {
+      // Register adapters to the version converter
+      registerAdapter(make_unique<NoPreviousVersionAdapter>("Cos",
```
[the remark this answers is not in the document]
Ac2zoom: The issue is that it's difficult to know which adapters to register for each.  Is there a particular approach for doing this that would be better?

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,160 @@
...
+      const std::string target_domain = target_version.domain();
+
+      ONNX_ASSERTM((initial_domain == "" || initial_domain == "ai.onnx") &&
+        (target_domain == "" || target_domain == "ai.onnx"),
+          "Warning: default onnx version converter can only convert "
+          " between default domain opset versions ('' or 'ai.onnx')\n"
+          "Provided initial_domain: %s, provided target_domain: %s",
+          initial_domain.c_str(), target_domain.c_str());
+
+      ONNX_ASSERTM(initial_domain == target_domain,
```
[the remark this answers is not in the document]
Ac2zoom: Made it easier to debug, but not really, yeah.

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,160 @@
...
+        "initial_version and target_version must have the same domains");
```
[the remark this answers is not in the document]
Ac2zoom: This is done on line 66 of the new convert.cc

onnx/version_converter/convert.h
Ac2zoom: We always use "" in versions_map, so this builds in compatibility for users providing "ai.onnx" in reference to the same domain.

onnx/version_converter/convert.h
Ac2zoom: Perfect. This makes your initial point much clearer.

onnx/version_converter/convert.h
Ac2zoom: The idea is just to store the current version of each op, which was useful before I switched to using since_version.  I will attempt to phase it out.

onnx/version_converter/convert.h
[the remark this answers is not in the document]
Ac2zoom: One thing about this: would it be fair to move the generation of all_schemas to the constructor?

onnx/version_converter/convert.cc
```diff
@@ -0,0 +1,122 @@
...
+  }
+  OpSetID target_struct = OpSetID(target_version);
+  return _version_converter.convert_version(mp_in, initial_struct, target_struct);
+}
+
+ModelProto DefaultVersionConverter::convert_version(
+    const ModelProto& mp_in,
+    const OpSetID& initial_version,
+    const OpSetID& target_version) const {
+  const std::string initial_domain = initial_version.domain();
```
smessmer: you might be able to make these two into references if `.domain()` returns a reference. If it returns by value, then you can't because you'd be keeping a reference to a dead object.

onnx/version_converter/convert.cc
```diff
@@ -0,0 +1,122 @@
...
+    if (it->domain() == "" || it->domain() == "ai.onnx") {
+      initial_struct.setVersion(it->version());
+      break;
+    }
+  }
+  OpSetID target_struct = OpSetID(target_version);
+  return _version_converter.convert_version(mp_in, initial_struct, target_struct);
+}
+
+ModelProto DefaultVersionConverter::convert_version(
```
smessmer: This method is quite long and hard to read. Can we split it into submethods?

onnx/version_converter/convert.cc
```diff
@@ -0,0 +1,122 @@
...
+      domain_index = i;
+    }
+  }
+  while (curr_version != target_version.version()) {
+    debug("curr_version: " + std::to_string(curr_version) + ", next_version: " +
+        std::to_string(curr_version + step));
+    // Iterate through and call adapter returned by adapter_lookup for ops from current_version opset
+    for (Node* op : nodes) {
+      debug(std::string("Finding schema for ") + std::string(op->kind().toString()));
+      const std::string& op_name = op->kind().toString();
```
smessmer: this is a reference to a dead object because `.toString()` returns a value, not a reference.

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,75 @@
...
+      if (DEBUG) std::cerr << str << std::endl;
+    }
+
+  public:
+    DefaultVersionConverter() {
+      const std::unordered_map<std::string, std::pair<int, int>>& versions_map =
+        OpSchemaRegistry::DomainToVersionRange::Instance().Map();
+      version_range = versions_map.at("");
+      // Register adapters to the version converter
+      const std::vector<OpSchema>& all_opschemas =
```
smessmer: reference to dead value, `get_all_schemas_with_history` returns a value, not a reference.

onnx/version_converter/convert.h
```diff
@@ -0,0 +1,75 @@
...
+          version_it->second.end() && up));
+    }
+
+    void debug(const std::string& str) const {
+      if (DEBUG) std::cerr << str << std::endl;
+    }
+
+  public:
+    DefaultVersionConverter() {
+      const std::unordered_map<std::string, std::pair<int, int>>& versions_map =
```
smessmer: also please check here that `.Map()` actually returns a reference. If it doesn't it'd be a reference to a dead value. If it does return a reference, you're fine.

smessmer:
Btw, just for fun, the references to dead values I noted above are actually not causing crashes because C++ has a corner case feature for exactly this (see https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/ ).

However, this feature is little known and comes with a whole bunch of pitfalls that are easy to get wrong, so I strongly recommend against using it.

onnx/common/ir.h
```diff
@@ -875,6 +929,11 @@ friend struct Value;
   const_graph_node_list nodes() const {
     return const_graph_node_list(output_, kNextDirection);
   }
+
+  std::vector<OpSetID>& opset_versions_mutable() {
```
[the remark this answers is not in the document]
Ac2zoom: As @houseroad and I discussed, the challenge with this, as mentioned above, is that it would require significant refactoring of the modification of these members in DefaultVersionConverter.  If it would be particularly important, I can do this.

onnx/version_converter/convert.cc
Ac2zoom: It does, so I will do that!

onnx/common/ir.h
```diff
@@ -767,6 +767,58 @@ struct Node : public Attributes<Node> {
...
+
+  public:
+    explicit OpSetID(const OperatorSetIdProto& proto)
+      :domain_(proto.domain()), version_(proto.version()) {}
+
+    // Default Domain Constructor
+    explicit OpSetID(const int64_t version)
+      :domain_(""), version_(version) {}
+
+    explicit OpSetID(const std::string& domain, int64_t version)
```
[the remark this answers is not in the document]
smessmer: Remember that you have to pass it in by value if you want it to be moved

docs/OpsetVersionConverter.md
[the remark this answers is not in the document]
houseroad: We can add another method in DefaultConverter. Let's keep it as it is in C++ for now and change it in the following PRs.

onnx/common/ir.h
```diff
@@ -770,6 +770,58 @@ struct Node : public Attributes<Node> {
...
+      :domain_(std::move(domain)), version_(version) {}
```
houseroad: Remove the std::move please. It's a const string reference, we are not suppose to change its content.

onnx/common/ir.h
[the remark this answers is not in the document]
houseroad: Yeah, add a note in the comment here, let's do it in the following PR.

onnx/common/ir_pb_converter.cc
```diff
@@ -533,9 +538,62 @@ void encodeGraph(ONNX_NAMESPACE::GraphProto * p_g, const std::shared_ptr<Graph>
...
+  // Add new opset_versions
+  p_m->clear_opset_import();
+  for (const OpSetID& opset : g->opset_versions_mutable()) {
+    OperatorSetIdProto *opset_version_output = p_m->add_opset_import();
+    opset_version_output->set_domain(opset.domain());
+    opset_version_output->set_version(opset.version());
+  }
+}
+
+// TODO (Optimizer): Should this just be copied instead?
```
houseroad: Optimization: probably do it in place

onnx/version_converter/adapters/adapter.h
```diff
@@ -0,0 +1,39 @@
...
+class Adapter {
+  private:
+    std::string name_;
+    OpSetID initial_version_;
+    OpSetID target_version_;
+
+  public:
+    virtual ~Adapter() noexcept = default;
+
+    explicit Adapter(std::string name, OpSetID initial_version, OpSetID target_version)
```
houseroad: Usually, I would prefer "const std::string& name" + name_(name). You will copy it once anyway. Same strategies apply to OpSetID params.

onnx/version_converter/convert.cc
```diff
@@ -0,0 +1,122 @@
...
+      .version() <= version_range.second,
+    "Warning: invalid target_version (must be between %s and %s",
+    version_range.first, version_range.second);
+  // Compile list of all ops used in the model
+  graph_node_list nodes = g->nodes();
+
+  // Iterate over all versions to target_version for specified
+  int64_t curr_version = initial_version.version();
+  int64_t step;
+  bool up;
```
houseroad: We can merge `step` and `up`

onnx/version_converter/convert.cc
houseroad: The sanity check part can be lifted into another function.

houseroad: @Ac2zoom some of the logic can be polished.

onnx/version_converter/convert.h
houseroad: Confirmed, this is fine :-)

onnx/common/ir_pb_converter.cc
Ac2zoom: Realized that this introduces challenges with over-writing existing nodes/inputs/outputs, so creating the new ModelProto makes a lot of sense, rather than messing with large existing ModelProtos.

onnx/common/ir.h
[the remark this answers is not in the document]
smessmer: this is not addressed yet

onnx/version_converter/adapters/adapter.h
```diff
@@ -0,0 +1,39 @@
...
+  private:
+    std::string name_;
+    OpSetID initial_version_;
+    OpSetID target_version_;
+
+  public:
+    virtual ~Adapter() noexcept = default;
+
+    explicit Adapter(const std::string& name, OpSetID initial_version, OpSetID target_version)
+      : name_(name), initial_version_(std::move(initial_version)),
```
smessmer: you can move in `name` too (i.e. pass it in by value and add an std::move)

onnx/version_converter/adapters/no_previous_version.h
```diff
@@ -0,0 +1,20 @@
...
+
+#pragma once
+
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+struct NoPreviousVersionAdapter final : public Adapter {
+  explicit NoPreviousVersionAdapter(const std::string& op_name, const OpSetID
+    initial, const OpSetID target): Adapter(op_name,
```
smessmer: You can move in `op_name` (i.e. pass by value and add an std::move)

Ac2zoom:
@smessmer Will implement your recommended changes on https://github.com/onnx/onnx/pull/1184

PR #1149, opened 2018-06-27
Some changes to enable Clang-cl build for Windows (AND TESTS)

peterjc123:
#1138 with tests for clang-cl builds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -278,6 +278,21 @@ target_include_directories(onnx_proto PUBLIC
   ${PROTOBUF_INCLUDE_DIRS})
 
 if (MSVC)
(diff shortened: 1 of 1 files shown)


peterjc123:
Well, I finally figured out the reason for that crash on `test_transpose_preexisting_incorrect_type`. It supposed to return `RuntimeError` and it was actually returned. But due to the poor support of SEH exception handling in clang-cl, it was not sometimes not caught at the Python side. So there are two solutions:
1. Forcibly compile code that contains pybind11 in MSVC (Bad and complex)
2. Wrap the code entries that throws the exception with a try..catch and then rethrow a Python exception (Good, but need to determine where the call is from)

PR #1154, opened 2018-06-28
Wrapper for vendor-specific ONNXIFI libraries

Maratyszcza:
Add a library (onnxifi.dll/libonnxifi.so/libonnxifi.dylib), which finds and load ONNXIFI implementations in the system, and exposes all their backends under a single interface. With this library, applications can just use
```c
onnxifi_load(ONNXIFI_LOADER_FLAG_VERSION_1_0, NULL, NULL, &library);
```
and get all ONNXIFI implementations in the system.

The libraries are searched in the following directories:
- GNU/Linux: `/usr/lib`
- macOS: `/opt/onnx/lib`
- Windows: TBD, currently not implemented
- Default location can be overriden with CMake configuration-time variable `ONNXIFI_SEARCH_DIR`

Vendor-specific libraries must follow the rules below to be successfully found:
- Library filename must be `onnxifi-${suffix}.dll` (Windows), `libonnxifi-${suffix}.dylib` (Mac), or `libonnxifi-${suffix}.so` (Linux and other OSes)
- The library must implement ONNXIFI functions with `${SUFFIX}` suffix, e.g. `libonnxifi-trt.so` must implement `onnxGetBackendIDsTRT`, `onnxInitGraphTRT`, etc

**TODO**:
- Manually test (I tested a while ago, but the interface changed a lot since then).
- Implement ONNXIFI implementations discovery on Windows.


onnx/onnxifi_wrapper.c
```diff
@@ -0,0 +1,870 @@
...
+  return graph_wrapper->library->onnxSetGraphIO(
+    graph_wrapper->graph,
+    inputsCount,
+    inputDescriptors,
+    outputsCount,
+    outputDescriptors);
+}
+
+ONNXIFI_PUBLIC onnxStatus ONNXIFI_ABI
+  onnxRunGraph(
```
yinghai: What's the usage of this function?

Maratyszcza: Typical use-case:
1. Create `graph` with `onnxInitGraph`
2. Set `inputFence.type = ONNXIFI_SYNCHRONIZATION_EVENT`
3. Create `inputFence.event` with `onnxInitEvent`
4. Set `outputFence.type = ONNXIFI_SYNCHRONIZATION_EVENT`
5. Set `outputFence.event = NULL` (this event will be created by `onnxRunGraph`)
6. Call `onnxRunGraph(graph, &inputFence, &outputFence)`

PR #1155, opened 2018-06-28
Update build from source instructions in README

vchuravy:
I run into issues while trying to build onnx from source because I did not checkout the submodules.


houseroad:
Thanks!

snnn:
It should be 
```
```

PR #1158, opened 2018-06-29
Convtranspose output_shape testcases

zrphercule:
new convtranspose testcases, testing output_shape feature.


houseroad:
rebase to master please :-)

fumihwh:
@zasdfgbnm Any difference between https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/convtranspose.py#L208 ?

zrphercule:
@fumihwh Yes there are three major differences:

1. The old tests implementing models without giving one crucial attribute: kernel_shape. This attr is neccesary in caffe2 now, and we cannot run the test of ConvTranspose without giving it.

2. Since the interface of caffe2 is quite different from onnx in ConvTranspose, we now choose to skip all tests of ConvTranspose: see https://l.facebook.com/l.php?u=https%3A%2F%2Fgithub.com%2Fpytorch%2Fpytorch%2Fblob%2F01f5ee77e3354d6892685309fa302349942eff13%2Fcaffe2%2Fpython%2Fonnx%2Ftests%2Fonnx_backend_test.py%23L40&h=AT0pJmneWCfKyWnGyFsWUxeddf2LIZbMXapgr0Da3W7LH4Fxm-Y7whR8ag1V-m-XZhC_tD_OXZf6DpOk76qrJSxNrgJ5Zibfx6lMwWjMZXezksPbcZhGPy_zg3oUJ3CXPQTgUlHuW50Tv2eLQT1ypzOZac8
Therefore, I wrote this new test for our new version of ConvTranspose. Also see task: T30052256. We have already solved the first two problems in this task and the diffs are waiting to be merged now. The third task has a lot of dependency, and it will take some time before it is solved. This test is designed to test the function of first two problems without triggering the buggy thrid one.

3. This is a model test, and we really do need to implement more model tests for this operator in the future.

fumihwh:
@zrphercule 
As `kernel_shape` is not a required attr, we should not add a particular test case just for one framework to pass their backend test. In other word, if a backend can not pass test in all conditions, it should be fixed first....
So the implementation in Caffe2 could not get `kernel_shape` from input W?

bddppq:
@fumihwh since this is an optional attribute, we should have test cases for both with and without it. Basically in general test cases should cover as many possible valid models as possible.

bddppq:
@zrphercule the cases you have added should be node test cases as they don't have initializers. You should also add test cases that put W as initializers of the model as simple model test case, since that's what normally the real world models will be.

zrphercule:
@bddppq Sure then I will change this test to a node test. I will submit another model test after I fixed the kernel_shape part of ConvTranspose =)

zrphercule:
@onnxbot retest this please

fumihwh:
> I will submit another model test after I fixed the kernel_shape part of ConvTranspose =)

I didn't get `fix the kernel_shape part`...

bddppq:
@fumihwh @zrphercule is referring to extending the caffe2 backend to support "kernel_shape". I believe before we have at least one backend supporting it or have an implementation by ourselves in onnx, there is no easy way to generate the reference inputs & outputs.

fumihwh:
@bddppq onnx-tf backend supports ConvTranspose. About `kernel_shape`, we just use it to check if it is same to shape of weight.

bddppq:
@fumihwh good to know. A minor correction to my previous comment: onnx-caffe2 supports convtranspose with kernel_shape, what @zrphercule is going to work on is when kernel_shape is absent, the converter needs to figure it out by looking into the initializers.

zrphercule:
@bddppq Thanks for your explanation. @fumihwh So far some machine learning frameworks like caffe2 still has no support for dynamic kernal shape in transposed conv operator, which means kernel_shape must be a given attribute when it is initialized. While I am working on this in caffe2 side, I would like to add more test cases of given kernal_shape and other attributes.

fumihwh:
@zrphercule OK, cool.

onnx/defs/nn/defs.cc
```diff
@@ -651,6 +651,9 @@ output_shape can also be explicitly specified in which case pads values are auto
   If (auto_pads != SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)
   Else: pads[start_i] = total_padding[i] - (total_padding[i]/2); pads[end_i] = (total_padding[i]/2).
 
+Notice although we define parameter pads as the input padding,
+for ease of calculation, we used following transformation for given input padding pads' before calculation:
+  pads = pads' - k - 1
```
bddppq: Sorry I don't quite understand this line, could you elaborate?

onnx/backend/test/case/node/convtranspose.py
```diff
@@ -279,3 +279,51 @@ def export_convtranspose_pads():  # type: () -> None
...
+                        [3., 3., 7., 4., 9., 5., 5., 0.],
+                        [3., 3., 7., 4., 9., 5., 5., 0.],
+                        [3., 3., 7., 4., 9., 5., 5., 0.],
+                        [6., 6., 13., 7., 15., 8., 8., 0.],
+                        [6., 6., 13., 7., 15., 8., 8., 0.],
+                        [6., 6., 13., 7., 15., 8., 8., 0.],
+                        [0., 0., 0., 0., 0., 0., 0., 0.]]]]).astype(np.float32)
+
+        expect(node, inputs=[x, W], outputs=[y],
+               name='test_convtranspose_with_kernel')
```
bddppq: you need to add kernel_shape attribute to the expect call here (kernel_shape=`W.shape`)

bddppq: actually you can just add an additional `expect` call to existing test cases (with kernel_shape attribute provided), so we can avoid copy-paste these pre-computed reference inputs & outputs.

zrphercule: Yes I was thinking if it is a better way...

onnx/defs/nn/defs.cc
zrphercule: @houseroad and I discussed about our definition of "output_pads" and "pads", and found that both of them are output padding, although we claim this "pads" is input padding. In order to translate this "pads" to input padding, we need this formula. 
See http://deeplearning.net/software/theano/tutorial/conv_arithmetic.html#transposed-convolution-arithmetic Relationship 8

onnx/backend/test/case/node/convtranspose.py
zrphercule: @bddppq btw I wonder why we need to add "kernel_shape = W.shape" here, since I have already added it to the node attribute.

zrphercule:
@fumihwh @bddppq Since test_convtranspose_output_shape, test_convtranspose_pad and the new added test_convtranspose_kernel_shape are using the same test data, I merged these three test into one function instead of three.

bddppq:
There are python lint error:
```
./onnx/backend/test/case/node/convtranspose.py:207:9: E303 too many blank lines (2)
```

onnx/defs/nn/defs.cc
```diff
@@ -651,7 +651,9 @@ output_shape can also be explicitly specified in which case pads values are auto
   If (auto_pads != SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)
   Else: pads[start_i] = total_padding[i] - (total_padding[i]/2); pads[end_i] = (total_padding[i]/2).
 
-    )DOC";
+The pads parameter represents the size of padding we need to add at both sides of output tensor;
```
bddppq: the `pads` are the number of pixels going to be **subtracted** in the output shape, while `output_padding` is going to be **added** at (only the end side) in the output shape.
Also worth noting is the number of `pads` attribute is 2x of the rank of the output tensor while for `output_padding` is equal to the rank of the output.

onnx/backend/test/case/node/convtranspose.py
bddppq: @zrphercule you are right, it should be in the node.

shinh:
Not directly related to this PR, but let me ask a question. Is `output_shape` of the original test, test_convtranspose_output_shape, right? The original test is passing [1,2,10,8], which is the entire shape of the output while the new test's output_shape is [10,8]. Reading the document, [1,2,10,8] seems incorrect, but I'm not sure.

PR #1174, opened 2018-07-07
Add a function for accessing the underlying data pointer in the Tensor class

bddppq:


onnx/common/tensor.h
```diff
@@ -62,6 +65,15 @@ struct Tensor final {
     return sizes_;
   }
 
+  int64_t size_from_dim(int dim) const {
+    static std::multiplies<int64_t> mul{};
```
yinghai: Does this actually same a copy each time we call this function?

bddppq: I put it as function static, so there should be only one copy got instantiated

yinghai: Yeah, it's inlined: https://godbolt.org/g/iGp6QB

:)

bddppq: wow good point! updated

PR #1177, opened 2018-07-09
Replacements for Loop and LoopIndexTensor

gramalingam:
This PR proposes a replacement of the current experimental Loop and LoopIndexTensor with a Scan and ScanWhile operator that are intended to provide a similar functionality. The motivation for this is that LoopIndexTensor is not well-defined as it exists (as it is intended to represent essentially the current value of "iterator", so to speak, in an iteration along some axis of a tensor). We have two alternatives: (a) use a general-purpose loop with index variables that can be used to index into tensors using well-defined operators, and (b) use a special-purpose scan loop that hides all of the index variables, their iteration, etc.  This PR proposal is along the lines of (b) to allow it to be efficiently implemented (using techniques that do not apply to general-purpose loops), following the intention of the original Loop and LoopIndexTensor.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -2311,208 +2311,6 @@ This version of the operator has been available since version 1 of the default O
 <dd>Constrain input and output types to float tensors.</dd>
 </dl>
 
-### <a name="Loop-1"></a>**Loop-1**</a>
-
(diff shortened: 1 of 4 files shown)


gramalingam:
Will update this as per discussion in the ONNX workshop.

Issue #1172, opened 2018-07-07
Why Min/Max supports 1 input?


kit1980:
What's the purpose of having element-wise Min/Max supporting 1 input?

houseroad:
Element-wise `Min` and `Max` with one input just behave as `Identify` node. It's easy for the frontend. However, not that many frameworks support that. We can have optimization passes to eliminate such operators.

PR #1179, opened 2018-07-09
Fix the input numbers for max, min, mean, sum

houseroad:
Like mentioned in https://github.com/onnx/onnx/issues/1172, Operator `Max`, `Min, `Sum`, and `Mean` should at least have two inputs. Even numpy does not support one input case. I would consider this as a schema bug, let's fix it.


houseroad:
Need further discussion. Close for now

PR #1183, opened 2018-07-10
pybind11 extension code should be compiled with hidden visibility

bddppq:
http://pybind11.readthedocs.io/en/stable/faq.html#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes


raymondxyang:
I actually tried https://github.com/silx-kit/pyFAI/pull/663/commits/a03f2bafe9ca963c4fbe7faadb7c38cb892e92c8#diff-2eeaed663bd0d25b7e608891384b7298 but the wheel still cause seg fault with caffe2.. but is it definitely a right practice to hide the symbols

bddppq:
@raymondxyang I found a similar issue in pybind11's repo https://github.com/pybind/pybind11/issues/1262, which says two pybind11 extensions built with different compilers can work well together. although we think as long as both extensions are built with gcc >= 4.9 then it should be fine. IIRC onnx wheels are built with gcc 4.9?

bddppq:
@raymondxyang damm alright then we need to update it to >= 4.9 :-)

raymondxyang:
https://github.com/pypa/manylinux/issues/118 I guess they are intentionally doing that for supporting centos5..

bddppq:
@raymondxyang IIUC from the post it says they initially chose to use gcc4.8 because they couldn't find newer version of gcc on centos5, however the follow up threads say they have built newer gcc for centos5

raymondxyang:
Agree.. I ll try updating it to 4.9 gcc to see if it can be solved

bddppq:
test failure is irrelevant (flaky appveyor infra)

raymondxyang:
Tried to upgrade to gcc >= 4.9. No official distribution found but used https://github.com/squeaky-pl/centos-devtools/releases . now pip install hangs during copying test files. Any ideas?
https://travis-ci.org/raymondxyang/wheel-builder/builds/405116217

PR #1185, opened 2018-07-11
[ONNXIFI] Use event handle in the memory fence structure rather than …

Maratyszcza:
…a pointer to even handle

Indirection through a pointer doesn't serve any purpose, and complicates implementation. I don't remember why I choose to use a pointer, and likely didn't put enough thought into this decision.

### DIFF
--- a/onnx/onnxifi.h
+++ b/onnx/onnxifi.h
@@ -596,13 +596,13 @@ typedef struct onnxMemoryFence {
   onnxEnum type;
   union {
     /**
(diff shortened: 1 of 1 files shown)


houseroad:
It seems that we also need to adjust onnxifi_op in Caffe2.

yinghai:
> It seems that we also need to adjust onnxifi_op in Caffe2.

Yes, I'm sending a PR.

PR #1188, opened 2018-07-12
[Minor] Add -pthread in compile time for multi-threading uses

raymondxyang:
A general good practice when use gcc per https://stackoverflow.com/questions/2127797/significance-of-pthread-flag-when-compiling @snnn


raymondxyang:
will reopen after changes

snnn:
find_package(Threads) doesn't add '-pthreads' to compiler flags.

Maratyszcza:
@snnn You need to add it to dependencies of the target.

snnn:
I have to say, compare to autoconf, The quality of cmake's FindXXX is too low.

This is how we do the same thing in autoconf:
http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_pthread.m4

And this is how cmake works:
https://github.com/Kitware/CMake/blob/master/Modules/FindThreads.cmake

PR #1198, opened 2018-07-14
Expose onnx-operators protobuf structs in Python

bddppq:


bddppq:
seems to be related to relative import in python3
upstream protobuf issue https://github.com/google/protobuf/issues/1491

raymondxyang:
see https://github.com/onnx/onnx/pull/1112/files#diff-2eeaed663bd0d25b7e608891384b7298R96 for a temp solution.. but the CI for python3 on mac always reported errors:
```
File "setup.py", line 100, in correct_import

    line.rstrip('\n')))

  File "/Users/travis/virtualenv/lib/python3.7/re.py", line 192, in sub

    return _compile(pattern, flags).sub(repl, string, count)

  File "/Users/travis/virtualenv/lib/python3.7/re.py", line 309, in _subx

    template = _compile_repl(template, pattern)

  File "/Users/travis/virtualenv/lib/python3.7/re.py", line 300, in _compile_repl

    return sre_parse.parse_template(repl, pattern)

  File "/Users/travis/virtualenv/lib/python3.7/sre_parse.py", line 1024, in parse_template

    raise s.error('bad escape %s' % this, len(this))

re.error: bad escape \x at position 73

```

could you help me look into it?  Thanks :P

bddppq:
@raymondxyang should now be fixed by using absolute import path in our proto files

raymondxyang:
Will merge it if there is no further comments today

PR #1200, opened 2018-07-15
Fix an error in GRU formula

tengyifei:


tengyifei:
@houseroad done and regenerated docs

PR #1203, opened 2018-07-17
Load ONNXIFI libraries with RTLD_LOCAL and remove suffix from functions

Maratyszcza:
We previously planned to load ONNXIFI libraries with `RTLD_GLOBAL` so that implementations relying on operator registration could accept new registered ops. However, we found that it would introduce more problems than it solves, in particular when implementations compiled with different compilers and distributed in binary are used together in one process. This diff changes `dlopen` flags to `RTLD_LOCAL` and removes suffixes in ONNXIFI implementations (suffixes were needed to avoid symbol conflict when loading several implementations with `RTLD_GLOBAL`).


Maratyszcza:
cc @varunjain99

onnx/onnxifi.h
```diff
@@ -47,16 +47,6 @@ extern "C" {
   #endif
 #endif
 
-#if defined(ONNXIFI_LIBRARY_SUFFIX)
-#define ONNXIFI_SYMBOL_CONCAT_(prefix, suffix) prefix##suffix
```
yinghai: Why do we remove suffix?

Maratyszcza: Suffix was needed to avoid symbol conflict where several ONNXIFI implementations are loaded with `RTLD_GLOBAL` within the same process. As we are no longer using `RTLD_GLOBAL`, and suffix didn't have any other purpose, its best to remove it to simplify loader, wrapper, implementations and users of ONNXIFI.

rdzhabarov:
Curious how pytorch depends on the particular version of onnxifi interface? Is that based on some good known SHA or it just uses whatever is on the latest onnx master?

Also, what's the policy for breaking changes in ONNXIFI?

Maratyszcza:
@rdzhabarov We are preparing to finalize ONNXIFI, and thats why there are some breaking changes. After ONNXIFI is finalized, we'd have to preserve stable API and ABI.

Issue #417, opened 2018-01-12
MaxPool return indices


ezyang:
A user here (https://discuss.pytorch.org/t/problems-converting-pytorch-model-into-onnx/12192?u=pengfei_wang) requested MaxPool with remembered indices, which ONNX doesn't support. Might be worth looking into.

ebarsoum:
Is that needed for backprop?

jendrikjoe:
Hi @ebarsoum,
it is needed for some people working with neural networks that down- and later upsample data, e.g. segmentation networks :)

fumihwh:
@ebarsoum It's something like [`tf.nn.max_pool_with_argmax`](https://www.tensorflow.org/api_docs/python/tf/nn/max_pool_with_argmax).

pgtgrly:
I am trying to export an Unet-like architecture trained on Pytorch and I am facing the same issue. Any updates on the same?

BenjaminBout:
I also have the same issue, any updates, workarounds ?

houseroad:
Look into the issue

houseroad:
@pgtgrly @BenjaminBout @simasima121 do you guys also need `MaxUnpool`?

houseroad:
https://github.com/onnx/onnx/issues/1217 is the follow up.

Margrate:
x1d = self.unpool1(drop_conv_2, id1)


    w.add_graph(parsing, (image, ))
    self.file_writer.add_graph(graph(model, input_to_model, verbose))
  File "python3.6/site-packages/tensorboardX/pytorch_graph.py", line 96, in graph
    torch.onnx._optimize_trace(trace, torch._C._onnx.OperatorExportTypes.ONNX_ATEN_FALLBACK)
  File "python3.6/site-packages/torch/onnx/__init__.py", line 41, in _optimize_trace
    trace.set_graph(utils._optimize_graph(trace.graph(), operator_export_type))
  File "python3.6/site-packages/torch/onnx/utils.py", line 107, in _optimize_graph
    graph = torch._C._jit_pass_onnx(graph, operator_export_type)
RuntimeError: symbolic for max_pool2d_with_indices returned None for the output 1 (indicating conversion for that particular output is not supported), but the network uses this output later

PR #1206, opened 2018-07-18
Support output indices in MaxPool op

houseroad:
Address https://github.com/onnx/onnx/issues/417

However, MaxUnpool is also missing. We probably should add it, too (in a different PR).

### DIFF
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -29,6 +29,14 @@ void convPoolTypeAndShapeInference(
     bool use_dilation,
     bool require_kernel_shape) {
   propagateElemTypeFromInputToOutput(ctx, 0, 0);
+  if (ctx.getNumOutputs() > 1) {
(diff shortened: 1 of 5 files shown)


onnx/defs/nn/defs.cc
```diff
@@ -257,6 +257,30 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "MaxPool",
+            "max",
+            "The output of each pooling window is maximum number of elements exclude pad."))
+        .Output(
+            1,
+            "Indices",
+            "Indices tensor from max pooling across the input tensor. "
+            "The dimensions of indices are the same as output tensor. "
+            "The values in indices of are the indices of the selected values during pooling. "
+            "The indices are computed as flatten 1-D tensor, "
```
ebarsoum: Why don't we make the output indices similar to TopK. Non flatten? To avoid row vs column major?

PR #1212, opened 2018-07-19
Enable cpp text by setting ONNX_BUILD_TESTS to "ON"

zrphercule:
The first step of making onnxifi cpp test.


bddppq:
As discussed offline, let's it turn in on in the CI instead of be default in cmake.

PR #1219, opened 2018-07-23
add low-precision support for conv, reducesum and matmul

linkerzhang:


onnx/defs/nn/defs.cc
```diff
@@ -1229,4 +1225,91 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "dilations",
+            "dilation value along each axis of the filter. If not present, the dilation defaults to 1 along each axis.",
+            AttributeProto::INTS,
+            OPTIONAL)
+        .Attr(
+            "strides",
+            "Stride along each axis. If not present, the stride defaults to 1 along each axis.",
+            AttributeProto::INTS,
+            OPTIONAL)
+        .Attr(
```
ebarsoum: auto_pad is deprecated in the standard conv, so it shouldn't be there.

linkerzhang: ok. was thinking to remove it in separate PR for both conv and conv_integer later. will remove it right now.

linkerzhang:
About having a new op for low-precision, there's another way of doing that without new op,

Having the original op’s input and output type specified as T1 and T2 (instead of T for both), and change the type inference function to,
1.	T2 always the same as T1 if T1 is NOT low-precision type.
2.	T2 is int32 if T1 is low-precision type (int8, uint8, int16 or uint16).

Which one do you guys prefer, use another op name or make the type inference function more complex as above?

docs/Changelog.md
```diff
@@ -8024,6 +7996,92 @@ This version of the operator has been available since version 7 of the default O
 </dl>
 
 ## Version 8 of the default ONNX operator set
+### <a name="Conv_Integer-8"></a>**Conv_Integer-8**</a>
+
+  The integer convolution operator consumes an input tensor, a filter and a padding value, and computes the output. The production MUST never overflow. The accumulation may overflow if and only if in 32 bits.
```
Maratyszcza: If inputs are 16 bits wide, the product will, generally, overflow. We should either remove 16-bit input types, specify overflow behavior, or specify 64-bit output for 16-bit inputs.

Maratyszcza:
Currently no operators in ONNX have an underscore in the name, any reasons why these are not named `ConvInteger`, etc?

docs/Changelog.md
linkerzhang: Let's remove 16 bits for now then.

onnx/defs/nn/defs.cc
linkerzhang: Think it twice, I'll still have the auto_pad removal in a separate PR for both conv and conv_integer soon.

linkerzhang:
@Maratyszcza  any more comments please?

bddppq:
@onnxbot retest this please

onnx/defs/reduction/defs.cc
```diff
@@ -6,6 +6,44 @@
...
+    axes.assign(axes_proto->ints().begin(), axes_proto->ints().end());
+
+  for (size_t i = 0; i < axes.size(); ++i) {
+    if (axes[i] < 0)
+      axes[i] += input_ndim;
+  }
+  // do we need handle negative axis?
+  for (int i = 0; i < input_ndim; ++i) {
+    // axes empty means reduce all dim
+    if (!axes.empty() && std::find(axes.begin(), axes.end(), i) == axes.end()) {
```
bddppq: Is `axes` required to be sorted?

onnx/defs/reduction/defs.cc
```diff
@@ -6,6 +6,44 @@
 #include "onnx/defs/schema.h"
 
 namespace ONNX_NAMESPACE {
+void reduceShapeInference(InferenceContext& ctx) {
+  if (!hasNInputShapes(ctx, 1)) {
+    return;
+  }
+
+  int64_t keep_dims = 1;
```
bddppq: you should get the default value from the schema

onnx/defs/reduction/defs.cc
```diff
@@ -6,6 +6,44 @@
...
+  auto attr_proto = ctx.getAttribute("keepdims");
+  if (attr_proto) {
+    keep_dims = attr_proto->i();
+  }
+  auto& input_shape = ctx.getInputType(0)->tensor_type().shape();
```
bddppq: const

onnx/defs/reduction/defs.cc
```diff
@@ -36,46 +74,44 @@ False instead of True.)DOC";
...
+        "Keep the reduced dimension or not, default 1 mean keep reduced dimension.",
+        AttributeProto::INT,
+        static_cast<int64_t>(1));
+    schema.Input(0, "data", "An input tensor.", "T");
+    schema.Output(0, "reduced", "Reduced output tensor.", "tensor(int32)");
+    schema.TypeConstraint(
+        "T",
+        {"tensor(int8)", "tensor(uint8)"},
+        "Constrain input type to 8-bits integer tensors.");
+    schema.TypeAndShapeInferenceFunction(reduceShapeInference);
```
bddppq: Set the output type to be int32

onnx/defs/nn/defs.cc
```diff
@@ -1229,4 +1225,84 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            {"tensor(int8)",
+             "tensor(uint8)"},
+            "Constrain input X and Z data types as 8-bits integer tensors.")
+        .TypeConstraint(
+            "T2",
+            {"tensor(int8)",
+             "tensor(uint8)"},
+            "Constrain input W data types as 8-bits integer tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          convPoolShapeInference(ctx, true, false);
```
bddppq: set output type

onnx/defs/nn/defs.cc
```diff
@@ -1229,4 +1225,84 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "tensor(int32)")
+        .TypeConstraint(
+            "T1",
+            {"tensor(int8)",
+             "tensor(uint8)"},
+            "Constrain input X and Z data types as 8-bits integer tensors.")
+        .TypeConstraint(
+            "T2",
+            {"tensor(int8)",
+             "tensor(uint8)"},
```
bddppq: dumb question: can X and W have different input types?

onnx/defs/nn/defs.cc
```diff
@@ -1229,4 +1225,84 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "W",
+            "The weight tensor that will be used in the "
+            "convolutions; has size (M x C x kH x kW), where C "
+            "is the number of channels, and kH and kW are the "
+            "height and width of the kernel, and M is the number "
+            "of feature maps. For more than 2 dimensions, the "
+            "kernel shape will be (M x C x k1 x k2 x ... x kn), "
+            "where is the dimension of the kernel",
+            "T2")
+        .Input(
```
bddppq: Wait padding become inputs now?

bddppq: what about bias?

onnx/defs/reduction/defs.cc
linkerzhang: output type is already int32 specified in type constraints, don't need to inferred in this function.

onnx/defs/nn/defs.cc
linkerzhang: output type is already int32 specified in type constraints, don't need to inferred in this function.

onnx/defs/reduction/defs.cc
bddppq: You do need. Shape/type inference is different with your type constraints (e.g. shape inference engine will put the inferred information into the model as ValueInfoProto).

onnx/defs/nn/defs.cc
linkerzhang: they can.

onnx/defs/nn/defs.cc
linkerzhang: yes, padding value is indeed the zero point of X, which is input.

linkerzhang: bias is kept to be separate as fusing it together with conv asks its scale and zero point to meet some equation. meanwhile, bias quantization is not that critical as input X and kernel W.

linkerzhang: so, no bias in convinteger here.

onnx/defs/reduction/defs.cc
linkerzhang: Good suggestion, however, I don't think the default value in opschema was designed to be accessible here. will discuss with @anderspapitto in separate thread and make it if doable in separate PR.

onnx/defs/nn/defs.cc
liwchang: I think yes. There is a popular one, in which one is int8 and another is uint8. I think it is from Intel

onnx/defs/reduction/defs.cc
linkerzhang: it's not needed based on current design and implementation.

onnx/defs/reduction/defs.cc
linkerzhang: why? type constraints is saying the output type is int32 only, it does not need any type inference. Type inference done by this function is when there're multiple options in output type constraints and no way to infer it with type constraints only. For example, input type is T1 and output type is T2, and T2 could be int32, double. From constraints, it's not doable to get to know which output type should be picked up. In the cases that output type constraint only have one (this case) or input and output types are the same (with type constraints specified both as "T"), type inference function is not needed. Make sense?

onnx/defs/nn/defs.cc
linkerzhang: @liwchang , if it's not check-able, then we should not add it. I assume that it will introduce ambiguity and correctness bugs.

docs/Changelog.md
```diff
@@ -8024,6 +7996,92 @@ This version of the operator has been available since version 7 of the default O
 </dl>
 
 ## Version 8 of the default ONNX operator set
+### <a name="ConvInteger-8"></a>**ConvInteger-8**</a>
+
+  The integer convolution operator consumes an input tensor, a filter and a padding value, and computes the output. The production MUST never overflow. The accumulation may overflow if and only if in 32 bits.
```
Maratyszcza: `The production MUST never overflow. The accumulation may overflow if and only if in 32 bits.`
An ONNX backend can't guarantee that result never overflows (which is implied by "MUST"). IMO, a better wording would be "If the any intermediate computation overflows 32-bit integer, the result is implementation-defined"

docs/Changelog.md
```diff
@@ -8024,6 +7996,92 @@ This version of the operator has been available since version 7 of the default O
...
+
+#### Inputs (2 - 3)
+
+<dl>
+<dt><tt>X</tt> : T1</dt>
+<dd>Input data tensor from previous layer; has size (N x C x H x W), where N is the batch size, C is the number of channels, and H and W are the height and width. Note that this is for the 2D image. Otherwise the size is (N x C x D1 x D2 ... x Dn)</dd>
+<dt><tt>W</tt> : T2</dt>
+<dd>The weight tensor that will be used in the convolutions; has size (M x C x kH x kW), where C is the number of channels, and kH and kW are the height and width of the kernel, and M is the number of feature maps. For more than 2 dimensions, the kernel shape will be (M x C x k1 x k2 x ... x kn), where is the dimension of the kernel</dd>
+<dt><tt>Z</tt> (optional) : T1</dt>
+<dd>padding value (zero_point normally).</dd>
```
Maratyszcza: If padding value is `zero_point`, we must require that the same `X` always use the same `zero_point` value. Otherwise, the same quantized tensor could have several dequantized representations.

docs/Changelog.md
```diff
@@ -8024,6 +7996,92 @@ This version of the operator has been available since version 7 of the default O
...
+</dl>
+
+#### Outputs
+
+<dl>
+<dt><tt>Y</tt> : tensor(int32)</dt>
+<dd>Output data tensor that contains the result of the convolution. The output dimensions are functions of the kernel size, stride size, and pad lengths.</dd>
```
Maratyszcza: Should we require `uint32` output types when both inputs are `uint8`?

Maratyszcza:
The spec should specify whether the result is exact or approximate. Note that targeting exact result would exclude popular implementations for x86, e.g. see the paper by Vanhoucke in the comments

docs/Changelog.md
linkerzhang: Give output data type is 32 bits, we're assuming that production between 2 8-bits should never overflow. Make sense? if there're exceptional cases, the behavior not specified will be undefined by default.

docs/Changelog.md
linkerzhang: zero point is dynamic in our design, thus, it should be flexible although same zero point may be used for same X, but it's not "MUST".

docs/Changelog.md
linkerzhang: make sense, I'll make the change accordingly.

docs/Changelog.md
```diff
@@ -8024,6 +7996,92 @@ This version of the operator has been available since version 7 of the default O
...
+</dl>
+
+#### Type Constraints
+
+<dl>
+<dt><tt>T1</tt> : tensor(int8), tensor(uint8)</dt>
```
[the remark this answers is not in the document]
linkerzhang: thanks a lot for great comments! do we want to filter the int8 X + uint8 W case specifically please? I'd suggest we don't need to filter it out.

docs/Changelog.md
```diff
@@ -8186,6 +8242,49 @@ This version of the operator has been available since version 8 of the default O
...
+
+<dl>
+<dt><tt>data</tt> : T</dt>
+<dd>An input tensor.</dd>
+</dl>
+
+#### Outputs
+
+<dl>
+<dt><tt>reduced</tt> : tensor(int32)</dt>
```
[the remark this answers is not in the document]
linkerzhang: fixed.

docs/TestCoverage.md
```diff
@@ -5609,6 +5609,9 @@ expect(node, inputs=[x, y], outputs=[z],
 <br/>
 
 ## &#x1F494;No Cover Common Operators
+### ConvInteger (call for test cases)
```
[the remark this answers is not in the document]
linkerzhang: Functionality of these ops is the same as their original ones (ones for full-precision). Only differences are data types, which are defined by type constraints and type inference function. besides adding same node test cases, any other test cases suggested to add please?

onnx/defs/math/defs.cc
```diff
@@ -734,6 +729,76 @@ static const char* MatMul_ver6_doc = R"DOC(
 Matrix product that behaves like numpy.matmul: https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.matmul.html
 )DOC";
 
+void malmulShapeInference(InferenceContext& ctx) {
```
bddppq: typo: malmul -> matmul

docs/TestCoverage.md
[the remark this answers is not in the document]
linkerzhang: am adding test data for convinteger. cases for the other two will be added separately. For the type checking, it will need some test API change per discussion with @houseroad  offline.

onnx/defs/math/defs.cc
linkerzhang: fixed.

onnx/defs/math/defs.cc
```diff
@@ -734,6 +729,76 @@ static const char* MatMul_ver6_doc = R"DOC(
...
+  if (shape0.dim_size() == 0 || shape1.dim_size() == 0) {
+    fail_shape_inference("Input tensors of wrong rank (0).");
+  }
+
+  TensorShapeProto shapeL, shapeR;
+
+  // First promote each shape to at least rank-2. This logic is
+  // specific to matmul, not generic broadcasting.
+  {
+    if (shape0.dim_size() == 1) {
```
houseroad: If shape0.dim_size() == 1... it seems an incorrect case. Either empty or at least rank 2, right?

onnx/defs/math/defs.cc
```diff
@@ -734,6 +729,76 @@ static const char* MatMul_ver6_doc = R"DOC(
...
+      *shapeR.add_dim() = shape1.dim(0);
+      shapeR.add_dim()->set_dim_value(1);
+    } else {
+      *shapeR.mutable_dim() = shape1.dim();
+    }
+  }
+
+  // Check for compatible matrix multiply dimensions
+  {
+    auto dimL = shapeL.dim(shapeL.dim_size() - 1);
```
houseroad: Nit: commonDimL?

onnx/backend/test/case/node/convinteger.py
```diff
@@ -0,0 +1,116 @@
...
+from ..base import Base
+from . import expect
+
+
+class ConvInteger(Base):
+
+    @staticmethod
+    def export():  # type: () -> None
+
+        x = np.array([[[[0, 1, 2, 3, 4],  # (1, 1, 5, 5) input tensor
```
houseroad: Can use some other case (not symmetric tensor), at least to tell people this is a co-relation, not a real convolution?

onnx/defs/math/defs.cc
linkerzhang: in this case, a dimension with value "1" was added to ensure its rank is greater than 2. This is the old logic (I didn't touch it) implemented by @anderspapitto 

@anderspapitto  may share more details on why we want to support this case.
Thanks!

onnx/defs/math/defs.cc
linkerzhang: This is not that common, I think :), so will keep the name. Again, I didn't touch this function, but make it to be used by both matmul and matmulinteger.

onnx/backend/test/case/node/convinteger.py
linkerzhang: good point. changed  two cases.

linkerzhang:
@bddppq @Maratyszcza @houseroad  any more comments please? thank you!

docs/Changelog.md
```diff
@@ -8024,6 +7996,92 @@ This version of the operator has been available since version 7 of the default O
...
+
+<dl>
+<dt><tt>Y</tt> : tensor(int32)</dt>
+<dd>Matrix multiply results from A * B</dd>
+</dl>
+
+#### Type Constraints
+
+<dl>
+<dt><tt>T1</tt> : tensor(int8), tensor(uint8)</dt>
```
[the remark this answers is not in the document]
linkerzhang: fixed.

docs/Changelog.md
Maratyszcza: Ok, I see, it would be better to phrase the first sentence as `The multiplication never overflows`. As for the second part, lets not introduce any undefined behavior, it is a horrible C++ concept that even C++ standardization committee tries to eradicate. It would suggest "If the any intermediate computation overflows 32-bit integer, the result is implementation-defined": this hints that operation must still produce some result (and not e.g. crash the whole system, which would be OK with undefined behavior), but this results depends on the backend implementation and is generally not portable.

docs/Changelog.md
Maratyszcza: @linkerzhang if you want to implement quantized operators/functions on top of ConvInteger/MatMulInteger, it needs to be the same. beware that backends which implement quantized operators directly may use a different internal representation (e.g. store everything as 9-bit signed integers instead of 8-bit unsigned + zero point). using the same `uint8` tensor with different `zero_point` values would make it impossible to use quantized operators with backends using any representation, but canonical `uint8` + `zero_point` + `scale`.

docs/Changelog.md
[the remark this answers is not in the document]
Maratyszcza: @linkerzhang is there a use-case for it?

docs/TestCoverage.md
```diff
@@ -1142,6 +1142,121 @@ expect(node_with_asymmetric_padding, inputs=[x, W], outputs=[y_with_asymmetric_p
 </details>
 
 
+### ConvInteger
+There are 2 test cases, listed as following:
```
Maratyszcza: Thanks for adding these @linkerzhang

prasanthpul:
superseded by https://github.com/onnx/onnx/pull/1908

PR #1224, opened 2018-07-24
Add version tags to tensor descriptor and memory fence structures

Maratyszcza:
As per discussion with WinML, this PR adds version tags to tensor descriptor and memory fence structures, to ensure that we can extend/revise these structures in the future without losing compatibility with existing backends.

Attn @yinghai, @rdzhabarov, @varunjain99, @zrphercule: this is a breaking change.

### DIFF
--- a/onnx/onnxifi.h
+++ b/onnx/onnxifi.h
@@ -112,32 +112,33 @@ typedef uint64_t onnxPointer;
 
 #define ONNXIFI_STATUS_SUCCESS 0x0000
 #define ONNXIFI_STATUS_FALLBACK 0x0001
-#define ONNXIFI_STATUS_INVALID_ID 0x0010
-#define ONNXIFI_STATUS_INVALID_SIZE 0x0011
-#define ONNXIFI_STATUS_INVALID_POINTER 0x0012
-#define ONNXIFI_STATUS_INVALID_PROTOBUF 0x0013
-#define ONNXIFI_STATUS_INVALID_MODEL 0x0014
-#define ONNXIFI_STATUS_INVALID_BACKEND 0x0015
-#define ONNXIFI_STATUS_INVALID_GRAPH 0x0016
-#define ONNXIFI_STATUS_INVALID_EVENT 0x0017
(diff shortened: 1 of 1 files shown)


onnx/onnxifi.h
```diff
@@ -127,9 +127,10 @@ typedef uint64_t onnxPointer;
 #define ONNXIFI_STATUS_UNSUPPORTED_PARAMETER 0x0022
 #define ONNXIFI_STATUS_UNSUPPORTED_SHAPE 0x0023
 #define ONNXIFI_STATUS_UNSUPPORTED_DATATYPE 0x0024
-#define ONNXIFI_STATUS_UNIDENTIFIED_NAME 0x0025
-#define ONNXIFI_STATUS_MISMATCHING_SHAPE 0x0026
-#define ONNXIFI_STATUS_MISMATCHING_DATATYPE 0x0027
+#define ONNXIFI_STATUS_UNSUPPORTED_TAG 0x0025
```
bddppq: What's reason to reuse 0x0025 here while skipping 0x0026 and 0x0027?

onnx/onnxifi.h
```diff
@@ -1467,6 +1511,10 @@ ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
  *                                        NULL.
  * @retval ONNXIFI_STATUS_INVALID_GRAPH The function call failed because
  *                                      graph is not an ONNXIFI graph handle.
+ * @retval ONNXIFI_STATUS_INVALID_TAG The function call failed because a tag
```
bddppq: `ONNXIFI_STATUS_UNSUPPORTED_TAG`?

onnx/onnxifi.h
Maratyszcza: To add space for `ONNXIFI_STATUS_UNSUPPORTED_something` codes I'll add later

Maratyszcza: Good point! Added more space in there

PR #1225, opened 2018-07-24
Add ONNXIFI_BACKEND_PROPERTY_LOG_LEVEL backend initialization property

Maratyszcza:
Most deep learning frameworks and higher-level software let users specify verbosity level. The new ONNXIFI_BACKEND_PROPERTY_LOG_LEVEL property lets a deep learning framework convey this information to an ONNXIFI backend.

### DIFF
--- a/onnx/onnxifi.h
+++ b/onnx/onnxifi.h
@@ -476,24 +476,67 @@ typedef int32_t onnxBackendInfo;
  *     ONNXIFI_OPTIMIZATION_LOW_DELAY
  */
 #define ONNXIFI_BACKEND_PROPERTY_OPTIMIZATION 1
+/**
+ * Logging verbosity level for the backend.
+ *
+ * If this property is not specified during initialization, the backend should
+ * assume ONNXIFI_LOG_LEVEL_WARNING logging verbosity level.
+ *
+ * Possible values:
+ *     ONNXIFI_LOG_LEVEL_ERROR
+ *     ONNXIFI_LOG_LEVEL_WARNING
+ *     ONNXIFI_LOG_LEVEL_INFO
+ *     ONNXIFI_LOG_LEVEL_DEBUG
+ */
+#define ONNXIFI_BACKEND_PROPERTY_LOG_LEVEL 2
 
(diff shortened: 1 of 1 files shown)


Maratyszcza:
We are going to have another header file for extensions (backend-specific options), but this option is intended to be standard.

onnx/onnxifi.h
```diff
@@ -493,6 +506,30 @@ typedef int32_t onnxBackendInfo;
  */
 #define ONNXIFI_OPTIMIZATION_LOW_DELAY 3
 
+/**
+ * Log events which caused a failure in an ONNXIFI function call.
```
bddppq: Does this line belong to `ONNXIFI_LOG_LEVEL_ERROR `? For `ONNXIFI_LOG_LEVEL_DEBUG` and `ONNXIFI_LOG_LEVEL_INFO ` they don't necessary correspond to some failures.

onnx/onnxifi.h
```diff
@@ -493,6 +506,30 @@ typedef int32_t onnxBackendInfo;
...
+ * about operation of a backend. Enabling this logging level MAY cause a small
+ * degradation in performance.
+ */
+#define ONNXIFI_LOG_LEVEL_INFO 2
+/**
+ * Log events in ONNXIFI_LOG_LEVEL_INFO and detailed status information about
+ * operations of a backend. Enabling this logging level MAY cause a large
+ * degradation in performance.
+ */
+#define ONNXIFI_LOG_LEVEL_DEBUG 1
```
bddppq: Add doc for this as well?

onnx/onnxifi.h
Maratyszcza: Good catch!

Maratyszcza: Yes, it is for `ONNXIFI_LOG_LEVEL_ERROR`. `ONNXIFI_LOG_LEVEL_INFO` and `ONNXIFI_LOG_LEVEL_DEBUG` also enable logging of errors, but can log many additional events too.

PR #1229, opened 2018-07-25
Clarify uniqueness of backend ID for the life-time of the process

Maratyszcza:
As agreed per discussion with WinML


bddppq:
`default` CI failure unrelated

PR #1233, opened 2018-07-26
Combined several similar chunk of code in Cmake into a function

zrphercule:
created msvc_mode_flag function for cmake to make the code shorter and reusable.


CMakeLists.txt
```diff
@@ -108,6 +108,22 @@ if(ONNX_ML)
   add_definitions("-DONNX_ML=1")
 endif()
 
+function(msvc_mode_flag lib)
```
yinghai: `add_msvc_runtime_flag` might be a better name?

zrphercule:
@yinghai Yes you are right, I am changing it...

PR #1236, opened 2018-07-27
fix type constraints for tensor ops.

linkerzhang:
For many tensor ops, they should support all tensor types. this PR is fixing the wrong type constraints specified before.


onnx/defs/data_type_utils.cc
```diff
@@ -391,8 +391,8 @@ TypesWrapper::TypesWrapper() {
   type_str_to_tensor_data_type_["uint16"] = TensorProto_DataType_UINT16;
   type_str_to_tensor_data_type_["uint32"] = TensorProto_DataType_UINT32;
   type_str_to_tensor_data_type_["uint64"] = TensorProto_DataType_UINT64;
-  type_str_to_tensor_data_type_["complext64"] = TensorProto_DataType_COMPLEX64;
-  type_str_to_tensor_data_type_["complext128"] =
+  type_str_to_tensor_data_type_["complex64"] = TensorProto_DataType_COMPLEX64;
```
bddppq: lol I'm wondering whether there is typo checker we can add to the CI :-)

linkerzhang: Fair point, will add some test cases.

PR #1238, opened 2018-07-28
shape inference: add initializers and inference for reshape operator

varunjain99:
1) Give access to initializers in InferenceContext
2) Do shape inference for reshape operator using the initializer


onnx/defs/tensor/defs.cc
```diff
@@ -89,7 +89,98 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          // we set the unresolvedZero flag. If targetShape dimension is positive,
+          // we set the dimension value in the outputShape. We track the product of 
+          // the dimensions we are setting outputShape in the outputProduct variable.
+          // The outputProduct will potentially be used for inferring a dimension marked -1.
+          auto* outputShape = ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+          TensorShapeProto::Dimension* negativeOneDim = nullptr;
+          const auto& dataInputTensorType = ctx.getInputType(0)->tensor_type();
+          bool unresolvedZero = false;
+          int64_t outputProduct = 1;
+          for (int i = 0; i < static_cast<int>(targetShape.size()); ++i)  {
```
bddppq: why use signed instead of unsigned for i here?

onnx/defs/tensor/defs.cc
```diff
@@ -89,7 +89,98 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              } else {
+                unresolvedZero = true;
+              }
+            } else if (targetShape[i] > 0) {
+              // Set the dimension value to targetShape[i]
+              new_dim->set_dim_value(targetShape[i]);
+              outputProduct *= targetShape[i];
+            } else {
+              // Check if value is less than -1; fail if so
+              fail_shape_inference("Target shape may not have dimensions less than -1");
```
bddppq: nit: just say the dim value is invalid? because it's not about less or greater than -1

onnx/defs/tensor/defs.cc
```diff
@@ -89,7 +89,98 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              // Check if data input has a shape, if the index i is within its bounds, and
+              // the dimension value is present at index i. If these conditions are satisfied,
+              // the dimension value can be inferred and set. Otherwise, set unresolvedZero flag.
+              if (dataInputTensorType.has_shape() && 
+                  i < dataInputTensorType.shape().dim_size() && 
+                  dataInputTensorType.shape().dim(i).has_dim_value()) {
+                const auto& dim_value = dataInputTensorType.shape().dim(i).dim_value();
+                new_dim->set_dim_value(dim_value);
+                outputProduct *= dim_value;
+              } else {
```
bddppq: in case the shape is a param, we should still propagate it to the output

onnx/defs/tensor/defs.cc
```diff
@@ -89,7 +89,98 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          // Type Inference
           propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          // Shape Inference if target shape is in initializer
+          const TensorProto* targetShapeInitializer = ctx.getInputInitializer(1);
+          if (!targetShapeInitializer)  {
+            return;
+          }
+          // Make targetShape (0 -> same as originalShape, -1 -> inferred).
+          // The targetShape vector represents the specified shape for output.
+          std::vector<int64_t> targetShape;
```
bddppq: you can use `data<int64_t>()` to get the tensor's data pointer and size_from_dim(0) to get the #elements

onnx/defs/tensor/defs.cc
```diff
@@ -89,7 +89,98 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          // information is present for the data input tensor.
+          if (negativeOneDim && !unresolvedZero) {
+            // First, attempt to compute number of elements in input data tensor, setting flag
+            // to false if not possible
+            int64_t dataElements = 1;
+            bool dataElementsFlag = true;
+            if (!dataInputTensorType.has_shape()) {
+              dataElementsFlag = false;
+            } else {
+              for (int i = 0; i < dataInputTensorType.shape().dim_size(); ++i)  {
```
bddppq: nit: probably can do 
```
for (const auto& dim : dataInputTensorType.shape.dim()) {
...
}
```

onnx/defs/tensor/defs.cc
```diff
@@ -89,7 +89,98 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            } else {
+              // Check if value is less than -1; fail if so
+              fail_shape_inference("Target shape may not have dimensions less than -1");
+            }
+          }
+
+          // If negativeOneDim has been set, we attempt to infer its value. This can be done if 
+          // all other dimensions have been set (unresolvedZero is false) and complete shape
+          // information is present for the data input tensor.
+          if (negativeOneDim && !unresolvedZero) {
```
bddppq: I think we can still infer the negativeOneDim when there is unresolvedZero, e.g.
```
input shape: 1 2 a 3
target shape: 1 1 0 -1
=>
output shape: 1 1 a 6
```

onnx/defs/tensor/defs.cc
varunjain99: Isn't TensorProto different from Tensor?

onnx/defs/tensor/defs.cc
varunjain99: Protobuf dim_size function returns an int, and I compare i < dataInputTensorType.shape().dim_size(). Also I think the dim function is supposed to take in an int

onnx/defs/tensor/defs.cc
varunjain99: Good point, but now I actually use the i

onnx/defs/tensor/defs.cc
varunjain99: True. To infer in this case, we need to keep track of which indices are unresolvedZero's. Only when the data input tensor has missing values corresponding to unresolvedZero's can we infer

onnx/defs/shape_inference.h
```diff
@@ -42,6 +42,7 @@ struct InferenceContext {
   virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
   virtual size_t getNumInputs() const = 0;
   virtual const TypeProto* getInputType(size_t index) const = 0;
+  virtual const TensorProto* getInputInitializer(size_t index) const = 0;
```
linkerzhang: Thanks for adding this! 

Suggestion: name it as getInput, since it does not have to be an initializer. What we want here is to access input tensor data.

onnx/defs/tensor/defs.cc
```diff
@@ -77,19 +77,126 @@ could also be 0, in which case the actual dimension value is unchanged (i.e. tak
...
-            {"tensor(float16)", "tensor(float)", "tensor(double)"},
-            "Constrain input and output types to float tensors.")
-        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+        .TypeConstraint("T",
+                        {"tensor(float16)", "tensor(float)", "tensor(double)"},
+                        "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext &ctx) {
+          // Type Inference
           propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          // Shape Inference if target shape is in initializer
```
linkerzhang: Shape inference if 2nd input is available.

onnx/defs/tensor/defs.cc
```diff
@@ -77,19 +77,126 @@ could also be 0, in which case the actual dimension value is unchanged (i.e. tak
...
+                               reinterpret_cast<const int64_t *>(bytes.c_str()),
+                               reinterpret_cast<const int64_t *>(bytes.c_str() +
+                                                                 bytes.size()));
+          } else {
+            const auto &data = targetShapeInitializer->int64_data();
+            targetShape.insert(targetShape.end(), data.begin(), data.end());
+          }
+
+          // Iterate through targetShape, adding dimensions in the outputShape
+          // TensorProto. If the targertShape dimension is -1, we do not set the
```
linkerzhang: Should we support the case that 2nd input is available and its value is -1? if yes, what's the scenario please?

onnx/shape_inference/implementation.h
```diff
@@ -58,6 +74,7 @@ struct InferenceContextImpl : public InferenceContext {
     }
     return &allOutputTypes_[index];
   }
+  std::vector<const TensorProto*> allInputInitializers_;
```
linkerzhang: name it as "all_inputs" or "all_input_tensors" may be better since you're putting nullptr for an input if it's not an initializer.

onnx/defs/tensor/defs.cc
bddppq: @varunjain99 oh yeah you are right

onnx/defs/tensor/defs.cc
bddppq: @linkerzhang Could you clarify by "its value is -1" you mean the second input is a scalar value `-1` or the second input is a 1D tensor which has an item being `-1`? The latter is supported in @varunjain99's current implementation, what is he saying in this comment is the whole inference process has two stages, and this first stage will mark the `-1` axis, and then compute it in the second stage (see negativeOneDim).

onnx/defs/shape_inference.h
varunjain99: Good point - making it getInputData

onnx/shape_inference/implementation.h
varunjain99: I think allInputData_ should keep it consistent

onnx/defs/tensor/defs.cc
```diff
@@ -77,19 +77,126 @@ could also be 0, in which case the actual dimension value is unchanged (i.e. tak
...
+          // can be done if all dimension values for the data input tensor shape
+          // are known other than the ones corresponding to unresolvedZeros
+          // flags.
+          if (negativeOneDim) {
+            // First, attempt to compute product of data input shape dimensions
+            // that are not marked by unresolvedZeros. If not possible, set the
+            // inputProductValid flag to false.
+            int64_t inputProduct = 1;
+            bool inputProductValid = true;
+            if (!dataInputTensorType.has_shape()) {
```
linkerzhang: add "i >= static_cast<int>(unresolvedZeros.size()" here.

onnx/defs/tensor/defs.cc
```diff
@@ -77,19 +77,126 @@ could also be 0, in which case the actual dimension value is unchanged (i.e. tak
...
+              }
+              negativeOneDim = new_dim;
+            } else if (targetShape[i] == 0) {
+              // Check if data input has a shape and if the index i is within
+              // its bounds. If these conditions are satisfied, any dimension
+              // value/param should be propogated. If dimension value cannot be
+              // inferred, set the corresponding  unresolvedZeros flag to true.
+              unresolvedZeros[i] = true;
+              if (dataInputTensorType.has_shape() &&
+                  i < dataInputTensorType.shape().dim_size()) {
```
linkerzhang: if (i >= dataInputTensorType.shape().dim_size()), there should be a data error, right?

onnx/defs/tensor/defs.cc
```diff
@@ -77,19 +77,126 @@ could also be 0, in which case the actual dimension value is unchanged (i.e. tak
...
+                  inputProduct *=
+                      dataInputTensorType.shape().dim(i).dim_value();
+                } else if (i >= static_cast<int>(unresolvedZeros.size()) || !unresolvedZeros[i]) {
+                  inputProductValid = false;
+                  break;
+                }
+              }
+            }
+            if (inputProductValid) {
+              if (inputProduct % outputProduct != 0) {
```
linkerzhang: add check to ensure outputProduct is not "0".

onnx/defs/tensor/defs.cc
linkerzhang: I mean the latter one.

I was thinking whether we should support the latter case in this shape inference function. It depends on when we want to run this shape inference function,  if it's before the data coming only, then no need to implement this case here I think. If we also run this function after we get the input data, it's fine to have it here.

bddppq: @linkerzhang Why are there differences between with and without real input?

varunjain99: The purpose of this function is that if the reshape shape (data of input 2) is present, we try to infer as much of the output shape. When there is a -1 in the 2nd input data, we can try to infer this dimension if we have enough *shape* information from input 1. We don't need the data of input 1, but enough shape information from input 1, to figure out what -1 should actually be.

onnx/defs/tensor/defs.cc
varunjain99: @linkerzhang what do you mean by this? We're not in a loop here?

onnx/defs/tensor/defs.cc
varunjain99: Yes, good point - I'll add a fail_inference for that

varunjain99:
@onnxbot retest this please

bddppq:
@onnxbot retest this please

onnx/test/shape_inference_test.py
```diff
@@ -44,7 +54,7 @@ def _assert_inferred(self, graph, vis):  # type: (GraphProto, List[ValueInfoProt
         vis = list(x for x in graph.value_info if x.name not in names_in_vis) + vis
         inferred_model = self._inferred(graph)
         inferred_vis = list(inferred_model.graph.value_info)
-
+        print(inferred_vis)
```
bddppq: remove this?

varunjain99: Oops - removed

onnx/defs/tensor/defs.cc
```diff
@@ -89,7 +88,118 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              // Check if value is less than -1; fail if so
+              fail_shape_inference("Dimension value is invalid");
+            }
+          }
+
+          // If negativeOneDim has been set, we attempt to infer its value. This
+          // can be done if all dimension values for the data input tensor shape
+          // are known other than the ones corresponding to unresolvedZeros
+          // flags.
+          if (negativeOneDim && outputProduct != 0) {
```
bddppq: should raise error if(negativeOneDim and outputProduct == 0)

onnx/defs/tensor/defs.cc
```diff
@@ -89,7 +88,118 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                  new_dim->set_dim_param(dim_param);
+                }
+              }
+            } else if (targetShape[i] > 0) {
+              // Set the dimension value to targetShape[i]
+              new_dim->set_dim_value(targetShape[i]);
+              outputProduct *= targetShape[i];
+            } else {
+              // Check if value is less than -1; fail if so
+              fail_shape_inference("Dimension value is invalid");
```
bddppq: nit: maybe also add the invalid value to the error message?

onnx/defs/tensor/defs.cc
varunjain99: fixed

onnx/defs/tensor/defs.cc
varunjain99: added

PR #1239, opened 2018-07-28
backports stoi on android

bddppq:
https://ci.pytorch.org/jenkins/job/caffe2-builds/job/py2-android-ubuntu16.04-build/12188/console


snnn:
two cents: We'd better to use strtol instead of stringstream in this case, because:
1. strtol is faster
2. strtol has better error check.  We should never ignore conversion errors, that's too dangerous.

PR #1242, opened 2018-07-31
move macro ONNXIFI_DUMMY_LIBRARY to onnxifi_dummy.h

zrphercule:
This macro is used in cpp test driver sample.


Maratyszcza:
Why the driver doesn't just dlopen with the default path?

PR #1248, opened 2018-08-01
Shape inference support for conv with groups

varunjain99:
@bddppq @houseroad


gramalingam:
I believe that the original check for groups was added because the behavior is different in the case of groups not being 1 (with a different shape inference required) … is this not the case?

varunjain99:
@onnxbot retest this please

bddppq:
@gramalingam IIUC the group number does not change the output shape

bddppq:
@houseroad will do a follow up PR to update the spec that the size of the kernel's second dimension should be `C / g`

Issue #1222, opened 2018-07-23
Spec clarity: contradiction on domain names


NiklasGustafsson:
Here's what IR.md has to say about model domain names:

_Models MUST specify a domain and use reverse domain names based on the responsible organization's identity,[...]_

And here is what onnx.proto has to say, in a comment:

  // Domain name of the model.
  // We use reverse domain names as name space indicators. For example:
  // `com.facebook.fair` or `com.microsoft.cognitiveservices`
  //
  // Together with `model_version` and GraphProto.name, this forms the unique identity of
  // the graph.

The commentary is silent on whether the property must be present or not, while the .proto file will specifically call out properties that must be present in other cases. While the IR.md establishes itself as normative ahead of comments in the .proto file when it comes to semantics, the inconsistency needs to be acknowledged and resolve.

PR #1257, opened 2018-08-03
Stricter model checking

NiklasGustafsson:
In this PR, the ONNX model checker tool is significantly more strict. It checks for a number of model errors that were previously skipped. It now performs type and shape inference and validates against existing value_info records, and flags any discrepancies.

As type and shape inference was added, a number of tests started failing. Many of those were simple test mistakes, which I fixed. However, the backend tests rely on models converted from PyTorch and other sources, including the ONNX Model Zoo. Many of these models contain names that are not valid -- the specification explicitly requires that all names follow C identifier syntax, so characters like '/' and '-' cannot appear in names. Also, names may not be simple numbers. Those models need to be fixed.

I added a simple tool 'fixmodel' to correct the most common errors -- missing domain names, and name syntax. For the latter, is simply replaces forbidden characters with '_'. Since this could conceivably result in name collisions (although the likelihood is very small) it would be better to address the issue at the source.

The specification is inconsistent on whether model domain names are required or not. See issue #1222.
 Since there is this discrepancy, I disabled the domain name check for the time being, until that issue is resolved.

Before this PR, the model checker only flagged one error and then gave up. I modified it to continue looking for errors after one is found, but that required making breaking changes to the API, since the checker context went from const to non-const.

In addition to the model checker work, I added more information on how to build ONNX on Windows, as I struggled a bit with that myself.

The model checker has been tested on Windows and OS X.


NiklasGustafsson:
I see that the check is failing building Caffe2, because of the breaking API changes in checker.h/.cc. I ran 'pip install -e .' to build this, but I overlooked running pytest. I assume that's when caffe2 is being built.

NiklasGustafsson:
Update: I moved the tests that cause the checker to error out with name errors to 'disabled_tests' @houseroad, if you have advice on how I should proceed with the caffe2 build errors, that would be appreciated -- that's doesn't appear in the regular install script, as far as I can see, so it must be something that's just done during CI.

onnx/validate.py
```diff
@@ -0,0 +1,51 @@
...
+import onnx
+from onnx import defs
+from onnx.defs import OpSchema
+import onnx.onnx_cpp2py_export.checker as c_checker
+import onnx.shape_inference
+
+import argparse
+
+
+def main():  # type: () -> None
```
houseroad: Please merge this file with [checker.py](https://github.com/onnx/onnx/blob/master/onnx/checker.py). It will cause a lot of confusion if we have one check and another validate.

onnx/defs/schema.cc
```diff
@@ -201,60 +309,39 @@ void OpSchema::Verify(const NodeProto& node) const {
...
-              "Attribute '", name, "' is expected to have field 'tensors'");
-        }
-        break;
       case AttributeProto::GRAPHS:
-        if (!attr_proto.graphs_size()) {
-          fail_check(
-              "Attribute '", name, "' is expected to have field 'graphs'");
-        }
+        // There's no check to be made here -- it is perfectly valid for a
+		// sequence-typed attribute to have zero elements.
```
houseroad: indent?

onnx/validate.py
NiklasGustafsson: Sure. checker.py seemed structured to be more of an API, so I wasn't sure if it was OK to add a 'main' to it, but I agree that it's better to have just one entry.

onnx/defs/schema.cc
NiklasGustafsson: Thanks for catching it!

onnx/validate.py
houseroad: You can add the main here: 
https://github.com/onnx/onnx/blob/master/onnx/bin/checker.py

onnx/checker.cc
```diff
@@ -35,15 +43,138 @@ namespace checker {
...
+}
+
+bool check_domain_syntax(const std::string& name) {
+  if (name.empty()) {
+    return false;
+  }
+
+  bool correct = true;
+
+  // Names should adhere to C identifier syntax.
```
houseroad: '.' (dot) is not in C identifier syntax

onnx/checker.cc
```diff
@@ -35,15 +43,138 @@ namespace checker {
...
+    }
+  }
+
+  return correct;
+}
+
+#define enforce_c_identifier(proto, field)       \
+  do {                                           \
+    const auto& str = proto.field();             \
+    if (!check_name_syntax(str)) {               \
```
houseroad: It's weird to call `check_name_syntax` in `enforce_c_identifier`.
We have to two options to solve the problem:
1) change `check_name_syntax` to `check_c_identifier`
2) change `enforce_c_identifier` to `enforace_name_syntax`

onnx/checker.cc
```diff
@@ -35,15 +43,138 @@ namespace checker {
...
+      correct = false;
+    }
+  }
+
+  return correct;
+}
+
+bool check_domain_syntax(const std::string& name) {
+  if (name.empty()) {
+    return false;
```
houseroad: if domain is empty, it's default domain. We should accept it.

onnx/checker.cc
```diff
@@ -35,15 +43,138 @@ namespace checker {
...
   const auto value_case = value_info.type().value_case();
   switch (value_case) {
     case TypeProto::kTensorType: {
       const auto& type = value_info.type().tensor_type();
       enforce_has_field(type, elem_type);
-      enforce_has_field(type, shape);
+
+      if (type.has_shape()) {
+
+        for (auto& dim : type.shape().dim()) {
```
houseroad: We need to check `has_dim` before accessing it.

onnx/checker.cc
```diff
@@ -35,15 +43,138 @@ namespace checker {
...
+          const auto dim_case = dim.value_case();
+          switch (dim_case) {
+            case 0:
```
houseroad: why 0? it should be "" in dim_param

onnx/checker.cc
```diff
@@ -246,61 +412,142 @@ void check_attribute(
...
 void check_node(
     const NodeProto& node,
-    const CheckerContext& ctx,
+    const GraphProto& graph,
+    CheckerContext& ctx,
+    const LexicalScopeContext& lex_ctx) {
+  check_node(node, ctx, lex_ctx);
+
+  // Resolve domain for node
+  auto domain = node.domain();
```
houseroad: If domain is missing, we use default one.

onnx/checker.cc
NiklasGustafsson: Thanks for catching the inconcistency. I will update the comment in the function to reflect that this is about domain name syntax, which does allow '.'

onnx/checker.cc
NiklasGustafsson: Excellent point. Going with your second suggestion.

onnx/checker.cc
NiklasGustafsson: True for nodes, and possibly true for models, depending on resolution of #1222. The check was intended for models, so I'm adding a separate check for nodes that accept an empty domain.

onnx/checker.cc
NiklasGustafsson: Yup, thanks for catching it.

onnx/checker.cc
NiklasGustafsson: That's right. I've added an if() to guard this check.

onnx/checker.cc
NiklasGustafsson: Hmm, dim is a repeated field, so protoc doesn't generate a 'has_dim()' for it in C++.

NiklasGustafsson:
@houseroad, thanks for your comments. I believe that I have addressed them all in my latest update of the PR.

onnx/defs/schema.h
```diff
@@ -22,6 +22,10 @@
 #include "onnx/defs/shape_inference.h"
 namespace ONNX_NAMESPACE {
 
+namespace checker {
```
houseroad: I think we correctly structure the code, `checker` is not necessary here.

onnx/defs/schema.h
```diff
@@ -194,7 +198,10 @@ class OpSchema final {
    * @brief Verifies if a NodeProto matches the pattern specified in
    * the schema.
    */
-  void Verify(const NodeProto& node) const;
+  void Verify(
```
houseroad: Can we leave this context independent? And move the verifyNode in the checker itself?

NiklasGustafsson: The context is used to report errors. I changed the logic to report more than one error at a time, but that requires using something (the context in this case) to accumulate error messages.

NiklasGustafsson: I've moved the operator schema verification to checker.cc.

houseroad:
@NiklasGustafsson sorry for the late reply. This is awesome work which improves the checker. However, it has added more checks, and if we squeeze it in, will break our whole pipeline. Is it possible to split it into smaller pieces? Each one contains 100-ish lines of code?

Thanks

zrphercule:
@NiklasGustafsson I generally agree with @houseroad 
I think this would be very useful, however if we merge it all together and bring all problem together, it will be hard to solve; But if you can break it into a few pieces, that would make the work much easier, and we can fix the problem one by one.
At least we can merge a few not-that-breaking changes, right?

PR #1260, opened 2018-08-03
Add a test case for abs op and int8/int16/... type

snnn:
We definitely need to cover the data types other than float32 in onnx backend test, otherwise who knows which works which not.
Please let me know what's the best way of doing this.

I've tried to add one.  Should duplicate this code many more times to cover the other types?

Thanks.


snnn:
"error: Function is missing a type annotation"

I'm fixing it.

onnx/backend/test/case/node/abs.py
```diff
@@ -10,17 +10,77 @@
 from . import expect
 
 
+def generate_test(x, name):
```
zrphercule: annotations here, describe the data type of `x` and `name`, and also return value of this function.

onnx/backend/test/case/node/abs.py
```diff
@@ -10,17 +10,77 @@
...
+        'Abs',
+        inputs=['x'],
+        outputs=['y'],
+    )
+    y = np.abs(x)
+    expect(node, inputs=[x], outputs=[y],
+           name=name)
+
+
+def generate_test_unsigned(x, name):
```
zrphercule: and here

onnx/backend/test/case/node/abs.py
```diff
@@ -10,17 +10,77 @@
...
+
+    @staticmethod
+    def export_uint32():  # type: () -> None
+        x = np.uint32([0, 1, 20, 4294967295])
+        generate_test_unsigned(x, 'test_abs_uint32')
+
+    @staticmethod
+    def export_int64():  # type: () -> None
+        number_info = np.iinfo(np.int64)
+        x = np.int64([-number_info.max, -4, 0, 3, number_info.max])
```
zrphercule: should better use number_info.min here to test extreme case?
I sugget you to use np.iinfo.max/min for all test cases.

snnn: No. Cannot use min. For example, if it's a 8 bits signed value, the min value is -256, but the abs(-256) is undefined.

zrphercule: @snnn Oh I see, yes I forgot this number if for testing abs.

snnn: Thanks. I'll add the annotations.

PR #1262, opened 2018-08-05
Update docs to reference optimizer.get_available_passes()

pliniosilveira:


docs/Optimizer.md
```diff
@@ -22,18 +22,21 @@ The optimizer may be invoked either via C++ or Python. The Python API
...
     const std::vector<std::string>& names);
+
+const std::vector<std::string> GetAvailablePasses();
 ```
 
-which accepts an input `ModelProto` and a list of optimization passes to
-apply, and which returns a new `ModelProto` which is the result of
-applying each of the named passes in sequence to the model. For a list
-of available passes, see [optimize.h](/onnx/optimizer/optimize.h).
+`GetAvailablePasses()` return a list of available passes.
```
houseroad: Move the introduction of GetAvailablePasses after Optimize?

PR #1274, opened 2018-08-09
[Doc Fix] Add the default value for auto_pad

houseroad:
Actually, we provide `NOTSET`as default value, let's add it to the real doc.


onnx/defs/nn/defs.cc
```diff
@@ -17,9 +17,11 @@ const char* auto_pad_doc =
     "auto_pad must be either SAME_UPPER, SAME_LOWER or VALID. Where "
```
NiklasGustafsson: Shouldn't 'NOTSET' be listed among the valid values? It sounds like it is one of the allowed values.

PR #1276, opened 2018-08-10
Honor visibility options when compiled as a static library

yinghai:
When compiled as static library and included as a lib in other project, current cmake OLD behavior won't honor the visibility settings from the outer projects, resulting in warning in linking. This fixes it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ if(NOT CMAKE_BUILD_TYPE)
       "Choose the type of build from: Debug Release RelWithDebInfo MinSizeRel Coverage."
     FORCE)
 endif()
-cmake_policy(SET CMP0063 NEW)
+#cmake_policy(SET CMP0063 NEW)
 
 # Project
 project(onnx C CXX)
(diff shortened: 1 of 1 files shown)


yinghai:
It should be ok.

houseroad:
@onnxbot retest this please

houseroad:
@onnxbot retest this please

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ if(NOT CMAKE_BUILD_TYPE)
       "Choose the type of build from: Debug Release RelWithDebInfo MinSizeRel Coverage."
     FORCE)
 endif()
+#cmake_policy(SET CMP0063 NEW)
```
rdzhabarov: temporary commented out?

houseroad: We are checking the problem with caffe2 build

yinghai: We saw test failures in CI. I comment this out to see if it's my PR that introduced the failure or it's master. Looks like it's in master. :(

yinghai:
@onnxbot retest this please

PR #1283, opened 2018-08-13
fix dtype mismatch in expand_op test

zrphercule:


onnx/backend/test/case/model/expand.py
```diff
@@ -34,32 +34,32 @@ def make_graph(node, input_shape, shape_shape, output_shape):  # type: (onnx.hel
         node = onnx.helper.make_node(
             'Expand', ['X', 'shape'], ['Y'], name='test')
         input_shape = [1, 3, 1]
-        x = np.ones(input_shape)
+        x = np.ones(input_shape, dtype=np.float32)
 
         #1st testcase
         shape = np.array([3, 1])
```
houseroad: Here, you also need to turn them into np.int64. Otherwise, on 32-bit machine, we will generate np.int32 instead.

zrphercule: yeah you are right

houseroad: Also do you need to update the generated data files too?

houseroad:
I guess you also need to run `python cmd_tools.py generate-data` to update the test data, and commit them.

zrphercule:
@houseroad Yes sure, I was planning to commit the data once the test code has no problem =)

houseroad:
Awesome, that party is easy to forget. Just wanna confirm :-)

onnx/backend/test/case/node/expand.py
```diff
@@ -20,11 +20,11 @@ def export_dim_changed():  # type: () -> None
             outputs=['expanded'],
         )
         shape = [3, 1]
```
houseroad: Similar apply here? The explicit type declaration.

zrphercule:
@houseroad Could you plz give me a stamp? ty!

PR #1184, opened 2018-07-11
Model Version Converter Adapters

Ac2zoom:
Implementation of Version Converter adapters, particularly for Model Zoo, as outlined in internal spec

Will be rebased onto version_converter until that is merged

PR #1284, opened 2018-08-14
Broadcast Version Conversion Adapters (Add, Mul, Gemm)

Ac2zoom:
Version Conversion Adapters for Add, Mul, and Gemm


onnx/version_converter.py
```diff
@@ -26,7 +26,134 @@
     RuntimeError when some necessary conversion is not supported
 
 Supported adapters:
-    --
+    --Add from Opset 7 to Opset 6
```
houseroad: Adjust this list accordingly?

onnx/test/test_backend_test.py
```diff
@@ -41,7 +42,10 @@ def prepare(cls,
         model = onnx.shape_inference.infer_shapes(model)
         value_infos = {vi.name: vi for vi in itertools.chain(model.graph.value_info, model.graph.output)}
 
-        if do_enforce_shape_inference_coverage(model):
+        if do_enforce_test_coverage_whitelist(model):
+            # test version conversion
```
houseroad: Can we remove this?

houseroad: I mean the comments

onnx/version_converter/adapters/adapter.h
```diff
@@ -13,12 +13,34 @@ class Adapter {
     OpSetID initial_version_;
     OpSetID target_version_;
 
+  protected:
+    bool c2_broadcastable(int64_t axis,
```
houseroad: Rename c2_broadcastable to onnx_opset1_broadcastable.

And make them as functions not member functions.

onnx/version_converter/adapters/adapter.h
```diff
@@ -13,12 +13,34 @@ class Adapter {
...
+        const std::vector<Dimension>& input1_sizes,
+        const std::vector<Dimension>& input2_sizes) const {
+      return axis != (int) (input1_sizes.size() - input2_sizes.size());
```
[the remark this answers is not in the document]
Ac2zoom: Restoring from https://github.com/onnx/onnx/pull/1184/commits/da507b8a42f68c6bcfc088cd33184d5d05fac154

onnx/version_converter/adapters/backwards_compatible.h
```diff
@@ -0,0 +1,17 @@
+// Adapter indicating backwards compatibility of op between opsets with
+// separate definitions
+
+#pragma once
+
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+struct BackwardsCompatibleAdapter final : public Adapter {
```
houseroad: Change the name to ComptibleAdapter?

onnx/version_converter/adapters/broadcast_backward_compatibility.h
```diff
@@ -0,0 +1,39 @@
...
+      // Main change: encode broadcasting commands as explicit attribute
+      const ArrayRef<Value*>& inputs = node->inputs();
+      ONNX_ASSERTM(inputs.size() == 2, "%s in opset version 6 can only broadcast"
+        " between 2 inputs", name().c_str());
+      ONNX_ASSERTM(inputs[0]->has_sizes(), "Shape of first input is not available.");
+      std::vector<Dimension> A_sizes = inputs[0]->sizes();
+      ONNX_ASSERTM(inputs[1]->has_sizes(), "Shape of second input is not available.");
+      std::vector<Dimension> B_sizes = inputs[1]->sizes();
+      // Ensure that first input is larger than or equal to the second
+      if(numpy_unibroadcastable(A_sizes, B_sizes)) {
```
houseroad: wait, this should be opset1_broadcastable, right?

onnx/version_converter/helper.cc
```diff
@@ -0,0 +1,46 @@
...
+      }
+      // Return true if broadcasting is required
+      return input1_sizes.size() > input2_sizes.size() || broadcast;
+    }
+
+    void numpy_multibroadcastable(const std::vector<Dimension>& input1_sizes,
+        const std::vector<Dimension>& input2_sizes) {
+      // Generalize above for multibroadcastable case
+      if (input1_sizes.size() < input2_sizes.size()) {
+        numpy_unibroadcastable(input2_sizes, input1_sizes);
```
houseroad: This is incorrect. check [2, 5, 1, 3] vs [2, 1, 5, 3]

onnx/version_converter/adapters/broadcast_forward_compatibility.h
```diff
@@ -0,0 +1,66 @@
...
+      // Assess whether axis requires reshaping
+      if (node->hasAttribute(kbroadcast)) {
+        const ArrayRef<Value*>& inputs = node->inputs();
+        ONNX_ASSERTM(inputs.size() == 2, "%s in opset version 6 can only broadcast"
+          " between 2 inputs", name().c_str());
+        ONNX_ASSERTM(inputs[0]->has_sizes(), "Shape of first input not available.");
+        std::vector<Dimension> A_sizes = inputs[0]->sizes();
+        ONNX_ASSERTM(inputs[1]->has_sizes(), "Shape of second input not available.");
+        std::vector<Dimension> B_sizes = inputs[1]->sizes();
+        // Also assert that broadcasting syntax are correct if axis is not present
```
houseroad: Probably add another assertion here: check all the dimensions are dims, not params.

https://github.com/onnx/onnx/blob/36517e6bac9c0f0cbf9390c97c39b8106dd4d33e/onnx/common/ir.h#L75

onnx/version_converter/adapters/broadcast_forward_compatibility.h
```diff
@@ -0,0 +1,66 @@
...
+        if (node->hasAttribute(kaxis)) {
+          if (!onnx_opset7_broadcastable(node->i(kaxis), A_sizes, B_sizes)) {
+            // Add a Reshape node before input B
+            Node * n = graph->create(kReshape);
```
houseroad: Instead of using `Reshape`, let's do unsqueeze, which is more flexible.

onnx/version_converter/adapters/compatible.h
```diff
@@ -0,0 +1,17 @@
+// Adapter indicating backwards compatibility of op between opsets with
```
houseroad: Remove backward here too

onnx/version_converter/adapters/broadcast_forward_compatibility.h
```diff
@@ -0,0 +1,66 @@
...
+        const ArrayRef<Value*>& inputs = node->inputs();
+        ONNX_ASSERTM(inputs.size() == 2, "%s in opset version 6 can only broadcast"
+          " between 2 inputs", name().c_str());
+        ONNX_ASSERTM(inputs[0]->has_sizes(), "Shape of first input not available.");
+        std::vector<Dimension> A_sizes = inputs[0]->sizes();
+        ONNX_ASSERTM(inputs[1]->has_sizes(), "Shape of second input not available.");
+        std::vector<Dimension> B_sizes = inputs[1]->sizes();
+        // Also assert that broadcasting syntax are correct if axis is not present
+        if (node->hasAttribute(kaxis)) {
+          if (!onnx_opset7_broadcastable(node->i(kaxis), A_sizes, B_sizes)) {
```
houseroad: This if-else seems incorrect.

onnx/version_converter/helper.cc
Ac2zoom: Resolved.  Didn't think of the double-sided broadcasting case.  Sorry about that!

onnx/version_converter/adapters/gemm_6_7.h
```diff
@@ -0,0 +1,46 @@
...
+      ONNX_ASSERTM(inputs[0]->has_sizes(), "Insufficient shape information for 1st input");
+      const auto& A_shape = inputs[0]->sizes();
+      ONNX_ASSERTM(inputs[1]->has_sizes(), "Insufficient shape information for 2nd input");
+      const auto& B_shape = inputs[1]->sizes();
+      // Determine if C is broadcastable
+      ONNX_ASSERTM(inputs[2]->has_sizes(), "Insufficient shape information for 3rd input");
+      const auto& C_shape = inputs[2]->sizes();
+      // Create (M, N) to input to numpy_unibroadcastable
+      // TODO: Reconcile fact that shapes aren't determined for 1st 2 inputs
+      std::vector<Dimension> MN;
```
houseroad: Since Gemm 6==>7, there is no axis, shouldn't we just remove `broadcast` flag?

onnx/version_converter/adapters/gemm_7_6.h
```diff
@@ -0,0 +1,46 @@
...
+      const ArrayRef<Value*>& inputs = node->inputs();
+      ONNX_ASSERTM(inputs.size() == 3, "3 Inputs must be provided to Gemm");
+      ONNX_ASSERTM(inputs[0]->has_sizes(), "Insufficient shape information for 1st input");
+      const auto& A_shape = inputs[0]->sizes();
+      ONNX_ASSERTM(inputs[1]->has_sizes(), "Insufficient shape information for 2nd input");
+      const auto& B_shape = inputs[1]->sizes();
+      // Determine if C is broadcastable
+      ONNX_ASSERTM(inputs[2]->has_sizes(), "Insufficient shape information for 3rd input");
+      const auto& C_shape = inputs[2]->sizes();
+      // Create (M, N) to input to numpy_unibroadcastable
```
houseroad: First make sure A, B, and C's sizes are ints (dims not param).

Then check whether C needs broadcasting or not.

onnx/version_converter/adapters/type_restriction.h
```diff
@@ -0,0 +1,42 @@
...
+#pragma once
+
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+class TypeRestriction : public Adapter {
+  public:
+    explicit TypeRestriction(const std::string& op_name, const OpSetID&
+      initial, const OpSetID& target): Adapter(op_name, initial, target) {}
```
houseroad: Can we make the unallowed_types and allowed_types as constructor parameters?

onnx/version_converter/adapters/broadcast_forward_compatibility.h
Ac2zoom: How would you recommend naming this function?  It asserts that the op is opset 7 broadcastable, but returns whether or not it requires broadcasting.  Should I call it onnx_opset7_requires_broadcasting?

onnx/version_converter/adapters/broadcast_backward_compatibility.h
```diff
@@ -0,0 +1,39 @@
...
+      initial, const OpSetID& target): Adapter(op_name, initial, target) {}
+
+    void adapt_broadcast_backward_compatibility(std::shared_ptr<Graph> graph, Node* node) const {
+      // Verify that broadcasts are allowed in limited spec of opset version 6
+      // Multidirectional broadcasting, as defined in Broadcasting.md
+      // MathDocGenerator provides differences
+      // Main change: encode broadcasting commands as explicit attribute
+      const ArrayRef<Value*>& inputs = node->inputs();
+      assertInputsAvailable(inputs, name().c_str(), 2);
+      std::vector<Dimension> A_sizes = inputs[0]->sizes();
```
houseroad: const reference?

onnx/version_converter/adapters/broadcast_backward_compatibility.h
```diff
@@ -0,0 +1,39 @@
...
+      std::vector<Dimension> B_sizes = inputs[1]->sizes();
```
houseroad: const reference? you won't change the content right?

onnx/version_converter/helper.cc
```diff
@@ -0,0 +1,79 @@
+// Helper Methods for Adapters
+
+#include "onnx/version_converter/helper.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+    bool onnx_opset7_requires_broadcasting(int64_t axis,
+        const std::vector<Dimension>& input1_sizes,
+        const std::vector<Dimension>& input2_sizes) {
+      // Assert that broadcasting semantics are correct
```
houseroad: Can you check that all the dimensions use int (dim) instead of param? Otherwise you may have 0 vs 0 cases.

onnx/version_converter/helper.cc
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    bool numpy_unibroadcastable(const std::vector<Dimension>& input1_sizes,
+        const std::vector<Dimension>& input2_sizes) {
+      // Assert that equal of input1 larger
+      ONNX_ASSERTM(input1_sizes.size() >= input2_sizes.size(),
+          "ONNX Opset 6 can only broadcast input 2 to input 1");
+      // Assert that axis is input1_sizes.size()-input2_sizes.size()
+      bool broadcast = false;
+      int axis = (int) (input1_sizes.size() - input2_sizes.size());
```
houseroad: I think Dimension has a method called `is_int` or something similar.

onnx/version_converter/helper.cc
```diff
@@ -0,0 +1,79 @@
...
+      for (int i = 0; i < (int) input2_sizes.size(); i++) {
+        ONNX_ASSERTM(input2_sizes[i].dim == input1_sizes[axis + i].dim ||
+            input2_sizes[i].dim == 1, "Dimension %d of input 2 does not match "
```
houseroad: You can return int as result, like 0 is no broadcast needed, 1 is broadcast is needed. -1 is non-broadcastable.

And give better error message in the caller.

onnx/version_converter/adapters/broadcast_forward_compatibility.h
```diff
@@ -0,0 +1,56 @@
...
+            for (int i = 0; i < (int) (A_sizes.size() - B_sizes.size()); i++) {
+              axes.emplace_back(B_sizes.size() + i);
+            }
+            n->is_(kaxes, std::forward<const std::vector<int64_t>>(axes));
+            // Set 2nd input to node to 1st of n and output of n to 2nd input to node
+            node->replaceInput(1, n->output());
+            // Move n before node
+            node->moveBefore(n);
+          } else {
+            onnx_opset7_requires_broadcasting(A_sizes.size() - B_sizes.size(), A_sizes,
```
Ac2zoom: Move outside else case to ensure correctness before numpy_multibroadcastable below

onnx/version_converter/helper.cc
Ac2zoom: The reason we don't distinguish between non-broadcastable and no broadcast needed/broadcast needed is that the asserts will interrupt whichever program if the inputs are non-broadcastable, so we'll never reach any such return value.

Where would the enum be useful here?

onnx/test/version_converter_test.py
```diff
@@ -31,17 +31,20 @@ def _converted(
...
                 [helper.make_tensor_value_info("X", TensorProto.FLOAT, (5,)),
-                    helper.make_tensor_value_info("shape", TensorProto.FLOAT, (1,))],
+                    helper.make_tensor_value_info("W", TensorProto.FLOAT, (1,)),
+                    helper.make_tensor_value_info("Z", TensorProto.FLOAT, (1,))],
                 [helper.make_tensor_value_info("Y", TensorProto.FLOAT, (5,))])
             self._converted(graph, helper.make_operatorsetid("", 8), 2)
         self.assertRaises(RuntimeError, test)
 
-    # Test 2: Backwards Compatible Conversion: Add: 8 -> 7
+    # Test 2: Backwards Compatible Conversion (No Adaptations): Add: 8 -> 7
```
houseroad: It's safer to use 3 -> 2, I guess.

onnx/version_converter/adapters/broadcast_forward_compatibility.h
```diff
@@ -0,0 +1,53 @@
...
+            // Add a Reshape node before input B
+            Node * n = graph->create(kUnsqueeze);
+            n->addInput(inputs[1]);
+            std::vector<int64_t> axes;
+            for (int i = 0; i < (int) (A_sizes.size() - B_sizes.size()); i++) {
+              axes.emplace_back(B_sizes.size() + i);
+            }
+            n->is_(kaxes, std::forward<const std::vector<int64_t>>(axes));
+            // Set 2nd input to node to 1st of n and output of n to 2nd input to node
+            node->replaceInput(1, n->output());
```
houseroad: Don't you need to set the sizes for output of node `n`?

onnx/version_converter/adapters/broadcast_forward_compatibility.h
```diff
@@ -0,0 +1,53 @@
...
+            // Move n before node
+            node->moveBefore(n);
+          }
+        }
+        node->removeAttribute(kbroadcast);
+      }
+      if (node->hasAttribute(kaxis)) node->removeAttribute(kaxis);
+      // Assert multi_broadcastable on inputs
+      const ArrayRef<Value*>& inputs = node->inputs();
+      numpy_multibroadcastable(inputs[0]->sizes(), inputs[1]->sizes());
```
houseroad: Seems to me, input[1]->sizes() should be available...

onnx/version_converter/helper.cc
```diff
@@ -0,0 +1,66 @@
+// Helper Methods for Adapters
+
+#include "onnx/version_converter/helper.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+    bool numpy_unibroadcastable(const std::vector<Dimension>& input1_sizes,
```
houseroad: assert_numpy_unibroadcastable_and_require_broadcast ?

onnx/version_converter/helper.cc
```diff
@@ -0,0 +1,66 @@
...
+        ONNX_ASSERTM(input2_sizes[i].dim == input1_sizes[axis + i].dim ||
+            input2_sizes[i].dim == 1, "Dimension %d of input 2 does not match "
+            "dimension %d of input 1 or the value 1", i, axis + i);
+        if (input2_sizes[i].dim != input1_sizes[axis + i].dim) broadcast = true;
+      }
+      // Return true if broadcasting is required
+      return input1_sizes.size() > input2_sizes.size() || broadcast;
+    }
+
+    void numpy_multibroadcastable(const std::vector<Dimension>& input1_sizes,
```
houseroad: assert_numpy_multibroadcastable_and_require_broadcast ?

Ac2zoom: Doesn't return anything, so not require broadcast, but will fix

Issue #1282, opened 2018-08-13
Convolution and Convolution transpose op with group spec clarification.


pranavsharma:
For convolution the spec says the following about the shape of the weight:

W : T
The weight tensor that will be used in the convolutions; has size (M x C x kH x kW), where C is the number of channels, and kH and kW are the height and width of the kernel, and M is the number of feature maps. For more than 2 dimensions, the kernel shape will be (M x C x k1 x k2 x ... x kn), where (k1 x k2 x ... kn) is the dimension of the kernel. Optionally, if dimension denotation is in effect, the operation expects the weight tensor to arrive with the dimension denotation of [FILTER_IN_CHANNEL, FILTER_OUT_CHANNEL, FILTER_SPATIAL, FILTER_SPATIAL ...].

However if you look at the test case (https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/pytorch-converted/test_Conv2d_groups), W.shape[1]*group_count = num_input_channels. It is not clear from the above spec if C should num_input_channels/group_count. Caffe2's implementation of this operator also assumes the same thing - https://github.com/pytorch/pytorch/blob/master/caffe2/operators/conv_op_impl.h#L29.

Please clarify the spec. This is the exact same problem with ConvTranspose as well.

Thanks!

pranavsharma:
Created a PR for this https://github.com/onnx/onnx/pull/1287

PR #1287, opened 2018-08-14
Clarify the spec for convolution and convolution transpose when group > 1.

pranavsharma:
See https://github.com/onnx/onnx/issues/1282


onnx/defs/nn/defs.cc
```diff
@@ -495,7 +495,9 @@ computes the output.)DOC";
         "Optionally, if dimension denotation is in effect, "
         "the operation expects the weight tensor to arrive "
         "with the dimension denotation of [FILTER_IN_CHANNEL, "
-        "FILTER_OUT_CHANNEL, FILTER_SPATIAL, FILTER_SPATIAL ...].",
+        "FILTER_OUT_CHANNEL, FILTER_SPATIAL, FILTER_SPATIAL ...]. "
```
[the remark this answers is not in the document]
pranavsharma: sure, will fix.

onnx/defs/nn/defs.cc
```diff
@@ -706,7 +709,9 @@ output_shape can also be explicitly specified in which case pads values are auto
         "height and width of the kernel, and M is the number "
         "of feature maps. For more than 2 dimensions, the "
         "weight shape will be (C x M x k1 x k2 x ... x kn), "
-        "where (k1 x k2 x ... x kn) is the dimension of the kernel",
+        "where (k1 x k2 x ... x kn) is the dimension of the kernel. "
```
houseroad: In convtranspose case, weight shape should be C/group x M x ...

onnx/defs/nn/defs.cc
```diff
@@ -495,7 +495,10 @@ computes the output.)DOC";
         "Optionally, if dimension denotation is in effect, "
         "the operation expects the weight tensor to arrive "
         "with the dimension denotation of [FILTER_IN_CHANNEL, "
-        "FILTER_OUT_CHANNEL, FILTER_SPATIAL, FILTER_SPATIAL ...].",
+        "FILTER_OUT_CHANNEL/group, FILTER_SPATIAL, FILTER_SPATIAL ...]. "
+        "Note that X.shape[1] should be equal to W.shape[1] * group "
```
houseroad: It should be [FILTER_OUT_CHANNEL, FILTER_IN_CHANNEL, FILTER_SPATIAL, ...]

And FILTER_IN_CHANNEL = DATA_CHANNEL / group

and you don't need the following note I think.

onnx/defs/nn/defs.cc
pranavsharma: I would actually prefer that we rename C and M. The naming is confusing. In the case of conv transpose, num_output_channels = W.shape[1]/group. W.shape[1] is actually a multiplier for num_output_channels. W.shape[0] corresponds to num_input_channels (X.shape[1]). 
If we keep the same names, the spec should be C x M/group.

onnx/defs/nn/defs.cc
pranavsharma: If you look at the caffe2 implementation, X.shape[1] = W.shape[1]*group. In the terminology of the spec, this translates to DATA_CHANNEL = FILTER_OUT_CHANNEL where W.shape[1] = FILTER_OUT_CHANNEL/group.

houseroad: What's the meaning of FILTER_OUT_CHANNEL? does it mean the channel of the output?

pranavsharma: Yes, that's how the implementation interprets it. FILTER_OUT_CHANNEL is number of channels in the output.

houseroad: There are two problems in this case, 
1) the denotation filter should be [out, in, spatial, spatial]
2) the w.shape[1] * group == x.shpae[1] == C, so the W's shape should be [M, C/group, kH, kW]

onnx/defs/nn/defs.cc
houseroad: Yes, this should be [C, M/group, spatial, spatial, ...]

houseroad: I would prefer to keep the names. But I am open to the new names, if you can think of any better name.

onnx/defs/nn/defs.cc
pranavsharma: +1 for #2. 
Regarding #1, are you suggesting that we change the denotation as well in the spec to [FILTER_OUT_CHANNEL, FILTER_IN_CHANNEL, FILTER_SPATIAL, FILTER_SPATIAL ...] ?

onnx/defs/nn/defs.cc
pranavsharma: Sure, lets keep [C, M/group, spatial, spatial, ...]

PR #1290, opened 2018-08-14
ONNXIFI cpp test driver

zrphercule:
the onnxifi cpp test driver.
using gtest to test all protobuf data in <model, <input, output>> structured dir.
need to specify the target dir.

Example of testing:
pip install -e .
./.setuptools-cmake-build/onnxifi_test_driver_gtests onnx/backend/test/data/node 

the second argument is the target dir.


onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+	}
+	DIR* directory = opendir(target_dir.c_str());
+	if (directory == NULL){
+		fprintf(stderr, "Error: cannot open directory %s when loading test data: %s\n",
+			target_dir.c_str(), strerror(errno));
+		return -1;
+	}
+
+	while (true){
+		errno = ERRNO_DIR_END;
```
Maratyszcza: Could you just use `0`? It is well-known that `0` means "no error" for `errno`, but `ERRNO_DIR_END` looks like an error code.

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+	TestDriver test_driver;
+	test_driver.fetchAllTestCases(location);
+	return test_driver.testcases_;
+}
+
+void loadSingleFile(const std::string& filename, std::string& filedata){
+	std::ifstream f(filename, std::ifstream::in);
+	if (f.is_open()){
+			std::string line;
+		getline(f, filedata);
```
Maratyszcza: Do you use this function only for text files?

zrphercule: This is actually a protobug file, but I would like to read it byte by byte

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+
+//Need to be migrated to util in the future
+onnxTensorDescriptorV1 ProtoToOnnxTensorDescriptor(const onnx::TensorProto& proto_tensor){
+	onnxTensorDescriptorV1 onnx_tensor;
+	onnx_tensor.tag = ONNXIFI_TAG_TENSOR_DESCRIPTOR_V1;
+	onnx_tensor.name = proto_tensor.name().c_str();
+	onnx_tensor.dataType = proto_tensor.data_type();
+	onnx_tensor.memoryType = ONNXIFI_MEMORY_TYPE_CPU;
+	onnx_tensor.dimensions = proto_tensor.dims().size();
+	onnx_tensor.shape = (uint64_t *)proto_tensor.dims().data();
```
Maratyszcza: what is the data type of `proto_tensor.dims().data()`? this doesn't look safe

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,140 @@
+#include "driver/gtest_utils.h"
+#include "onnx/checker.h"
+#include "onnx/onnxifi.h"
+#include "onnx/onnxifi_loader.h"
+#define ONNXIFI_BACKEND_USED false
```
Maratyszcza: `const bool` please

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,140 @@
...
+	  std::vector<ProtoTestData> protos_;
+	  onnx::ModelProto model_;
+	  void SetUp() override {
+		ProtoTestCase t = GetParam();
+		protos_ = t.proto_test_data_;
+		model_ = t.model_;
+	  }
+	  uint64_t getDescriptorSize(onnxTensorDescriptorV1 *t){
+		uint64_t d_size = 1;
+		if (t->dimensions == 0) return 0;
```
Maratyszcza: Tensor with 0 dimensions is a scalar, it has size 1 (in # of elements)

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,140 @@
...
+                for (auto output : proto_test_data.outputs_) {
+                  output_descriptor.push_back(
+                      ProtoToOnnxTensorDescriptor(output));
+                  onnxTensorDescriptorV1 result;
+                  result.tag = ONNXIFI_TAG_TENSOR_DESCRIPTOR_V1;
+                  result.name = "result";
+                  result.dataType = output.data_type();
+                  result.memoryType = ONNXIFI_MEMORY_TYPE_CPU;
+                  result.dimensions = output.dims().size();
+                  result.shape = (unsigned long long*)output.dims().data();
```
Maratyszcza: Why `unsigned long long*`? IIRC, it should be `uint64_t`

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,140 @@
...
+                  result.buffer =
+                      (onnxPointer) new char[sizeof(output.raw_data().size())];
```
Maratyszcza: Memory leak here. Don't use new, instead allocate `std::vector<uint8_t>` of required size

onnx/backend/test/cpp/driver/test_driver.cc
zrphercule: It is a long long *, so the only difference is the "unsigned". I think this is safe?

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,80 @@
...
+class TestDriver{
+
+	std::string default_dir_;
+
+	void setDefaultDir(const std::string& s);
+	public:
+	std::vector<TestCase> testcases_;
+	TestDriver(const std::string default_dir = "."){
+		default_dir_ = default_dir_;
+		testcases_.clear();
```
bddppq: why is this needed?

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,140 @@
...
+              EXPECT_EQ(lib.onnxReleaseGraph(graph), ONNXIFI_STATUS_SUCCESS);
+            }
+        }
+};
+TEST_P(ONNXCppDriverTest, ONNXCppDriverUnitTest){
+	onnxifi_library lib;
+	onnxBackendID backendID;
+	onnxBackend backend;
+        // using default onnxifi backend
+        if (ONNXIFI_BACKEND_USED) {
```
bddppq: Is this used to distinguish our dummy backend and the other real backends?

onnx/backend/test/cpp/driver/test_driver.h
zrphercule: You are right...

onnx/backend/test/cpp/driver_test.cc
zrphercule: By now I only check the model and the tensor as we dont have real backends, since loading dummy backend here is useless.

onnx/backend/test/cpp/driver_test.cc
zrphercule: Yes it is uint64_t*, but I think there is no difference between these two? But for consistent I will change this to uint64_t*.

onnx/backend/test/cpp/driver_test.cc
bddppq: Got it. In Python land test cases can throw a special exception to indicate skipping a test at *runtime*. Just found gtest doesn't it yet. There is an open pr https://github.com/google/googletest/pull/1544 to add it to gtest, would be good to use it once it's merged to avoid special logic for the dummy backend.

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,79 @@
...
+	TestData(const std::vector<std::string> &input_filenames, const std::vector<std::string> &output_filenames){
+		input_filenames_ = input_filenames;
+		output_filenames_ = output_filenames;
+	}
+};
+struct TestCase{
+	TestCase(
+		const std::string& model_filename,
+		const std::string& model_dirname,
+		const std::vector<TestData>& test_data){
```
yinghai: You can use the ctor initializer to initialize the members. 

https://en.cppreference.com/w/cpp/language/initializer_list

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,79 @@
...
+
+#include "gtest/gtest.h"
+
+struct TestData{
+	std::vector<std::string> input_filenames_;
+	std::vector<std::string> output_filenames_;
+
+	TestData(){}
+
+	TestData(const std::vector<std::string> &input_filenames, const std::vector<std::string> &output_filenames){
```
yinghai: Same here.

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,79 @@
...
+	std::string default_dir_;
+
+	void setDefaultDir(const std::string& s);
+	public:
+	std::vector<TestCase> testcases_;
+	TestDriver(const std::string default_dir = "."){
+		default_dir_ = default_dir_;
+	}
+
+	int fetchAllTestCases(const std::string& _target_dir);
```
yinghai: `_target_dir` seems to be an odd naming convention here.

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,79 @@
...
+#include <cstdio>
+
+#include "onnx/onnx_pb.h"
+#include "onnx/proto_utils.h"
+#include "onnx/onnxifi_loader.h"
+#include "onnxifi.h"
+
+#include "gtest/gtest.h"
+
+struct TestData{
```
yinghai: It would be good to add high level doxygen style comments on what the class and functions are about.

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+	return false;
+}
+
+void TestDriver::setDefaultDir(const std::string &s){
+	default_dir_ = s;
+}
+
+//load single test case in case_dir to _testcases
+int TestDriver::fetchSingleTestCase(const std::string& case_dir) {
+  std::string model_name = case_dir;
```
yinghai: Do we need to check `/` at the end of `case_dir`?

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+  model_name += "model.onnx";
+  if (fileExist(model_name)) {
```
yinghai: nit: `FileExists`

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+    TestCase test_case;
+    test_case.model_filename_ = model_name;
+    test_case.model_dirname_ = case_dir;
+    int case_count = 0;
```
yinghai: Move this into the `for` loop?

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+    for (;; case_count++) {
+      std::vector<std::string> input_filenames, output_filenames;
+      std::string input_name, output_name;
+      std::string case_dirname = case_dir;
+      case_dirname += "test_data_set_" + std::to_string(case_count);
```
yinghai: `std::to_string` doesn't work on Android. Check `caffe2::to_string`.

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+      for (int data_count = 0;; data_count++) {
+        input_name =
+            case_dirname + "/input_" + std::to_string(data_count) + ".pb";
+        output_name =
+            case_dirname + "/output_" + std::to_string(data_count) + ".pb";
+        if (!fileExist(input_name) && !fileExist(input_name)) {
+          break;
+        }
+        if (fileExist(input_name)) {
+          input_filenames.push_back(input_name);
```
yinghai: small optimization would be 
```
input_filenames.emplace_back(std::move(input_name));
```

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+        }
+        if (fileExist(output_name)) {
+          output_filenames.push_back(output_name);
+        }
+      }
+      TestData test_data(input_filenames, output_filenames);
+      test_case.test_data_.push_back(test_data);
+    }
+    testcases_.push_back(test_case);
```
yinghai: `emplace_back` here too.

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+        output_name =
+            case_dirname + "/output_" + std::to_string(data_count) + ".pb";
+        if (!fileExist(input_name) && !fileExist(input_name)) {
+          break;
+        }
+        if (fileExist(input_name)) {
+          input_filenames.push_back(input_name);
+        }
+        if (fileExist(output_name)) {
+          output_filenames.push_back(output_name);
```
yinghai: What if we have input_file but not output_file?

onnx/backend/test/cpp/driver/test_driver.cc
yinghai: There seems to be one-liner: https://stackoverflow.com/questions/3203452/how-to-read-entire-stream-into-a-stdstring

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+	f.close();
+}
+
+ProtoTestCase loadSingleTestCase(const TestCase& t){
+
+	ProtoTestCase st;
+	loadSingleFile(t.model_filename_, st.raw_model_);
+	ParseProtoFromBytes(&st.model_, st.raw_model_.c_str(), st.raw_model_.size());
+
+	for (auto test_data : t.test_data_){
```
yinghai: We are making copies here. Is this intended?

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+		ProtoTestData proto_test_data;
+
+		for (auto input_file : test_data.input_filenames_){
```
yinghai: Same here and below.

onnx/backend/test/cpp/driver/test_driver.cc
yinghai: Doesn't look safe to me, either.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,138 @@
+#include "driver/gtest_utils.h"
+#include "onnx/checker.h"
+#include "onnx/onnxifi.h"
+#include "onnx/onnxifi_loader.h"
+const bool ONNXIFI_BACKEND_USED = false;
+class ONNXCppDriverTest
```
yinghai: no namespace?

onnx/backend/test/cpp/test_main.cc
```diff
@@ -0,0 +1,19 @@
+#include <iostream>
+#include "driver/gtest_utils.h"
+
+std::vector<ProtoTestCase> all_test_cases;
+GTEST_DEFINE_string_(
+    TARGET_DIR,
+    ::testing::internal::StringFromGTestEnv("TARGET_DIR", "auto"),
+    "The target repo of onnxifi data.");
+GTEST_API_ int main(int argc, char** argv){
+	std::cout << "Running main() from cpp test driver: cpp/test_main.cc" << std::endl;
```
yinghai: remove?

onnx/backend/test/cpp/driver/gtest_utils.h
```diff
@@ -0,0 +1,4 @@
+#include "gtest/gtest.h"
+#include "test_driver.h"
+
+extern std::vector<ProtoTestCase> all_test_cases;
```
yinghai: Is this some quirky requirement from gtest?

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,79 @@
...
+
+struct ProtoTestData{
+	std::vector<std::string> raw_inputs_;
+	std::vector<std::string> raw_outputs_;
+	std::vector<onnx::TensorProto> inputs_;
+	std::vector<onnx::TensorProto> outputs_;
+};
+
+struct ProtoTestCase{
+	std::string raw_model_;
```
yinghai: This seems to be a temporary variable to me, because its purpose is to create the `model_`.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,138 @@
...
+            for (auto proto_test_data : protos_) {
+              for (auto input : proto_test_data.inputs_) {
+                onnx::checker::check_tensor(input, ctx);
+              }
+              for (auto output : proto_test_data.outputs_) {
+                onnx::checker::check_tensor(output, ctx);
+              }
+            }
+            /* TO DO:
+             * This chunk of code is to test the performance of onnxifi backend.
```
yinghai: test performance or correctness?

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,138 @@
...
+             * Since we are not using a real backend, we should wait and not
+             * enable these tests. */
+
+            if (ONNXIFI_BACKEND_USED) {
+              onnxGraph graph;
+              uint32_t weightCount = model_.graph().initializer_size();
+              onnxTensorDescriptorV1 weightDescriptors =
+                  ProtoToOnnxTensorDescriptor(model_.graph().initializer(0));
```
yinghai: Why `initializer(0)`?

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,138 @@
...
+                      backend,
+                      NULL,
+                      sizeof(model_),
+                      &model_,
+                      weightCount,
+                      &weightDescriptors,
+                      &graph),
+                  ONNXIFI_STATUS_SUCCESS);
+
+              for (auto proto_test_data : protos_) {
```
yinghai: `const auto&`

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,138 @@
...
+                    lib.onnxRunGraph(graph, &inputFence, &outputFence),
+                    ONNXIFI_STATUS_SUCCESS);
+                EXPECT_EQ(
+                    lib.onnxWaitEvent(outputFence.event),
+                    ONNXIFI_STATUS_SUCCESS);
+                for (int i = 0; i < output_descriptor.size(); i++) {
+                  auto output_size = getDescriptorSize(&output_descriptor[i]);
+                  for (int j = 0; j < output_size; j++) {
+                    // size might be a problem!
+                    EXPECT_EQ(
```
yinghai: This is not good. We cannot compare them byte by byte because there might be numerical noise.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,138 @@
...
+                      ProtoToOnnxTensorDescriptor(output));
+                  onnxTensorDescriptorV1 result;
+                  result.tag = ONNXIFI_TAG_TENSOR_DESCRIPTOR_V1;
+                  result.name = "result";
+                  result.dataType = output.data_type();
+                  result.memoryType = ONNXIFI_MEMORY_TYPE_CPU;
+                  result.dimensions = output.dims().size();
+                  result.shape = (uint64_t*) output.dims().data();
+                  result.buffer =
+                      (onnxPointer) new char[sizeof(output.raw_data().size())];
```
yinghai: Who will release this?

onnx/backend/test/cpp/driver/test_driver.cc
Maratyszcza: To be safe, please copy shape values into `std::vector<uint64_t>` as use it from there. You can do it in one line with `std::vector(proto_tensor.dims().cbegin(), proto_tensor.dims().cend())`

onnx/backend/test/cpp/driver/test_driver.cc
Maratyszcza: Don't read binary files line-by-line! I bet you don't want to debug what happens when some lines end with `\n`, others with `\r\n`.

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+				return -1;
+			} else{
+				break;
+			}
+		}
+                std::string entry_dname = entry->d_name;
+                entry_dname = _target_dir + '/' + entry_dname + "/";
+                fetchSingleTestCase(entry_dname);
+        }
+cleanup:
```
Maratyszcza: label is unused. Also, `goto label` doesn't play well with C++ destructors

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,180 @@
...
+
+//load all test data in target_dir to _testcases
+int TestDriver::fetchAllTestCases(const std::string& _target_dir){
+	std::string target_dir = _target_dir;
+	if (target_dir == ""){
+		target_dir = default_dir_;
+	}
+	DIR* directory = opendir(target_dir.c_str());
+	if (directory == NULL){
+		fprintf(stderr, "Error: cannot open directory %s when loading test data: %s\n",
```
Maratyszcza: Since you write in C++, you should prefer `std::cerr` vs `fprintf(stderr, ...`

cmake/unittest.cmake
```diff
@@ -1,5 +1,6 @@
 set(UT_NAME ${PROJECT_NAME}_gtests)
 set(ONNX_ROOT ${PROJECT_SOURCE_DIR})
+set(DRIVER_NAME onnxifi_test_driver_gtests)
```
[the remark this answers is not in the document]
zrphercule: any recommendations?... I am lack of imagination now...

onnx/backend/test/cpp/driver/test_driver.cc
zrphercule: I guess it doesnt make sense if a node has no output? But I will check this at the beginning of this chunk of code.

onnx/backend/test/cpp/driver/test_driver.cc
zrphercule: oop...

onnx/backend/test/cpp/driver/gtest_utils.h
zrphercule: It is hard to pass these data from our backend to gtest. There I set this global var for both side to comunicate. Gtest has similar solution of using GFLAG, and it does the same thing of using a gloabl var.

onnx/backend/test/cpp/driver_test.cc
zrphercule: you are right....

onnx/backend/test/cpp/driver_test.cc
zrphercule: since we need the address of the first initializer. Or we should put all initializers in a vector first?

onnx/backend/test/cpp/driver_test.cc
zrphercule: yes, I agree with you, thats why I set a reminder here... wonder if you have any suggestions?

onnx/backend/test/cpp/driver_test.cc
zrphercule: I added namespace for test driver, but this part I think it is ok?

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,170 @@
...
+
+      output_name = case_dirname + "/output_" + "0" + ".pb";
+      if (!FileExists(output_name) && !FileExists(input_name)) {
+        break;
+      }
+
+      for (int data_count = 0;; data_count++) {
+        input_name = case_dirname + "/input_" + to_string(data_count) + ".pb";
+        output_name = case_dirname + "/output_" + to_string(data_count) + ".pb";
+        if (!FileExists(input_name) && !FileExists(input_name)) {
```
Maratyszcza: Could you cache the result of checks? `FileExists` is relatively expensive

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,112 @@
...
+ *	No real data was loaded in this type of class,
+ *	but only data location.
+ */
+struct TestData{
+	std::vector<std::string> input_filenames_;
+	std::vector<std::string> output_filenames_;
+
+	TestData(){}
+
+        TestData(
```
Maratyszcza: indentation off

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,112 @@
...
+	onnx::ModelProto model_;
+	std::vector<ProtoTestData> proto_test_data_;
+};
+
+/**
+ *	Store all unloaded test cases in one repo.
+ */
+class TestDriver{
+
+	std::string default_dir_;
```
yinghai: weid indent?

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,170 @@
...
+ProtoTestCase LoadSingleTestCase(const TestCase& t) {
+  ProtoTestCase st;
+  LoadSingleFile(t.model_filename_, st.raw_model_);
+  ParseProtoFromBytes(&st.model_, st.raw_model_.c_str(), st.raw_model_.size());
+
+  for (auto& test_data : t.test_data_) {
+    ProtoTestData proto_test_data;
+
+    for (auto& input_file : test_data.input_filenames_) {
+      std::string input_data = "";
```
yinghai:
```
std::string input_data;
```
is good enough.

onnx/backend/test/cpp/driver/utils.h
```diff
@@ -0,0 +1,21 @@
...
+#include <string>
+
+#include "onnx/onnx_pb.h"
+#include "onnx/proto_utils.h"
+#include "onnxifi.h"
+
+namespace onnx {
+namespace testing {
+
+template <typename T>
```
yinghai: nit: we can define this only for Android.

yinghai:
Windows build is failing?

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,187 @@
...
+namespace testing {
+bool FileExists(const std::string& filename) {
+  struct stat stats;
+  if (lstat(filename.c_str(), &stats) != 0 || !S_ISREG(stats.st_mode)) {
+    return false;
+  }
+  return true;
+}
+
+void TestDriver::SetDefaultDir(const std::string& s) {
```
rdzhabarov: is this planned to be used, do not see any references.

onnx/backend/test/cpp/test_main.cc
```diff
@@ -0,0 +1,14 @@
+#include <iostream>
+#include "driver/gtest_utils.h"
+
+std::vector<onnx::testing::ProtoTestCase> all_test_cases;
+GTEST_API_ int main(int argc, char** argv){
+        if (argc != 2) {
+          std::cout << "target directory must be given!" << std::endl;
+          return 0;
```
rdzhabarov: should probably return 1 on failure here.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,139 @@
...
+                result_descriptor.data()),
+            ONNXIFI_STATUS_SUCCESS);
+
+        onnxMemoryFenceV1 inputFence, outputFence;
+        inputFence.tag = ONNXIFI_TAG_MEMORY_FENCE_V1;
+        outputFence.tag = ONNXIFI_TAG_MEMORY_FENCE_V1;
+        inputFence.type = ONNXIFI_SYNCHRONIZATION_EVENT;
+        outputFence.type = ONNXIFI_SYNCHRONIZATION_EVENT;
+
+        // EXPECT_EQ(lib.onnxInitEvent(backend, inputFence),
```
rdzhabarov: why commented out? (seems legit)

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,139 @@
...
+        for (const auto& input : proto_test_data.inputs_) {
+          input_descriptor.push_back(
+              onnx::testing::ProtoToOnnxTensorDescriptor(input));
+        }
+        for (auto& output : proto_test_data.outputs_) {
+          output_descriptor.push_back(
+              onnx::testing::ProtoToOnnxTensorDescriptor(output));
+          onnxTensorDescriptorV1 result;
+          result.tag = ONNXIFI_TAG_TENSOR_DESCRIPTOR_V1;
+          result.name = "result";
```
rdzhabarov: since we expect multiple outputs should the name be unique per output?

onnx/backend/test/cpp/driver/test_driver.cc
zrphercule: It is not yet to be used (including default_dir it self), but I think it would be useful in the future, and unharm to keep it now.

onnx/backend/test/cpp/driver_test.cc
zrphercule: @rdzhabarov I am not sure about the mechanism of this name, must it be unique? Or it should be unique to distinguish. 
Also, these results are only a temporary vars.
@Maratyszcza What do you think?

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,195 @@
...
+  default_dir_ = s;
+}
+
+//load single test case in case_dir to _testcases
+int TestDriver::FetchSingleTestCase(const std::string& case_dir) {
```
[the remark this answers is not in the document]
zrphercule: actually I have doxygen comment of this function in test_driver.h, although it is not clear enough. I will update that one. @rdzhabarov

cmake/unittest.cmake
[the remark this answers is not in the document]
rdzhabarov: maybe, just ONNXIFI_TEST_DRIVER

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ *	Store all unloaded test cases in one repo.
+ */
+class TestDriver {
+  std::string default_dir_;
+
+ public:
+  void SetDefaultDir(const std::string& s);
+  std::vector<TestCase> testcases_;
```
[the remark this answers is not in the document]
zrphercule: I will change them to public...

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,118 @@
...
+   *
+   *	Regular file: model.onnx, store the protobuf of the model.
+   *	Repo(s): test_data_set_X: store the input&output data.
+   *
+   *	Each repo in test_data_set_X should include following contents.
+   *
+   *	Regular file(s): input_X.pb, store one input tensor.
+   *	Reuglar file(s): output_X.pb, stroe one output tensor.
+   */
+  int FetchSingleTestCase(const std::string& case_dir);
```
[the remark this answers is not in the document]
zrphercule: oop, it has no use. I will remove it.

onnx/backend/test/cpp/driver/test_driver.h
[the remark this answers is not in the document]
zrphercule: @rdzhabarov Actually... I dont really want anyone to simply modify default_dir_ by themselves, and in order to do this they need to cal SetDefaultDir. default_dir_ is only useful for this class internally, so no need to expose to others.

onnx/backend/test/cpp/driver_test.cc
Maratyszcza: Name in tensor descriptors correspond to tensor names in the model, which must be unique

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,240 @@
+#include "driver/gtest_utils.h"
+
+#include "onnx/checker.h"
+#include "onnx/onnxifi.h"
+#include "onnx/onnxifi_loader.h"
+
+const bool ONNXIFI_BACKEND_USED = false;
```
rdzhabarov: how is it planned to be used in the future?
Say I have an implementation of the ONNXIFI backend, how would I use the driver to run integration test?

If steps are not obvious it would be good to document.

zrphercule: I think you are right. I would need to document this.

rdzhabarov: Is the plan to use #ifdefs here for ONNXIFI_BACKEND_USED?

zrphercule: @rdzhabarov In fact we are doing such things:
1. We dont specify onnxifi backend for test driver, but only use by default backend. That does not mean we only use one backend, but means we do not indicate the target backend in test driver, but indicate it in onnxifi_loader. You can check the logic of onnxifi_loader when given empty backend location for more details.
2. As for ONNXIFI_BACKEND_USED, we will finnally enable it when we have valid by default backend to test. Now since we dont have any valid backend, in order to pass gtest, we set this temporary var. We will finnally remove it, or set it to true if by default backend detected.

zrphercule:
@yinghai @Maratyszcza @bddppq @rdzhabarov @houseroad Any more comments? btw I will add support for float16 in a seperate pr immediately after the merge of this pr, and also I will keep maintaining this test driver so please dont worry if there is anything imperfect, just tell me and I will fix it =)

onnx/proto_utils.h
```diff
@@ -3,6 +3,8 @@
 #include <google/protobuf/io/coded_stream.h>
 #include <google/protobuf/io/zero_copy_stream_impl_lite.h>
 
+#include "onnx/onnx_pb.h"
```
houseroad: Why add this?

zrphercule: Because it is needed... It used some function under onnx_pb.h, therefore if you include proto_utils.h without include onnx_pb.h, there will be a compile error.

onnx/backend/test/cpp/driver/utils.h
```diff
@@ -0,0 +1,21 @@
+#include <sstream>
```
houseroad: Seems this file is pretty useful for our users, shall we move it out of `test/cpp/driver`? to some onnxifi_utils.h/cc?

cmake/unittest.cmake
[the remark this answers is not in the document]
zrphercule: @rdzhabarov but I would like to keep the integrity of name, since all other gtest have "_gtest" as the end of their name...

onnx/backend/test/cpp/driver/utils.h
zrphercule: @houseroad Yes I do agree with you. Just wondering where should it be moved to.

onnx/proto_utils.h
houseroad: I don't think so... I guess because in some other files, we didn't correctly import #onnx/onnx_pb.h"

onnx/backend/test/cpp/driver/utils.h
houseroad: How about the folder as onnxifi.h? which is onnx/

zrphercule: Sounds decent for now. At last I guess most of the onnxifi code in there will need to be migrated to somewhere else, since there are too many onnxifi files already...

onnx/proto_utils.h
zrphercule: @houseroad "class AttributeProto" is defined under onnx/onnx_pb.h, and used here...

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,230 @@
...
+#include <io.h>
+#include <windows.h>
+#else
+#include <dirent.h>
+#endif
+
+namespace ONNX_NAMESPACE {
+namespace testing {
+bool FileExists(const std::string& filename) {
+  struct stat stats;
```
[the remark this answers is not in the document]
zrphercule: yes you are right

onnx/backend/test/cpp/test_main.cc
```diff
@@ -0,0 +1,14 @@
+#include <iostream>
+#include "driver/gtest_utils.h"
+
+std::vector<ONNX_NAMESPACE::testing::ProtoTestCase> all_test_cases;
+GTEST_API_ int main(int argc, char** argv){
+        if (argc != 2) {
```
rdzhabarov: clang-format on all changes maybe.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,247 @@
...
+TEST_P(ONNXCppDriverTest, ONNXCppDriverUnitTest){
+	onnxifi_library lib;
+	onnxBackendID backendID;
+	onnxBackend backend;
+        // using default onnxifi backend
+        if (ONNXIFI_BACKEND_USED) {
+          EXPECT_TRUE(onnxifi_load(1, NULL, &lib));
+
+          size_t numBackends;
+          lib.onnxGetBackendIDs(&backendID, &numBackends);
```
rdzhabarov: i think we can fix that later, but as of now, numBackends is initialized with whatever is on a stack and call to onnxGetBackendIDs will likely fail.

onnx/backend/test/cpp/test_main.cc
```diff
@@ -0,0 +1,14 @@
...
+          std::cout << "target directory must be given!" << std::endl;
```
rdzhabarov: std::cerr

cmake/unittest.cmake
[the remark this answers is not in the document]
bddppq: @zrphercule I was talking about the "DRIVER_NAME" part, maybe rename it to "ONNXIFI_TEST_DRIVER" as @rdzhabarov suggested.

cmake/unittest.cmake
```diff
@@ -26,9 +27,11 @@ add_whole_archive_flag(onnx tmp)
 list(APPEND ${UT_NAME}_libs ${tmp})
 list(APPEND ${UT_NAME}_libs onnx_proto)
 list(APPEND ${UT_NAME}_libs onnxifi_loader)
+list(APPEND ${UT_NAME}_libs onnxifi)
```
bddppq: Better to separate the dependencies of onnxifi test driver from normal unittests (aka create another ${ONNXIFI_TEST_DRIVER}_libs instead of mixing it into ${UT_NAME}_libs.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,247 @@
+#include "driver/gtest_utils.h"
```
bddppq: use include path relative to onnx top directory

onnx/backend/test/cpp/driver_test.cc
bddppq: I agree with @rdzhabarov to make ONNXIFI_BACKEND_USED as an ifdef so it can be configured at compile time.


nit: I suggest using ONNXIFI_DUMMY_BACKEND (instead of ONNXIFI_BACKEND_USED) and in your code do
```
#ifndef ONNXIFI_DUMMY_BACKEND
  blablablabla(...)
#endif
```
so that normal real backends don't need to add special -D define to build (at the end it's the dummy backend being the special one to the test driver).

onnx/backend/test/cpp/driver/utils.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "onnx/onnx_pb.h"
+#include "onnx/onnxifi.h"
+#include "onnx/proto_utils.h"
+
+namespace ONNX_NAMESPACE {
+namespace testing {
+
+template <typename T>
+std::string to_string(T value) {
```
bddppq: this already exist in onnx/string_utils.h

onnx/backend/test/cpp/driver/gtest_utils.h
```diff
@@ -0,0 +1,3 @@
+#include "gtest/gtest.h"
+#include "test_driver.h"
+extern std::vector<ONNX_NAMESPACE::testing::ProtoTestCase> all_test_cases;
```
bddppq: make a wrapper function for this global variable to avoid SIOF issue

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,249 @@
...
+
+#include "onnx/checker.h"
+#include "onnx/onnxifi.h"
+#include "onnx/onnxifi_loader.h"
+#include "onnx/string_utils.h"
+
+// Now we set this macro to be enbaled,
+// but once we have valid onnxifi backend,
+// we should remove it and set it to be a compiler option.
+#define ONNXIFI_DUMMY_BACKEND
```
rdzhabarov: someone could start benefit from this driver_test if they have an implementation of the ONNXIFI interface.

If you allow it via the following:
```
#ifndef USE_ONNXIFI_BACKEND
#define USE_ONNXIFI_BACKEND false
#endif
```

If someone wants to enable ONNXIFI_BACKEND they would compile with -DONNXIFI_BACKEND=true.

onnx/backend/test/cpp/gtest_utils.cc
```diff
@@ -0,0 +1,12 @@
+#include "gtest_utils.h"
+std::vector<ONNX_NAMESPACE::testing::ProtoTestCase>
+    OnnxifiGtestWrapper::all_test_cases;
+std::vector<ONNX_NAMESPACE::testing::ProtoTestCase>
+OnnxifiGtestWrapper::GetTestCases() {
+  return OnnxifiGtestWrapper::all_test_cases;
```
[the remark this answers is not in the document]
zrphercule: Oh sure it is...

onnx/backend/test/cpp/driver_test.cc
zrphercule: Yes I guess that is better

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,249 @@
...
+    EXPECT_EQ(lib.onnxReleaseGraph(graph), ONNXIFI_STATUS_SUCCESS);
+#endif
+  }
+};
+
+/**
+ *	When testing backend, we do not specify target backend here,
+ *	but always use the by-default backend of onnxifi_loader.
+ *	Check onnxifi_loader.h for more detail of how to indicate specific
+ *backend.
```
houseroad: Nit: indent

onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -0,0 +1,122 @@
...
+   *	Fetch one test case from repo case_dir.
+   *	The structure of case_dir should include following contents.
+   *
+   *	Regular file: model.onnx, store the protobuf of the model.
+   *	Repo(s): test_data_set_X: store the input&output data.
+   *
+   *	Each repo in test_data_set_X should include following contents.
+   *
+   *	Regular file(s): input_X.pb, store one input tensor.
+   *	Reuglar file(s): output_X.pb, store one output tensor.
```
rdzhabarov: typo

onnx/backend/test/cpp/driver_test.cc
zrphercule: oops

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,249 @@
...
+ */
+TEST_P(ONNXCppDriverTest, ONNXCppDriverUnitTest){
+	onnxifi_library lib;
+	onnxBackendID backendID;
+	onnxBackend backend;
+#if !ONNXIFI_DUMMY_BACKEND
```
rdzhabarov: I think you could use regular `if` here (not #if)

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,232 @@
...
+    test_case.model_filename_ = model_name;
+    test_case.model_dirname_ = case_dir;
+    for (int case_count = 0;; case_count++) {
+      std::vector<std::string> input_filenames, output_filenames;
+      std::string input_name, output_name;
+      std::string case_dirname = case_dir;
+      case_dirname += "test_data_set_" + ONNX_NAMESPACE::to_string(case_count);
+
+      output_name = case_dirname + "/output_" + "0" + ".pb";
+      if (!FileExists(output_name) && !FileExists(input_name)) {
```
houseroad: input_name is empty?

onnx/backend/test/cpp/driver_test.cc
zrphercule: Sounds decent

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -0,0 +1,232 @@
...
+      do {
+        std::string entry_dname = file.name;
+        if (entry_dname != "." && entry_dname != "..") {
+          entry_dname = target_dir + "/" + entry_dname + "/";
+          FetchSingleTestCase(entry_dname);
+        }
+      } while (_findnext(lf, &file) == 0);
+    } catch (const std::exception& e) {
+      _findclose(lf);
+      throw;
```
houseroad: Shall we add some error info?

onnx/backend/test/cpp/driver/test_driver.cc
zrphercule: Oh at first I only used output_name to judge, so I didnt add input_name here. I will fix it.

onnx/backend/test/cpp/driver/test_driver.cc
zrphercule: OK

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,249 @@
...
+#ifndef ONNXIFI_DUMMY_BACKEND
+#define ONNXIFI_DUMMY_BACKEND false
+#endif
+
+const float ONNXIFI_TESTDATA_EPS = 1e-5;
+
+template <typename T>
+class CompareOnnxifiData {
+ public:
+  bool IsEqual(void* P, void* Q) {
```
houseroad: Shall we add a parameter `eps`, and make ONNX)FI_TESTDATA_EPS as default value?

Also feel this is quite useful, probably lift it util?

onnx/backend/test/cpp/driver_test.cc
zrphercule: Yes you are right, but I wonder what number should be set to this numBackends?

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,249 @@
...
+    model_ = t.model_;
+  }
+  uint64_t GetDescriptorSize(const onnxTensorDescriptorV1* t) {
+    uint64_t d_size = 1;
+    for (int i = 0; i < t->dimensions; i++) {
+      d_size *= t->shape[i];
+    }
+    return d_size;
+  }
+  bool IsDescriptorEqual(
```
houseroad: This also seems very general, maybe lift it to the common/utils folder?

onnx/backend/test/cpp/driver_test.cc
zrphercule: @houseroad For the eps I think you are right, I will do it. As for IsEqual, I think it is fine to leave it here now. If there is other scenario gonna use this in the future, we can do it then. 

In fact, I am thinking of building a whole operator system for onnxifi backend, since I dont think we have now...

houseroad: I am about to do it in the python world (seriously), are you interested in ?

zrphercule: Sure, glad if I can help

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,249 @@
...
+ *	When testing backend, we do not specify target backend here,
+ *	but always use the by-default backend of onnxifi_loader.
+ *	Check onnxifi_loader.h for more detail of how to indicate specific
+ *  backend.
+ */
+TEST_P(ONNXCppDriverTest, ONNXCppDriverUnitTest){
+	onnxifi_library lib;
+	onnxBackendID backendID;
+	onnxBackend backend;
+        if (!ONNXIFI_DUMMY_BACKEND) {
```
houseroad: Nit: indent

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -0,0 +1,249 @@
+#include "gtest_utils.h"
+
+#include "onnx/checker.h"
+#include "onnx/onnxifi.h"
+#include "onnx/onnxifi_loader.h"
+#include "onnx/string_utils.h"
+
+#ifndef ONNXIFI_DUMMY_BACKEND
```
houseroad: We should add some comments to explain how to use the macro :-)

PR #1296, opened 2018-08-15
Scan operation

gramalingam:
Here is an updated PR for a scan operation based on the ONNX workshop discussions.


gramalingam:
A couple of questions where feedback would help:

(1)	The scan axis is basically the sequence axis. Question: do we need to support sequence-axis other than 1 (with batch-axis being 0)? If sequence axis is other than 1, then some form of copying (transposition) will be required in each iteration. Is it useful enough to add?
(2)	If we want only sequence-axis = 1, we can drop the scan_axes attribute. We would then need some other way then to determine the number of scan inputs and number of state variables. We could either add an attribute (number_of_scan_inputs, may be with a default value of 1) or make directions non-optional.
(3)	For variable length inputs, should the specification require the padded values in the output to be zero? Or, can the padded values be “undefined”?

docs/Changelog.md
```diff
@@ -8224,6 +8224,173 @@ This version of the operator has been available since version 8 of the default O
...
+#### Version
+
+This version of the operator has been available since version 8 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>body</tt> : graph (required)</dt>
+<dd>The graph run each iteration. It has N+M inputs: (loop state variables..., scan_input_elts...). It has N+K outputs: (loop state variables..., scan_output_elts...). Each scan_output is created by concatenating the value of the specified scan_output_elt value at the end of each iteration of the loop. It is an error if the dimensions of these values change across loop iterations.</dd>
+<dt><tt>directions</tt> : list of strings</dt>
```
ebarsoum: Do we need direction per inputs or one global direction?

docs/Changelog.md
```diff
@@ -8224,6 +8224,173 @@ This version of the operator has been available since version 8 of the default O
...
+      Scan <
+          scan_axes = [axis_1, ..., axis_m],
+          body = loop-body
+      > (sequence_lengths, init_1, ..., init_n, scan_1, ..., scan_m)
+  
+  is equivalent to the following pseudo-code:
+  
+      // T.shape[0] denotes the batch-size of T
+      // The batch-size of scan_1, ..., scan_m are all required to be equal
+      batch_size = scan_1.shape[0];
```
ebarsoum: Do we need to support [batch, seq,...] and [seq, batch,...] through flag?

docs/Changelog.md
gramalingam: I think direction per input is better.

docs/Changelog.md
```diff
@@ -8224,6 +8224,173 @@ This version of the operator has been available since version 8 of the default O
...
+  to as loop-carried dependences in the context of loops). All these tensors are required to
+  have the same shape in each iteration of the loop (a restriction imposed to enable efficient
+  memory allocation). Many common usages involve a single scan_input tensor (where functionality
+  similar to scan, fold and map can be obtained). When more than one scan_input is used,
+  a behavior similar to zip is obtained.
+  
+  The attribute body must be a graph, specifying the computation to be performed in
+  every iteration. It takes as input the current values of the state_variables and
+  the current iterated element of the scan_inputs. It must return the (updated) values
+  of the state_variables and zero or more scan_output_element tensors. The values of the
```
skottmckay: Is the number of output tensors constant across loop iterations, so the body potentially updates none of them rather than returns zero scan_output_element tensors?

docs/Changelog.md
```diff
@@ -8224,6 +8224,173 @@ This version of the operator has been available since version 8 of the default O
...
+This version of the operator has been available since version 8 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>body</tt> : graph (required)</dt>
+<dd>The graph run each iteration. It has N+M inputs: (loop state variables..., scan_input_elts...). It has N+K outputs: (loop state variables..., scan_output_elts...). Each scan_output is created by concatenating the value of the specified scan_output_elt value at the end of each iteration of the loop. It is an error if the dimensions of these values change across loop iterations.</dd>
+<dt><tt>directions</tt> : list of strings</dt>
+<dd>An optional list of M directions. The i-th element of the list specifies the direction (forward or reverse) to be scanned for the i-th scan_input tensor. If omitted, all scan_input tensors will be scanned in the forward direction.</dd>
+<dt><tt>scan_axes</tt> : list of ints (required)</dt>
```
skottmckay: Are there any known use cases where scan_axes is required or could that be added later if necessary?

If required, the value assumes the first dimension of the input is the batch size right, so 0 would be valid and equate to the first axis of each row in the batch?

docs/Changelog.md
```diff
@@ -8224,6 +8224,173 @@ This version of the operator has been available since version 8 of the default O
...
+
+#### Version
+
+This version of the operator has been available since version 8 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>body</tt> : graph (required)</dt>
+<dd>The graph run each iteration. It has N+M inputs: (loop state variables..., scan_input_elts...). It has N+K outputs: (loop state variables..., scan_output_elts...). Each scan_output is created by concatenating the value of the specified scan_output_elt value at the end of each iteration of the loop. It is an error if the dimensions of these values change across loop iterations.</dd>
```
dzhulgakov: does it make sense to also add iteration_number to match Loop's format? It might be useful for doing some masking, but I'm not that sure

ebarsoum: I am not fan of one loop OP with multiple way to terminate.

docs/Changelog.md
gramalingam: I am waiting for other people to provide input … I think this is related to the question I asked up at the top also: do we want to allow flexibility with sequence-axis (and/or batch-axis) or just fix them (batch-axis = 0 and sequence-axis = 1) for now? I am fine either way.

docs/Changelog.md
gramalingam: Sorry, I don't understand the question. Yes, the number of outputs produced by the body sub-graph is constant across loop iterations. The values of some of these outputs may not vary from loop-iteration to loop-iteration. I didn't understand the point about zero scan_output_element.

docs/Changelog.md
gramalingam: Yes, I have the same question (your first question) ... and I am hoping somebody will speak up if they think it is needed. Otherwise, let's drop it for now.

docs/Changelog.md
gramalingam: I believe @dzhulgakov is talking about just providing the iteration number to the loop body, not about adding a trip-count … basically a scan/fold with index. We could do this … but if there is no compelling need right now, we can add further such ops (or extend this) later (just my opinion).

docs/Changelog.md
skottmckay: If there are zero output elements, what is the purpose of the iterations? Don't they need to produce something, or is it valid to just update the state variables?

gramalingam: We will still return the final-value of state-variables (this will be like the output Y_h of RNN). The scan_output_element tensors are concatenated and returned and that will be like the output Y of RNN.

gramalingam:
I updated the PR as per the discussions above: simplified things by assuming sequence-axis / scan-axis is 1. As Emad says, we can extend this by providing a Boolean flag (for each scan input) in case we want to flip the batch-axis and sequence-axis. We can add this in the future if there is demand.

onnx/defs/controlflow/defs.cc
```diff
@@ -227,4 +227,189 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        "number of scan_inputs M is assumed to be 1.",
+		AttributeProto::INT,
+        false
+	)
+    .Attr(
+		"directions",
+		"An optional list of M directions. The i-th element of the list specifies the direction "
+		"(forward or reverse) to be scanned for the i-th scan_input tensor. "
+        "If omitted, all scan_input tensors will be scanned in the forward direction.",
+		AttributeProto::STRINGS,
```
bddppq: I think it's better to use boolean or enum/int to indicate the direction rather than string.

onnx/defs/controlflow/defs.cc
```diff
@@ -227,4 +227,189 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+	.Attr(
+		"body",
+		"The graph run each iteration. It has N+M inputs: "
+		"(loop state variables..., scan_input_elts...). It has N+K outputs: "
+		"(loop state variables..., scan_output_elts...). Each "
+		"scan_output is created by concatenating the value of the specified "
+		"scan_output_elt value at the end of each iteration of the loop. It is an error"
+		" if the dimensions of these values change across loop iterations.",
+		AttributeProto::GRAPH,
+		true)
```
bddppq: nit: you don't need to specify `true` for required attributes.

onnx/defs/controlflow/defs.cc
```diff
@@ -227,4 +227,189 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+	.Attr(
```
bddppq: I would suggest making this attribute required.

docs/Changelog.md
bddppq: IIUC the loop index can be simulate by the loop state variable. This is not ideal, but if we see this pattern come up often than we can add it in a backward compatible way.

onnx/defs/controlflow/defs.cc
gramalingam: I agree … but my goal was to be consistent with the current RNN/LSTM/GRU spec. It might be better to change all of these uniformly, if we want to use an enum/int.

bddppq: Here there are only two choices, while for RNN there can be potentially many different activation functions.

gramalingam: I was referring to the "direction: STRING" attribute (Specify if the RNN is forward, reverse, or bidirectional. Must be one of forward (default), reverse, or bidirectional.)

gramalingam: But I can change it to Boolean … but the inconsistency might cause some confusion too.

bddppq:
@gramalingam ping for regenerating the docs for adding new operators

gramalingam:
@bddppq , which docs? I ran "python onnx\defs\gen_doc.py" and added all changed doc files. What am I missing?

bddppq:
@gramalingam If you look at the failed CI, it says
```
+git status
HEAD detached at FETCH_HEAD
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)
    
    modified: docs/TestCoverage.md
...
```
Which means TestCoverage.md file is outdated. Please run `python onnx/backend/test/stat_coverage.py` to re-generate it.
@houseroad Maybe add this command to https://github.com/onnx/onnx/blob/master/docs/CONTRIBUTING.md?

PR #1297, opened 2018-08-15
ControlFlow operators update

bddppq:
Move If and Loop out of experimental
Drop LoopIndexTensor

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -41,6 +41,7 @@
   * <a href="#HardSigmoid">HardSigmoid</a>
   * <a href="#Hardmax">Hardmax</a>
   * <a href="#Identity">Identity</a>
+  * <a href="#If">If</a>
   * <a href="#InstanceNormalization">InstanceNormalization</a>
   * <a href="#LRN">LRN</a>
   * <a href="#LSTM">LSTM</a>
@@ -48,6 +49,7 @@
   * <a href="#Less">Less</a>
   * <a href="#Log">Log</a>
   * <a href="#LogSoftmax">LogSoftmax</a>
+  * <a href="#Loop">Loop</a>
   * <a href="#LpNormalization">LpNormalization</a>
   * <a href="#LpPool">LpPool</a>
   * <a href="#MatMul">MatMul</a>
@@ -111,10 +113,7 @@
   * <sub>experimental</sub> <a href="#Crop">Crop</a>
   * <sub>experimental</sub> <a href="#GRUUnit">GRUUnit</a>
   * <sub>experimental</sub> <a href="#GivenTensorFill">GivenTensorFill</a>
-  * <sub>experimental</sub> <a href="#If">If</a>
   * <sub>experimental</sub> <a href="#ImageScaler">ImageScaler</a>
-  * <sub>experimental</sub> <a href="#Loop">Loop</a>
-  * <sub>experimental</sub> <a href="#LoopIndexTensor">LoopIndexTensor</a>
   * <sub>experimental</sub> <a href="#MeanVarianceNormalization">MeanVarianceNormalization</a>
   * <sub>experimental</sub> <a href="#ParametricSoftplus">ParametricSoftplus</a>
   * <sub>experimental</sub> <a href="#Scale">Scale</a>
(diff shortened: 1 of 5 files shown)


docs/Operators.md
```diff
@@ -4107,6 +4106,47 @@ expect(node, inputs=[data], outputs=[data],
...
+</dl>
+
+#### Inputs
+
+<dl>
+<dt><tt>cond</tt> : B</dt>
+<dd>Condition for the if</dd>
+</dl>
+
+#### Outputs (1 - &#8734;)
```
ebarsoum: Does both statements need to return the same Outputs, and types?

docs/Operators.md
```diff
@@ -4943,6 +4983,163 @@ expect(node, inputs=[x], outputs=[y],
 </details>
 
 
+### <a name="Loop"></a><a name="loop">**Loop**</a>
+
+  Generic Looping construct. This loop has multiple termination conditions:
+  
+  1) Trip count. Iteration count specified at runtime. Set by
```
ebarsoum: With Scan being acceptable also, do we still need trip_count?

docs/Operators.md
```diff
@@ -4943,6 +4983,163 @@ expect(node, inputs=[x], outputs=[y],
...
+        %b[INT32, scalar]
+      ) {
+        %my_local = Add(%a, %b)
+        %b_out = Sub(%a, %b)
+        %keepgoing_out = Greater(%my_local, %b_out)
+        %user_defined_vals = Add(%b, %b)
+        return %keepgoing_out, %b_out, %user_defined_vals
+      }
+  
+  *Sample equivalent C code*
```
ebarsoum: How nested Loop will look like?

for (size_t r=0; r< height; r++) {
    for (size_t c=0; c< width; c++) {
        process(m[r][c])
    }
}

docs/Operators.md
```diff
@@ -4943,6 +4983,163 @@ expect(node, inputs=[x], outputs=[y],
...
+
+#### Version
+
+This version of the operator has been available since version 1 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>body</tt> : graph (required)</dt>
+<dd>The graph run each iteration. It has 2+N inputs: (iteration_num, condition, loop carried dependencies...). It has 1+N+K outputs: (condition, loop carried dependencies..., scan_outputs...). Each scan_output is created by concatenating the value of the specified output value at the end of each iteration of the loop. It is an error if the dimensions of these values change across loop iterations.</dd>
```
ebarsoum: Why we have this restriction?

docs/Operators.md
dzhulgakov: It's a provided number of iterations (as a scalar tensor) vs picking the dim in Scan op

I think fixed trip count is useful - it saves from having to do add-1 manually. Also it'd make it easier if we want to translate ScanOp into generic Loop

docs/Operators.md
dzhulgakov: Because it's invalid to concatenate non-matching tensor shapes? I think the same is the case for Loop op actually (padding on max seq size is different)

docs/Operators.md
dzhulgakov: Yep, it sounds as we discussed.

docs/Operators.md
wschin: @dzhulgakov , is it better to explain how the tensors gets concatenated (and add an example)? For example, if I have 3 iterations and each iteration generates a 7x5 tensor as the only scan_output, is the final scan_output a 3x7x5 tensor? There are multiple ways to concatenate tensors, I believe.

docs/Operators.md
wschin: I am also wondering how scan_output can be used in nested Loops. Can the inner loop accumulate scan_output variables onto the outer loop's scan_output?

dzhulgakov: yes, why not. scan_output of the inner loop is a tensor and it can be passed as the output of the outer loop's body and also become the scan output

docs/Operators.md
dzhulgakov: it's probably more general to concatenate  by the first dimension. Then individual loop iteration can return not only slices of dim 1 (and can insert unsqeeze as needed).

So in your example each iteration would return 1x7x5 to end up with 3x7x5 output. Or each iteration can return 2x7x5, 0x7x5 and 4x7x5 making the output 6x7x5

Then we also need to add which  dimension to concat on similarly to Scan? That one actually has only input axis but hardcoded output axis to be 1.

ebarsoum: @dzhulgakov do we need the concatenation? Can we pass the needed data as extra parameter?

docs/Operators.md
bddppq: Updated the doc

docs/Operators.md
ebarsoum: I just talked to Rama, can we clarify the doc, is the restriction only for the scan_outputs or the scan_outputs and loop carry dependencies? We don't need restriction for the loop carry dependency, correct?

PR #1301, opened 2018-08-17
Coverage Report Tools for Backend Scoreboard

Ac2zoom:
Reporting tools  to enable backend scoreboard: capability to write table to CSV including other backend results.


onnx/backend/test/stat_coverage.py
```diff
@@ -5,6 +5,7 @@
 from __future__ import unicode_literals
 
 import os
+import sys
```
houseroad: This can be removed

onnx/backend/test/report/coverage.py
```diff
@@ -3,17 +3,23 @@
...
 from tabulate import tabulate  # type: ignore
 
 import onnx
 from onnx import defs, helper, GraphProto
-from typing import Optional, Text, Set, Dict, IO
+from typing import Optional, Text, Set, Dict, IO, List, Any
 
 _all_schemas = defs.get_all_schemas()
 
+model_coverage_whitelist = set(
```
houseroad: why do we need this whitelist?

onnx/backend/test/report/coverage.py
```diff
@@ -114,7 +128,116 @@ def report_text(self, writer):  # type: (IO[Text]) -> None
             else:
                 attrs_column = 'No attributes'
             rows.append([op_cov.op_type, attrs_column])
+            passed.append(op_cov.op_type)
         writer.write(tabulate(
             rows,
             headers=['Operator', 'Attributes\n(name: #values)'],
             tablefmt='plain'))
+        if os.environ.get(str('CSVDIR')) is not None:
```
houseroad: Can we split this part into a separate function? seems the logic is independent, and probably we should add some comments about how to use it?

onnx/backend/test/report/coverage.py
```diff
@@ -114,7 +125,119 @@ def report_text(self, writer):  # type: (IO[Text]) -> None
...
+            self.report_csv(all_ops, passed, experimental)
+
+        def report_csv(self, all_ops, passed, experimental):  # type: (List[Text], List[Text], List[Text]) -> None
```
houseroad: Add a summary, including the purpose, which env var will be used

PR #1304, opened 2018-08-17
Add note for Gather operator for index out of bound case

HectorSVC:
add notes for Gather operator if an out of bound index is encountered in indices

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -3369,6 +3369,9 @@ expect(node, inputs=[input, W, R, B], outputs=[Y_h.astype(np.float32)], name='te
         ],
     ]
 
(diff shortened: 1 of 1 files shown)


HectorSVC:
1. it aligns with TF: https://www.tensorflow.org/api_docs/python/tf/gather
2. The reason for not report error directly on GPU is that it's not easy to do this on device normally, it may slow the processing a lot, just set a default value will a lot faster for GPU case

houseroad:
I think probably we should leave the error handling to backends. Especially, in different frameworks, people have different handling. Especially, how to report the error is not well defined in ONNX.

prasanthpul:
CLosing due to lack of activity in a long time. Please reopen if needed

PR #1307, opened 2018-08-20
[Optimization Pass] Add optimization pass eliminating nop Pad

0wu:
[Why]
A Pad layer with argument pads = zeros does nothing and should be removed.
Such a nop pad is observed when converting torchvision.resnet to onnx

[How]
* We mimic the optimization pass of nop_transpose and created the nop_pads optimizer.
* Test cases are added ensuring the correctness.

### DIFF
--- a/onnx/common/interned_strings.h
+++ b/onnx/common/interned_strings.h
@@ -40,6 +40,7 @@ _(Tile) \
 _(SubConstant) \
 _(Scale) \
 _(Transpose) \
+_(Pad) \
 _(Reshape) \
 _(split) \
 _(chunk) \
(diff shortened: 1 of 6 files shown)


houseroad:
Can you also update the doc here? https://github.com/onnx/onnx/blob/master/onnx/examples/optimize_onnx.ipynb

houseroad:
I mean update the doc (ipynb) to show this optimization pass in the lists

0wu:
@houseroad The list indeed contains the `eliminate_nop_pad` in the list
```
Available optimization passes:
eliminate_identity
eliminate_nop_pad
eliminate_nop_transpose
eliminate_unused_initializer
extract_constant_to_initializer
fuse_add_bias_into_conv
[deleted]
```
 https://github.com/onnx/onnx/blob/a14c0f756b418183d7f347f23417a130fc2508a0/onnx/examples/optimize_onnx.ipynb

houseroad:
@0wu cool, I was blind.

PR #1317, opened 2018-08-23
Add the ability to deprecate an OpSchema

RyanUnderhill:
The case has come up for us where we would like to deprecate an op so that it's no longer available starting in a certain version. Our case is to deprecate an op in an experimental domain and move it to the core domain.

Originally we were thinking of specifying an until_version for a schema. This had some drawbacks, like needing to modify existing ops to set their until_versions, plus adding some complexity to the schema lookups to deal with ranges of versions. A simple 'deprecated' flag keeps things simple and to deprecate an op, one just registers it as deprecated starting with a particular version.

For example, to deprecate 'Add' starting in version 8:

`ONNX_OPERATOR_SET_SCHEMA(
    Add,
    8,
    OpSchema().Deprecate());
`

I tested deprecating an op in our backend for onnx to verify externally, before submitting here for discussion.


RyanUnderhill:
Doc generation calls the schema() functions, so deprecating an op leads to errors during document generation. I changed it so that the schema() function will still return a deprecated schema, and the caller has to do the check. This should be easy since a deprecated schema will have no inputs/outputs/type constraints in it's definition.

With the updated doc generation, this is a sample output:

`### <a name="Asin"></a><a name="asin">**Asin** (deprecated)</a>
#### Version
This version of the operator has been deprecated since version 7 of the default ONNX operator set.
#### Inputs
#### Outputs
#### Type Constraints
#### Examples
`

I used 'asin' as an example since deprecating operators that are used in node tests causes them to fail, so this was easier.

onnx/defs/schema.h
```diff
@@ -711,8 +725,8 @@ class OpSchemaRegistry final : public ISchemaRegistry {
         // All versions are less than specified version, or,
         // The <pos> version is greater than specified version.
         pos--;
-        return &(pos->second);
```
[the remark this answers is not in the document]
RyanUnderhill: This case fixes up the pos to be at the right index, the fallthrough case is the same line.

It was more obvious when I had the deprecated check, as it looked very redundant to do it once within the if and once right after it.

RyanUnderhill:
After the latest change, the documentation now looks like this:

`### <a name="Asin"></a><a name="asin">**Asin** (deprecated)</a>

#### Version

This version of the operator has been deprecated since version 7 of the default ONNX operator set.
`

linkerzhang:
@bddppq  any comments please?

onnx/defs/schema.cc
```diff
@@ -79,6 +79,10 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
 }
 
 void OpSchema::Verify(const NodeProto& node) const {
+  if (deprecated_) {
+    fail_check("Using a deprecated operator.");
```
bddppq: nit: Maybe say "Operator xxx has been deprecated since version yyy"

RyanUnderhill: Done, changed to this and verified:

`fail_check("Operator '", name_, "' has been deprecated since version ", since_version_);
`

RyanUnderhill:
@houseroad : My next step was to deprecate OneHotEncoder from ONNX_ML and add OneHot to the ONNX domain. Would that be a better example than adding back Upsample? (Plus experimental ops are just deleted, they shouldn't be deprecated)

PR #1318, opened 2018-08-23
Use target_compile_options instead of add_definitions when setting ONNX_NAMESPACE

daquexian:
All programs linked to onnx need to manually add `-DONNX_NAMESPACE=..` in their own compile flags, for example, [caffe2](https://github.com/pytorch/pytorch/blob/f0d8a36e709dbf6a4c3aed7faf0da2b113668fe7/cmake/Dependencies.cmake#L793). 

It is unfriendly for developers, and hard to figure out.

After `replacing add_definitions` with `target_compile_options`, one can just use `add_library(third_party/onnx)` to use onnx source in his program.


CMakeLists.txt
```diff
@@ -103,7 +103,6 @@ else()
   set(ONNX_NAMESPACE "onnx")
   set(MY_ONNX_NAMESPACE "-DONNX_NAMESPACE=onnx")
 endif()
-add_definitions(${MY_ONNX_NAMESPACE})
 
 if(ONNX_ML)
   add_definitions("-DONNX_ML=1")
```
daquexian: There is also an `add_definitions` here, but I don't know whether it is ok to replace it.

snnn:
It's better,  but I think onnx has 3 C/C++ projects and you only changed two of them.
And, if another static library, let's say, lib A, depends on onnx. it still has to manually add "-DONNX_NAMESPACE=.."

daquexian:
@snnn Thanks!

> if another static library, let's say, lib A, depends on onnx. it still has to manually add "-DONNX_NAMESPACE=.."

I think it is not the case. As long as a developer use `add_subdirectory` to add onnx sources into his library, the flag will automatically take effect.

In fact, a perfect c++ && cmake project should let downstreams directly use `find_package(xxx)` in CMakeLists.txt if the project has been installed, it is a feature of modern(namely 3.0+, I think) cmake ([link](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/)). For example, if onnx supports this feature, then developers can simply link their lib to onnx lib using the following code

```
find_package(onnx)
add_library(my_lib my_lib.cpp)
target_link_libraries(my_lib onnx)
```

on any PC which onnx is installed, the compile flags will automatically take effects too.

Sadly onnx doesn't support it, but my PR also helps when using `add_subdirectory` to directly build onnx sources rather than link the installed onnx lib.

snnn:
For a static library, you won't need to call "target_link_libraries(my_lib onnx)". As there is no link step for a static library.

daquexian:
@snnn Yes there is no link for static libraries, but compiler flags will also be spread from onnx to the intermediate static library, and then from the intermediate static library to the executable, as long as the intermediate library and the executable are using cmake properly (e.g. In my project, my static lib A depends on onnx, and my executable B depends on A, `-DONNX_NAMESPACE=...` also takes effect when compiling B). I have verified it in my project, but maybe there is something I'm overlooking. I'm looking forward to your opinion.

snnn:
Sounds good. Would you please also change the definition of "-DONNX_ML=1"?

daquexian:
@snnn @bddppq Thanks for your approving. It's better to replace `add_definition` with modern `target_...`.
 Do I need to make the `ONNX_ML=1` definition public (also take effect on downstream project) or private (only take effect on onnx itself)?

daquexian:
@bddppq Thanks! I have fixed ONNX_ML define, and also used more proper `target_compile_definitions` instead of `target_compile_options` for ONNX_NAMESPACE.

bddppq:
@daquexian Cool Thanks. I will merge it once CI have finished (and passed).

daquexian:
@bddppq Thanks! CI has passed :)

PR #1320, opened 2018-08-24
Add shape inference function of Upsample operator.

rockindy:
Add shape inference function of Upsample operator.


onnx/defs/tensor/defs.cc
```diff
@@ -1085,7 +1085,21 @@ ONNX_OPERATOR_SET_SCHEMA(
             "T",
             OpSchema::all_tensor_types(),
             "Constrain input and output types to all tensor types.")
-        .SetDoc(Upsample_ver7_doc));
+        .SetDoc(Upsample_ver7_doc)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateShapeAndTypeFromFirstInput(ctx);
+          auto& input_shape = getInputShape(ctx, 0);
```
houseroad: Please call `hasNInputShapes(ctx, 1)` to make sure we have the shape information of the first input.

houseroad:
@zrphercule this PR won't touch the doc. :-)

onnx/defs/tensor/defs.cc
```diff
@@ -1085,7 +1085,24 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .SetDoc(Upsample_ver7_doc)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          if (!hasNInputShapes(ctx, 1)) {
+            return;
+          }
+          propagateShapeAndTypeFromFirstInput(ctx);
+          auto& input_shape = getInputShape(ctx, 0);
+          auto* output_shape = getOutputShape(ctx, 0);
+          auto* scale_attr = ctx.getAttribute("scales");
+          if (input_shape.dim_size() != scale_attr->floats_size())
```
bddppq: Could you add `{}` around the body of if?

onnx/defs/tensor/defs.cc
```diff
@@ -1085,7 +1085,24 @@ ONNX_OPERATOR_SET_SCHEMA(
             "T",
             OpSchema::all_tensor_types(),
             "Constrain input and output types to all tensor types.")
-        .SetDoc(Upsample_ver7_doc));
+        .SetDoc(Upsample_ver7_doc)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          if (!hasNInputShapes(ctx, 1)) {
+            return;
+          }
+          propagateShapeAndTypeFromFirstInput(ctx);
```
bddppq: You probably only want to copy the input type to output here?

onnx/defs/tensor/defs.cc
rockindy: Sure.

onnx/defs/tensor/defs.cc
rockindy: My original intention is to prepare an output object for modifying. I will use propagateElemTypeFromInputToOutput() for less code instructions.
Thanks for your advice.

PR #1324, opened 2018-08-24
Fix non root user install issue

chinhuang007:
The install from source code is trying to create
files in /usr/local/bin where a non root user might
not have write permissions. The fix is to ensure
non root users wouldn't get 'error: [Errno 13] Permission
denied'.


chinhuang007:
Okay, in that case, we might need to look into alternative ways to setuptools + entry_points to set up these console script files.

houseroad:
@chinhuang007 feel free to reopen the pr if you have new proposal :-)

PR #1325, opened 2018-08-24
Best way to add_msvc_runtime_flag

lygstate:
Best way to implement add_msvc_runtime_flag for choose msvc runtime properly.
Refer to https://stackoverflow.com/questions/24460486/cmake-build-type-not-being-used-in-cmakelists-txt
So that cmake_args doesn't need to append build type. we choose the build type at the build phase.


zrphercule:
@onnxbot retest this please

lygstate:
Do not know why it's failed. weird.

setup.py
```diff
@@ -148,6 +148,7 @@ def run(self):
             os.makedirs(CMAKE_BUILD_DIR)
 
         with cd(CMAKE_BUILD_DIR):
+            build_type = 'Release'
```
bddppq: I think cmake's default build type is not necessary "Release"?

cmake/Utils.cmake
```diff
@@ -2,17 +2,9 @@
 # Add MSVC RunTime Flag
 function(add_msvc_runtime_flag lib)
   if(${ONNX_USE_MSVC_STATIC_RUNTIME})
-    if(${CMAKE_BUILD_TYPE} MATCHES "Debug")
-      target_compile_options(${lib} PRIVATE /MTd)
-    else()
-      target_compile_options(${lib} PRIVATE /MT)
-    endif()
+    target_compile_options(${lib} PRIVATE $<$<NOT:$<CONFIG:Debug>>:/MT> $<$<CONFIG:Debug>:/MTd>)
```
bddppq: What's the benefit of changing to use generator expression for this? IMO this new code looks more cryptic than the old version.

setup.py
bddppq: @raymondxyang `build_type` will be unconditionally appended to `cmake_args`

cmake/Utils.cmake
lygstate: https://stackoverflow.com/questions/34490294/what-does-configdebugrelease-mean-in-cmake

setup.py
lygstate: Refer to https://stackoverflow.com/questions/24460486/cmake-build-type-not-being-used-in-cmakelists-txt

lygstate: After https://stackoverflow.com/questions/24460486/cmake-build-type-not-being-used-in-cmakelists-txt, there is no need apped build_type to cmake_args.

lygstate:
@zrphercule Please take a look at this and merge this,

PR #1326, opened 2018-08-25
Add support for building with protobuf-lite

bddppq:


bddppq:
CI errors are unrelated to this PR, waiting for the fix in #1327

yinghai:
What's `protobuf-lite` for?

bddppq:
@yinghai it's a "lite" version of protobuf :-). From the [official doc](https://developers.google.com/protocol-buffers/docs/proto):
> LITE_RUNTIME: The protocol buffer compiler will generate classes that depend only on the "lite" runtime library (libprotobuf-lite instead of libprotobuf). The lite runtime is much smaller than the full library (around an order of magnitude smaller) but omits certain features like descriptors and reflection. This is particularly useful for apps running on constrained platforms like mobile phones. The compiler will still generate fast implementations of all methods as it does in SPEED mode. Generated classes will only implement the MessageLite interface in each language, which provides only a subset of the methods of the full Message interface.

.travis.yml
```diff
@@ -34,6 +34,14 @@ matrix:
     - os: osx
       osx_image: xcode9.3
       env: PYTHON_VERSION=python3 ONNX_ML=1
+    - os: linux
+      sudo: required
+      env: PYTHON_VERSION=python2 LITE=1
```
yinghai: Can we do `ONNX_ML=1` as well as `LITE=1`?

Maratyszcza:
Thanks a lot, @bddppq!

PR #1327, opened 2018-08-25
Fix build error caused by using std::set without including corresponding header

bddppq:
```
/home/travis/build/onnx/onnx/onnx/optimizer/passes/lift_lexical_references.h:80:3: error: ‘set’ in namespace ‘std’ does not name a type
   std::set<std::string> liftReferences(Graph* g, EnvStack *es) {
```


bddppq:
Fold into #1326 to avoid waiting since it's small

PR #1328, opened 2018-08-25
Adding a Go definition file

owulveryck:
I am trying to implement a completely new backend in Go.

# Purpose of the PR

To add an official Go Definition/API to the project.

# Why do I do this PR:

There is no, at the present time, Go API for ONNX. Everyone needs to compile it on their own.
I think that the community would benefit an "Official API".

This file I could be hosted it in any third party repo but: 
* this shall not be manually modified;
* It has to be generated by the official protobuf definition;
* **It has to be** the **official** (and **only**) API for any backend implementation in the Go language.

# About the code

The Go file has not been modified manually and is the result of a processing by protocol buffer compilers only. I have simply added a `Makefile` to be able to reproduce this step.
It could be good to add the mechanism into the ONNX integration mechanism but I have no clue on how to do that.


owulveryck:
The separate repo sounds like the best idea indeed. And to be honest, it will be a lot better to have a lighter repo. It will be more efficient for "go-getting" and update it.

jspisak:
Completely agree. A separate repo is definitely the way to go. @owulveryck - thanks for contributing to the project!

owulveryck:
How do we proceed then?
Is there any special Licence required for the repo? Any requirement that needs to be fulfilled? I have read the contribution guide but nothing seems to be related to this.

houseroad:
@owulveryck you may want to take a look at https://github.com/onnx/onnx-r

Since ONNX uses MIT License, you can stick to it you like. Setting up the CI and running tests will be recommended.

If you can create one repo under your user account for pre-review, it will be great.

bddppq:
Closing this one as discussed.

PR #1330, opened 2018-08-26
Add epsilon attribute and operator tests to LpNormalization operator 

anirudhacharya:
Issue - #1043

- add extra attribute
- Test cases.


anirudhacharya:
@houseroad I tried running these commands to generate the data and documentation for the changes -
```
python onnx/backend/test/cmd_tools.py generate-data
python onnx/defs/gen_doc.py
```
The `gen_doc` and `generate-data` tools errored with `ImportError: cannot import name FunctionProto` and 
```
File "onnx/backend/test/cmd_tools.py", line 31, in generate_data
    args.output, case.kind, case.name)
AttributeError: 'TestCase' object has no attribute 'kind'
```
respectively. 

The `stat-coverage` tool changed a lot of unrelated files like `bvlc_alexnet`, `densenet` etc.. Are these tools up-to-date and usable or is there something wrong in my usage?

bddppq:
@anirudhacharya tools are up-to-date, they are run in the CI for every PR and master commits. From the "missing FunctionProto error message" I guess you have invoked the python that installed the onnx release package, you need to instead install your local onnx checkout (that has the changes in the PR).

onnx/backend/test/data/real/test_inception_v2/data.json
```diff
@@ -1 +1 @@
-{"model_name": "inception_v2", "url": "https://s3.amazonaws.com/download.onnx/models/opset_8/inception_v2.tar.gz"}
\ No newline at end of file
+{"model_name": "inception_v2", "url": "https://s3.amazonaws.com/download.onnx/models/inception_v2.tar.gz"}
```
bddppq: Why changing this?

anirudhacharya: this change was made by the `stat-coverage` tool.

anirudhacharya:
the CI failed on this test `OnnxBackendNodeModelTest.test_mvn_cpu`. The data file for the above test is badly formed, but my PR does not touch this test.

houseroad:
Let’s don’t commit the mvn tests at this moment

bddppq:
@raymondxyang @linkerzhang for mvn

onnx/backend/test/case/node/lpnormalization.py
```diff
@@ -0,0 +1,122 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import math
```
bddppq: unused import

onnx/backend/test/case/node/lpnormalization.py
```diff
@@ -0,0 +1,122 @@
...
+import numpy as np  # type: ignore
+import onnx
+from ..base import Base
+from . import expect
+
+
+class LRN(Base):
```
anirudhacharya: wrong name. will correct

anirudhacharya:
the two `appveyor` tests are stuck

bddppq:
@anirudhacharya Not stuck, just the queue is super long ... @prasanthpul Maybe upgrade our plan again? :-)

houseroad:
may be from 2 instances to 4 instances :-)

onnx/defs/nn/defs.cc
```diff
@@ -1077,6 +1086,11 @@ ONNX_OPERATOR_SET_SCHEMA(
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to float tensors.")
         .SetDoc(LpNormalization_ver1_doc)
+        .Attr(
+            "epsilon",
+            "The epsilon value to use to avoid division by zero, default is 1e-10f.",
+            AttributeProto::FLOAT,
+            1e-10f)
```
houseroad: For the single floating number, 1e-10f is too small. Let's make it 1e-5.

onnx/backend/test/case/node/lpnormalization.py
```diff
@@ -0,0 +1,121 @@
...
+import onnx
+from ..base import Base
+from . import expect
+
+
+class LpNormalization(Base):
+
+    @staticmethod
+    def export_l2normalization_axis_0():  # type: () -> None
+        node = onnx.helper.make_node(
```
houseroad: We can have a variable called epsilon, since you use it twice in every cases.

onnx/defs/nn/defs.cc
anirudhacharya: Tensorflow uses 1e-12 - https://www.tensorflow.org/api_docs/python/tf/nn/l2_normalize 
and MXNet uses 1e-10 - https://mxnet.incubator.apache.org/api/python/symbol/symbol.html#mxnet.symbol.L2Normalization. 

IMO 1e-10 is a pretty reasonable default value to have. Let me know if you still want the default value changed.

houseroad: I checked with the floating converter, it is fine. Let's keep 1e-10.

anirudhacharya:
unrelated CI error - `ERROR at setup of onnx/examples/load_model.ipynb::Cell 0`. Have retriggered the CI.

docs/Operators.md
```diff
@@ -5144,6 +5144,15 @@ This version of the operator has been available since version 1 of the default O
 ### <a name="LpNormalization"></a><a name="lpnormalization">**LpNormalization**</a>
 
   Given a matrix, apply Lp-normalization along the provided axis.
+  
+  For p=2,
+  
+  y = x  / (sqrt(sum(x ** 2)) + epsilon),
```
ebarsoum: This might make it difficult for converter, due to how each framework add epsilon:

    TF: x / sqrt(max(sum(x**2), epsilon))
    Others: x / sqrt(sum(x**2) + epsilon)

houseroad: Any suggestion?

ebarsoum: I don't have a good solution. What Caffe2 and PyTorch use?

bddppq: caffe2 doesn't have epsilon, pytorch uses `max(sqrt(sum(x**2)), epsilon)`
yep yet another two variants :)

anirudhacharya: so what is the resolution for this?

anirudhacharya: @bddppq @ebarsoum ping?

vandanavk:
Is this PR good to go? Would be great if we could have this param and tests for LPNormalization.

Roshrini:
@houseroad @bddppq Due to 'eps' attribute not being present, users wont be able to use LPNormalization operator in mapping framework specific Normalization operators. 

Seems like [onnx-tf](https://github.com/onnx/onnx-tensorflow/blob/master/onnx_tf/handlers/backend/lp_normalization.py ) maps this operator to tf.norm which is same as reduceL1/reduceL2.. 

Pytorch maps norm to reducel1/reducel2 but does not export [normalize](https://pytorch.org/docs/master/nn.html#torch.nn.functional.normalize) operator due to the same restriction.

I see implementation to add eps attribute to LPNormalization in this PR but due to eps usage being different in different frameworks, its kind of stuck. Any idea how we can add this attribute or if we should just remove the operator as it cannot be used in any framework converters? Correct me if I am wrong.

docs/Operators.md
ebarsoum: If both TF and PT use max, and based on both equations we can easily convert between them. Then let's use max and not plus.

ebarsoum: Also, max doesn't add bias like the plus.

ebarsoum:
If this is still important, please bring a proposal to the OP SIG meeting.

Issue #1319, opened 2018-08-23
Test case for argmin is wrong with 32 bits python


snnn:
backend/test/case/node/argmin.py

If the python is 32 bits, then the output type of argmin_no_keepdims_random model is INT32, not INT64.

PR #1332, opened 2018-08-27
Arg tests output int64

bddppq:
@snnn #1319 

test data files are already correct though


bddppq:
@houseroad no doc needs to be updated

houseroad:
@bddppq right, the changes are not part of the expect functions.

snnn:
Thanks!

Issue #1313, opened 2018-08-22
Test Cases for FuncMeanVarianceNormalization Missing


Ac2zoom:
Test cases for FuncMeanVarianceNormalization are missing.  This means that, on a separate branch of ONNX, when test data is generated, relevant tests start failing (schema for FuncMeanVarianceNormalization can't be found).

linkerzhang:
MVC test cases were added.

PR #1335, opened 2018-08-27
Align mvn test with current schema

raymondxyang:
Fix for #1313. Will further modify this once the MVN schema is removed.


raymondxyang:
FYI I ignore some changes caused by generate-data script (parts I didnot touch..)

PR #1337, opened 2018-08-27
Add instructions for non root install from source

chinhuang007:
The install from source code is trying to create
files in /usr/local/bin where a non root user might
not have write permissions. The fix is to provide
instructions for non root users to avoid getting
'error: [Errno 13] Permission denied'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ git submodule update --init --recursive
 python setup.py install
 ```
 
+You can run the commands above in a virtual environment such as virtualenv if you are not logged in as root.
+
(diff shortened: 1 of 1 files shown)


chinhuang007:
@houseroad Is it okay to instruct non root users to use a virtual environment to avoid the permissions exception? I believe it is fairly common to run development builds in python virtual envs

chinhuang007:
The situation is that non-root user would get the exception when strictly following the current instructions, not a very good first impression.

chinhuang007:
I believe most developers log in as non-root users. Some python applications do not require adding files to /usr/local/bin. So a little reminder here should help the first time onnx user's experience.

PR #1338, opened 2018-08-28
added ne/le/ge/expand_as operators

### LINKED ISSUE #1333: Adding not equal operator and expand_as operator.
It seems like ONNX does not have operators for pytorch's expand_as and ne (not equal) operators.

Can I work on adding these operators to ONNX?

ArmenAg:
fixes #1333


houseroad:
Can you sign the CLA?

ArmenAg:
I have signed it.

houseroad:
It shows you have one more to sign :-)

ArmenAg:
Sorry git troubles :)

prasanthpul:
Agree with @bddppq.

prasanthpul:
This seems like a good use for functions (https://github.com/onnx/onnx/blob/master/docs/Functions.md). Functions are basically compound ops and enable optimization in backends that can but still allow others to run the models without implementing lots of ops.

ArmenAg:
In that case should things like sigmoid\tanh be functions and not operators?

snnn:
No test case?

ArmenAg:
Are the only tests I need to implement [here](https://github.com/onnx/onnx/tree/master/onnx/backend/test/case/node)? What's the consensus on which of these ops need to be added?

houseroad:
You also need to generate the data using these tests. `pythno onnx/backend/test/cmd_tools.py generate-data`

ArmenAg:
I wrote all the tests for ne/ge/le/expand_as.

onnx/defs/logical/defs.cc
```diff
@@ -121,6 +121,48 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "Constrains output to boolean tensor."));
+
+ONNX_OPERATOR_SET_SCHEMA(
+    LE,
+    8,
+    OpSchema()
+        .FillUsing(BinaryLogicDocGenerator("less than or equal"))
+        .TypeConstraint(
+            "T",
+            {"tensor(bool)", "tensor(int32)", "tensor(int64)"},
```
snnn: What's the semantic of ‘<=' on bool?

ArmenAg: This is a good point. I wanted to stay consistent with numpy.less_equal functionality which would return a tensor filled with True. But I see the argument as to not supporting boolean tensors for this op.

ArmenAg: What we could do is to maintain support <= on boolean tensors (for consistency reasons), but write an optimization pass that would replace this comparison with a Boolean Tensor of True values.

onnx/backend/test/case/node/ge.py
```diff
@@ -0,0 +1,41 @@
...
+from ..base import Base
+from . import expect
+
+
+class GE(Base):
+
+    @staticmethod
+    def export():  # type: () -> None
+        node = onnx.helper.make_node(
+            'LE',
```
zrphercule: I think this should be GE?

onnx/backend/test/case/node/ge.py
```diff
@@ -0,0 +1,41 @@
...
+        x = (np.random.randn(3, 4, 5) * 10).astype(np.int32)
+        y = (np.random.randn(3, 4, 5) * 10).astype(np.int32)
+        z = np.greater_equal(x, y)
+        expect(node, inputs=[x, y], outputs=[z],
+               name='test_ge')
+
+    @staticmethod
+    def export_equal_broadcast():  # type: () -> None
+        node = onnx.helper.make_node(
+            'NE',
```
zrphercule: And here?

onnx/backend/test/case/node/le.py
```diff
@@ -0,0 +1,41 @@
...
+        x = (np.random.randn(3, 4, 5) * 10).astype(np.int32)
+        y = (np.random.randn(3, 4, 5) * 10).astype(np.int32)
+        z = np.less_equal(x, y)
+        expect(node, inputs=[x, y], outputs=[z],
+               name='test_le')
+
+    @staticmethod
+    def export_equal_broadcast():  # type: () -> None
+        node = onnx.helper.make_node(
+            'NE',
```
zrphercule: Should be LE?

ArmenAg:
For consistency with numpy it makes sense. Logical operators should be defined for all numerical types.

ArmenAg:
@bddppq @snnn  @zrphercule What are your opinions on this PR? What are the next steps?

spandantiwari:
May I suggest that we reconsider the names for these three ops, `NE`, `LE`, `GE`. We currently have op names such as `Less`, `Greater`, `Equal`. It would be more consistent naming if we choose something like `NotEqual`, `LessEqual`, `GreaterEqual` for these new ops.

PR #1342, opened 2018-08-29
Fix handling of 0-element tensors in checker

jamesr66a:
Previously serialized models with 0-element tensors (e.g. Float[0, 1, 10]) would be rejected by the checker. This fixes the issue by checking for this condition explicitly, since there is not really a null state for the repeated data fields


onnx/checker.cc
```diff
@@ -95,7 +95,12 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& /*ctx*/) {
 
 #undef check_data_field
 
-  if (num_value_fields != 1) {
+  int32_t nelem = 1;
```
bddppq: int64_t

onnx/checker.cc
```diff
@@ -95,7 +95,12 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& /*ctx*/) {
...
+  for (auto x : tensor.dims()) {
+    nelem *= x;
+  }
+
+  if (nelem != 0 && num_value_fields != 1) {
```
bddppq: if nelem is 0 we should explicitly check num_value_fields is 0

onnx/checker.cc
houseroad: Good catch

PR #1348, opened 2018-08-29
Require exactly one node in a model passed to onnxGetBackendCompatibility

Maratyszcza:
Based on discussion with @rdzhabarov and @bddppq, there isn't a good use-case today for passing a graph with more than one node, so to simplify implementations it is better to add this constraint. If we find that we do need to pass whole subgraphs in the future, we can relax this constraint without breaking backward compatibility.

### DIFF
--- a/onnx/onnxifi.h
+++ b/onnx/onnxifi.h
@@ -1015,12 +1015,13 @@ ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
     size_t* infoValueSize);
 
 /**
- * Query if an ONNX model graph is compatible with the backend.
+ * Query if an ONNX node (operator) is compatible with the backend.
  *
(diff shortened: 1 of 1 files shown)


yinghai:
This might be an issue. One use case is we convert C2 op to ONNX node and run `onnxGetBackendCompatibility`. Sometimes one C2 op will map to multiple ONNX nodes. Then this rule will break things.

Maratyszcza:
@yinghai: in this case we can still check the subgraph nodes one-by-one, right?

Maratyszcza:
Albeit it does it create inconvenience on framework side.

yinghai:
That's multiple calls, which isn't really necessary. The rule that this PR proposes looks a bit strict to me.

rdzhabarov:
@Maratyszcza @yinghai I think it should not be a problem to supply multiple nodes, go for it.

PR #1358, opened 2018-08-30
PRelu, Sub, and LeakyRelu Version Converter Adapters

Ac2zoom:
Remaining Version Converter Adapters to fully cover ONNX Model Zoo (particularly TinyYoloV2 and ResNet100)

TODO: Need to add tests

### DIFF
--- a/onnx/version_converter/adapters/prelu_7_6.h
+++ b/onnx/version_converter/adapters/prelu_7_6.h
@@ -1,0 +1,28 @@
+// Adapter for PRelu in default domain from version 8 to 7
+
+#pragma once
+
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+class PRelu_7_6 final : public Adapter {
+  public:
+    explicit PRelu_7_6(): Adapter("PRelu", OpSetID(7), OpSetID(6)) {}
+
+    void adapt_prelu_7_6(std::shared_ptr<Graph> graph, Node* node) const {
+      // Throw an exception if any broadcasting occurs
+      const ArrayRef<Value*>& inputs = node->inputs();
+      std::vector<Dimension> X_sizes = inputs[0]->sizes();
+      std::vector<Dimension> slope_sizes = inputs[1]->sizes();
+      ONNX_ASSERTM(slope_sizes.size() == 1 || check_numpy_unibroadcastable_and_require_broadcast(X_sizes,
+            slope_sizes) == 0,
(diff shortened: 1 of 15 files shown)


onnx/version_converter/adapters/prelu_7_6.h
```diff
@@ -0,0 +1,28 @@
...
+
+class PRelu_7_6 final : public Adapter {
+  public:
+    explicit PRelu_7_6(): Adapter("PRelu", OpSetID(7), OpSetID(6)) {}
+
+    void adapt_prelu_7_6(std::shared_ptr<Graph> graph, Node* node) const {
+      // Throw an exception if any broadcasting occurs
+      const ArrayRef<Value*>& inputs = node->inputs();
+      std::vector<Dimension> X_sizes = inputs[0]->sizes();
+      std::vector<Dimension> slope_sizes = inputs[1]->sizes();
```
houseroad: It's possible, we don't have the shape information. If there is no enough information, also fail the conversion.

onnx/version_converter/convert.h
```diff
@@ -161,6 +162,25 @@ class DefaultVersionConverter : public BaseVersionConverter {
...
+      registerAdapter(make_unique<TypeRestriction>("Sub",
+        OpSetID(6), OpSetID(5), broadcast_unallowed_types));
+      registerAdapter(make_unique<RemoveConsumedInputs>("Sub",
+        OpSetID(5), OpSetID(6)));
+      registerAdapter(make_unique<CompatibleAdapter>("PRelu",
+        OpSetID(6), OpSetID(5)));
+      registerAdapter(make_unique<RemoveConsumedInputs>("PRelu",
+        OpSetID(5), OpSetID(6)));
+      registerAdapter(make_unique<CompatibleAdapter>("PRelu",
+        OpSetID(6), OpSetID(7)));
```
houseroad: We need to add a unsqueeze to make sure the second input is in the shape of [C, 1, 1, ...]

onnx/version_converter/helper.cc
```diff
@@ -16,7 +16,7 @@ namespace ONNX_NAMESPACE { namespace version_conversion {
             input2_sizes[i].dim != 1) return -1;
         if (input2_sizes[i].dim != input1_sizes[axis + i].dim) broadcast = true;
       }
-      // Return true if broadcasting is required
+      // Return 1 if broadcasting is required
```
rdzhabarov: curious why check_numpy_unibroadcastable_and_require_broadcast does not return `bool`

PR #1360, opened 2018-08-31
initialize dim with unknown dimension when doing shape inference.

linkerzhang:
In cases dimension can't be inferred (either with dim value or symbolic dim), an unknown dim should still be set/added into shape as a placeholder.


linkerzhang:
@houseroad  reformatted. Thanks!

linkerzhang:
@houseroad  this PR is doing the same fix as the fix you mentioned. Abandoning this PR in this way.

PR #1365, opened 2018-08-31
New tool to profile ops in model

Ac2zoom:


tools/profile_model_ops.py
```diff
@@ -0,0 +1,10 @@
+# A tool to profile the ops in ONNX models
+# Run by providing path to an ONNX model as the only CLI arg
+# A set of all unique ops present in the model will be printed
+
+import onnx
+import sys
+
```
houseroad: Probably have main function and other functions?

We can do more profile on the model, not only op_types, we can keep it extensible

tools/profile_model_ops.py
```diff
@@ -0,0 +1,10 @@
...
+if len(sys.argv) == 2:
```
bddppq: use argparse :-)

tools/profile_model_ops.py
```diff
@@ -0,0 +1,19 @@
...
+import sys
+import argparse
+
+
+def uniqueOps(model):
+    print(set([n.op_type for n in model.graph.node]))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Calculate stats on ONNX models')
```
houseroad: add an arg, like --unique-op?

bddppq:
```
+python setup.py typecheck
running typecheck
tools/profile_model_ops.py:10: error: Function is missing a type annotation
```

Need to add type annotations

xadupre:
Do you have other features in mind for this tool?

tools/profile_model_ops.py
```diff
@@ -0,0 +1,22 @@
...
+# Run by providing path to an ONNX model as the only CLI arg
+# A set of all unique ops present in the model will be printed
+
+import onnx
+import sys
+import argparse
+
+
+def uniqueOps(model):  # type: (onnx.onnx_pb2.ModelProto) -> None
+    print(set([n.op_type for n in model.graph.node]))
```
xadupre, line 11: Recursivity? (control flow) What about functions?

PR #1377, opened 2018-09-05
Add DynamicSlice experimental op

jamesr66a:
Being able to slice tensors with boundaries only available at runtime is crucial for many sequence processing models. This PR adds an experimental operator DynamicSlice, which is equivalent to the existing Slice operator, but the `starts`, `ends` and `axes` are specified via inputs, not attributes

A sample implementation (PyTorch -> ONNX -> Caffe2) can be found here: https://github.com/pytorch/pytorch/pull/11255


onnx/defs/experiments/defs.cc
```diff
@@ -421,4 +421,60 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  starts = [0, 1]
+  ends = [-1, 1000]
+  result = [
+      [2, 3, 4],
+  ]
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    DynamicSlice,
+    1,
```
bddppq: New operator needs to be put into new operator set. Since we just released opset 8 with onnx v1.3 last week, you need to bump the opset to 9.

onnx/defs/experiments/defs.cc
```diff
@@ -421,4 +421,60 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+Produces a slice of the input tensor along multiple axes. Similar to numpy:
+https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html
+Slices uses `axes`, `starts` and `ends` inputs to specify the start and end
+dimension for each axis in the list of axes, it uses this information to
+slice the input `data` tensor. If a negative value is passed for any of the
+start or end indices, it represent number of elements before the end of that
+dimension. If the value passed to start or end is larger than the `n` (the
+number of elements in this dimension), it represents `n`. For slicing to the
+end of a dimension with unknown size, it is recommended to pass in `INT_MAX`.
+If `axes` are omitted, they are set to `[0, ..., ndim-1]`.
```
bddppq: Is axes optional input? If so you need to annotate the corresponding Input in the schema

docs/Changelog.md
```diff
@@ -990,6 +990,43 @@ This version of the operator has been available since version 1 of the default O
 <dd>Constrain input and output types to float tensors.</dd>
 </dl>
 
+### <a name="DynamicSlice-1"></a>**DynamicSlice-1**</a>
```
ebarsoum: When we decide to remove it from experimental, it should replace slice, right?

jamesr66a: Yes, I think that makes sense, unless people want the specialized version with the slices as attributes for optimization

onnx/defs/operator_sets.h
```diff
@@ -146,6 +147,7 @@ class OpSet_Onnx_ver1 {
            Onnx, 1, DepthToSpace)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Div)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Dropout)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, DynamicSlice)>());
```
bddppq: This should be put in OpSet_Onnx_ver9:: ForEachSchema (you need to create the new OpSet_Onnx_ver9 since this is the first change in opset 9).
Mixing schemas in different opset versions will cause issue when __ONNX_DISABLE_STATIC_REGISTRATION is false.

onnx/defs/operator_sets.h
```diff
@@ -456,7 +454,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 8, Mean);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 8, MaxPool);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 8, Scan);
 
-// Iterate over schema from ai.onnx version 7
+// Iterate over schema from ai.onnx version 8
```
bddppq: lol good catch

PR #1381, opened 2018-09-05
WIP: statCoverage(model) Fixes

Ac2zoom:
Abstracted _prepare_model_data from Runner, fix 0 node model case, remove ml check for model_written

### DIFF
--- a/onnx/backend/test/stat_coverage.py
+++ b/onnx/backend/test/stat_coverage.py
@@ -205,10 +205,11 @@ def gen_model_test_coverage(schemas, f, ml):
                     .format(op, len(attrs[op].keys()), len(schema_dict[op]
                         .attributes)))
             for attribute in sorted(schema_dict[op].attributes):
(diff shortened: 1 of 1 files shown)


onnx/backend/test/stat_coverage.py
```diff
@@ -205,10 +205,11 @@ def gen_model_test_coverage(schemas, f, ml):
                     .format(op, len(attrs[op].keys()), len(schema_dict[op]
                         .attributes)))
             for attribute in sorted(schema_dict[op].attributes):
+                f.write('<summary>{}</summary>\n<details>'.format(attribute))
                 if attribute in attrs[op]:
-                    f.write('{}: {}\n'.format(attribute, len(attrs[op][attribute])))
-                else:
-                    f.write('{}: 0\n'.format(attribute))
+                    for value in attrs[op][attribute]:
```
zrphercule: What if attribute is not in attrs[op]?

Issue #1364, opened 2018-08-31
"-Wl,--exclude-libs,ALL" should be used when linking onnx_cpp2py_export


snnn:
Symbols in onnx_cpp2py_export.cpython-36m-x86_64-linux-gnu.so:

  Before:  6222
  After:   347 

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index dce1d98..6beb053 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -344,7 +344,7 @@ if(BUILD_ONNX_PYTHON)
     # Assume everything else is like gcc
     target_link_libraries(onnx_cpp2py_export
                           PRIVATE "-Wl,--whole-archive" $<TARGET_FILE:onnx>
-                                  "-Wl,--no-whole-archive")
+                                  "-Wl,--no-whole-archive" "-Wl,--exclude-libs,ALL")
   endif()
```

PR #1388, opened 2018-09-07
Do not export onnx symbols in the python extension

bddppq:
~~Also fixed the issue of duplicating libonnx in the link command of onnx_cpp2py_export~~

Results of `nm -g .setuptools-cmake-build/onnx_cpp2py_export.so | wc -l`:
Before: 2677
After: 396

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -345,6 +345,8 @@ if(BUILD_ONNX_PYTHON)
     target_link_libraries(onnx_cpp2py_export
                           PRIVATE "-Wl,--whole-archive" $<TARGET_FILE:onnx>
                                   "-Wl,--no-whole-archive")
+    set_target_properties(onnx_cpp2py_export
+                          PROPERTIES LINK_FLAGS "-Wl,--exclude-libs,ALL")
   endif()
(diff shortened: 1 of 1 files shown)


bddppq:
#1364

houseroad:
@zrphercule it's generated from `onnx-operators.proto`

zrphercule:
@houseroad I see... So I think this error is because onnx-operators.pb.h is a part of lib "onnx"?

bddppq:
@houseroad @zrphercule I bet it's cmake treats
```
target_link_libraries(onnx_cpp2py_export
                          PRIVATE "-Wl,--whole-archive" $<TARGET_FILE:onnx>
                                  "-Wl,--no-whole-archive")
```
differently than
```
target_link_libraries(onnx_cpp2py_export PRIVATE onnx)
```
(I have removed the latter in my previous diff) that with the former it doesn't treat onnx as a dependency of onnx_cpp2py_export and so onnx's public includes (and probably also the defines) are not propagated to onnx_cpp2py_export.

Let me remove that cleanup from this PR to fix #1364 first.

PR #1390, opened 2018-09-07
Supporting int32 and int64 in Less and Greater

houseroad:
More like a spec fix instead of adding new features. So prefer not bumping the opset version for these two ops (really no need).

### DIFF
--- a/onnx/defs/logical/defs.cc
+++ b/onnx/defs/logical/defs.cc
@@ -81,16 +81,12 @@ ONNX_OPERATOR_SET_SCHEMA(
 
 ONNX_OPERATOR_SET_SCHEMA(
     Greater,
-    7,
+    9,
     OpSchema()
         .FillUsing(BinaryLogicDocGenerator("greater"))
         .TypeConstraint(
             "T",
-            {"tensor(int32)",
-             "tensor(int64)",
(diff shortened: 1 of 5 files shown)


spandantiwari:
Maybe we should consider bumping the opset version for those runtimes that support ONNX 1.2 and 1.3 that did not not have the integer types as part of the implementation of these ops.

houseroad:
@spandantiwari ok, let me update the pr later :-)

onnx/defs/logical/defs.cc
```diff
@@ -86,7 +86,11 @@ ONNX_OPERATOR_SET_SCHEMA(
         .FillUsing(BinaryLogicDocGenerator("greater"))
         .TypeConstraint(
             "T",
-            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            {"tensor(int32)",
```
linkerzhang: thank you Lu! shall we add all numeric types please?

houseroad: Sure, I will do that and bump up the opset version too

houseroad:
@linkerzhang @spandantiwari comments have been addressed, ping for review :-)

PR #1396, opened 2018-09-07
Change onnxifi test driver classname

zrphercule:
Change the name of some classes in onnxifi test driver for better description, also added some documents.

### DIFF
--- a/onnx/backend/test/cpp/driver/test_driver.h
+++ b/onnx/backend/test/cpp/driver/test_driver.h
@@ -15,11 +15,11 @@ namespace testing {
  *	No real data was loaded in this type of class,
  *	but only data location.
  */
-struct TestData {
+struct UnsolvedTestData {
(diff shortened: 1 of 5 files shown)


onnx/backend/test/cpp/driver/test_driver.h
```diff
@@ -32,28 +32,28 @@ struct TestData {
...
+  std::vector<UnsolvedTestData> test_data_;
 };
 
 /**
  *	Store one chunk of test data,
  *	including raw input/output and protos
  *	Real data was loaded in this type of class..
  */
-struct ProtoTestData {
+struct ResolvedTestData {
```
rdzhabarov: can you comment what Unsolved/Resolved actually mean here :) in the comment for the classes?

zrphercule: Sure!

PR #1401, opened 2018-09-10
Enable onnxifi_test_driver_gtests in CI 

zrphercule:
Using onnxifi dummy backend.


cmake/unittest.cmake
```diff
@@ -62,6 +62,10 @@ function(AddTest)
     target_link_libraries(${_UT_TARGET} ${PROTOBUF_LIBRARIES})
   endif()
 
+  if(ONNX_DUMMY_BACKEND)
+	  target_compile_definitions(${_UT_TARGET} PUBLIC "ONNX_DUMMY_BACKEND=true")
```
rdzhabarov: Don't you need "-DONNX_DUMMY_BACKEND"? (but i'm not 100% certain)

zrphercule: Actually whether -D or not it doesnt works...

zrphercule:
@rdzhabarov oh I forgot telling you I updated the doc of "resolved" "unsolved" here =)

PR #1405, opened 2018-09-11
Adding generator ops Ones, Zeros, OnesLike, ZerosLike

spandantiwari:
These are generator ops for producing all-zero and all-ones tensors, that are useful for models. 
* `Ones` and `Zeros` take a shape and data type as attribute and produce a tensor.
* `OnesLike` and `ZerosLike` take an input tensor as an example and produce a tensor of the same shape and type (unless datatype is explicitly specified as attribute).


spandantiwari:
I see you point. Since we have other generator ops such as `RandomNormal` and `RandomNormalLike`, which could also be replaced with `Constant`, but are still there for convenience and clarity, I went with a design similar to those on this. I think this design lends more clarity, but I am OK with something like `ConstantLike` as well. 

It will be great to get some more feedback on this. What do others reviewers think?

spandantiwari:
I have created another PR https://github.com/onnx/onnx/pull/1406 for a different design using `ConstantLike` op. Please review that design as well. If that one is preferable, we can abandon this PR. 

Also, as noted in PR https://github.com/onnx/onnx/pull/1406, there's a experimental op `ConstantFill` which is along similar lines. The new PR PR https://github.com/onnx/onnx/pull/1406 is a streamlined version with more datatype support.

spandantiwari:
Closing this one out because the other design (https://github.com/onnx/onnx/pull/1406) is now merged.

spandantiwari:
Reviving this

spandantiwari:
@houseroad - While `ConstantLike` (https://github.com/onnx/onnx/pull/1406) addressed the requirement for `ZerosLike` and `OnesLike` from this PR, it does not address the need for `EyeLike` (since it is not a constant value). Also, existing `Constant` op is not enough because it requires a tensor of known shape and does not allow for dynamic shape during runtime. I will open a different PR to address requirements for EyeLike.

PR #1406, opened 2018-09-11
Adding generator op ConstantLike

spandantiwari:
This is a generator op for producing tensors with constant values (e.g. ones, zeros), that are useful for models. The shape and datatype of the generated tensor can be specified through an input tensor or through attributes.

Please note that there is a similar experimental op `ConstantFill`. This op is a somewhat streamlined version with more datatype support.


spandantiwari:
Not sure what the failure is. @houseroad could you please point to the issue? Is it related to test coverage?

houseroad:
Hi @spandantiwari, you probably need to run `python onnx/backend/test/stat_coverage.py` to update the test coverage page.

spandantiwari:
Hi @houseroad - Thanks for the pointer. I tried running `stat_coverage.py`, but I see a `urllib.error.HTTPError: HTTP Error 403: Forbidden` error. Do I need some permission to download the models? I see that it is trying to access `*opset_9` models, which are not listed on onnx/models repo.

spandantiwari:
@houseroad - still not getting stat_coverage to run. 
Also, ping for review...

houseroad:
@spandantiwari sorry, the models haven't been updated yet. I will do it probably tomorrow. Could you use opset_8 models to generate the results?

onnx/defs/operator_sets.h
```diff
@@ -23,6 +23,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Clip);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Concat);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Constant);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, ConstantFill);
+class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, ConstantLike);
```
houseroad: we should add it as 9, because it is just added, right?

onnx/defs/generator/defs.cc
```diff
@@ -34,6 +34,91 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+argument. If 'dtype' is not specified, then the type of input tensor is used. If input 
+tensor is also not specified, then the type defaults to 'float'.
+
+The 'dtype' argument must be one of the data types specified in the 'DataType' enum field in the
+TensorProto message and be valid as an output type.
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    ConstantLike,
+    1,
```
houseroad: Same here: opset version 9?

spandantiwari:
@houseroad - thanks for the feedback. I have updated the PR based on your comments. I was also able to use  opset_8 models to generate test coverage results so all the checks are now passing. Could you please review. Thanks.

spandantiwari:
@houseroad - I have updated the branch to remove some conflicts with master (maybe from your latest Less, Greater check-in). All checks are passing. Could you please review and see if this can be approved.

onnx/backend/test/case/node/constantlike.py
```diff
@@ -0,0 +1,53 @@
...
+    @staticmethod
+    def export_threes_with_shape_and_dtype():  # type: () -> None
+        shape = (3, 4)
+        node = onnx.helper.make_node(
+            'ConstantLike',
+            shape=shape,
+            inputs=[],
+            outputs=['y'],
+            value=3.0,
+            dtype=11,  # 11: DOUBLE
```
houseroad: Can you directly refer to the Enum, like `https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/constant.py#L24`?

spandantiwari: Good point. Done.

spandantiwari:
@houseroad Addressed the issue in your latest feedback.

onnx/defs/generator/defs.cc
```diff
@@ -34,6 +34,91 @@ ONNX_OPERATOR_SET_SCHEMA(
           updateOutputShape(ctx, 0, tensor_proto);
         }));
 
+static const char* ConstantLike_ver1_doc = R"DOC(
```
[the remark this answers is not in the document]
spandantiwari: Fixed.

spandantiwari:
@houseroad - Addressed your comments.

spandantiwari:
@houseroad - Thanks.

PR #1407, opened 2018-09-12
[build] Add ONNX_API for protos in all cases

orionr:
This is needed as a part of https://github.com/pytorch/pytorch/pull/11385/

cc @houseroad @mingzhe09088

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,11 +118,8 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP NAME SRCS HDRS ROOT_DIR DEPEND)
     return()
   endif()
 
-  if(MSVC AND BUILD_SHARED_LIBS)
-    set(ONNX_DLLEXPORT_STR "dllexport_decl=ONNX_API:")
-  else()
-    set(ONNX_DLLEXPORT_STR "")
-  endif()
+  # Add ONNX_API prefix to protobuf classes and methods in all cases
+  set(ONNX_DLLEXPORT_STR "dllexport_decl=ONNX_API:")
 
   set(${SRCS})
   set(${HDRS})


houseroad:
I think you need to remove the condition here: https://github.com/onnx/onnx/blob/6bedd27b0307c9295039bd847895a27275160a98/onnx/onnx_pb.h#L54 to define ONNX_API in all cases.

snnn:
Please wait. 

Please add some document of who will use this macro, at where, for which reason?

Who will build onnx as a dll on Windows?
If that is supported, will that dll statically link to protobuf or dynamically?
Which symbols will be exported?

Generally speaking, it's not good to use C++ types in DLL interface. It means you shouldn't build protobuf library as a DLL.  If you really want to do this, please add some documents, to let users know the possible issues.

orionr:
Hi @snnn. Happy to talk more here before landing.

ONNX won't actually be built as a DLL in our case, but rather included in another DLL where we want the ONNX interface to be exported. Note that this is really just making the ONNX_API functional where it wasn't before. Per the code we always wanted to support this case, but let me know if that's wrong.

Also FYI these changes simply match what we've done on the PyTorch / Caffe2 side.

houseroad:
@onnxbot retest this please

orionr:
@snnn I just improved the docstring around ONNX_API. Let me know your thoughts and thanks.

CMakeLists.txt
```diff
@@ -274,6 +271,16 @@ target_include_directories(onnx_proto PUBLIC
...
+if (MSVC)
+  if (BUILD_SHARED_LIBS OR ONNX_BUILD_MAIN_LIB)
+    set(ONNX_API_DEFINE "-DONNX_API=__declspec(dllexport)")
+  else()
+    set(ONNX_API_DEFINE "-DONNX_API=")
+  endif()
+else()
+  set(ONNX_API_DEFINE "-DONNX_API=")
+endif()
+target_compile_definitions(onnx_proto PRIVATE ${ONNX_API_DEFINE})
```
yinghai: Setting this to `PRIVATE` will cause compilation issue for other code which used ONNX and include `onnx.pb.h`.

PR #1408, opened 2018-09-12
Add opaque type support

yuslepukhin:
Cc: @gramalingam, @linkerzhang


yuslepukhin:
Version check naturally fails. Will need to learn the due process.
```
 assert reference_output == test_output failed:

  'The ir_versi...is checked!\n' == 'The ir_versio...is checked!\n'
  Skipping 778 identical trailing characters in diff, use -v to show
  - The ir_version in model: 3
  ?                          ^
  + The ir_version in model: 4
  ?                          ^
```

yuslepukhin:
Seems the build errors out on Java exception

onnx/defs/data_type_utils.cc
```diff
@@ -119,25 +119,48 @@ std::string DataTypeUtils::ToString(
...
+      std::string result;
+      std::string param_str;
+      const auto& op_type = type_proto.opaque_type();
+      const auto param_size = op_type.parameters_size();
+      for (int p = 0; p < param_size; ++p) {
+        const auto& param_proto = op_type.parameters(p);
+        param_str.append(
+            ToString(param_proto, (p == 0) ? empty : ",", empty));
+      }
+      result.append("opq(");
```
gramalingam: Just some minor suggestions: why not "opaque" instead of "opq"? Also, not sure if we really need the designations "d()" and "n()" around the domain and name … we can treat them positionally … except I guess that they can be empty …

yuanbyu: Yes "opaque" is better.

yuslepukhin: I do not mind. Would `sequence` instead of `seq` be a better choice then?

yuslepukhin:
@houseroad The Opaque type is meant to serve as placeholder for types defined by ML clients. Our runtime provides C++ definitions for Tensors, Sequences and Maps associated with the corresponding C++ implementation. However, users may want/need to define their own types that are recursively composed of other  other onnx/ml types. To that end we provide an Opaque type with name/domain serving as discriminators and an array of TypeProto params that describe the composition.

gramalingam:
@houseroad , to expand on Dmitri's explanation: opaque types are meant to be an extensibility mechanism, to introduce new types into ONNX. In particular, it can be used with custom ops, as a way to introduce custom types. (It can also be used to introduce experimental types, which we can promote later on as standard types within ONNX if it is considered useful enough.)

onnx/onnx-ml.proto
```diff
@@ -435,6 +435,15 @@ message TypeProto {
     optional TypeProto value_type = 2;
   };
 
+  message Opaque {
+   // When missing, the domain is the same as the model's.
+    optional string domain = 1;
+  // The name is optional but significant when provided.
+    optional string name = 2;
+  // parameters that help defining the type
+    repeated TypeProto parameters = 3;
```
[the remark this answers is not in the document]
gramalingam: Parametrized types are quite common (in fact, all the existing types, Sequences, Maps, and Tensors are parametrized types). So, allowing parameters adds flexibility. In the special case when the list is empty, we get unparametrized types. If we decide to introduce "sparse tensor" as an experimental type, or a "quantized tensor" as an experimental type (I realize that there is ongoing discussion to introduce a standardized type, but just giving examples here), they will be parametric types … if we want to introduce "Tuple" (or "Pair") as a type constructor, they will require type-parameters.

onnx/defs/data_type_utils.cc
linkerzhang: keep "sep".

onnx/onnx-ml.proto
```diff
@@ -452,6 +461,8 @@ message TypeProto {
     // The type of a map.
     Map map_type = 5;
 
+    Opaque opaque_type = 100;
```
linkerzhang: the id should be 6 here (given it's being checked in onnx repo now).

yuslepukhin: @linkerzhang denotation string is already 6 so we have to put 7

yuslepukhin:
@houseroad Checked formatting and it takes the .clang-format in the repo. Could not find any abnormalities.

onnx/onnx.in.proto
```diff
@@ -433,6 +433,15 @@ message TypeProto {
     optional TypeProto value_type = 2;
   };
 
+  message Opaque {
+   // When missing, the domain is the same as the model's.
+    optional string domain = 1;
+  // The name is optional but significant when provided.
```
houseroad: For example here, indent is not right. I guess clang-format won't check proto files.

yuslepukhin: Will do.

PR #1409, opened 2018-09-12
Update split pass to not move control operators

jamesr66a:


jamesr66a:
@onnxbot retest this please

PR #1410, opened 2018-09-13
add default value to doc

fumihwh:


fumihwh:
@houseroad 
Maybe remove the existing description of default value is a good idea to avoid duplicating?

houseroad:
@fumihwh that will be great, could you do that?

PR #1418, opened 2018-09-15
Add version table to Versioning.md to provide a clear mapping

fdwr:
ONNX has three comingled but independent versioning sequences:
- the ONNX version (e.g. 1.2)
- file format version (e.g. ir_version 3)
- op set versions (e.g. op set 7).

This includes a table to make the various relationships clearer.


docs/Versioning.md
```diff
@@ -140,3 +140,12 @@ ISSUE: what's our policy when a model takes a dependency on new `IR_VERSION` cha
 ### Accuracy or performance changes
 
 Assuming that there are no breaking changes to the signature of the model's graph or any operator dependencies, the shape and contents of the graph can change freely provided there are no semantic changes to the model. However, changes to the shape and contents of the graph can impact model accuracy and/or model performance.
+
+## Released Versions
+
+ONNX version|File format version|Operator set version ai.onnx|Operator set version ai.onnx.ml
+------------|-------------------|----------------------------|-------------------------------
+1.0|3|1|1
```
houseroad: Here do we only cover the major version or we also want to cover the minor version? For example, in 1.1.2, opset version is 6.

fdwr: @houseroad Certainly - I let this PR stall. Updating...

docs/Versioning.md
```diff
@@ -1,142 +1,152 @@
...
+ISSUE: what's our policy when a model takes a dependency on new `IR_VERSION` change and/or new operator change?
+
+
+### Accuracy or performance changes
+
+Assuming that there are no breaking changes to the signature of the model's graph or any operator dependencies, the shape and contents of the graph can change freely provided there are no semantic changes to the model. However, changes to the shape and contents of the graph can impact model accuracy and/or model performance.
+
+## Released Versions
+
+ONNX version|File format version|Operator set version ai.onnx|Operator set version ai.onnx.ml
```
smk2007: Can we keep this IR Version unless we update it everywhere?

fdwr: @smk2007 : What do you mean? This PR is just documenting the relationship to existing versions, not changing the file format version.

PR #1421, opened 2018-09-18
Add BFLOAT16 data type (FLOAT32 truncated to 16 bits)

Maratyszcza:
BFLOAT16 is a 16-bit floating-point data type equivalent to truncated IEEE single-precision floating-point format. Although non-standard, BFLOAT16 is getting traction in hardware, e.g.

> We also will support bfloat16, a numerical format being adopted industrywide for neural networks, in the Intel Nervana NNP-L1000. Over time, Intel will be extending bfloat16 support across our AI product lines, including Intel Xeon processors and Intel FPGAs.

https://newsroom.intel.com/editorials/artificial-intelligence-requires-holistic-approach/

This patch makes it possible to represent BFLOAT16 tensors and operators in ONNX, but doesn't yet update existing operators to support BFLOAT16 type.

cc @rdzhabarov and @nadavrot


onnx/onnx.in.proto
```diff
@@ -296,6 +296,10 @@ message TensorProto {
     UINT64 = 13;
     COMPLEX64 = 14;     // complex with float32 real and imaginary components
     COMPLEX128 = 15;    // complex with float64 real and imaginary components
+
+    // IEEE single-precision floating-point number truncated to 16 bits.
+    // This format has 1 sign bit, 8 exponent bits, and 7 mantissa bits.
+    BFLOAT16 = 16;
```
gramalingam: It would be helpful to add a comment/description to FLOAT16 also to clarify the distinction between that and BFLOAT16.

onnx/mapping.py
```diff
@@ -36,6 +36,7 @@
     TensorProto.INT64: TensorProto.INT64,
     TensorProto.BOOL: TensorProto.INT32,
     TensorProto.FLOAT16: TensorProto.UINT16,
+    TensorProto.BFLOAT16: TensorProto.UINT16,
```
bddppq: Actually now this whole `TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE` mapping seems unnecessary to me and I don't remember why I added it :-). I will remove this map in a follow up PR.

Maratyszcza: I thought an initial type is mapped to storage type, which is then mapped to name of the field

bddppq: yeah but IIRC we never really used storage type directly (it's always followed by using it to look up in the STORAGE_TENSOR_TYPE_TO_FIELD map).

Maratyszcza:
@prasanthpul, @linkerzhang, @yuanbyu: is this PR good with you?

jspisak:
@prasanth and I connected. They are reviewing.

jspisak:
@lupesko - Can you guys take a look and approve if you are good?

bddppq:
ping @gramalingam @ebarsoum @linkerzhang @pranavsharma @lupesko

linkerzhang:
LGTM, thank you for adding this! We may also check which operators need to add this type support.

jgong5:
@bddppq Do you have plan to add bfloat16 type support to operators?

PR #1423, opened 2018-09-19
Adding checker/typeshape inference logic for Function

raymondxyang:
1. Adding checker/typeshape inference logic for Function
2. Remove experimental mvn op, promote mvn function in opset 9
3. Regenerate test case to mvn


onnx/checker.cc
```diff
@@ -485,6 +501,22 @@ void check_model(const ModelProto& model) {
...
+    const CheckerContext& ctx,
+    const LexicalScopeContext& lex_ctx) {
+  // Create a temproary graphproto to hold the expanded subgraph
+  GraphProto g;
+  // To Generate unique internal tensor names
+  // while preserving node's input/output names
+  FunctionExpandHelper(node, func, g);
+  for (const NodeProto& n : g.node()) {
+    // Check the nodes in subgraph
+    check_node(n, ctx, lex_ctx);
```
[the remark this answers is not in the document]
raymondxyang: I considered the possibility of changing function into a graph. But we are missing value_info here.

onnx/shape_inference/implementation.cc
```diff
@@ -161,5 +179,98 @@ void InferShapes(
...
+      if (attr.has_ref_attr_name()) {
+        if (attr_map.count(attr.ref_attr_name())) {
+          copy_n.add_attribute()->CopyFrom(*attr_map[attr.ref_attr_name()]);
+        }
+      } else {
+        copy_n.add_attribute()->CopyFrom(attr);
+      }
+    }
+
+    InferenceContextImpl temp_ctx(
```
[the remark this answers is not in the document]
raymondxyang: We actually need to have a new context for every node. It will pick out the existing shape/type info from the temp_valuetypesbyname in its constructor.

onnx/checker.cc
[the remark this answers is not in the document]
raymondxyang: I ll try constructing a new context object here.

onnx/shape_inference/implementation.cc
```diff
@@ -161,5 +179,98 @@ void InferShapes(
...
+      return;
+    }
+    NodeProto copy_n(n);
+    // Add attribute information into the temproary node
+    copy_n.clear_attribute();
+    for (auto attr : n.attribute()) {
+      if (attr.has_ref_attr_name()) {
+        if (attr_map.count(attr.ref_attr_name())) {
+          copy_n.add_attribute()->CopyFrom(*attr_map[attr.ref_attr_name()]);
+        }
```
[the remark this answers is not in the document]
raymondxyang: I thought about it. We should not throw an error seems there can be default attribute. ie for mvn, if no attribute is passed into the function node, thus no attribute should be passed to reduced_mean op, and it has its own default values for the attributes

docs/FunctionsChangelog.md
```diff
@@ -3,14 +3,14 @@
             [def files](/onnx/defs) via [this script](/onnx/defs/gen_doc.py).
             Do not modify directly and instead edit function definitions.*
 # ai.onnx (default)
-## Version 8 of domain ai.onnx (default)
-### <a name="MeanVarianceNormalization-8"></a>**MeanVarianceNormalization-8**</a>
+## Version 9 of domain ai.onnx (default)
```
houseroad: Can we also run the doc comparison in the CI to make sure the doc is up to date?

raymondxyang: We run the doc_gen and git diff in CI but I think the default domain is not checked (we set to build onnx_ml and thus only -ml files are generated). 
I plan to work on a long-pending PR (which is to merge function and operator.md) files after this is merged. I can enable the script to build default docs no matter onnx_ml is set.

linkerzhang: @houseroad , good point, however, this doc (Function.md) will be merged into Operator.md, so let's keep it as it is right now. sounds good?

onnx/checker.cc
```diff
@@ -485,6 +501,20 @@ void check_model(const ModelProto& model) {
...
+void VerifyFunctionNode(
+    const NodeProto& node,
+    const FunctionProto& func,
+    const CheckerContext& ctx,
+    const LexicalScopeContext& lex_ctx) {
+  // Create a temporary graphproto to hold the expanded subgraph
+  GraphProto g;
+  g.set_name("func_" + func.name() + "_expanded_subgraph");
+  // To Generate unique internal tensor names
+  // while preserving node's input/output names
```
houseroad: Before expand to graph, we should check whether the node's inputs and outputs are the same as defined in the function?

docs/FunctionsChangelog.md
houseroad: @raymondxyang @linkerzhang that sounds good to me, thanks.

onnx/checker.cc
raymondxyang: The function does not have any class defined with io type constraints. It is only a proto class. So here what I did is to expand the function into a sub-graph while preserving the tensor names in input and output list. Then the actual nodes takes the tensors (partially or wholly) will do the checking job

onnx/defs/function.cc
```diff
@@ -100,8 +101,108 @@ Status FunctionBuilderRegistry::GetFunctions(
...
+  }
+  std::string node_name =
+      node.has_name() ? node.name() : func.name() + uniq_prefix;
+  int version = (int)func.since_version();
+  std::unordered_map<std::string, std::string> input_names_map;
+  std::unordered_map<std::string, std::string> output_names_map;
+  std::unordered_map<std::string, AttributeProto> attr_map;
+
+  for (int idx = 0; idx < node.input_size(); ++idx) {
+    input_names_map[func.input().Get(idx)] = node.input().Get(idx);
```
houseroad: Here, you access the func inputs without checking whether it is out of boundary. I mean we should check the inputs and outputs of the node match the spec of the function.

onnx/shape_inference/implementation.cc
```diff
@@ -161,5 +179,98 @@ void InferShapes(
...
 }
 
+void InferShapeForFunctionNode(
+    const FunctionProto& func,
+    const ISchemaRegistry* schema_registry,
+    InferenceContext& ctx) {
+  int domain_version = (int)func.since_version();
+  GraphProto g;
+  // Get a temproary tensor-shape map
+  std::unordered_map<std::string, TypeProto*> temp_valueTypesByName;
```
houseroad: Can we add another function called `InferShape(GraphProto g)`, so you just need to expand the node and do the shape inference?

houseroad: Our API is only for model right now, should be easy to add one for graph: https://github.com/onnx/onnx/blob/master/onnx/shape_inference/implementation.h#L91

raymondxyang: Sure will do so

linkerzhang:
@houseroad any more comment please? thanks!

PR #1426, opened 2018-09-19
Add node name for error & Fix typo

pengwa:


houseroad:
I have updated the doc here. `Operators.md` cannot be changed directly. Changing it is a little bit complicated. So I created https://github.com/onnx/onnx/pull/1427 to fix it.

PR #1428, opened 2018-09-19
Adding EyeLike generator op.

spandantiwari:
This is a generator ops for producing identity-like matrices with ones on the major diagonal and zeros everywhere else. This generator op is quite useful for models and most major frameworks have it. 

Note that  the input is restricted to 2D because that is the most dominant use case and most of the frameworks have support restricted to 2D for this (similar) op.


onnx/defs/generator/defs.cc
```diff
@@ -119,6 +119,81 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+             "tensor(uint8)",
+             "tensor(uint16)",
+             "tensor(uint32)",
+             "tensor(uint64)",
+             "tensor(bool)"},
+             "Constrain output types. Strings and complex are not supported.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          if (ctx.getAttribute("dtype") != nullptr)
+            propagateElemTypeFromAttributeToOutput(ctx, "dtype", 0);
+          else
```
houseroad: Nit: add {} even if there is only one stmt.

onnx/defs/generator/defs.cc
```diff
@@ -119,6 +119,81 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+The 'dtype' argument must be one of the data types specified in the 'DataType' enum field in the
+TensorProto message and be valid as an output type.
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    EyeLike,
+    9,
+    OpSchema()
+        .SetDoc(EyeLike_ver9_doc)
+        .Attr(
```
houseroad: Do you also want to add `k` parameter to indicate the index of the diagonal: 
0 (the default) refers to the main diagonal, a positive value refers to an upper diagonal, and a negative value to a lower diagonal.

onnx/defs/generator/defs.cc
```diff
@@ -119,6 +119,81 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "dtype",
+            "(Optional) The data type for the elements of the output tensor. If not specified,"
+            "the data type of the input tensor T1 is used. If input tensor T1 is also not" 
+            "specified, then type defaults to 'float'.",
+            AttributeProto::INT,
+            OPTIONAL)
+        .Input(
+            0,
+            "input",
```
anirudhacharya: why not have number of rows and columns as parameters rather than infer it from an input tensor? numpy, pytorch and mxnet all have explicit rows and cols as parameters.

houseroad: I think we can have another operator called `Eye` which takes rows and columns as inputs.

anirudhacharya: won't that be like duplicating the functionality? are there examples where this Eyelike operator is used?

spandantiwari: Yes, that's the idea. The `*Like` ops take the properties such as shape and type from the input tensor. This design is similar to Numpy (`ones_like`, `zero_like`) and also other ops in ONNX such as `ConstantLike`, `RandomNormalLike` and `RandonUniformLike`. 

We can have a separate PR for `Eye` if needed.

onnx/defs/generator/defs.cc
spandantiwari: Sure. I will add that.

onnx/defs/generator/defs.cc
spandantiwari: OK.

onnx/defs/generator/defs.cc
spandantiwari: @anirudhacharya - I don't think this is exactly duplicating functionality. There are several frameworks that subscribe to the design where there are two parallel set of generator ops - one that take tensor as input and one that take explicit shape as input, e.g. (`ones`, `ones_like`), (`zeros`, `zeros_like`). This kind of op is quite useful, e.g. creating epsilon covariance matrices matrices for regularization, and matrix factorization scenarios.

Also, note that I have not added `Eye` in this PR in order avoid adding more ops than necessary, because something like `Eye`, with static shape, can be supported using `Constant` op in current spec.

onnx/defs/generator/defs.cc
spandantiwari: Added.

spandantiwari:
Updated the PR based on the feedback.

spandantiwari:
@houseroad , @anirudhacharya - Thanks for the review.

fdwr:
@spandantiwari : Would it be useful to keep the 2D restriction (I can't think of any use case extending identity matrices to 3D/4D given matmul's are 2D anyway) but treat any higher dimensions as a batch count? I mean, if this operator is useful for processing one image, then it should be useful for processing a *batch* of images in parallel like MatMul and Conv2D (without Concat'enating a bunch of EyeLike outputs). I'm asking because GPU's are better at processing a lot of similar data in parallel than repeating lots of smaller varied executions.

fuzzyBatman:
Hello. I am new to ONNX and I am trying to understand how to add operators.
I am not unable to find the part where the logic for the op is.
Can someone help me understand this?

PR #1430, opened 2018-09-19
Optimization Pass: remove monotonic growing nop before argmax

### LINKED ISSUE #1429: Optimization Pass: Monotonic growing function is useless before argmax calculation.
It's standard to use softmax to generate a probability distribution at the end of a model. If we want to get the predicted class we end up with `net(x).softmax(dim=1).argmax(dim=1)`. 

In reality though because softmax is a monotonic growing operation we can replace the expression with `net(x).argmax(dim=1)`

ArmenAg:
fixes #1429


ArmenAg:
@houseroad How does this PR look? Is there a way I can request you as a reviewer. I can't seem to find a way to do that.

houseroad:
@ArmenAg I will take a look later. I guess you don't have the write access of this repository.

onnx/test/optimizer_test.py
```diff
@@ -758,6 +758,69 @@ def test_fuse_consecutive_softmax_log_multiple_out(self):  # type: () -> None
         assert optimized_model.graph.node[0].attribute[0].i == 2
         assert optimized_model.graph.node[1].op_type == "Exp"
 
+    def test_eliminate_nop_monotone_argmax_basic_no_node_axis(self):  # type: () -> None
+        for node_name in ["log", "exp", "sqrt"]:
```
houseroad: Please use `Log`, `Exp`, `Sqrt`. `log`, `exp` and `sqrt` are not legal names.

onnx/optimizer/passes/eliminate_nop_monotone_argmax.h
```diff
@@ -0,0 +1,56 @@
...
+
+#pragma once
+
+#include "onnx/optimizer/passes/optimize_pass.h"
+
+namespace ONNX_NAMESPACE {
+namespace optimization {
+
+const std::unordered_set<NodeKind> monotone_node_no_axis_kind{kLog,
+                                                              klog,
```
houseroad: Here we should use kLog, kExp, kSqrt. And remove klog, kexp, ksqrt from intern_strings

onnx/test/optimizer_test.py
```diff
@@ -758,6 +758,69 @@ def test_fuse_consecutive_softmax_log_multiple_out(self):  # type: () -> None
...
+                    [node, node2, argmax],
+                    "test",
+                    [helper.make_tensor_value_info(
+                        "X", TensorProto.FLOAT, (5, 7, 11))],
+                    [helper.make_tensor_value_info("Z", TensorProto.FLOAT, (5, 7, 11)),
+                     helper.make_tensor_value_info("Z1", TensorProto.FLOAT, (5, 7, 11))])
+                optimized_model = self._optimized(
+                    graph, ["eliminate_nop_monotone_argmax"])
+                assert optimized_model.graph == graph
+
```
houseroad: Another test you may want to add is two consecutive nop monotone + argmax.

houseroad:
@ArmenAg btw, by default, please make sure fix point mode is off. I have some concern about that some of our optimization passes may be not robust enough. We may get infinite loop there.

PR #1434, opened 2018-09-20
Removing experimental ConstantFill op.

spandantiwari:
As recommended, now that the [ConstantLike](https://github.com/onnx/onnx/blob/master/docs/Operators.md#ConstantLike) op is in, this PR removes the experimental `ConstantFill` op.


spandantiwari:
@houseroad - OK, sounds fine.

snnn:
Hi,

I feel concerned on it. Converters like onnxmltools, has to support all the recent ONNX versions. One converter version is mapped to multiple onnx versions. If you modified the schema history, you break this.

spandantiwari:
@snnn - I am not sure I understand the concern, specifically what you mean by 'schema history'. Are you referring to the change in Changelog.md or is it a larger concern about removing this op altogether?

snnn:
Hi @spandantiwari , 

First, if you get a file how could you know if it is a valid ONNX model file? You may run the onnx model checker. But, if the checker reports error, how could you know, it's the model is wrong, or just you are using the wrong ONNX version? It's likely this model is valid in ONNX 1.3 but not ONNX 1.4 because it contains an ConstantFill op. And for a model, there is no easy way to figure out which ONNX version works which doesn't.  

Second, if a backend need to support multiple ONNX versoins, it would be very hard.  If it statically linked to ONNX 1.4, which doesn't contain the ConstantFill OP, but the program got a model from some converter which is using ONNX 1.3, and the model contains a ConstantFill OP. The backend cannot find the schema for that OP.  The same thing also applies to converters.

bddppq:
We do have the mechanism to remove an op with proper opset versioning (thanks to #1317).
But this op was in experimental, so we delete it without dealing with opset.

snnn:
We have models in onnx zoo, and onnx get a new release for every 1-2 months. We don't guarantee these model will be supported in newer ONNX releases. Is that true?

spandantiwari:
@snnn - I understand you concern now. As @bddppq mentions, this is the reason for having opset versioning. The idea is that when removing an op (and I agree with you that we should not take op removal lightly), we bump up the opset version and that it will be known that the op is no longer part of the updated opset version (which is encoded in the model).

In this particular case, `ConstantFill` is an experimental op, which was not part of the main spec, so we did not bump the opset.

snnn:
In this particular case, for ConstantFill op, I'm fine. I'm not using it and I don't care the other people. 
But please remember ImageScaler is also an experimental op. If somebody created a PR to remove it, with a reason of it's an experimental op,  I wouldn't agree. Because ImageScaler is used in the tiny-yolo model in ONNX model zoo.

bddppq:
@prasanthpul @lupesko According to @snnn, the tiny-yolo model in model zoo has used an experimental op, could you guys help fixing it?

spandantiwari:
@houseroad - now that `ConstantOfShape` is in ONNX spec, I think we have all the use cases supported. Do you think we can get `ConstantFill` removed before 1.4 release?

linkerzhang:
@houseroad  can you push to fix the internal cases in your side and we can push this PR before 1.4 release? Thanks a lot! We may target to merge this mid next week.

spandantiwari:
Reviving this PR for ONNX 1.4 release. 
@houseroad - With `ConstantAsShape` in the spec now, we can move ahead with removing `ConstantFill` from PyTorch exporter. I can help with that. But I see that the ONNX version that is submoduled to PT is bit older. Is it OK to bring it up to latest ONNX so we can get `ConstantOfShape`?

If this PR breaks any ConstantFill related tests, maybe we can disable them (similar to what we did for `ConstantLike` https://github.com/onnx/onnx/pull/1716)?

houseroad:
Yeah, sure, we can do that, let's see

spandantiwari:
There were test failures for tests that are based on `ConstantFill`. I have updated the tests that were using `ConstantFill`, and removed the shape inference tests for `ConstantFill`.

The following PyTorch backend tests will still fail because of `ConstantFill`:
`OnnxBackendPyTorchOperatorModelTest.test_operator_lstm_cpu
OnnxBackendPyTorchOperatorModelTest.test_operator_rnn_cpu
OnnxBackendPyTorchOperatorModelTest.test_operator_rnn_single_layer_cpu`

Should we disable them for the interim?

spandantiwari:
@houseroad - the PyTorch backend tests listed above are the ones failing. Can we disable them?

houseroad:
@spandantiwari Rui and I are working on this. Hopefully we should be able to migrate the PyTorch exporter and Caffe2 frontend and backend to new ops.

houseroad:
@spandantiwari feel free to review https://github.com/pytorch/pytorch/pull/15938

spandantiwari:
@houseroad - I reviewed the PyTorch PR and left a comment.Overall it looks good. Waiting for it to get merged so that this PR can get unblocked.

houseroad:
@spandantiwari can we also have a whitelist in the checker to make sure the legacy models containing ConstantFill won't be rejected by the checker? We have a lot of models relying on this... I know it's ugly, but I don't want to break anything. We can print some warnings when detect such legacy experimental ops.

spandantiwari:
@houseroad - we can do that. But if it's OK with you, let's do that in a separate PR. It is related but doesn't block this PR, and may need some time. We are hoping to get this PR in for ONNX 1.4.

PR #1437, opened 2018-09-21
Fix output_shape of a testcase for ConvTranspose

shinh:
As per the documentation of the operator, `output_shape` seems
to expect the same rank as other attributes such as `strides`
or `pads`. In other words, `output_shape` should not have
dimensions for batch size and number of channels. So this
changes `output_shape` from [1, 2, 10, 8] to [10, 8] for the
2D ConvTranspose.

Note that test_convtranspose_kernel_shape, the other test which
uses `output_shape`, is already OK.


shinh:
I failed to get an attention in https://github.com/onnx/onnx/pull/1158 so decided to send a PR to see if my understanding is right. Comments are welcomed, thanks!

shinh:
Thanks for your quick review! I've updated those documents. I think my original commit had updated model.onnx and test_data_set_0 was unchanged since this PR changes inputs/outputs. I've just double-checked there are no update by doing

```
$ cd onnx/backend/test
$ python3 cmd_tools.py generate-data -o /tmp/onnx
$ cp -r /tmp/onnx/node/test_convtranspose* data/node/
$ git diff  # no update
```

shinh:
I'm not sure why stat_coverage.py used vgg16 for me although it was renamed, but anyway, now the tests seem to be passing.

zrphercule:
@shinh Hi shinh, I wonder if this is ready to be merged?

PR #1438, opened 2018-09-21
add Merge op

suexu1025:
- Add new operation: Merge.
It's based on MergeDim in caffe2.
Caffe2 only support axis = 1, which is merging first 2 dims, while this operation offers flexible for axis is arbitrary number.

before merge: Input [d0, d1, ... dn]
after merge: Output [d0d1...daxis, daxis+1, ... dn ]


houseroad:
Close, since it's able to covered by Reshape.

PR #1445, opened 2018-09-26
Fix undefined ONNX_API

yinghai:
https://github.com/onnx/onnx/pull/1407 added `ONNX_API` as `PRIVATE`. For upstream project which uses ONNX, if they include `onnx.pb.h`, compilation will fail due to undefined `ONNX_API`. This PR fixes it by propagating the definition.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,7 +286,7 @@ if (MSVC)
 else()
   set(ONNX_API_DEFINE "-DONNX_API=")
 endif()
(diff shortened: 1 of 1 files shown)


yinghai:
Actually this is not the correct fix. Downstream project should include `onnx/onnx_pb.h` instead of `onnx/onnx.pb.h` where `ONNX_API` is already defined.

orionr:
Thanks @yinghai. You are right - all downstream projects should include `onnx/onnx_pb.h` and not onnx.pb.h directly. Having the ONNX_API_DEFINE public would be a problem on Windows.

PR #1451, opened 2018-09-26
[ONNXIFI]Fix the crash when weightCount = 0

zrphercule:
Fix a few nits for onnxifi test driver.

### DIFF
--- a/onnx/backend/test/cpp/driver_test.cc
+++ b/onnx/backend/test/cpp/driver_test.cc
@@ -147,17 +147,17 @@ class ONNXCppDriverTest
         ONNX_NAMESPACE::checker::check_tensor(output, ctx);
       }
     }
-    /* TO DO:
-     * This chunk of code is to test the correctness of onnxifi backend.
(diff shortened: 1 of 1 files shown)


onnx/backend/test/cpp/driver_test.cc
```diff
@@ -233,17 +232,16 @@ class ONNXCppDriverTest
...
 TEST_P(ONNXCppDriverTest, ONNXCppDriverUnitTest){
   onnxifi_library lib;
-  onnxBackendID backendID;
+  onnxBackendID backendID = NULL;
   onnxBackend backend;
   if (!ONNXIFI_DUMMY_BACKEND) {
     EXPECT_TRUE(onnxifi_load(1, NULL, &lib));
-
-    size_t numBackends;
+    size_t numBackends = 1;
```
Maratyszcza: would be safer to initialize to `0`.

zrphercule: sure, thanks

PR #1452, opened 2018-09-27
ONNX Optimization Rewrite

ArmenAg:
I'd like to start off this convo by providing a skeleton of what the optimization framework will look like. 

We first define a set of enums which describe general attributes of the pass. For example we have PassType (e.g. fuse/nop), PassEfficiency (e.g. partial, complete), PassOptimizationType (e.g. memory, compute, stability). These attributes of the pass will help up build better PassManagers.

A pass contains a couple of methods. It can give you it's name as well as it's attributes. It has the ability to initialize a pass given a graph as well as finalize it. Finally it contains the runPass method which works both on IR and Proto (although we should think about only supporting IR. I don't see a fundamental benefit to optimizing over Proto). 

The runPass method returns a PostPassAnalysis which provides some information about what the pass has done (e.g. did initialization? did finalization? number transformers applied?) this is useful when deciding to do things such as fixed point optimization.

One fundamental type of pass we implement is PredicateBasedPass. A lot of code reuse is happening in the code framework wrt DescendOnGraphAttributes. The way we solve this is by using a PredicateBasedPass where one implements the predicate for when the transform will be applied as well as the transform. Our backend takes care of DescendOnGraphAttributes/PassAnalysis in a way agnostic to the user writing the pass.

@houseroad What do you think so far? Any design ideas which look wrong?


houseroad:
@ArmenAg overall, the idea sounds good to me. Some comments:
1) Let's add more comments and examples to explain how we define PassType, PassEfficiency, and PassOptimizationType.
2) Even if we provide onnx fix point optimizer, user can still be able to run customized passes. This is for backward compatibility.
3) Your PredicateBasedPass proposal sounds pretty useful, especially avoid copy-paste code.

ArmenAg:
@houseroad Perfect. I'll start implementing this PR end to end. I'll also rewrite the current optimizations through PredicateBasedPass. PassManagers will also be introduced.

ArmenAg:
@houseroad I've implemented the rewrite. All the optimization passes are now on the new framework for optimization. Could you please review the code. Also I'm still getting a CI error when compiling with MSVC, but I'm unsure what these errors mean. Would you mind helping me out?

houseroad:
@ArmenAg Sure, I will take a look tonight or tomorrow

ArmenAg:
@houseroad Sounds good. I also fixed the errors that appeared in MSVC. This PR doesn't implement fixed point optimization. Once this is accepted I can put fixed point optimization into a separate PR.

onnx/optimizer/optimize.cc
```diff
@@ -3,7 +3,8 @@
 
 #include "onnx/optimizer/optimize.h"
 
-namespace ONNX_NAMESPACE { namespace optimization {
+namespace ONNX_NAMESPACE {
+namespace optimization {
 
 // TODO: Remove this static reference
 static Optimizer _optimizer;
```
houseroad: We can have a static method to return the _optimizer, make sure _optimizer is already initialized.

onnx/optimizer/pass.cc
```diff
@@ -0,0 +1,98 @@
...
+      fn(*n->g(name));
+    }
+    if (kind == AttributeKind::gs) {
+      for (auto& g : n->gs(name)) {
+        fn(*g);
+      }
+    }
+  }
+}
+PredicateBasedPass::~PredicateBasedPass() {}
```
houseroad: Empty line?

onnx/optimizer/pass.h
```diff
@@ -0,0 +1,179 @@
...
+namespace optimization {
+
+// Base struct representing result of a pass.
+struct PostPassAnalysis {};
+
+// Enum that represents the type of optimization it is.
+enum PassType {
+  // Class of optimizations that fuses operations.
+  Fuse = 0,
+  // Class of optimizations that remove useless operations.
```
houseroad: removes

onnx/optimizer/pass.h
```diff
@@ -0,0 +1,179 @@
...
+  Nop = 1,
+  // Class of optimizations that includes some form of seperation.
+  Seperate = 2,
+  // Immutable pass, also sometimes referred to as an analysis pass.
+  Immutable = 3,
+  // Unknown type of pass.
+  Unk = 4
```
houseroad: To me, unknown sounds better, shall we call it others?

onnx/optimizer/pass_registry.h
```diff
@@ -0,0 +1,75 @@
...
+  Pass* find(std::string pass_name) {
+    auto it = this->passes.find(pass_name);
+    ONNX_ASSERTM(
+        it != this->passes.end(), "pass %s is unknown.", pass_name.c_str());
+    return it->second;
+  }
+  const std::vector<std::string> GetAvailablePasses();
+
+  template <typename T>
+  void registerOptimizer() {
```
houseroad: Shall we call it registerPass?

onnx/optimizer/pass.h
ArmenAg: I think Other is better. A user will not write a pass that is of unknown type but rather other.

ArmenAg:
@houseroad Fixed the issues mentioned above.

ArmenAg:
@houseroad I added fixed point optimization with deadend elimination. I'm again getting some MSVC error for protobuf. Could you take a look please?

houseroad:
@ArmenAg seems the master is broken.

onnx/optimizer/pass_manager.cc
```diff
@@ -0,0 +1,48 @@
...
+    delete pass_analysis;
+  }
+  return new EmptyPassManagerAnalysis();
+}
+
+PassManagerAnalysis* FixedPointPassManager::run(Graph& graph) {
+  bool fixed_point_optimization_done;
+
+  do {
+    fixed_point_optimization_done = false;
```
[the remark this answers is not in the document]
ArmenAg: Determining in what order to do passes is very challenging and is a core problem in optimization. This fixed point manager implements the simplest type of fixed point manager. In the future we can think about adding more types of pass managers.

houseroad:
I prefer to use smarter pointers everywhere, unless you have a strong argument for some cases. If so, I am open to discuss. Otherwise, let's just stick to smart pointer :-)

ArmenAg:
I don't have a good argument :) I can move everything to smart pointers. The old passes all used Node* should I convert all these to std::shared_ptr<Node>?

houseroad:
@ArmenAg sure, if you want, we can use shared_ptr to replace Node*. But it may need to change the IR API.

ArmenAg:
I'd prefer not to touch IR code in this PR. Could I keep it Node* for this PR. And sometime in the future we could migrate our IR to shared_ptr?

houseroad:
@ArmenAg sure, I think the pytorch-onnx-caffe2 tests are broken for some reason. Let me check what's going here

ArmenAg:
@houseroad Could you please review.

houseroad:
@ArmenAg I am investigating why with this change, the CI of ONNX becomes really unstable. Very likely, it will hang on the test_pytorch_onnx_caffe2.py... I am not able to reproduce the problem locally.

ArmenAg:
@houseroad I believe the bug is fixed. It was due to it.destroyCurrent() being called too many times. Running test_pytorch_onnx_caffe2.py:
```
master:
	real    1m27.064s
	user    3m9.906s
	sys     0m27.422s
rewrite:
	real    1m32.443s
	user    3m20.391s
	sys     0m29.016s
```
The extra time is from running the DeadEndElimination pass. Let's see how the CI turns out.

ArmenAg:
@houseroad Seems like the error is fixed. The CI is failing on test/onnx/test_operators.py but it seems like this is happening on master as well, based on your PR (https://github.com/onnx/onnx/pull/1501)

houseroad:
I think i have fixed the error, let's see

ArmenAg:
@houseroad looks like it's fixed!

houseroad:
@onnxbot retest this please

ArmenAg:
@houseroad Seems like there are more ONNX tests that need to be updated in PyTorch. The same error is happening in other PR's as well.

houseroad:
Yeah, I am working on that... annoying, the tests are not triggered in pytorch repo... I need to fix that

ArmenAg:
@houseroad The tests look to be figured out. How is this PR looking?

houseroad:
@onnxbot retest this please

onnx/optimizer/pass.h
```diff
@@ -0,0 +1,195 @@
...
+  ComputeMemory = 3,
+  // Optimizes for stability (e.g. log-sum-exp trick).
+  Stability = 4
+};
+
+enum NodeDestroyType {
+  // Does not destroy node
+  NoDestroy = 0,
+  // Equivalent to calling it.destroyCurrent() once.
+  WeakDestroy = 1,
```
houseroad: DestroyOne, DestroyTwo, and so on?

onnx/optimizer/pass_manager.cc
```diff
@@ -0,0 +1,46 @@
...
+        fixed_point_optimization_done = true;
+      }
+    }
+  } while (fixed_point_optimization_done);
+
+  return std::shared_ptr<PassManagerAnalysis>(new EmptyPassManagerAnalysis());
+}
+
+} // namespace optimization
+} // namespace ONNX_NAMESPACE
```
houseroad: newline

onnx/optimizer/optimize.h
```diff
@@ -63,34 +34,13 @@ struct Optimizer {
...
     return mp_out;
   }
 
-  template <class Optimizer, class... Args>
-  void registerOptimizer(Args&&... args) {
-    auto optimizer = make_unique<Optimizer>(std::forward<Args>(args)...);
-    passes[optimizer->name] = std::move(optimizer);
-  }
+ private:
+  PassManager* pass_manager;
```
houseroad: Add some comments why we choose raw pointer here?

onnx/optimizer/pass.h
```diff
@@ -0,0 +1,195 @@
...
+  explicit PredicateBasedPass(
+      PassType pass_type,
+      PassEfficiency pass_efficiency,
+      PassOptimizationType pass_optimization_type)
+      : Pass(pass_type, pass_efficiency, pass_optimization_type) {}
+  ~PredicateBasedPass() override;
+
+  virtual bool patternMatchPredicate(Node* node) = 0;
+  virtual bool
+  runTransform(Node* node, Graph& graph, NodeDestroyType& destroy_current) = 0;
```
houseroad: Add some comments about what should be put in the `runTransform`? (do everything but not deleting the node?)

onnx/optimizer/passes/eliminate_unused_initializer.h
```diff
@@ -58,8 +66,9 @@ struct EliminateUnusedInitializer final : public OptimizePass {
     }
   }
 
-  void optimize(Graph& graph) override {
+  std::shared_ptr<PostPassAnalysis> runPass(Graph& graph) {
```
houseroad: override?

onnx/optimizer/passes/fuse_add_bias_into_conv.h
```diff
@@ -15,140 +15,143 @@
...
-        n->replaceAllUsesWith(orig_conv->node());
-        it.destroyCurrent();
+  bool patternMatchPredicate(Node* node) override {
+    return node->kind() == kAdd && node->inputs()[0]->node()->kind() == kConv &&
+        node->inputs()[0]->node()->inputs().size() == 2;
+  }
+  bool runTransform(Node* n, Graph& graph, NodeDestroyType& destroy_current)
+      override {
+    // due to current broadcasting's constraint, Conv has to be the first
+    // oprand
```
houseroad: operand

onnx/optimizer/passes/fuse_add_bias_into_conv.h
```diff
@@ -15,140 +15,143 @@
...
-#include "onnx/optimizer/passes/optimize_pass.h"
+#include "onnx/common/assertions.h"
+#include "onnx/optimizer/pass.h"
 
-namespace ONNX_NAMESPACE { namespace optimization {
+namespace ONNX_NAMESPACE {
+namespace optimization {
 
-struct FuseAddBiasIntoConv final : public OptimizePass {
+struct FuseAddBiasIntoConv final : public PredicateBasedPass {
```
houseroad: Although this one can be considered as a predicatebase pass, however, it's very complicated, I would suggest to make it pass to make sure, the iterator will be valid...

onnx/optimizer/passes/fuse_bn_into_conv.h
```diff
@@ -93,45 +109,45 @@ struct FuseBNIntoConv final : public OptimizePass {
...
-      var.add(eps);                                                            \
-      var.sqrt();                                                              \
-      s.divide(var);                                                           \
-      W.scale_by_first_dim(s);                                                 \
-      bc.subtract(m);                                                          \
-      bc.multiply(s);                                                          \
-      bc.add(bbn);                                                             \
-
-    switch(s.elem_type()) {
+#define DO_COMPUTATION(TENSOR_TYPE, vec)                                 \
```
houseroad: undefine after use?

onnx/test/optimizer_test.py
```diff
@@ -901,6 +883,50 @@ def test_fuse_bn_into_conv_simple(self):  # type: () -> None
             np.testing.assert_almost_equal(
                 W * f[:, np.newaxis, np.newaxis, np.newaxis], new_W)
 
+    def test_deadend_elimination_simple(self):  # type: () -> None
```
houseroad: Can we reuse some code between test_dataend_elimination_simple and fix?

onnx/optimizer/pass_manager.cc
```diff
@@ -0,0 +1,46 @@
...
+
+std::shared_ptr<PassManagerAnalysis> FixedPointPassManager::run(Graph& graph) {
+  bool fixed_point_optimization_done;
+
+  do {
+    fixed_point_optimization_done = false;
+    for (std::shared_ptr<Pass> pass : this->passes) {
+      std::shared_ptr<PostPassAnalysis> analysis = pass->runPass(graph);
+      std::shared_ptr<CountBasedPassAnalysis> count_analysis =
+          std::dynamic_pointer_cast<CountBasedPassAnalysis>(analysis);
```
houseroad: Can we get ride of `dynamic_pointer_cast`?

onnx/optimizer/pass_manager.cc
```diff
@@ -0,0 +1,46 @@
...
+
+      while (nullptr != count_analysis &&
+             count_analysis->fixedPointOptimizationNeeded()) {
+        count_analysis = std::dynamic_pointer_cast<CountBasedPassAnalysis>(
```
houseroad: Can we get ride of `dynamic_pointer_cast`? Some projects would like to disable RTTI.

onnx/optimizer/passes/fuse_add_bias_into_conv.h
ArmenAg: I think we have one to one parity with the old implementation when using NodeDestroyType. It would make the code pretty messy especially if we want to use CountBasedAnalysis. Are you sure we should rewrite it using FullGraphBasedPass?

houseroad: I am fine to keep this as PredicateBasedPass, just want to make sure the logic is correct, since we did pretty complicated manipulations on the graph. I will double check the implementation on this again.

onnx/optimizer/pass_manager.cc
ArmenAg: How should we circumvent the use of RTTI? We need to know the type of analysis done in order to know whether or not to do fixed point optimization and this cannot be inferred at a compile time right?

houseroad: Can you add one field in the parent class? For different type of children, you assign with different values, and do the static cast at runtime. Does that sound good to you?

ArmenAg: Good idea. I'm working on the revisions now.

ArmenAg:
@houseroad I removed the need for RTTI following your suggestions.

ArmenAg:
My bad. I added the nop test back. Also all tests have passed disregarding unstable runs. Feel free to rerun the tests.

ArmenAg:
@houseroad There we go. All test's pass. Can we merge?

houseroad:
@ArmenAg done, thanks!

PR #1454, opened 2018-09-27
Add Compress Op

liqunfu:


fdwr:
@liqunfu : I appreciate that you included a summary in the description, justification, and a pointer to existing usage (per the contribution guidance https://github.com/onnx/onnx/blob/master/docs/CONTRIBUTING.md), as these are valuable for implementers.

liqunfu:
Hi @ezyang @bddppq @houseroad , it will be great if any of you can review this PR. Thanks.

onnx/defs/tensor/defs.cc
```diff
@@ -1121,4 +1121,29 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+ONNX_OPERATOR_SET_SCHEMA(
+    Compress,
+    9,
+    OpSchema()
+        .SetDoc(Compress_ver1_doc)
+        .Attr(
+            "axis",
+            "Axis along which to take slices.",
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
```
houseroad: If we want to align with Numpy, we should not set default value as 0

liqunfu: Gather takes indices. Compress takes condition of bools as input tensor. condition input may naturally come from ops like logical operations. It is not possible to use Gather to achieve what is needed here.

liqunfu: regarding default axis, I agree we shall follow numpy's default behavior - if not set, assume flatten source.

houseroad: Make sense.

houseroad: How about the default value of axis?

liqunfu: I am thinking there is no default value for axis attribute. So if the axis attribute is not set, the op will perform as stated: to flatten the input and to select from the flattened tensor according to the condition evaluation. The output, in this case, is a rank 1 tensor. I will add an example for this case.

liqunfu:
@houseroad Please take a look at this updated PR. Also I am getting ci/circleci failure.   ###211. Can you please also take a look? Thanks

liqunfu:
@houseroad, thanks for fixing build for this PR. Is there anything I need to do to complete it? Thanks.

liqunfu:
Hi @houseroad Still need your approval before I can merge. Thanks.

houseroad:
According to the log, seems your local onnx models are out of date. Please remove the cached model in ~/.onnx folder. and rerun https://github.com/onnx/onnx/blob/master/tools/update_doc.sh

liqunfu:
thanks @houseroad , I have updated my change with your comment. May you please approve and merge it ?

onnx/defs/tensor/defs.cc
```diff
@@ -1121,4 +1121,32 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::all_tensor_types(),
             "Constrain input and output types to all tensor types.")
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
+
+static const char* Compress_ver1_doc = R"DOC(
```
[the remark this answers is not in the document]
liqunfu: udpated to ver9

liqunfu:
@houseroad @bddppq @linkerzhang 
Hi, why is this? Do I need to do anything? I have signed cla and been able to contribute to ONNX before.
Thanks 
license/cla Expected — Waiting for status to be reported

prasanthpul:
@liqunfu - CLA refreshed; you are good to go

liqunfu:
@prasanthpul @houseroad Thank you so much guys!

PR #1455, opened 2018-09-27
[ONNXIFI]serialized model data in test driver, ir version is now corrected

zrphercule:
Bug: using unserialized model in test driver. It is now fixed.

### DIFF
--- a/onnx/backend/test/cpp/driver_test.cc
+++ b/onnx/backend/test/cpp/driver_test.cc
@@ -158,12 +158,14 @@ class ONNXCppDriverTest
                 model_.graph().initializer(0));
         weightDescriptors_pointer = &weightDescriptors;
       }
+      std::string serialized_model;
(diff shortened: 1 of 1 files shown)


onnx/backend/test/cpp/driver_test.cc
```diff
@@ -158,12 +158,14 @@ class ONNXCppDriverTest
                 model_.graph().initializer(0));
         weightDescriptors_pointer = &weightDescriptors;
       }
+      std::string serialized_model;
+      model_.SerializeToString(&serialized_model);
       EXPECT_EQ(
           lib.onnxInitGraph(
               backend,
               NULL,
-              sizeof(model_),
```
yinghai: Did I review this and let it slip through? :)

zrphercule: (shhhhhh, lets keep it as a secret

PR #1456, opened 2018-09-27
[ONNXIFI]gtest:expect to assert

zrphercule:
Change the expect to assert for some tests in onnxifi test driver. This can prevent onnxifi test driver from crashing once model is not supported by backend.


yinghai:
Interesting, what's the difference?

bddppq:
@yinghai `expect` will not immediately exit and subsequent code can cause segfault

PR #1457, opened 2018-09-28
Extract ONNX_API macro to a common header and export OpSchemaRegistry

bddppq:


onnx/api_macros.h
```diff
@@ -0,0 +1,46 @@
+#ifndef ONNX_API_H
```
houseroad: I think better to use `ONNX_API_MACROS_H`
in case some day we have onnx/api.h

bddppq: oops that's because I initially named this file as onnx/api.h

cmake/Utils.cmake
```diff
@@ -27,4 +27,11 @@ function(add_onnx_global_defines target)
     target_compile_definitions(${target} PUBLIC "ONNX_USE_LITE_PROTO=1")
   endif()
 
+  if (BUILD_SHARED_LIBS)
+    target_compile_definitions(${target} PRIVATE "ONNX_BUILD_SHARED_LIBS=1")
+  endif()
+
+  if (ONNX_BUILD_MAIN_LIB)
+    target_compile_definitions(${target} PRIVATE "ONNX_BUILD_MAIN_LIB=1")
```
yinghai: What's the purpose of these build variants?

zrphercule: @yinghai See the the followed api_macros.h, they are used there. I think it is used to deal with some ms dll stuff.

PR #1460, opened 2018-09-28
Change Function registry flow; Get rid of whole-archive in compile

raymondxyang:
* Change Function registry flow into the Opschema registering way.
* Get rid of whole-archive during compiling

### DIFF
--- a/onnx/defs/experiments/experiments_functions.cc
+++ b/onnx/defs/experiments/experiments_functions.cc
@@ -4,8 +4,8 @@
 #include "onnx/common/constants.h"
 #include "onnx/common/model_helpers.h"
 #include "onnx/defs/function.h"
-using namespace ONNX_NAMESPACE;
 
+namespace ONNX_NAMESPACE {
 static Common::Status BuildMVN(std::unique_ptr<FunctionProto>* func_proto) {
   if (nullptr == func_proto) {
     return Common::Status(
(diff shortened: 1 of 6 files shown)


onnx/defs/experiments/experiments_functions.cc
```diff
@@ -171,5 +171,11 @@ static Common::Status BuildMVN(std::unique_ptr<FunctionProto>* func_proto) {
...
 }
 
-ONNX_FUNCTION(
+ONNX_FUNCTION_BUILD(
+    MeanVarianceNormalization,
+    9,
     FunctionBuilder().SetDomain(ONNX_DOMAIN).SetBuildFunction(BuildMVN));
+
+} // namespace ONNX_NAMESPACE
+// ONNX_FUNCTION(
```
houseroad: clean up?

raymondxyang: This means to leave a example for registering functions. I ll update the marco to be the latest one.

raymondxyang:
@RyanUnderhill please help take a look thanks😄

onnx/defs/function.cc
```diff
@@ -46,6 +47,17 @@ Common::Status FunctionBuilderRegistry::GetFunctions(
...
 
+  class FunctionBuilderRegisterer {
+   public:
+    FunctionBuilderRegisterer() {
+      RegisterOnnxFunctionBuilder();
+    }
+  };
+
+#ifndef __ONNX_DISABLE_STATIC_REGISTRATION
+  static FunctionBuilderRegisterer functionBuilder_registerer;
```
RyanUnderhill: If you want to remove the dummy class, you can write this:

`static bool functionBuilder_registerer=(RegisterOnnxFunctionBuilder(), false);`

onnx/defs/function.h
```diff
@@ -62,6 +62,25 @@ class FunctionBuilderRegistry : public IFunctionBuilderRegistry {
...
+
+#define ONNX_FUNCTION_BUILDER_CLASS_NAME(domain, ver, name) \
+  name##_##domain##_ver##ver
+
+#define ONNX_FUNCTION_BUILD(name, ver, build_func) \
+  ONNX_FUNCTION_BUILD_HELPER(                      \
+      __COUNTER__, name, Onnx, ONNX_DOMAIN, ver, build_func)
+
+#define ONNX_FUNCTION_BUILD_HELPER(                                           \
+    counter, name, domain, domain_str, ver, build_func)                       \
```
jeffbloo: The counter parameter can be removed here

PR #1461, opened 2018-09-28
Remove default value for 'dtype' attribute in ConstantLike op.

spandantiwari:
This is a bug in the spec of `ConstantLike` op. The spec says that if `dtype` is not specified then the type of the input tensor should be used as the output tensor type. But currently, with the default value specified for this op, this attribute will always be _present_, and therefore, it is contrary to the spec. With this fix the default value is removed, which will allow the `dtype` attribute to be absent, in which case the input tensor type can be used for output type. 

Since this op was checked in very recently (less than two weeks ago), I am not updating the opset for this change.

### DIFF
--- a/onnx/defs/generator/defs.cc
+++ b/onnx/defs/generator/defs.cc
@@ -113,7 +113,14 @@ ONNX_OPERATOR_SET_SCHEMA(
              "tensor(bool)"},
              "Constrain output types. Strings and complex are not supported.")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
-          propagateElemTypeFromAttributeToOutput(ctx, "dtype", 0);
+          if (ctx.getAttribute("dtype") != nullptr) {
+            propagateElemTypeFromAttributeToOutput(ctx, "dtype", 0);
+          }
+          else {
+            if (hasNInputShapes(ctx, 1)) {
+              propagateElemTypeFromInputToOutput(ctx, 0, 0);
+            }
+          }
           if (hasNInputShapes(ctx, 1)) {
             propagateShapeFromInputToOutput(ctx, 0, 0);
           }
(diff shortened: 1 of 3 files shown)


onnx/defs/generator/defs.cc
```diff
@@ -59,10 +59,10 @@ ONNX_OPERATOR_SET_SCHEMA(
         .Attr(
             "dtype",
             "(Optional) The data type for the elements of the output tensor. If not specified,"
-            "the data type of the input tensor T1 is used. If input tensor T1 is also not" 
-            "specified, then type defaults to 'float'.",
+            "the data type of the input tensor T1 is used. If input tensor T1 is also not " 
+            "specified, then output tensor type defaults to 'float'.",
```
houseroad: default type should be the same as the input tensor?

spandantiwari: Yes, if `dtype` is not specified, then output type should be the same as input tensor. This is the case when even input tensor is not provided (i.e. output created using `shape` attribute).

houseroad: Yeah, I mean you didn't change the description :-)

spandantiwari: umm, sorry, I am not sure if the description needs to be changed. The lines you are pointing to (62-63) are for the case when both `dtype` and input tensor are not provided, in which case the output tensor type, by default, should be float. 

Maybe I am misunderstanding your comment?

spandantiwari:
@houseroad - oh right, there hasn't be a release since this went in. So we're good. Thanks for reminding. :-)

onnx/defs/generator/defs.cc
spandantiwari: Cool, thanks.

houseroad:
Our CI is broken, let's wait for it, will be back soon.

PR #1462, opened 2018-09-28
[ONNXIFI]Add examine of onnxGetBackendCompatibility and onnxGetBackendInfo in onnxifi test driver

zrphercule:
Add two more expect/assert in test driver.
The examine of compat will finally be changed to skipping the test instead of failing it.

### DIFF
--- a/onnx/backend/test/cpp/driver_test.cc
+++ b/onnx/backend/test/cpp/driver_test.cc
@@ -48,6 +48,32 @@ class ONNXCppDriverTest
     }
     return d_size;
   }
+  bool IsUnsupported(onnxStatus s, std::string& msg) {
+    if (s == ONNXIFI_STATUS_UNSUPPORTED_VERSION) {
+      msg = "Unsupported version of ONNX or operator.";
+      return true;
+    }
+
+    if (s == ONNXIFI_STATUS_UNSUPPORTED_OPERATOR) {
+      msg = "Unsupported operator.";
+      return true;
+    }
+
+    if (s == ONNXIFI_STATUS_UNSUPPORTED_ATTRIBUTE) {
+      msg = "Unsupported attribute.";
+      return true;
+    }
+
+    if (s == ONNXIFI_STATUS_UNSUPPORTED_SHAPE) {
+      msg = "Unsupported shape.";
+      return true;
+    }
+    if (s == ONNXIFI_STATUS_UNSUPPORTED_DATATYPE) {
+      msg = "Unsupported datatype";
+      return true;
+    }
+    return false;
+  }
   bool IsDescriptorEqual(
       const onnxTensorDescriptorV1& x,
       const onnxTensorDescriptorV1& y) {
@@ -134,7 +160,10 @@ class ONNXCppDriverTest
     }
     return true;
   }
-  void RunAndVerify(onnxifi_library& lib, onnxBackend& backend) {
+  void RunAndVerify(
+      onnxifi_library& lib,
+      onnxBackend& backend,
+      onnxBackendID backendID) {
     // Check Model
     ONNX_NAMESPACE::checker::check_model(model_);
     // Check Input&Output Tensors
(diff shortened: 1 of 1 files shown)


onnx/backend/test/cpp/driver_test.cc
```diff
@@ -160,6 +163,10 @@ class ONNXCppDriverTest
       }
       std::string serialized_model;
       model_.SerializeToString(&serialized_model);
+      ASSERT_EQ(
+          lib.onnxGetBackendCompatibility(
+              backendID, serialized_model.size(), serialized_model.data()),
+          ONNXIFI_STATUS_SUCCESS);
```
houseroad: So if it is not compatible, do we need to record/report something?

zrphercule: Right now we will fail this test directly. Guess if we should skip this test once we found it is not compatible?

rdzhabarov: I think it's a valid response if the model cannot be supported by the ONNXIFI implementation.

I'd be in favor to make the test pass if the response is `ONNXIFI_STATUS_UNSUPPORTED_OPERATOR` or `ONNXIFI_STATUS_SUCCESS`.

* In case of `ONNXIFI_STATUS_SUCCESS` do other checks as we do currently.
* In case of `ONNXIFI_STATUS_UNSUPPORTED_OPERATOR` make test pass and report warning for the given test. 
* In case of any other results fail the test immediately.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -241,14 +248,21 @@ TEST_P(ONNXCppDriverTest, ONNXCppDriverUnitTest){
...
   onnxBackendID backendID = NULL;
   onnxBackend backend;
+  const size_t max_info_size = 50; // The maximum backend info size
   if (!ONNXIFI_DUMMY_BACKEND) {
     ASSERT_TRUE(onnxifi_load(1, NULL, &lib));
     size_t numBackends = 0;
     lib.onnxGetBackendIDs(&backendID, &numBackends);
     const uint64_t backendProperties[] = {ONNXIFI_BACKEND_PROPERTY_NONE};
     lib.onnxInitBackend(backendID, backendProperties, &backend);
+    char infovalue[max_info_size];
```
rdzhabarov: how does this work?

```
onnxBackendID backendID = NULL;
...
size_t numBackends = 0;
    lib.onnxGetBackendIDs(&backendID, &numBackends);
```
In case backendID == nullptr, ONNXIFI implementation will just populate numBackends to be equal to number of supported backends.

Don't you need to call `onnxGetBackendIDs` again to actually initialize backendID?

onnx/backend/test/cpp/driver_test.cc
zrphercule: @houseroad @rdzhabarov Just realized gtest doesnt have skipping... I will let these test passed. 
https://github.com/google/googletest/pull/1544

onnx/backend/test/cpp/driver_test.cc
yinghai: What @rdzhabarov said. Normally, we need two passes. First to get the number of backends. And second to query.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -160,6 +189,16 @@ class ONNXCppDriverTest
       }
       std::string serialized_model;
       model_.SerializeToString(&serialized_model);
+      std::cout << model_.graph().name() << std::endl;
```
rdzhabarov: do you want to dump graph name every time? (maybe just for soft failures as you do in IsUnsupported).

zrphercule: Dont worry, I will submit another pr to deal with this.
Right now TEST_P only output a meaningless number of testcase, I just found a way to output the model name instead.

PR #1465, opened 2018-10-02
add more types support for constant and matmul op.

linkerzhang:


onnx/defs/operator_sets.h
```diff
@@ -481,6 +483,9 @@ class OpSet_Onnx_ver9 {
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, EyeLike)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, Greater)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, Less)>());
+	// Add more types' support to Constant/MatMul op.
```
houseroad: indent :-)

onnx/defs/math/defs.cc
```diff
@@ -730,22 +715,28 @@ ONNX_OPERATOR_SET_SCHEMA(
...
     OpSchema()
         .Input(0, "A", "N-dimensional matrix A", "T")
         .Input(1, "B", "N-dimensional matrix B", "T")
         .Output(0, "Y", "Matrix multiply results from A * B", "T")
         .TypeConstraint(
             "T",
-            {"tensor(float16)", "tensor(float)", "tensor(double)"},
-            "Constrain input and output types to float tensors.")
-        .SetDoc(MatMul_ver6_doc)
+            {"tensor(float16)",
```
houseroad: Shall we use all_numeric_types instead?

linkerzhang: no, all_numeric_types contains low precision ones there. To support low-precision, we'd have different definitions. for example, for matmul, the input may be 8 bits integer but the output should be 16-bit or 32-bit (common use) integer to handle overflow. Make sense please?

PR #1473, opened 2018-10-03
[ONNXIFI]Add skip logic in onnxifi test driver

zrphercule:
Add gtest_skip in test driver to skip no-compatible test cases.
now we have 275 skipped test cases.


bddppq:
@rdzhabarov It uses cmake's external project feature which will git clone during build time (I personally prefer submodule though)

Issue #1443, opened 2018-09-25
Shall we bump operator set version for both backward and forward compatibility breaking changes? 


linkerzhang:
It's stated that operator set version is bumped if and only if there's a backward compatibility breaking change in ONNX. However, there're two facts/issues,
1. Previously changes didn't fully follow this rule. For example, adding new operators is always against new operator set version.
2. For non backward breaking but forward breaking changes, say, adding type constraints for an op, without bumping operator set version,  it's hard for an onnx runtime to elegantly fail to run a model referring the updated op.

Thoughts? @bddppq @houseroad  @yuanbyu @ebarsoum @gramalingam @lupesko

gramalingam:
Yes, I agree … it makes sense to bump opset version in both cases

houseroad:
I think we already did both in practice. For example, adding new op, even adding new attributes.

houseroad:
@linkerzhang yeah, that sounds good on our side. :-)

linkerzhang:
@fdwr

fdwr:
@linkerzhang Should the issue be closed now, after 1475?

linkerzhang:
Closing this issue as it is resolved by #1415

PR #1475, opened 2018-10-03
Operator set versioning - tighten wording regarding breaking changes

fdwr:
It is highly desirable for model loaders to quickly gauge their compatibility with the operators in a model when loading it and to reliably know that they can compute correct results. The current wording though for op set versioning is unclear regarding optional attributes, which makes it sound like it's legal to add a new attribute to an operator without updating the version number so long as (a) it's marked optional, and (b) it's absence would yield the same behavior as the earlier version of the operator. This is okay when the attribute *is* missing, but it would be bad for the model loader when the new attribute is present as the execution would think that it could successfully process a model and all operators in it, only to yield incorrect results because it was unaware of a newly added attribute that it should have considered.

So, any parameter additions which affect behavior, optional or not, should increment the operator version. Unlike some other file formats which can afford more lenient fallback in the results presented to the user (e.g. color SVG fonts falling back to black&white glyph outlines or JPEG-HDR falling back to LDR gamut), ML must either have a clear up-front failure or should have correct results (e.g. the difference between a 75% chance vs 5% of cancer diagnosis matters).

This tightens wording to include:
- attribute additions/changes/removals
- tensor additions/removals/reorderings.
- data type additions/removals. Previously, type additions which didn't affect behavior were just considered documentation updates, without incrementing the op version, but it's desirable to know type compatibility immediately.

### DIFF
--- a/docs/Versioning.md
+++ b/docs/Versioning.md
@@ -69,27 +69,19 @@ ONNX is defined such that the IR can evolve independently from the set of operat
 
 A given operator is identified by a three-tuple: `(domain, op_type, and op_version)`. This is written as `domain.op_type:op_version` in prose (e.g., `com.acme.FastConv:3`).  Nodes in graphs always refer to operators by their three-part identifier.
 
-The semantics of an operator MAY be extended in a backwards-compatible
-way without requiring the `op_version` of an operator to be increased.
-A backwards compatible semantics change is defined to be a change to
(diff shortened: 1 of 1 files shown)


fdwr:
#1443 @linkerzhang

Lall-091:
07f359a1433caa341094693d7059b364e73db95e

PR #1476, opened 2018-10-04
Clean memory when failed test

zrphercule:
Right now, we dont clean when our test failed, which may cause memory leaking.
In this pr we introduced a macro to handle the clean up in test driver.
We also updated the testing process in onnxifi test driver, fixed a few bug of not using signal.

### DIFF
--- a/onnx/backend/test/cpp/driver_test.cc
+++ b/onnx/backend/test/cpp/driver_test.cc
@@ -20,17 +20,6 @@
 #define ONNXIFI_TESTDATA_EPS 1e-5
 #endif
 
-#define ASSERT_MEMORYSAFE_SUCCESS(X, isGraphInitialized)  \
-  {                                                       \
-    onnxStatus sta = X;                                   \
-    if (sta != ONNXIFI_STATUS_SUCCESS) {                  \
-      if (isGraphInitialized) {                           \
-        lib.onnxReleaseGraph(graph);                      \
-      }                                                   \
-    }                                                     \
-    ASSERT_EQ(sta, ONNXIFI_STATUS_SUCCESS);               \
-  }
-
 const float onnxifi_testdata_eps = ONNXIFI_TESTDATA_EPS;
 
 template <typename T>
@@ -52,6 +41,7 @@ class ONNXCppDriverTest
     protos_ = t.proto_test_data_;
     model_ = t.model_;
   }
+
   uint64_t GetDescriptorSize(const onnxTensorDescriptorV1* t) {
     uint64_t d_size = 1;
     for (int i = 0; i < t->dimensions; i++) {
@@ -59,6 +49,19 @@ class ONNXCppDriverTest
     }
     return d_size;
   }
+
+  void GtestAssertMemorySafeSuccess(
+      onnxStatus x,
+      onnxGraph* graph_pointer,
+      onnxifi_library& lib) {
+    if (x != ONNXIFI_STATUS_SUCCESS) {
+      if (graph_pointer != NULL) {
+        lib.onnxReleaseGraph(*graph_pointer);
+      }
+    }
+    ASSERT_EQ(x, ONNXIFI_STATUS_SUCCESS);
+  }
+
   bool IsUnsupported(onnxStatus s, std::string& msg) {
     if (s == ONNXIFI_STATUS_UNSUPPORTED_VERSION) {
       msg = "Unsupported version of ONNX or operator.";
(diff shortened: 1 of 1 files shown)


onnx/backend/test/cpp/driver_test.cc
```diff
@@ -20,6 +20,17 @@
 #define ONNXIFI_TESTDATA_EPS 1e-5
 #endif
 
+#define ASSERT_MEMORYSAFE_SUCCESS(X, isGraphInitialized)  \
+  {                                                       \
+    onnxStatus sta = X;                                   \
+    if (sta != ONNXIFI_STATUS_SUCCESS) {                  \
+      if (isGraphInitialized) {                           \
+        lib.onnxReleaseGraph(graph);                      \
```
houseroad: So, we have to assume that graph is named graph, hmm...

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -20,6 +20,17 @@
 #define ONNXIFI_TESTDATA_EPS 1e-5
 #endif
 
+#define ASSERT_MEMORYSAFE_SUCCESS(X, isGraphInitialized)  \
```
rdzhabarov: i'm not sure what's coding standards for ONNX. I'd favor just a simple method instead of #define (easier write/read/debug/... )

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -199,8 +210,8 @@ class ONNXCppDriverTest
         GTEST_SKIP();
         return;
       }
-      ASSERT_EQ(is_compatible, ONNXIFI_STATUS_SUCCESS);
-      ASSERT_EQ(
+      ASSERT_MEMORYSAFE_SUCCESS(is_compatible, false);
```
rdzhabarov: why do we specifically care about releasing graph but not onnxReleaseBackend or onnxReleaseBackendID?

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -237,25 +248,24 @@ class ONNXCppDriverTest
...
         outputFence.tag = ONNXIFI_TAG_MEMORY_FENCE_V1;
         inputFence.type = ONNXIFI_SYNCHRONIZATION_EVENT;
         outputFence.type = ONNXIFI_SYNCHRONIZATION_EVENT;
 
-        ASSERT_EQ(
-            lib.onnxRunGraph(graph, &inputFence, &outputFence),
-            ONNXIFI_STATUS_SUCCESS);
-        ASSERT_EQ(lib.onnxWaitEvent(outputFence.event), ONNXIFI_STATUS_SUCCESS);
+        ASSERT_MEMORYSAFE_SUCCESS(
+            lib.onnxRunGraph(graph, &inputFence, &outputFence), true);
```
rdzhabarov: I might be missing something, but how does the whole workflow work without actually signaling on inputFence?

onnx/backend/test/cpp/driver_test.cc
zrphercule: In fact we do care about releasing backend, backendID and free memory. However, they are done in TEST_P() instead of in this test class, therefore even the ASSERT failed in class, the release of backend would still be excuted. I have tested this mechanism myself.

onnx/backend/test/cpp/driver_test.cc
zrphercule: I think you are right about this! I wonder why some of the results are correct...
It might because I didnot initialie the event at all, and therefore both output.event and input.event are set to signaled by default.
Let me fix this, thanks!

onnx/backend/test/cpp/driver_test.cc
rdzhabarov: People would always argue about things like that :D

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es31-dont-use-macros-for-constants-or-functions

onnx/backend/test/cpp/driver_test.cc
zrphercule: (And also even we dont release&free, OS would always do these things for us. quote: a google testing engineer.... Considering to uninstall all google product in my computer...

onnx/backend/test/cpp/driver_test.cc
houseroad: Can we pass X and graph? if you the graph is not initialized, we just pass nullptr, does that sound good to you?

zrphercule: @houseroad I guess it is better.

onnx/backend/test/cpp/driver_test.cc
rdzhabarov: "And also even we dont release&free, OS would always do these things for us" This is true, but ASAN would often report leaks in this cases.

zrphercule:
After merging of this pr, we would observe more failed, crashed and deadlocked testcases...

houseroad:
Let's kill all the bugs

zrphercule:
mind give a stamp? @houseroad

PR #1480, opened 2018-10-05
[ONNXIFI]Reliable memory of shape in test driver

zrphercule:
We didnt use persistent memory in test driver, which may cause memory leaking.

### DIFF
--- a/onnx/backend/test/cpp/driver_test.cc
+++ b/onnx/backend/test/cpp/driver_test.cc
@@ -175,7 +175,10 @@ class ONNXCppDriverTest
     return true;
   }
 
+  // A memory pool of shape information of onnxTensorDescriptorV1 used in this
+  // test driver.
   std::vector<std::vector<uint64_t>> shape_pool;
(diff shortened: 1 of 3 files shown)


onnx/backend/test/cpp/driver_test.cc
```diff
@@ -174,6 +174,10 @@ class ONNXCppDriverTest
     }
     return true;
   }
+
+  std::vector<std::vector<uint64_t>> shape_pool;
```
houseroad: could you add some comments about these variables?

yinghai: Separate member variables from member functions?

zrphercule: @yinghai  Yes I think this pool is better to be designed as a member variable of class instead of function, is it?

zrphercule: @yinghai oops I know what you mean now...... okay I will separate them hhh

PR #1484, opened 2018-10-09
Some CMake script changes for clang-cl

peterjc123:


raymondxyang:
LGTM.. @snnn do we depend on $MP_FLAG?

snnn:
@raymondxyang no

PR #1491, opened 2018-10-10
[ONNXIFI]Maintain Gtest main.cc argument 

zrphercule:
Not to ruin gtest argument by always assuming the last argument to be the data target dir

### DIFF
--- a/onnx/backend/test/cpp/test_main.cc
+++ b/onnx/backend/test/cpp/test_main.cc
@@ -2,11 +2,12 @@
 #include "gtest_utils.h"
 
 GTEST_API_ int main(int argc, char** argv) {
-  if (argc != 2) {
+  if (argc < 2) {
     std::cerr << "target directory must be given!" << std::endl;
     return EXIT_FAILURE;
   }
-  std::string target_dir = argv[1];
+  argc--;
+  std::string target_dir = argv[argc];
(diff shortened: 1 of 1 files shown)


onnx/backend/test/cpp/test_main.cc
```diff
@@ -2,11 +2,12 @@
 #include "gtest_utils.h"
 
 GTEST_API_ int main(int argc, char** argv) {
-  if (argc != 2) {
+  if (argc < 2) {
     std::cerr << "target directory must be given!" << std::endl;
     return EXIT_FAILURE;
   }
-  std::string target_dir = argv[1];
+  argc--;
```
houseroad: later, argc is used again by init google test, do we want expose the change on argc to initgoogletest?

zrphercule: @houseroad Yes, because I dont want google test to receive this dir anymore.

rdzhabarov: Sorry for late comment.

But it's not clear why this `argc--` magic happens here, can you add a comment? (and that we always expect target directory to be the last param).

zrphercule: oops, sure we need to add comments here, let me do it.

PR #1492, opened 2018-10-10
Change function ownership to ONNX

raymondxyang:
This avoids redundant copies of functionproto returned to any API users.


raymondxyang:
Reopen when ready

PR #1495, opened 2018-10-11
fix onnx checker to support proto3 models.

linkerzhang:
In proto3 models, if the attribute value is set to be the default value (say 0 for int), it will not really have the value set, then used_fields will be 0. We'd relax the checker logic to pass such models.

### DIFF
--- a/onnx/checker.cc
+++ b/onnx/checker.cc
@@ -229,7 +229,7 @@ void check_attribute(
 	  fail_check(
 		  "Attribute (name: ",
 		  attr.name(),
-		  ") should contain less than one value field.");
+		  ") should not contain more than one value field.");
   }
 
   if (!ctx.is_main_graph()) {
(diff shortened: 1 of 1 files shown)


onnx/checker.cc
```diff
@@ -223,20 +223,23 @@ void check_attribute(
...
-          ") should contain one and only one value field.");
-    }
-  } else {
+  // Normally, used_fields is expected to be 1.
+  // In proto3, when the value to be set is type default value (say 0 for int), used_fields may be 0.
+  if (used_fields > 1) {
+	  fail_check(
+		  "Attribute (name: ",
+		  attr.name(),
+		  ") should contain less than one value field.");
```
gramalingam: "should not contain more than one value field"

gramalingam: or "contains more than one value field."

linkerzhang: Fixed.

PR #1494, opened 2018-10-11
Add MaxUnpool op to ONNX.

spandantiwari:
This PR adds a new op, MaxUnpool, to ONNX. Unpooling is a operation commonly used to invert a pooling op and is typically useful for visualizing intermediate feature layers ([ref.]( https://arxiv.org/pdf/1311.2901v3.pdf)) in a convolutional network. Several frameworks have an unpooling op and this operation cannot be easily supported by any of the existing ONNX ops.

PR #1499, opened 2018-10-11
Add several hyperbolic function ops.

kit1980:
Add acosh, asinh, atanh, cosh, sinh. These ops are used in CNTK and Tensorflow.


kit1980:
Unrelated test_addmm failing, same as in another recent PR https://github.com/onnx/onnx/pull/1494

kit1980:
Can someone looks at this?
The ops are very similar to the existing ops, and are needed to convert CNTK and TF models.

ankkhedia:
@Roshrini @vandanavk

ebarsoum:
@houseroad  and @bddppq can someone take a look?

kit1980:
Rebased and updated the docs.

PR #1507, opened 2018-10-13
Fix googletest building when using ninja

zrphercule:
googletest target cannot be found when using ninja to build onnx.
this pr fixed it by adding BUILD_BYPRODUCTS macro.


bddppq:
https://stackoverflow.com/a/43714505

PR #1517, opened 2018-10-17
Add Scatter op to ONNX

pk-g:
Adding Scatter op and corresponding tests to ONNX


onnx/backend/test/case/node/scatter.py
```diff
@@ -0,0 +1,51 @@
...
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+class Scatter(Base):
+
+    @staticmethod
+    def export_scatter_0():  # type: () -> None
```
spandantiwari: I would suggest using better names for these test points. For example, we can eliminate the explicit use of the `axis=0` attribute here, and call this `export_without_axis`. In the next test point, we can have the axis attribute (`axis=1`) and that can be called `export_with_axis`.

onnx/defs/operator_sets.h
```diff
@@ -93,6 +93,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Relu);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Reshape);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Scale);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, ScaledTanh);
+class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Scatter);
```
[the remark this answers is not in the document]
pk-g: fixed

onnx/defs/operator_sets.h
```diff
@@ -225,6 +226,7 @@ class OpSet_Onnx_ver1 {
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Reshape)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Scale)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, ScaledTanh)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 1, Scatter)>());
```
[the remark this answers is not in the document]
pk-g: fixed

onnx/defs/tensor/defs.cc
```diff
@@ -610,6 +610,94 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
         }));
 
+static const char* Scatter_ver1_doc = R"DOC(
```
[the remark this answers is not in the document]
pk-g: fixed

onnx/defs/tensor/defs.cc
```diff
@@ -610,6 +610,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  data = [[1.0, 2.0, 3.0, 4.0, 5.0]]
+  indices = [[1, 3]]
+  updates = [[1.1, 2.1]]
+  axis = 1
+  output = [[1.0, 1.1, 3.0, 2.1, 5.0]]
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    Scatter,
+    1,
```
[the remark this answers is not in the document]
pk-g: fixed

onnx/defs/tensor/defs.cc
```diff
@@ -610,6 +610,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          if (!hasNInputShapes(ctx, 3)) {
+            return;
+          }
+
+          int r = ctx.getInputType(0)->tensor_type().shape().dim_size();
+
+          int out_rank = r;
+
+          if (out_rank == 0) {
```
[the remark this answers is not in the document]
pk-g: fixed.

onnx/defs/tensor/defs.cc
```diff
@@ -610,6 +610,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            2,
+            "updates",
+            "Tensor of rank r >=1 (same rank and type as input data)",
+            "T"
+        )
+        .Output(0, "output", "Tensor of rank r >= 1 (same rank as input).", "T")
+        .TypeConstraint(
+            "T",
+            OpSchema::all_tensor_types(),
+            "Constrain input and output types to any tensor type.")
```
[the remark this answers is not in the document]
pk-g: fixed

onnx/backend/test/case/node/scatter.py
pk-g: fixed

onnx/backend/test/case/node/scatter.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @staticmethod
+    def export_scatter_without_axis():  # type: () -> None
+        node = onnx.helper.make_node(
+            'Scatter',
+            inputs=['data', 'indices', 'updates'],
+            outputs=['y'],
+        )
+        data = np.zeros((3, 3))
+        indices = np.array([[1, 0, 2], [0, 2, 1]])
```
houseroad: please specify the dtype as int64, otherwise, we may generate different type of data on 32bit platform.

onnx/backend/test/case/node/scatter.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @staticmethod
+    def export_scatter_with_axis():  # type: () -> None
+        node = onnx.helper.make_node(
+            'Scatter',
+            inputs=['data', 'indices', 'updates'],
+            outputs=['y'],
+            axis=1,
+        )
+        data = np.array([[1.0, 2.0, 3.0, 4.0, 5.0]])
```
houseroad: ditto

onnx/defs/tensor/defs.cc
```diff
@@ -610,6 +610,91 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T"
+        )
+        .Output(0, "output", "Tensor of rank r >= 1 (same rank as input).", "T")
+        .TypeConstraint(
+            "T",
+            OpSchema::all_tensor_types(),
+            "Input and output types can be of any tensor type.")
+        .TypeConstraint(
+            "Tind",
+            {"tensor(int32)", "tensor(int64)"},
```
houseroad: just int64? I think other ops only support int64 as index, not much sense to support int32 here.

onnx/defs/tensor/defs.cc
```diff
@@ -610,6 +610,91 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(0, "data", "Tensor of rank r >= 1.", "T")
+        .Input(
+            1,
+            "indices",
+            "Tensor of int32/int64 indices, of r >= 1 (same rank as input).",
+            "Tind")
+        .Input(
+            2,
+            "updates",
+            "Tensor of rank r >=1 (same rank and type as input data)",
```
houseroad: Same shape as indices?

onnx/defs/tensor/defs.cc
pk-g: since we are already supporting both int32 and int64 in gather(based on https://github.com/onnx/onnx/blob/master/docs/Operators.md#Gather), and given the fact that scatter can conceptually be viewed as opposite of gather, it may make sense for scatter to support what is already being supported by gather ?

onnx/backend/test/case/node/scatter.py
pk-g: sure, resolved. good point!

onnx/backend/test/case/node/scatter.py
pk-g: resolved.

onnx/defs/tensor/defs.cc
pk-g: Sure, resolved.

pk-g:
@houseroad, thanks for the initial review. Can you please review the updated version so we could move forward with adding this op? thanks.

docs/Changelog.md
```diff
@@ -8873,3 +8873,76 @@ This version of the operator has been available since version 9 of the default O
 <dd>Constrain input and output types to float/int tensors.</dd>
 </dl>
 
+### <a name="Scatter-9"></a>**Scatter-9**</a>
+
+  Given `data`, `updates` and `indices` input tensors of rank r >= 1, write the values provided by `updates` 
+  into `data` along `axis` dimension of `data` (by default outer-most one as axis=0) at corresponding `indices`. 
+  For each entry in `updates`, the target index in `data` is specified by corresponding entry in `indices`
+  for dimension = axis, and index in source for dimension != axis. For instance, in a 2-D tensor case, 
+  data[indices[i][j]][j] = updates[i][j] if axis = 0, or data[i][indices[i][j]] = updates[i][j] if axis = 1.
```
fdwr: Something that was unclear to me is that `i` and `j` are not loop counters in terms of `data` or `output` but rather `indices` and `updates`. This greatly bewildered me when looking at example 2 because it appeared that "data[i][indices[i][j]] = updates[i][j] if axis = 1" would yield out of bounds array accesses. Propose clarifying that...

```suggestion
  data[indices[i][j]][j] = updates[i][j] if axis = 0, or data[i][indices[i][j]] = updates[i][j] if axis = 1, where i and j are loop counters from 0 up to the respective size in `updates` - 1.
```

onnx/defs/tensor/defs.cc
```diff
@@ -610,6 +610,91 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
         }));
 
+static const char* Scatter_ver9_doc = R"DOC(
+Given `data`, `updates` and `indices` input tensors of rank r >= 1, write the values provided by `updates` 
+into `data` along `axis` dimension of `data` (by default outer-most one as axis=0) at corresponding `indices`. 
+For each entry in `updates`, the target index in `data` is specified by corresponding entry in `indices`
+for dimension = axis, and index in source for dimension != axis. For instance, in a 2-D tensor case, 
+data[indices[i][j]][j] = updates[i][j] if axis = 0, or data[i][indices[i][j]] = updates[i][j] if axis = 1.
```
fdwr: Something that was unclear to me is that `i` and `j` are not loop counters in terms of `data` or `output` but rather `indices` and `updates`. This greatly bewildered me when looking at example 2 because it appeared that "data[i][indices[i][j]] = updates[i][j] if axis = 1" would yield out of bounds array accesses. Propose clarifying that...

```suggestion
  data[indices[i][j]][j] = updates[i][j] if axis = 0, or data[i][indices[i][j]] = updates[i][j] if axis = 1, where i and j are loop counters from 0 up to the respective size in `updates` - 1.
```

pk-g: sure, thanks for the proposal.

pk-g:
@houseroad, thanks again for the initial review and feedbacks. I have incorporated the feedback and updated the PR. can you please review the updated PR so we can move forward with adding this op ? thanks!

pk-g:
@houseroad thanks for the review, the CI is now passing, can you please take another look so we can move forward with adding this op ?

onnx/defs/tensor/defs.cc
```diff
@@ -606,6 +606,91 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "Tind",
+            {"tensor(int32)", "tensor(int64)"},
+            "Constrain indices to integer types")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          if (!hasNInputShapes(ctx, 3)) {
+            return;
+          }
+
+          int r = ctx.getInputType(0)->tensor_type().shape().dim_size();
```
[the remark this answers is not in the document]
pk-g: That's right, fixed and added the shape inference tests.

pk-g:
@houseroad sure, updated the PR and applied feedback. can you please take a look so we can move forward ? thanks.

pk-g:
@houseroad just a friendly reminder here, can you please take a look at updated PR so we can move forward with adding this op? thanks.

onnx/defs/tensor/defs.cc
```diff
@@ -606,6 +606,81 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T",
+            OpSchema::all_tensor_types(),
+            "Input and output types can be of any tensor type.")
+        .TypeConstraint(
+            "Tind",
+            {"tensor(int32)", "tensor(int64)"},
+            "Constrain indices to integer types")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          if (hasNInputShapes(ctx, 3)) {
```
houseroad: `hasNInputShapes(ctx, 1)`? Since you just need the shape information of the first input.

pk-g: the other two inputs, indices and updates are not optional for this op. So we should always have three inputs when calling this op.

houseroad: Yeah, they are required inputs. However, their shape information is not required for our shape inference. `hasNInputShapes` checks whether we have necessary information for us to do the shape inference.

pk-g: makes sense, I'll update this to 1 then.

pk-g: just updated the PR with this fix and resolved conflicts.

PR #1519, opened 2018-10-17
add eliminate nop monotone argmax pass

ArmenAg:
fixes #1430

### DIFF
--- a/onnx/common/interned_strings.h
+++ b/onnx/common/interned_strings.h
@@ -2,170 +2,167 @@
 // Adventurous users should note that the APIs will probably change.
 
 #pragma once
-#include <vector>
 #include <stdint.h>
 #include <string>
 #include <unordered_map>
+#include <vector>
 
 namespace ONNX_NAMESPACE {
 
 #define FORALL_BUILTIN_SYMBOLS(_) \
-_(PythonOp) \
-_(CppOp) \
-_(Param) \
-_(Select) \
-_(Return) \
-_(Eval) \
-_(add) \
-_(Add) \
-_(Div) \
-_(Mul) \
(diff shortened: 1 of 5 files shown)


ArmenAg:
@houseroad This is the original optimization passed that triggered the rewrite. We also test the fixed point capability of the new framework. Let me know how it looks.

houseroad:
Yeah, I will take a look this weekend

ArmenAg:
@houseroad I'm not sure why we should handle the default axis any differently. What I did though was to unsure that we have the kaxis using `node->hasAttribute(kaxis)`. Let me know if you had something different in mind.

ArmenAg:
@houseroad It looks like the CI has changed and now I'm getting an error: 
```
/home/travis/build/onnx/onnx/onnx/optimizer/passes/eliminate_nop_monotone_argmax.h:53:8: error: unused parameter ‘graph’ [-Werror=unused-parameter]
   bool runTransform(Node* node, Graph& graph, NodeDestroyType& destroy_current)
```

Unfortunatly since we've taken an OOP approach to building the optimization framework we will have unused variables like the ones above. What do you think we should do about this?

bddppq:
@ArmenAg drop/comment out the parameter name like here https://github.com/onnx/onnx/pull/1547

ArmenAg:
@houseroad @bddppq One of the tests is failing because of an operator check in pytorch unrelated to this PR. Does something need to be fixed on their side?

zrphercule:
@ArmenAg Yes I identified it is because of the recent merge of new upsample_op in onnx. Dont worry, I am going to fix it asap.

PR #1524, opened 2018-10-18
Use vector instead of set to keep the order of the opt passes

houseroad:
Use set is dangerous. The order is important for some passes. Use vector to replace set.

### DIFF
--- a/onnx/optimizer/pass_manager.h
+++ b/onnx/optimizer/pass_manager.h
@@ -35,7 +35,7 @@ class GeneralPassManager : public PassManager {
   std::shared_ptr<PassManagerAnalysis> run(Graph& graph) override;
 
  protected:
-  std::set<std::shared_ptr<Pass>> passes;
+  std::vector<std::shared_ptr<Pass>> passes;
 };
 
 // Exhibits the same behavior as GeneralPassManager but will instead check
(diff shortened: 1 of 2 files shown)


onnx/optimizer/pass_manager.h
```diff
@@ -35,7 +35,7 @@ class GeneralPassManager : public PassManager {
   std::shared_ptr<PassManagerAnalysis> run(Graph& graph) override;
 
  protected:
-  std::set<std::shared_ptr<Pass>> passes;
+  std::vector<std::shared_ptr<Pass>> passes;
```
bddppq: Maybe add a comment saying the order we add the passes matters, so people won't accidentally change it back to set.

houseroad: Sure, will add that

ArmenAg:
My bad. Yes it should be vector. But if we really want to insure the constraint that split is called last, we should implement it the pass manager.

houseroad:
@ArmenAg yeah, we can add some mechanism to enforce some invariance.

ArmenAg:
@houseroad Agreed. This can also break the fixed point optimization, since it might call split multiple times. We could make a new type of pass. Maybe call it something like SingularPass that is guaranteed to only run once throughout the lifetime of the optimization. It would also include a field that constrains where in the optimization pass has to run (beginning, end, any).

houseroad:
@ArmenAg You can add this support in another diff. :-)

PR #1527, opened 2018-10-20
Fix warning in onnx/optimizer/passes/nop.h

yinghai:
It's missing `override`


snnn:
BTW， why graph is not used ? why this function is not const?
If graph is not used, it should generate a warning?

Issue #1538, opened 2018-10-23
Build ONNX with -Wall


Flamefire:
Enabling more warnings will reveal potential bugs. The current "quasi-standard" is `-Wall -Wextra`, which I would highly recommend. Building on CI should happen with `-Werror` then (not only for some warnings as currently is the case)

Minor things that come up with GCC 6.4 are unused variables.

A major bug I just found are unhandled enum values (will open separate PR)

PR #1540, opened 2018-10-23
Fix tensor checker

Flamefire:
Having an initializer of type int16, int8 or uint8 will make the model fail the model-checker although it is valid.

This is caused by missing switch cases which this fixes.

Note that issues like this could be detected automatically by CI if `-Wall -Werror` was used and `default` is not used in the switch handling. I include a commit ("Remove default case...") to do the latter and opened an issue (#1538) for the former.

### DIFF
--- a/onnx/checker.cc
+++ b/onnx/checker.cc
@@ -151,8 +151,10 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& /*ctx*/) {
         break;
 
       case TensorProto::INT32:
-      case TensorProto::UINT8:
+      case TensorProto::INT16:
+      case TensorProto::INT8:
       case TensorProto::UINT16:
+      case TensorProto::UINT8:
       case TensorProto::BOOL:
       case TensorProto::FLOAT16:
       case TensorProto::BFLOAT16:
@@ -235,10 +237,10 @@ void check_attribute(
   // In proto3, when the value to be set is type default value (say 0 for int),
   // used_fields may be 0.
   if (used_fields > 1) {
-    fail_check(
-        "Attribute (name: ",
-        attr.name(),
-        ") should not contain more than one value field.");
+	  fail_check(
+		  "Attribute (name: ",
+		  attr.name(),
+		  ") should not contain more than one value field.");
   }
 
   if (!ctx.is_main_graph()) {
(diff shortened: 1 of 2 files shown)


onnx/checker.cc
```diff
@@ -120,6 +120,7 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& /*ctx*/) {
     }
     return;
   } else {
+    bool isChecked = false;
```
linkerzhang: this is not needed, right?

onnx/checker.cc
```diff
@@ -164,7 +169,8 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& /*ctx*/) {
         check_field(string_data);
         break;
 
-      default:
+    }
+    if (!isChecked) {
```
linkerzhang: use the "default" to check all other unsupported cases should be good. no?

Flamefire: No. As I explained in the commit/PR message `-Wall` includes `-Wswitch` which warns about unhandled enum switch-cases. The problem I fixed here would have not been there (because detected by CI) if `-Wall -Werror` was enabled by CI and `default:` was not used.

It is better to explicitly check all other unsupported cases with e.g.:

```
case TensorProto.Foo:
case TensorProto.Bar:
  break; // Unhandled
```

When now another member is added to the enum, the compiler will warn (and on CI error) that it is not handled. Isn't this a great benefit?

Flamefire:
Hmm, seems (on Windows only???) protobuf does not accept `numpy.bool_` as an `int32` field, but it does for the rest...
Is this a bug in protobuf or shall I convert the numpy integer (Int <=32, UInt <=16) arrays to int32 first?

onnx/checker.cc
Flamefire: This is required to remove the default case

Flamefire:
> Hmm, seems (on Windows only???) protobuf does not accept numpy.bool_ as an int32 field, but it does for the rest...
Is this a bug in protobuf or shall I convert the numpy integer (Int <=32, UInt <=16) arrays to int32 first?

Looking into this I fail to understand **HOW** it passed on any platform. There is a check in place which MUST fail when numpy.bools are passed to the int32 repeated field. Very strange that it does not.

I changed this accordingly and it should now pass on all platforms

Flamefire:
@linkerzhang Any required changes?

PR #1542, opened 2018-10-23
Fuse MatMul and Add into Gemm

vloncar:
Add optimizer pass that fuses MatMul and Add into Gemm, i.e.:
```
Before:
   Z = MatMul(X, Y)
   B = Z + A
 After:
   B = Gemm(X, Y, A)
```
Few notes:
- There doesn't seem to be a `kMatMul` kind, why?
- Had to add default attributes manually, is this ok?


linkerzhang:
This actually tells us MatMul and Gemm are redundant to each other. we may only need one in our op spec? @houseroad @ebarsoum @bddppq @gramalingam

bddppq:
@linkerzhang MatMul can support "batch" semantic [like numpy](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.matmul.html), while Gemm strictly requires 2D inputs.

onnx/optimizer/passes/fuse_matmul_add_bias_into_gemm.h
```diff
@@ -0,0 +1,94 @@
...
+      : PredicateBasedPass(
+            PassType::Fuse,
+            PassEfficiency::Complete,
+            PassOptimizationType::Compute) {}
+  std::string getPassName() const override {
+    return "fuse_matmul_add_bias_into_gemm";
+  }
+  bool patternMatchPredicate(Node* node) override {
+    return node->kind() == kAdd &&
+        strcmp(node->inputs()[0]->node()->kind().toString(), "MatMul") == 0 &&
```
ArmenAg: Instead of doing a string comparison what you should do is add MatMul to interned_strings.h. Here's an example with the dropout op: https://github.com/onnx/onnx/pull/1555/files

vloncar: Thanks Armen, I was wondering if I should do it that way, but I wanted to ask first. I'll make the change.

onnx/test/optimizer_test.py
```diff
@@ -613,6 +613,23 @@ def test_fuse_add_bias_into_conv_squeeze_4d_bias_no_fuse(self):  # type: () -> N
...
+        graph = helper.make_graph(
+            [matmul, add],
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (32, 10)),
+             helper.make_tensor_value_info("Y", TensorProto.FLOAT, (8, 16)),
+             helper.make_tensor_value_info("A", TensorProto.FLOAT, (1, 16))],
+            [helper.make_tensor_value_info("B", TensorProto.FLOAT, (32, 16))]
+        )
+        optimized_model = self._optimized(graph, ["fuse_matmul_add_bias_into_gemm"])
+        print(helper.make_model(graph, producer_name='onnx-test'))
```
ArmenAg: Could you remove this print.

onnx/test/optimizer_test.py
```diff
@@ -613,6 +613,23 @@ def test_fuse_add_bias_into_conv_squeeze_4d_bias_no_fuse(self):  # type: () -> N
         assert optimized_model.graph.node[0].op_type == 'Conv'
         assert optimized_model.graph.node[1].op_type == 'Add'
```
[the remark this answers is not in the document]
vloncar: Sure, I'll add a few tests. Thanks for looking into this.

ArmenAg:
@houseroad LGTM. What do you think?

onnx/optimizer/passes/fuse_matmul_add_bias_into_gemm.h
```diff
@@ -0,0 +1,93 @@
...
+    } else {
+      return false;
+    }
+    // check if bias_shape is compatible
+    int64_t num_el = 1;
+    for (int i = 0; i < static_cast<int64_t>(bias_shape.size()); ++i) {
+      if (bias_shape[i].is_int) {
+        num_el *= bias_shape[i].dim;
+      } else {
+        num_el = -1;
```
houseroad: This line can be removed.

onnx/optimizer/passes/fuse_matmul_add_bias_into_gemm.h
```diff
@@ -0,0 +1,93 @@
+// ATTENTION: The code in this file is highly EXPERIMENTAL.
+// Adventurous users should note that the APIs will probably change.
+
+#pragma once
+
+// Before:
+//   Z = MatMul(X, Y)
+//   B = Z + A
```
houseroad: Would suggest use A = Z + Bias, easier to read

onnx/test/optimizer_test.py
```diff
@@ -653,6 +653,69 @@ def test_fuse_add_bias_into_conv_squeeze_4d_bias_no_fuse(self):  # type: () -> N
...
 
+    def test_fuse_matmul_add_bias_into_gemm(self):  # type: () -> None
+        matmul = helper.make_node("MatMul", ["X", "Y"], ["Z"])
+        add = helper.make_node("Add", ["Z", "A"], ["B"])
+        graph = helper.make_graph(
+            [matmul, add],
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (32, 10)),
```
houseroad: This shape of X and Y does not match for matmul...

onnx/optimizer/passes/fuse_matmul_add_bias_into_gemm.h
```diff
@@ -0,0 +1,93 @@
...
+    // try to get feature M from weight_shape
+    if (static_cast<int64_t>(weight_shape.size()) == 2 &&
+        weight_shape[1].is_int) {
+      M = weight_shape[1].dim;
+    } else {
+      return false;
+    }
+    // check if bias_shape is compatible
+    int64_t num_el = 1;
+    for (int i = 0; i < static_cast<int64_t>(bias_shape.size()); ++i) {
```
houseroad: Here, to be safe, instead of checking num_el, I would suggest checking len(dim) == 1 or 2, and dim[size-1] == M
Otherwise you will get incorrect output shape.

vloncar:
I believe the latest changes address the issues raised. Is there anything else you would like me to change?

I also have one general question. Should the optimizer defend against invalid models (i.e., check the nodes it works on and bail out if they are invalid, or assume that they are valid and only check what it need to perform the optimization)? For example, how should optimizer handle wrong shapes, missing required attributes, invalid attribute values etc? I am working on another optimizer which, while simple, has many cases where it would fail if the model is invalid.

vloncar:
On second thought (and after noticing that the shape of `MatMul`'s X input may not be available), I have simplified the bias shape check, to do what @houseroad recommended.

ArmenAg:
@houseroad Correct me if I'm wrong but the optimization framework should operate under the assumption that the model is constructed correctly. Otherwise this puts to much responsibility on the optimization framework. Either way we have a model checker that we can run before doing optimization.

houseroad:
@ArmenAg the principle here is that: having basic check to make sure we are not running into segmentation fault, and leave complete check to the checker. (Even the model is wrong, I think the optimizer should not give segfault)

vloncar:
@houseroad Can you review the latest changes? I believe the current code satisfies the principle you outlined above.

onnx/test/optimizer_test.py
```diff
@@ -653,6 +653,85 @@ def test_fuse_add_bias_into_conv_squeeze_4d_bias_no_fuse(self):  # type: () -> N
...
+
+    def test_fuse_matmul_add_bias_into_gemm_2d_bias(self):  # type: () -> None
+        matmul = helper.make_node("MatMul", ["X", "Y"], ["Z"])
+        add = helper.make_node("Add", ["Z", "B"], ["A"])
+        graph = helper.make_graph(
+            [matmul, add],
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (32, 10)),
+             helper.make_tensor_value_info("Y", TensorProto.FLOAT, (10, 16)),
+             helper.make_tensor_value_info("B", TensorProto.FLOAT, (32, 16))],
```
houseroad: please add two more cases:
1) B ==> 1, 16
2) X = (1, 10), Y = (10, 16), B ==> 16, 16

onnx/optimizer/passes/fuse_matmul_add_bias_into_gemm.h
```diff
@@ -0,0 +1,97 @@
...
+    } else {
+      return false;
+    }
+    // check if bias_shape is compatible
+    auto bias_shape = orig_bias->sizes();
+    auto bias_dim = static_cast<int64_t>(bias_shape.size());
+    int64_t bias_M = -1;
+    if (bias_dim == 1 && bias_shape[0].is_int) {
+      bias_M = bias_shape[0].dim;
+    } else if (bias_dim == 2 && bias_shape[1].is_int) {
```
houseroad: there is still a problem here. Since in GEMM we only broadcast the bias to the A*B. so we should check here whether bias's second dimension is 1 or the same as A*B.

onnx/test/optimizer_test.py
vloncar: Done.

onnx/optimizer/passes/fuse_matmul_add_bias_into_gemm.h
vloncar: I had the stricter check in the previous version, but thought it was too much. I have reintroduced it now.

PR #1548, opened 2018-10-24
Add hook to InferenceContext to allow running type/shape inferencing …

skottmckay:
…for a sub-graph represented by a GraphProto attribute in the current node.

Split out from PR 1503 so this InferenceContext interface change can be discussed separately.


onnx/shape_inference/implementation.h
```diff
@@ -76,10 +105,47 @@ struct InferenceContextImpl : public InferenceContext {
...
+    if (entry == graphAttributeInferers_.cend()) {
+      // create GraphInferencer instance
+      auto attrNameToGraphProto = graphProtoAttributesByName_.find(attr_name);
+      if (attrNameToGraphProto == graphProtoAttributesByName_.cend()) {
+        fail_type_inference(
+            "Attribute ", attr_name, " does not contain a graph.");
+      }
+
+      std::unique_ptr<GraphInferencer> new_inferer =
+          std::make_unique<GraphInfererImpl>(
```
bddppq: `std::make_unqiue` is available starting from c++14, while onnx needs to support c++11. Use the backports here instead https://github.com/onnx/onnx/blob/master/onnx/common/stl_backports.h (`ONNX_NAMESPACE::make_unqiue`).

onnx/defs/shape_inference.h
```diff
@@ -6,6 +6,13 @@
 
 namespace ONNX_NAMESPACE {
 
+class GraphInferencer {
+  // Perform inferencing on the graph contained in GraphInferencer.
+  // Returns the graph output types post-inferencing.
+  virtual std::vector<const TypeProto*> doInferencing(
```
linkerzhang: Thanks a lot for this change! Please also consider the case of using input data to do shape inference in this interface.

skottmckay: Added.

linkerzhang:
@bddppq  @anderspapitto  @houseroad  any more comments on this API change please? this is for the attribute graph type and shape inference.

skottmckay:
It's used by the Scan inferencing in https://github.com/onnx/onnx/pull/1503 so that covers some aspects of it. As we don't use the implementation.cc version directly I wasn't sure how or where to add explicit tests for that. The logic there replicates what we do in our inferencing code, using existing pieces from implementation.cc/h. I didn't see any existing test setup for testing implementation.cc though. Did I miss it?

linkerzhang:
how about we merge this PR to unblock PR #1503 and test could also be added for scan shape inference in #1503. @houseroad

PR #1553, opened 2018-10-25
Fix output type bug in MaxUnpool definition.

spandantiwari:
The type of output tensor should be type of the first input tensor (the pooled input image). This change fixes it to be so.


spandantiwari:
An unrelated test (Upsample) seems to be failing.

spandantiwari:
@bddppq - Thanks!

PR #1554, opened 2018-10-25
Introduce SparseTensor ML proto

yuslepukhin:


yuslepukhin:
Cc: @gramalingam

gramalingam:
I think introducing SparseTensor into ONNX main spec would be useful eventually. This will help us experiment with it first.

gramalingam:
The Travis CI failure appears to be a stalled build. What is the protocol here? Should we try to redo all builds? This seems to have the disadvantage that it will potentially slow down other builds, triggering a cascade of such issues. Or, go ahead and merge it? If so, when is it okay to go ahead and merge it?

PR #1555, opened 2018-10-26
Optimization Pass: Remove Dropout if Ratio == 0.0

### LINKED ISSUE #1521: Optimization Pass: Dropout with ratio 0.0 should be a Nop

ArmenAg:
fixes #1521

### DIFF
--- a/onnx/optimizer/passes/eliminate_nop_dropout.h
+++ b/onnx/optimizer/passes/eliminate_nop_dropout.h
@@ -1,0 +1,36 @@
+// ATTENTION: The code in this file is highly EXPERIMENTAL.
+// Adventurous users should note that the APIs will probably change.
+
+#pragma once
+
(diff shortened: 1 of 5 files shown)


onnx/optimizer/passes/eliminate_nop_dropout.h
```diff
@@ -0,0 +1,36 @@
...
+            PassEfficiency::Complete,
+            PassOptimizationType::Compute) {}
+
+  std::string getPassName() const override {
+    return "eliminate_nop_dropout";
+  }
+
+  bool patternMatchPredicate(Node* node) override {
+    return (node->kind() == kDropout && node->hasAttribute(kratio)) &&
+        node->f(kratio) == 0.0;
```
houseroad: Even it's not 0, in inference, dropout can be considered as a non-op.

PR #1557, opened 2018-10-26
implement fusing of consecutive concats

### LINKED ISSUE #1556: Optimization Pass: Fusing consecutive concats
I was looking at an exported model at work and saw the following behavior:

image

In this case we can replace the 3 concats with a single concat along axis=0. I saw this type of pattern in multiple places of the DAG and it seems to be a biproduct of pytorch exportation of GRU (or maybe all RNN) ops.

ArmenAg:
fixes #1556


ArmenAg:
Here is a example of the optimization: 

**Before:**
image

**After:**
image

ArmenAg:
@houseroad Looks like all the tests passed!

Issue #1531, opened 2018-10-21
incorrect ConvTranspose Bias shape


fumihwh:
The current spec `Optional 1D bias to be added to the convolution, has size of C.` maybe incorrect.
Pytorch: https://pytorch.org/docs/stable/nn.html#torch.nn.ConvTranspose2d
Tensorflow: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/layers/convolutional.py#L839
Bias should have size `M` (say feature maps or filters or out channel) not `C`

houseroad:
Fixed by @fumihwh , thanks!

PR #1566, opened 2018-10-31
Fix conv transpose spec

fumihwh:
#1531


houseroad:
CI failure is unrelated

PR #1567, opened 2018-10-31
Add OneHot op to ONNX.

spandantiwari:
This PR adds a new op, `OneHot`, to ONNX spec. Representing an input as one-hot coded vector(s) is a common operation that's useful in many scenarios.


onnx/defs/tensor/defs.cc
```diff
@@ -1185,4 +1185,172 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(
+            1,
+            "depth",
+            "Scalar specifying the number of classes in one-hot tensor. This is also the size "
+            "of the one-hot dimension (specified by 'axis' attribute) added on in the output "
+            "tensor and the values in the 'indices' input tensor are expected to be "
+            "in the range [0, depth). The"
+            "In case 'depth' is of non-integer type, it will be casted to int64 before use.",
+            "T1")
+        .Input(
```
linkerzhang: how about we remove attribute "dtype" and have this input as required? looks to me it's more simple.

spandantiwari: @linkerzhang - Yes, we could do that. A minor reason to keep it this way could be the situation in packages such as TF, when `values` are available in a different datatype (from prior computation) than the type specified by the explicit `dtype` argument. In such a situation the exporter will have to insert an additional `Cast` node before `OneHot`. Having `dtype` in the ONNX op may provide some convenience and eliminate the need for the extra op.
This is a minor point and I am OK either way. What do you think?

Does anyone else have any preference/feedback?

linkerzhang: Hmmm, agree this is a minor point. but don't quite understand the case you said. "The type of 'values' should be the same as 'dtype', if 'dtype' is specified." this means anyway, the values's type is indeed indicating output's type, if values are always specified, then output type is also specified (no datatype attribute needed). If values are from previous node ('s output) and the type is not expected, then a CAST is anyway needed, am I right?

spandantiwari: Yes, you are right, and the `Cast` op is needed anyway. I agree that the `dtype` attribute is not really adding much and just making `values` a required input is a cleaner design. Thanks for pointing this out. Will update the signature.

spandantiwari: @linkerzhang as per your suggestion, I have removed `dtype` and made `values` a required input.

spandantiwari:
@houseroad, @bddppq - Gentle ping for your review.

onnx/defs/tensor/defs.cc
```diff
@@ -1185,4 +1185,122 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            OpSchema::all_tensor_types(),
+            "Constrain to any tensor type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          // Check that the node has two or three onputs.
+          if (!hasNInputShapes(ctx, 3)) {
+            fail_shape_inference(
+                  "OneHot node must have three inputs.");
+          }
+          // If input 'values' is specified, it must be a two-element vector.          
+          auto& input_shape = getInputShape(ctx, 2);
```
houseroad: nit: input2_shape / value_shape

onnx/defs/tensor/defs.cc
```diff
@@ -1185,4 +1185,122 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            OpSchema::all_numeric_types(),
+            "Constrains input to only numeric types.")
+        .TypeConstraint(
+            "T2",
+            OpSchema::all_tensor_types(),
+            "Constrain to any tensor type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          // Check that the node has two or three onputs.
+          if (!hasNInputShapes(ctx, 3)) {
+            fail_shape_inference(
```
houseroad: Probably not throwing exception, just return to stop the shape inference.

onnx/defs/tensor/defs.cc
```diff
@@ -1185,4 +1185,122 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                  "OneHot node must have three inputs.");
+          }
+          // If input 'values' is specified, it must be a two-element vector.          
+          auto& input_shape = getInputShape(ctx, 2);
+          if (input_shape.dim_size() != 1) {
+            fail_type_inference(
+                  "Input 'values' must be rank 1 tensor.");
+          }
+          if (!input_shape.dim((int)0).has_dim_value() ||
+            (input_shape.dim((int)0).has_dim_value() &&
```
houseroad: ditto, if we don't have enough information, just exit the shape inference.

onnx/defs/tensor/defs.cc
```diff
@@ -1185,4 +1185,122 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          auto* output_shape = getOutputShape(ctx, 0);
+          for (int i = 0; i < out_rank; ++i) {
+            auto* dim = output_shape->add_dim();
+            if (i < axis) {
+              if (indices_shape.dim(i).has_dim_value()) {
+                dim->set_dim_value(indices_shape.dim(i).dim_value());
+              } else if (indices_shape.dim(i).has_dim_param()) {
+                dim->set_dim_param(indices_shape.dim(i).dim_param());
+              }
+            }
```
houseroad: how about i == axis case?

onnx/defs/tensor/defs.cc
```diff
@@ -1185,4 +1185,122 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    OneHot,
+    9,
+    OpSchema()
+        .SetDoc(OneHot_ver9_doc)
+        .Attr(
+            "axis",
+            "(Optional) Axis along which one-hot representation in added. Default: axis=-1.",
```
houseroad: Let's mention axis=-1 means the last dimension.

spandantiwari: Done.

onnx/defs/tensor/defs.cc
spandantiwari: Agreed. I have updated the shape inference at multiple places to return whenever there isn't enough info, instead of throwing.

onnx/defs/tensor/defs.cc
spandantiwari: Updated.

onnx/defs/tensor/defs.cc
spandantiwari: Agreed. Updated shape inference code at multiple places to do that.

onnx/defs/tensor/defs.cc
spandantiwari: In the `i == axis` case the size of that dimension comes from the `depth` input, which may not be known or inferrable. Therefore, for that case, we just add a dimension but do not add a value. Also, added test points for this scenario.

spandantiwari:
@houseroad - Thanks for your feedback. I have overhauled the shape inference code based on your feedback. Also added a couple of test points in `shape_inference_test.py` to test the code.

spandantiwari:
@houseroad - I have updated based on your comments. All the checks are passing. Could you please review? Thanks.

linkerzhang:
Thank you very much! @spandantiwari

spandantiwari:
@linkerzhang and @houseroad - Thanks!

Issue #870, opened 2018-05-02
More operators: where, >= and <=?


zasdfgbnm:
How about adding the following three new operators?
`Where`, `GreaterOrEqual`, `LessOrEqual`

Would these be helpful?

zasdfgbnm:
Can anyone make any comment on this?  I just need to know if they are contribute welcome. These ops are useful to me, and if it is contribute welcome, I can easily add it as I did in https://github.com/onnx/onnx/pull/869

For the `GreaterOrEqual` and `LessOrEqual`, we already have `Greater` , `Less`, `Equal`, and `Or`, but it is not a good idea to express `>=` as `> or =`, since many ISAs support `>=` instructions natively.

Also for the `Where`, it is also widely supported:
https://www.tensorflow.org/api_docs/python/tf/where
https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.where.html
https://pytorch.org/docs/master/torch.html#torch.where
http://docs.chainer.org/en/stable/reference/generated/chainer.functions.where.html
https://mxnet.incubator.apache.org/api/python/symbol/symbol.html#mxnet.symbol.where

bddppq:
For `GreaterOrEqual` and `LessOrEqual`, since as you have pointed out they can be composed by existing ops, so I feel it's better to not add them (or we could use the newly added `Function` to introduce these composed ops). @linkerzhang 
For `Where`, this can be implemented as `If` with the two graph attributes only containing one single Constant Node. @jamesr66a @zdevito

zasdfgbnm:
@bddppq Thanks a lot for your quick reply!

As a user, I'm totally happy with the solution of composing `GreaterOrEqual` and `LessOrEqual` by existing ops. It might just be a bit annoying for DL framework developers, e.g. they might need to write code to optimize `a > b or a = b` or `not a < b` to `a >= b` to turn 2 tensor operations into 1 for better performance.

For the `Where`, your reply completely sounds good to me. Maybe I should go to my framework and implement the onnx export part this way.

zasdfgbnm:
@bddppq Can we really implement `Where` as a condition plus two constant node? I don't understand the document of `If` well, but to me it looks like the `cond` should be a bool tensor that contains only one element, and there is no broadcasting at all for the condition and two graphs.

kit1980:
[...]

PR #1569, opened 2018-10-31
Add Where op.

kit1980:


kit1980:
@zrphercule, there is an issue suggesting using If: https://github.com/onnx/onnx/issues/870
But as I mentioned there, it's not possible to use it for Where because you don't just select one of the output tensors, you select some elements from the first tensor, and some elements from the second tensor. And if there is only one input, Where returns indices of true elements.

I'll update with regenerated docs and test data.

liqunfu:
Please update onnx/defs/operator_sets.h with the new op

gramalingam:
The two usage scenarios (one where X and Y are specified and one where X and Y are not specified) seem different. Why bundle them into one op? I understand numpy does it, but should we duplicate the same? Why not make these different ops? It is easier to describe/implement/understand when they are separate.

kit1980:
@gramalingam Tensorflow also has the dual-intent Where as numpy. With two different ops, we'll need to invent some non-standard name for one variant.

gramalingam:
Ok. I think it would help to clarify in the documentation what happens when X and Y are not specified: e.g., "returns the indices of positions in the first tensor where the value is 1, as a 2 dimensional tensor where the first index is … and second index is ...".  It looks like the shape you specify for output has the two dimensions in the opposite order to what is in tensorflow. Adding a shape inference function for the op would also be helpful.

kit1980:
Modified the Where op spec to accept only three arguments (the one argument case is to be handled by NonZero). Also added shape and type inference.

kit1980:
A separate PR for NonZero: https://github.com/onnx/onnx/pull/1714

kit1980:
Can someone look at this? The Where Op is much simpler now.

kit1980:
Rebased on master.

PR #1573, opened 2018-11-01
Add attributes to scan op

gramalingam:
Add two new attributes to the scan op:
(a) One to specify whether scan_outputs should be constructed in forward or reverse direction, and
(b) One to specify whether the scan op should explicitly handle batch (axis 0) or treat it as unbatched.
Update shape inference accordingly.


gramalingam:
Hi @bddppq @linkerzhang , any feedback on this update?. Thanks

gramalingam:
Updated PR appears in https://github.com/onnx/onnx/pull/1653 .

PR #1574, opened 2018-11-02
Add a virtual destructor to GraphInferencer

snnn:
In ./onnx/shape_inference/implementation.h:
We have
```
      std::unique_ptr<GraphInferencer> new_inferencer{new GraphInferencerImpl(...)}
```
Therefore, the destructor of GraphInferencer must be virtual.


skottmckay:
Nice catch. Would be great it some static analysis could catch these things earlier.

rdzhabarov:
We could add "-Wnon-virtual-dtor" compilation flag to detect missing virtual destructor when there is a virtual method declared for the class.

PR #1578, opened 2018-11-02
Fix MaxUnpool shape inference when output_shape is provided as input

spandantiwari:
Currently there is a bug in shape inference for `MaxUnpool` op. Specifically, the problem happens when `output_shape` input is provided. In this case, the current code checks (wrongly) the `dim_size` for `output_shape` with the `dim_size` for first input tensor. The right thing to do is to match the number of elements in `output_shape` (which must be rank 1) to the `dim_size` of first input tensor. 

This PR fixes that. I have also included two test points in `shape_inference_test.py` for this scenario.

### DIFF
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -295,13 +295,15 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain index tensor to int64"));
 
 void maxUnpoolShapeInference(InferenceContext& ctx) {
-  propagateElemTypeFromInputToOutput(ctx, 0, 0);
-
   // we need at least two inputs to have a shape for this inference.
-  if (!hasNInputShapes(ctx, 2)) {
-    fail_shape_inference("MaxUnpool op must have atleast 2 inputs.");
+  if (ctx.getNumInputs() != 2 && ctx.getNumInputs() != 3) {
+    fail_type_inference(
+            "MaxUnpool op must have either two or three inputs.");
   }
-
+  propagateElemTypeFromInputToOutput(ctx, 0, 0);
+  if (!hasInputShape(ctx, 0)) {
+    return; // If first input does not have shape, we cannot infer much.
+  }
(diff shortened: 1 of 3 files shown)


onnx/defs/nn/defs.cc
```diff
@@ -340,12 +340,22 @@ void maxUnpoolShapeInference(InferenceContext& ctx) {
...
-    if (output_shape.dim_size() != input_shape.dim_size()) {
-        fail_shape_inference("output_shape must be the same size as the shape of input tensor X.");
+    auto& output_shape = getInputShape(ctx, 2);
+    if (output_shape.dim_size() != 1) {
+      fail_type_inference(
+            "'output_shape' must be rank 1 tensor.");
     }
-    else
-      return; // output_shape is specified as input. Actual shape will be detrermined at runtime.
+    if (!output_shape.dim((int)0).has_dim_value()) {
```
gramalingam: I think this check-and-fail is not needed … it is valid if the input-dimension is not statically known

onnx/defs/nn/defs.cc
```diff
@@ -340,12 +340,22 @@ void maxUnpoolShapeInference(InferenceContext& ctx) {
   if (hasNInputShapes(ctx, 3)) {
```
gramalingam: This check is not the same as checking if the optional 3rd input is specified. If the optional 3rd input is specified, we will need to return, I believe.

onnx/defs/nn/defs.cc
spandantiwari: Fixed.

onnx/defs/nn/defs.cc
spandantiwari: Fixed.

spandantiwari:
@gramalingam - I have overhauled the shape inference code based on your feedback.

spandantiwari:
@gramalingam - Thanks.

PR #1580, opened 2018-11-05
Fuse Pad into Conv optimizer

vloncar:
This is an optimizer that fuses Pad operation into Conv.
```
Before:
  P = Pad(X)
  Z = Conv(P, Y)
After:
  Z = Conv(X, Y) with "pads" attribute set
```
What would be the best way to avoid working with strings directly (lines 50 and 59) in this case? If i put the value in `BuiltinSymbol` enum I can't compare it with string value of the attribute. Or am I missing something?


onnx/optimizer/passes/fuse_pad_into_conv.h
```diff
@@ -0,0 +1,99 @@
...
+            PassType::Fuse,
+            PassEfficiency::Complete,
+            PassOptimizationType::Compute) {}
+  std::string getPassName() const override {
+    return "fuse_pad_into_conv";
+  }
+  bool patternMatchPredicate(Node* node) override {
+    return node->kind() == kConv && node->inputs()[0]->node()->kind() == kPad;
+  }
+  bool runTransform(Node* n, Graph& graph, NodeDestroyType& destroy_current)
```
zrphercule: It seems the "graph" here has not been used, and that's why the CI failed.
Wonder if it is designed to be like this?

vloncar: I noticed and fixed in the latest commit, but I don't understand how is this not an issue in other optimizers, like `fuse_consecutive_squeezes` or `eliminate_nop_pad`?

ArmenAg: @zrphercule I believe that CI checks for unused parameters.

@vloncar both fuse_consecutive_squeezes and eliminate_nop_pad have definitions which don't use graph:

`bool runTransform(Node* n, Graph&, NodeDestroyType& destroy_current)`

vloncar:
Is there any hope for getting this (and #1542) optimizer into ONNX or should we handle things like this in our downstream tools? Both this PR and #1542 came from working with models exported to ONNX format in a standard way (torch.onnx.export for Pytorch and onnxmltools for Keras, respectively) so they should be useful.

ArmenAg:
@houseroad Could you please review the two passes that @vloncar has written. Both could be very useful to us.

houseroad:
@vloncar yeah, I also reviewed the other one, could you address my comments?

vloncar:
@houseroad Thanks a lot!

PR #1582, opened 2018-11-06
Add ConstantOfShape op

zrphercule:
Still in testing...


linkerzhang:
Several questions:
1. looks to me the constantlike op does meet all requirements, isn't it? @spandantiwari  may comment more on this.
2. A production operator can't be removed like this. It should be deprecated in next operator set version, since there may be models replying on it.

zrphercule:
@linkerzhang Thanks for your comments!
My comments:
1. ConstantLike has useful functionality, however its design is a little confusion and different from other similar operators. It takes both optional input and optional attribute to indicate the shape of output, which we believe is not a good design.
2. ConstantLike was added a 3~4 weeks ago in version 9, which means it is still not released yet. I believe we may be able to modify it before we release it?
Thanks!

spandantiwari:
@zrphercule 
1) Yes, the design could be more streamlined, but it was created intentionally so as try to support features from the experimental op `ConstantFill`. `ConstantFill` is used in export by PyTorch and requires support. In fact, there's a PR still open (https://github.com/onnx/onnx/pull/1434) for removing `ConstantFill`, but it is on hold till PyTorch migrates away from `ConstantFill`.
Also, the name design of existing `ConstantLike` is largely in line with design of similar generator ops in ONNX, e.g. `RandomUniform` and `RandomUniformLike`, and `RandomNormal` and `RandomNormalLike`, not to mention similar Numpy APIs, e.g. `zeros_like`, `ones_like`. If we want to streamline the signature of `ConstantLike`, we should consider improving the existing op, because the proposed name `ConstantAsShape` is quite different from the existing convention. 

2) Regarding this part, I agree that it isn't the best idea to remove productions ops without deprecation. `ConstantLike` was committed more than two months ago. Given the fast release times, this is already part of frameworks' exporters.

spandantiwari:
Circling back on this. Having looked at `torch.zeros` I understand that there're use cases when the size has to be provided as input on the fly. We should support both use cases:
1) Taking shape (and possibly type) info from the shape of the input tensor. This is because:
a) There are framework APIs that support this directly, so having an analogous op is convenient, because it eliminates the need for inserting multiple nodes such as shape.
b) We already similar generator ops such as `RandomUniformLike` and `RandomUniformLike`.
2) Taking shape directly as an input. This will satisfy the torch.zeros type use case. 

One design is update `ConstantLike` to support both use cases. Another option is to support these two in two separate ops. Both have their pros and cons. Thoughts?

spandantiwari:
@zrphercule - Having looked at this in detail and discussing with others, I am OK with moving forward with this spec. The question is whether should we keep `ConstantLike` also, or should we deprecate it. I can see argument for both. The arguments for keeping `ConstantLike` are similar to the reasoning in the other open PR for `LessOrEqual` and `GreaterOrEqual`. It is a common op seen in many frameworks such as CNTK and TF, and it lends to better optimization. But at the same time, we are increasing the number of ops by having multiple ops. What do you think? Shall we keep `ConstantLike`.

onnx/defs/generator/defs.cc
```diff
@@ -33,69 +33,36 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-             "tensor(int64)",
-             "tensor(uint8)",
+            {"tensor(uint8)",
              "tensor(uint16)",
              "tensor(uint32)",
-             "tensor(uint64)",
-             "tensor(bool)"},
-            "Constrain input types. Strings and complex are not supported.")
+             "tensor(uint64)"},
+            "Constrain input types. Shape must be unsigend integers.")
```
spandantiwari: nit: spelling "unsigned"

onnx/defs/generator/defs.cc
```diff
@@ -33,69 +33,36 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-
-The 'dtype' argument must be one of the data types specified in the 'DataType' enum field in the
-TensorProto message and be valid as an output type.
+static const char* ConstantAsShape_ver9_doc = R"DOC(
+Generate a tensor with given value and shape.
 )DOC";
 
 ONNX_OPERATOR_SET_SCHEMA(
-    ConstantLike,
+    ConstantAsShape,
```
spandantiwari: Since the shape is an input to this op, may I suggest `ConstantOfShape` as possible name for this op.

onnx/defs/generator/defs.cc
```diff
@@ -33,69 +33,36 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-        .Attr(
-            "shape",
-            "(Optional) The shape of the output tensor. If input tensor T1 is provided, then"
-            " 'shape' attribute is ignored and the output follows the shape of the input."
-            " One of either input tensor T1 or 'shape' attribute must be provided.",
-            AttributeProto::INTS,
-            OPTIONAL)
+            "(Optional) The value of the output elements."
+            "Should be a one-element tensor, "
+            "If 'dtype' is not given, the datatype of "
```
spandantiwari: Did you mean: "The value for this attribute defaults to 0."?

onnx/defs/generator/defs.cc
```diff
@@ -33,69 +33,36 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            "Input tensor to copy shape, and optionally, type information from."
-            " One of either input tensor T1 or 'shape' attribute must be provided.",
-            "T1",
-            OpSchema::Optional)
-        .Output(
-            0,
-            "output",
-            "Output tensor, same shape as input tensor T1.",
-            "T2")
+            "1D tensor, The shape of the expected output tensor.",
```
spandantiwari: Can we clarify if we can create a scalar from this (empty tensor)?

onnx/defs/generator/defs.cc
```diff
@@ -110,27 +77,22 @@ ONNX_OPERATOR_SET_SCHEMA(
              "tensor(uint32)",
              "tensor(uint64)",
              "tensor(bool)"},
-            "Constrain output types. Strings and complex are not supported.")
+            "Constrain output types")
```
spandantiwari: Do we need to constrain output types?

zrphercule:
@spandantiwari Sorry I missed you prior message... Yeah the reason I didnt push this spec anymore is I also agree it might be a good thing to keep ConstantLike, while ConstantAsShape should be added or not is still under discussing.
I wonder if you are willing to use ConstantLike or the new ConstantAsShape? Feel free to use ConstantLike since I dont think I will do any breaking commit recently.

spandantiwari:
@zrphercule @houseroad - can we move to include this op (`ConstantAsShape`) in the spec. I will be happy to help land this.

houseroad:
@spandantiwari unfortunately, @zrphercule is sick this week, feel free to take over the PR, I can help on the review.

zrphercule:
@spandantiwari Sorry I was having a fever while you ping me. Now I am back, and am working on this. Should be fast, no worries. Thanks!

onnx/defs/generator/defs.cc
zrphercule: Since we have the same concern in ConstantLike... I am ok with removing this constrain.

onnx/defs/generator/defs.cc
zrphercule: I guess we can? "If empty tensor provided, then output tensor would be a scalar"

onnx/defs/generator/defs.cc
spandantiwari: @zrphercule - Minor point - this probably needs some clarity. I think we want the datatype of the output tensor to be the same as the type of one-element `value` tensor, correct? If yes, then this can be worded better. For one, there's no `dtype` attribute.

spandantiwari:
@zrphercule - No worries. Hope you are feeling better. 

You have addressed most of the feedback and the  spec looks good to me. I have added just one comment on the description of the `value` attribute. If you take a look that'll be great.

onnx/defs/generator/defs.cc
zrphercule: Oh yes, but the value tensor is optional. Let's say if value is not given, then by default the output dtype is float32.

spandantiwari:
@zrphercule - I may have mentioned this earlier but missed in my later reviews. Could you please add a shape inference also to the `TypeAndShapeInferenceFunction` for `ConstantOfShape`? Right now it is doing type inference only. Thanks!

zrphercule:
@spandantiwari  sure, no problem!

spandantiwari:
@zrphercule - Actually, if you haven't started on it yet, let me do it. I am creating a PR to remove `ConstantLike` anyway and I will add shape inference to `ConstantOfShape` as well. Is that OK?

zrphercule:
@spandantiwari Thanks! If you have any problem, feel free to ping me!

spandantiwari:
@zrphercule - Sure. I have created a PR https://github.com/onnx/onnx/pull/1716 that removes `ConstantLike` and also adds shape inference to `ConstantOfShape`. Please take a look and let me know your feedback. Thanks!

PR #1587, opened 2018-11-07
Clarify some aspects of the Loop spec.

skottmckay:
* Mark first two inputs as optional to match the description
 * Use 'tensor(bool)' and 'tensor(int64)' as type constraints instead of 'bool' and 'int64'.
   * Utils::DataTypeUtils::ToType converts bool/int64 to tensor(bool)/tensor(int64) 
   * Using tensor(bool)/tensor(int64) is consistent will other ops that expect scalars (e.g. Size)


houseroad:
@skottmckay do you mind updating the PR to pass the CI?

PR #1588, opened 2018-11-07
Add domain as an optional parameter for make_node function

youngkim93:
Currently make_node function in helper.py does not take in domain as a parameter. Without this change if we specify domain as a keyword argument for this helper it will treat "domain" as an attribute, not as a field of NodeProto. This change will allow assigning domain field when creating a NodeProto with make_node helper function.


linkerzhang:
Please make Linux ci happy, btw.

houseroad:
Looks good, can you add/modify one test case to cover this change? Thanks

You can find existing test here: https://github.com/onnx/onnx/blob/master/onnx/test/helper_test.py#L174

onnx/helper.py
```diff
@@ -33,6 +34,7 @@ def make_node(
         outputs (list of string): list of output names
         name (string, default None): optional unique identifier for NodeProto
         doc_string (string, default None): optional documentation string for NodeProto
+        domain (string, default None): optional domain for NodeProto
```
houseroad: If it's none, we will just use default domain (which is empty)

onnx/helper.py
```diff
@@ -45,6 +48,8 @@ def make_node(
         node.name = name
     if doc_string:
         node.doc_string = doc_string
+    if domain:
```
bddppq: You need to explicitly check for `None` here, otherwise here will skip setting the domain when it's empty string

PR #1591, opened 2018-11-07
Loop type shape inferencing

skottmckay:
Add type/shape inferencing for the subgraph in the Loop 'body' attribute.


onnx/defs/controlflow/defs.cc
```diff
@@ -272,6 +272,107 @@ void IfInferenceFunction(InferenceContext& ctx) {
...
+  // loop state values get propagated to outputs
+  for (size_t i = 2; i < num_inputs; ++i) {
+    auto input_type = ctx.getInputType(i);
+    subgraph_input_types.push_back(input_type);
+
+    propagateElemTypeFromInputToOutput(ctx, i, i - 2);
+
+    if (hasInputShape(ctx, i)) {
+      propagateShapeFromInputToOutput(ctx, i, i - 2);
+    }
```
gramalingam: My reading of the op documentation is that the scan-outputs are required to have a shape that is invariant across the loop iterations … but I don't see a similar restriction imposed on the loop-state variables. (We do have such a restriction in the case of "scan"). I would drop the above 3 lines (for loop-state vars) unless the op. spec is changed.

gramalingam: Likewise the type pushed into subgraph_input_types should be a copy of the type with shape information omitted, for the same reason, since this shape could vary over iterations.

onnx/defs/controlflow/defs.cc
```diff
@@ -272,6 +272,107 @@ void IfInferenceFunction(InferenceContext& ctx) {
   }
 }
 
+void LoopInferenceFunction(InferenceContext& ctx) {
+  auto num_inputs = ctx.getNumInputs();
+  auto num_loop_state_vars = num_inputs - 2; // skip 'M' and 'cond'
+
+  std::vector<const TypeProto*> subgraph_input_types;
+  subgraph_input_types.push_back(nullptr); // iteration number. not inferred.
```
gramalingam: Adding the expected numeric type would catch a type-error where the subgraph's first parameter is not the right type. (The spec doesn't clarify the type, but int64 seems best, and we could probably clarify the spec too.)

onnx/test/shape_inference_test.py
```diff
@@ -1053,6 +1049,41 @@ def test_onehot_with_axis(self):  # type: () -> None
             [])
         self._assert_inferred(graph, [make_tensor_value_info('Y', TensorProto.FLOAT, (2, None, 3, 5))])  # type: ignore
 
+    def test_loop(self):    # type: () -> None
+        # can't use self._make_graph for the subgraph as it add more inputs for the Reshape operations it inserts.
+        # this breaks the subgraph inferencing as it expects the number of inputs passed from Loop to match
+        # the GraphProto, but Loop knows nothing about the additional inputs.
+        input_value_infos = [make_tensor_value_info('iter_num_in', TensorProto.FLOAT, (1,)),
```
gramalingam: Wonder if this is meant to be float or int?

skottmckay: INT64 is more correct. Doesn't play any role in the inferencing though.

onnx/defs/controlflow/defs.cc
skottmckay: Should that be handled here or in the checker where it validates types match the constraints?

skottmckay: Sorry - I see what you mean now. The checker would only validate the max iterations type not this. Will add a check it's int64.

PR #1594, opened 2018-11-09
Shape inference fix for broadcast, concat and scan

KeDengMS:
Shape inference fix for broadcasting and concat
Now shape inference is correct for following cases:
    broadcasting: ['S', 1, 2] + ['S', 2, 1] => ['S', 2, 2]
    concat: concat((['S', 2], ['S', 3]), axis=1) => ['S', 5]


onnx/defs/tensor/defs.cc
```diff
@@ -330,16 +330,18 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-                  if (shape.dim(j).dim_value() !=
-                      output_shape->dim(j).dim_value()) {
+              } else {
+                const auto& output_dim = output_shape->dim(j);
+                const auto& input_dim = shape.dim(j);
+                if (!output_dim.has_dim_value() && !output_dim.has_dim_param())
+                  *output_shape->mutable_dim(j) = input_dim;
+                else
+                  if ((input_dim.has_dim_value() !=
+                       output_dim.has_dim_value()) ||
```
gramalingam: I think it is possible for one of these to be a symbolic dimension and the other to have a value. I don't think we should throw a dimension-mismatch exception in that case.

gramalingam: I think it would be useful to extract a utility function "mergeDim" (e.g. https://github.com/onnx/onnx/blob/fa568e4281f5008921e7b677ead901504cab491b/onnx/defs/shape_inference.h#L499 also tries to do the same thing) so that it can be used by multiple shape-inference methods

onnx/helper.py
```diff
@@ -290,7 +290,7 @@ def make_empty_tensor_value_info(name):  # type: (Text) -> ValueInfoProto
 def make_tensor_value_info(
         name,  # type: Text
         elem_type,  # type: TensorProto.DataType
-        shape,  # type: Optional[Sequence[int]]
+        shape,  # type: Optional[Sequence[TypeVar['dimension', Text, int]]]
```
gramalingam: Would Sequnce[Union(Text,int)] be simpler? I don't understand TypeVar usage, so I am not sure. Just wondering.

KeDengMS: Yes I've switched to Union in latest commit

onnx/defs/tensor/defs.cc
KeDengMS: Thanks for the suggestion. Done

KeDengMS:
@houseroad Could you please take a look at this PR when you have time?

houseroad:
@KeDengMS sure, I will take a look this weekend.

PR #1596, opened 2018-11-09
Upgrade label encoder to support more input types

wschin:
Label encoder (e.g., [scikit-learn's](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.LabelEncoder.html)) is a general mechanism of mapping values to integers or vice versa. Today, the label encoder in ONNX only supports integer-to-string and it's inverse transform. Thus, this PR proposes a new signature to fill the gap.


linkerzhang:
don't change the .md directly, change the defs.cc and call defs/gen_doc.py --ml to generate the .md.

wschin:
@linkerzhang, there is no `--ml` and by setting up `ONNX_ML=1`, that script just remove everything in *-ml.md. Any idea?

onnx/defs/traditionalml/defs.cc
```diff
@@ -355,6 +355,124 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+        .Attr(
+            "classes_floats",
+            "A list of floats.",
+            AttributeProto::FLOATS,
+            OPTIONAL)
+        .Attr(
+            "default_string",
+            "A string.",
+            AttributeProto::STRING,
+            std::string("\"\", empty string"))
```
linkerzhang: do you want to reuse the default string value as in op set 1? "_Unused". Which looks to me better.

linkerzhang:
Please move the old version (v1) to a separate file, like old.cc.

onnx/defs/traditionalml/defs.cc
wschin: Sure.

wschin:
@linkerzhang, no problem. Your comments are addressed.

onnx/defs/traditionalml/defs.cc
```diff
@@ -302,57 +302,122 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+    Label Encoder has two operation modes. The first mode, named
+    indexing mode, only works if and only if the input is an integer
+    tensor. Indexing mode uses each input element as an index to retrieve
+    the associated value in the specified 'classes_*' field. For example, if
+    input is [1, 0, 2, 0], 'classes_floats' is [5.5, 6.6], and
+    'default_float' is -8.7, the expected output would be [6.6, 5.5, -8.7,
+    5.5]. Let's consider another example. Assume that input is [1, 2, 2, 0]
+    and 'classes_strings' is ["Sally", "Dori", "Amy"]. The output would be
+    ["Dori", "Amy", "Amy", "Sally"]. The other mode maps non-integer types
+    to integer types, called loop-up mode. Each element in the input
```
gramalingam: look-up mode?

onnx/defs/traditionalml/defs.cc
```diff
@@ -302,57 +302,122 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+    tensor is used as a position index to the associated 'classes_*'
+    attribute. Let ["A", "B", "B", "C"] be the input, 'classes_strings' be
+    ["C", "B"], 'default_int64' be -99. Then, the corresponding output
+    should be [-99, 1, 1, 0] because "C" and "B" are respectively the first
+    (indexed by 0) and the second (indexed by 1) elements in
+    'classes_strings' while "A" cannot be found in 'classes_strings'.<br>
+    Under loop-up mode, only default_int64 can be set because it is the
```
gramalingam: look-up?

onnx/defs/traditionalml/defs.cc
```diff
@@ -302,57 +302,122 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+    integer value used whenever indexing fails (e.g., when an integer
+    larger than the size of 'classes_* is given). In the meanwhile, input
```
gramalingam: This is confusing. Do you mean "when a value not present in the classes_* is given"?

onnx/defs/traditionalml/defs.cc
```diff
@@ -302,57 +302,122 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+    type is identical to the element type of the specified 'classes_' while
+    output is always an integer tensor. For indexing mode, the input type
+    is always integer tensor and the output type is identical to the
+    specified specified 'classes_*'. The 'default_*' field is the output
+    value whenever loop-up fails for an element. Note that only one of
```
gramalingam: look-up? It looks like look-up and indexing may be mixed up in the above few lines.

onnx/defs/traditionalml/defs.cc
wschin: Yep. I will improve it.

onnx/defs/traditionalml/defs.cc
wschin: Yes!

onnx/defs/traditionalml/defs.cc
```diff
@@ -302,57 +302,122 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+
+          // Look-up mode so output must be ints
+          if (TensorProto::INT64 != input_elem_type) {
             output_elem_type->set_elem_type(TensorProto::INT64);
-          } else if (TensorProto::INT64 == input_elem_type) {
-            output_elem_type->set_elem_type(TensorProto::STRING);
+            return;
           }
+
+          // The following code is for indexing mode because input_elem_type
```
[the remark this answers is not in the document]
wschin: Spec largely revised. :)

gramalingam:
Hi, the new specification looks good. However, I wonder if it would be better to introduce this as a new op, rather than replacing the old one op by a breaking change? A breaking change would force all existing exporters that export the older form of the op to update to the new version: do we really want this?

wschin:
I believe the uses of this op are very limited. We just need to minorly change Onnxmltools and ML.NET.

PR #1598, opened 2018-11-09
Bugfix in function shape/type inference

raymondxyang:
Fix a bug in accessing vector element

### DIFF
--- a/onnx/shape_inference/implementation.cc
+++ b/onnx/shape_inference/implementation.cc
@@ -232,9 +232,9 @@ void InferShapeForFunctionNode(
   GraphProto g;
   // Get a temporary tensor-shape map
   std::unordered_map<std::string, TypeProto*> temp_valueTypesByName;
-  std::vector<TypeProto> temp_types_cache;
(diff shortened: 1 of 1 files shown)


raymondxyang:
Well it actually behave differently once the capacity of the vector is met and the memory spaces got reassigned..

houseroad:
@raymondxyang could you give an example to elaborate the problem?

raymondxyang:
Well.. it either caused seg fault or get typeproto with uninitialized fields. The test case is Full Connect function op, which has three inputs and triggers the reassignment of the vector elements addresses. we store them to quickly access its shape/type info within the function.

snnn:
Or you may let temp_valueTypesByName take the ownership of these protobuf objects? Then you can delete temp_types_cache.

raymondxyang:
@snnn its not easy to do it in that way.. The ctx.getInputType will return a const pointer, Which means there has to be a mutable copy of the typeproto object while there should also be a std::unordered_map<std::string, TypeProto*> to pass to the InferenceContextImpl()

PR #1610, opened 2018-11-13
Separate types of inputs 1 and 2 in OneHot op.

spandantiwari:
Input `indices` and `depth` do not have to be the same type. This change separates the types for the two.


spandantiwari:
@linkerzhang and @gramalingam - Thanks!

PR #203, opened 2017-11-07
Adding Activation Ops

SherlockNoMad:
Adding following activation ops: 
HardSigmoid
ScaledTanh
ThresholdedRelu
LogSoftmax
Hardmax
Softsign
Softplus
ParametricSoftplus

PR #1616, opened 2018-11-14
Add attribute 'threshold' and 'value' in ReLU Op

zrphercule:
In #203, we discussed about a possible alternative of operator "ThresholdReLU" by adding more attributes in ReLU.
Since many backends have thresholded relu now, we would like to add these attributes in ReLU Op. After adding these attr, ReLU won't change its behavior if these attributes are not given.


onnx/defs/math/defs.cc
```diff
@@ -227,14 +227,24 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-the tensor elementwise.
+(Tensor<T>). For each element x in input, if x > threshold then y = x;
+else y = value. By default the value and threshold all equal to 0.
 )DOC";
 
 ONNX_OPERATOR_SET_SCHEMA(
     Relu,
     6,
     OpSchema()
+        .Attr(
```
linkerzhang: we're having a design issue here.

The attribute "threshold" type should be same as input type, while the attribute "value" type should be the same as output type.

Before we unifying the type system for attribute, input and output, let's set the attribute type as "Tensor", which looks better than current setting, though it's not perfect.

zrphercule: @linkerzhang  btw since in relu, the output type should be the same to the input type, that means "x" "y" "value" "threshold" should all share one type.

zrphercule:
@linkerzhang @gramalingam We are planning to enable ThresholdedRelu instead of adding new attrs in Relu, for the consistency of our operator set (Since we have celu, elu, prelu...). What is your opinion? Thanks!

linkerzhang:
@zrphercule let's promote ThresholdedRelu then. Do you want to make a PR to do that please?

zrphercule:
@linkerzhang Yeah we would love to promote ThresholdedRelu. I will close this pr for now, and add a new one. If you have any emergency use, feel free to promote it by your self, and I can review it for you :)

prasanthpul:
https://github.com/onnx/onnx/pull/1856 has been created to promote ThresholdedRelu. So we can close this

PR #1623, opened 2018-11-16
Show the op's type and name when the shape inference is failed.

JerryShih:
When we hit a shape inference failed, we only have the mismatched shape error message. It's still hard to find that wrong op in the whole model. This patch try to catch the runtime_error from shape inference, and show the op's name and op_type before the program terminates.


JerryShih:
The message is like:

> (op_type:Conv, name:test_layer) :Inferred shape and existing shape differ in rank: (4) vs (5)
> RuntimeError: Inferred shape and existing shape differ in rank: (4) vs (5)

JerryShih:
@zrphercule 
Should I need another reviewer(@bddppq ?) for this pr?
All tests passed.

zrphercule:
@JerryShih yeah I think we can merge it now, it looks good. Sorry didnt reply you last week because of thanks giving...

PR #1626, opened 2018-11-16
Use int instead of enum to store data type

bddppq:
Protobuf enum has a pitfall that when it has encountered an unknown value during parsing, it uses the first enum value. This means when we add new data types [BFLOAT16](https://github.com/onnx/onnx/blob/eb4b7c2cc2a0d34c0127e26c2c1cb5e712467e1e/onnx/onnx.proto#L306) to the enum, model files that are generated by the new version of onnx that have used the new data type get interpreted by the old onnx runtime as "UNDEFINED" data type.

Note this change **is backward compatible** at protobuf serialization level, so existing model files will continue working.


bddppq:
`ci/circleci: py2-gcc7-ubuntu16.04` failure is known issue, please ignore for now.

bddppq:
@houseroad 

> the value of the elem_type will be changed to 0 if they don't know how to interpret it
yes old runtime still uses enum and there is no way we can change it anymore.

> Shall we completely comment out the DataType in proto file
we still need it to define the values as constants

> The typechecker is also unhappy.
oops I will fix it soon

bddppq:
ping @linkerzhang @gramalingam

gramalingam:
Since the first enum is UNDEFINED, mapping a datatype that the runtime does not know/understand to UNDEFINED seems like reasonable behavior. May be I am missing something, but I don't quite understand the motivation for the change. When a new datatype is added, the ir-version will be changed too, right? So, when a version v1 runtime loads a v2 version model (v2 > v1), it treats any newly added datatypes in v2 as being UNDEFINED. If we change the enum to int, then all the code dealing with datatype need to handle a value outside the legal range (in a given version) in some fashion.

bddppq:
@gramalingam 
> When a new datatype is added, the ir-version will be changed too, right?

It has never been the case, both when we added complex and added bloat16. IMO Bumping ir_version is more intrusive to all existing integrations, I think unless there is a big semantics changes that we can not make it backward compatible, it's better to not bump the ir_version. For adding new dyptes, once we make the change in the PR, we can make it backward compatible at the serialization format level.

linkerzhang:
Data can't be recognized as valid enums is marked as undefined is a by-design behavior, isn't it? (I thought that was why we add undefined).

Changing it to "int" looks to me make the proto definition not good (we defined Enum DataType but we don't use it).

bddppq:
I think we should leave backends the flexibility to decide what to do (error out or gracefully pass it to downstream) when an unknown dtype is encountered.

linkerzhang:
I guess the problem you're facing with here is there's no good way for a backend to distinguish a new model (with bfloat) from an old model. This strongly reminds me of bringing the topic of when and how to bump the ir version. Looks to me, adding bfloat into the standard should bump the ir version to 4.

With version bump, a backend can support ir version <= 3 only is by design not supporting bfloat.

bddppq:
@linkerzhang I think bumping the ir_version or not when adding new dtypes is a little bit orthogonal to the proposed change here. And as I mentioned above, my goal is to let a framework that delegates execution to onnxifi can pass through models with new dtypes, not any framework is by design only support a specific ir_version. 
Also I don't see there is any downside of doing this change, backends that want to make stricter assumption can still check whether there are new dtypes that it doesn't recognize and error out.

gramalingam:
@linkerzhang and I talked about this. Summarizing our thoughts: changing the enum to int seems fine. However, when we add new types (like bfloat16), we are extending the set of serialized models (in binary format) that are considered legal. Technically, this requires the ir_version number to be bumped. Does this make sense? It seems like it would make sense for us to increment the ir_version for the upcoming release, with all these changes. Is there a problem with incrementing the ir_version?

bddppq:
@gramalingam Technically any change in the .proto file change the set of legal serialized models, I think we need further discussions about what change should and should not cause an ir_version bump. 

Could you stamp this PR? As mentioned above this change is kinda orthogonal to ir_version bump.

onnx/onnx.in.proto
```diff
@@ -311,7 +311,7 @@ message TensorProto {
   repeated int64 dims = 1;
 
   // The data type of the tensor.
-  optional DataType data_type = 2;
+  optional int32 data_type = 2;
```
gramalingam: For each of these lines, can we add a comment "// This field MUST have a valid TensorProto.DataType value".
This will at least help capture what a valid ONNX model is.

bddppq: Sure added

bddppq:
@gramalingam Sure
cc @prasanthpul @jspisak

gramalingam:
Updating the IR_VERSION in https://github.com/onnx/onnx/pull/1718 .

PR #1630, opened 2018-11-18
Fixes the missing words in README.md

cmpt376Kor:
- "We are an early stage" -> "We are at an early stage"


houseroad:
Sign the CLA please

houseroad:
Since the author didn't reply for a long time, close the PR.

Issue #1625, opened 2018-11-16
CMake Error Could NOT find PythonLibs While building ONNX on windows7


hanzhaogang:
-- Found PythonLibs: C:/ProgramData/Miniconda3/libs/Python36.lib
  -- Found PythonInterp: C:/ProgramData/Miniconda3/python.exe (found suitable ve
rsion "3.6.6", minimum required is "3.6")
  CMake Error at C:/ProgramData/Miniconda3/Lib/site-packages/cmake/data/share/cm
ake-3.12/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
    Could NOT find PythonLibs (missing: PYTHON_LIBRARIES) (found suitable
    version "3.6.6", minimum required is "3.6")
  Call Stack (most recent call first):
    C:/ProgramData/Miniconda3/Lib/site-packages/cmake/data/share/cmake-3.12/Modu
les/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
    C:/ProgramData/Miniconda3/Lib/site-packages/cmake/data/share/cmake-3.12/Modu
les/FindPythonLibs.cmake:265 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
    CMakeLists.txt:388 (find_package)

**Can updating python to 3.7 solve this issue?**

hanzhaogang:
updated to python3.7, still hit the issue:

    -- Found PythonInterp: C:/Users/Administrator/AppData/Local/Programs/Python/
Python37/python.exe (found version "3.7.1")
    -- Found PythonLibs: C:/Users/Administrator/AppData/Local/Programs/Python/Py
thon37/libs/Python37.lib
    -- Found PythonInterp: C:/Users/Administrator/AppData/Local/Programs/Python/
Python37/python.exe (found suitable version "3.7.1", minimum required is "3.7")
    CMake Error at C:/Users/Administrator/AppData/Local/Programs/Python/Python37
/Lib/site-packages/cmake/data/share/cmake-3.12/Modules/FindPackageHandleStandard
Args.cmake:137 (message):
      Could NOT find PythonLibs (missing: PYTHON_LIBRARIES) (found suitable
      version "3.7.1", minimum required is "3.7")
    Call Stack (most recent call first):
      C:/Users/Administrator/AppData/Local/Programs/Python/Python37/Lib/site-pac
[...]

PR #1631, opened 2018-11-19
Update CMakeLists.txt

hanzhaogang:
since cmake v3.12, variable "PythonInterp" and "PythonLibs" has been changed into "Python"
see https://cmake.org/cmake/help/v3.12/module/FindPython.html for details.

see details for this issue: #1625


snnn:
The old code works fine with cmake 3.12.

hanzhaogang:
@snnn see this link: [https://cmake.org/cmake/help/v3.12/module/FindPythonLibs.html?highlight=findpythonlibs](https://cmake.org/cmake/help/v3.12/module/FindPythonLibs.html?highlight=findpythonlibs), it says: FindPythonLibs 

> Deprecated since version 3.12: Use FindPython3, FindPython2 or FindPython instead.


But actually I don't know the reason why you didn't hit this issue under cmake 3.12. Maybe there are some other reasons for the issue #1625 ? For me I fix this issue by the updated CMakeList.txt (with cmake 3.12).

snnn:
It's deprecated, not removed. We don't have to do anything for that.

houseroad:
Close the PR since the change is not necessary. Btw, thanks for the PR anyway.

skottmckay:
I think it's necessary. Without the change the cmake generation fails on Windows with CMake > 3.12.

_CMake Error at C:/Program Files/CMake/share/cmake-3.13/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PythonLibs (missing: PYTHON_LIBRARIES) (found suitable
  version "3.7.1", minimum required is "3.7")_

If you run the cmake generation again it will work so something gets cached from the first attempt, however it would be better to work the first time wouldn't it?

ayermolo:
Also seeing this issue. Running twice bypasses it, but it's kind of a hack.

snnn:
LGTM.

Issue #1496, opened 2018-10-11
Please cleanup the code related to ONNX_NAMESPACE::ConvertVersion


snnn:
```c++
ModelProto ConvertVersion(
    const ModelProto& mp_in,
    const int target_version);
```

First, at here, "const int" make no sense. Please remove const.
Second, find all the callers, who is invoking this function. You'll find more problems.

snnn:
BTW, please stop abusing const. Use it only when such use increases readability. 
https://google.github.io/styleguide/cppguide.html#auto

houseroad:
PR is merged

PR #1632, opened 2018-11-19
[Easy] Remove inappropriate const in version converter code

houseroad:
Address https://github.com/onnx/onnx/issues/1496


houseroad:
cc @snnn

snnn:
LGTM.

Maratyszcza:
what is the point of removing `const`? `const` hints the reader that this variable never changes, and unless is breaks compilation / forces less efficient overload I don't see a good reason to remove it.

houseroad:
For the function declaration, const int and int are the same thing, since they are passing value. Meanwhile, in the implementation, we almost never change the value of the parameters, and we would rather to make APIs consistent. In this diff, we also would like to remove the const of the return value, because we may have some post processing afterward.

snnn:
Please make it consistent. Either you always write code like:
```
void func(int c);
```
or 

```
void func(const int c);
```
Most people will choose the first one. 
Most open source code bases don't have the second one.

Issue #1515, opened 2018-10-16
batch norm non-spatial mode test case and spec clarification


pranavsharma:
It looks like the spec is incorrect w.r.t the dimensions for B, mean, var and scale when spatial = 0. The dimensions should be 1xCxHxW or 1xCxDxHxW (at least as per the [cudnn APIs](https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#cudnnDeriveBNTensorDescriptor)). 

Also we should add a test case.

snnn:
@pranavsharma  please verify if this issue can be closed.

PR #1633, opened 2018-11-19
Fix batchnorm doc

houseroad:
Address https://github.com/onnx/onnx/issues/1515


houseroad:
cc: @pranavsharma

pranavsharma:
LGTM 👍  Btw, we were planning to remove the non-spatial case altogether from the op definition since no other framework (other than CNTK) implements it. @liqunfu will probably send a PR for it.

cloudhan:
> LGTM  Btw, we were planning to remove the non-spatial case altogether from the op definition since no other framework (other than CNTK) implements it. @liqunfu will probably send a PR for it.

Actually mxnet is using non-spatial batchnorm. The doc suggests using flatten to walkaround, but I am frustrating with "unflattening" the tensor...

PR #1635, opened 2018-11-19
Fix shape inference of subgraph

zrphercule:
In #1571 we modified the shape_inference of subgraph, make it accept all out scope values as part of its own values. But I believe this does not always make sense, especially when there are name conflicts between main graph and subgraphs.

In this PR I would like to discuss about whether we should keep this outer_scope_value_types_by_name or not.


zrphercule:
@skottmckay @gramalingam Could you please tell me the purpose of using outer_scope_value_types_by_name, and if it is safe to not use it? Thanks!

skottmckay:
They are required for control flow operators like If and Loop. You need to have that information to do type/shape inferencing for subgraphs within those operators. 

The initial approach was to take a copy of the outer scope values to allow local values to override.

PR #1637, opened 2018-11-20
deprecate no-spatial mode of BN

liqunfu:
Most AI frameworks do not have non-spatial mode for BN. For example, all 3 ops in Pytorch (torch.nn.BatchNorm1d, torch.nn.BatchNorm2d, torch.nn.BatchNorm3d) are spatial mode. In TF, BN is treaded in a flexible way that statistics are computed over dimensions whose size if 1. In all cases, what is needed is to specify dimensions over which statistics are computed. With this update, dimensions except the channel dimension are dimensions over which statistics are computed. We therefore deprecate the non-spatial mode.


liqunfu:
@houseroad, the purpose of this PR is to simplify BN spec by removing the no-spatial mode. One can view BN in its most generic form as an op to compute statistics across a subset of the input tensor's dimensions. Dimensions not involved in statistics computation are grouped into a single dimension as channel. Therefore, it is unnecessary to call out spatial and non-spatial modes. 
As state in this PR, no-spatial mode can be done by pre reshaping (to group all dimensions other than N into the channel dimension). There is no performance hit. CuDNN per-activation mode can be done similarly in spatial mode.

liqunfu:
@houseroad please any feedback? Thanks

jspisak:
LGTM!

PR #1639, opened 2018-11-21
Fix an error under -Werror by removing redundant const

daquexian:
I got an error

```
/home/daquexian/repos/onnx/onnx/shape_inference/implementation.cc: In function ‘void onnx::shape_inference::InferShapeForFunctionNode(const onnx::Function
Proto&, const onnx::ISchemaRegistry*, onnx::InferenceContext&)’:                                                                                         
/home/daquexian/repos/onnx/onnx/shape_inference/implementation.cc:242:53: error: type qualifiers ignored on cast result type [-Werror=ignored-qualifiers]
   for (int i = 0; i < (const int)(ctx.getNumInputs()); ++i) {                                                                                           
                                                     ^         
```

when using gcc 8.2.1.

It is no need to cast a rvalue to const, so I removed it. :)

### DIFF
--- a/onnx/shape_inference/implementation.cc
+++ b/onnx/shape_inference/implementation.cc
@@ -239,7 +239,7 @@ void InferShapeForFunctionNode(
   }
   // Get a temporary initial value map
   std::unordered_map<std::string, const TensorProto*> temp_initializersByName;
-  for (int i = 0; i < (const int)(ctx.getNumInputs()); ++i) {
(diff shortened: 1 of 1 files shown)


rdzhabarov:
any reason not to add -Werror flag to the CI?

houseroad:
Even with -Werror, such problem may not be detected. Also if we set -Werror, projects which use onnx may get frustrated.

Some useful blog: https://embeddedartistry.com/blog/2017/5/3/-werror-is-not-your-friend

PR #1648, opened 2018-11-26
Implement Op Annotation's for ONNX

### LINKED ISSUE #1560: Proposal: Op Registry
Once we start getting into more complicated passes having knowledge about operators becomes incredibly useful. What I propose is to maintain a registry of operators mapping to a set of descriptors. 

For example let's take a look at the [eliminate_nop_monotone_argmax](https://github.com/onnx/onnx/blob/master/onnx/optimizer/passes/eliminate_nop_monotone_argmax.h) pass. Within the pass we have to maintain an unordered set of ops that are monotone. The correct design would be to have a registry which we can ask whether or not if a certain op is monotone. Once such a registry is implemented, passes can be generalized through descriptors, therefore as we add new ops and their respective descriptors our passes naturally scale out.

Some possible descriptors we could have are:
- monotonic
- elementwise independent (op works on an individual element without knowledge of other elements, such as log, but not softmax since it requires calculating a sum over an axis)
- elementwise dependent (softmax, logsoftmax, softmin, softplus)
- reduction
- shape transform (e.g. transpose, view)

@houseroad @bddppq What do you guys think?

ArmenAg:
fixes #1560 

This PR implements _Op Annotations_ as discussed in the Issue linked above. As recommended by @houseroad annotations are added to [OpSchema](https://github.com/onnx/onnx/blob/master/onnx/defs/schema.h) and are completly optional. These Op Annotations provide some level of generalization to the optimization framework, for example if we add another Op which happens to be monotonic increasing we automatically reap the benefit of our `eliminate_nop_monotone_argmax` pass.


ArmenAg:
@houseroad Any updates on this PR?

spandantiwari:
This looks like a better way to manage optimization passes. 

Since it is possible to add multiple annotations to an op, I like the name 'op_annotation', as compared to something else like 'op_category' or 'op_kind', which have connotations of mutual exclusivity.

onnx/defs/op_annotation.h
```diff
@@ -0,0 +1,196 @@
...
+namespace ONNX_NAMESPACE {
+
+struct EnumHash {
+  template <typename T>
+  std::size_t operator()(T t) const {
+    return static_cast<std::size_t>(t);
+  }
+};
+
+enum OpAnnotationFlag : uint16_t {
```
houseroad: How about enum class?

onnx/defs/schema.cc
```diff
@@ -502,7 +502,22 @@ OpSchema& OpSchema::AllowUncheckedAttributes() {
...
+OpSchema& OpSchema::AddOpAnnotation(std::shared_ptr<OpAnnotation> annotation) {
+  this->op_annotations_.insert(*annotation);
+  this->op_annotation_flags_.insert(
+      annotation->GetAnnotations()->begin(),
+      annotation->GetAnnotations()->end());
+  return *this;
+}
+OpSchema& OpSchema::AddOpAnnotation(OpAnnotationFlag annotation_flag) {
+  auto registry = OpAnnotationRegistry::GetInstance();
+  return this->AddOpAnnotation(registry->GetOpAnnotation(annotation_flag));
```
houseroad: two return?

onnx/defs/schema.cc
```diff
@@ -502,7 +502,22 @@ OpSchema& OpSchema::AllowUncheckedAttributes() {
   allows_unchecked_attributes_ = true;
   return *this;
 }
-
+OpSchema& OpSchema::AddOpAnnotation(std::shared_ptr<OpAnnotation> annotation) {
```
houseroad: Do we really need this?

onnx/defs/op_annotation.h
```diff
@@ -0,0 +1,196 @@
...
+  ElementwiseWeakMonotonicDecreasing,
+  Reduction,
+  ShapeTransform
+};
+
+using OpAnnotationFlagSet = std::unordered_set<OpAnnotationFlag, EnumHash>;
+
+class OpAnnotationHash;
+
+class OpAnnotation {
```
houseroad: Add some comments about the purpose of OpAnnotation

onnx/defs/op_annotation.h
houseroad: Also, could you add comments for each item to explain the meaning?

ArmenAg:
@houseroad Great. I'll polish up my code and add some comments tomorrow. Do we have any other documentation that will need updating with respect to OpSchema?

houseroad:
According to https://github.com/onnx/onnx/tree/master/docs, we probably have no doc describing how our schema works. Feel free to add one to cover general schema, or specific to the op annotation. :-)

ArmenAg:
@houseroad Sorry, this week is incredibly busy for me. I definitely won't be able to work on this today. Hopefully I'll have some time towards the end of the week to finish this.

onnx/defs/op_annotation.h
ArmenAg: Done.

onnx/defs/schema.cc
ArmenAg: I think we should keep this. What I was envisioning is that if you have a some family of operators that contain a lot of annotations. You would create that a new OpAnnotation class and pass that instance of that class in.

onnx/defs/op_annotation.h
ArmenAg: Done.

onnx/defs/schema.cc
ArmenAg: What do you mean? What should we do in change in this function?

onnx/defs/schema.cc
ArmenAg: Also in the future I was thinking OpAnnotation could itself contain attributes. For example if the OpAnnotationFlag is Bounded, OpAnnotation would have a min/max attribute.

onnx/defs/op_annotation.cc
```diff
@@ -0,0 +1,6 @@
+#include "onnx/defs/op_annotation.h"
+
+namespace ONNX_NAMESPACE {
+std::shared_ptr<OpAnnotationRegistry> OpAnnotationRegistry::instance_ =
+    std::shared_ptr<OpAnnotationRegistry>(new OpAnnotationRegistry());
+} // namespace ONNX_NAMESPACE
```
houseroad: newline here

onnx/defs/schema.cc
houseroad: The following instruction is also a `return`

ArmenAg: Oh sorry about that. I didn't see the `return *this` in the github view.

ArmenAg:
> Thanks for the effort. I am wondering whether we should make the relation between different annotations more clear. for example, whether ElementwiseStrictMonotonicIncreasing should be a subclass of ElementwiseWeakMonotonicIncreasing

@houseroad I agree with you that there is a hierarchy in the annotations which is what inspired the OpAnnotation class which contain's this hierarchy. Furthermore this is a good reason to keep 
`OpSchema& OpSchema::AddOpAnnotation(std::shared_ptr<OpAnnotation> annotation)` in case we want to introduce more complicated hierarchies in the future, or allow OpAnnotation to contain values itself (e.g. Sigmoid is InLimitBoundedElementWise(min=0, max=1)).

ArmenAg:
@houseroad Any updates?

onnx/defs/op_annotation.h
```diff
@@ -0,0 +1,238 @@
...
+  // ElementwiseDependent is also an elementwise, but the elementwise dependent
+  // Op relies on information outside the current value. The softmax family of
+  // operators is a good example of an ElementwiseDependent Op. Although softmax
+  // is applied elementwise the partition portion of the calculation requires a
+  // summation across axis, therefore utilizing information outside of the
+  // current value.
+  ElementwiseDependent,
+  // Elementwise operation that is strictly monotonic increasing. In other words
+  // op(x) > op(y) iff x > y
+  ElementwiseStrictMonotonicIncreasing,
```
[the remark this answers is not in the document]
houseroad: Yeah, actually, I would like to ask whether we should make ElementwiseStrictMonotonicIncreasingOpAnn a subclass of ElementwiseWeakMonotonicIncreasingOpAnn? Since if you are strict, you must be weak too.

ArmenAg:
@houseroad

Apologies for not working on this for the last couple of weeks. I've left Microsoft and will be joining Facebook starting next week and in between was vacationing. 

In terms of introducing inheritance between the OpAnnotation classes, I think we should shy away from this. Since it's possible to have multiple super classes per OpAnnotation we would have to rely on multiple inheritance to stay consistent. I believe the way we introduced multiple flags per OpAnnotation is the cleanest.

houseroad:
@ArmenAg welcome, and no worries, I will do some check later this week. If it's fine, I will merge it.

linkerzhang:
Thank you!

I'm seeing that we'll need a more comprehensive proposal and review meeting on op annotations, and it's also a good timing to discuss move non-standard related codes/scripts out of onnx repo to make onnx standard development faster.

ArmenAg:
@houseroad @linkerzhang Any updates on this?

houseroad:
Overall I think this is a good move. Let's merge for now. We can discuss for the future later.

linkerzhang:
Oh no, I was assuming that there should be some overall discussion on how to standardizing the annotation stuff in ONNX. @houseroad , I'll revert this change and let's discuss.

PR #1653, opened 2018-11-27
Update scan

gramalingam:
(a) Simplify scan specification. It does not iterate over batch axis (but leaves it to the subgraph to handle the batch axis). 
(b) Add attribute to specify whether scan_outputs should be constructed in forward or reverse direction


gramalingam:
The old version is opset 8. (Yes, some of variables in the older version said "ver1" since it was the first version, but it was opset 8.)

houseroad:
@gramalingam yep, do you mind to clean the code (changing variable name) a bit?

gramalingam:
@houseroad, changed it

PR #1656, opened 2018-11-28
Add IsNaN operator.

pranavsharma:
This operator is available in tensorflow and numpy and required for AutoML uses case.


pranavsharma:
@linkerzhang @gramalingam @wschin can you take a look? thanks!

docs/Operators.md
```diff
@@ -4827,6 +4828,59 @@ expect(node, inputs=[x, s, bias], outputs=[y],
...
+
+<dl>
+<dt><tt>Y</tt> : T2</dt>
+<dd>output</dd>
+</dl>
+
+#### Type Constraints
+
+<dl>
+<dt><tt>T1</tt> : tensor(uint32), tensor(uint64), tensor(int32), tensor(int64), tensor(float16), tensor(float), tensor(double)</dt>
```
gramalingam: Can integral types be NaN ? I thought it was mostly for floats?

onnx/defs/tensor/defs.cc
```diff
@@ -1319,4 +1319,22 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    IsNaN,
+    9,
+    OpSchema()
+        .SetDoc(R"DOC(Returns which elements of the input are NaN.)DOC")
+      .Input(0, "X", "input", "T1")
+      .Output(0, "Y", "output", "T2")
+      .TypeConstraint(
+          "T1",
+          {"tensor(float16)","tensor(float)","tensor(double)"},
+          "Constrain to any numeric tensor type. If the dtype attribute is not provided this must be a valid output type.")
```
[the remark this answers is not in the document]
pranavsharma: fixed.

pranavsharma:
Not obvious why some of the build jobs are failing.

onnx/defs/tensor/defs.cc
```diff
@@ -1319,4 +1319,22 @@ ONNX_OPERATOR_SET_SCHEMA(
             }
           }
         }));
+
+ONNX_OPERATOR_SET_SCHEMA(
+    IsNaN,
+    9,
+    OpSchema()
+        .SetDoc(R"DOC(Returns which elements of the input are NaN.)DOC")
```
linkerzhang: please add some test cases for this operator. Refer to https://github.com/onnx/onnx/tree/master/onnx/backend/test/case/node for examples.

onnx/defs/tensor/defs.cc
```diff
@@ -1319,4 +1319,22 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+      .Input(0, "X", "input", "T1")
+      .Output(0, "Y", "output", "T2")
+      .TypeConstraint(
+          "T1",
+          {"tensor(float16)","tensor(float)","tensor(double)"},
```
linkerzhang: inputs are limited to float/double?

gramalingam: Because NaN does not apply to integral types. (Or, are you asking about float16?)

pranavsharma: Thinking more about it. The op returns a boolean tensor if a value is NAN or not. So even if we have integers we'll simply return false for them. So we don't necessarily have to constrain to only floating points. It probably make sense to support all numeric types. Thoughts?

gramalingam: But why? It seems strange to support it for int, if it is just going to return false always. Is there some use case? Otherwise, it seems to suggest that there is a way to represent NaN in the int type, which is likely to cause confusion. The real issue lies in the original (preceding) int ops that would have produced a NaN (if it could be represented), may be via overflow or divide-by-zero … so a proper solution would require extending the int ops (like multiplication or division) to return a value to indicate that this happened.

onnx/defs/tensor/defs.cc
pranavsharma: i had one. forgot to check it in.

onnx/defs/tensor/defs.cc
pranavsharma: Ok, let's keep it float only.

pranavsharma:
> Not obvious why some of the build jobs are failing.

@gramalingam @linkerzhang do any of you know what's wrong with the partial build failures? doesn't seem to be obvious.

gramalingam:
I think you need to regenerate test-coverage data when you add tests. https://github.com/onnx/onnx/blob/master/tools/update_doc.sh is a useful reference to the steps involved.

gramalingam:
Hi Pranav, I think you are running into the same kind of problems I ran into recently :-) .
(a) After test data generation, just commit the files corresponding to the NaN operator. There are some platform incompatibilities due to which I also see changes in the test data for other ops (e.g., dynamic-slice), but I don't commit them.
(b) I think you might have previously cached model files in your user-root-directory/.onnx/models, which might be affecting the test-coverage results. Deleting those will force it to get the correct current models. (But this download takes several hours! If we know which model has been updated, we might be able to do it more efficiently.)

pranavsharma:
> Hi Pranav, I think you are running into the same kind of problems I ran into recently :-) .
> (a) After test data generation, just commit the files corresponding to the NaN operator. There are some platform incompatibilities due to which I also see changes in the test data for other ops (e.g., dynamic-slice), but I don't commit them.
> (b) I think you might have previously cached model files in your user-root-directory/.onnx/models, which might be affecting the test-coverage results. Deleting those will force it to get the correct current models. (But this download takes several hours! If we know which model has been updated, we might be able to do it more efficiently.)

thanks. let me try one more time. if it doesn't work this time, i'll abandon this and create a new PR.

linkerzhang:
@houseroad @bddppq any comment on this PR please?

pranavsharma:
@gramalingam @linkerzhang this can now be merged.

PR #1658, opened 2018-11-29
Add Sign Op

zrphercule:
if x > 0,   sign(x) = 1
if x < 0,   sign(x) = -1
if x == 0, sign(x) = 0


onnx/defs/math/defs.cc
```diff
@@ -1168,4 +1168,33 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+	9,
+	OpSchema()
+		.SetDoc(Sign_ver9_doc)
+		.Input(0, "input", "Input tensor", "T")
+		.Output(
+			0,
+			"output",
+			"The sign of the input tensor "
+			"computed element-wise",
+			"T")
```
gramalingam: The output type should be int32

onnx/defs/math/defs.cc
```diff
@@ -1168,4 +1168,33 @@ ONNX_OPERATOR_SET_SCHEMA(
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to float tensors.")
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
+
+
+static const char* Sign_ver9_doc = R"DOC(
+lalala
```
gramalingam: Documentation? The documentation needs to be generated also.

zrphercule: lol yeah, I havent delete all debugging info either, just wanna ask for you and @linkerzhang 's opinion on this Op.

onnx/defs/math/defs.cc
zrphercule: I think it is not neccesary to be a int32? Or we can set an optional attribute?

houseroad: I think output tensor should have the same type as the input tensor. At least both tensorflow and pytorch follow this rule.

gramalingam: Okay, then the shape inference function below is incorrect. I was going by that.

zrphercule: @gramalingam Wonder why it is incorrect? The output here should use the same shape&type as the input...

gramalingam: But it says "updateOutputElemType(ctx, 0, TensorProto::INT32);". I think the whole thing can be replaced by .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput).

zrphercule: oh damn I didnt push this. Thanks!

onnx/defs/math/defs.cc
```diff
@@ -1168,4 +1168,27 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    Sign,
+    9,
+    OpSchema()
+        .SetDoc(Sign_ver9_doc)
+        .Input(0, "input", "Input tensor", "T")
+        .Output(
+            0,
+            "output",
+            "The sign of the input tensor "
+            "computed element-wise",
```
houseroad: Let's make it clear here: the output has the same shape and type as the input. Although this can be inferred from shape inference function.

zrphercule:
@linkerzhang Any comments? Thanks!

gramalingam:
I notice that numpy and tensorflow documentation for sign differ in the treatment of NaN. May be we should start clarifying the spec for NaN (for all ops, not just this op … this is a more general issue, especially as I see another PR for adding a IsNaN op).

PR #1660, opened 2018-11-29
repaire convtranspose shape inference

peteryang1:
convtranspose shape inference does not work right, since convtranspose has two ways of determine the output shape, but the old code only supports one, I added code to support two ways of shape inference.


peteryang1:
I solved the shape inference of convtranspose when the node is like bolow:

in this case, the shape of the output is not given, we should determine it from pad and other attributes.

hope to be reviewed, thanks!

peteryang1:
@houseroad I have added tests already.

onnx/defs/nn/defs.cc
```diff
@@ -822,24 +836,22 @@ void convTransposeShapeInference(InferenceContext& ctx) {
...
-    auto newdim = final_output_shape->add_dim();
-    if (!input_shape.dim(2 + i).has_dim_value()) {
-      continue;
+  else
+  {
+    size_of_output = input_shape.dim_size() - 2;
+    for (int i = 0; i < size_of_output; ++i)
+    {
+      int64_t output_shape_dim =
+          strides[i] * (input_shape.dim(i + 2).dim_value() - 1) +
```
gramalingam: The input_shape may have a symbolic dim and not a concrete dim_value. There needs to be a check if input_shape.dim(i+2).has_dim_value() before using that dim_value().

onnx/defs/nn/defs.cc
```diff
@@ -741,6 +741,12 @@ void convTransposeShapeInference(InferenceContext& ctx) {
     return;
   }
 
+  int64_t group = 1;
```
gramalingam: You can replace these lines with "int64_t group = getAttribute(ctx, "group", 1)".

onnx/defs/nn/defs.cc
```diff
@@ -809,37 +818,44 @@ void convTransposeShapeInference(InferenceContext& ctx) {
...
+      if (input_shape.dim(i + 2).has_dim_value()) {
+        int64_t output_shape_dim =
+            strides[i] * (input_shape.dim(i + 2).dim_value() - 1) +
+            output_padding[i] + kernel_shape[i] - pads[i] -
+            pads[i + n_input_dims];
+        
+
+        final_output_shape->add_dim()->set_dim_value(output_shape_dim);
+      } else if (input_shape.dim(i + 2).has_dim_param()){
+        final_output_shape->add_dim()->set_dim_param(input_shape.dim(i + 2).dim_param());
```
gramalingam: Just "final_output_shape->add_dim()" would be correct. We can't set the dim_param to be the same as input_shape.dim(i + 2).dim_param() (say "N") because that implies that these two dimensions are equal, which is not the case.

onnx/defs/nn/defs.cc
```diff
@@ -809,37 +818,44 @@ void convTransposeShapeInference(InferenceContext& ctx) {
...
+      }else{
```
gramalingam: Just omit the if condition "input_shape.dim(i + 2).has_dim_param()" above … the same solution is valid for both cases. If neither dim_value nor dim_param is set, it is treated as an unknown dimension. (A dim_param is useful primarily when there are two unknown dimensions of the same value.)

onnx/defs/nn/defs.cc
peteryang1: I have resolved this problem.
But I found that seems other ops also have this problem, because on one of my model, I set my input's height and width "*", after the shape inference, the height and width of the output is still "*"while it supposed to be different from the input, so I infer that param means dim_value unknown.

PR #1664, opened 2018-11-30
Replace np.long by np.int64

gramalingam:
np.long is a platform dependent type. Replace use of np.long with np.int64.


houseroad:
Please update the doc as well. Thanks

Issue #1121, opened 2018-06-16
CELU as an ONNX operator?


zasdfgbnm:
I'm adding CELU ([paper](https://arxiv.org/pdf/1704.07483.pdf)) to PyTorch (https://github.com/pytorch/pytorch/pull/8551). I want to be able to export this operator to ONNX, but I don't know what is the best way to do that.

One option is to add a new operator `Celu` to onnx, and PyTorch just export a node named 'Celu'. Another option is to export it as composed by existing ops.

I don't know which is better:

On the one hand, `Elu`, `Selu`, `Relu`, etc. are all there in ONNX, although they can be easily composed by existing ops. Which makes it looks better to make `Celu` part of the family.

On the other hand, there are some other operators that is also very common but can be composed by existing ops not in ONNX, such as https://github.com/onnx/onnx/issues/870.

prasanthpul:
Probably best to do it as a function. That way it can be optimized by runtimes that know about it but can still be executed by those that don't. @linkerzhang what do you think? do we need any enhancement to functions to fully enable this?

ebarsoum:
Yes, if still interested please add it as function: https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md

jeremycochoy:
Hello. I do need CELU unit in production setup. Is someone working on it or can I add it for you? :)

PR #1676, opened 2018-12-05
Add Celu Op

zrphercule:
Celu(x) = max(x, 0) + min(0, alpha * (exp(x/alpha) - 1))
See #1121 for related paper.


zrphercule:
I will update the docs later for not making so many conflicts.

ArmenAg:
This is more of a philosophical question, but at what point should we prefer Functions to adding new Op's? It seems like it's going to become tougher and tougher for back-ends to be ONNX compliant if we have an ever-growing addition of ops.

houseroad:
According to the doc https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md, if CELU can be added as simple combination, we should use function, such as LE => LT + EQ.

ArmenAg:
So in this case CELU can be implemented using max/min/exp/mul/sub. So should it be a Function?

houseroad:
@ArmenAg in general it depends, if it is (relatively) complicated, and in deep learning frameworks we should implement it as a separate operator. For Celu, personally, I prefer to add it as an operator, not function... because the decomposition is not straightforward. Also, for functions, I think it should not conflict with operator, actually, it can be optional attribute as part of the operator schema, just like the operator attribute.

zrphercule:
@gramalingam @linkerzhang Any comments? Thanks!

zrphercule:
@ArmenAg @spandantiwari @linkerzhang  How do you like if we should merge this op? Thanks!

onnx/defs/math/defs.cc
```diff
@@ -1059,6 +1059,29 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            AttributeProto::FLOAT,
+            1.0f)
+        .SetDoc(Celu_ver9_doc)
+        .Input(0, "input", "Input tensor", "T")
+        .Output(0, "output", "Output tensor", "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
```
spandantiwari: It will be nice to have a shape inference test also.

docs/Changelog.md
```diff
@@ -8481,6 +8481,43 @@ This version of the operator has been available since version 9 of the default O
...
 </dl>
 
+### <a name="Celu-9"></a>**Celu-9**</a>
+
+  Calculate the Celu of input tensor element-wise.
+  Celu(x) = max(x, 0) + min(0, alpha * (exp(x/alpha) - 1)).
+
+#### Version
+
+This version of the operator has been available since version 9 of the default ONNX operator set.
```
faxu: Is this for opset10? Opset9 released with ONNX 1.4

faxu:
@zrphercule : is this op still needed? ONNX 1.5 will release soon - are you targeting to include this as part of opset10?

zrphercule:
@faxu Thanks for your review. I think it is still needed, I will update it to version 10 next week with the operator LE and GE. Sounds good?

faxu:
@zr

> @faxu Thanks for your review. I think it is still needed, I will update it to version 10 next week with the operator LE and GE. Sounds good?

ONNX 1.5 release branch is planned to be cut by the end of today. If you can only get to it next week, it'll be released as part of opset 11 in ONNX 1.6 most likely.

zrphercule:
> @zr
> 
> > @faxu Thanks for your review. I think it is still needed, I will update it to version 10 next week with the operator LE and GE. Sounds good?
> 
> ONNX 1.5 release branch is planned to be cut by the end of today. If you can only get to it next week, it'll be released as part of opset 11 in ONNX 1.6 most likely.

Then I am okay with it to leave it to next version. Thanks!

faxu:
@zrphercule is this change still needed? If so, can you please sign the CLA so it can be included in the 1.6 release?

Data-drone:
was this added in the end?

zrphercule:
> was this added in the end?

Not really since this pr is out-dated. Feel free to add it into modern onnx ;)

PR #1684, opened 2018-12-07
Add LessOrEqual and GreaterOrEqual Ops.

zrphercule:
Right now, we have:

tf.math.less_equal() tf.math.greater_equal()
c2.LE() c2.GE()
mxnet::symbol.__le__() mxnet::symbol.__ge__()
np.less_equal() np.greater_equal()

Therefore it is neccesary to add two ops in ONNX as well.

### DIFF
--- a/community/working-groups.md
+++ b/community/working-groups.md
@@ -12,7 +12,6 @@ Working Groups have most of their discussions on Gitter. If you are interested i
 
 | Working Group      | Objectives    |
 | ------------------ | ------------- |
-| [Quantization](https://gitter.im/onnx/quantization) | Enhance ONNX to support quantized data types and operators on a variety of runtimes and hardware devices |
 | [Training](https://gitter.im/onnx/training) | Expand ONNX to support training as well as inference |
 | [Edge/Mobile](https://gitter.im/onnx/edge) | Enable deployment of ONNX models to edge platforms by identifying the ONNX operators that must be supported by mobile and IoT runtimes, defining tool chain requirements, and contributing to ONNX compatibility tests |
 | [Data Pipelines](https://gitter.im/onnx/pipelines) | Define new operators for processing the data that goes in and comes out |
@@ -22,6 +21,7 @@ Working Groups have most of their discussions on Gitter. If you are interested i
 | Working Group      | Objectives    | Status |
 | ------------------ | ------------- | ------ |
 | [Control Flow and Loops](https://gitter.im/onnx/ControlFlowWG) | Enable dynamic control structures to enable advanced models for NLP, speech, and video/image processing | Completed - If, Loop, and Scan operators were added in ONNX 1.3 release. |
+| [Quantization](https://gitter.im/onnx/quantization) | Enhance ONNX to support quantized data types and operators on a variety of runtimes and hardware devices | Completed - added in ONNX 1.5 release. |
 
 ## Inactive working groups
 


zrphercule:
I will generate and upload docs once it is approved.

houseroad:
@linkerzhang @ebarsoum @spandantiwari @pk-g do you mind to take a look?

onnx/defs/logical/defs.cc
```diff
@@ -121,6 +121,34 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+	OpSchema()
+		.FillUsing(BinaryLogicDocGenerator("greater or equal"))
+		.TypeConstraint(
+			"T",
+			OpSchema::all_numeric_types(),
+			"Constarins input to float tensors.")
+		.TypeConstraint(
+			"T1",
+			{"tensor(bool)"},
+			"Constrains output to boolean tensor."));
```
spandantiwari: Maybe consider adding type and shape inference function as they are increasingly being used by backends for model validation.

onnx/defs/logical/defs.cc
```diff
@@ -138,4 +166,6 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrains input/output to boolean tensors.")
         .TypeAndShapeInferenceFunction(unaryLogicalOpInference));
 
+
```
spandantiwari: line break - intentional?

onnx/defs/logical/defs.cc
```diff
@@ -121,6 +121,34 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 
+ONNX_OPERATOR_SET_SCHEMA(
+	LessOrEqual,
+	9,
+	OpSchema()
+		.FillUsing(BinaryLogicDocGenerator("less or equal"))
+		.TypeConstraint(
+			"T",
+			OpSchema::all_numeric_types(),
+			"Constarins input to float tensors.")
```
spandantiwari: The comment does not match the types allowed. Also nit: spelling for "constrains" .

onnx/defs/logical/defs.cc
```diff
@@ -121,6 +121,34 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+ONNX_OPERATOR_SET_SCHEMA(
+	GreaterOrEqual,
+	9,
+	OpSchema()
+		.FillUsing(BinaryLogicDocGenerator("greater or equal"))
+		.TypeConstraint(
+			"T",
+			OpSchema::all_numeric_types(),
+			"Constarins input to float tensors.")
```
spandantiwari: Same comment as above.

onnx/defs/logical/defs.cc
spandantiwari: nit: spelling

onnx/backend/test/case/model/greater_or_equal.py
```diff
@@ -0,0 +1,42 @@
...
+                                                       onnx.TensorProto.FLOAT,
+                                                       x.shape),
+                    onnx.helper.make_tensor_value_info('y',
+                                                       onnx.TensorProto.FLOAT,
+                                                       y.shape)],
+            outputs=[onnx.helper.make_tensor_value_info('z',
+                                                        onnx.TensorProto.FLOAT,
+                                                        z.shape)])
+        model = onnx.helper.make_model(graph, producer_name='backend-test')
+        expect(model, inputs=[x], outputs=[y],
```
spandantiwari: Great to see both node and model tests for these ops. :-)

docs/Changelog.md
```diff
@@ -8855,6 +8855,42 @@ This version of the operator has been available since version 9 of the default O
 <dd>Constrains output to boolean tensor.</dd>
 </dl>
 
+### <a name="GreaterOrEqual-9"></a>**GreaterOrEqual-9**</a>
+
+  Returns the tensor resulted from performing the `greater or equal` logical operation
```
spandantiwari: I suggest slight rephrasing: "Returns tensor resulting from..."

spandantiwari:
@zrphercule @houseroad - The spec looks good to me. 

I do have a high-level question. Do we want to introduce these as separate ops given that we can obtain the effect by simple negation of existing ops. E.g. `LessOrEqual` can be implemented by `Greater` followed by `Not`. I feel that it is OK to include them in the spec because they are common logical operators and have clear design. At the same time, such additions add to the number of supported ops. 

For e.g., we have an `Equal` op and a `Not` op. Does the spec have place for a `NotEqual` op?

shinh:
`Greater` and `Not` is wrong for NaNs?

```
>>> nan = np.array(float('nan'))
>>> nan <= nan
False
>>> not (nan > nan)
True
```

onnx/defs/logical/defs.cc
zrphercule: Yeah, and I just noticed the schema of Greater and Less were also wrong here.

zrphercule:
@spandantiwari Thanks for your review! I have two things would like to clear:

1. For these two ops, we have a need to add them because many backend frameworks have them in their op list, and may have some special optimize for them. Also, as @shinh mentioned, sometimes "not greater" is not fully equal to "less or equal" , e.g. the two numbers are not comparable. As for NotEqual, if it meet the these two conditions, I guess it is also okay to have it in ONNX?

2. I think you mentioned a very good point: when should we add one op as a "function", and when should we add it as a normal operator. @houseroad  @bddppq  and I think it might be better to give both options to frameworks and let themselves decide. 
For example, in <= case, we create a standard op LessOrEqual, meanwhile we add an optional attribute "isFunc" in it.
If isFunc is given, then we will treat LessOrEqual as a function instead of a standard operator, and dismantle it to be "Not Greater" in our exported model. This solution is more flexible, we can have a standard op while have one or even more pre-setting ways of translating it to a function. We can also let backend decide how to dismantle an op on their own by providing a subgraph as optional Attribute. Right now, our function design uses a totally different workflow, and one op can only be a function OR a normal op.

What is your oppinion? Thanks!

docs/Changelog.md
zrphercule: ok sure, this will update the docs of all logical ops, and I think it is fine =w=

spandantiwari:
@zrphercule - I understand and agree with your points about being able to optimize better when `LessOrEqual` is an op in itself and also about it producing different values for edge cases with `NaNs`. Regarding the second point - while I agree that we should let the frameworks, my point was about the number of new ops that this introduces (that's why I included the example for `NotEqual`) in the ONNX spec. Let's say that we let the framework decide which way to export, the ONNX spec still needs to have the `LessOrEqual` op in case the framework wants to export it. And this increases the number of ops. 

In this particular case I am fine with including these ops.But I have heard concerns from others about increasing number ops and that's why I would love to hear their thoughts.

prasanthpul:
One benefit of adding as a function is that runtimes can run models that use it right away, even before they have implemented an optimized implementation of that function

PR #1695, opened 2018-12-15
Add NonMaxSupression operator

HectorSVC:
Add schema for NonMaxSupression operator


onnx/defs/nn/defs.cc
```diff
@@ -1398,4 +1398,96 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            0.0f)
+        .Attr(
+            "score_threshold",
+            "Float tensor representing the threshold for deciding when to remove boxes based on score.",
+            AttributeProto::FLOAT)
+        .Attr(
+            "pad_to_max_output_size",
+            "Optional. 1(true) - the output selected_indices is padded to be of length max_output_size. Defaults to 0(false).",
+            AttributeProto::INT,
+            OPTIONAL)
```
linkerzhang: put default value here please.

onnx/defs/nn/defs.cc
```diff
@@ -1398,4 +1398,96 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            OpSchema::Optional)
+        .TypeConstraint(
+            "T1",
+            {"tensor(float)"},
+            "Constrain input "
+            " types to float tensors.")
+        .TypeConstraint(
+            "T2",
+            {"tensor(int32)"},
+            "Constrain input "
```
linkerzhang: output type.

onnx/defs/nn/defs.cc
```diff
@@ -1398,4 +1398,96 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T1")
+        .Output(
+            0,
+            "selected_indices",
+            "selected indices from the boxes tensor.",
+            "T2")
+        .Output(
+            1,
+            "valid_outputs",
+            "Optional. A 0-D integer tensor representing the number of valid elements in selected_indices, with the valid elements appearing first.",
```
linkerzhang: An integer scalar?

onnx/defs/nn/defs.cc
```diff
@@ -1398,4 +1398,96 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Attr(
+            "pad_to_max_output_size",
+            "Optional. 1(true) - the output selected_indices is padded to be of length max_output_size. Defaults to 0(false).",
+            AttributeProto::INT,
+            OPTIONAL)
+        .Input(
+            0,
+            "boxes",
+            "An input tensor. 2D tensor with shape [num_boxes, 4].",
+            "T1")
```
linkerzhang: you may put "tensor(float)" here directly as T1 defined below only has float.

linkerzhang: Same for the T2 below.

onnx/defs/nn/defs.cc
HectorSVC: I didn't find a overload Attr which can accept the (..., bool required, T defaultvalue)

---
In reply to: [242298590](https://github.com/onnx/onnx/pull/1695#discussion_r242298590) [](ancestors = 242298590)

linkerzhang:
@HectorSVC please also add shape inference test.

PR #1700, opened 2018-12-19
 Update definition of Cast Op to support casting to/from string

raymondxyang:
* Update the Cast op to support casting to/from string tensors
* Enable float <-> string test cases


raymondxyang:
Reopen after fix

PR #1703, opened 2018-12-19
Add NonMaxSuppression operator

HectorSVC:
NMS is from Tensorflow: 
Definition refers to: https://www.tensorflow.org/api_docs/python/tf/image/non_max_suppression

Here we define the "pad_to_max_output_size" as optional input, and "valid_outputs" as optional output, so it can support both TF NMS v3 and v4. More details refers to TF implementation: 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/non_max_suppression_op.cc


HectorSVC:
@bddppq @linkerzhang

HectorSVC:
anyway to re-run the CircleCI? I even close and re-open the PR, still can't re-run it. From the log "Connection reset by peer". What's wrong with the CI?

linkerzhang:
@houseroad

HectorSVC:
I'll update this according our latest changes.

HectorSVC:
@houseroad  Could you review the updated schema? We have a model converted from Keras. Need to get this op in to support that model.

linkerzhang:
let's also create a PR to add the model, so that the usage of this op is clear.

guschmue:
We have a few models working using this op:
From [this list](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/detection_model_zoo.md) ssd_mobilenet_v1_coco_2018_01_28 is working (we think a good part of the list will work).
There is a pytorch implementation for the ssd-resnet34 (needed for mlperf) [here](https://github.com/BowenBao/inference/tree/master/cloud/single_stage_detector/pytorch#6-onnx) that is working.
And this [tensorflow version of yolov3](https://github.com/YunYang1994/tensorflow-yolov3)  works as well.

jywu-msft:
@houseroad , given @guschmue 's list, it seems like we can enable many object detection models in onnx with this change. Can you please review and approve if no objections?

onnx/defs/nn/defs.cc
```diff
@@ -2125,4 +2125,66 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(
+            1,
+            "scores",
+            "An input tensor with shape [num_batches, num_classes, spatial_dimension]",
+            "tensor(float)")
+        .Input(
+            2,
+            "max_output_boxes_per_class",
+            "Integer representing the maximum number of boxes to be selected per batch per class. It is a scalar.",
+            "tensor(int32)",
```
houseroad: In ONNX, we use int64 as indices/sizes, shall we align with other ops as well here?

onnx/defs/nn/defs.cc
```diff
@@ -2125,4 +2125,66 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            OpSchema::Optional)
+        .Input(
+            4,
+            "score_threshold",
+            "Float representing the threshold for deciding when to remove boxes based on score. It is a scalar",
+            "tensor(float)",
+            OpSchema::Optional)
+        .Output(
+            0,
+            "selected_indices",
```
houseroad: ditto

onnx/defs/nn/defs.cc
HectorSVC: updated

---
In reply to: [276524091](https://github.com/onnx/onnx/pull/1703#discussion_r276524091) [](ancestors = 276524091)

onnx/defs/nn/defs.cc
HectorSVC: updated

---
In reply to: [276524111](https://github.com/onnx/onnx/pull/1703#discussion_r276524111) [](ancestors = 276524111)

HectorSVC:
@houseroad Any further comments?

ebarsoum:
@houseroad  anything else missing? If not can we merge it?

HectorSVC:
Why I still can't merge it? Need approval from @bddppq ?

PR #1704, opened 2018-12-20
Update definition of Cast Op to support casting to/from string

raymondxyang:
* Update the Cast op to support casting to/from string tensors
* Enable float <-> string test cases


onnx/backend/test/case/node/cast.py
```diff
@@ -25,18 +25,26 @@ def export():  # type: () -> None
...
         for from_type, to_type in test_cases:
-            input = np.random.random_sample(shape).astype(
-                TENSOR_TYPE_TO_NP_TYPE[getattr(TensorProto, from_type)])
+            if 'STRING' != from_type:
+                input = np.random.random_sample(shape).astype(
+                    TENSOR_TYPE_TO_NP_TYPE[getattr(TensorProto, from_type)])
+
+                output = input.astype(TENSOR_TYPE_TO_NP_TYPE[getattr(TensorProto, to_type)])
+            else:
+                input = np.array([['0.47892547', '0.48033667', '0.49968487', '0.81910545'],
```
wschin: Do you have tests for inf and nan? The spec also needs to mention the string literals of them.

onnx/defs/tensor/defs.cc
```diff
@@ -60,8 +62,8 @@ ONNX_OPERATOR_SET_SCHEMA(
              "tensor(uint16)",
              "tensor(uint32)",
              "tensor(uint64)",
-             "tensor(bool)"},
-            "Constrain output types. Casting to strings and complex are not supported.")
+             "tensor(string)"},
```
linkerzhang: bool was removed by mistake?

raymondxyang: oops... fixed

docs/Changelog.md
```diff
@@ -8481,6 +8481,50 @@ This version of the operator has been available since version 9 of the default O
 <dd>Constrain input and output types to float tensors.</dd>
 </dl>
 
+### <a name="Cast-9"></a>**Cast-9**</a>
+
+  The operator casts the elements of a given input tensor to a data type
+  specified by the 'to' argument and returns an output tensor of the same size in
+  the converted type. The 'to' argument must be one of the data types specified
+  in the 'DataType' enum field in the TensorProto message. Litrals string 'NaN' 
+  and 'INF' are defined to be supported in a case-insensitive manner when cast from
```
wschin: There are two `INF`, positive and negative. Which one does `INF` mean? If `INF' stands for positive inf, can we do `-INF` for negative inf? In addition, due to the existence of `-INF`, would it be nice to support `+INF`?

onnx/defs/tensor/defs.cc
```diff
@@ -7,19 +7,21 @@
...
 
 namespace ONNX_NAMESPACE {
-static const char* Cast_ver6_doc = R"DOC(
+static const char* Cast_ver9_doc = R"DOC(
 The operator casts the elements of a given input tensor to a data type
 specified by the 'to' argument and returns an output tensor of the same size in
 the converted type. The 'to' argument must be one of the data types specified
-in the 'DataType' enum field in the TensorProto message.
-NOTE: Casting to and from strings is not supported yet.
+in the 'DataType' enum field in the TensorProto message. Litrals string 'NaN'
```
houseroad: You should explicitly say when converting `String` tensor to numeric type of tensors. Also you should say which numeric representation supported.

docs/Changelog.md
raymondxyang: Sure. Will clarify that

raymondxyang:
@houseroad Reverted the format changes in def.cc

docs/Changelog.md
```diff
@@ -8481,6 +8481,56 @@ This version of the operator has been available since version 9 of the default O
...
 </dl>
 
+### <a name="Cast-9"></a>**Cast-9**</a>
+
+  The operator casts the elements of a given input tensor to a data type
+  specified by the 'to' argument and returns an output tensor of the same size in
+  the converted type. The 'to' argument must be one of the data types specified
+  in the 'DataType' enum field in the TensorProto message.
+  
+  Casting from string tensor in plain (e.g., "3.14" and "1000") and scientific numeric representation
```
```suggestion
  Casting from string tensor in plain (e.g., "3.14" and "1000") and scientific numeric representations
```

docs/Changelog.md
```diff
@@ -8481,6 +8481,56 @@ This version of the operator has been available since version 9 of the default O
...
+  (e.g., "1e-5" and "1E8") to all numeric types is supported. Values will be implicitly truncated
+  based on the target tensor type's precision. For example, converting string "100.5" to an integer may
+  result 100. There are some string literals reserved for special floating-point values;
+  "+INF" (and "INF"), "-INF", and "NaN" are positive infinity,  negative infinity, and not-a-number, respectively.
```
```suggestion
  "+INF" (and "INF"), "-INF", and "NaN" are positive infinity, negative infinity, and not-a-number, respectively.
```

docs/Changelog.md
```diff
@@ -8481,6 +8481,56 @@ This version of the operator has been available since version 9 of the default O
...
+  Any string which can exactly match "+INF" in a case-insensitive way would be mapped to positive infinite. Similarly,
+  this case-insensitive rule is applied to "INF" and "NaN". When casting from numeric tensors
+  to string tensors, plain floating-point representation (such as "3.1415926") would be used.
```
wschin: Let's make it more different than scientific representation.
```suggestion
  to string tensors, plain floating-point representation (such as "314.15926") would be used.
```

docs/Changelog.md
```diff
@@ -8481,6 +8481,56 @@ This version of the operator has been available since version 9 of the default O
 <dd>Constrain input and output types to float tensors.</dd>
 </dl>
 
+### <a name="Cast-9"></a>**Cast-9**</a>
+
+  The operator casts the elements of a given input tensor to a data type
+  specified by the 'to' argument and returns an output tensor of the same size in
+  the converted type. The 'to' argument must be one of the data types specified
+  in the 'DataType' enum field in the TensorProto message.
+
```
wschin: Conversion from a numerical type to any numerical type is always allowed. User must be aware of precision loss and value change caused by range difference between two types. For example, a 64-bit float 3.1415926459 may be round to a 32-bit float 3.141592. Similarly, converting an integer 36 to Boolean may produce 1 because we truncate bits which can't be stored in the targeted type.

docs/Changelog.md
raymondxyang: ooops.. clicked the auto correction.. should do fix in .cc source

docs/TestCoverage.md
```diff
@@ -910,21 +910,29 @@ test_cases = [
...
 for from_type, to_type in test_cases:
-    input = np.random.random_sample(shape).astype(
-        TENSOR_TYPE_TO_NP_TYPE[getattr(TensorProto, from_type)])
+    if 'STRING' != from_type:
+        input = np.random.random_sample(shape).astype(
+            TENSOR_TYPE_TO_NP_TYPE[getattr(TensorProto, from_type)])
+
+        output = input.astype(TENSOR_TYPE_TO_NP_TYPE[getattr(TensorProto, to_type)])
+    else:
+        input = np.array([['0.47892547', '0.48033667', '0.49968487', '0.81910545'],
```
wschin: This test needs to contain all special literal numbers. You already have "NaN" and "INF" so only "-INF" and "+INF" are missing.

docs/Operators.md
```diff
@@ -1563,19 +1563,33 @@ expect(node, inputs=[x, s, bias, mean, var], outputs=[y],
   specified by the 'to' argument and returns an output tensor of the same size in
   the converted type. The 'to' argument must be one of the data types specified
   in the 'DataType' enum field in the TensorProto message.
-  NOTE: Casting to and from strings is not supported yet.
+  
+  Casting from string tensor in plain (e.g., "3.14" and "1000") and scientific numeric representations
+  (e.g., "1e-5" and "1E8") to all numeric types is supported. Values will be implicitly truncated
+  based on the target tensor type's precision. For example, converting string "100.5" to an integer may
```
houseroad: I tried this in numpy, `100.5` to int will cause error. So I would suggest throw the error as well. If user really want to do so, they can convert to float first, then convert to int.

raymondxyang: +@wschin for discussion

wschin: I guess users specify float-to-int conversion only when they really want (if not by rare accident).

houseroad: So to prevent such accident, let's throw error in such cases.

wschin: No I mean this is an operator, not a programming language. In addition, what's the definition of throwing in an IR? You can have at most an undefined behavior because exception is runtime behavior which can't be controlled by IR.

raymondxyang: Discussed offline. Define this rare case (string representing 'float' -> int) as undefined behavior.

houseroad:
Please resolve the conflict.

magneter:
Hi ,a similar problem occured to me .
It tips : Unsupported ONNX ops of type: Cast, onnx version is onnx-1.6.0.
It seems ,1.6.0 did not support Cast op.
How Could I solve this ?

PR #1705, opened 2018-12-20
Make NodeProto and TypeProto.Tensor extensible

bddppq:
To allow downstream frameworks to export additional annotations

### DIFF
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -3,5 +3,10 @@
 set -ex
 
 source /tmp/venv/bin/activate
+
+# install torchvision from master
+# the one on pypi requires cuda
+pip install -q git+https://github.com/pytorch/vision.git
+
 cd /tmp/pytorch
 CI=1 exec "scripts/onnx/test.sh"


bddppq:
@houseroad why? :-)

houseroad:
There is a tiny possibility, we may get have more than 100 attributes in some message. To me, (1000, max) or (100, 1000) sounds better. Although, there is no big difference.

onnx/onnx.proto
```diff
@@ -424,6 +425,7 @@ message TypeProto {
     // This field MUST be present for this version of the IR.
     optional int32 elem_type = 1;
     optional TensorShapeProto shape = 2;
+    extensions 100 to max;
```
gramalingam: Do you want this for TypeProto or Tensor? The title of the PR says TypeProto, but it is being added to Tensor.

bddppq: Oh you are right, it should be `TypeProto.Tensor`

houseroad:
@linkerzhang any objection?

shinh:
This is great! Any reason ValueInfoProto doesn't have it? For example, if ValueInfoProto is extensible, I'd consider using the room to store information about devices where a value will be placed. Though we can also store such information to NodeProto or TypeProto, I'd imagine there could be some information which ValueInfoProto is the most suitable place.

prasanthpul:
It's not clear how this increases interoperability. The extensions would not be centrally managed so they may conflict and could be interpreted differently by different backends/tools/etc.

Which specific annotations/extensions you are looking to add via these fields?

bddppq:
@prasanthpul 
> The extensions would not be centrally managed so they may conflict and could be interpreted differently by different backends/tools/etc.

I think the policy is the same as custom operator domain.

> Which specific annotations/extensions you are looking to add via these fields?

It's similar to @shinh's, i.e. putting some device information into the model.

> if that is something of interest to others, we should see how to add a field for that specific annotation.

I believe runtime is a specific thing that we want to avoid adding to onnx core since the very beginning.

linkerzhang:
I'm thinking in a little bit different way. Looks to me this kind of extension is not part of standard. Shall we make them in each framework locally? This reminds me of reserving ids for ONNX standard for each message actually.

linkerzhang:
@shinh  dumb question please. I don't quite understand why using model file to store debugging information. Are you going to serializing those debugging info into a model file (*.onnx) please? Thanks!

shinh:
@linkerzhang Sorry, but I'm not sure if I understand your question, so let me describe my usecase in detail. Let's assume we have the following Python code

```
1: def foo(x, y, z):
2:   a = x + y
3:   return a + z
```

and this code was exported as an ONNX model file, which have two `Add` ops. If a user passes a tensor with wrong shape as `z` to an ONNX importer, the ONNX importer will fail due to dimension mismatch. However, if the user don't know much about ONNX, the user may not see whether line 2 or 3 is wrong. I'm thinking of serializing information like "foo.py:3" to ops so an ONNX importer can show Python-level errors to end-users.

We could put this kind of information in a separate file, though.

linkerzhang:
@shinh Thank you very much for detail clarification. I understand the use case, and I believe this kind of debugging information is very useful, but I personally don't think it should be serialized in model file, but in a separate file (debugging log, say) as you mentioned.

houseroad:
@linkerzhang I think we should add such supports, so users can easily do experiments with their new ideas. It won't bring us any disadvantage, instead, it will make users' life much easier if the standard onnx cannot meet all their requirements. ONNX standard iterates slowly (at least 3 month per release). When user thinks the design of their new feature is mature enough , they can propose/promote it to an official feature to the core ONNX standard.

bddppq:
@linkerzhang This change is exactly enabling partners to do extension outside of onnx. If we don't provide a way to do extension then people don't have a choice but can only force adding special features into onnx core or hard fork onnx which will be even worse.

faxu:
@bddppq : is this change still needed? ONNX 1.5 will release very soon.

bddppq:
@faxu I still think this is a good change, but it should not block 1.5 release.

PR #1709, opened 2018-12-21
Add operator MurmurHash3

HectorSVC:
MurmurHash is a non-cryptographic hash function suitable for general hash-based lookup. (Wiki: https://en.wikipedia.org/wiki/MurmurHash)
MurmurHash3_x86_32 is used to generating low latency 32bits hash suitable for implementing lookup tables, Bloom filters, count min sketch or feature hashing.
It is used in scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/src/MurmurHash3.cpp

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -1176,19 +1176,19 @@ If input > 0, output 1. if input < 0, output -1. if input == 0, output 0.
 )DOC";
 
 ONNX_OPERATOR_SET_SCHEMA(
-	Sign,
-	9,
-	OpSchema()
-		.SetDoc(Sign_ver9_doc)
-		.Input(0, "input", "Input tensor", "T")
-		.Output(
-			0,
-			"output",
-			"The sign of the input tensor "
-			"computed element-wise. It has the same shape and type of the input.",
-			"T")
-		.TypeConstraint(
-			"T",
+    Sign,
+    9,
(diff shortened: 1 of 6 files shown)


pranavsharma:
Adding @wschin

HectorSVC:
@wschin

onnx/defs/math/defs.cc
```diff
@@ -1198,20 +1198,54 @@ Computes the error function of the given input tensor element-wise.
...
+            "seed",
+            "Seed for the hashing algorithm, unsigned 32-bit integer, default to 0.",
+            AttributeProto::INT,
+            static_cast<int64_t>(0LL))
+        .Input(0, "X", "An input tensor to hash.", "T1")
+        .Output(0, "Y", "32-bit hash value.", "T2")
+        .TypeConstraint(
+            "T1",
+            {"tensor(uint32)", "tensor(int32)", "tensor(string)"},
+            "Constrain input type to unsigned or signed 32-bit integer tensor, or string tensor. It should be utf-8 encoded if using unicode.")
```
houseroad: At this moment, we can say that string must be encoded using utf-8. Since in ONNX, we only support utf-8.

postrational:
Would it be possible to implement this as an ONNX `Function`? 
If not, then what primitive op are we missing?

PR #1713, opened 2018-12-27
Changed the MaxRoiPool index input tensor type from float to int64

MrGeva:
Int64 seems like a better type for represent indices


houseroad:
@MrGeva do you have any concrete example for this op implementation in systems like mxnet, cntk, which use int64? I remember in Caffe2 the operator's implementation (not only the schema) is using floats to store the indices.

PR #1714, opened 2019-01-02
Add NonZero op.

kit1980:


docs/Changelog.md
```diff
@@ -9003,6 +9003,38 @@ This version of the operator has been available since version 9 of the default O
 <dd>Constrain index tensor to int64</dd>
 </dl>
 
+### <a name="NonZero-9"></a>**NonZero-9**</a>
+
+  Returns the indices of the elements that are non-zero
```
wschin: Would you mind to add an example here? The code used in numpy could be changed in the future.

houseroad:
@kit1980 could you update the PR using https://github.com/onnx/onnx/blob/master/tools/update_doc.sh?

onnx/defs/tensor/defs.cc
```diff
@@ -1417,4 +1417,26 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    https://docs.scipy.org/doc/numpy/reference/generated/numpy.nonzero.html
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    NonZero,
+    9,
+    OpSchema()
+    .SetDoc(NonZero_ver9_doc)
+    .Input(0, "X", "input", "T")
+    .Output(0, "Y", "output", "tensor(int64)")
```
houseroad: Let's make it clear here the output is always a 2d tensor.

kit1980: done

kit1980:
Rebased on master and added mention of always 2D output.

PR #1716, opened 2019-01-03
Remove ConstantLike op. Updates to ConstantOfShape op.

spandantiwari:
This PR has two main changes:
1. Remove `ConstantLike`: Now that `ConstantOfShape` has been merged, `ConstantLike` can be removed. Since `ConstantLike` was added only in `Opset 9`, and has not been part of any release, no formal deprecation is needed.

1. Update `ConstantOfShape`: The following updates have been made:

- Shape inference has been added.
- Shape inference tests have been added.
- Some verbiage has been added to the description of `value` attribute to clarify that the default is `0.0f`. 
- `int64` has been added as allowed type for the shape input. This is because the `Shape` op returns `tensor(int64)` (because -1 can be part of shape), and it is a common scenario to provide the output of `Shape` op as input to `ConstantOfShape` op. Also, `uint64` has been removed as allowed input type, because internally we read the values of shape input as `int64`, and if we allow `uint64`, it may create issues for very large positive values.


spandantiwari:
@houseroad @zrphercule - I see that PyTorch tests for zeros_like, ones_like, and full_like are failing. It is expected because these PT ops are exported using `ConstantLike`, which is being removed here. 

How should we handle this? Is there a way to disable the test so that this PR can make it through? I can then update the export of these ops and enable the tests here. 

Or do you think we should update the PT export first and then get the updated tests here? Not sure how much the lag will be in the latter case to get the updated backend here. We need to be mindful of the ONNX 1.4 release date which may be near.

zrphercule:
@spandantiwari Yes if we can fix pytorch side, the only question is in what level. If we only remove this constantlike op, them we can use constantofshape in oneslike\fulllike\zerolike symbolic; If we are going to remove all these XXXlike ops, it is going to be another story, we need to remove all related symbolic & tests.

spandantiwari:
@zrphercule  - that's a valid point and one that I have discussed with @linkerzhang and @gramalingam. @linkerzhang and @gramalingam can chime in too, but my understanding is that we agree on removing `ConstantLike` now that `ConstantOfShape` has been added. There are no separate ops for OnesLike and ZerosLike in ONNX. It makes sense to overhaul all the `*Like` generator ops, but that will have to be a separate PR, especially because those have been released previously (but not `ConstantLike`) and would need to be deprecated. 

On the PyTorch side, maybe we can start by moving `oneslike\fulllike\zerolike` to using `ConstantOfShape` in export. That way this PR will get unblocked.

It will be great if we can clean up these Constant* ops before the upcoming ONNX 1.4 release.

houseroad:
@spandantiwari I had a discussion with @zrphercule in person. Yeah, we agree on that after adding ConstantOfShape, we should remove ConstantLike, since the functionality has overlap. Let's disable the ConstantLike test in PyTorch first (Rui is working on that, and will keep you updated), then we can push this PR in. After that, we can update PyTorch ONNX exporter. 

For other *Like ops, let's change them in the same way, but in different PRs.

Thanks for the push on this. :-)

spandantiwari:
@houseroad - that sounds like a good plan. 
@zrphercule - please let me know when the tests are disabled and we are ready to try pushing in this PR. Thanks!

zrphercule:
@spandantiwari https://github.com/pytorch/pytorch/pull/15740 The related pr has been landed. Please try to rerun the test and see if it has been skipped.

spandantiwari:
@zrphercule - thanks! Let me give it a shot.

docs/Operators.md
```diff
@@ -2203,27 +2091,27 @@ This version of the operator has been available since version 9 of the default O
...
 <dt><tt>output</tt> : T2</dt>
-<dd>Output tensor, using input as shape.If attribute 'value' exist, all value and datatype of output tensor equal to 'value'.Else, all value equal to 0, and datatype default to float32</dd>
+<dd>Output tensor of shape specified by 'input'.If attribute 'value' is specified, the value and datatype of the output tensor is taken from 'value'.If attribute 'value' is not specified, the value in the output defaults to 0, and the datatype defaults to float32.</dd>
 </dl>
 
 #### Type Constraints
 
 <dl>
-<dt><tt>T1</tt> : tensor(uint8), tensor(uint16), tensor(uint32), tensor(uint64)</dt>
+<dt><tt>T1</tt> : tensor(uint8), tensor(uint16), tensor(uint32), tensor(int64)</dt>
```
houseroad: also add int32 and remove uint8, uint16, and uint32? They are  rarely used in the real systems

onnx/test/shape_inference_test.py
```diff
@@ -1202,6 +1202,23 @@ def test_loop(self):    # type: () -> None
             [make_tensor_value_info('loop_state_final', TensorProto.FLOAT, None),  # shape may change between iterations
              make_tensor_value_info('loop_output', TensorProto.FLOAT, (None, 3))])  # type: ignore
 
+    def test_constantofshape_with_input_shape(self):  # type: () -> None
+        graph = self._make_graph([],
+            [make_node("Constant", [], ['shape'],
+                       value=make_tensor('shape', TensorProto.INT64, (3,), (3, 4, 5))),
+             make_node("ConstantOfShape", ['shape'], ['y'], value=make_tensor('shape', TensorProto.INT32, (1, ), (2, )))],
```
houseroad: The name of the `value` tensor (attribute) is confusing. Can we rename it to `value` or something else?

onnx/test/shape_inference_test.py
```diff
@@ -1202,6 +1202,23 @@ def test_loop(self):    # type: () -> None
...
+            [])
+        self._assert_inferred(graph,
+            [make_tensor_value_info('shape', TensorProto.INT64, (3,)),
+             make_tensor_value_info('y', TensorProto.INT32, (3, 4, 5))])  # type: ignore
+
+    def test_constantofshape_without_input_shape(self):  # type: () -> None
+        graph = self._make_graph([('shape', TensorProto.INT64, (3, ))],
+            [make_node("ConstantOfShape", ['shape'], ['y'], value=make_tensor('shape', TensorProto.INT32, (1, ), (2, )))],
```
houseroad: Same here

onnx/defs/generator/defs.cc
```diff
@@ -180,6 +90,59 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            return;
+          }
+
+          // This is the second case when exact shape data is available.
+          // In this case, we extract the shape values from input tensor
+          // and create output tensor of that shape.
+          // First, extract target shape value.
+          std::vector<int64_t> targetShape;
+          if (targetShapeInitializer->has_raw_data()) {
+            const std::string& bytes = targetShapeInitializer->raw_data();
```
houseroad: Nit: assert targetShapeInitializer's data_type is int64

onnx/defs/generator/defs.cc
```diff
@@ -111,6 +111,7 @@ ONNX_OPERATOR_SET_SCHEMA(
                     if (input_shape_dim_value > 0) {
                         for (int i = 0; i < input_shape_dim_value; ++i) {
                             auto newdim = final_output_shape->add_dim();
+                            (void)(newdim); // To eliminate "unused variable" compiler warning.
```
houseroad: Or just call the function, not creating the variable newdim to store the return value?

docs/Operators.md
spandantiwari: Agreed. Done.

onnx/test/shape_inference_test.py
spandantiwari: Fixed.

onnx/test/shape_inference_test.py
spandantiwari: Fixed.

spandantiwari:
@houseroad - updated based on your feedback. If all goes well, I will merge this.

spandantiwari:
@houseroad - Shall we merge?

houseroad:
Done :-)

Issue #1449, opened 2018-09-26
Initializers and graph inputs


gramalingam:
The ONNX spec requires every initializer X to also appear in the list of graph inputs. I suggest relaxing this constraint. It appears to have no advantage and several disadvantages (due to changes that have happened since the original spec). With the transformations to lift "Constants" into "Initializers", this restriction looks a bit odd. Furthermore, the original spec allowed callers to override the values specified in the initializers, which also seems to have been removed. Any thoughts?

houseroad:
One advantage is that, we can use input to replace the value in the initializer. If we don't want this feature, we can relax the restriction.

snnn:
If we changed that, it means we need to update IR version. Then the new model cannot be loaded in an old runtime. This is not good.

houseroad:
We don't need to change proto, but probably bumping up the IR version. And backends have to support new version, I think it won't bring us a lot of benefits.

gramalingam:
@houseroad : yes, compatibility is a good reason to leave things as they are for now. But I feel that there is still ambiguity about whether inputs are allowed to replace the value in the initializer (though that was the original intent; it seems to have been changed along the way.) Optimizations like lifting a "Constant" to be an "Initializer" do complicate things. Okay ... let me leave this open just to see if anybody has any other opinion before closing it.

skottmckay:
In the scenario of lifting a Constant to an initializer, isn't it a confusing and unexpected side effect to have to add a graph input? Doesn't the usage of a Constant node declare an intention to not allow it to be overridden by an input? 

Is an alternative to have a separate list of non-graph input initializers that are used in scenarios like Constant lifting, to clarify which initializers match graph inputs and are overridable, and which ones don't?

gramalingam:
Yes, I agree that it is confusing.

The alternative you propose is one possibility. It has the advantage of supporting both kinds of initializers. (However, I am not sure if even the current standard permits overriding of initializers by inputs. The changes in https://github.com/onnx/onnx/pull/690 seemed to have changed the intention/goals of allowing such overriding.)

It would be nice to eliminate the confusion one way or another.

PR #1718, opened 2019-01-04
Relax constraint that the initializers must be a subset of graph inputs

gramalingam:
Drop the constraint that the initializers must be a subset of graph inputs. An initializer is allowed to have the same name as a graph-input, but with this change it is not required to be a graph-input.

This is important now that we have control-flow ops with subgraphs. It becomes impossible to have an initializer to represent a constant in a sub-graph of a control-flow op otherwise.

Since we will be changing the ONNX ir-version for the upcoming release, this seems a good time to make this change as well.

See previous discussions on this at: https://github.com/onnx/onnx/issues/1449 and https://github.com/onnx/onnx/pull/1579.


onnx/onnx-ml.proto
```diff
@@ -252,7 +252,7 @@ message GraphProto {
 
   // A list of named tensor values, used to specify constant inputs of the graph.
   // Each TensorProto entry must have a distinct name (within the list) that
-  // also appears in the input list.
+  // MAY also appear in the input list.
```
houseroad: We should also tell users which are recommended to appear in the input list, for example, some weights may be overwritten. Some constants won't be changed for sure, they should not be included in the initializers.

gramalingam: I added this explanation to IR.md, since that seemed like a good place for this.

gramalingam: @houseroad I am in the middle of updating the IR_VERSION. It looks like I may have to regenerate the python notebooks in the onnx\examples folder since I am running into a failure there during testing. Please let me know how these files are generated or how to update them.

onnx/onnx.in.proto
```diff
@@ -73,7 +73,12 @@ enum Version {
   //    - Added new message OperatorSetIdProto
   //    - Added opset_import in ModelProto
   // - For vendor extensions, added domain in NodeProto
-  IR_VERSION = 0x0000000000000003;
+  IR_VERSION_2017_11_3 = 0x0000000000000003;
+
+  // IR VERSION 4 published on Jan 22, 2019
+  // - Relax constraint that initializers should be a subset of graph inputs
+  // - Add type BFLOAT16
```
houseroad: Please record the change in https://github.com/onnx/onnx/pull/1626

houseroad:
I will take care of the pytorch ci part.

gramalingam:
@houseroad thanks! I assume you are offering to take care of the CircleCI failure … if there is something that I need to do, please let me know. Thanks again.

houseroad:
Yeah, I mean the CricleCI, I will create a PR to update the expects in pytorch. I will do the review for this PR again soon.

linkerzhang:
@houseroad  ok to merge this one please? this PR also contains the change of bumping IR version.

PR #1723, opened 2019-01-10
Support rtol and atol at the model granularity

houseroad:
Some models, rtol=1e-3 and atol=1e-7 may be too tight. Let's enable that different models may have different rtol and atol.


houseroad:
cc: @BIT-silence

PR #1724, opened 2019-01-10
Fix the OSX CI

houseroad:


houseroad:
See whether installing package separately solves the problem. If not, we can just go with the first way.

PR #1729, opened 2019-01-10
Fix shape propagation for Dropout to mask (2nd output)

### LINKED ISSUE #1728: propagateShapeAndTypeFromFirstInput only fills first output breaking e.g. Dropout shape inference
In https://github.com/onnx/onnx/blob/8384c788939bc65463f9754b6a7a00b212b18ba1/onnx/defs/shape_inference.h#L315 the function only fills the shape and type for first output. This makes use of it in Operators with multiple outputs (like `Dropout`) incomplete.

Solution: Fill all passed outputs

Flamefire:
Fixes #1728

Extends test case for Dropout to test all 3 variants:
  - 1 output, optional missing
  - 1 output, optional empty
  - 2 outputs

### DIFF
--- a/onnx/defs/shape_inference.h
+++ b/onnx/defs/shape_inference.h
@@ -158,7 +158,7 @@ inline void validateAndSetElemType(
     TypeProto* output_type) {
   if (output_type->value_case() == TypeProto::VALUE_NOT_SET) {
     output_type->mutable_tensor_type()->set_elem_type(data_type);
-  } else if (output_type->value_case() == TypeProto::kTensorType) {
+  } else if (output_type->has_tensor_type()) {
     if (output_type->tensor_type().has_elem_type()) {
       if (data_type != output_type->tensor_type().elem_type()) {
         fail_type_inference(
@@ -201,7 +201,7 @@ inline void propagateElemTypeWithValidation(
     fail_type_inference("Input type was null");
   }
 
-  if (input_type->value_case() != TypeProto::kTensorType) {
+  if (!input_type->has_tensor_type()) {
     fail_type_inference(
         "Input was expected to have tensor type. Got ",
         input_type->value_case());
@@ -276,11 +276,11 @@ inline void appendSingleDimCopiedFromInputTypeToOutputType(
     size_t fromDimIndex) {
   auto output_type = ctx.getOutputType(outputIndex);
   auto input_type = ctx.getInputType(inputIndex);
-  if (TypeProto::kTensorType != output_type->value_case()) {
+  if (!output_type->has_tensor_type()) {
     fail_type_inference(
         "Output ", outputIndex, " expected to have tensor type");
   }
-  if (TypeProto::kTensorType != input_type->value_case()) {
+  if (!input_type->has_tensor_type()) {
     fail_type_inference("Input ", inputIndex, " expected to have tensor type");
   }
   auto* dim = ctx.getOutputType(outputIndex)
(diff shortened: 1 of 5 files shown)


onnx/defs/shape_inference.h
```diff
@@ -313,11 +313,15 @@ inline void propagateShapeFromInputToOutput(
 }
 
 inline void propagateShapeAndTypeFromFirstInput(InferenceContext& ctx) {
-  propagateElemTypeFromInputToOutput(ctx, 0, 0);
+  for (int i = 0; i < static_cast<int>(ctx.getNumOutputs()); ++i) {
```
gramalingam: This changes the behavior for all existing uses of this function. I think it would be better to explicitly indicate whether we want to propagage just 0 => 0 or 0 => all. What if we add a Boolean parameter to the function (with a default value) to choose between the two? Or, alternatively, take an integer parameter N to indicate propagation from first input to the first N outputs (with a negative value to indicate all)?

Flamefire: In practice this only affects `BatchNormalization` which has multiple outputs and uses this. All other uses have only 1 output.

What about: Have this function take an optional param **to**  which output it should be propagated. Additionally add another function `propagateShapeAndTypeFromFirstInputToAllOutputs` which will be used by `Dropout`. 

Reasoning: Using parameters the meaning gets obscured.   `propagateShapeAndTypeFromFirstInput(ctx, 2)` can probably still be understood as `to output 2` and this is more useful than an "all-or-nothing" parameter. The "all" function can then be used directly assuming this might be quite common (more common than to only a subset which can still be implemented by means of the base function)

gramalingam: That sounds fine.

Flamefire: Ok done. Had to do some minor changes to that, so best see the commits and messages please

onnx/defs/shape_inference.h
```diff
@@ -312,12 +313,24 @@ inline void propagateShapeFromInputToOutput(
       ctx.getInputType(inputIndex)->tensor_type().shape();
 }
 
-inline void propagateShapeAndTypeFromFirstInput(InferenceContext& ctx) {
-  propagateElemTypeFromInputToOutput(ctx, 0, 0);
-  if (!hasNInputShapes(ctx, 1)) {
+inline void propagateShapeAndTypeFromInputToOutput(InferenceContext& ctx, size_t inputIndex, size_t outputIndex) {
+  propagateElemTypeFromInputToOutput(ctx, inputIndex, outputIndex);
```
houseroad: Check whether we have the type information for input[inputIndex] first.

onnx/defs/shape_inference.h
```diff
@@ -312,12 +313,24 @@ inline void propagateShapeFromInputToOutput(
...
+  if (!hasNInputShapes(ctx, inputIndex + 1)) {
```
houseroad: You should use `hasInputShape` here. Check how `hasNInputShapes` is implemented here: https://github.com/onnx/onnx/blob/master/onnx/defs/shape_inference.h#L266

onnx/defs/shape_inference.h
Flamefire: According to https://github.com/onnx/onnx/blob/765f5ee823a67a866f4bd28a9860e81f3c811ce8/onnx/onnx.proto#L425 there should be. If there isn't the function will throw which I deem OK

onnx/defs/shape_inference.h
Flamefire: Thanks: 6aa269b

onnx/defs/shape_inference.h
houseroad: https://github.com/onnx/onnx/blob/765f5ee823a67a866f4bd28a9860e81f3c811ce8/onnx/onnx.proto#L143, we may not have valueinfo for some intermediate value. In that case, the type information is not available.

Flamefire: And this shall be changed from a hard failure (https://github.com/onnx/onnx/blob/84a0441ae28795a928005863dc142bee81827566/onnx/defs/shape_inference.h#L212) to a soft failure (do nothing)?   
Just asking again as this changes existing behaviour.

houseroad: if you cannot do the shape inference, it's not a good idea to throw exception, instead you should just skip it. This logic is just to avoid segfault.

Issue #1725, opened 2019-01-10
Impossible to create Tensor(string) as input in python tests


yuslepukhin:
When creating an input of strings for `expect` function `onnx.mapping.NP_TYPE_TO_TENSOR_TYPE` directs us to create an array of `np.dtype('str')` which is the same thing as `np.string_`.

However, the code errors out producing the following output:

>  File "d:\dev\onnx\onnx\backend\test\case\node\__init__.py", line 35, in expect
    for arr, arr_name in zip(inputs, present_inputs)]
  File "d:\dev\onnx\onnx\backend\test\case\node\__init__.py", line 35, in <listcomp>
    for arr, arr_name in zip(inputs, present_inputs)]
  File "d:\dev\onnx\onnx\backend\test\case\node\__init__.py", line 23, in _extract_value_info
    elem_type=onnx.mapping.NP_TYPE_TO_TENSOR_TYPE[arr.dtype],
KeyError: dtype('<U9')

PR #1734, opened 2019-01-12
Add a capability to input/output unicode strings

yuslepukhin:
Adjust internal encoding from ascii to UTF-8 so
  UTF-8 encoding is fed to the actual stirng processing
  operators.

Address issue https://github.com/onnx/onnx/issues/1725


yuslepukhin:
Cc: @zrphercule

onnx/mapping.py
```diff
@@ -4,8 +4,23 @@
 from __future__ import unicode_literals
 
 from onnx import TensorProto
+from typing import Text, Any
 import numpy as np  # type: ignore
 
+
+class StringHolder:
```
houseroad: Please add some comments to explain the purpose and usage of this class.

houseroad: Feel free to add it in another PR.

PR #1745, opened 2019-01-16
Add stringnormalizer operator to ONNX

yuslepukhin:
String normalizer is found in the following frameworks:
* [StopWordsRemoverTransform](https://github.com/dotnet/machinelearning/blob/fe71bb82b8ef70e20f98d6fa77f246daba6f7a51/src/Microsoft.ML.Transforms/Text/StopWordsRemoverTransform.cs#L950) in ML.NET
* [CounterVectorizer/TfIdfVectorizer in scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/f0ab589f1541b1ca4570177d93fd7979613497e3/sklearn/feature_extraction/text.py#L128)

StringNormalization performs string operations for basic cleaning. This operator has only one input (denoted by X) and only one output (denoted by Y). This operator first examines the elements in the X, and remove elements specified in "stopwords" attribute. Note that an implementation should sequentially remove "stopwords[0]," then "stopwords[1]," and so on. After removing stop words, the intermediate result can be further lowercased, uppercased, or just returned depending the "casechangeaction" attribute.


yuslepukhin:
Cc: @houseroad @HectorSVC @yuanbyu @gramalingam @wschin @linkerzhang

onnx/defs/nn/defs.cc
```diff
@@ -1416,4 +1402,67 @@ ONNX_OPERATOR_SET_SCHEMA(
         .SetDoc(LRN_ver1_doc)
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
+static const char* StringNormalizer_ver9_doc = R"DOC(
+[optional] Step1: Remove elements in X if they match any of the stop words so
```
wschin: Here is an `optional Step 1` but I don't see a `Step 2`. Because I was involved in its original design, I can say its step 2 is `changing the case according to casechangeaction attribute. After reading this again, I feel it actually consists of some independent operators:
- Stop words remover.
- Lower, Upper
Do you think we should break it down?

onnx/defs/nn/defs.cc
```diff
@@ -345,16 +344,17 @@ void maxUnpoolShapeInference(InferenceContext& ctx) {
...
-              "'output_shape' must be rank 1 tensor.");
+        fail_type_inference("'output_shape' must be rank 1 tensor.");
       }
       if (output_shape.dim((int)0).has_dim_value() &&
-          static_cast<int>(output_shape.dim((int)0).dim_value()) != input_shape.dim_size()) {
-          fail_shape_inference(
-                  "'output_shape' must have same number of elements as the shape of input tensor X.");
+          static_cast<int>(output_shape.dim((int)0).dim_value()) !=
+              input_shape.dim_size()) {
+        fail_shape_inference(
```
linkerzhang: please run the clang-format.

onnx/defs/nn/defs.cc
linkerzhang: I think it's better to have the two covered in one op. Covering the two in one op does not add much complexity but help performance.

onnx/defs/nn/defs.cc
```diff
@@ -1416,4 +1402,68 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    9,
+    OpSchema()
+        .Input(0, "X", "Strings to normalize", "T")
+        .Output(0, "Y", "Normalized strings", "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(string)"},
+            "Input/Output is a UTF-8 string tensor")
+        .Attr(
+            "casechangeaction",
```
linkerzhang: do you want to make them as "optional" attribute with default value? say, casechangeaction with default value "NONE"; and is_case_sensitive with default value "false".

onnx/defs/nn/defs.cc
```diff
@@ -1416,4 +1402,68 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            // Last axis dimension is unknown if we have stop-words since we do
+            // not know how many stop-words are dropped
+            if (dim_size == 0) {
+              output_shape.add_dim();
+            } else if (dim_size == 1) {
+              // Copy C if no stopwords otherwise unknown
+              output_shape.add_dim();
+            } else if (dim_size == 2) {
+              // Copy B-dim
+              *output_shape.add_dim() = input_shape.dim(0);
```
linkerzhang: In this case, the first dim value has to be "1", right? per the description above. if yes, check it please.

onnx/defs/nn/defs.cc
```diff
@@ -1416,4 +1402,68 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          auto output_elem_type = ctx.getOutputType(0)->mutable_tensor_type();
+          output_elem_type->set_elem_type(TensorProto::STRING);
+          if (hasInputShape(ctx, 0)) {
+            TensorShapeProto output_shape;
+            auto& input_shape = ctx.getInputType(0)->tensor_type().shape();
+            auto dim_size = input_shape.dim_size();
+            // Last axis dimension is unknown if we have stop-words since we do
+            // not know how many stop-words are dropped
+            if (dim_size == 0) {
+              output_shape.add_dim();
```
linkerzhang: this is an invalid case?

onnx/defs/nn/defs.cc
```diff
@@ -1416,4 +1402,68 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+[optional] Step1: Remove elements in X if they match any of the stop words so
+ that the output tensor will not contain any stop words. This operator only accepts [C]-
+ and [1, C]-tensors. If all elements in X are dropped, the output will be the default value
+ of string tensor with shape [1] if input shape is [C] and shape [1, 1] if input shape is [1, C].
+[optional] Step2: Lower all characters (if action is LOWER) in X or capitalize them (when action is UPPER).
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    StringNormalizer,
+    9,
```
linkerzhang: this should be in opset 10 now. 

We just release onnx 1.4, together with opset 9.

yuslepukhin:
CI fails on Python 2.7 quirks. I have to jump through the hoops bc its inferior Unicode support and it is reaching the end of its life. Should we continue doing CI on it?

onnx/defs/nn/defs.cc
yuslepukhin: In case of 2-dim the first value can be an int, dim_param or nothing, so I simply copy it.

onnx/defs/nn/defs.cc
yuslepukhin: Ok.

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,71 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "casechangeaction",
+            "string enum that cases output to be lowercased/uppercases/unchanged. Valid values are \"LOWER\", \"UPPER\", \"NONE\"",
+            AttributeProto::STRING)
+        .Attr(
+            "is_case_sensitive",
+            "Boolean. Whether the identification of stop words in X is case-sensitive.",
+            AttributeProto::INT)
+        .Attr(
+            "stopwords",
+            "List of stop words",
```
wschin:
```suggestion
            "List of stop words. If not set, no work would be removed from the input tensor.",
```

onnx/defs/nn/defs.cc
wschin: Sound reasonable to me. :)

yuslepukhin:
> 
> 
> Guess we should better have a model test as well?

@zrphercule Added model tests

VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.4.0
+1.4.1
```
linkerzhang: This should be changed by mistake?

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,73 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .TypeConstraint(
+            "T",
+            {"tensor(string)"},
+            "Input/Output is a UTF-8 string tensor")
+        .Attr(
+            "casechangeaction",
+            "string enum that cases output to be lowercased/uppercases/unchanged. Valid values are \"LOWER\", \"UPPER\", \"NONE\"",
+            AttributeProto::STRING)
+        .Attr(
+            "is_case_sensitive",
```
wschin: No sure if we want
```suggestion
            "iscasesensitive",
```
@linkerzhang, any preference?

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,73 @@ ONNX_OPERATOR_SET_SCHEMA(
         })
         .SetDoc(TfIdfVectorizer_ver9_doc));
 
+static const char* StringNormalizer_ver10_doc = R"DOC(
+StringNormalization performs string operations for basic cleaning. 
+This operator has only one input (denoted by X) and only one output 
+(denoted by Y). This operator first examines the elements in the X, 
+and remove elements specified in "stopwords" attribute. 
+Note that an implementation should sequentially remove "stopwords[0]," then "stopwords[1],"
```
linkerzhang: "Note that an implementation should sequentially remove "stopwords[0]," then "stopwords[1]," "
why? is this because that the stop word may not be unigram?

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,73 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+Note that an implementation should sequentially remove "stopwords[0]," then "stopwords[1]," 
+and so on. After removing stop words, the intermediate result can be further lowercased, 
+uppercased, or just returned depending the "casechangeaction" attribute.
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    StringNormalizer,
+    10,
+    OpSchema()
+        .Input(0, "X", "Strings to normalize", "T")
```
linkerzhang: Change "T" to "tensor(string)" since there's only one option for T, so that you don't need to specify .TypeConstraint(...), which simplifies a little bit.

onnx/defs/nn/defs.cc
linkerzhang: attribute named as "case_change_action" (consistent with the other attribute)? btw, how do you think of my comments of making them as an optional attribute with default value?

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,73 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            {"tensor(string)"},
+            "Input/Output is a UTF-8 string tensor")
+        .Attr(
+            "casechangeaction",
+            "string enum that cases output to be lowercased/uppercases/unchanged. Valid values are \"LOWER\", \"UPPER\", \"NONE\"",
+            AttributeProto::STRING)
+        .Attr(
+            "is_case_sensitive",
+            "Boolean. Whether the identification of stop words in X is case-sensitive.",
+            AttributeProto::INT)
```
linkerzhang: same comment: I'd suggest to make this attribute as optional with default value.

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,73 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "stopwords",
+            "List of stop words. If not set, no work would be removed from X.",
+            AttributeProto::STRINGS,
+            OPTIONAL)
+        .Attr(
+            "locale",
+            "Environment dependent string that denotes the locale according to which output strings needs to be upper/lowercased."
+            "Default en_US or platform specific equivalent",
+            AttributeProto::STRING,
+            OPTIONAL)
```
linkerzhang: specify the default value via API explicitly please, though I see you specified default value "en-US" in the description.

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,73 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .SetDoc(StringNormalizer_ver10_doc)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          auto output_elem_type = ctx.getOutputType(0)->mutable_tensor_type();
+          output_elem_type->set_elem_type(TensorProto::STRING);
+          if (hasInputShape(ctx, 0)) {
```
linkerzhang: small comment: 

change to if(!hasInputShape(ctx, 0)) {...} to avoid big if block?

onnx/defs/nn/defs.cc
linkerzhang: I suggest the other way, see my comments inline. :)

VERSION_NUMBER
yuslepukhin: I have not changed this, only merged the master

onnx/defs/nn/defs.cc
yuslepukhin: The default value is platform dependent and, therefore, can not be specified uniformly.

onnx/defs/nn/defs.cc
wschin: Right. It's for avoiding undefined behavior in that case.

onnx/defs/nn/defs.cc
yuslepukhin: I will make this change

onnx/defs/nn/defs.cc
linkerzhang: I discussed with Dmitri offline. Looks to us the order/sequantial does not matter here. Say, input is "x", "y", "z", and stop words are "xy", "yz", then nothing will be removed from input.

wschin: Ah, yeah. Makes sense. Thank you.

linkerzhang:
@zrphercule  any more comments please? Thank you!

zrphercule:
@houseroad Any comments? Ke is targeting to merge it as soon as possible, could you plz give a verify review of it? Thanks!

linkerzhang:
@houseroad  any comments?

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,74 @@ ONNX_OPERATOR_SET_SCHEMA(
         })
         .SetDoc(TfIdfVectorizer_ver9_doc));
 
+static const char* StringNormalizer_ver10_doc = R"DOC(
+StringNormalization performs string operations for basic cleaning. 
+This operator has only one input (denoted by X) and only one output 
+(denoted by Y). This operator first examines the elements in the X, 
+and remove elements specified in "stopwords" attribute.
```
houseroad: removes

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,74 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            std::string("NONE"))
+        .Attr(
+            std::string("is_case_sensitive"),
+            std::string(
+                "Boolean. Whether the identification of stop words in X is case-sensitive. Default is false"),
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
+        .Attr(
+            "stopwords",
+            "List of stop words. If not set, no work would be removed from X.",
```
houseroad: word

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,74 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          // not know how many stop-words are dropped
+          if (dim_size == 1) {
+            // Unknown output dimension
+            output_shape.add_dim();
+          } else if (dim_size == 2) {
+            // Copy B-dim
+            auto& b_dim = input_shape.dim(0);
+            if (b_dim.has_dim_value() && b_dim.dim_value() != 1) {
+              fail_shape_inference(
+                  "Input shape must have either [C], [?], [1,C] or [?,C] dimensions where C > 0");
```
houseroad: I am confused here... What is the expected shape of input?

yuslepukhin: Yes, this needs to be fixed. B dimension *must* have dim_value() and it has to be 1.
The error message *must* read "Input shape must have either [C] or [1,C] where C > 0"

yuslepukhin:
One of the Travises failed indicating some issue with the build itself.

onnx/defs/nn/defs.cc
yuslepukhin: @houseroad Pls, take another look

onnx/defs/nn/defs.cc
```diff
@@ -1542,4 +1534,74 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+and removes elements specified in "stopwords" attribute. 
+After removing stop words, the intermediate result can be further lowercased, 
+uppercased, or just returned depending the "case_change_action" attribute.
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    StringNormalizer,
+    10,
+    OpSchema()
+        .Input(0, "X", "UTF-8 strings to normalize", "tensor(string)")
```
houseroad: Shall we explicitly say that input's shape should [C] and [1, C]? Also, I am wondering why do we need to support [1, C] shape. Sounds a bit redundant to me. Since we always add an unsqueeze before.

onnx/numpy_helper.py
```diff
@@ -83,15 +83,17 @@ def from_array(arr, name=None):  # type: (np.ndarray[Any], Optional[Text]) -> Te
     if arr.dtype == np.object:
         # Special care for strings.
         tensor.data_type = mapping.NP_TYPE_TO_TENSOR_TYPE[arr.dtype]
-        for e in arr:
+        flat_array = arr.flatten()
```
houseroad: why do we need flatten here? We already handle the np.ndarray case already. (although, we probably should throw exception if it's 3d string array.

yuslepukhin: @houseroad Current infrastructure does not handle strings input at all. I tried to avoid investing too much time in it. This line was introduced to accommodate another test Cast which I had to change as well.

houseroad: Since it's not handled appropriately, please add some comments on why flatten is called here, and leave a TODO mark here.

onnx/defs/nn/defs.cc
wschin: In traditional ML pipelines in, for example, scikit-learn and Core ML, batch size 1 is very common. It's ok to remove but I'd like to support it to avoid adding extra ops. In addition, the reason that we can't have batch size > 1 is the lack of sparse tensor. This restriction will be eventually removed, I believe.

houseroad: Okay, let's keep it as it is now. But please explicitly describe what input shape will be accepted.

yuslepukhin: So the doc string clearly states that but I will also mention this in the input spec.

yuslepukhin:
Travis fails on git issue

yuslepukhin:
@houseroad Do you have any outstanding issues?

PR #1748, opened 2019-01-18
Handle negative axis in scan shape inference

gramalingam:
(Follow up to PR https://github.com/onnx/onnx/pull/1737) 
Extend shape inference to handle negative axes. Add shape inference test with negative axes.


gramalingam:
@houseroad this addresses a couple of comments on PR 1737, to handle negative axes in the shape inference of the scan operator. Sorry I couldn't get it into PR 1737 before it was merged in.

skottmckay:
Looks good to me. 

Also includes a bugfix when setting up the inputs to the subgraph inferencing where it was incorrectly starting at input 1 (which is correct for Scan 8 to skip the sequence_lens input, but not for Scan 9).

linkerzhang:
@houseroad please kindly review, though I'm assuming that this change should be good as it only adds more coverage for shape inference function and adds more test cases accordingly.

houseroad:
The code randomly hangs on some CI instances.

PR #1773, opened 2019-01-25
Support down sampling for Upsample with scales < 1.

linkerzhang:
This is to support down sampling, to relax the restrictions that "scales should be greater or equal to 1".


linkerzhang:
@houseroad please comment on this requirement/PR to support down sampling. It's verified that using scales < 1 works good.

linkerzhang:
Something wrong with doc generation, will fix.

linkerzhang:
@houseroad @zrphercule any comments on this change please?

onnx/backend/test/case/node/resize.py
```diff
@@ -0,0 +1,62 @@
...
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+class Resize(Base):
+
+    @staticmethod
+    def export_upsample_nearest():  # type: () -> None
```
houseroad: Can we also add some bilinear examples?

houseroad:
This PR also needs a corresponding PR in PyTorch, cc: @spandantiwari could you help on this? Thanks.

spandantiwari:
@houseroad @linkerzhang - sure. Let me send a fix to PT. I will let you know

spandantiwari:
I have started a PyTorch PR to disable the failing tests https://github.com/pytorch/pytorch/pull/17696. Once that is merged we can trigger CI build again to move this forward.

spandantiwari:
@linkerzhang - the PyTorch PR is in. We can try another CI run.

onnx/backend/test/case/node/resize.py
linkerzhang: added.

linkerzhang: added.

linkerzhang:
@houseroad  any more comments please? Thank you!

onnx/defs/tensor/old.cc
```diff
@@ -396,6 +396,65 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "Constrain input 'X' and output 'Y' to all tensor types.")
+        .SetDoc(Upsample_ver9_doc)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          if (!hasNInputShapes(ctx, 1)) {
+            return;
+          }
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          auto& input_shape = getInputShape(ctx, 0);
+          auto* output_shape = getOutputShape(ctx, 0);
+          output_shape->clear_dim();
```
houseroad: Can we check whether output_shape already has enough information? 

If so, we don't need to erase the data. Instead, we just finish the shape inference here.

linkerzhang: @houseroad  I'd suggest the check happens outside when running shape inference, otherwise, this kind of check needs happening for every shape inference function implementation. Make sense?

PR #1776, opened 2019-01-30
Fix Cast testcase data

raymondxyang:
Fix testcase data for Cast related to utf8 encode handling


raymondxyang:
I pushed the fix in generating scripts in this PR.

raymondxyang:
Me and Dimitri were working on the same script during the same time, so I mistakenly implemented the generating function for the string tensor. Just found out and fixed it.

raymondxyang:
oh u mean the second push? I add the wrong generated version so yeah it was a wrong submit

PR #1757, opened 2019-01-22
Document Facebook's ONNXIFI extensions

Maratyszcza:
- "onnx_extension_function" extension and onnxGetExtensionFunctionAddress function
- "fb_clone_graph" extension and onnxCloneGraphFB function
- "ext_alternative_model_graph" extension with
onnxGetBackendCompatibilityEXT function,
ONNXIFI_GRAPH_PROPERTY_MODEL_FORMAT enumeration value (for auxGraphProperties
key), and ONNXIFI_GRAPH_FORMAT_ONNX enumeration value (for
auxGraphProperties value)
- "fb_caffe2_model" extension with ONNXIFI_GRAPH_FORMAT_CAFFE2
enumeration value for ONNXIFI_GRAPH_PROPERTY_MODEL_FORMAT,
ONNXIFI_TAG_CAFFE2_QTENSOR_DESCRIPTOR_V1 tag, and
onnxCaffe2QTensorDescriptorV1 structure.

PR #1781, opened 2019-01-31
[ONNXIFI]Add extension of onnxSetIOAndRunGraph

zrphercule:
Today we have onnxSetGraphIO (https://github.com/onnx/onnx/blob/9a26a21a5b8dc8ff35ca1fffb84338fe250603f9/onnx/onnxifi.h#L1874) and onnxRunGraph (https://github.com/onnx/onnx/blob/9a26a21a5b8dc8ff35ca1fffb84338fe250603f9/onnx/onnxifi.h#L1973) as two separate functions, which makes it difficult to do atomic evaluation. We would like to have a new interface onnxSetIOAndRunGraph() which basically combines two.

### DIFF
--- a/onnx/onnxifi_ext.h
+++ b/onnx/onnxifi_ext.h
@@ -57,8 +57,9 @@ typedef ONNXIFI_CHECK_RESULT onnxStatus
  *                                            the backend was disconnected or
  *                                            uninstalled from the system.
  */
+
 ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
-  onnxGetExtensionFunctionAddress(
+onnxGetExtensionFunctionAddress(
     onnxBackendID backendID,
     const char* name,
     onnxExtensionFunctionPointer* function);
@@ -71,14 +72,16 @@ typedef ONNXIFI_CHECK_RESULT onnxStatus
     const onnxTensorDescriptorV1* inputDescriptors,
     uint32_t outputsCount,
     const onnxTensorDescriptorV1* outputDescriptors,
-    const onnxMemoryFenceV1* inputFence,
     onnxMemoryFenceV1* outputFence);
 
-/* Extension functions */
 /**
  * A combination of onnxSetIO and onnxRunGraph, functionally equals to first run
- * onnxSetIO(graph, inputsCount, inputDescriptors, outputsCount, outputDescriptors),
- * then run onnxRunGraph(graph, inputFence, outputFence)
+ * onnxSetIO(graph, inputsCount, inputDescriptors, outputsCount,
+ * outputDescriptors), then run onnxRunGraph(graph, inputFence, outputFence)
+ *
+ * As two separate functions, it is difficult to do atomic evaluation.
+ * Therefore, we would like to unify this process and make it evaluable.
+ *
  * @param graph - graph handle created by onnxInitGraph.
  * @param inputsCount - number of elements in the inputDescriptors array.
  * @param[in] inputDescriptors - descriptors of input tensors for the graph.
@@ -211,18 +214,16 @@ typedef ONNXIFI_CHECK_RESULT onnxStatus
  *                                       internal error.
  */
 
- ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
-    onnxSetIOAndRunGraph(
-        onnxGraph graph,
-        uint32_t inputsCount,
-        const onnxTensorDescriptorV1* inputDescriptors,
-        uint32_t outputsCount,
-        const onnxTensorDescriptorV1* outputDescriptors,
-        const onnxMemoryFenceV1* inputFence,
-        onnxMemoryFenceV1* outputFence);
+ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI onnxSetIOAndRunGraph(
+    onnxGraph graph,
+    uint32_t inputsCount,
+    const onnxTensorDescriptorV1* inputDescriptors,
+    uint32_t outputsCount,
+    const onnxTensorDescriptorV1* outputDescriptors,
+    onnxMemoryFenceV1* outputFence);
 
 #ifdef __cplusplus
 } /* extern "C" */
 #endif
 
-#endif
+#endif /* !defined(ONNXIFI_EXT_H) */
--- a/onnx/onnxifi_ext.h
+++ b/onnx/onnxifi_ext.h
@@ -72,7 +72,6 @@ typedef ONNXIFI_CHECK_RESULT onnxStatus(
     const onnxTensorDescriptorV1* inputDescriptors,
     uint32_t outputsCount,
     const onnxTensorDescriptorV1* outputDescriptors,
-    const onnxMemoryFenceV1* inputFence,
     onnxMemoryFenceV1* outputFence);
 
 /**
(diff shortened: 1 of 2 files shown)


onnx/onnxifi_ext.h
```diff
@@ -0,0 +1,163 @@
+#ifndef ONNXIFI_EXT_H
+#define ONNXIFI_EXT_H 1
```
rdzhabarov: #define ONNXIFI_EXT_H

should be enough

rdzhabarov:
cc: @jackm321 who is going to wire that on glow end

onnx/onnxifi_ext.h
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * The caller should convert this generic function pointer to the function
+ * pointer specific for an extension function type.
+ */
+typedef onnxStatus (ONNXIFI_ABI* onnxExtensionFunctionPointer)(void);
+
+/**
+ * A combination of onnxSetIO and onnxRunGraph, functionally equals to first run
+ * onnxSetIO(graph, inputsCount, inputDescriptors, outputsCount, outputDescriptors),
+ * then run onnxRunGraph(graph, inputFence, outputFence)
```
rdzhabarov: I'd mention why we need this to support atomicity

onnx/onnxifi_ext.h
```diff
@@ -0,0 +1,163 @@
...
+ *                               If inputsCount is non-zero, inputDescriptors
+ *                               pointer must be non-NULL.
+ * @param outputsCount - number of elements in the outputDescriptors array.
+ *                       Must be greater than zero.
+ * @param[in] outputDescriptors - descriptors of output tensors for the graph.
+ *                                outputDescriptors pointer must be non-NULL.
+ *                                Elements of this array must provide a location
+ *                                for each ValueInfoProto.name listed in
+ *                                ModelProto.graph.output of the ONNX graph.
+ * @param[in] inputFence - synchronization primitive that signals when graph
```
rdzhabarov: do we need to care about this? or should inputs be available right away on the `onnxSetIOAndRunGraph` call

onnx/onnxifi_ext.h
```diff
@@ -0,0 +1,163 @@
...
+ * @retval ONNXIFI_STATUS_BACKEND_UNAVAILABLE The function call failed because
+ *                                            the backend was disconnected or
+ *                                            uninstalled from the system.
+ * @retval ONNXIFI_STATUS_INTERNAL_ERROR The function call failed because the
+ *                                       backend experienced an unrecovered
+ *                                       internal error.
+ */
+
+ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
+	onnxSetIOAndRunGraph(
```
rdzhabarov: this looks good to me

onnx/onnxifi_ext.h
```diff
@@ -0,0 +1,163 @@
...
+ *                                            tensor locations.
+ * @retval ONNXIFI_STATUS_BACKEND_UNAVAILABLE The function call failed because
+ *                                            the backend was disconnected or
+ *                                            uninstalled from the system.
+ * @retval ONNXIFI_STATUS_INTERNAL_ERROR The function call failed because the
+ *                                       backend experienced an unrecovered
+ *                                       internal error.
+ */
+
+ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
```
yinghai: You need a typedef to define a new function type. Check https://github.com/onnx/onnx/pull/1757/files#diff-8312e1aaff6ed2027647302d102f0d44R271 for an example.

onnx/onnxifi_ext.h
jackm321: Yes it would be cool if we could get rid of this also if possible

onnx/onnxifi_ext.h
```diff
@@ -0,0 +1,163 @@
...
+extern "C" {
+#endif
+
+/**
+ * Generic ONNXIFI extension function pointer.
+ *
+ * The caller should convert this generic function pointer to the function
+ * pointer specific for an extension function type.
+ */
+typedef onnxStatus (ONNXIFI_ABI* onnxExtensionFunctionPointer)(void);
```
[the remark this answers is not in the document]
zrphercule: Oh yes you are right

onnx/onnxifi_ext.h
zrphercule: Actually I am not quite sure about what is the meaning of this, just copying codes from Marat. Mind telling me why we should add this macro?

rdzhabarov: this is to avoid redefinition in the same translation unit. So when you define those things `ONNXIFI_EXT_H`, header file will be included just once.

onnx/onnxifi_ext.h
rdzhabarov: or if we have that param for general purpose in this API, we can make C2/Glow to make sure that Glow does not need to wait on that event.
That option also works well.

zrphercule: Quite a good point. My thought is since this is a general API, we should better give less restrictions or potential bug maker to our backend, so I would like to remove this InputFence from this API for now. If adding this param become very important in the future, I guess we can add it back then. Agree?

rdzhabarov: That sounds good to me.

onnx/onnxifi_ext.h
```diff
@@ -0,0 +1,229 @@
...
+ *                               If inputsCount is non-zero, inputDescriptors
+ *                               pointer must be non-NULL.
+ * @param outputsCount - number of elements in the outputDescriptors array.
+ *                       Must be greater than zero.
+ * @param[in] outputDescriptors - descriptors of output tensors for the graph.
+ *                                outputDescriptors pointer must be non-NULL.
+ *                                Elements of this array must provide a location
+ *                                for each ValueInfoProto.name listed in
+ *                                ModelProto.graph.output of the ONNX graph.
+ * @param[in] inputFence - synchronization primitive that signals when graph
```
rdzhabarov: do not forget to remove this from the comment, since it was removed from the function declaration.

zrphercule: Yes!

jackm321:
Looks good! `inputFence` looks like it's still mentioned in the documentation but should not be.

zrphercule:
> Looks good! `inputFence` looks like it's still mentioned in the documentation but should not be.

Yeah, once CI is passed I will update the documentation. I have been fighting with CI for a while, and finally with help of yinghai this problem is solved I believe.

onnx/onnxifi_ext.h
```diff
@@ -0,0 +1,226 @@
...
+ *                                            tensor locations.
+ * @retval ONNXIFI_STATUS_BACKEND_UNAVAILABLE The function call failed because
+ *                                            the backend was disconnected or
+ *                                            uninstalled from the system.
+ * @retval ONNXIFI_STATUS_INTERNAL_ERROR The function call failed because the
+ *                                       backend experienced an unrecovered
+ *                                       internal error.
+ */
+
+ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI onnxSetIOAndRunGraph(
```
yinghai: Add a typedef for this function pointer too?

zrphercule: I did, see https://github.com/onnx/onnx/pull/1781/files#diff-8312e1aaff6ed2027647302d102f0d44R68

PR #1794, opened 2019-02-06
Enfore Protobuf 2.6.1 at cmake time

WilliamTambellini:


CMakeLists.txt
```diff
@@ -101,7 +101,9 @@ else()
     find_path(Protobuf_INCLUDE_DIR google/protobuf/service.h
               PATHS ${_PROTOBUF_INSTALL_PREFIX}/include
               NO_DEFAULT_PATH)
-    find_package(Protobuf REQUIRED)
+    # Protobuf 2.5.0 is no more compatible with onnx, and 2.7.0 not yet released.
+    # https://github.com/protocolbuffers/protobuf/releases
+    find_package(Protobuf 2.6.1 REQUIRED)
```
houseroad: The version parameter in `find_package` is tricky, it's defined package by package. I am not sure if it is a good idea to use it here. cc: @zrphercule @bddppq @snnn

yuslepukhin: If you built protobuf from source then 2.6.1 is good. But when it come to packages the version is all over the place. Some Conda packages on Windows are broken, so we recently updated instructions to use 3.9.2 because it works. I think updating instructions is a better way of addressing it.

yuslepukhin:
This would conflict with the most recent build instructions bc it would take away package version flexibility. The flexibility is needed for two reasons:

- Some Conda packages versions are built improperly. Some versions are build in a form of a static lib, others are in the form of shared lib. We want to use the versions that work.
- Some users have reported that they want to use ONNX with products that are already using different version of protobuf. For them, rebuilding ONNX with different version of protobuf solves the issue.

prasanthpul:
Closing this old PR. Please reopen if necessary

PR #1796, opened 2019-02-07
[ONNXIFI]Add extension to be implementable

zrphercule:
In this PR, we give a way of implementing the new onnxifi_extension, and an example of using and testing it using onnxifi dummy backend.


zrphercule:
cc @jackm321

onnx/onnxifi_dummy.c
```diff
@@ -136,3 +141,25 @@ ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
...
+/*
+ * ONNXIFI Extension Functions
+ */
+
+ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
+onnxGetExtensionFunctionAddress(
+    onnxBackendID backendID,
+    const char* name,
+    onnxExtensionFunctionPointer* function) {
+  return ONNXIFI_STATUS_SUCCESS;
```
rdzhabarov: could we test that actual function will be assigned when name == "onnxSetIOAndRunGraph"?

zrphercule: Yeah, that's exactly I am doing... The main reason why this is "Unfinished" hahah

onnx/onnxifi_loader.c
```diff
@@ -7,6 +7,11 @@
 #include <dlfcn.h>
 #endif
 
+/*
+ * temp enable ONNXIFI_ENABLE_EXT by all means
```
houseroad: i guess this is only for temporary test?

zrphercule: Yeah, I will add this macro in CMAKE, so it is not ready to be merged now.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -322,6 +322,17 @@ class ONNXCppDriverTest : public ::testing::TestWithParam<
                 true);
           }
         }
+        /*
+         * Examine functions in onnxifi_ext
+         */
+#ifdef ONNXIFI_ENABLE_EXT
```
yinghai: ENABLE_ONNXIFI_EXT might be more natural?

zrphercule: Well, since all onnxifi related macros are start with an "ONNXIFI_...", I wonder if we really want to change this.

zrphercule:
@yinghai @rdzhabarov  This is now ready to be merged. Noticing since we disabled onnxifi_dummy_backend in test driver, I will look back to the dummy backend part after this. But for now, this pr is okay to be merged.

yinghai:
CI failure?

onnx/onnxifi_dummy.c
```diff
@@ -136,3 +141,71 @@ ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
...
+/*
+ * ONNXIFI Extension Functions
+ */
+
+ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
+onnxGetExtensionFunctionAddress(
+    onnxBackendID backendID,
+    const char* name,
+    onnxExtensionFunctionPointer* function) {
+  /*		const int extension_function_number = 2;
```
yinghai: Where is `extension_function_number` defined? It is commented out here?

zrphercule: oh it's my bad, I thought I have already deleted this.

onnx/onnxifi_ext.h
```diff
@@ -7,6 +7,11 @@
 extern "C" {
 #endif
 
+static const int extension_function_number = 2;
+static const char extension_function_list[] =
```
yinghai: Why do we need to put these things here? The factor that we include "onnxifi_ext.h" also means we have access to these names, right? Such a table should be kept at the backend side IMO.

zrphercule: I also took some consideration when design it like this. My reason of putting these define here is we would like to make this two consts represent the super set of functions we support. Like, a backend may only support a few of functions in it, but still we want to know overall what is supported. Backend can have their own list and number if they need, or just use this two (like in dummy backend), meaning it support all fucntions.

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -322,6 +323,17 @@ class ONNXCppDriverTest : public ::testing::TestWithParam<
                 true);
           }
         }
+/*
+ * Examine functions in onnxifi_ext
+ */
+#ifdef ONNXIFI_ENABLE_EXT
+        onnxExtensionFunctionPointer* f;
+        ASSERT_TRUE(IsGtestAssertMemorySafeSuccess(
```
rdzhabarov: why not test `onnxSetIOAndRunGraph` function extension as part of the driver_test (given that the function is available)

onnx/onnxifi_dummy.c
```diff
@@ -136,3 +141,69 @@ ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
...
+    /* count mismatch */
+    if (*p == '\0') {
+      count++;
+    }
+    p++;
+  }
+
+  if (*function == NULL) {
+    return ONNXIFI_STATUS_UNIDENTIFIED_NAME;
+  } else {
```
rdzhabarov: this `else` branch is not needed.

onnx/onnxifi_ext.h
```diff
@@ -7,6 +7,11 @@
 extern "C" {
 #endif
 
+static const int ALL_EXT_FUNCTION_NUMBER = 2;
+static const char ALL_EXT_FUNCTION_LIST[] =
```
rdzhabarov: what's the reason to represent all functions in this long char str? and not
char *list[] = {"a", "b"} etc.

onnx/backend/test/cpp/driver_test.cc
zrphercule: That's exactly I would like to do. But the thing is since the driver test is down, we cannot test it anyway. Since I dont want to cause some potential bugs here, I will add related test in test driver once we fixed onnxifi driver test.

onnx/onnxifi_dummy.c
```diff
@@ -160,7 +160,8 @@ onnxGetExtensionFunctionAddress(
   }
   int count = 0;
   *function = NULL;
-  for (int i = 0; i < extension_function_number; i++) {
+  int i;
```
rdzhabarov: what's this :P

CMakeLists.txt
```diff
@@ -543,6 +544,11 @@ endif()
 
 # ---[ ONNXIFI dummy backend
 add_library(onnxifi_dummy SHARED onnx/onnxifi_dummy.c)
+
+if(ONNXIFI_ENABLE_EXT)
+	add_definitions(-DONNXIFI_ENABLE_EXT=ON)
```
rdzhabarov: indentation looks weird.

onnx/onnxifi_dummy.c
```diff
@@ -136,3 +141,52 @@ ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
...
+  if (name == NULL || function == NULL) {
+    return ONNXIFI_STATUS_INVALID_POINTER;
+  }
+  int count = 0;
+  *function = NULL;
+  int i;
+  for (i = 0; i < extension_function_number; i++) {
+    /* target function found */
+    if (strcmp(name, extension_function_list[i]) == 0) {
+      switch (count) {
```
rdzhabarov: that's incorrect, we have `i` for this.

onnx/onnxifi_dummy.c
zrphercule: It is because I just realize C lang does not accept "for (int i = 0; ...)".... T^T

onnx/onnxifi_dummy.c
```diff
@@ -146,7 +146,7 @@ onnxReleaseGraph(onnxGraph graph) {
  * ONNXIFI Extension Functions
  */
 
-const int extension_function_number = ALL_EXT_FUNCTION_NUMBER;
+const int extension_function_number = 2;
```
jackm321: What is this? Can you add a comment?

zrphercule:
@rdzhabarov mind if you unblock it and we merge it & update submodule in pytorch? Thanks!

onnx/onnxifi_dummy.c
jackm321: Hmm, what's wrong with that?

onnx/onnxifi_ext.h
```diff
@@ -7,6 +7,10 @@
 extern "C" {
 #endif
 
+static const int ALL_EXT_FUNCTION_NUMBER = 2;
```
jackm321: please comment these

zrphercule: ok..
the "for(int i = 0....)" thing is because, this is C, and for (int i = 0; ...) is not allowed in C...

jackm321: Ohh wow ok got it thanks

zrphercule:
@onnxbot retest this please

PR #1804, opened 2019-02-11
Adding Reverse op

pk-g:


onnx/defs/tensor/defs.cc
```diff
@@ -622,6 +622,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    OpSchema()
+        .SetDoc(Reverse_ver9_doc)
+        .Attr(
+            "axes",
+            "A list of integers indicating the axes to reverse. "
+            "Negative values mean counting dimensions from the back."
+            "By default, reverse over all axes of input dimensions",
+            AttributeProto::INTS,
+            OPTIONAL)
+        .Input(0, "inpu6", "Tensor of rank r >= 1.", "T")
```
gramalingam: inpu6 => input

onnx/defs/operator_sets.h
```diff
@@ -497,6 +497,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, Where);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, Cast);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, NonZero);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, TfIdfVectorizer);
+class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 9, Reverse);
```
linkerzhang: the operator set should be 10. 

9 was the one released in ONNX 1.4.

houseroad: Please use 10 instead.

pk-g:
Thanks for the review, updated the PR with addressed feedback.

onnx/backend/test/case/node/reverse.py
```diff
@@ -0,0 +1,60 @@
...
+        input = np.arange(6.0).reshape(2, 3)
+
+        node = onnx.helper.make_node(
+            'Reverse',
+            inputs=['input'],
+            outputs=['output'],
+            axes=[0]
+        )
+
+        output = np.flip(input)
```
gramalingam: Is this correct? Shouldn't the axes=0 be passed to np.flip?

pk-g: the default is to reverse on all axes.

pengwa:
shall we enhance this op to support reverse sequences of different lengths? in tf, it is called tf.reversesequence().

pk-g:
@pengwa  If you are referring to the case where sequential data of different lengths has been _padded_, it may be cleaner to have a separate op to cover such scenario.

pk-g:
@linkerzhang can you please review the updated PR so we can move ahead with this op? thanks.

docs/Changelog.md
```diff
@@ -9451,6 +9451,83 @@ This version of the operator has been available since version 9 of the default O
...
+
+#### Version
+
+This version of the operator has been available since version 10 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>axes</tt> : list of ints</dt>
+<dd>A list of integers indicating the axes to reverse. Negative values mean counting dimensions from the back.By default, reverse over all axes of input dimensions</dd>
```
hariharans29: Sorry for this clarification so late on - but I assume this operator doesn't support "implicit" copying of Tensors by specifying duplicates in axes ?  For example, would axes of [0,0] mean no reversing but force the tensor's buffer to be copied ? Or simply, don't allow duplicates in axes at all - just like the way tf operates - https://www.tensorflow.org/api_docs/python/tf/reverse ?

pk-g: Great question, this is more in line with the latter, no duplicate is expected here.

linkerzhang:
@pk-g can you update the PR by fixing the conflicts?

@houseroad any more comments on this op please?

Thank you!

houseroad:
I will do the review later this week, but please don't rush.

Will revert it here: https://github.com/onnx/onnx/pull/1882

Please resubmit the changes.

fdwr:
> Will revert it here: #1882
> 
> Please resubmit the changes.

Was this op reverted and then simply forgotten to re-add? It's found in the following libraries...

- PyTorch [flip](https://pytorch.org/docs/stable/generated/torch.flip.html)
TensorFlow [reverse](https://www.tensorflow.org/api_docs/python/tf/reverse)
- CoreML [tensor_transformation.reverse](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS17.tensor_transformation.reverse)
- TFLite [tfl.reverse_v2](https://www.tensorflow.org/mlir/tfl_ops#tflreverse_v2_tflreversev2op)
- StableHLO reverse
- tosa.reverse (mlir::tosa::ReverseOp)
- ANEURALNETWORKS_REVERSE
- PyTorch prims rev

...and we're looking to [add it to WebNN](https://github.com/webmachinelearning/webnn/issues/773) for a voice to text model. It *can* be achieved in ONNX via the overloaded interpretation of setting ONNX [Slice](https://onnx.ai/onnx/operators/onnx__Slice.html)'s to -1, but loses semantics and is more awkward (needing to specify `starts: [2147483647, 2147483647], ends: [-2147483648, -2147483648], steps: [-1,-1], axes: [0,1]` rather than just `axes: [0,1]`).

PR #1809, opened 2019-02-12
More extendable Runner

postrational:
Using `@staticmethods` and hardcoding the class name when using those methods (as in `Runner._prepare_model_data`) makes it more difficult to extend these classes and reuse them.

I would recommend switching from `@staticmethod` to `@classmethod` and from `Runner.method` to `cls.method`.

For now I'm proposing a smaller change to avoid hardcoding the class in one method invocation.

What do you think about this?


postrational:
@houseroad Let me know what you think.

postrational:
> I wonder what is the purpose to change methods like _prepare_model_data from static to a cls method? Thanks!

@zrphercule: The purpose would be to allow another library to use the ONNX test `Runner` class by inheriting from it, overriding some methods to modify behavior, but keeping the rest as is. 

When the methods are `@staticmethod` and are called from other methods as `Runner.method_name`, then the inheriting class can't override their behavior. They will always be called as `Runner.method_name`, never as `RunnerSubclass.method_name`.

Changing these to `@classmethod` would imply calling them as `cls.method_name`, so a subclass can override them.

I realize these are marked as `_hidden` methods, but it is still very useful to use the same testing logic as the ONNX library does in out customized tests.

houseroad:
@postrational I like this idea.

bddppq:
I suggest let's rename `_prepare_model_data` to `prepare_model_data` to make it more clear that it's ok to override.

postrational:
I addressed the comments - changed static to class methods and renamed `_prepare_model_data` to `prepare_model_data`.

Could someone restart the failed Travis jobs to see if they will pass?

onnx/backend/test/runner/__init__.py
```diff
@@ -163,8 +163,8 @@ def tests(self):  # type: () -> Type[unittest.TestCase]
                 setattr(tests, name, item.func)
         return tests
 
-    @staticmethod
-    def _assert_similar_outputs(ref_outputs, outputs, rtol, atol):  # type: (Sequence[Any], Sequence[Any], float, float) -> None
+    @classmethod
+    def _assert_similar_outputs(cls, ref_outputs, outputs, rtol, atol):  # type: (Sequence[Any], Sequence[Any], float, float) -> None
```
houseroad: since we already change this to classmethod, why not also turning it to public method?

onnx/backend/test/runner/__init__.py
```diff
@@ -174,9 +174,9 @@ def _assert_similar_outputs(ref_outputs, outputs, rtol, atol):  # type: (Sequenc
                 rtol=rtol,
                 atol=atol)
 
-    @staticmethod
+    @classmethod
     @retry_excute(3)
-    def _download_model(model_test, model_dir, models_dir):  # type: (TestCase, Text, Text) -> None
+    def _download_model(cls, model_test, model_dir, models_dir):  # type: (TestCase, Text, Text) -> None
```
houseroad: same here, if we turn it to classmethod, that means it can be overridden. Let's turn it public

postrational:
This is done, unless you have any other comments.

PR #1810, opened 2019-02-13
Enable ONNX_ML by default

shinh:
Apparently, official wheel packages enable it. However, if a
user builds ONNX by herself/himself, ONNX_ML will be disabled
by default. For example, as of writing, ONNX does not have a
wheel package for Python 3.7 and users need to build ONNX.

WDYT?


prasanthpul:
yeah CI needs to be fixed.

would love to figure out how to merge

prasanthpul:
@houseroad CI is passing now.

let's open a new issue to discuss how/if to merge the profiles.

prasanthpul:
#1813 opened

onnx/defs/gen_doc.py
```diff
@@ -20,7 +20,7 @@
 
 SNIPPETS = collect_snippets()
 SAMPLE_IMPLEMENTATIONS = collect_sample_implementations()
-ONNX_ML = bool(os.getenv('ONNX_ML') == '1')
+ONNX_ML = not bool(os.getenv('ONNX_ML') == '0')
```
houseroad: The file permission has been changed.

shinh: Done. I guess you'd like to keep 644 because users need to specify python/python3 explicitly? If this is the case, how about removing shebangs from 644 files?

Issue #1775, opened 2019-01-29
doc issue for Greater op


nbcsm:
in Opset9, it is not limited to ```float tensors``` anymore.

https://github.com/onnx/onnx/blob/15c33c945851907411619f599900c3852108e7e3/onnx/defs/logical/defs.cc#L82-L94

linkerzhang:
yes, all numeric types were added. is there any concern please?

nbcsm:
the description for T is still **float tensors**

prasanthpul:
thanks for the catch @nbcsm. Are you able to submit a PR to fix?

PR #1811, opened 2019-02-13
Fix TypeConstraint description for Greater and Less in Opset9

nbcsm:
fix https://github.com/onnx/onnx/issues/1775


houseroad:
Please run https://github.com/onnx/onnx/blob/master/tools/update_doc.sh to update the doc.

nbcsm:
fixed, please take another look.

PR #1818, opened 2019-02-15
Delete OpsetVersionConverter.md which is a duplicate of VersionConverter.md

prasanthpul:


docs/VersionConverter.md
```diff
@@ -1,12 +1,7 @@
...
 
-ONNX provides a C++ library for converting ONNX models between different
-opset versions. The library leverages the convenient in-memory
-representation that is much more convenient to manipulate than the raw
-protobuf structs, and converters to and from the protobuf format which
-were developed for the ONNX Optimizer.
-
-The primary motivation is to improve backwards compatibility of ONNX
+ONNX provides a library for converting ONNX models between different
+opset versions. The primary motivation is to improve backwards compatibility of ONNX
```
houseroad: Do we have a plan to replace the whole internal IR structure here and rewrite the whole version converter?
If so, why not change this until then?

I am not sure why we cannot tell the truth here :-)

cc: @jspisak @bddppq

prasanthpul: no changes, just streamlining the documentation. it's weird to talk about how it's implemented before talking about what it does or why. while it's relevant to developers of the tool, it's not for users of it. I've added the sentence to the end. what do you think

jspisak: Looks like you @prasanthpul added back the docs @houseroad mentioned. Should be good..

PR #1821, opened 2019-02-19
Support defined ONNX_ML in parent cmake files

houseroad:
Not only defined by env variable, we should also support it in cmake files.


houseroad:
So if we set ONNX_ML in the parent cmake already, we don't need to check the environment variable any more.

PR #1822, opened 2019-02-19
Infer shape of the second output of Dropout op

shinh:


onnx/defs/nn/defs.cc
```diff
@@ -1278,7 +1278,15 @@ ONNX_OPERATOR_SET_SCHEMA(
             "T",
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to float tensors.")
-        .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+            propagateShapeAndTypeFromFirstInput(ctx);
+            if (ctx.getNumOutputs() == 2) {
+                propagateElemTypeFromInputToOutput(ctx, 0, 1);
```
houseroad: I think mask tensor should be bool type

gramalingam: Yes, "Output(1)" type should be changed in the spec (line 1276) as well.

shinh: Done. As for the spec change, I sent a separate PR: https://github.com/onnx/onnx/pull/1826

Issue #1823, opened 2019-02-19
test data for shrink_test is of incorrect shape


hariharans29:
Both onnx/backend/test/data/simple/test_shrink/test_data_set_0/input_0.pb and onnx/backend/test/data/simple/test_shrink/test_data_set_0/output_0.pb have shape [5] whereas the graph input and output in onnx/backend/test/data/simple/test_shrink/model.onnx has shape [1,5].

In test_shrink_soft and test_shrink_hard, the test input data, test output data, the graph input defined in the model, the graph output defined in the model all have consistent shape of [5].

Tagging @zrphercule as fyi

zrphercule:
Thanks for raising this, I will look into it once I get back to office tomorrow.

zrphercule:
Hi @hariharans29 , I have checked this operator. When dealing with model test and node test, we use totally different test cases and data, that is why the input/output of a node test could be very different from the input/output of a model test. So long as a node test/model test is consistent itself, then it is fine.
After checking the code, it seems we need to modify the test data of model test. Sounds decent?

hariharans29:
Hi @zrphercule 

1) Agreed. Sorry for the confusion -  I did not mean to say the test cases between model test and node test are required to be the same. So long as they are consistent within the individual test itself, it is fine. 

2) Agreed. You can modify the shape of the test data to make it consistent with graph input/output.

Thanks for the quick turnaround.

zrphercule:
Closed due to the merging of #1825

PR #1825, opened 2019-02-20
Update test data of shrink_op

zrphercule:
fix issue #1823


zrphercule:
@houseroad yes it is......

houseroad:
I think you probably need to update the doc as well.

zrphercule:
@houseroad It seems the docs remain unchanged, since we dont have model test related information in docs?

zrphercule:
Cause it is a model test...

PR #1829, opened 2019-02-22
Change TopK operator to allow dynamic 'k'

hariharans29:
This PR is modifying TopK operator from having 'k' as attribute to having it as an input


hariharans29:
CC: @linkerzhang

linkerzhang:
@houseroad  any comments please? thanks a lot!

hariharans29:
@houseroad @linkerzhang - Is this change good to go ? Thanks!

spandantiwari:
@hariharans29 - there was a failing backend test which is why the CI has failed. That test (`test_densenet`) is disabled for now. Could you try running the CI job again to see if this passes?

spandantiwari:
@houseroad  - just FYI - the failing test was `test_densenet`, which we, coincidentally, disabled today (https://github.com/pytorch/pytorch/pull/17696) for the `Upsample` PR (https://github.com/onnx/onnx/pull/1773).

hariharans29:
> Sorry to block:
> 
> Yes, we do use topk in pytorch-onnx-caffe2
> https://github.com/pytorch/pytorch/blob/master/torch/onnx/symbolic.py#L1305

@houseroad - Sorry I don't quite understand. Could you please elaborate on the changes you are requesting ?

houseroad:
@hariharans29 I don't mean you need to change anything, but we need to prepare for the change. Otherwise, it will break our internal pipeline. I am thinking of setting the pytorch onnx exporter to some stable opset by default. So such changes won't cause problem anymore.

hariharans29:
@houseroad - Got it. I will wait for your go-ahead. Thanks.

hariharans29:
@houseroad - Thanks a lot!

PR #1832, opened 2019-02-25
Remove exp op: Affine, ImageScaler,ParametricSoftplus, Crop.

linkerzhang:


linkerzhang:
@houseroad @zrphercule  any comments on this change? Thanks!

PR #1834, opened 2019-02-26
Merge function into opschema

raymondxyang:
This PR implemented the proposed merge of function representation and OpSchema. User can later define a function proto in complicated ops. Different onnx frameworks can run it as a function in order to support the op.

In the process of defining a function attached op, user will:
* Define meta info (name, domain, version..), input/output, attribute, type constraints (Required)
* Define type/shape inference function (Optional)

This PR contains:
* The implementation of function body in opschema
* Updated checker, type/shape inference functions
* The implementation of python APIs, doc generating scripts
* A replaced definition in opset 9 of MVN as an example
* Verification in CI for consistency between op and its function body
* Test cases of implementation


onnx/defs/schema.cc
```diff
@@ -622,6 +622,12 @@ void OpSchema::ParseAndSetTypes(
   }
 }
 
+OpSchema& OpSchema::FunctionBody(const FunctionProto& func_proto) {
```
gramalingam: I think we need to specify only a list of nodes here. FunctionProto has extra fields (name, inputs, outputs, attributes) which replicate what is separately specified in an OpSchema. This is an unnecessary duplication (what happens if different values are specified in the two places? There is no check for that.)

raymondxyang: I checked that in verification.. but its definitely right we need to remove this. Ill change it.

onnx/defs/__init__.py
```diff
@@ -36,12 +44,10 @@ def _Attribute_default_value(self):  # type: ignore
...
-    function_byte_map = C.get_all_functions(domain)  # type: ignore
-    for function_name, raw_functions in function_byte_map.items():
-        for function_bytes in raw_functions:
-            function_proto = FunctionProto()
-            function_proto.ParseFromString(function_bytes)
-            function_map[function_name].append(function_proto)
-    return function_map
+def get_function_ops():  # type: () -> List[OpSchema]
+    schemas = C.get_all_schemas()
+    function_ops = []
```
gramalingam: "return [schema for schema in schemas if schema.has_function_body]" would be more compact.

onnx/onnx-operators-ml.proto
```diff
@@ -36,22 +36,6 @@ enum OperatorStatus {
 message FunctionProto {
   // The name of the function, similar usage of op_type in OperatorProto.
   optional string name = 1;
-
```
linkerzhang: no, this should not be touched, Removing these fields means the function is not complete.

houseroad:
I will look into the CircleCI problem.

onnx/defs/schema.h
```diff
@@ -558,6 +560,14 @@ class OpSchema final {
     return tensor_inference_function_ ? true : false;
   }
 
+  bool has_function_body() const {
```
linkerzhang: maybe "IsFunction()"

onnx/defs/schema.h
```diff
@@ -558,6 +560,14 @@ class OpSchema final {
...
+    return !function_body_nodes_.empty();
+  }
+
+  OpSchema& FunctionBody(const std::vector<NodeProto>& func_nodes);
+
+  FunctionProto GetFunctionBody() const;
```
linkerzhang: 1. GetFunction
2. The function may be owned by the schema.

onnx/defs/schema.h
```diff
@@ -20,6 +21,7 @@
 #include "data_type_utils.h"
 #include "onnx/common/constants.h"
 #include "onnx/defs/shape_inference.h"
+#include "onnx/onnx-operators_pb.h"
```
linkerzhang: I'm seeing we're also having onnx-operators-ml.proto.  You may either remove this proto to have only one proto file, or handle both cases (ONNX_ML=0 and ONNX_ML=1).

onnx/defs/schema.h
```diff
@@ -558,6 +560,14 @@ class OpSchema final {
     return tensor_inference_function_ ? true : false;
   }
 
+  bool has_function_body() const {
+    return function_body_.node_size() > 0;
+  }
+
+  OpSchema& FunctionBody(std::vector<NodeProto>& func_nodes);
+
+  const FunctionProto* GetFunctionBody() const;
```
linkerzhang: a better name  is GetFunciton, as here's not only getting its body (bunch of nodes), but all function related stuff.

onnx/defs/schema.h
```diff
@@ -568,6 +578,9 @@ class OpSchema final {
   void ParseAndSetTypes(
       /*out*/ std::vector<OpSchema::FormalParameter>* formalParameters);
 
+  // Build function with inofrmation stored in opschema
+  void FunctionBuilder();
```
linkerzhang: per the comment, name it as "BuildFunction"?

linkerzhang: typo: inofrmation ?

onnx/defs/schema.h
raymondxyang: This file is actually a auto-switch between onnx-operators-ml.pb.h and onnx-operators.pb.h

gramalingam:
@houseroad : please let us know if you have any comments on this PR. Thanks!

onnx/checker.cc
```diff
@@ -342,25 +342,15 @@ void check_node(
...
-    auto func = func_registry->GetFunction(
-        node.op_type(), domain_version, node.domain());
-    if (nullptr == func) {
-      fail_check(
-          "No Op or Function registered for " + node.op_type() +
-          " with domain_version of " +
-          ONNX_NAMESPACE::to_string(domain_version));
-    }
-    VerifyFunctionNode(node, *func, ctx, lex_ctx);
+  } else if (schema->Deprecated()) {
```
houseroad: This is better handling, thanks.

onnx/cpp2py_export.cc
```diff
@@ -45,9 +45,15 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
           &OpSchema::has_type_and_shape_inference_function)
       .def_property_readonly(
           "type_constraints", &OpSchema::typeConstraintParams)
-      .def_static("is_infinite", [](int v) {
-        return v == std::numeric_limits<int>::max();
-      });
+      .def_static(
+          "is_infinite",
+          [](int v) { return v == std::numeric_limits<int>::max(); })
+      .def_property_readonly("is_function", &OpSchema::is_function)
```
houseroad: can we call it `has_function` or `contains_function`?

onnx/defs/experiments/experiments_functions.cc
```diff
@@ -3,177 +3,189 @@
...
-  func.add_input("X");
-  func.add_output("X_MVN");
-  func.add_attribute("axes");
-  func.set_status(OperatorStatus::STABLE);
-
-  NodeProto* initial_node0 = func.add_node();
+using SupportType = OpSchema::SupportType;
+using SupportType = ONNX_NAMESPACE::OpSchema::SupportType;
+
+static std::vector<NodeProto> BuildMVN() {
```
houseroad: Nit: BuildMVNFunction?

onnx/defs/schema.h
```diff
@@ -558,6 +560,14 @@ class OpSchema final {
     return tensor_inference_function_ ? true : false;
   }
 
+  bool is_function() const {
```
houseroad: same here... `contains_function` or `has_function` feels more appropriate to me.

onnx/test/cpp/function_verify_test.cc
```diff
@@ -0,0 +1,152 @@
...
+// Verify registered ops with function body has compatible
+// definition on TypeConstraints between ops and function body
+TEST(FunctionVerification, VerifyFunctionOps) {
+  const std::vector<OpSchema> schemas = OpSchemaRegistry::get_all_schemas();
+  
+  for (const auto s : schemas) {
+    if (!s.is_function()) continue;
+    try{
+      auto function_body = s.GetFunction();
+      VerifyFunction(s, function_body);
```
houseroad: Can we add a counter here, to assert, we at least verify one function?

docs/Operators.md
```diff
@@ -139,6 +139,9 @@
   * <sub>experimental</sub> <a href="#ScaledTanh">ScaledTanh</a>
   * <sub>experimental</sub> <a href="#ThresholdedRelu">ThresholdedRelu</a>
 
+  **Operators with function registered:**
```
linkerzhang: Change this to ""Functions""

onnx/defs/experiments/experiments_functions.cc
```diff
@@ -3,177 +3,189 @@
 
 #include "onnx/common/constants.h"
```
linkerzhang: this file should not be in experiments folder any more, and its file name should not have "experiments_" prefix neither.

onnx/defs/schema.h
```diff
@@ -558,6 +560,14 @@ class OpSchema final {
     return tensor_inference_function_ ? true : false;
   }
 
+  bool has_function() const {
```
linkerzhang: HasFunction? I'm seeing the other two functions you added with name "FunctionBody" and "GetFunction". Let's keep the name consistent.

docs/Operators.md
houseroad: Only call it Functions may be misleading (trust me, users may get confused by functions again), I think Operators with function registered is Okay. Because they are operators, right?

raymondxyang:
> as talked offline, the way to ensure/verify the function body's correctness (compliance) is, when folks adding test case for a function, there should be two models generated, one model is with function node, one model is with function inlined. With same inputs/outputs, the two models should work as same in terms of correctness.

is this verification supposed to be the next PR?

onnx/defs/experiments/functions.cc
```diff
@@ -3,177 +3,189 @@
 
 #include "onnx/common/constants.h"
```
linkerzhang: don't put it under "experiments" folder, I suggest.

linkerzhang:
Thank you!

PR #1835, opened 2019-02-26
Add ability to override atol and rtol in BackendTest.Runner

skottmckay:
For operators that involve a lot of matmul such as RNN/LSTM/GRU the atol default of 1e-7 can be too strict. Provide a way in BackendTest.Runner to override the default rtol and atol values that are set in TestCase, so that appropriate values for the backend implementation being tested can be set.


skottmckay:
Can use changes from https://github.com/onnx/onnx/commit/873ddbbc33c6e54d90c5628387edd391fb651dfc instead

PR #1836, opened 2019-02-26
Update Slice with dynamic input & optional input steps

BowenBao:
* Replacing Slice with DynamicSlice: now `starts`, `ends` and `axes` are inputs, instead of attributes.
* Remove DynamicSlice from Experiment Ops.
* Add optional input `steps`.


onnx/defs/tensor/defs.cc
```diff
@@ -471,55 +474,96 @@ Example 2:
...
-            "Starting indices of corresponding axis in `axes`",
-            AttributeProto::INTS)
-        .Attr(
-            "ends",
-            "Ending indices (exclusive) of corresponding axis in axes`",
-            AttributeProto::INTS)
+        .Input(1, "starts", "1-D tensor of starting indices of corresponding axis in `axes`", "Tind")
+        .Input(2, "ends", "1-D tensor of ending indices (exclusive) of corresponding axis in `axes`", "Tind")
+        .Input(3, "axes", "1-D tensor of axes that `starts` and `ends` apply to.", "Tind", OpSchema::Optional)
+        .Input(4, "steps", "1-D tensor of slice step of corresponding axis in `axes`. Default to 1. ", "Tind", OpSchema::Optional)
```
linkerzhang: so the "steps" is used to specify the axis order/step of slicing? if that's the case, can we just use/follow the "axes" indices order please?

BowenBao: I'm not sure I understand the question correctly. Is your question suggesting that for a given axis `j = axes[i]`, `output[i]` should be computed based on `starts[j]`, `ends[j]`, `steps[j]`? If that's the case then yes that is the current design, I'll update the description to be more precise.

BowenBao:
@linkerzhang part of the CI is failing on various caffe2 & pytorch test cases & models using the old slice spec(<opset9). What is the recommended way of working around that?

houseroad:
This change looks good to me. But we should prepare for it.

spandantiwari:
@houseroad - all the tests are passing. Shall we go ahead and merge it now?

houseroad:
Yeah, it should be fine to merge.

ayermolo:
@BowenBao So to clarify. This is for opset 10. For opset < 10 starts/ends is still part of the attributes?

BowenBao:
> @BowenBao So to clarify. This is for opset 10. For opset < 10 starts/ends is still part of the attributes?

yes.

Issue #1772, opened 2019-01-25
convtranspose with dynamic output shape


zhijxu-MS:
I have a tensorflow model which uses "**tf.nn.conv2d_transpose**" and the op's **output shape** is specified at runtime.

While when i tried to convert it to onnx format i found that onnx requires its user to specified the output shape as an attribute of "**ConvTranspose**". And also i tried "auto_pad" but its result is still different with tensorflow's.

Is there any plan to support "ConvTranspose" with dynamic output shape? I think this is should be a common requirement at CNN domain, for example people want their model to accept dynamic input  image size.

linkerzhang:
Thanks a lot for creating the issue! I'm thinking of moving "pads" attribute as input to solve the problem. Meanwhile, ONNX spec marked "auto_pad" as "deprecated" but never deprecates it. Without having "pads" as input, the deprecation can't be done (otherwise, no way to support any kind of dynamic output shape case). Thus, moving "pads" as input will also make the deprecation of "auto_pad" happen. What do you think? @houseroad  @ebarsoum @bddppq

linkerzhang:
Dynamic "Pad" operator can't solve the problem actually. I'd suggest to update the spec of ConvTranspose, to have "output_shape" as an input, and remove the "pads" attribute.

zhijxu-MS:
> Dynamic "Pad" operator can't solve the problem actually. I'd suggest to update the spec of ConvTranspose, to have "output_shape" as an input, and remove the "pads" attribute.

agree to make "output_shape" as a dynamic input instead of an attribute since this is very common requirement at CNN domain to have a flexible input size.

thanks very much for the support!

jcwchen:
I don't think it has been resolved yet. Reopening this issue to track.

justinchuby:
Marked for discussion for 1.16

PR #1839, opened 2019-02-28
support dynamic output shape for ConvTranspose.

linkerzhang:
This PR changes the spec of ConvTranspose,
1. Support dynamic output shape cases.
    a. Move "output_shape" as an input, from an attribute.
    b. Remove "pads" attribute.
2. Fix the spec inconsistency on how to pad in odd padding number case.


linkerzhang:
This is solving the issue #1772

PR #1847, opened 2019-03-07
opset converter backward compatibility support for opset versions 9 and 8

pk-g:


houseroad:
@Ac2zoom feel free to review if you get time. :-)

prasanthpul:
@Ac2zoom / @houseroad - any feedback on this?

ebarsoum (approved):

houseroad (approved):

onnx/version_converter/adapters/upsample_9_8.h
```diff
@@ -0,0 +1,84 @@
...
+
+struct Upsample_9_8 final : public Adapter {
+  explicit Upsample_9_8()
+    : Adapter("Upsample", OpSetID(9), OpSetID(8)) {
+    }
+
+  void adapt_upsample_9_8(std::shared_ptr<Graph> graph, Node* node) const {
+      const ArrayRef<Value*>& inputs = node->inputs();
+      const std::vector<Tensor>& initializers = graph->initializers();
+      std::string scale_input_name = node->inputs()[1]->uniqueName();
```
spandantiwari: Since the code here and below strongly assumes that input[1], it may  be useful to have a quick check that input[1] is there, so that just in case there's an invalid graph without it, we throw a clear and meaningful error message.

onnx/version_converter/adapters/upsample_9_8.h
```diff
@@ -0,0 +1,84 @@
...
+          {
+            std::vector<float> value = initializers[i].floats();
+            if (initializers[i].is_raw_data()){
+              const std::string& bytes = initializers[i].raw();
+              value.insert(
+                  value.end(),
+                  reinterpret_cast<const float*>(bytes.c_str()),
+                  reinterpret_cast<const float*>(bytes.c_str() + bytes.size()));
+            }            
+            std::vector<double> d_values;
```
spandantiwari: Wold it be possible to use float here instead of double given that spec takes scales as floats?

spandantiwari (approved):

Ac2zoom:
> @Ac2zoom feel free to review if you get time. :-)

Other than @spandantiwari's comments, looks good to me!

onnx/version_converter/adapters/upsample_9_8.h
pk-g: It would not. fs is using double in IR.

onnx/version_converter/adapters/upsample_9_8.h
pk-g: Yes, the code strongly assumes there is an input[1] here because in upsample 9 input[1], is "scale", which is required and is not optional. Therefore, if there's no input[1] here we are dealing with an invalid model. From design perspective, model verification in terms of having required inputs and parameters is not part of converters role. so, if an invalid model is passed to converter in first place, there's not much that can be done and issues shall bubble up.

spandantiwari: Yes, I agree the model is invalid, but if you have an opportunity to throw a meaningful error message with minimal perf impact, it is preferable to have it.

PR #1848, opened 2019-03-08
bump the version for dropout

linkerzhang:
fix the issue that the version was not bumped when changing its type constraint declaration.


linkerzhang:
@shinh

PR #1851, opened 2019-03-08
Handle string comparision in tests

yuslepukhin:
Strings are represented as np.objects  and all other objects.


yuslepukhin:
I am reluctant to write code that tests the tester.

houseroad:
So I guess the existing test cases of StringNormalizer should already cover this. Is it true?

yuslepukhin:
> 
> 
> So I guess the existing test cases of StringNormalizer should already cover this. Is it true?

Yes, this is how the issue was discovered.

PR #1858, opened 2019-03-12
ARGN is not a variable

xsacha:
Latest CMake is complaining about this for me.
The current code returns the error:
Error: RELATIVE_PROTOBUF_GENERATE_CPP() called without any proto files

From the docs:
https://cmake.org/cmake/help/v3.14/command/macro.html
Since ARGN, ARGC, ARGV, ARGV0 etc. are not variables, you will NOT be able to use commands like
if (ARGV1)
...
In the first case, you can use if(${ARGV1}).


xsacha:
This patch actually isn't required it seems. Sorry about that.

PR #1859, opened 2019-03-12
Add a utils function update_inputs_outputs_dims

BowenBao:
The function updates the dimension of the inputs and outputs of the model, to the provided values in the parameter. These values can be either static or dynamic. This can be useful for extending models with fixed input size to support dynamic input sizes.

### DIFF
--- a/docs/PythonAPIOverview.md
+++ b/docs/PythonAPIOverview.md
@@ -214,3 +214,16 @@ import onnx.utils
 model = onnx.load('path/to/the/model.onnx')
 polished_model = onnx.utils.polish_model(model)
 ```
+
+### Updating Model's Inputs Outputs with Variable Length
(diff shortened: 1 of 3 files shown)


docs/PythonAPIOverview.md
```diff
@@ -214,3 +214,17 @@ import onnx.utils
 model = onnx.load('path/to/the/model.onnx')
 polished_model = onnx.utils.polish_model(model)
 ```
+
+### Updating Model's Inputs Outputs with Variable Length
+Function `update_inputs_outputs_dims` updates the dimension of the inputs and outputs of the model,
```
houseroad: I think we should explain more about how we handle dynamic/variable dimension in ONNX. You may link the corresponding doc as well.

BowenBao:
Closing, replacing with #2148

PR #1861, opened 2019-03-12
Add tokenizer to ONNX-ML

yuslepukhin:
Tokenizer is an op that is used in conjunction with StringNormalizer and TfIdfVectorizer ops.

  Tokenizer divides each string in X into a vector of strings along the last axis. Allowed input shapes are [C] and [N, C].
  If the maximum number of tokens found per input string is D, the output shape would be [N, C, D] when input shape is [N, C].
  Similarly, if input shape is [C] then the output should be [C, D]. Tokenizer has two different operation modes.
  The first mode is selected when "tokenexp" is not set and "separators" is set. If "tokenexp" is set and "separators" is not set,
  the second mode will be used. The first mode breaks each input string into tokens by removing separators.

  Let's assume "separators" is [" "] and consider an example.
  If input is

  ["Hello World", "I love computer science !"] whose shape is [2],

  then the output would be

 [["Hello", "World", padvalue, padvalue, padvalue],
 ["I", "love", "computer", "science", "!"]]

 whose shape is [2, 5] because you can find at most 5 tokens per input string.
 Note that the input at most can have two axes, so 3-D and higher dimension are not supported.

 For each input string, the second mode searches matches of "tokenexp" and each match will be a token in Y.
 The matching of "tokenexp" is conducted greedily (i.e., a match should be as long as possible).
 This operator searches for the first match starting from the beginning of the considered string,
 and then launches another search starting from the first remained character after the first matched token.
 If no match found, this operator will remove the first character from the remained string and do another search.
 This procedure will be repeated until reaching the end of the considered string.

  Let's consider another example to illustrate the effect of setting "mark" to true.
  If input is ["Hello", "World"],
  then the corresponding output would be [0x02, "Hello", "World", 0x03].
  This implies that if mark is true, [C]/[N, C] - input's output shape becomes [C, D+2]/[N, C, D+2].

If tokenizer removes the entire content of [C]-input, it will produce [[]].
I.e. the output shape should be [C][0] or [N][C][0] if input shape was [N][C].

If the tokenizer receives empty input of [0] then the output is [0] if empty input
of [N, 0] then [N, 0].


linkerzhang:
@wschin

onnx/defs/operator_sets-ml.h
```diff
@@ -28,6 +28,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(OnnxML, 1, Scaler);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(OnnxML, 1, TreeEnsembleClassifier);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(OnnxML, 1, TreeEnsembleRegressor);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(OnnxML, 1, ZipMap);
+class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(OnnxML, 1, Tokenizer);
```
houseroad: Are we sure we intentionally want to add this to ONNX-ML opset version 1, not version 2?

yuslepukhin: Will bump the version

docs/Changelog-ml.md
```diff
@@ -964,3 +964,59 @@ This version of the operator has been available since version 2 of the 'ai.onnx.
...
+      whose shape is [2], then the output would be [["Hello", "World", padvalue, padvalue, padvalue],
+      ["I", "love", "computer", "science", "!"]] whose shape is [2, 5] because you can find at most 5 tokens per input string.
+      Note that the input at most can have two axes, so 3-D and higher dimension are not supported. 
+      For each input string, the second mode searches matches of "tokenexp" and each match will be a token in Y.
+      The matching of "tokenexp" is conducted greedily (i.e., a match should be as long as possible).
+      This operator searches for the first match starting from the beginning of the considered string,
+      and then launches another search starting from the first remained character after the first matched token.
+      If no match found, this operator will remove the first character from the remained string and do another search.
+      This procedure will be repeated until reaching the end of the considered string.
+      If the tokenizer receives empty input of [0] then the output is [0] if empty input
```
gramalingam: Changing "[0]" to "shape [0]" will make this slightly easier to parse/read.

onnx/defs/traditionalml/defs.cc
```diff
@@ -1098,5 +1120,107 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+            // Add unknown D
+            output_shape.add_dim();
+          } else if (dim_size == 2) {
+            // Input is [N, C], Output is [N][C][?]
+            // Copy N, C
+            *output_shape.add_dim() = input_shape.dim(0);
+            *output_shape.add_dim() = input_shape.dim(1);
+            // Add unknown D
+            output_shape.add_dim();
+          }
```
gramalingam: Copy (local) output_shape to output, or make it a reference. (But the reference would be a problem if neither branch is taken, since it will appear as though output-shape has rank 0.)

gramalingam: If rank > 2, may be it would be better (for debugging) to throw a shape-inference exception.

yuslepukhin:
Upon further discussion and due to high volatility of the spec, we'd like to postpone this operator introduction into ONNX. This PR will be closed until further notice.

PR #1862, opened 2019-03-13
Add testcase generator for functions

raymondxyang:
* Automatically generate test cases for function's functionbody
* Fix a bug in MVN test


raymondxyang:
Added testcase validated with backend implementation 
image

linkerzhang:
This means, for an op with function body checked in to onnx, for any test case added, there will be two models generated, one with the function, the other with function body inline-ed.  With same inputs, it's expected to have same outputs when running these two models respectively. In this way, we can ensure that the function body attached to this "composed" op is correct.

onnx/backend/test/case/node/__init__.py
```diff
@@ -15,8 +15,74 @@
...
+                           op_prefix  # type:  Text
+                           ):  # type:  (...) -> List[NodeProto]
+    node_list = []
+    input_names_map = dict()
+    output_names_map = dict()
+    attribute_map = node.attribute
+
+    for idx in range(len(node.input)):
+        if idx not in range(len(function_proto.input)):
+            continue
```
gramalingam: Isn't this an error-situation? May be it would be better to return an empty list, instead of continuing, if there is an error like this. Similarly, what if for a function_proto.input there is no matching node.input? In that case, I think we should treat it as an empty string (a missing optional input).

onnx/backend/test/case/node/__init__.py
```diff
@@ -15,8 +15,74 @@
...
+        input_names_map[function_proto.input[idx]] = node.input[idx]
+
+    for idx in range(len(node.output)):
+        if idx not in range(len(function_proto.output)):
```
gramalingam: Same comments as for input.

onnx/backend/test/case/node/__init__.py
raymondxyang: tbh the checker in gtests should catch this later. so however we handle the mismatched # of io tensors is fine here

onnx/backend/test/case/node/__init__.py
```diff
@@ -15,8 +15,74 @@
...
+    output_names_map = dict()
+    attribute_map = node.attribute
+
+    for idx in range(len(node.input)):
+        if idx not in range(len(function_proto.input)):
+            return []
+        input_names_map[function_proto.input[idx]] = node.input[idx]
+
+    for idx in range(len(node.output)):
+        if idx not in range(len(function_proto.output)):
```
gramalingam: Hi, the more important change is: make "input_names_map[function_proto.input[idx]]" the empty string if "idx >= len(node.input)". E.g., iterating over range(len(function_proto.input)) would be fine.

Issue #1302, opened 2018-08-17
Problem installing onnx from source


elboyran:
After installing onnx from binaries and encountering problems (missing functions) when running the Python API notebooks and after an advice to install from source, I have attempted several times to do so. Unfortunately, the installation check does not work. 
Windows host, VBox 5.2.18 virtual machine Ubuntu 18.04. 

Below is exactly what I did (following the installation instructions) in my last attempt:
```
elena@elena-VirtualBox:~$ cd eStep/XAI/Software/
elena@elena-VirtualBox:~/eStep/XAI/Software$ conda update -n base -c defaults conda
elena@elena-VirtualBox:~/eStep/XAI/Software$ conda create --name onnx_elena
elena@elena-VirtualBox:~/eStep/XAI/Software$ source activate onnx_elena
(onnx_elena) elena@elena-VirtualBox:~/eStep/XAI/Software$ conda config --add channels conda-forge
(onnx_elena) elena@elena-VirtualBox:~/eStep/XAI/Software$ conda install -c conda-forge protobuf numpy
(onnx_elena) elena@elena-VirtualBox:~/eStep/XAI/Software$ git clone https://github.com/onnx/onnx.git
(onnx_elena) elena@elena-VirtualBox:~/eStep/XAI/Software$ cd onnx
(onnx_elena) elena@elena-VirtualBox:~/eStep/XAI/Software/onnx$ git submodule update --init --recursive
(onnx_elena) elena@elena-VirtualBox:~/eStep/XAI/Software/onnx$ python setup.py install
(onnx_elena) elena@elena-VirtualBox:~/eStep/XAI/Software/onnx$ cd onnx
(onnx_elena) elena@elena-VirtualBox:~/eStep/XAI/Software/onnx/onnx$ python -c "import onnx"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/elena/anaconda3/envs/onnx_elena/lib/python3.6/site-packages/onnx-1.2.2-py3.6-linux-x86_64.egg/onnx/__init__.py", line 11, in <module>
    import onnx.helper  # noqa
  File "/home/elena/anaconda3/envs/onnx_elena/lib/python3.6/site-packages/onnx-1.2.2-py3.6-linux-x86_64.egg/onnx/helper.py", line 13, in <module>
    import onnx.defs as defs
[...]

PR #1863, opened 2019-03-13
update readme install command to fix issue 1302

maxwillzq:
see https://github.com/onnx/onnx/issues/1302


maxwillzq:
@houseroad ,

Thanks a lot for your comments.  Could you help take a look at issue #1302 while install does not work to when load defs in python?

maxwillzq:
After clean build, it have no problem. Thanks a lot

houseroad:
@maxwillzq since the problem is solved, can we close the PR?

maxwillzq:
Yes, we can

PR #1865, opened 2019-03-14
promote scaledtanh

askhade:
Promoting ScaledTanh operator


linkerzhang:
it's discussed and agreed to remove this op.

PR #1866, opened 2019-03-14
Remove scaledtanh

askhade:
Remove scaledtanh


zchrissirhcz:
@linkerzhang @askhade Why this op is removed from ONNX, is there any details?

PR #1868, opened 2019-03-15
Introduce Function Body Helper 

SherlockNoMad:
In this PR, we introduce an easier way to compose a function body. 

It should serve as a shortcut to reduce the effort to define an op with function body.


onnx/defs/experiments/functions.cc
```diff
@@ -175,17 +25,35 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-			OPTIONAL)
+            AttributeProto::INTS,
+            OPTIONAL)
         .TypeConstraint(
             "T",
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to all numeric tensors.")
-       .FunctionBody(BuildMVNFunctionBody()));
+        .FunctionBody(FunctionBodyHelper::BuildNodes(
+            {// nodes: {outputs, op, inputs, attributes}
```
gramalingam: An example to illustrate might be helpful: e.g., "{ {Z}, "Add", {"X", "Y"} }" represents "Z = Add(X,Y)".

gramalingam:
Please look at the CI error. It seems to be complaining about "onnx/defs/experiments/functions.cc" … not sure what that is.

SherlockNoMad:
Hi @houseroad, could you please have a look on this PR when you got a chance ? 

This PR is intended to simplify the process for defining a Function Body.

onnx/defs/function.h
```diff
@@ -19,4 +21,80 @@ void FunctionExpandHelper(
...
     const std::string& node_prefix = "");
+
+class FunctionBodyHelper {
+ public:
+  struct AttributeProtoWrapper {
+    AttributeProto proto;
+
+    AttributeProtoWrapper() {}
+
+    AttributeProtoWrapper(const AttributeProto& attr_prot) {
```
linkerzhang: looks to me this wrapper is not that useful, but creating one more copy. I may miss sth though.

onnx/defs/experiments/functions.cc
```diff
@@ -3,163 +3,13 @@
 
 #include "onnx/common/constants.h"
```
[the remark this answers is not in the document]
SherlockNoMad: Moved to nn/defs.cc

onnx/defs/function.h
SherlockNoMad: This class serve as a syntactic sugar for easily define an attribute, so that user simply write following without worrying about how to compose a AttributeProto. 

{"attribute", 1.0f}
{"attribute", int64_t(1)}
{"attribute", "string_value"}
{MakeRefAttribute("axes", AttributeProto::INTS)}}

PR #1869, opened 2019-03-16
Adding RoIAlign operator

sampepose:
This picks up where #1010 left off for RoiAlign.


sampepose:
cc @rbgirshick for api review

onnx/defs/object_detection/defs.cc
```diff
@@ -0,0 +1,114 @@
...
+          "rois",
+          "RoIs (Regions of Interest) to pool over; rois is "
+          "2-D input of shape (num_rois, 4) given as "
+          "[[x1, y1, x2, y2], ...]. "
+          "The RoIs' coordinates are in the coordinate system of the input image.",
+          "T1")
+      .Input(
+          2,
+          "batch_splits",
+          "Tensor of shape (num_images) with each element denoting the number "
```
BowenBao: This doesn't seem to support all cases in the current api design? Currently (both pytorch and onnx) rois are 2-D input of shape (num_rois, 5) given as [[batch_id, x1, y1, x2, y2], ...], which might not be sorted by batch_id. How about having the input as
```
.Input(
  2,
  "batch_indices",
  "Tensor of shape (num_rois) with each element denoting "
  "the index of the corresponding image in the batch.",
  "T2")
```

onnx/defs/object_detection/defs.cc
```diff
@@ -0,0 +1,114 @@
...
+      .TypeConstraint(
+          "T2",
+          {"tensor(int16)", "tensor(int32)", "tensor(int64)", "tensor(uint8)",
+           "tensor(uint16)", "tensor(uint32)", "tensor(uint64)"},
+          "Constrain types to int tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          if (!hasNInputShapes(ctx, 3)) {
+            return;
+          }
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
```
BowenBao: Similar to https://github.com/onnx/onnx/blob/c05f2ae412daf8fd64136ca354b97ccf73e0ea6c/onnx/defs/nn/defs.cc#L683, consider more checks (e.g.: input_shape.dim_size() == 4, rois_shape.dim_size() == 2, etc).

onnx/defs/object_detection/defs.cc
```diff
@@ -0,0 +1,114 @@
...
+          "2-D input of shape (num_rois, 4) given as "
+          "[[x1, y1, x2, y2], ...]. "
+          "The RoIs' coordinates are in the coordinate system of the input image.",
+          "T1")
+      .Input(
+          2,
+          "batch_splits",
+          "Tensor of shape (num_images) with each element denoting the number "
+          "of RoIs in the corresponding image in the batch.",
+          "T2")
```
houseroad: Do we need to support so many int types for this input?

prasanthpul:
@sampepose can you please address the feedback and resolve the conflicts?

sampepose:
Yeah, I should be able to get to it this week.

prasanthpul:
@sampepose any update?

docs/Operators.md
```diff
@@ -9931,6 +9932,252 @@ for test_name, shape in test_cases.items():
...
+</dl>
+
+#### Inputs
+
+<dl>
+<dt><tt>X</tt> : T1</dt>
+<dd>Input data tensor from the previous operator; 4-D feature map of shape (N x C x H x W), where N is the batch size, C is the number of channels, and H and W are the height and the width of the data.</dd>
+<dt><tt>rois</tt> : T1</dt>
+<dd>RoIs (Regions of Interest) to pool over; rois is 2-D input of shape (num_rois, 4) given as [[x1, y1, x2, y2], ...]. The RoIs' coordinates are in the coordinate system of the input image.</dd>
+<dt><tt>batch_splits</tt> : T2</dt>
```
wenbingl: Can we seperate batch_split from this op?

jspisak:
@sampepose - let us know if you need help landing this one. Right now it is gating support for detection models for ONNX v1.5 so I'd like to see what the open issues are and try to address them.

pranavsharma:
Does it make sense to add a 'mode' attribute that determines the pooling method? "avg" or "max"?

onnx/defs/object_detection/defs.cc
pranavsharma: Or we can just follow the pytorch spec ([[batch_id, x1, y1, x2, y2], ...]) for this? Will probably make conversions easier.

pranavsharma:
I think we should make 2 concrete changes here:
* rois should be specified along with the batch_id (just like pytorch). 
```      .Input(1, "rois", "RoIs (Regions of Interest2) to pool over; rois is 2-D input of shape (num_rois, 5) given as [[batch_id, x1, y1, x2, y2], ...]. The RoIs' coordinates are in the coordinate system of the input image.", "T")```
* Introduce a pooling mode to specify avg/max with default as avg.
```      .Attr(
          "mode",
          "The pooling method. Two modes are supported: 'avg' and 'max'. "
          "Default is 'avg'.",
          AttributeProto::STRING,
          std::string("avg"))
```
cc @spandantiwari @BowenBao

daquexian:
I noticed @Maratyszcza said "storing batch index in a float tensor is a 100% bad idea." in https://github.com/onnx/onnx/pull/1010#issuecomment-459268400. Should we still follow the current pytorch spec?

onnx/defs/object_detection/defs.cc
```diff
@@ -0,0 +1,109 @@
...
+
+RoIAlign is proposed to avoid the misalignment by removing
+quantizations while converting from original image into feature
+map and from feature map into RoI feature; in each ROI bin,
+the value of the sampled locations are computed directly
+through bilinear interpolation.
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    RoIAlign,
```
prasanthpul: minor: capitalization of I. exiting MaxRoiPool op does not capitalize the I. should keep it consistent

spandantiwari:
> I noticed @Maratyszcza said "storing batch index in a float tensor is a 100% bad idea." in [#1010 (comment)](https://github.com/onnx/onnx/pull/1010#issuecomment-459268400). Should we still follow the current pytorch spec?

I agree with @Maratyszcza that batch_index should ideally not be stored in float tensor. How about creating a separate input, with correct type, for batch_index as @BowenBao has suggested above.

onnx/defs/object_detection/defs.cc
```diff
@@ -0,0 +1,109 @@
...
+            return;
+          }
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+
+          auto& input_shape = getInputShape(ctx, 0);
+          auto& rois_shape = getInputShape(ctx, 1);
+          auto* output_shape = getOutputShape(ctx, 0);
+
+          output_shape->clear_dim();
+          output_shape->add_dim()->set_dim_value(static_cast<int64_t>(
```
houseroad: Minimum check/assert is recommended before accessing the shape. It will help us detect the problem early and print out reasonable error message.

pranavsharma:
> I noticed @Maratyszcza said "storing batch index in a float tensor is a 100% bad idea." in [#1010 (comment)](https://github.com/onnx/onnx/pull/1010#issuecomment-459268400). Should we still follow the current pytorch spec?

Yup, this makes sense. Creating a separate input for batch_index is fine.

sampepose:
need to update the changelog, one sec

onnx/defs/object_detection/defs.cc
```diff
@@ -0,0 +1,131 @@
...
+          "(num_rois, C, output_height, output_width). The r-th batch element Y[r-1] "
+          "is a pooled feature map corresponding to the r-th RoI X[r-1].",
+          "T1")
+      .TypeConstraint(
+          "T1",
+          {"tensor(float16)", "tensor(float)", "tensor(double)"},
+          "Constrain types to float tensors.")
+      .TypeConstraint(
+          "T2",
+          {"tensor(int16)", "tensor(int32)", "tensor(int64)", "tensor(uint8)",
```
pranavsharma: One minor comment: do the batch indices really need to be so many different types? Shouldn't just int64 suffice to be super conservative?

onnx/defs/object_detection/defs.cc
```diff
@@ -0,0 +1,131 @@
...
+          "X",
+          "Input data tensor from the previous operator; "
+          "4-D feature map of shape (N, C, H, W), "
+          "where N is the batch size, C is the number of channels, "
+          "and H and W are the height and the width of the data.",
+          "T1")
+      .Input(
+          1,
+          "rois",
+          "RoIs (Regions of Interest) to pool over; rois is "
```
pranavsharma: nit: would be good to call out that each of the co-ordinate sets has a 1-1 correspondence with the batch_index tensor. It might be obvious to some and not to others.

BowenBao:
Comments on the update
1. The description of input batch_index needs to be updated. It is still the old one for batch_split. 
2. Similarly, the input data for input batch_index in roi_align.py needs to be updated.

onnx/defs/object_detection/defs.cc
pranavsharma: @sampepose Can you update the documentation for this input? Also, since we'll be storing multiple indices, let's call it batch_indices like @BowenBao mentioned. I think this should be the final change.

sampepose:
@BowenBao, @pranavsharma , @houseroad: Let me know if you want any additional changes. All comments should be addressed now.

BowenBao:
> @BowenBao, @pranavsharma , @houseroad: Let me know if you want any additional changes. All comments should be addressed now.

@sampepose thank you for the updates!

Two minor changes:
1. Update batch_indices description phrasing
```
          "1-D tensor of shape (num_rois,) with each element denoting "
          "the index of the corresponding image in the batch.",
```
2. In roialign.py
change
```batch_indices = np.array([0, 0, 0], dtype=np.int32)```
to
```batch_indices = np.array([0, 0, 0], dtype=np.int64)```

pranavsharma:
LGTM 👍 (except @BowenBao's roialign.py comment).

spandantiwari:
Thanks @sampepose . Looks good.

PR #1872, opened 2019-03-18
quantization support in onnx

linkerzhang:
ONNX quanitzation support:
Requirements:
1.	Interoperability MUST be ensured.
ONLY widely accepted quantization schema can be standardized in ONNX. In this design, 8 bits linear (scale/zero_point) quantization will be standardized.
2.	Customized quantization schema should be allowed.
ONNX should be able to represent customized quantization schemas (the schema hasn’t been standardized in ONNX yet) with a subgraph consisting of primitive operators.
3.	All ONNX operators must define a mathematical function of the following form: 
outputs = OP(inputs, attrs)
It means the data needed for mathematical calculation defined by an op must be either an input or an attribute.
4.	Enable both static and dynamic quantization.
Quantization parameters used in defining an op will be defined as inputs/outputs. Static quantization will be a special case of dynamic one, where the quantization parameter inputs are from either initializers or constant nodes.
NOTE: as a best practice, weights in an inference model should be statically quantized.
5.	Support model verification for static quantization models. The verification includes,
a.	Same tensor should have same real-value representation.
If they use same static quantization parameters, then this can be ensured.
b.	Any other kind of quantization parameters’ value check before sending a model to a hardware vendor.

Goals of this design/PR:
1.	Add a small set of operators to standardize 8 bits linear (scale/zero_point) quantization.
        QLinearMatMul/QLinearConv are added in this PR, more ops may be added in separate PRs as needed later.
2.	Add a small set of operators to further enable ONNX to represent other quantization schemas. 
        MatmulInteger/ConvInteger are added in this PR, more ops may be added in separate PRs as needed later.
3.	Add quantization information as model level annotation for easy model verification.


onnx/defs/quantization/defs.cc
```diff
@@ -0,0 +1,108 @@
...
+        .Input(0, "x", "N-D full precision Input tensor to be quantized.", "T1")
+        .Input(
+            1,
+            "y_scale",
+            "Scale for doing quantization to get 'y'. It's a scalar, which means a per-tensor/layer quantization.",
+            "tensor(float)")
+        .Input(
+            2,
+            "y_zero_point",
+            "Zero point for doing quantization to get 'y'. It's a scalar, which means a per-tensor/layer quantization.",
```
raghuramank100: Better to have y_zero_point as optional in the quantizer if it is optional in the dequantizer so that we are consistent.

onnx/defs/math/defs.cc
```diff
@@ -1226,20 +1236,144 @@ Computes the error function of the given input tensor element-wise.
...
+    10,
+    OpSchema()
+        .SetDoc(MatMulInteger_ver10_doc)
+        .Input(0, "A", "N-dimensional matrix A", "T1")
+        .Input(1, "B", "N-dimensional matrix B", "T2")
+        .Input(
+            2,
+            "a_zero_point",
+            "Zero point tensor for input 'A'. It's optional and default value is 0. It could be a scalar or a 1-D tensor, "
+            "which means a per-tensor or per-row quantization. If it's a 1-D tensor, its number of elements "
```
raghuramank100: We should explicitly pass the axis as an argument here to be consistent with the definition of per-row/per-column/per-channel quantization.

onnx/defs/nn/defs.cc
```diff
@@ -864,6 +885,298 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "Otherwise the size is (N x C x D1 x D2 ... x Dn). "
+            "Optionally, if dimension denotation is "
+            "in effect, the operation expects input data tensor "
+            "to arrive with the dimension denotation of [DATA_BATCH, "
+            "DATA_CHANNEL, DATA_FEATURE, DATA_FEATURE ...].",
+            "T1")
+        .Input(
+            1,
+            "x_scale",
+            "Scale tensor for input 'x'. It's a scalar, which means a per-tensor/layer quantization.",
```
raghuramank100: Lets explicitly provide the axis as a parameter here. For example if I have 3x3x3x32 (3 input channels, 3x3 kernel and 32 output channels), I would specify axis=3 and have a vector of length 32 to specify the per-channel scale and zero-point

onnx/defs/math/defs.cc
linkerzhang: I don't think that we need this flexibility, since for per-channel, it has to be per output channel (for conv) for weights. For matmul, if it's not per tensor, per row for the first and per-column for the 2nd is a good one, but per-column for the 1st input and per-row for 2nd input is not that useful (math is not that straight-forward). So I'd suggest to keep this specification.

onnx/defs/quantization/defs.cc
linkerzhang: Sounds good.

onnx/defs/nn/defs.cc
linkerzhang: for conv, the most frequent used one is per-layer (per-tensor), or per output channel (for weights). I don't see much usage about other cases. I'd suggest to keep the spec and extend it when we need the other cases (flexibility). Sounds good?

onnx/defs/quantization/defs.cc
linkerzhang: sounds good.

onnx/defs/math/defs.cc
linkerzhang: I think for matmul, per-row quant matmul per-column quant makes sense in terms of math. the other way (per-column quant matmul per-row quant) does not make sense, right?

For Conv, normally, folks just use per-tensor for activation and per output channel for weights some times. Don't really see other use cases. 

To not add complexity/burden to the spec and partner's implementation, I'd keep the spec as it is, and we may extend it when seeing more other cases.

Sounds good?

linkerzhang:
am going to have all test cases covered in separate PR soon.

linkerzhang:
@raghuramank100 I'm going to check it in today. Please let me know if you have more comments.

@spandantiwari please help to drive the pytorch test failure. I'm not taking the failure as a check-in blocker.

Thank you!

onnx/defs/math/defs.cc
raghuramank100: I see what your saying but it might be good to be consistent across the spec. For Quantize_Linear, we specify axis as attribute:
   .Attr(
            "axis",
            "The axis along which same quantization parameters are applied. It's optional. If it's not specified, it means per-tensor quantization and input 'x_scale' and 'x_zero_point' must be scalars. If it's specified, it means per 'axis' quantization and input 'x_scale' and 'x_zero_point' must be 1-D tensors.",
            AttributeProto::INT,
            false)

Might be good to make this consistent across all ops. We could still enforce that the op itself only supports per-row for the first tensor and per column for the second one.

onnx/defs/quantization/defs.cc
```diff
@@ -0,0 +1,110 @@
...
+The dequantization formula is y = (x - x_zero_point) * x_scale. 'x_scale' and 'x_zero_point' must have same shape.
+'x_zero_point' and 'x' must have same type. 'x' and 'y' must have same shape. In the case of dequantizing int32,
+there's no zero point (zero point is supposed to be 0).
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    DequantizeLinear,
+    10,
+    OpSchema()
+        .Input(0, "x", "N-D quantized input tensor to be de-quantized.", "T")
```
raghuramank100: Should we also have an axis argument, like we have with the quantizer?

onnx/defs/math/defs.cc
linkerzhang: My bad. "axis" for quant should be removed.

onnx/defs/quantization/defs.cc
linkerzhang: clarified above, "axis" in quant should be removed. given only "weights" has per output channel quant (which is static).

onnx/defs/math/defs.cc
```diff
@@ -1226,20 +1236,144 @@ Computes the error function of the given input tensor element-wise.
...
+            "b_zero_point",
+            "Scale tensor for input 'B'. It's optional and default value is 0.  It could be a scalar or a 1-D tensor, "
+            "which means a per-tensor or per-column quantization. If it's a 1-D tensor, its number "
+            "of elements should be equal to the number of columns of input 'B'.",
+            "T2",
+            OpSchema::Optional)
+        .Output(0, "Y", "Matrix multiply results from A * B", "T3")
+        .TypeConstraint(
+            "T1",
+            {"tensor(int8)", "tensor(uint8)"},
```
KeDengMS: Please consider support int16 as well. There's cblas_gemm_s16s16s32 for int16 inputs to generate int32 outputs.

linkerzhang: @darrenscrews is going to make that (more types support) in separate PR.

houseroad:
CI side is okay. Waiting @raghuramank100 for the final approval :-)

linkerzhang:
as clarified in mail thread, the "adding axis" comment from @raghuramank100  may be added when we see use cases and want to support per axis quantization for activations. I'm merging this PR now. @raghuramank100  please feel free to share more comments if any. we can keep tuning it.

onnx/defs/quantization/defs.cc
```diff
@@ -0,0 +1,104 @@
...
+// Licensed under the MIT license.
+
+#include "onnx/defs/schema.h"
+
+namespace ONNX_NAMESPACE {
+
+static const char* QuantizeLinear_ver10_doc = R"DOC(
+The linear per-tensor/layer quantization operator. It consumes a high precision tensor, a scale, a zero point to compute the low precision / quantized tensor.
+The quantization formula is y = saturate ((x / y_scale) + y_zero_point). For saturation, it saturates to [0, 255] if it's uint8, or [-128, 127] if it's int8.
+For (x / y_scale), it's rounding to nearest ties to even. Refer to https://en.wikipedia.org/wiki/Rounding for details. 'y_zero_point' and 'y' must have same type.
```
fdwr: Guessing this was input from Intel to change from round-evens-away-from-zero (which was our earlier choice, which was TensorFlow's default and C++ std::round's default)?

onnx/defs/quantization/defs.cc
fdwr: Well it's certainly simpler to implement DequantizeLinear with `zero-point` and `scale` being scalars. We originally included the `axis` attribute because it was a compromise between space savings and accuracy, and with channel-wise weight packing, @youngkim93 noticed significant accuracy improvements across the various image recognition models. Are you dropping it because QLinearConv doesn't have it? (granted, getting QLinearConv to support it would greatly complicate things) This should be compatible with our current code because the `axis` was optional anyway.

onnx/defs/nn/defs.cc
```diff
@@ -21,35 +21,35 @@ const char* auto_pad_doc =
...
-    auto output_type = ctx.getOutputType(1);
-    if (output_type->value_case() == TypeProto::kTensorType ||
-        output_type->value_case() == TypeProto::VALUE_NOT_SET) {
-      output_type->mutable_tensor_type()->set_elem_type(TensorProto::INT64);
-    }
-  }
+    bool require_kernel_shape,
+    int input1Idx,
+    int input2Idx) {
+  // propagateElemTypeFromInputToOutput(ctx, 0, 0);
```
houseroad: what is this?

onnx/defs/nn/defs.cc
```diff
@@ -872,6 +893,298 @@ ONNX_OPERATOR_SET_SCHEMA(
     1,
     OpSchema().FillUsing(ConvOpSchemaGenerator("a filter")));
 
+static const char* QLinearConv_ver10_doc = R"DOC(
+The convolution operator consumes a quantized input tensor, its scale and zero point,
+a quantized filter, its scale and zero point, and output�s scale and zero point,
```
houseroad: output's is non-utf8 character.

houseroad:
The merged version is prematured. Please address my comments and resubmit the PR. Thanks

PR #1877, opened 2019-03-21
Fix wrongly handled attribute in MVN and test generating scripts

raymondxyang:
* Fixed missing default value in updated MVN op
* Fixed bugs in function test generating script
* Revert wrong change in numpy validation data generation step
* Updated generated data/model
* Testing on backend:
image


onnx/defs/nn/defs.cc
```diff
@@ -1768,6 +1768,8 @@ static const char* mvn_ver9_doc = R"DOC(
       on the input tensor X using formula: <br/> ``` (X-EX)/sqrt(E(X-EX)^2) ```
 )DOC";
 
+static std::vector<int64_t> default_axes = {0, 2, 3};
```
houseroad: nit: default_axes sounds too general. how about mvn_default_axes?

docs/Operators.md
```diff
@@ -6911,7 +6911,7 @@ This version of the operator has been available since version 9 of the default O
 #### Attributes
 
 <dl>
-<dt><tt>axes</tt> : list of ints</dt>
+<dt><tt>axes</tt> : list of ints (default is ['0', '2', '3'])</dt>
```
gramalingam: The following documentation line "The default is to reduce over all the dimensions of the input tensor." sounds contradictory. Please delete that line as well.

PR #1878, opened 2019-03-21
fix the problems introduced in previous PRs in operator registration

houseroad:
Dropout, https://github.com/onnx/onnx/pull/1848, didn't add the dropout v10 to the operator_set.h
cc: @linkerzhang 

ImageScaler and Crop deleted but added back in https://github.com/onnx/onnx/pull/1836. cc: @linkerzhang @BowenBao


houseroad:
I remember we agreed on having a better registration mechanism for operators, which should help us avoid such bugs introduced in the previous PRs. cc: @bddppq @linkerzhang

houseroad:
The check fails on the check here: https://github.com/onnx/onnx/blob/master/onnx/defs/schema.cc#L820

We should have a CI instance which enables the debug mode.

PR #1883, opened 2019-03-22
Adding Reverse Operator

pk-g:
resubmitting #1804 changes for review after #1882 revert 
@houseroad , @linkerzhang : please feel free to continue review here, thanks.


onnx/backend/test/case/node/reverse.py
```diff
@@ -0,0 +1,60 @@
...
+import onnx
+from ..base import Base
+from . import expect
+
+
+class Reverse(Base):
+
+    @staticmethod
+    def export_default():  # type: () -> None
+        input = np.arange(6.0).reshape(2, 3)
```
houseroad: please specify the element data type explicitly here. and regenerate the test data to make sure we are using the right type of data.

onnx/backend/test/case/node/reverse.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+
+        output = np.flip(np.flip(input, 0), 1)
+
+        expect(node, inputs=[input], outputs=[output],
+               name='test_reverse_default')
+
+    @staticmethod
+    def export_with_axes():  # type: () -> None
+        input = np.arange(6.0).reshape(2, 3)
```
houseroad: ditto

onnx/backend/test/case/node/reverse.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+
+        output = np.flip(input, 0)
+
+        expect(node, inputs=[input], outputs=[output],
+               name='test_reverse_with_axes')
+
+    @staticmethod
+    def export_with_negative_axes():  # type: () -> None
+        input = np.arange(12.0).reshape(2, 2, 3)
```
houseroad: ditto

wschin:
`Slice` now can have negative steps. Is this op overlapping with `Slice`?

onnx/backend/test/case/node/reverse.py
pk-g: sure.

pk-g:
@wschin, thanks for your input. Yes, indeed; with opset 10 recent changes in slice, we can now cover the functionalities of this op as a slice scenario. There will be cons and pros for having/not having a stand-alone `reverse` operator vs. relying on more generic `slice`. However, after considering both side on cons and pros, I'm leaning more towards relying on `slice` instead due to following reasons: although `reverse` may sound more intuitive wording from clarity standpoint only for simple reverse scenarios, `slice` is more flexible and generic in handling different cases such as greater than one step size, and support for dynamic steps. In addition, from runtime standpoint, reducing amount of overlaps between ops shall help with overall efficiency. 

@houseroad , @linkerzhang , @wschin Therefore, I'd recommend withdrawing/abandoning this PR and using `slice` instead for scenarios addressed by this op . Having said that, please feel free to provide feedback.

PR #1884, opened 2019-03-22
Add IsInf to detect infinity values

wschin:
In addition to `NaN`, positive and negative values are also very important special values in floating-number system, so it'd be nice if `IsInf` can support them.


docs/Changelog.md
```diff
@@ -9531,6 +9531,49 @@ This version of the operator has been available since version 10 of the default
...
+      attributes.
+
+#### Version
+
+This version of the operator has been available since version 10 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>detectnegativeinfinity</tt> : int (default is 0)</dt>
```
ebarsoum: Use the same style for attribute naming as the rest of ONNX. The above should be 'detect_negative_infinity'.

docs/Changelog.md
```diff
@@ -9531,6 +9531,49 @@ This version of the operator has been available since version 10 of the default
...
+<dd>(Optional) Whether map negative infinity to true. Default to 1 so that negative infinity induces true. Set this attribute to 0 if positive infinity should be mapped to false.</dd>
```
gramalingam: "positive" => "negative"

onnx/defs/tensor/defs.cc
```diff
@@ -1563,13 +1563,40 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "(Optional) Whether map positive infinity to true. Default to 1 "
+            "so that positive infinity induces true. Set this attribute to 0 "
+            "if positive infinity should be mapped to false.",
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
+        .Attr(
+            "detectnegativeinfinity",
+            "(Optional) Whether map negative infinity to true. Default to 1 "
+            "so that negative infinity induces true. Set this attribute to 0 "
+            "if positive infinity should be mapped to false.",
```
gramalingam: "positive" => "negative"

gramalingam:
Adding a test-case to test this would be helpful.

spandantiwari:
@wschin - This is useful functionality and will be a welcome addition to ONNX.

But have we considered an alternate design where we have a separate op for detecting `Inf` and `-Inf`, possibly named `IsInf`? 

`Inf` and `NaN` are two different floating-point numbers with totally different meaning and IEEE representation. Putting Infs under the `IsNan` op is not the best design. I am aware that we do not want to add too many ops to the spec, but I think this is one of those cases where clarity of design trumps that concern with adding another op. As example, they are separate ops in TF, Numpy, MATLAB, and PyTorch (in PyTorch there is no isinf, but isnan is just for NaNs not Infs).

wschin:
@spandantiwari, sounds good. I will update this PR to add `IsInf` with `detect_positive` and `detect_negative` attributes.

docs/Changelog.md
wschin: Test files are added and now the strategy becomes
- Not to touch IsNaN
- Add IsInf instead.

docs/Changelog.md
wschin: Ok.

onnx/defs/tensor/defs.cc
wschin: Ok.

spandantiwari:
@wschin - OK, that sounds good. Thanks for your consideration.

spandantiwari:
> @spandantiwari, sounds good. I will update this PR to add `IsInf` with `detect_positive` and `detect_negative` attributes.

LGTM.

PR #1889, opened 2019-03-28
Nullary variadic

gramalingam:
The existing interpretation of a variadic parameter is that it represents 1 or more parameters. However, for a Loop we may have zero or more loop-carried-state parameters. This PR adds an extension to support such ops that may have a variadic parameter representing zero or more parameters. (The PR actually adds an ability to specify a minimum arity, which defaults to 1, which is the original behavior.)


gramalingam:
@houseroad any feedback on this? We could also do this by adding another value in the enum FormalParameterOption (say "ZeroOrMore"). The current form allows variadic ops that require a minimum number of parameters (other than 0 or 1) as well, in case we need it in the future.

onnx/defs/operator_sets.h
```diff
@@ -533,6 +533,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 10, IsInf);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 10, NonMaxSuppression);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 10, ReverseSequence);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 10, RoiAlign);
+class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 10, Loop);
```
houseroad: This should be opset 11, since we already release 1.5. 10 should be frozen.

onnx/defs/operator_sets.h
```diff
@@ -568,14 +564,13 @@ class OpSet_Onnx_ver10 {
...
-           Onnx, 10, IsInf)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 10, IsInf)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(
            Onnx, 10, NonMaxSuppression)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(
            Onnx, 10, ReverseSequence)>());
-     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(
-            Onnx, 10, RoiAlign)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 10, RoiAlign)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 10, Loop)>());
```
houseroad: same here.

onnx/defs/controlflow/defs.cc
```diff
@@ -529,9 +529,9 @@ point-wise operators (e.g. dropout, residual connections, linear layer).
 
 ONNX_OPERATOR_SET_SCHEMA(
     Loop,
-    1,
+    10,
```
houseroad: here, bump up please

onnx/defs/operator_sets.h
gramalingam: Thanks for catching this. I have updated it. (I created this PR a while back when it was still opset 10 and forgot about it now.)

gramalingam:
@houseroad @linkerzhang : any other comments on this PR? Thanks!

PR #1890, opened 2019-03-28
Add Operator to Update Indexed Values

lara-hdr:


lara-hdr:
I could modify the operator's name for "UpdateIndex" or "Update" or other

gramalingam:
How does this op related to Scatter (https://github.com/onnx/onnx/blob/master/docs/Operators.md#Scatter ) ? Will Scatter be sufficient for the current purpose?

lara-hdr:
@gramalingam, both this op and scatter would be used to update values in an input tensor.
However, Scatter has more restrictions that make it impossible to cover all cases of this operator.
Scatter follows an equation to update the values (for example, in a 2-D tensor case, data[indices[i][j]][j] = updates[i][j] with an axis of 0); this makes it hard to update some of the values without updating others.

Let's assume data is : 
 [ [ 1, **2**, **3**],
   [ 4, 5, **6**],]
and we want to update the bold values.

onnx/defs/tensor/defs.cc
```diff
@@ -691,6 +691,117 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+ONNX_OPERATOR_SET_SCHEMA(
+    IndexPut,
+    10,
+    OpSchema()
+        .SetDoc(IndexPut_ver10_doc)
+        .Input(0, "data", "Tensor of rank r >= 1.", "T")
+        .Input(
+            1,
+            "indices",
+            "Tensor of int32/int64 indices of rank q <=r, can be the indexes of a particular element or a slice (Example 2).",
```
```suggestion
            "Tensor of int32/int64 indices of rank q <= r, can be the indexes of a particular element or a slice (Example 2).",
```

onnx/defs/tensor/defs.cc
```diff
@@ -691,6 +691,117 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+      [1, 1, 1],
+      [2, 2, 2],
+      [0, 0, 0],
+      [0, 0, 0],
+      [3, 3, 3]
+  ]
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    IndexPut,
```
wschin: I am not sure if `IndexPut` is a good name. How about `Replace`? It's closer to what it's doing in your examples.

lara-hdr: That's a good name. I'll change the name for Replace

onnx/backend/test/case/node/index_put.py
```diff
@@ -0,0 +1,73 @@
+from __future__ import absolute_import
```
wschin: Please also rename this file.

onnx/backend/test/case/node/index_put.py
```diff
@@ -0,0 +1,73 @@
...
+        updates = np.array([1, 2, 3], dtype=np.float32)
+
+        y = np.array([[0, 1, 0],
+                     [0, 0, 2],
+                     [0, 0, 0],
+                     [0, 0, 0],
+                     [0, 3, 0]], dtype=np.float32)
+
+        expect(node, inputs=[data, indices, updates], outputs=[y],
+               name='test_index_put')
```
wschin: Test names are not renamed.

onnx/backend/test/case/node/index_put.py
lara-hdr: The file is already renamed, this code is outdated :)

onnx/backend/test/case/node/index_put.py
lara-hdr: The file is already renamed, this code is outdated :)

PR #1891, opened 2019-03-29
Update Gather Op

lara-hdr:
Add attribute elem_index to allow indexing on elements in the input.


onnx/defs/tensor/defs.cc
```diff
@@ -806,68 +810,106 @@ Example 2:
...
+        .SetDoc(Gather_ver10_doc)
         .Attr(
             "axis",
             "Which axis to gather on. Negative value means "
-            "counting dimensions from the back. Accepted range in [-r, r-1]",
+            "counting dimensions from the back. Accepted range in [-r, r-1]. ",
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
+        .Attr(
+            "elem_index",
```
spandantiwari: Maybe we can consider a more evocative name for this attribute. Something like, `GatherElement` or `GatherSlice`, or something else.

spandantiwari:
To give some context, the functionality added in this spec is definitely needed. For example, PyTorch's `torch.gather` op would need this to export correctly and efficiently. However, there are two things that we need to consider:

1) The added functionality is almost like a second "mode" of operation, i.e. the interpretation of indices and the shape of the output is completely different from the existing semantics of `Gather`. We can add this new "mode" under this `Gather` op itself, but since these two modes are different (both modes are needed), we should consider adding a new op for this. My worry is that `Gather` is already a somewhat complicated op, and adding more complexity would confuse users. 

2) This will need a deeper look on my side, but on initial look, I think `Gather` and `Scatter` ops in ONNX can be streamlined more in terms of their semantics. On a quick read of the `Scatter` op spec, this second mode being added to `Gather` in this PR may be more in line with current ONNX `Scatter`. For good design it may be useful to make sure that these ops align with each other. (I can take a deeper look).

@gramalingam @linkerzhang @ebarsoum @houseroad - thoughts?

gramalingam:
Yes, these ops are difficult to understand.
(a) I prefer making this a different op, given that it has a distinct meaning. 
(b) I don't see any example added to illustrate this new op/mode. That would be helpful.
(c) I think that the ops will be easier to understand if we add documentation of the form
    output[i][j][k] = input[.....] (for case 1), etc.
as shown in the TF and Torch documentation at https://www.tensorflow.org/api_docs/python/tf/gather and https://pytorch.org/docs/stable/torch.html#torch.gather)

gramalingam:
Just trying to understand the ops first before considering the name for the op. Is the following description correct? We can consider all the axes other than the indexed-axis as one axis called the non-index axis. (E.g., if it helps, we can visualize all the other axes as reshaped into one axis, just to help understand this.) For example, the non-index axis could be used for, say, batching. In the original gather op, we use the same index-tensor to do indexing for all elements of a batch (thinking of the non-index axis as a batch). In the new gather op, each batch-element has its own index-tensor to do the indexing. However, the original gather op allows the index to be multi-dimensional (e.g., we can think of this as "batching" for the index, which is not correlated to the "batching" for the indexed tensor). 

So, in short: with the original mode, we have "output[data-batch][index-batch] = data[data-batch][indices[index-batch]]", while with the new mode we have "output[data-batch][i] = data[data-batch][indices[data-batch][i]]"

Is that correct? I guess it is more important to explain clearly what the two modes do. I am fine with keeping these as different modes of the same op.

onnx/defs/tensor/defs.cc
```diff
@@ -765,10 +765,14 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-static const char* Gather_ver1_doc = R"DOC(
+static const char* Gather_ver10_doc = R"DOC(
 Given `data` tensor of rank r >= 1, and `indices` tensor of rank q, gather
 entries of the axis dimension of `data` (by default outer-most one as axis=0) indexed by `indices`, and concatenates
-them in an output tensor of rank q + (r - 1).
+them in an output tensor.
+If elem_index is disabled (default),  the indices represent the indices of the slices in the axis chosen, and the 
+output is of rank q + (r - 1).
+Otherwise the indices represent the indices of the elements in the tensor in the chosen axis, and the output shape is
+the same as the input shape.
```
gramalingam: Should this be the "the output shape is the same as the indices shape"?

onnx/defs/tensor/defs.cc
```diff
@@ -806,68 +810,106 @@ Example 2:
           [4.5, 5.9],
       ],
   ]
+Example 3:
```
gramalingam: Is this for the new mode? Shouldn't it have "elem_index" attribute?

onnx/defs/tensor/defs.cc
gramalingam: I am not sure I understand the description of the attribute below. It sounds like the main distinction is whether the same indices tensor is used for all elements of the non-indexed axis (the original mode) or whether a different indices (sub)tensor is used for each element of the non-indexed axis (the new mode). Is this correct? If so, may be an attribute like "SharedIndices" or "SameIndices" might help.

lara-hdr:
@gramalingam, correct, "output[data-batch][index-batch] = data[data-batch][indices[index-batch]]" for the old mode, vs "output[data-batch][i] = data[data-batch][indices[data-batch][i]]" for the new mode, is a good way of putting it.
I will add doc similar to TF and Pytorch to explain it.

onnx/defs/tensor/defs.cc
lara-hdr: yes I will update this

onnx/defs/tensor/defs.cc
lara-hdr: yes I will update this

PR #1892, opened 2019-03-29
Quantization tool supporting Conv and MatMul nodes

PhaniShekhar:
Quantization tool for converting an onnx model into quantized onnx model. Currently only conversion of Conv and MatMul nodes is supported.

### DIFF
--- a/tools/quantization/README.md
+++ b/tools/quantization/README.md
@@ -1,0 +1,82 @@
+# Quantization tool Overview
+This tool supports quantization of an onnx model. quantize() takes a model in ModelProto format and returns the quantized model in ModelProto format.
+
+## Quantize an onnx model
+```python
+import onnx
+from quantize import quantize, QuantizationMode
+
+# Load the onnx model
+model = onnx.load('path/to/the/model.onnx')
+# Quantize
+quantized_model = quantize(model, per_channel=False, quantization_mode=QuantizationMode.IntegerOps_Dynamic)
+# Save the quantized model
+onnx.save(quantized_model, 'path/to/the/quantized_model.onnx')
+```
+
+## Quantization modes
+
+The following quantization modes are supported.
+- **QuantizationMode.IntegerOps_Static**:
+    Quantize using integer ops. Only [ConvInteger](https://github.com/onnx/onnx/blob/master/docs/Operators.md#ConvInteger) and [MatMulInteger](https://github.com/onnx/onnx/blob/master/docs/Operators.md#MatMulInteger) ops are supported now.
+    Inputs/activations are quantized using static scale and zero point values.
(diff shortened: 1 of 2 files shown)


tools/quantization/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Quantize an onnx model
+```python
+import onnx
+from quantize import quantize, QuantizationMode
+
+# Load the onnx model
+model = onnx.load('path/to/the/model.onnx')
+# Quantize
+quantized_model = quantize(model, per_channel=False, quantization_mode=QuantizationMode.IntegerOps_Dynamic)
```
askhade: Static mode involves more inputs right? like the quantization params… can you include an example for static mode too...

tools/quantization/quantize.py
```diff
@@ -0,0 +1,899 @@
...
+            z: zero point
+    '''
+    rmin = min(data)
+    rmax = max(data)
+
+    if mode == DataQuantizationMode.Linear_Scaled:
+        max_range = max(abs(rmin), abs(rmax))
+        scale = (float(max_range)*2) / quantize_range
+        zero_point = 0
+        quantized_data = (np.asarray(data) / scale).round().astype('b') #signed byte type
```
askhade: Does this round half to even?

tools/quantization/quantize.py
```diff
@@ -0,0 +1,899 @@
...
+                m = max(abs(rmin), abs(rmax))
+
+        and add necessary intermediate nodes to trasnform quantized weight to full weight using the equation
+        r = S(q-z), where
+            r: real original value
+            q: quantized value
+            S: scale
+            z: zero point
+    '''
+    rmin = min(data)
```
askhade: what happens when range does not include 0? For example is the range is 2-10 then in this case we dont have a unique representation for 0. Can you do rmin = min(min(data), 0) and similar for max...

tools/quantization/quantize.py
```diff
@@ -0,0 +1,899 @@
...
+            Given a weight object, update the graph by doing the following:
+             - remove old initializer, update new initializers for quantized weight, zero point, and scale
+             - remove old weight input, update with new inputs for quantized weight, zero point, and scale
+            This function does NOT update the nodes in the graph, just initializers and inputs
+        '''
+        packed_weight_name = weight.name + '_quantized'
+        scale_name = weight.name + '_scale'
+        zero_point_name = weight.name + '_zero_point'
+
+        # Remove existing weight initializer
```
askhade: What happens when this input is also being used by another node? This condition should be checked.

askhade:
@PhaniShekhar please fix the comments and update the branch with master. Thanks

PR #1910, opened 2019-04-04
Create CODEOWNERS for automatic reviewer assignment for PRs

prasanthpul:


CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* @onnx/sig-archinfra-approvers
```
houseroad: seems archinfra is a super set of almost everything else... if we trigger it every time, other list maybe meaningless? Or if it's onnx/onnx/defs, sig-archinfra-approvers won't be trigger, but instead, it triggers sig-operators-approvers.

prasanthpul: Operators SIG is responsible for operator changes, Currently most of our operator work is in the def files, so changes there should go to the operator sig approvers. I realize some of the tests are outside of that folder, so maybe we need to add more rules.

everything else in the repo should go to archinfra-approvers, until the sig carves out other approver areas

houseroad: yeah, better put `onnx/backend/test` and `doc/` (maybe specific files in doc folder, TestCoverage.md, Operators.md, and *-ml.md) under @onnx/sig-operators-approvers as well

prasanthpul: updated with individual entries for now, but suggest we move all the operator related documentation into an operator subdirectory soon.

PR #1917, opened 2019-04-09
[Draft] Update Operator (and Motivation to Have Wait Operator)

wschin:
One essential concept in training is the update of model parameters based on, for example, their gradient. This PR proposes a new operator for simulating this semantic.

Because the existence of `Update` (or commonly known as `Assign`) can create circular dependencies among operators, we need to introduce another operator, `Wait`, to schedule the computation. Here is a possible signature of `Wait`.

name: `Wait`
input: X, which can be any type.
output: Y, which can be any type.
upstream attribute: a string list which stores names of operators which should be evaluate before producing output Y.
description: `Wait` is conceptually an `Identity` operator with a side effect. `Wait` suspends downstream until the evaluation of all upstream operators is finished.


SherlockNoMad:
I think both X and Y should be input. Just like tf.assign(ref, value). Also, maybe let's call the inputs 'ref' and 'value'. 

Currently, the weights are stored as initializers in an onnx model, if we were to use this op for updating the weights, we are performing in-place update for the initializers. And initializers should be feed as inputs.

gramalingam:
Hi, I don't understand this operator's description at all. All ONNX ops are functional in nature: the output should be computable as a function of the inputs. I would expect an update operator to take the old-value as an input, and the changes/delta as another input.

wschin:
@SherlockNoMad, I am ok with having two inputs because this change doesn't affect this operator's behavior. Just feel `modifying input` is a bit more ambiguous then `modifying output`.

@gramalingam, I will improve its doc and add more context. In addition, `Update` is not necessary to have a delta. Considering gradient descent method with a fixed step length as an example, it first compute gradient, sum up the current model parameters and their gradient, and finally assign the sum back to the model. There are conceptually three steps (and therefore three sub-graphs).

1. Forward graph
2. Backward graph for computing gradient of each adjustable variable
3. Updating graph which combines assign gradient (or so-called `delta`) with the current state and assign the combined value back to model.

This operator aims at capturing the `assignment` in step 3.

gramalingam:
I feel I am missing something important regarding the high-level picture. Is each of the forward/backward/update graphs to be represented by a separate ONNX model? Or, is the goal to have a single model that has these different components? I think some fundamental extension to ONNX may be required, not just adding another operator.

gramalingam:
Or, you could have "Train" as an operator which has 3 graph-valued attributes representing the forward/backward/update parts. Within a single graph, the existing ONNX (static single assignment) requirement means that the same Y cannot appear as the output of two different nodes.

gramalingam:
Or, better still: we could use a single model that takes incoming values of (learnable) weights as well as other inputs, and computes the updated values of the weights as outputs and returns them. Thus, the model combines the three graphs into one, but it has different names for the tensors representing the initial value of a weight at the beginning of an iteration and the tensors representing the final value of the weight in the end of that iteration. If we want, we can represent the iterations also using some form of loop.

gramalingam:
So, I don't understand the need for the proposed "Update" operator.

wschin:
> 
> 
> I feel I am missing something important regarding the high-level picture. Is each of the forward/backward/update graphs to be represented by a separate ONNX model? Or, is the goal to have a single model that has these different components? I think some fundamental extension to ONNX may be required, not just adding another operator.

I don't feel `update` and `backward` should be in separated models. They are designed specifically to the associated forward graph. I like to view them as a single model.

I feel introducing `Sequential` operator will be enough. It allows ordered computation of multiple graphs.

wschin:
> 
> 
> Or, you could have "Train" as an operator which has 3 graph-valued attributes representing the forward/backward/update parts. Within a single graph, the existing ONNX (static single assignment) requirement means that the same Y cannot appear as the output of two different nodes.

I am not sure if `Train` is really a three-graph thing. There are tons of heuristics (such as warm-start) in training neural networks and that's why I propose `Sequential` to allow writing graphical code in ONNX.

wschin:
> 
> 
> Or, better still: we could use a single model that takes incoming values of (learnable) weights as well as other inputs, and computes the updated values of the weights as outputs and returns them. Thus, the model combines the three graphs into one, but it has different names for the tensors representing the initial value of a weight at the beginning of an iteration and the tensors representing the final value of the weight in the end of that iteration. If we want, we can represent the iterations also using some form of loop.

Combining different graphs into one looks super ambiguous. Using `different name` to identify `new value` is a big problem --- how could we make sure all partners follow this rule?

hobei:
@wschin.
I presume that one of the main points of this operation is to update variables based on their gradients. At this stage it is unclear to me how the gradients will be named in the graph and so how this operation is to be used. One proposal is for the backward pass to follow an auto-grad solution, where the backwards operations (and their inputs/outputs) are not explicitly identified, in which case how do we identify the name of the gradient to use in this operation?

wschin:
> 
> 
> @wschin.
> I presume that one of the main points of this operation is to update variables based on their gradients. At this stage it is unclear to me how the gradients will be named in the graph and so how this operation is to be used. One proposal is for the backward pass to follow an auto-grad solution, where the backwards operations (and their inputs/outputs) are not explicitly identified, in which case how do we identify the name of the gradient to use in this operation?

The only person who knows the corresponding between gradient and variable is the auto-grad module, so auto-grad will be responsible to produce those `Update` operators. If gradient (and backward operators) are not explicitly specified, how can ONNX user write their own optimization algorithm? It'd not be scalable to implement an operator to each training algorithm.

shinh:
I think I like outputting updated weights/values as mentioned in https://github.com/onnx/onnx/pull/1917#issuecomment-481631741 . To correspond which `input` should be updated by which `output`, I'd propose to add a repeated message like

```
message UpdateValueProto {
  string input = 1;  // must be one of initializer
  string output = 2;  // must be one of output
}
message GraphProto {
   ...
  repeated UpdateValueProto update;
}
```

(this was mentioned in https://docs.google.com/document/d/1_rfkxCZbvoj-0T9KuDuMLecSpsgJHJK6ExOw4aW-DeY/edit#heading=h.2u9zcp1xt95)

This would look like `Update` op, but I think `Update` op has a downside. If `Update` op exists in `Loop`, you should always run the `Loop` op even if the outputs of `Loop` are not used by other ops. I think this would complicate pruning unnecessary nodes.

Just my two cents

gramalingam:
@wschin : is there some context for this, describing the requirements? That would make it easier to understand what is needed. If we assume that the generator of the ONNX model does the necessary (auto) differentiation, then we just need to capture the computations representing the forward pass, the backward pass, and the update. We can generate an ONNX model representing a single iteration of the training loop (which will be a single model combining all the three steps mentioned above) or we could generate an ONNX model that includes even the training loop. None of these require any changes/extensions to ONNX.

wschin:
> I think I like outputting updated weights/values as mentioned in #1917 (comment) . To correspond which input should be updated by which output, I'd propose to add a repeated message like
> message UpdateValueProto {
>   string input = 1;  // must be one of initializer
>   string output = 2;  // must be one of output
> }
> message GraphProto {
>    ...
>   repeated UpdateValueProto update;
> }
> 
> (this was mentioned in https://docs.google.com/document/d/1_rfkxCZbvoj-0T9KuDuMLecSpsgJHJK6ExOw4aW-DeY/edit#heading=h.2u9zcp1xt95)
> This would look like Update op, but I think Update op has a downside. If Update op exists in Loop, you should always run the Loop op even if the outputs of Loop are not used by other ops. I think this would complicate pruning unnecessary nodes.
> Just my two cents

Right but why would someone add an `Update` to change unused variable? The benefit of having `Update` is not only to make ONNX a formal language but also save 50% of memory.

wschin:
> @wschin : is there some context for this, describing the requirements? That would make it easier to understand what is needed. If we assume that the generator of the ONNX model does the necessary (auto) differentiation, then we just need to capture the computations representing the forward pass, the backward pass, and the update. We can generate an ONNX model representing a single iteration of the training loop (which will be a single model combining all the three steps mentioned above) or we could generate an ONNX model that includes even the training loop. None of these require any changes/extensions to ONNX.

My opinion is we need to describe every bit of the training algorithm, so that training algorithm becomes protable.

We can NOT describe the whole training behavior by defining a single iteration. Let's consider warm-start as an example. One first execute ADAGRAD to get an initial result and then switch to Gauss-Newton method for better final accuracy ---- this is not a thing can be described in an iteration.

We should not mix training and prediction graphs together. They are used for different purposes. ONNX will finally become a language (or dead), so I feel it's fine to steal some features from existing languages.

onnx/defs/controlflow/defs.cc
```diff
@@ -774,4 +774,21 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "(denoted by \"ref\"). This operator can be evaluated once both inputs are "
+            "produced by other operators.")
+        .Input(0, "ref", "input", "T")
+        .Input(0, "value", "input", "T")
+        .TypeConstraint(
+            "T",
+            OpSchema::all_tensor_types(),
+            "Constrain to all tensor types.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          updateOutputElemType(ctx, 0, TensorProto::INT64);
```
wschin: Remove it.

onnx/backend/test/case/model/update.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class Update(Base):
+
+    @staticmethod
+    def export():  # type: () -> None
+        node = onnx.helper.make_node(
+            'Add', ['x', 'y'], ['z'], name='MyAdd')
+        updateNode = onnx.helper.make_node(
+            'Update', ['z'], ['x'], name='MyUpdate')
```
gramalingam: Sorry, this example seems to violate several fundamental principles of ONNX. Unclear if 'x' is an input or output to Update (the above seems to conflict with the opschema). In either case, we have problems. May be it will be easier if we discuss this face to face.

PR #1927, opened 2019-04-11
add ReverseSequence op

nbcsm:
ReverseSequence is commonly used in bi-directional RNN models.
E.g. TensorFlow has [ReverseSequence](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/reverse-sequence) op.

Currently, seems there is no easy way to compose ReverseSequence with existing onnx ops.
Slice cannot deal with batch/time/seq_lengths easily.
It would be very helpful to have ReverseSequence in onnx.


linkerzhang:
@houseroad please help to review. Thank you!

nbcsm:
travis failure is due to timeout, does not seem to be code issue.
could anyone help to rerun it?

onnx/defs/tensor/defs.cc
```diff
@@ -1657,4 +1657,83 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "batch_axis",
+            "(Optional) Specify which axis is batch axis. Must be one of 1 (default), or 0.",
+            AttributeProto::INT,
+            static_cast<int64_t>(1))
+        .Input(0, "input", "Tensor of rank r >= 2.", "T")
+        .Input(
+            1,
+            "sequence_lens",
+            "Tensor specifying lengths of the sequences in a batch. It has shape `[batch_size]`.",
+            "tensor(int32)")
```
houseroad: shall we also support int64 here? since we use int64 as type for indices in most of the cases.

onnx/defs/tensor/defs.cc
```diff
@@ -1657,4 +1657,83 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Output(0, "Y", "Tensor with same shape of input.", "T")
+        .TypeConstraint(
+            "T",
+            OpSchema::all_tensor_types(),
+            "Input and output types can be of any tensor type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
```
houseroad: add a test to cover the following code here: https://github.com/onnx/onnx/blob/master/onnx/test/shape_inference_test.py

nbcsm:
@houseroad, PR updated, please help to take another look. 
thanks.

faxu:
> @houseroad, PR updated, please help to take another look.
> thanks.

@houseroad do you mind reviewing this and merge if all is addressed? we'd like for this to be included in 1.5. Thanks!

onnx/defs/tensor/defs.cc
```diff
@@ -1747,4 +1747,83 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          if (!hasNInputShapes(ctx, 2)) {
+            return;
+          }
+
+          auto& first_input_shape = getInputShape(ctx, 0);
```
houseroad: Is this check really necessary?

onnx/defs/tensor/defs.cc
```diff
@@ -1747,4 +1747,83 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          if (first_input_shape.dim_size() < 2) {
+            fail_shape_inference("'input' must have rank >= 2");
+          }
+          auto& seq_len_input_shape = getInputShape(ctx, 1);
```
houseroad: Is this check really necessary?

nbcsm: according the op spec, ```sequence_lens``` is required to be a scalar.
so maybe it is better to have check?

onnx/defs/tensor/defs.cc
nbcsm: this op is designed to deal with ```input``` with dim_size >=2.
i feel it is better to have check.

nbcsm:
@houseroad, thanks for the review.
i have rebased the PR onto latest master, and replied your comments.
i think it is better to have a check to ensure the inputs fit the op spec requirements, but i can remove the check if you have concern.

onnx/defs/tensor/defs.cc
houseroad: The check is to make sure our shape inference logic won't crash in the middle, or at least print out useful error message. However, I don't the checked shape information is used. So we should move it to the checker. Minimal necessary check is the principle in ONNX shape inference.

nbcsm: i got your point, i will take a look at the checker code.
thanks.

PR #1929, opened 2019-04-11
Shape Inference Tests for QOps

askhade:
* Adding shape inference tests for quantized ops 
* Bug fix for QLinearConv shape inference
* Adding more checks for QLinearConv and QLinearMatMul type inference


askhade:
@houseroad , @linkerzhang : This is the pull request for shape inference tests for quantized ops. Adding node tests for these ops in separate PR as it needs thorough testing.

askhade:
> Looks good. Can you lint the code? One tool I recommend is clang-format, which will only lint your changes.

Thanks @houseroad . I do use that tool not sure why it did not trigger earlier... updated the files with formatting changes.

PR #1931, opened 2019-04-12
Add bit-shift operators for supporting hashing

wschin:
There are two essential operators used in many hashing algorithms. This PR proposes to add [bit-shift](https://en.wikipedia.org/wiki/Bitwise_operations_in_C#Shift_operators) to cover them. Here is reference to a popular [hashing algorithm](https://en.wikipedia.org/wiki/MurmurHash#Implementations).


onnx/defs/logical/defs.cc
```diff
@@ -138,4 +138,68 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrains input/output to boolean tensors.")
         .TypeAndShapeInferenceFunction(unaryLogicalOpInference));
 
+static const char* LeftShift_ver10_doc = R"DOC(
```
linkerzhang: I think it's better to have on op to current both case (right and left).

onnx/defs/logical/defs.cc
```diff
@@ -138,4 +138,68 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+ binary representation toward the right side so that the input value 
+ is decreased. The input X is the tensor to be shifted while another
+ input Y specifies the amounts of shift. For example, if X and S are
+ [1, 4] and [1, 1], respectively, then the corresponding output Z
+ would be [0, 2]. Because this operator supports Numpy-style broadcasting,
+ X's and Y's shapes are not necessarily identical.
+
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
```
linkerzhang: please also add test case for the op and also shape inference test.

linkerzhang:
run https://github.com/onnx/onnx/blob/master/onnx/backend/test/stat_coverage.py to update the test coverage doc (that's why Linux ci failed). meanwhile, a trick when running this script, you'd have to explicitly set ML=false to update the right doc. https://github.com/onnx/onnx/blob/master/onnx/backend/test/stat_coverage.py#L232

I know this kind of doc update experience is bad in windows, it should be improved (some unified script should be created for it).

onnx/defs/logical/defs.cc
wschin: Sure.

onnx/defs/logical/defs.cc
wschin: No problem.

wschin:
> run https://github.com/onnx/onnx/blob/master/onnx/backend/test/stat_coverage.py to update the test coverage doc (that's why Linux ci failed). meanwhile, a trick when running this script, you'd have to explicitly set ML=false to update the right doc. https://github.com/onnx/onnx/blob/master/onnx/backend/test/stat_coverage.py#L232
> I know this kind of doc update experience is bad in windows, it should be improved (some unified script should be created for it).

Just tried some combinations of reinstall/ONNX_ML=0 and none of them worked..

PR #1938, opened 2019-04-16
Fix failed Window dll build and refine the corresponding logic

daquexian:
The logic in onnx_pb.h is wrong, and the logic in CMakeLists.txt just follows the wrong one.

ONNX_IMPORT is only set when ONNX_BUILD_SHARED_LIBS or ONNX_BUILD_MAIN_LIB _is_ defined, but only used when ONNX_BUILD_SHARED_LIBS or ONNX_BUILD_MAIN_LIB _is not_ defined. It is not the expected behavior and causes windows shared lib build fails (CI doesn't check it).

CI log: https://ci.appveyor.com/project/daquexian/dnnlibrary/builds/23876196#L3723 (without this patch, build fails on unresolved symbols of protobuf generated files)
https://ci.appveyor.com/project/daquexian/dnnlibrary/builds/23876935#L13092 (with this patch, build passes)


daquexian:
What's more the code of ONNX_API is confusing. Why is ONNX_API set twice? What is the correct way to set it for projects depending on ONNX?

daquexian:
Hmm.. CI fails since ONNX_BUILD_MAIN_LIB is set to OFF, and then dllimport is used.

I believe it is that "two wrongs make a right"

daquexian:
I refined the logic of ONNX_EXPORT and ONNX_IMPORT. 

1. The redundant definitions in CMakeLists.txt are removed. In CMakeLists.txt, the ONNX_API is set only for compiling onnx_proto itself, so nothing about visibility need to be set.

2. Two new definitions, ONNX_IMPORT_SYMBOLS and ONNX_EXPORT_SYMBOLS are introduced. The documentation is also updated accordingly.

CMakeLists.txt
```diff
@@ -303,18 +303,24 @@ if (MSVC)
...
-  if (BUILD_SHARED_LIBS OR ONNX_BUILD_MAIN_LIB)
-    set(ONNX_API_DEFINE "-DONNX_API=__declspec(dllexport)")
+  if (BUILD_SHARED_LIBS)
+    target_compile_definitions(onnx_proto INTERFACE "-DONNX_IMPORT_SYMBOLS")
+    # ONNX_API is set here for compiling onnx_proto itself
+    target_compile_definitions(onnx_proto PRIVATE "-DONNX_API=__declspec(dllexport)")
   else()
-    set(ONNX_API_DEFINE "-DONNX_API=")
+    if (ONNX_BUILD_MAIN_LIB)
+      target_compile_definitions(onnx_proto PUBLIC "-DONNX_EXPORT_SYMBOLS")
```
daquexian: There are two ONNX_BUILD_MAIN_LIB in the original code, a cmake variable (in here) and a compile definition (in onnx_pb.h). 
PUBLIC here is because in the original code ONNX_BUILD_MAIN_LIB cmake variable set the PRIVATE part and the ONNX_BUILD_MAIN_LIB definition in onnx_pb.h set the INTERFACE part. However, I haven't met a use case of both the cmake variable and compile definition.

daquexian:
I feel confused about the BUILD_SHARED_LIBS and ONNX_BUILD_MAIN_LIB so I conclude the logic about them as following. I think it respect the original behavior, correct me if I am wrong :

 **BUILD_SHARED_LIBS:**
We want to build onnx_proto as dll. In this case, we set ONNX_API=export for onnx_proto itself and set ONNX_API=import for anything depending on onnx_proto.

**ONNX_BUILD_MAIN_LIB and BUILD_SHARED_LIBS:**
No difference from above.

**ONNX_BUILD_MAIN_LIB and not BUILD_SHARED_LIBS:**
We use ONNX as a subproject and want to build onnx_proto as a lib and the whole project as a dll. In this case, we want ONNX_API to be "export" for onnx_proto itself to export the symbols to dll, and also for anything depending on onnx_proto to "fix a Windows global-variable-in-dll issue" according to documentation in onnx_pb.h.

snnn:
Well, your logic is correct, but
1. why do people want to build onnx_proto as a separated dll, instead of putting onnx and onnx_proto together?
2. So far, I didn't see onnx or onnx_proto was marked as shared lib. There should be something like:

add_library(onnx_proto **SHARED** ${ONNX_PROTO_SRCS} ${ONNX_PROTO_HDRS})

I didn't see it.

snnn:
> We use ONNX as a subproject and want to build onnx_proto as a lib and the whole project as a dll
In general. "__declspec(dllexport) " in static lib doesn't work.  "https://devblogs.microsoft.com/oldnewthing/20140321-00/?p=1433"

daquexian:
@snnn Thanks for your review!

> 1. why do people want to build onnx_proto as a separated dll, instead of putting onnx and onnx_proto together?

Frankly speaking, I didn't know when people want to build onnx_proto as a separate dll, either. However, an `onnx_proto.dll` may be generated by `BUILD_SHARED_LIBS` global flag. My edition remains 
the behavior of the original code as far as possible.

> 2. So far, I didn't see onnx or onnx_proto was marked as shared lib. There should be something like:
> 
> add_library(onnx_proto **SHARED** ${ONNX_PROTO_SRCS} ${ONNX_PROTO_HDRS})
> 
> I didn't see it.

Since we didn't add "SHARED" or "STATIC" explicitly here, whether onnx_proto is a shared lib or a static lib is controlled by the global [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) flag. So if onnx is a submodule of another project, it is possible that onnx_proto is a shared lib.

> And what is the "Windows global-variable-in-dll issue"? How did it get fixed?

It is directly from the comment in onnx_pb.h, which is introduced by @linkerzhang from the result of git blame. Maybe we need to discuss with Ke of this specific question. :)

snnn:
If you want to build the whole project as a dll, I think the more straight forward way is adding a new target.
add_library(onnx_whole_dll SHARED ${ONNX_SRCS} ${ONNX_PROTO_SRCS} ${ONNX_PROTO_HDRS})
As I mentioned, "__declspec(dllexport)" doesn't work in static lib. And also, this dll should dynamically link to libprotobuf.

daquexian:
> In general. "__declspec(dllexport) " in static lib doesn't work. "https://devblogs.microsoft.com/oldnewthing/20140321-00/?p=1433"
> 
> May I know your use case?

I think what this article points out is that if the obj file is not seen by linker, then the dllexport will not work. 

If the obj file is seen by linker, and the static lib is linked into a shared lib, dllexport will work. Although I have not verified it, exporting symbols of a static onnx lib when linked into a shared lib is exactly what `ONNX_BUILD_MAIN_LIB` wants to do. Reference: https://github.com/onnx/onnx/blob/master/onnx/onnx_pb.h#L34

daquexian:
> If you want to build the whole project as a dll, I think the more straight forward way is adding a new target.
> add_library(onnx_whole_dll SHARED ${ONNX_SRCS} ${ONNX_PROTO_SRCS} ${ONNX_PROTO_HDRS})

I agree. It is unwise to build an onnx_proto.dll. However, it will happen if `BULID_SHARED_LIBS` is set.

snnn:
In general, DLL shouldn't export a symbol which is provided in another lib. Otherwise the symbol should be handled in a special way.

daquexian:
> In general, DLL shouldn't export a symbol which is provided in another lib. Otherwise the symbol should be handled in a special way.

I have no experience in windows programming, and just try to remain the logic of original code ( however, the original logic is confusing). If you are sure that we shouldn't export symbol in a static onnx lib to a shared lib, I will be glad to remove the confusing `ONNX_BUILD_MAIN_LIB`

RyanUnderhill:
> In general, DLL shouldn't export a symbol which is provided in another lib. Otherwise the symbol should be handled in a special way.
> 
> I have no experience in windows programming so I chose to remain the logic of original code as far as possible (however, the original logic is confusing). If you are sure that we shouldn't export symbol in a static onnx lib to a shared lib, I will be glad to remove the confusing ONNX_BUILD_MAIN_LIB

From the outside it's not visible how the symbol was exported, but it's a rather unexpected to link to a library and cause your DLL to export additional symbols.

daquexian:
Thanks @snnn @RyanUnderhill 

> From the outside it's not visible how the symbol was exported, but it's a rather unexpected to link to a library and cause your DLL to export additional symbols.

Frankly speaking, I believe we need check the logic about dll thoroughly and refactor it.

snnn:
Please hold on a moment.  We completely understand the current setting first. To me, it's even not clear why this option exists.

linkerzhang:
@yinghai @bddppq  any comments on "Should we remove the ONNX_BUILD_MAIN_LIB completely?"

linkerzhang:
Aha, I noticed that I added these stuff, but my memory is telling me that they're from some other file and I just did the movement.

How about remove it and see whether there're concerns from other folks? :)

daquexian:
@linkerzhang @snnn I have removed ONNX_BUILD_MAIN_LIB

Another question is, `ONNX_EXPORT` in the header `onnx_pb.h` is never used now. The reason is that only onnx_proto exports symbols, onnx and libraries/executables depending on onnx import symbols from onnx_proto, and `onnx_pb.h` is only included by the latter ones rather than onnx_proto.

Should `ONNX_EXPORT` be removed too?

onnx/onnx_pb.h
```diff
@@ -8,14 +8,9 @@
 // different declarations (dllexport and dllimport). On Linux/Mac, it just
 // resolves to the same "default visibility" setting.
 #if defined(_MSC_VER)
-#if defined(ONNX_BUILD_SHARED_LIBS) || defined(ONNX_BUILD_MAIN_LIB)
```
snnn: Why do you remove this line?
What if people want to build a static lib?
What if people want to consume this header and static link to onnx?

snnn:
Are you trying to split onnx into two DLLs: onnx.dll and onnx_proto.dll? If not, you don't need to change any code here.

onnx/onnx_pb.h
daquexian: Thanks for your review. This line is not removed accidentally in the commit of removing ONNX_BUILD_MAIN_LIB.

The dllexport and dllimport are now managed by ONNX_EXPORT_SYMBOLS and ONNX_IMPORT_SYMBOLS, which are set properly in cmake.

The reason to change this line is to extract the common codes. For example, in the original code, there are two "ONNX_BUILD_MAIN_LIB", one is [a cmake variable](https://github.com/onnx/onnx/blob/master/CMakeLists.txt#L306) and another is [a compiler definition](https://github.com/onnx/onnx/blob/master/onnx/onnx_pb.h#L11). Setting the cmake variable to ON doesn't define its namesake. How can users define this compiler definition? No easy way. I believe it is not the expected behavior. This bug clearly shows that it is error-prone to maintain the duplicated logic in two different places. So I extract the logic to cmake, and set the two extra definitions to control the behavior of onnx_pb.h.

daquexian:
> Are you trying to split onnx into two DLLs: onnx.dll and onnx_proto.dll? If not, you don't need to change any code here.

Thanks for your review! I am not trying to split onnx into two dlls. onnx looks already split into [onnx_proto](https://github.com/onnx/onnx/blob/master/CMakeLists.txt#L284) and [onnx](https://github.com/onnx/onnx/blob/master/CMakeLists.txt#L339). Please correct me if I am wrong.

snnn:
Then you can do this:
1. In your project, add a new dynamic lib target named 'onnx_dll'
2. Let onnx_dll link to onnx and onnx_proto, group them within wholearchive flag
3. define ONNX_BUILD_SHARED_LIBS when building onnx and onnx_proto

Would it work?

snnn:
BTW, is there an option for forcing onnx dynamically link to libprotobuf.dll on Windows? If your program will dynamic link to onnx, then it should also dynamic link to protobuf, as onnx's interface contains protobuf symbols.

daquexian:
> BTW, is there an option for forcing onnx dynamically link to libprotobuf.dll on Windows? If your program will dynamic link to onnx, then it should also dynamic link to protobuf, as onnx's interface contains protobuf symbols.

Yes, I think it is doesn't matter that the cmake target `protobuf::libprotobuf` is a shared library or a static library. It is the [build log in CI](https://ci.appveyor.com/project/daquexian/dnnlibrary/builds/23876935#L13092) of my own project, it compiles and links successfully (line 13092).

daquexian:
> Then you can do this:
> 
> 1. In your project, add a new dynamic lib target named 'onnx_dll'
> 2. Let onnx_dll link to onnx and onnx_proto, group them within wholearchive flag
> 3. define ONNX_BUILD_SHARED_LIBS when building onnx and onnx_proto
> 
> Would it work?

I am not familiar with windows programming. Looks like it will work. It reminds me of the `ONNX_BUILD_MAIN_LIB` :D 

BTW, what do you think of my opinion [here](https://github.com/onnx/onnx/pull/1938#discussion_r284531608)?  :)

onnx/onnx_pb.h
snnn: Yes, it's a bug. It can be easily fixed by adding:
```
if(ONNX_BUILD_MAIN_LIB)
  target_compile_definitions(xxx  PRIVATE ONNX_BUILD_MAIN_LIB)
endif()
```
to cmake files.

snnn:
> Yes, I think it is doesn't matter that the cmake target `protobuf::libprotobuf` is a shared library or a static library. It is the [build log in CI](https://ci.appveyor.com/project/daquexian/dnnlibrary/builds/23876935#L13092) of my own project, it compiles and links successfully (line 13092).

It matters. You may hit link error at later time when you are using the dll. Not always, but it may happen.

snnn:
Let's say:
If you have a library, foo.dll, which depends on libprotobuf.

Then, let's say, if it doesn't have protobuf symbols on its interface, it's better to let foo.dll static link to protobuf.
otherwise,  if it has, and it choose to static link to protobuf, and you have an app named app.exe, which depends on foo.dll, also, protobuf. Because foo.dll has protobuf in its interface,  app.exe can't use foo.dll without include protobuf headers.
However, app.exe may or may not need to link to protobuf. Because foo.dll will provide some of the protobuf symbols but not all of them. 
If app.exe also link to libprotobuf, whether statically or dynamically, you may have duplicated symbols between app.exe and foo.dll. I mean, two copies of one symbol, and linker won't complain. Because it's not an error, it's the design. If you know ODR, you should also know in DLL world, it's broken. And it would be a disaster if you allocate a protobuf object in one place(e.g. app.exe) but  deallocate it in another place(e.g. foo.dll). Because protobuf has a local memory pool, if you have two copies of the same symbol, you'll have two copy of the pool.  If you allocate it in one pool but deallocate it in another, it will just crash. However, it often happens when you are passing protobuf objects between dlls.

So, as long as you need to pass the ownership of one protobuf from one dll(or exe) to another, you should let them dynamic link to libprotobuf.

It's a not problem specific to Windows. Linux has the similar problem.

onnx/onnx_pb.h
linkerzhang: I personally prefer this idea of removing the duplicated logic in both header and cmake files. :)

onnx/onnx_pb.h
```diff
@@ -31,17 +26,22 @@
...
 // https://msdn.microsoft.com/en-us/library/3y1sfaz2.aspx
 //
 // This solution is similar to
 // https://github.com/pytorch/pytorch/blob/master/caffe2/core/common.h
-#if defined(ONNX_BUILD_SHARED_LIBS) || defined(ONNX_BUILD_MAIN_LIB)
+
+#if defined(_MSC_VER)
+// dllimport will fail non-dll build, so extra flags is used.
+#if defined(ONNX_EXPORT_SYMBOLS)
 #define ONNX_API ONNX_EXPORT
```
linkerzhang: this is also not needed? Anyway, let's clean up firstly, and we may add it back if needed later :)

linkerzhang:
@yinghai @houseroad  any concerns from your side on this change please?

snnn:
Could you please also clean up the comments?
Please describe what's the relationship between

1. ONNX_BUILD_SHARED_LIBS
2. ONNX_BUILD_MAIN_LIB
3. ONNX_IMPORT_SYMBOLS
4. ONNX_API 

When should they get defined.

Thanks.

daquexian:
> Could you please also clean up the comments?
> Please describe what's the relationship between
> 
> 1. ONNX_BUILD_SHARED_LIBS
> 2. ONNX_BUILD_MAIN_LIB
> 3. ONNX_IMPORT_SYMBOLS
> 4. ONNX_API
> 
> When should they get defined.
> 
> Thanks.

All right. I find the origin of `ONNX_BUILD_SHARED_LIBS` and `ONNX_BUILD_MAIN_LIB`. It is from  Caffe2 and especially this commit https://github.com/pytorch/pytorch/commit/91d76f5dbde782396d2777c9d2ae5cde2c83160d. After checking this commit I figured out these definitions.

daquexian:
The history of these macros:

**ONNX_API**
The usage of _MYDLL_API_ should be like it: https://stackoverflow.com/a/26966255/5283216. When building the dll, _MYDLL_API_ should be dllexport, and when building a program depending on the dll, _MYDLL_API_ should be dllimport.

**ONNX_BUILD_SHARED_LIBS**
However, we cannot always set _ONNX_API_ as "dllimport" when linking onnx_proto, because we need support not only shared lib but also static lib. So a definition _ONNX_BUILD_SHARED_LIBS_ is introduced.

**ONNX_BUILD_MAIN_LIB**
_XXX_BUILD_MAIN_LIB_ is from Caffe2, and indeed not needed for ONNX.

We first learn how it works in Caffe2: In Caffe2, the exported symbols consist of symbols in generated pb.cc/pb.h and also a few symbols in the source code (please check out this commit: https://stackoverflow.com/a/26966255/5283216). In the source code, _CAFFE2_API_ is set in caffe2/core/macros.h (a common header included by every headers/sources). In pb.h/pb.cc, _CAFFE2_API_ is set by cmake since they are generated and do not include any Caffe2 headers. The compiler definition _CAFFE2_BUILD_MAIN_LIB_ is then introduced, which is [always set](https://github.com/pytorch/pytorch/blob/91d76f5dbde782396d2777c9d2ae5cde2c83160d/caffe2/CMakeLists.txt#L102) when compiling Caffe2. As a result, the symbols in the **source code** are exported when building Caffe2 and are imported when linking to Caffe2.

However, in ONNX, only symbols in pb.cc/pb.h are exported, no _ONNX_API_ is in the source code. So in onnx/onnx_pb.h, _ONNX_API_ is never to be "export". The only use case of _ONNX_API_ is importing the symbols of pb.h/pb.cc. So 1. _ONNX_BULID_MAIN_LIB_ is not needed. 2. _ONNX_API_ is always "import".

daquexian:
Setting ONNX_API in cmake makes pytorch build [broken](https://circleci.com/gh/onnx/onnx/1941?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). The problem is not on ONNX side but on PyTorch side. ONNX is linked privately in [caffe2/CMakeLists.txt]( https://github.com/pytorch/pytorch/blob/master/caffe2/CMakeLists.txt#L853) so public compiler definitions of ONNX cannot be spread to targets depending on caffe2 (e.g., [caffe2_pybind11_state](https://github.com/pytorch/pytorch/blob/master/caffe2/CMakeLists.txt#L1198)) (It not only causes the CI fails but also causes [these hardcoded definitions](https://github.com/pytorch/pytorch/blob/master/cmake/Dependencies.cmake#L1044)). 

So I defined an ONNX_API macro in macros.h now, which is generated by cmake.

snnn:
Don't create a DLL that:
1. has a libprotobuf inside
2. And exported some of the libprotobuf symbols in its interface

You should:
1. build protobuf as a dynamic DLL
2. let your library dynamic link to protobuf

daquexian:
Closing because it is outdated. I don't know there is a proper way to solve this issue (or whether this issue has already been fixed)

PR #1941, opened 2019-04-16
Fix shape inference for matmul

BowenBao:
Recent commit changes the logic
https://github.com/onnx/onnx/blob/6de42d7d7f8ad6e601a8388147400b344b2728ea/onnx/defs/math/defs.cc#L774
to
https://github.com/onnx/onnx/blob/3717dc617fa06e4eea326e85dc0ccfdcdf4f4ab5/onnx/defs/math/defs.cc#L754
to accommodate qlinearmatmul when there are more than 2 inputs.
However it seems the logic should be
```
if (!hasInputShape(ctx, input1Idx) || !hasInputShape(ctx, input2Idx)) {
```

This is causing a bug when only one of the input does not have shape. Also these cases are not captured in the previous unit tests.

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -751,7 +751,7 @@ void matmulShapeInference(
     ONNX_NAMESPACE::InferenceContext& ctx,
     int input1Idx,
     int input2Idx) {
-  if (!hasInputShape(ctx, input1Idx) && !hasInputShape(ctx, input2Idx)) {
+  if (!hasInputShape(ctx, input1Idx) || !hasInputShape(ctx, input2Idx)) {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


linkerzhang:
Thank you very much!

onnx/defs/math/defs.cc
```diff
@@ -751,7 +751,7 @@ void matmulShapeInference(
     ONNX_NAMESPACE::InferenceContext& ctx,
     int input1Idx,
     int input2Idx) {
-  if (!hasInputShape(ctx, input1Idx) && !hasInputShape(ctx, input2Idx)) {
+  if (!hasInputShape(ctx, input1Idx) || !hasInputShape(ctx, input2Idx)) {
```
houseroad: Thanks for fixing this bug!

PR #1942, opened 2019-04-17
README.md: Update badge style for build status

fs-eire:
Change to a slim stype. Use [shields.io](https://shields.io/) so that label can be embedded.


raymondxyang:
@fs-eire please sign the CLA thx

PR #1944, opened 2019-04-17
add node tests for quantized ops

askhade:
Adding tests for all 6 quantized ops. I have validated these tests using onnxruntime.


askhade:
@houseroad , @linkerzhang : This PR is to add node tests for the quantized ops. Please take a look.

PR #1945, opened 2019-04-17
Support dynamic 'pads' and 'value' in Pad operator

hariharans29:
* Make interface change in Pad operator as described in title
    ** Additionally support 2D pads input of shape [1, 2 * input_rank]. This is useful if using Flatten operator (which produces 2D output only) to flatten pads input while converting tensorflow models to onnx. 
* Make related node test case changes
* Make related shape inference test changes
* Make related changes in optimizer passes
* Make related optimizer test changes


hariharans29:
CC: @linkerzhang

onnx/defs/tensor/defs.cc
```diff
@@ -1657,4 +1567,119 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  ]
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    Pad,
+    10,
+    OpSchema()
+        .Attr(
+            "mode",
+            "Three modes: constant(default), reflect, edge",
```
linkerzhang: Add more description or reference to clarify these modes?

onnx/defs/tensor/defs.cc
```diff
@@ -1657,4 +1567,119 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "at the beginning and end of each axis. For 2D input tensor, it is the number of pixels. "
+            "`pads` should be a 1D tensor of shape [2 * input_rank] or a 2D tensor of shape [1, 2 * input_rank]. "
+            "`pads` format (1D example) should be as follow [x1_begin, x2_begin,...,x1_end, x2_end,...], "
+            "where xi_begin is the number of pixels added at the beginning of axis `i` and "
+            "xi_end, the number of pixels added at the end of axis `i`.",
+            "T1")
+        .Input(
+            2,
+            "value",
+            "(Optional) Rank 1 tensor containing 1 float indicating the value to be filled if the mode chosen is `constant` (by default it is 0.0f).",
```
linkerzhang: More clarification about this 1-D tensor please (though the previous one is also not clear enough :)). Its shape should be the same as "pads".

onnx/defs/tensor/defs.cc
```diff
@@ -1657,4 +1567,119 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T2",
```
linkerzhang: padding value's type should be the same as input "data".

onnx/defs/tensor/defs.cc
```diff
@@ -1657,4 +1567,119 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(
+            1,
+            "pads",
+            "Tensor of integers indicating the number of padding elements to add or remove (if negative) "
+            "at the beginning and end of each axis. For 2D input tensor, it is the number of pixels. "
+            "`pads` should be a 1D tensor of shape [2 * input_rank] or a 2D tensor of shape [1, 2 * input_rank]. "
+            "`pads` format (1D example) should be as follow [x1_begin, x2_begin,...,x1_end, x2_end,...], "
+            "where xi_begin is the number of pixels added at the beginning of axis `i` and "
+            "xi_end, the number of pixels added at the end of axis `i`.",
+            "T1")
```
linkerzhang: replace "T1" with "tensor(int64)", as T1 only has one option defined.

linkerzhang:
@houseroad  @spandantiwari this update breaks pytorch test again, I think. Please look into it accordingly. Thank you!

onnx/defs/tensor/defs.cc
hariharans29: @linkerzhang - Actually the shape need not be the same as 'pads' , this the equivalent of the previous 'value' attribute - it is just a single float value (a scalar essentially) - so it's shape should be either [] or [1] - I ll update it

onnx/defs/tensor/defs.cc
hariharans29: I will update this

onnx/defs/tensor/defs.cc
hariharans29: Fixed.

onnx/defs/tensor/defs.cc
hariharans29: Fixed.

onnx/defs/tensor/defs.cc
hariharans29: Updated.

onnx/defs/tensor/defs.cc
linkerzhang: I thought you would like to pad different values in each direction. if it's a [] or [1], then let's clarify it as a "Scalar".

hariharans29: It is explicitly mentioned as a "scalar" in the latest version

linkerzhang:
@houseroad @bddppq  please kindly review this op update. Thank you!

hariharans29:
> Please bump up to opset 11.

@houseroad - Done

linkerzhang:
@houseroad  any more comments please? Thank you!

hariharans29:
Please hold off merging this until #2005 is merged

hariharans29:
Pinging....

@houseroad

hariharans29:
> Can we make the optimizer works for both version of Pad operator?

Hi @houseroad - can we test the same in optimizer_test.py (by creating a Pad v10 node) ? Is it possible to create an older opset node ?

hariharans29:
Closing this in favor of #2031

PR #1950, opened 2019-04-18
Fix shape inference for slice

### LINKED ISSUE #1949: shape inference bug of op "Slice 10"
this shape inference function fails even in simple cases, for example a ONNX graph that only contains 1 slice whose inputs such as "starts" are all const nodes except the "data".

hariharans29:
Resolves #1949


onnx/defs/tensor/defs.cc
```diff
@@ -536,96 +537,156 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                initializer->data_type() == TensorProto::INT64) {
               const std::string& bytes = initializer->raw_data();
               vec.insert(
                   vec.end(),
                   reinterpret_cast<const int64_t*>(bytes.c_str()),
                   reinterpret_cast<const int64_t*>(
                       bytes.c_str() + bytes.size()));
-            } else {
+            } else if (
+                initializer->has_raw_data() &&
```
snnn: It may have byte-order issue.

hariharans29: you mean endianness of the machine we are trying to parse the raw data ? That's a good point.  Thanks. I will fix it.

snnn: You may use [this](https://github.com/Microsoft/onnxruntime/blob/master/onnxruntime/core/framework/tensorprotoutils.cc#L49) as a reference.

hariharans29: Oh I didn't see your comment - I implemented it already and tested it. You may take a look.

snnn:
LGTM.

linkerzhang:
@houseroad @spandantiwari please take care of the pytorch failure accordingly. Thank you! btw, @hariharans29 the pytorch ci failure is not a blocker for the PR merge. Will merge this later but give pytorch folks some time  to be aware of it.

PR #1951, opened 2019-04-19
Fix shape inference for ConstantOfShape op

askhade:
Bug fix for ConstantOfShape shape inference. mutable_shape for output should be called only when we can do the shape inference.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -313,15 +313,6 @@ ONNX_OPERATOR_SET_SCHEMA(
 
           auto rank = ctx.getInputType(0)->tensor_type().shape().dim_size();
 
-          // rank == 0 may mean that shape inference can't be done yet.
-          // today shape inference is not complete for all the operators.
-          // when such an op preceeds concat rank can be 0.
-          // TODO : Fix shape inferencing for all ops to fall back to rank
-          // inference when shape inferening cant be done.
-          if (rank == 0) {
-            return;
-          }
-
           auto axisAttr = ctx.getAttribute("axis");
           if (!axisAttr) {
             fail_shape_inference("Required attribute axis is missing");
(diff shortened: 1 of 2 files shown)


onnx/defs/tensor/defs.cc
```diff
@@ -313,6 +313,15 @@ ONNX_OPERATOR_SET_SCHEMA(
 
           auto rank = ctx.getInputType(0)->tensor_type().shape().dim_size();
 
+          // rank == 0 may mean that shape inference can't be done yet.
```
linkerzhang: maybe it's better to check whether ctx.getInputType(0)->tensor_type().has_shape().

This reminds me that a "Status" is still needed to be returned for shape inference function, so that, frameworks calling shape inference in a graph can do early stop (without having to do that via throwing and catching exceptions).

@gramalingam  thoughts?

gramalingam: The existing check for “hasNInputShapes(…)” already checks for this. So, I am guessing that the real bug lies elsewhere … it could be in the shape-inference for a preceding node.   

This is somewhat fragile, because of the way “protobuf”’s optional fields work. The “shape” field is optional and not set if it is not known (e.g., if it’s rank is not known). So, every line of code that tries to access “.shape()” should first check if it “.has_shape()”. Otherwise, protobuf accessors will allocate the shape to a default-value when someone accesses it. So, my guess is that some recently added shape-inference code accesses the “.shape()” field without first checking if it “.has_shape()”. 

It is likely to be in the shape-inference for some preceding node that has the same tensor-variable as input or output.

askhade: As @gramalingam mentioned:  ctx.getInputType(0)->tensor_type().has_shape() is already being checked as part of hasNInputShapes(…)

lucienwang1009: rank() == 0 may mean the tensor is a scalar. And concatenating a scalar with an array is obviously wrong.

PR #1955, opened 2019-04-21
[Training] Add Adagrad optimizer operator

wschin:
PR #2314 is a single place for reviewing the whole training story.

This PR proposes a stateless optimizer based on a stochastic gradient method. This operator covers [Tensorflow](https://www.tensorflow.org/api_docs/python/tf/train/AdagradOptimizer)'s and [Pytorch](https://pytorch.org/docs/stable/_modules/torch/optim/adagrad.html#Adagrad)'s ADAGRAD optimizer.

Optimizer itself can be defined as a stateful operator which, for example, maintain accumulated momentum and accumulated squared gradient. However, this way might implicitly introduce `assignment` semantic into ONNX and therefore breaks SSA assumption.

Below is a script I used to verify the Adagrad function added in this PR's test. It shows that consistent results in TF, Pytorch, and the Adagrad function used to create ONNX tests in this PR.
```python
import numpy as np
import torch
import tensorflow as tf
from torch import optim
from torch import nn
import torch.random as rnd

lr=0.2
lr_decay=0.0
weight_decay=0.0
n = 10
l = 5
X_ = torch.randn(l, n)
Y_ = torch.randn(l, 1)

def apply_adagrad(r, t, x, g, h, norm_coefficient, epsilon, decay_factor):  # type: ignore
    # Compute adjusted learning-rate.
    r_ = r / (1 + t * decay_factor)
    # Add gradient of regularization term.
    g_regularized = norm_coefficient * x + g
    # Update squared accumulated gradient.
    h_new = h + g_regularized * g_regularized
    # Compute ADAGRAD's gradient scaling factors
    h_sqrt = np.sqrt(h_new) + epsilon
    # Apply ADAGRAD update rule.
    x_new = x - r_ * g_regularized / h_sqrt
    return (x_new, h_new)

def show_pytorch():
    X = X_.clone()
    Y = Y_.clone()

    rnd.manual_seed(0)
    model = nn.Sequential(
        nn.Linear(n, 1, bias=False)
    )

    loss_fn = nn.MSELoss(reduction='sum')

    solver = optim.Adagrad(model.parameters(), lr=lr, lr_decay=lr_decay, weight_decay=weight_decay)

    for t in range(8):
        Y_pred = model(X)
        loss = loss_fn(Y_pred, Y)
        model.zero_grad()
        loss.backward()
        solver.step()
        print(loss)

def show_tensorflow():
    rnd.manual_seed(0)
    layer = nn.Linear(n, 1, bias=False)

    X = tf.placeholder('float', shape=[l, n])
    W = tf.Variable(torch.Tensor.numpy(layer.weight.detach()))
    Y = tf.placeholder('float', shape=[l, 1])
    Y_pred = tf.matmul(X, W, transpose_b=True)
    loss = tf.reduce_sum(tf.square(Y - Y_pred))
    optimizer = tf.train.AdagradOptimizer(learning_rate=lr, initial_accumulator_value=0.0000000001)
    minimizer = optimizer.minimize(loss)

    sess = tf.Session()

    init = tf.global_variables_initializer()

    sess.run(init)

    for t in range(8):
        print(sess.run(loss, {X: X_, Y: Y_}))
        result = sess.run([minimizer], {X: X_, Y: Y_})


def show_onnx():
    X = X_.clone()
    Y = Y_.clone()

    rnd.manual_seed(0)
    model = nn.Sequential(
        nn.Linear(n, 1, bias=False)
    )

    loss_fn = nn.MSELoss(reduction='sum')

    solver = optim.Adagrad(model.parameters(), lr=lr, lr_decay=lr_decay, weight_decay=weight_decay)

    for t in range(8):
        Y_pred = model(X)
        loss = loss_fn(Y_pred, Y)
        model.zero_grad()
        loss.backward()

        with torch.no_grad():
            for param in model.parameters():
                new_tensor, new_state = apply_adagrad(r=lr, t=t, x=param.data, g=param.grad.data, h=solver.state[param]['sum'].data,
                                                      norm_coefficient=weight_decay, epsilon=1e-10, decay_factor=lr_decay)
                solver.state[param]['sum'].data = new_state.data
                param.data = new_tensor.data

        print(loss)

show_tensorflow()
print('---------')
show_pytorch()
print('---------')
show_onnx()
```


SherlockNoMad:
I echo that having a stateless optimizer is a good design with existing ONNX without breaking SSA assumption. 

Different backends can come up with their own implementation on how to handle the new state and update the weights.

onnx/defs/controlflow/defs.cc
```diff
@@ -774,4 +774,105 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            OpSchema::Variadic,
+            false)
+        .Output(0, "T_new", "New update count of inputs", "T3")
+        .Output(
+            1,
+            "outputs",
+            "It sequentially contains the new values of optimized tensors and then the new "
+            "values of accumulated gradient. For example, if two tensor \"X_1\" and \"X_2\" are "
+            "optimized, the output list would be [new value of \"X_1,\" new value of \"X_2\" "
+            "new accumulated gradient of \"X_1\", new accumulated gradient of \"X_2\"].",
```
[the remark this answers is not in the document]
wschin: Right. It should be called `new accumulated squared gradient`.

onnx/defs/controlflow/defs.cc
```diff
@@ -774,4 +774,105 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T2",
+            OpSchema::Variadic,
+            false)
+        .TypeConstraint(
```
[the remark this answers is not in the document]
wschin: One for tensor and the other one for scalar.

onnx/defs/controlflow/defs.cc
```diff
@@ -774,4 +774,105 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            5,
+            "inputs",
+            "It sequentially contains the current values of optimized tensors and then the "
+            "current values of accumulated gradient. For example, if two tensor \"X_1\" and \"X_2\" "
+            "are optimized, The input list would be [\"X_1\", \"X_2\", gradient of \"X_1\", "
+            "gradient of \"X_2\", accumulated squared gradient of \"X_1\", accumulated squared gradient of \"X_2\"].",
+            "T2",
+            OpSchema::Variadic,
+            false)
+        .Output(0, "T_new", "New update count of inputs", "T3")
```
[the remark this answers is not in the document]
wschin: We need to add all `states` to the output list. `T_new` is added because learning-rate is a function of `T`.

onnx/defs/controlflow/defs.cc
```diff
@@ -774,4 +774,102 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    In that reference paper, this operator is a spacial case of the Figure 1's composite mirror
+    descent update.
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    Adagrad,
+    10,
+    OpSchema()
+        .SetDoc(Adagrad_ver10_doc)
+        .Input(0, "R", "The initial learning rate.", "T1")
```
postrational: Perhaps some of these inputs could be static Attributes instead.

onnx/defs/controlflow/defs.cc
```diff
@@ -774,4 +774,102 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(1, "D", "The decay factor of learning rate after one update.", "T1")
+        .Input(2, "T", "The update count of \"X\". It should be a scalar.", "T3")
+        .Input(3, "Lambda", "Regularization coefficient of 0.5 * Lambda * ||X||_F^2.", "T1")
+        .Input(4, "Eps", "Small scalar to avoid dividing by zero.", "T1")
+        .Input(
+            5,
+            "inputs",
```
wschin: Gradient and accumulated squared gradient should be optional because a model can come without backward graph.

onnx/defs/controlflow/defs.cc
wschin: Will do.

gramalingam:
Hi, Is there a plan to add the complete Adagrad optimization (loop) as another operator? The proposed operator seems only one part of the solution. It would help to understand how the whole thing would work.

onnx/defs/controlflow/defs.cc
wschin: Maybe I am wrong. It causes ambiguation between [X, G] and [X_1, X_2].

wschin:
> Hi, Is there a plan to add the complete Adagrad optimization (loop) as another operator? The proposed operator seems only one part of the solution. It would help to understand how the whole thing would work.

There are two phases of supporting training in ONNX.

1. Training the model with one iteration.
2. Training the model with the whole data set.

Those operators are defined for Phase 1 and will be used in Phase 2 to compose multi-iteration algorithms. Several difficults that we can't go to Phase 2 directly are 

1. Defining a full training algorithm is similar to writing a program.
2. Defining a full training algorithm may put extra constraints on the inference stage. For example, if a written training algorithm wants to run 100 iterations, user will have to feed 100 batches into ONNXRuntime.
3. If we can define one training iteration properly, conceptually users can execute multiple iterations.

onnx/backend/test/case/node/adagrad.py
```diff
@@ -0,0 +1,95 @@
...
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+def apply_adagrad(r, t, x, g, h, norm_coefficient, epsilon, decay_factor):
+    # Compute adjusted learning-rate.
+    r_ = r * (1 + t * decay_factor)
```
wschin:
```suggestion
    r_ = r / (1 + t * decay_factor)
```

onnx/backend/test/case/node/adagrad.py
```diff
@@ -0,0 +1,96 @@
...
+    # Add gradient of regularization term.
+    g_regularized = norm_coefficient * x + g
+    # Update squared accumulated gradient.
+    h_new = h + g_regularized * g_regularized
+    # Compute ADAGRAD's gradient scaling factors
+    h_sqrt = np.sqrt(h_new) + epsilon
+    # Apply ADAGRAD update rule.
+    x_new = x - r_ * g_regularized / h_sqrt
+    return (x_new, h_new)
+
```
wschin:
```suggestion

```

onnx/defs/training/defs.cc
```diff
@@ -324,4 +324,134 @@ ONNX_TRAINING_OPERATOR_SET_SCHEMA(
...
+              size_t i_out = i;
+              propagateElemTypeFromInputToOutput(ctx, i_in, i_out);
+              propagateShapeFromInputToOutput(ctx, i_in, i_out);
+
+              // Pass H1's and H2's shapes to H1_new and H2_new, respectively.
+              i_in = 2 + 2 * num_optimized_tensors + i;
+              i_out = i + num_optimized_tensors;
+              propagateElemTypeFromInputToOutput(ctx, i_in, i_out);
+              propagateShapeFromInputToOutput(ctx, i_in, i_out);
+            }}));
```
postrational: It seems that this operator could be decomposed into more primitive operations. Can we add a `FunctionBody` for it as well?

wschin: Yes... but we need to think about how to process variable-length inputs using a static graph.

onnx/defs/training/defs.cc
```diff
@@ -324,4 +324,134 @@ ONNX_TRAINING_OPERATOR_SET_SCHEMA(
...
+    from the given inputs following the pseudo code below.
+
+    Let "+", "-", "*", and "/" are all element-wise arithmetic operations with
+    numpy-style broadcasting support. The pseudo code to compute those outputs is:
+
+      // Compute a scalar learning-rate factor. If X is never updated, T should be 0.
+      r = R / (1 + T * decay_factor);
+
+      // Add gradient of 0.5 * norm_coefficient * ||X||_2^2, where ||X||_2 is the 2-norm.
+      G_regularized = norm_coefficient * X + G;
```
gramalingam: The comment and pseudo-code above don't seem to match.

onnx/defs/training/defs.cc
```diff
@@ -324,4 +324,134 @@ ONNX_TRAINING_OPERATOR_SET_SCHEMA(
...
+    this operator's input list are sequentially "R", "T", "X", "G", and "H". Other
+    parameters are given as attributes because they are usually constants. Also, the
+    corresponding output tensors are the new value of "X" (called "X_new"), and then
+    the new accumulated squared gradient (called "H_new"). Those outputs are computed
+    from the given inputs following the pseudo code below.
+
+    Let "+", "-", "*", and "/" are all element-wise arithmetic operations with
+    numpy-style broadcasting support. The pseudo code to compute those outputs is:
+
+      // Compute a scalar learning-rate factor. If X is never updated, T should be 0.
```
gramalingam: Above line is unclear. Is the "T" supposed to be "R"? Is the goal to say "If the specified R value is 0, X will never be updated."?

onnx/defs/training/defs.cc
```diff
@@ -324,4 +324,134 @@ ONNX_TRAINING_OPERATOR_SET_SCHEMA(
...
+      r = R / (1 + T * decay_factor);
+
+      // Add gradient of 0.5 * norm_coefficient * ||X||_2^2, where ||X||_2 is the 2-norm.
+      G_regularized = norm_coefficient * X + G;
+
+      // Compute new accumulated squared gradient.
+      H_new = H + G_regularized * G_regularized;
+
+      // Compute the adaptive part of per-coordinate learning rate. Note that Sqrt(...)
+      // compute square root element-wisely.
```
gramalingam: Nit: "compute square root element-wisely" => "computes square-root element-wise"

onnx/defs/training/defs.cc
```diff
@@ -324,4 +324,134 @@ ONNX_TRAINING_OPERATOR_SET_SCHEMA(
...
+    1,
+    OpSchema()
+        .SetDoc(Adagrad_ver1_doc)
+        .Input(0, "R", "The initial learning rate.", "T1")
+        .Input(1, "T", "The update count of \"X\". It should be a scalar.", "T2")
+        .Input(
+            2,
+            "inputs",
+            "It sequentially contains the current values of optimized tensors and then the "
+            "current values of accumulated gradient. For example, if two tensor \"X_1\" and \"X_2\" "
```
gramalingam: "the current values of optimized tensors, followed by their respective gradients, followed by their respective accumulated squared gradients."

onnx/defs/training/defs.cc
```diff
@@ -324,4 +324,134 @@ ONNX_TRAINING_OPERATOR_SET_SCHEMA(
...
+            "are optimized, The input list would be [\"X_1\", \"X_2\", gradient of \"X_1\", "
+            "gradient of \"X_2\", accumulated squared gradient of \"X_1\", accumulated squared gradient of \"X_2\"].",
+            "T3",
+            OpSchema::Variadic,
+            false)
+        .Output(
+            0,
+            "outputs",
+            "It sequentially contains the new values of optimized tensors and then the new "
+            "values of accumulated gradient. For example, if two tensor \"X_1\" and \"X_2\" are "
```
gramalingam: "gradient" => "squared gradient"?

onnx/defs/training/defs.cc
```diff
@@ -324,4 +324,134 @@ ONNX_TRAINING_OPERATOR_SET_SCHEMA(
...
+            "optimized, the output list would be [new value of \"X_1,\" new value of \"X_2\" "
+            "new accumulated squared gradient of \"X_1\", new accumulated squared gradient of \"X_2\"].",
+            "T2",
```
gramalingam: T2 => T3?

onnx/defs/training/defs.cc
gramalingam: Or, say "Specify a T value of 0 if the learning-rate should stay the same."

onnx/defs/training/defs.cc
wschin: It matches. Let `y=0.5 * coef * x^2`. We have `dy/dx = coef * x`.

gramalingam: Sorry, I misunderstood the comment.

onnx/defs/training/defs.cc
wschin: I will change it to
```
      // Compute a scalar learning-rate factor. At the first update of X, T is generally
      // 0 (0-based update index) or 1 (1-based update index).
```

onnx/defs/training/defs.cc
wschin: Yes. Many thanks.

PR #1966, opened 2019-04-25
Optional flag for visual studio to set runtime

ayermolo:
When creating a project through cmake for visual studio BUILD_TYPE can be dropped, this allows visual studio to set it runtime library automatically in VS IDE. It's useful for applications that link against ONNX because it allows to have one solution that supports all the configs.


houseroad:
@snnn @raymondxyang any thoughts?

snnn:
I don't like this change. Because it looks wired, there are two branches in add_msvc_runtime_flag  function, why the new option only applies one of them? If the new toption doesn't work when ONNX_USE_MSVC_STATIC_RUNTIME is ON, at least it should be mentioned at somewhere.


I think
```
if(${ONNX_USE_MSVC_STATIC_RUNTIME})
    if(${CMAKE_BUILD_TYPE} MATCHES "Debug")
      target_compile_options(${lib} PRIVATE /MTd)
    else()
      target_compile_options(${lib} PRIVATE /MT)
    endif()
  else()
    if(${CMAKE_BUILD_TYPE} MATCHES "Debug")
      target_compile_options(${lib} PRIVATE /MDd)
    else()
      target_compile_options(${lib} PRIVATE /MD)
    endif()
  endif()
```
can be shorten to
```
if(${ONNX_USE_MSVC_STATIC_RUNTIME})
    if(${CMAKE_BUILD_TYPE} MATCHES "Debug")
      target_compile_options(${lib} PRIVATE /MTd)
    else()
      target_compile_options(${lib} PRIVATE /MT)
    endif()
endif()
```
Then we don't need the new option.

BTW,  you can omit CMAKE_BUILD_TYPE from cmake command line args only if this project(ONNX) and all its dependencies doesn't use this var in any of its cmake files.  As it is hard to verify, do it with your own risk.

snnn:
Let's ask if add_msvc_runtime_flag function is still needed. If not, then it can be removed, this PR can also be closed.

ayermolo:
Well I wasn't sure what the original purpose of it was, so didn't want to change too much and break someone else build. Removing else part entirely would also works. The if case can still be there to provide option for static linking to anyone who wants it?
@snnn Are you going to make the change, or should I do another PR and just remove else case?

snnn:
@houseroad can we remove the whole function?

ayermolo:
@snnn @houseroad So... what's the conclusion? :)

linkerzhang:
Thank you @ayermolo  !

@zrphercule Rui, any idea about "add_msvc_runtime_flag function is still needed" I noticed that this was added by one of your commit.

ayermolo:
It has been over a week, with no replies from what looks like key stake holders. Can this PR go through, and I can file an issue, tagging this PR, for more general question of is those calls are still necessary?

snnn:
I suggest
1. Either we remove the add_msvc_runtime_flag  function.
2. Or please add a clear comment there, let people know this option only works partially. (I feel it doesn't sounds good).

ayermolo:
I am all for removing this, but people who are tagged about it are not replying, and I don't want to break someone else build. :) As such option 2 I think would suffice for now? I can add a comment.

zrphercule:
@linkerzhang @ayermolo  Sorry for the late response.
Although I created this function, but it was there already when I created it - I was just wrap it to a function. I have checked the code and feel it should be Okay if we remove it.

Let's try first to open a pr removing it, and see what the CI says. While I will double check with @houseroad to make sure it does not break anything.

Thanks!

snnn:
@ayermolo Would you like starting a new PR of removing the add_msvc_runtime_flag function?

ayermolo:
will do

ayermolo:
Tried this: https://github.com/onnx/onnx/pull/1995

Looks like the first part for Mt, Mtd is needed.

snnn:
> When creating a project through cmake for visual studio BUILD_TYPE can be dropped

It can be dropped only if onnx and all its dependencies doesn't use 'CMAKE_BUILD_TYPE' in any of their source file. So far, it's not true.  

I tried to clean it , so I created PR #2022. However, I found it's not easy to solve. 
Let's hold on a little bit, until we can remove 'CMAKE_BUILD_TYPE'.

sutario:
good

PR #1969, opened 2019-04-26
 Fix a shapeinference bug in upsample v9/10

raymondxyang:
Fixed the raw_data handling in upsample shape inference for the input 'scales'


raymondxyang:
Close it right now to fix something.. will reopen soon

onnx/defs/tensor/defs.cc
```diff
@@ -1324,22 +1324,55 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            if (scales->data_type() == TensorProto::FLOAT &&
-                scales->float_data_size() == input_shape.dim_size()) {
-              for (int i = 0; i < input_shape.dim_size(); ++i) {
-                float dim_value =
+            if (scales->data_type() == TensorProto::FLOAT) {
+              bool invalid_scale_shape = false;
+              std::vector<float> vec;
+              if (scales->has_raw_data()) {
+                const auto& data = ParseRawData<float>(scales);
+                vec.insert(vec.end(), data.begin(), data.end());
```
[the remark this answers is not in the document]
raymondxyang: I was following the same way we did in other ops. see https://github.com/onnx/onnx/blob/master/onnx/defs/tensor/defs.cc#L123 and L551.
Any specific reason to make assign() preferred?

onnx/defs/tensor/defs.cc
```diff
@@ -1324,22 +1324,51 @@ ONNX_OPERATOR_SET_SCHEMA(
           auto& input_shape = getInputShape(ctx, 0);
           auto* output_shape = getOutputShape(ctx, 0);
           output_shape->clear_dim();
-          auto scales = ctx.getInputData(1);
+          auto scales = ctx.getInputData(1);
```
xkszltl: Still 1 extra trailing space (ーー;)
Feel free to do it in another PR if you prefer.

onnx/defs/tensor/defs.cc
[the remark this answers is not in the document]
xkszltl: Because the vector here is brand new.
If it's not used afterward, you can just have something like
```
if (scales->has_raw_data()) { 
    const auto& data = ParseRawData<float>(scales); 
    vector<float> vec(data.begin(), data.end());
    ...
```
or if you want it later:
```
vector<float> vec;
if (scales->has_raw_data()) { 
    const auto& data = ParseRawData<float>(scales); 
    vec.assign(data.begin(), data.end());
    ...
```

This is a general suggestion, not something serious.
It's just a shorter way.

onnx/defs/tensor/defs.cc
```diff
@@ -1327,19 +1327,47 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             // Infer output shape's dimension value if 'scales' is known.
-            if (scales->data_type() == TensorProto::FLOAT &&
-                scales->float_data_size() == input_shape.dim_size()) {
-              for (int i = 0; i < input_shape.dim_size(); ++i) {
-                float dim_value =
+            if (scales->data_type() == TensorProto::FLOAT) {
+              bool invalid_scale_shape = false;              
+              if (scales->has_raw_data()) {
+                const auto& data = ParseRawData<float>(scales);
+                std::vector<float> vec(data.begin(), data.end());
```
pranavsharma: This copy can be avoided.

onnx/defs/tensor/defs.cc
```diff
@@ -1383,44 +1411,55 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-                          ").");
-                    }
-                  } else {
-                    output_shape->mutable_dim(i)->set_dim_value(dim_value);
+            if (scales->data_type() == TensorProto::FLOAT) {
+              bool invalid_scale_shape = false;
+              std::vector<float> vec;
+              if (scales->has_raw_data()) {
+                const auto& data = ParseRawData<float>(scales);
+                vec.insert(vec.end(), data.begin(), data.end());
```
pranavsharma: This copy can be avoided.

onnx/defs/tensor/old.cc
```diff
@@ -435,17 +436,41 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             // Infer output shape's dimension value if 'scales' is known.
-            if (scales->data_type() == TensorProto::FLOAT &&
-                scales->float_data_size() == input_shape.dim_size()) {
-              for (int i = 0; i < input_shape.dim_size(); ++i) {
-                float dim_value =
+            if (scales->data_type() == TensorProto::FLOAT) {
+              bool invalid_scale_shape = false;
+              if (scales->has_raw_data()) {
+                const auto& data = ParseRawData<float>(scales);
+                std::vector<float> vec(data.begin(), data.end());
```
pranavsharma: This copy can be avoided.

pranavsharma:
@raymondxyang have you verified this fix with onnxruntime using the model that had raw_data?

raymondxyang:
Yes. verified with both model with raw or float data

PR #1973, opened 2019-04-29
Create archivefileformat doc based on the wiki equivalent

jspisak:
will delete the wiki when done..


houseroad:
We use camel naming method in docs/: https://github.com/onnx/onnx/tree/master/docs

Shall we name the .md file in the same way?

PR #1984, opened 2019-04-30
fix shape inference

askhade:


askhade:
@houseroad : This PR is to fix shape inference for Max Pool and Avg Pool version. Can you please take a look at this PR.

hariharans29:
Some minor nits that I noticed along the way that can potentially be fixed in the same PR -

1) Line 53: `if (input_shape.dim_size() < 2)` should actually be `if (input_shape.dim_size() < 3)` (I think!) since the first two dimensions are `N` and `C`, and the actual data dimensions start from the third dimension

2) Line 60: Comment is stale. `MaxPool` and `Conv` support dilations

3) Line 118: `second_input_shape.dim_size() < 1` can probably be modified to check if it matches the input `shape second_input_shape.dim_size() !=  input_shape.dim_size()`. This check doesn't seem to be present anywhere....

askhade:
> Some minor nits that I noticed along the way that can potentially be fixed in the same PR -
> 
> 1. Line 53: `if (input_shape.dim_size() < 2)` should actually be `if (input_shape.dim_size() < 3)` (I think!) since the first two dimensions are `N` and `C`, and the actual data dimensions start from the third dimension
> 2. Line 60: Comment is stale. `MaxPool` and `Conv` support dilations
> 3. Line 118: `second_input_shape.dim_size() < 1` can probably be modified to check if it matches the input shape `second_input_shape.dim_size() !=  input_shape.dim_size()`. This check doesn't seem to be present anywhere....

For 
1. This shape inference method is common for a lot of ops... Let's figure out how this will affect other operators and test before making this change. 

2. I addressed the comment

3. This is out of scope for MaxPool and AvgPool because these 2 ops only have 1 input. Let's address this as part of a separate PR and add relevant test cases to support these fixes.

PR #1988, opened 2019-05-01
Add shape inference for legacy auto_pad modes

stevenlix:
Add shape inference for legacy auto_pad modes


linkerzhang:
Thank you! Can you also add some test cases to cover this change https://github.com/onnx/onnx/blob/master/onnx/test/shape_inference_test.py please?

onnx/defs/nn/defs.cc
```diff
@@ -1248,6 +1256,44 @@ void convTransposeShapeInference(InferenceContext& ctx) {
     }
   }
 
+  std::vector<int64_t> pads;
+  if (getRepeatedAttribute(ctx, "pads", pads)) {
```
hariharans29: I think this if condition should be (parse the auto_pad attribute prior to this like in line 1266)- 

`(nullptr == auto_pad_attr) || (nullptr != auto_pad_attr && auto_pad_attr->s() == "NOTSET" )`

Probably it is safe to also ensure both `auto_pad` and `pads` attributes are not present in this node - if both are detected fail shape inference with appropriate message

onnx/defs/nn/defs.cc
```diff
@@ -1248,6 +1256,44 @@ void convTransposeShapeInference(InferenceContext& ctx) {
...
+    if (pads.size() != n_input_dims * 2) {
```
hariharans29: According to the spec, it is totally possible for pads size to be less than 2 * input dimensions. In this case, default values to 0. Probably instead of failing shape inference, you could do a resize with values set to 0. This is the line in the spec which makes it possible - 

`If not present, the padding defaults to 0 along start and end of each axis.`

onnx/test/shape_inference_test.py
```diff
@@ -996,6 +996,20 @@ def test_maxpool_with_dilations(self):  # type: () -> None
             [])
         self._assert_inferred(graph, [make_tensor_value_info("Y", TensorProto.FLOAT, (5, 3, 2, 2))])
 
+    def test_maxpool_with_same_upper_padding_and_stride(self):  # type: () -> None
```
hariharans29: Maybe add a few more tests ? 
* NOTSET case
* strides = 1 with either SAME_UPPER or SAME_LOWER (kind of basic, but good to have)

onnx/defs/nn/defs.cc
stevenlix: Since auto_pad is deprecated attribute, "pads" has higher priority than auto_pad and should be checked first. If "pads" and "NOTSET" are not present, it means no padding.

onnx/defs/nn/defs.cc
stevenlix: I checked the spec again. My understanding is that if "pads" presents, its size should be n_input_dims*2. If it doesn't present, its default value is 0, but its size should still be n_input_dims*2.

onnx/test/shape_inference_test.py
stevenlix: NOTSET is covered by old tests already. I added one more test for strides = 1

stevenlix:
> Thank you! Can you also add some test cases to cover this change https://github.com/onnx/onnx/blob/master/onnx/test/shape_inference_test.py please?
Done

PR #1995, opened 2019-05-03
Removing add_msvc_runtime_flag.

ayermolo:
Removing add_msvc_runtime_flag. Not clear it is still necessary, and interfered in creating VS Solutions that support multiple configurations.
Came out of: https://github.com/onnx/onnx/pull/1966


cmake/Utils.cmake
```diff
@@ -1,18 +1,12 @@
 #
-# Add MSVC RunTime Flag
+# Add MSVC RunTime Flag, this part is necessary for tests in CI
```
houseroad: Is it only required by CI? or real world case also need it?

CI should just be one instance of real world case in my opinion.

houseroad:
To be honest, I still didn't see a strong motivation to drop this options...

Could you further elaborate?

linkerzhang:
@houseroad  this PR is based on the discussion of #1966, I believe.

cmake/Utils.cmake
ayermolo: Well I added in case few down the road same question is raised.

ayermolo:
Any way to re-kick CI? Doesn't look like failure is related. "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

ayermolo:
@houseroad great, thanks!

ayermolo:
@zrphercule All good on your side?

ayermolo:
Awesome, thanks!

PR #2002, opened 2019-05-06
Support Unique operator in ONNX

hariharans29:
Support an equivalent of tf.unique in ONNX


onnx/defs/tensor/defs.cc
```diff
@@ -1848,4 +1848,52 @@ ONNX_OPERATOR_SET_SCHEMA(
 
           propagateShapeFromInputToOutput(ctx, 0, 0);
         }));
+
+static const char* Unique_ver11_doc = R"DOC(
+Finds unique elements in an input 1-D tensor sorted in the same order of first occurrence in the input.
```
linkerzhang: not sure it's because of my personal understanding of English. Is it better to say dedup? My first thought of unique of input {2, 1, 1, 3, 4, 3} is {2, 4} :)

hariharans29: This is good feedback. I will try and refine the description - but I was going based on the tf.unique description - 

`This operation returns a tensor y containing all of the unique elements of x sorted in the same order that they occur in x. This operation also returns a tensor idx the same size as x that contains the index of each value of x in the unique output y` from https://www.tensorflow.org/api_docs/python/tf/unique

hariharans29:
@linkerzhang - I refined the op description. Please take a look. It aligns more with tensorflow's description. For more details on model usage, @wenbingl @jiafatom maybe can share more details. 

@jeffsaremi - I included the output you requested. Please review to make sure it if what you need.

Thanks!

jeffsaremi:
I only had one comment. Everything else looked good.

docs/Changelog.md
```diff
@@ -10282,3 +10282,46 @@ This version of the operator has been available since version 10 of the default
...
+<dt><tt>x</tt> : T</dt>
+<dd>A 1-D input tensor that is to be processed.</dd>
+</dl>
+
+#### Outputs
+
+<dl>
+<dt><tt>y</tt> : T</dt>
+<dd>A 1-D tensor of the same type as 'x' containing all the unique values in 'x' sorted in the same order that they occur in the input 'x'</dd>
+<dt><tt>idx</tt> : tensor(int64)</dt>
```
jiafatom: tf.unique output idx supports both int64 and int32, but we only support int64. Shall we need support int32 also?

hariharans29: @linkerzhang - I initially preferred to keep it to just INT64 only, but what's your opinion ?

hariharans29:
Pinging....

@houseroad

docs/Changelog.md
linkerzhang: int64 is good enough, I think.

linkerzhang:
@houseroad  would you help to review this op please? Thank you!

hariharans29:
Closing in favor of #2042

PR #2005, opened 2019-05-07
Fix shape inference logic for TopK operator

hariharans29:
Support initializer containing raw data


onnx/defs/math/defs.cc
```diff
@@ -955,29 +954,50 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          // (2) axis_dim has dim value
+          // Othewise cannot reliably compute output shape as axis dim value is
+          // unknown and hence cannot determine if axis dim value >= k (which
+          // should be enforced)
+          if (nullptr != k && axis_dim.has_dim_value()) {
+            int64_t k_value = 0;
+            if (k->dims_size() != 1 || k->dims(0) != 1)
+              fail_shape_inference(
+                  "K input must be a one-dimensional tensor of size 1.");
+            if (k->has_raw_data() && k->data_type() == TensorProto::INT64)
```
linkerzhang: Thank you!

We may do a little bit refactoring to have a function (template) to get data from TensorProto. That function will cover the raw_data check, otherwise, all shape inference functions (using input data / initializers) will have these two "if"s.

hariharans29: Thanks @linkerzhang. I refactored the function that was previously only parsing raw data to generically handle data parsing (not just raw). So, two "if"s should go away now.

hariharans29:
@raymondxyang (FYI - as the `Upsample` shape inference method has changed due to refactor, could you please review as well ?)

onnx/defs/tensor_proto_util.cc
```diff
@@ -27,45 +27,47 @@ namespace ONNX_NAMESPACE {
...
-  }
+inline bool is_platform_little_endian() {
+  int num = 1;
+  if (*(char*)&num == 1)
+    return true;
+  return false;
+}
 
-#define DEFINE_PARSE_RAW_DATA(type)                                        \
+#define DEFINE_PARSE_DATA(type, typed_data_fetch)                          \
```
raymondxyang: Nice refactor. We can unify the data parsing step after this! Thanks.

btw for testing purpose, did you have the model modified (pass the tensor as initializor instead of input) and tested with the changed code? If you need I can send you the script to do so offline.

hariharans29: Thanks @raymondxyang. Not sure if this will cover the case - but I added a shape inference test for upsample that has an initializer with raw data. Will this help ?

As an aside  - I see that the test seems to be for eopset 9, should I add similar tests for opset 10 ?

hariharans29: Never mind - I see that the op is deprecated in opset 10. Please just check if he test I added will cover your case. Thanks!

raymondxyang: Looks it will cover the case. Thanks!

hariharans29:
@linkerzhang  - When you have a chance, please review again.

PR #2009, opened 2019-05-08
Relax constraint on subgraph input/output type and shape

BowenBao:
In many cases, it is not obvious to know the shape information of inputs/outputs for the subgraph, especially in models where dynamic shape is involved. This PR relaxes this check for subgraph.


onnx/checker.h
```diff
@@ -98,7 +98,8 @@ void check_node(
 void check_graph(
     const GraphProto& graph,
     const CheckerContext&,
-    const LexicalScopeContext&);
+    const LexicalScopeContext&,
+    bool is_main_graph=true);
```
gramalingam: CheckerContext already has the "is_main_graph" attribute … it would be better to set/use that.

BowenBao: Thanks for the comment, update to use CheckerContext.

onnx/checker.cc
```diff
@@ -371,14 +371,17 @@ void check_node(
...
   enforce_non_empty_field(graph, name);
 
-  for (const auto& value_info : graph.input()) {
-    check_value_info(value_info, ctx);
-  }
-  for (const auto& value_info : graph.output()) {
-    check_value_info(value_info, ctx);
+  if (is_main_graph) {
+    for (const auto& value_info : graph.input()) {
+      check_value_info(value_info, ctx);
```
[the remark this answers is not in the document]
BowenBao: Update to use CheckerContext. check_value_info checks name, type, and shape. Now for subgraph the checker will skip type and shape.

onnx/checker.h
```diff
@@ -50,7 +50,7 @@ class CheckerContext final {
   bool is_main_graph() const {
     return is_main_graph_;
   }
-  void set_is_main_graph(bool is_main_graph) {
+  void set_is_main_graph(bool is_main_graph) const {
```
linkerzhang: this looks to me very bad. a "const" setter....

gramalingam: Yes, I agree. I think we should be creating a copy of the CheckerContext to modify it. Furthermore, we can make all components of CheckerContext that do not change "const *" or "const &" so it can be shared (if desired).

gramalingam: Or, it should be a non-const parameter.

BowenBao: Agree.. Let's keep the ctx immutable, creating copy for subgraph.

onnx/test/checker_test.py
```diff
@@ -217,6 +217,39 @@ def test_nested_graph(self):  # type: () -> None
...
+        )
+
+        i1 = helper.make_node(
+            "If", ["cond"], ["Z"], then_branch=graph, else_branch=graph)
+
+        graph = helper.make_graph(
+            [i1],
+            "test",
+            inputs=[
+                helper.make_tensor_value_info("cond", TensorProto.BOOL, [1])
```
houseroad: Define X?

onnx/test/checker_test.py
```diff
@@ -217,6 +217,39 @@ def test_nested_graph(self):  # type: () -> None
...
+            ],
+            outputs=[],
```
houseroad: Define the output?

onnx/test/checker_test.py
BowenBao: Added, also added to the original nested_graph test case that I copied from.

onnx/test/checker_test.py
BowenBao: same above

BowenBao:
cc @ebarsoum @spandantiwari

PR #2013, opened 2019-05-11
ONNX Training Proposal

wschin:
**PR #2314 is a single place for reviewing the whole training story.**

This proposal aims at capturing the information required to perform
stochastic gradient-based training on the inference graph.

Although training itself can encoded as a sub-graph by adding
a single backward operator, we currently prefer a more strongly-typed
way to store that information. We also think separating training graph
from inference graph can make backends' life easier.

The major change happens in the introduction of TrainingInfoProto:
```protobuf
// Training information
// TrainingInfoProto contains the minimized loss function (e.g., mean
// squared error), the used optimization algorithm (a mapping from gradient
// and the current tensors to their new values; that is, one training iteration),
// and the binding between the optimized tensors and their gradient tensors'
// names (we need to give a name (e.g., "gradient_of_X") to the gradient of
// tensor "X", so that gradient tensors can be referenced in the training 
// algorithm).
//
// This design is mainly designed for popular stochastic gradient-based methods.
// Conceptually, a TrainingInfoProto defines a "partial computation graph"
// which will be merged into the inference graph to form a full training
// computation graph.
message TrainingInfoProto {
  // Training-specific initializers, for example, accumulated squared gradient
  // in ADAGRAD optimizer, momentum, or number of executed training iterations.
  // The values can only be inputs of "loss" and "optimizer" defined below.
  repeated TensorProto additional_initializer = 1;

  // Training-specific inputs, for example, labels in classification problems.
  // The values can only be inputs of "loss" and "optimizer" defined below.
  // The full input list of the training computation graph is the concatenation
  // of "ModelProto.graph.input" and
  // "ModelProto.training_info.additional_input".
  repeated ValueInfoProto additional_input = 2;

  // Training-specific outputs, for example, new tensor values computed by the
  // specified training algorithm. The full output list of the training
  // computation graph is the concatenation of "ModelProto.graph.output" and
  // "ModelProto.training_info.additional_output".
  repeated ValueInfoProto additional_output = 3;
  
  // Loss node which consumes tensors in "ModelProto.graph", "additional_inputs"
  // , and "additional_initializer". Loss nodes produces a scalar loss value.
  // In training phase, the optimization algorithm would be launch to minimizes
  // loss node's output value. "loss" can be an ONNX primitive operator or a
  // function defined in "ModelProto.function."
  //
  // The field MUST be present.
  optional NodeProto loss = 4;

  // Optimized tensors and their gradient tensor names. Each pair describes a
  // tensor's name (key) and its gradient's name (value). The gradient tensors
  // are only visible to the input list of "optimizer."
  repeated StringStringEntryProto gradient_binding = 5;

  // Optimization algorithm used to minimize the loss function. It can be an
  // ONNX primitive operator or a function defined in "ModelProto.function."
  // This node can reference tensors in the "ModelProto.graph",
  // "additional_inputs," "additional_initializer," and gradient tensors
  // defined by "gradient_bing" below. The field "optimizer"'s outputs are
  // accessible only in "additional_training_output."
  //
  // The field MUST be present.
  optional NodeProto optimizer = 6;

  // Gradient-based training is usually an iterative procedure. In one gradient
  // descent iteration, we apply
  //
  // x = x - r * g
  //
  // where "x" is the optimized tensor, "r" stands for learning rate, and "g" is
  // gradient of "x" with respect to "loss" node's output value. To avoid the
  // direct use of an assignment, user should explicitly specify the new
  // tensor's name for each optimized tensor. For example, if the new value of
  // "x" is "y," the field "update_binding" may contain a pair of "x" (key) and
  // "y" (value).
  //
  // Similarly, optimizer's state tensors (e.g., accumulated squared gradient
  // and momentum) should be handled by using the same strategy. 
  //
  // Each StringStringEntryProto binds an initizliaer defined in one of
  // "ModelProto.graph.initializer" and "additional_training_initializer" to
  // an element of "additional_training_output." An initializer can be bound
  // to at most one training output.
  repeated StringStringEntryProto update_binding = 7;
}
```
An optional `TrainingInfoProto` will be added into ModelProto so that users know how to apply the specified optimization algorithm to conduct further training iterations.

To allow customized gradient-based optimization algorithm, a `FunctionProto` list is also added into `ModelProto`. User can store their training algorithm as a `FunctionProto` in that list and reference that `FunctionProto` using `TrainingInfoProto.optimizer`.

The existence of `TrainingInfoProto.additional_initializer` has a reason. There are many per-tensor states in training phase. Common examples are momentum and accumulated squared gradient. If we reuse `ModelProto.graph.initializer` to store those training-specific tensors, loading a model for inference may cause a lot more memory (2 times or 3 times).

Common optimizers and loss functions will be proposed `FunctionProto`s subsequently. For example, ADAGRAD optimizer has a WIP PR, https://github.com/onnx/onnx/pull/1955.


onnx/onnx-ml.proto
```diff
@@ -188,6 +194,115 @@ message NodeProto {
...
+  // computation graph is the concatenation of "ModelProto.graph.output" and
+  // "ModelProto.training_info.additional_output".
+  repeated ValueInfoProto additional_output = 3;
+  
+  // Loss node which consumes tensors in "ModelProto.graph", "additional_inputs"
+  // , and "additional_initializer". Loss nodes produces a scalar loss value.
+  // In training phase, the optimization algorithm would be launch to minimizes
+  // loss node's output value. "loss" can be an ONNX primitive operator or a
+  // function defined in "ModelProto.function."
+  required NodeProto loss = 4;
```
gramalingam: If I remember right, we make all fields optional for Proto2 / Proto3 compatibility? And document required fields in the comment only (e.g., see the other fields with the "The <name> field MUST be present" comments.)

onnx/onnx-ml.proto
```diff
@@ -188,6 +194,115 @@ message NodeProto {
...
+  
+  // Optimization algorithm used to minimize the loss function. It can be an
+  // ONNX primitive operator or a function defined in "ModelProto.function."
+  // This node can reference tensors in the "ModelProto.graph",
+  // "additional_inputs," "additional_initializer," and gradient tensor names
+  // defined below.
+  // To connect the optimizer's parameter, optimized tensors, optimized tensors'
+  // gradient, and optimizer correctly without breaking
+  // single-static-assignment, "optimizer" has some restruction on its input
```
gramalingam: I don't think we need the restrictions below now, do we? If we have pre-defined ONNX ops (for the optimizer), the op spec will indicate the required operand-ordering. We can recommend the following ordering as a preferred practice, but it is  not clear to me why this is a compulsory requirement. (Also: spelling correction "restruction" => "restriction" … but I don't think we need restrictions.)

onnx/onnx-ml.proto
```diff
@@ -188,6 +194,115 @@ message NodeProto {
...
+  //      "new_accumulate_grad_X1", "new_accumulate_grad_X2"].
+  //   3. Optimized tensors and optimizer states must be defined in
+  //      "ModelProto.graph.initializer" or
+  //      "ModelProto.training_info.additional_initializer."
+  required NodeProto optimizer = 5;
+
+  // Optimized tensors and their gradient tensor names. Each pair describes a
+  // tensor's name (key) and its gradient's name (value). The gradient tensors
+  // are only visible to the input list of "optimizer."
+  repeated StringStringEntryProto gradient_bindings = 6;
```
gramalingam: Just a minor suggestion: can we swap the order of the optimizer and gradient_bindings? Just to correspond to the order of evaluation (cost function first, then computation of gradients, and then the optimizer step).

gramalingam:
Thanks Wei-sheng. Looks good to me, with a few minor points I mentioned above.

onnx/onnx-ml.proto
wschin: They are for customized training algorithms. As you mentioned, pre-defined ONNX optimizers have some kind of special schemas for correcting wiring tensors. For the same reason, customized optimizers (defined as FunctionProto in `ModelProto.function`) should use the same schema.

onnx/onnx-ml.proto
wschin: I am ok to change them back to `optional` but is it really meaningful to have a `TrainingInfo` without loss/optimizer?

onnx/onnx-ml.proto
wschin: Sure.

onnx/onnx-ml.proto
gramalingam: Yes, but the runtime doesn't care really since we already capture the relationship between optimized-tensors and their-gradients separately, and the optimizer NodeProto's input list captures the relationship between the tensors/gradients computed in the preceding steps and the inputs required by the optimizer custom-function. So, I don't understand what extra-restrictions we need to talk about. (There is already an implicit requirement that the inputs listed in the optimizer NodeProto's input should be consistent with the input-ordering in the optimizer custom-function/onnx-op.)

gramalingam: In other words, there is no requirement that different customized optimization functions should all follow the same pattern/style for ordering inputs/outputs. However, if they all follows the same pattern/style, it may end up being helpful in the future (or to users). So, I can understand the value in a "recommended ordering" (but it doesn't seem like it needs to be a "mandated ordering").

wschin: Let's consider an example. Can user specify an optimization function which outputs `[Gradient_X1, new_X1, new_X2, Gradient_X2]`? How would runtime know "new_X1" is the new value of input tensor "X"? In addition, there can be other FunctionProto optimizers which produce `[A, B, C, D]`, `[C, A, B, D]`, and so on. Can runtime bind `[A, B, C, D]` correctly to updated tensors and updated states without those restrictions?

gramalingam: new_X1 would be in the additional_input list (of TrainingInfoProto), right? It will also appear in the optimizer NodeProto's input list. (The relationship between items in these lists are simply by name.) A corresponding formal_parameter (call it formal_new_X1) will appear in the customized FunctionProto. The correspondence between items in the Optimizer NodeProto input list and the customized FunctionProto input list is by position. So, what else does the runtime need to know?

gramalingam: Sorry, new_X1 would probably in the inference graph's input list. Other things (like state_X1) may be in the additional_input list. It doesn't matter. But they are referred to by name in the optimizer NodeProto's input list.

gramalingam: Sorry again (several typos!): I am talking about inputs, but the same thing applies for outputs. new_X1 in optimizer NodeProto's output list is associated with the additional_outputs of training ny name.

gramalingam: Spoke with Wei-sheng offline. Now I understand his goals. It sounds like this is more a schema restriction for the training-graph (rather than the optimizer function). Can we specify the restriction that the additional_training_outputs must correspond to the concat(graph initializers, training graph additional initializers) one-to-one?

gramalingam: Otherwise, we may need to add the mapping as another field.

wschin: One-to-one looks too much because not every tensor is updated in the training phase. Let's add another string-to-string entry list.

onnx/onnx.proto
```diff
@@ -186,6 +192,123 @@ message NodeProto {
...
+  repeated ValueInfoProto additional_output = 3;
+  
+  // Loss node which consumes tensors in "ModelProto.graph", "additional_inputs"
+  // , and "additional_initializer". Loss nodes produces a scalar loss value.
+  // In training phase, the optimization algorithm would be launch to minimizes
+  // loss node's output value. "loss" can be an ONNX primitive operator or a
+  // function defined in "ModelProto.function."
+  //
+  // The field MUST be present.
+  optional NodeProto loss = 4;
```
wschin: Please clarify name look-up when referencing `FunctionProto` in `ModelProto.functions`.

onnx/onnx.proto
```diff
@@ -186,6 +192,123 @@ message NodeProto {
...
+
+  // Optimized tensors and their gradient tensor names. Each pair describes a
+  // tensor's name (key) and its gradient's name (value). The gradient tensors
+  // are only visible to the input list of "optimizer."
+  repeated StringStringEntryProto gradient_binding = 5;
```
wschin: Please explain how the gradient tensors are associated with `loss`.

chinhuang007: Do we really need this binding? Based on the document https://github.com/onnx/onnx/files/3208156/ONNX.Training.Discussion.pptx, the optimizer has W and Gradient W as input. The name of "gradient W", which is output from the backward prop, seems could be derived from the backend and runtime implementation. The current Pytorch and TF APIs do not support user-defined names for the gradients for the optimizers as a reference.

wschin: We need it. Pytorch creates such a binding in another (but equivalent) way. They have a dictionary where key is parameter and value is its gradient tensor.

hobei:
@wschin During reviewing of the TrainingInfoProto structure I find that the TrainingInfoProto in the onnx.in.proto has a 'algorithm' field but the TrainingInfoProto in onnx.proto (and others) have a 'loss' & 'optimizer' field. Are they intentionally different? Thanks

chinhuang007:
The cardinality of TrainingInfoProto in model seems out of sync in different proto files. I believe the latest proposal is to have many (repeated) to support multiple training stages?

wschin:
@hobei, @chinhuang007, That's my bad. `onnx.in.proto` contains the right information.

wschin:
This PR has been merged into #2314 so we can close it now.

Issue #2008, opened 2019-05-07
Adding Sparse Tensor support in ONNX


gramalingam:
In a previous ONNX workshop, there was agreement to add SparseTensor support to ONNX. Please find a design document for this attached:
[SparseDesign.docx](https://github.com/onnx/onnx/files/3154766/SparseDesign.docx)

gramalingam:
Adding updated design document:
[SparseDesign-V2.docx](https://github.com/onnx/onnx/files/3347252/SparseDesign-V2.docx)

jxchenus:
I found that a GraphProto loses its sparse tensor initializer going through the optimizer:

#0 onnx::graphProtoToGraph (gp=..., nested=nested@entry=false)
at .../onnx/common/ir_pb_converter.cc:188
#1 0x0000000000568d3f in onnx::ImportModelProto (mp=...)
at .../onnx/common/ir_pb_converter.cc:338
#2 0x000000000054e39a in onnx::optimization::Optimizer::optimize (
this=this@entry=0x7fffffffcf00, mp_in=...)
at .../onnx/optimizer/optimize.h:26
#3 0x0000000000534d3c in onnx::optimization::Optimize (mp_in=..., names=...)
at .../onnx/optimizer/optimize.cc:32

Is this a known issue?

jxchenus:
Filed a bug with more information for the above issue: https://github.com/onnx/onnx/issues/3009

garymm:
Is this done? https://github.com/onnx/onnx/pull/2019 is merged.

lokitoth:
@garymm Based on the design-doc, it seems like there are still more operators to be enlightened to sparse support (Add/Mul, etc.). 

@gramalingam: Will there be more work here to support additional operators?

askhade:
SparseTensor was added to ONNX's type system starting this PR #3398 

I think we can close this issue now. In future will most likely need to extend more operators to support SparseTensor type but that can be tracked by a separate issue and it should be done on a case by case basis driven by models. 

@gramalingam  thoughts?

PR #2019, opened 2019-05-14
Adding sparse tensor to ONNX

gramalingam:
Proposal to add sparse-tensor support in ONNX proto. (Please see https://github.com/onnx/onnx/issues/2008  and design document https://github.com/onnx/onnx/files/3154766/SparseDesign.docx )

The original proposal has been split into two parts. The current PR addresses only the addition of sparse-tensor constants into a model file. The type system extension is not part of the current proposal. This PR introduces SparseTensorProto to represent tensors in a sparse-format in a model, and its use in AttributeProto and initializers. It extends the Constant op to create tensors from this sparse representation.


onnx/onnx.in.proto
```diff
@@ -442,6 +447,21 @@ message TensorProto {
...
+// A serialized sparse-tensor value
+message SparseTensorProto {
+  // The sequence of non-default values are encoded as a tensor of shape [NNZ].
+  // The default-value is zero for numeric tensors, and empty-string for string tensors.
+  optional TensorProto values = 1;
+
+  // The indices of the non-default values are encoded as a tensor of shape [NNZ].
+  // Note that the indices are linearized-indices. Each linearized index can be
+  // converted into a tuple of indices using the shape provided below.
+  optional TensorProto indices = 2;
```
gramalingam: Note that we could generalize this and allow the user to choose to encode the indices in linearized form or directly: e.g., consider a 100x100 sparse matrix. A direct representation would be [ [row-indices], [column-indices] ] e.g., [ [0,1,2], [3,4,5] ] while a linearized form would be more compact as [ 0x100+3, 1x100+4, 2x100+5 ] = [3, 104, 205].

onnx/onnx.in.proto
```diff
@@ -442,6 +447,21 @@ message TensorProto {
   repeated uint64 uint64_data = 11 [packed = true];
 }
 
+// A serialized sparse-tensor value
+message SparseTensorProto {
+  // The sequence of non-default values are encoded as a tensor of shape [NNZ].
+  // The default-value is zero for numeric tensors, and empty-string for string tensors.
+  optional TensorProto values = 1;
```
gramalingam: Should we generalize this and allow values to be a tensor of rank more than 1? That would be useful if we want to capture sparse-tensors with s sparse-dimensions followed by d dense-dimensions where d > 1.

onnx/checker.cc
```diff
@@ -221,6 +225,31 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
 }
 
+void check_sparse_tensor(
+    const SparseTensorProto& sparse,
+    const CheckerContext& ctx) {
+  int64_t nnz = 0;
+  if (sparse.has_values()) {
+    const TensorProto& values = sparse.values();
+    check_tensor(values, ctx);
+    // TBD: Should we generalize and permit tensors of rank > 1?
```
wschin: I'd like to have higher rank. Examples:
- In multi-class linear classification, we usually store the linear coefficient vectors for all classes as a 2-D tensor.
- In some training algorithms (see truncated Newton methods as an example), a lot of sparse-matrix-vector produces are needed.
- The output of OneHotEncoder/TFIDF should be a sparse tensor and that sparse tensor's rank can be larger than 1.

onnx/checker.cc
```diff
@@ -328,14 +366,22 @@ void check_node(
   }
 
   // Put the removed experimental ops here
-  static std::set<std::string> experimental_ops = {"ATen", "Affine",
-    "ConstantFill", "Crop", "DynamicSlice", "GRUUnit", "GivenTensorFill",
-    "ImageScaler", "ParametricSoftplus", "Scale", "ScaledTanh"};
+  static std::set<std::string> experimental_ops = {"ATen",
```
wschin: Do we still have experimental ops?

onnx/onnx.in.proto
wschin: Yes, then we need to have checker being aware of this.

onnx/onnx.in.proto
wschin: I feel yes for reasons mentioned above.

onnx/onnx.in.proto
gramalingam: In linearized case, the rank is 1 and shape is [NNZ]. In the other case, the rank is 2, and the shape is either [NNZ, dims] or [dims, NNZ] (we need to pick one of these). Note that the values are stored separately (as they may have a different type) from the indices.

onnx/checker.cc
gramalingam: Hi, this question is not about the rank of the sparse-tensor: we do permit sparse-tensors of arbitrary rank. The question is about whether the value stored for a given sparse-index can be, say, N floats instead of a single float. Pytorch, for example, allows this. It allows us to take a d-dimensional tensor, and say that the first s-dimensions are sparse and the remaining (d-s) dimensions are dense. In this case, the values field becomes a (d-s)-dimensional tensor, while the indices becomes a [NNZ, s] dimensional tensor. Representation-wise, this is easy enough to support. However, the implementation of ops will be somewhat more complex to support this generality (vs. treating each value as a scalar).

onnx/checker.cc
gramalingam: I didn't touch this. It was reformatted automatically by VisualStudio using the clang format. (I believe this might be used for backward-compatibility … Ke would know.)

onnx/checker.cc
wschin: I got your point. This might be helpful for scientific simulations where their matrices can have block-diagonal structure. However, I don't feel they will be ONNX's major users at this stage. Let's not to allow the mixture of sparse and dense for now (as your design can be extended to cover that in the future).

onnx/checker.cc
wschin: Ah.. For BC, yes, we need to know those operator names.

onnx/onnx.in.proto
wschin: I'd prefer `[NNZ, dims]` for more compact representation in extremely sparse cases.

wschin: I'd prefer `[NNZ, dims]`. In general, sparse matrices are used to store `data points` (a collection of sparse feature vectors). I'd imagine that most data transformations would want to access the coordinates of one `non-zero` at the same time.

gramalingam: Yes, conceptually [NNZ, dims] is more intuitive. I think TF does this. However, the MKL sparse library, for example, requires row-indices as a separate array and column-indices as a separate array (see: https://software.intel.com/en-us/mkl-developer-reference-c-sparse-blas-coordinate-matrix-storage-format ), which can be generated easily if we use [dims, NNZ]. So, it doesn't look like there is one uniform standard anyway.

gramalingam: PyTorch sparse constructors (at Python level) seem to take a list of lists in [dims,NNZ] ordering: see: https://pytorch.org/docs/stable/sparse.html

onnx/onnx.proto
```diff
@@ -443,6 +448,21 @@ message TensorProto {
...
+// A serialized sparse-tensor value
+message SparseTensorProto {
+  // The sequence of non-default values are encoded as a tensor of shape [NNZ].
+  // The default-value is zero for numeric tensors, and empty-string for string tensors.
+  optional TensorProto values = 1;
+
+  // The indices of the non-default values are encoded as a tensor of shape [NNZ].
+  // Note that the indices are linearized-indices. Each linearized index can be
+  // converted into a tuple of indices using the shape provided below.
+  optional TensorProto indices = 2;
```
linkerzhang: this should be just a repeated int64 indices?

onnx/onnx.in.proto
wschin: Let's only consider 2-D tensors for simplicity. Right, there is no uniform standard, but having `[NNZ, dims]` is enough to cover all cases. As COO format is not efficient for matrix computation, runtime will have to do a conversion (to, e.g., CSR) by sorting those non-zeros according to either row indexes and column indexes. Because the sorting will be something like `sorting those 1-D index vectors by their first/second value`, I feel using `[NNZ, dims]` may lead to better cache locality.

On the other hand, when producing sparse tensors to COO format using `OneHotEncoder`, we generally produce `[dims]` index vector at one time, so keep appending `[dims]` to `[NNZ, dims]`-output could also lead to a better cache locality.

onnx/onnx.proto
wschin: No. For a 2-D sparse tensor,
```
[ [0, 0],
  [1, 0] ],
```
the first element in `indices` would be `[1, 0]`.

onnx/onnx.in.proto
wschin: Another behavior I'd like to have is that `indices` should be sorted first by its first element, then by its second element, .., and finally its last element. This is very helpful when searching for a specified value in a sparse tensor.

onnx/onnx.proto
gramalingam: @linkerzhang : it is a int64 TensorProto, if that is your point. Using a TensorProto instead of "repeated int64" has a couple of advantages: one, as Wei-sheng mentions, it allows us to capture the "shape" of indices; second, it allows us to use the different formats supported by TensorProto for free.

onnx/onnx.in.proto
gramalingam: Yes, we should require that the indices should appear in sorted order. We need to clarify that in the documentation.

onnx/checker.cc
```diff
@@ -221,6 +225,31 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+void check_sparse_tensor(
+    const SparseTensorProto& sparse,
+    const CheckerContext& ctx) {
+  int64_t nnz = 0;
+  if (sparse.has_values()) {
+    const TensorProto& values = sparse.values();
+    check_tensor(values, ctx);
+    // TBD: Should we generalize and permit tensors of rank > 1?
+    if (values.dims().size() != 1)
+      fail_check("Sparse tensor values must have rank 1.");
```
wschin: Please print out the tensor's name if possible.

onnx/checker.cc
```diff
@@ -221,6 +225,31 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+    nnz = values.dims(0);
+  }
+  int64_t num_indices = 0;
+  if (sparse.has_indices()) {
+    const TensorProto& indices = sparse.indices();
+    check_tensor(indices, ctx);
+    // TBD: Should we generalize and permit tensors of rank > 1?
+    if (indices.dims().size() != 1)
+      fail_check("Sparse tensor indices must have rank 1.");
```
wschin: Please print out the tensor's name if possible.

onnx/checker.cc
```diff
@@ -221,6 +225,31 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+    nnz = indices.dims(0);
+  }
+  if (nnz != num_indices)
+    fail_check("Number of values and indices must be same in a sparse tensor.");
```
wschin: Please print out the tensor's name if possible.

onnx/checker.cc
```diff
@@ -221,6 +225,31 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+  if (sparse.has_values()) {
+    const TensorProto& values = sparse.values();
+    check_tensor(values, ctx);
+    // TBD: Should we generalize and permit tensors of rank > 1?
+    if (values.dims().size() != 1)
+      fail_check("Sparse tensor values must have rank 1.");
+    nnz = values.dims(0);
+  }
+  int64_t num_indices = 0;
+  if (sparse.has_indices()) {
```
wschin: Should we throw if sparse tensor has no indices? Also, would you think checking if those indices are properly bound by `dims`  is one thing we need to do here?

onnx/onnx.proto
linkerzhang: Fair enough.

gramalingam: On second thoughts, I think it seems simpler to go with @linkerzhang 's suggestion of repeated int64 indices. (Of course, it will contain NNZ * rank entries, in the required order, which the spec should clarify.) It is unclear we will need the extra flexibility provided by the different formats in a TensorProto for a sparse tensor. So, I am inclined to change it as suggested.

gramalingam:
Do we need a separate name for the sparse-tensor? Or, is the name in the "values" tensor (which MUST be present) sufficient?

wschin:
> Do we need a separate name for the sparse-tensor? Or, is the name in the "values" tensor (which MUST be present) sufficient?

I think `values` tensor is good enough but we still need to explicitly define this in the spec.

onnx/onnx.in.proto
```diff
@@ -145,6 +147,9 @@ message AttributeProto {
   repeated bytes strings = 9;         // list of UTF-8 strings
   repeated TensorProto tensors = 10;  // list of tensors
   repeated GraphProto graphs = 11;    // list of graph
+
+  optional SparseTensorProto sparse_tensor = 22;  // sparse tensor value
```
wschin: It seems that this message groups fields by `repeated` and `optional`. Do you feel this line needs to go to line 144?

onnx/onnx.proto
gramalingam: After more discussions, we decided to leave it as a TensorProto, given that protobuf repeated fields seem to use int rather than int64 for indexing. TensorProto's flexible options for representing data may end up being useful.

onnx/checker.cc
gramalingam: Added the checks.

gramalingam:
Hi @wschin , thanks for all your feedback. I believe I have addressed them all. If there's anything more, please do let me know.

gramalingam:
@houseroad : I added the helper make_sparse_tensor. However, I have a concern about the python support in protobuf, which makes these kind of utility functions potentially expensive and inefficient. Protobuf, in python, does not let us assign or update field values (except for primitive types). Thus, we end up deep copying in such utility functions, which could be expensive for large tensors. For example, if we create a tensor (using make_tensor) and then create an attribute (using make_attribute), we will likely end up copying the data first in make_tensor (e.g., from a numpy array) and then copying it again in make_attribute. I don't see a easy solution to this. If the user directly does it (without using these utility functions) they can probably eliminate the extra copy. This may not be a big deal for small objects, but could matter for tensors that are 100s of megabytes.

gramalingam:
Hi @linkerzhang : this PR is ready. Please let me know if it is okay to merge it in.

gramalingam:
I can merge in master once we have approval.

onnx/checker.cc
```diff
@@ -226,6 +228,160 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+    size_t nnz) {
+  int dense_rank = sparse_tensor_proto.dims_size();
+  int64_t dense_size = 1;
+  for (int i = 0; i < dense_rank; ++i)
+    dense_size *= sparse_tensor_proto.dims(i);
+  if (static_cast<size_t>(indices.dims(0)) != nnz)
+    fail_check(
+        "Sparse tensor indices (",
+        indices.name(),
+        ") does not have NNZ values.");
```
wschin:
```suggestion
        ") does not have the right number of values.");
```
Please also print out the right value, `nnz`, and the existing value, `indices.dims(0)`.

onnx/checker.cc
```diff
@@ -226,6 +228,160 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+
+  // Check if indices appear in ascending order, and if they have valid
+  // values.
```
wschin:
```suggestion
  // values. The i-th value in index_data is the linear index of the i-th non-zero value.
```

onnx/checker.cc
```diff
@@ -226,6 +228,160 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+  int64_t prev_index = -1;
+  for (size_t i = 0; i < nnz; ++i) {
+    int64_t curr_index = index_data[i]; // linearized index of i-th value
+    if (curr_index < 0 || curr_index >= dense_size)
+      fail_check(
+          "Sparse tensor (",
+          indices.name(),
+          ") index value at position [",
+          i,
+          "] out of range.");
```
wschin: Please print out the expected range, `[0, dense_size]`, if possible.

onnx/checker.cc
```diff
@@ -226,6 +228,160 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+          i,
+          "] not in sorted order.");
+    }
+    prev_index = curr_index;
+  }
+}
+
+// Check that the index data stored in a SparseTensorProto is valid.
+// indices: a 2-dimensional tensor; indices[i,j] represents the j-th
+// index value for the i-th nonzero value.
```
wschin: What does it mean? An non-zero value can have multiple indices?

onnx/onnx-ml.proto3
```diff
@@ -445,6 +458,28 @@ message TensorProto {
...
 }
 
+// A serialized sparse-tensor value
+message SparseTensorProto {
+  // The sequence of non-default values are encoded as a tensor of shape [NNZ].
+  // The default-value is zero for numeric tensors, and empty-string for string tensors.
+  TensorProto values = 1;
+
+  // The indices of the non-default values, which may be stored in one of two formats.
+  // (a) Indices can be a tensor of shape [NNZ, rank] with the [i,j]-th value
```
wschin: Is it for sparse block matrix?

onnx/onnx-ml.proto3
```diff
@@ -445,6 +458,28 @@ message TensorProto {
   repeated uint64 uint64_data = 11 [packed = true];
 }
 
+// A serialized sparse-tensor value
+message SparseTensorProto {
+  // The sequence of non-default values are encoded as a tensor of shape [NNZ].
```
wschin: If user uses indices following case (a), would this shape still be [NNZ]?

onnx/onnx-ml.proto3
```diff
@@ -445,6 +458,28 @@ message TensorProto {
...
+  // The default-value is zero for numeric tensors, and empty-string for string tensors.
+  TensorProto values = 1;
+
+  // The indices of the non-default values, which may be stored in one of two formats.
+  // (a) Indices can be a tensor of shape [NNZ, rank] with the [i,j]-th value
+  // corresponding to the j-th index of the i-th value (in the values tensor).
```
wschin:
```suggestion
  // corresponding to the j-th index of the i-th value (in the values tensor).
  // That is, the linear index of values[i][j] is indices[i][j].
```

onnx/onnx-ml.proto
```diff
@@ -477,6 +512,14 @@ message TypeProto {
     optional TensorShapeProto shape = 2;
   }
 
+  message SparseTensor {
```
wschin: [nit] Not sure if we need test for this.

onnx/checker.cc
```diff
@@ -226,6 +228,160 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+  }
+  for (int i = 0; i < dense_rank; ++i) {
+    if (sparse_tensor_proto.dims(i) <= 0)
+      fail_check(
+          "Sparse tensor (", values.name(), ") dimensions are not positive.");
+  }
+
+  if (sparse_tensor_proto.has_indices()) {
+    // We keep the representation simple for now.
+    // indices must be a tensor of shape [NNZ, dense_rank].
```
wschin: `Must`? I saw two formats below. One uses linearized indexing and the other is the same as what described here.
```

File: onnx/checker.cc [hunk 20]
```diff
@@ -226,6 +228,160 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+    // integer value (the linearized index), the value the index would
+    // have if the tensor is reshaped to be 1-dimensional. This is more
+    // compact. We may add support for this later if required.
+    const TensorProto& indices = sparse_tensor_proto.indices();
+    check_tensor(indices, ctx);
+    if (indices.data_type() != TensorProto::INT64)
+      fail_check(
+          "Sparse tensor indices (", indices.name(), ") must have INT64 type.");
+    switch (indices.dims().size()) {
+      case 1:
```
wschin:
```suggestion
      // Check linearized index.
      case 1:
```

onnx/checker.cc
```diff
@@ -226,6 +228,160 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+        check_sparse_tensor_indices_1(indices, sparse_tensor_proto, nnz);
+        return;
+      case 2:
```
wschin:
```suggestion
      // Check COO-style index. Note that, for example, in 3-D tensor, an index means a 3-element vector.
      case 2:
```

onnx/checker.cc
gramalingam: Thanks. I was puzzled about the original message … I guess I meant to write "does not have ", nnz, " values" !

gramalingam:
Close and reopen to trigger CI.

PR #2022, opened 2019-05-16
Cleanup cmake

snnn:
There are many ways of setting compiler flags globally.

See: https://gitlab.kitware.com/cmake/cmake/issues/18390

It doesn't need any special code in the project itself.
Let's keep it simple.


snnn:
The problem is:  
In our Windows CI build, onnx depends on the prebuilt libprotobuf from conda forge.
However, the older version of libprotobuf  was static link to VC Runtime,
But the latest one is dynamic link to VC Runtime.
And, currently, onnx's cmake can only support one of these!

Any suggestion?

PR #2023, opened 2019-05-16
Bug fix in Conv Shape Inference

hariharans29:
Support cases where stride (value) > kernel (value). Needed in ResNet50 case.

### DIFF
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -101,20 +101,12 @@ void convPoolShapeInference(
     if ((nullptr != auto_pad_attr) && (auto_pad_attr->s() != "VALID")) {
       int input_dims_size = static_cast<int>(n_input_dims);
       for (int i = 0; i < input_dims_size; ++i) {
-        int64_t residual = 0;
-        if (strides[i] > 1) {
-          if (!input_shape.dim(2 + i).has_dim_value()) {
-            continue;
-          }
-          residual =  input_shape.dim(2 + i).dim_value();
-          while (residual > 0) {
-            residual -= strides[i];
-          }
+        if (!input_shape.dim(2 + i).has_dim_value()) {
+          continue;
         }
-        int64_t total_pad = residual == 0 ? kernel_shape[i] - strides[i] : kernel_shape[i] + residual;
-        if (total_pad < 0) {
-          fail_shape_inference("Stride is bigger than Kernel shape");
-        }
+        int64_t dim_value = input_shape.dim(2 + i).dim_value();
+        int64_t legacy_target_size = (dim_value + strides[i] - 1) / strides[i];
+        int64_t total_pad = (legacy_target_size - 1) * strides[i] + kernel_shape[i] - dim_value;
         int64_t half_pad_small = total_pad >> 1;
         int64_t half_pad_big = total_pad - half_pad_small;
         if (auto_pad_attr->s() == "SAME_UPPER") {
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -95,6 +95,11 @@ void convPoolShapeInference(
     if (pads.size() != n_input_dims * 2) {
       fail_shape_inference("Attribute pads has incorrect size");
     }
+    for (const auto& pad : pads) {
+      if (pad < 0) {
+        fail_shape_inference("Attribute pads has negative values");
+      }
+    }
   } else {
     pads.assign(n_input_dims * 2, 0);
     const auto* auto_pad_attr = ctx.getAttribute("auto_pad");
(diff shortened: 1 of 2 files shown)


onnx/defs/nn/defs.cc
```diff
@@ -101,20 +101,12 @@ void convPoolShapeInference(
...
-        }
-        int64_t total_pad = residual == 0 ? kernel_shape[i] - strides[i] : kernel_shape[i] + residual;
-        if (total_pad < 0) {
-          fail_shape_inference("Stride is bigger than Kernel shape");
+        if (!input_shape.dim(2 + i).has_dim_value()) {
+          continue;
         }
+        int64_t dim_value = input_shape.dim(2 + i).dim_value();
+        int64_t legacy_target_size = (dim_value + strides[i] - 1) / strides[i];
+        int64_t total_pad = (legacy_target_size - 1) * strides[i] + kernel_shape[i] - dim_value;
```
stevenlix: What if there is the corner case,
input shape: 4, kernel shape: 2, strides: 4
then legacy_target_size = 1, total_pad will be negative value -2.

onnx/defs/nn/defs.cc
```diff
@@ -101,20 +101,12 @@ void convPoolShapeInference(
...
-          if (!input_shape.dim(2 + i).has_dim_value()) {
-            continue;
-          }
-          residual =  input_shape.dim(2 + i).dim_value();
-          while (residual > 0) {
-            residual -= strides[i];
-          }
-        }
-        int64_t total_pad = residual == 0 ? kernel_shape[i] - strides[i] : kernel_shape[i] + residual;
-        if (total_pad < 0) {
```
BowenBao: is(was) negative padding supported?

stevenlix: Not sure if negative padding means anything in real cases. For above corner case, we can safely set total_pad to 0 if it's negative, but there might be more corner cases.

onnx/defs/nn/defs.cc
```diff
@@ -101,20 +101,12 @@ void convPoolShapeInference(
     if ((nullptr != auto_pad_attr) && (auto_pad_attr->s() != "VALID")) {
```
BowenBao: it seems the check for ```auto_pad_attr->s() != "NOTSET"``` was removed in the previous PR. One edge case slipped through that when auto_pad=NOTSET, and pads are not provided. This doesn't affect the outcome though, as pads values will remain zero for NOTSET.

hariharans29: I think I did make this comment in the original PR - https://github.com/onnx/onnx/pull/1988

onnx/defs/nn/defs.cc
hariharans29: Added a check if negative pads are present when explicitly provided. Since this is not supported, it will result in shape inference error.

onnx/defs/nn/defs.cc
hariharans29: Thanks @stevenlix. Actually I had a similar concern when I first looked at the ORT shape computation logic. I feel this computation is more complex than it should be. If you substitute `legacy_target_size` from line 108 in the formula for `total_pad` in line 109, and after simplification, it is simply - `kernel_shape[i] - 1` (and this makes sense to me). Pads computation need not invlove strides at all as they are not related. Pads are to be computed only so that the filter is able to be fully accomodated for each pixel. So, I have made a change now. Please correct me if I am wrong.

hariharans29: @linkerzhang - Please comment on this line - 

`int64_t total_pad = kernel_shape[i] - 1;`

BowenBao: @hariharans29 I think this will result in different pad number, which changes the kernel center position and leads to different result.

hariharans29: @BowenBao - Thank you! But thinking out loud naively here, let's forget about the corner case that @stevenlix  pointed out for this discussion. The current pad computation logic used by ORT is here - 
https://github.com/microsoft/onnxruntime/blob/a7039601c4bf469131569d0c1261f13cd6bd67d9/onnxruntime/core/providers/cpu/nn/conv_base.h#L39

This was the previous logic in this same PR - 
`int64_t legacy_target_size = (dim_value + strides[i] - 1) / strides[i];
int64_t total_pad = (legacy_target_size - 1) * strides[i] + kernel_shape[i] - dim_value;`

Please substitute `legacy_target_size` into `total_pad` and simplify by doing some simple math, the result will be (if my math is correct) `kernel_shape[i] - 1` (which is the new logic I proposed). [Here](https://stackoverflow.com/questions/37674306/what-is-the-difference-between-same-and-valid-padding-in-tf-nn-max-pool-of-t) is another detailed discussion about this and it aligns with the ` kernel_shape[i] - 1`

As long as the total pads computed is the same, it should not affect kernel center position.

hariharans29:
Closing in favor of #2025

PR #2025, opened 2019-05-17
Relax shape inference failure when strides > kernel shape in conv shape inference

hariharans29:
When pad values are computed to be negative, it doesn't affect the output dim value in any way. In reality, it only means that a few values are ignored in the conv computation. Relaxing this condition because ResNet 50 from the ONNX model zoo has case(s) where strides > kernel shape and this causes shape inference failure


hariharans29:
@BowenBao @stevenlix - Any concerns with just relaxing the shape failure ?

stevenlix:
I can't edit the code in your fork, so I created a new PR here https://github.com/onnx/onnx/pull/2028 to fix the bug. Now resnet50 passed with the fix.

hariharans29:
@stevenlix - sounds good, thanks. Closing this.

PR #2028, opened 2019-05-17
Fix auto_pad shape inference bug

stevenlix:


onnx/defs/nn/defs.cc
```diff
@@ -102,19 +102,19 @@ void convPoolShapeInference(
...
+          while (residual >= stride) {
+            residual -= stride;
           }
         }
-        int64_t total_pad = residual == 0 ? kernel_shape[i] - strides[i] : kernel_shape[i] + residual;
-        if (total_pad < 0) {
-          fail_shape_inference("Stride is bigger than Kernel shape");
-        }
+        int64_t total_pad = residual == 0 ? kernel_shape[i] - stride : kernel_shape[i] - residual;
+        if (total_pad < 0)
```
hariharans29: LGTM.

Just 1 naive question - do we really need this mutation ? Even if total pads is negative, in reality, this means just ignoring a few values in the raw data ("unpad" a few values) right ? Also, I think it should not affect the output dim value when we use the formula - floor((in_dim - kernel_shape + total_pad) / s) + 1. Does this seem reasonable ?

stevenlix: I checked the corner cases and I think probably there is no impact if you leave total_pad as negative value, but for simplicity and clarity, it's better to keep it as non-negative value.

stevenlix: @linkerzhang Please review this PR

linkerzhang:
can you also add a test case to cover this bug case please? Thank you!

PR #2029, opened 2019-05-17
added initial passes for converting mythic keras to onnx for translat…

markbeards:
added initial passes for converting mythic keras to onnx for translating to compilerIR

In addition, it has 2 bug fixes:
1. construction of the optimizer graph
2. crash in identity driving an output

Minor change to improve the readability of the DOT graph generated from ONNX


markbeards:
Accidental PR

PR #2031, opened 2019-05-18
Support dynamic 'pads' and 'value' in Pad operator

### LINKED ISSUE #1765: Meaning of edge and reflect in Pad operator?
The meaning of edge and reflect are not described in the documentation for the Pad operator. There is an example, but please tell us the intended meaning in English if you would.

Thanks

### LINKED ISSUE #2313: Allow creation of older opset ops/nodes
It will be great if the onnx helper's `make_node()` abstraction helps us create older opset nodes/ops that we can use to cover shape inference tests and support testing the optimizer passesfor the older version of the op's spec. 

This will help guard against shape inference regressions for older opsets and test to make sure optimizer passes can handle correctly all versions of an op's spec.

hariharans29:
Change description:

This change is to make two static attributes of the Pad op - 'pads' and 'value' into dynamic inputs to support conversion of some TensorFlow models into ONNX which allow these 2 values to be fed in dynamically to the op. This change is to basically support conversion of such TF models into ONNX.

Summary of changes:

* Make interface change in Pad operator as described in title
* Rename 'value' to 'constant_value' to reflect its usage only for 'constant' mode
* Make related node test case changes
* Make related shape inference test changes
* Make related changes in optimizer passes
* Make related optimizer test changes

Issues that are resolved:
Resolves #1765 - Adds description and examples for 'Edge' and 'Reflect' mode
Resolves #2313 - Allows testing optimizer(s) for previous opset specs of ops/nodes


hariharans29:
> > Overall a good PR. Thanks for updating the optimizer as well.
> > Couple of minor points in line with best practices for op specifications:
> > 
> > 1. For posterity, if there's an op in any of the popular frameworks that has similar semantics or which this change is intended to support, could you capture that in the description?
> > 2. We are trying to capture a reference implementation for the ops in the tests. For this we have a np.pad which is close in semantics. For clarity, could you do a simple refactor to capture the implementation in a separate method (say `pad_reference_impl`), that uses np.pad, to be quite clear and readable.
> 
> Thanks Spandan. Will address your comments soon!

Addressed the feedback. Thanks.

wschin:
Please also sync with master. There is a minor conflict. You can just replace your lines with those from master.

hariharans29:
> Just realized that we still have an open item about the allowed tensor shapes for the `pads` input. Could you please address that as well before merging.

Hi Spandan - Sorry, I don't understand this. Should we limit pads input shapes to something specific ?

docs/Changelog.md
```diff
@@ -10488,3 +10488,58 @@ This version of the operator has been available since version 11 of the default
...
+<dd>Three modes: `constant`(default) - pads with a given constant value, `reflect` - pads with the reflection of the vector mirrored on the first and last values of the vector along each axis, `edge` - pads with the edge values of array</dd>
+</dl>
+
+#### Inputs (2 - 3)
+
+<dl>
+<dt><tt>data</tt> : T</dt>
+<dd>Input tensor.</dd>
+<dt><tt>pads</tt> : tensor(int64)</dt>
+<dd>Tensor of integers indicating the number of padding elements to add or remove (if negative) at the beginning and end of each axis. For 2D input tensor, it is the number of pixels. `pads` should be a 1D tensor of shape [2 * input_rank] or a 2D tensor of shape [1, 2 * input_rank]. `pads` format (1D example) should be as follow [x1_begin, x2_begin,...,x1_end, x2_end,...], where xi_begin is the number of pixels added at the beginning of axis `i` and xi_end, the number of pixels added at the end of axis `i`.</dd>
```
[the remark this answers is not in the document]
hariharans29: Sure. I will make it only 1D. Thanks.

hariharans29:
> Overall, looks good. Added a few minor comments. Two things to consider:
> 
> 1. Pursuant to previous feedback - if there are any equivalent ops in popular frameworks, consider adding them in the description of the spec for posterity.
> 2. I see that there's limited datatype support (float16, float32, double) for input/output for this op. Padding could be needed for tensors of other numeric types as well. Since we are updating this op, consider extending the datatype to other numeric types as well (ints).

1. The only op that i can think of is the tf.Pad op as this is basically to support a Keras (tf backed) model into ONNX. I have included that in the PR description already. 
2. I will expand type support to all int types.

docs/Changelog.md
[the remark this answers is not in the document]
spandantiwari: Great. Thanks.

docs/Operators.md
```diff
@@ -10026,19 +10046,19 @@ expect(node, inputs=[x], outputs=[y],
...
     )
-    x = np.random.randn(1, 3, 4, 5).astype(np.float32)
-    y = np.pad(
+    x = np.random.randn(1, 3, 4, 5).astype(np.int32)
+    pads = np.array([0, 0, 1, 1, 0, 0, 1, 1]).astype(np.int64)  # pad order [x1_begin, x2_begin, ..., x1_end, x2_end, ...]
+    y = pad_impl(
         x,
-        pad_width=((0, 0), (0, 0), (1, 1), (1, 1)),
-        mode=mode,
+        ((0, 0), (0, 0), (1, 1), (1, 1)),  # re-order to np.pad accepted order ((x1_begin, x1_end), (x2_begin, x2_end), ...)
```
wschin: `pad_impl` is a reference implementation of ONNX `Pad`. It should accept ONNX `pads` directly.

onnx/defs/tensor/defs.cc
```diff
@@ -2594,4 +2514,122 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(
+            1,
+            "pads",
+            "Tensor of integers indicating the number of padding elements to add or remove (if negative) "
+            "at the beginning and end of each axis. For 2D input tensor, it is the number of pixels. "
+            "`pads` should be a 1D tensor of shape [2 * input_rank]. "
+            "`pads` format should be: [x1_begin, x2_begin,...,x1_end, x2_end,...], "
+            "where xi_begin is the number of pad values added at the beginning of axis `i` and "
+            "xi_end, the number of pad values added at the end of axis `i`.",
+            "tensor(int64)")
```
wschin: I feel `int32` is big enough for `pads`.

onnx/defs/tensor_util.cc
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) ONNX Project Contributors.
+// Licensed under the MIT license.
+
+#include "tensor_util.h"
+#include <vector>
+#include "onnx/common/platform_helpers.h"
+
+namespace ONNX_NAMESPACE {
+
+#define DEFINE_PARSE_DATA(type, typed_data_fetch)                          \
```
wschin: Why do you need this function? Is it for processing int64 `pads`? Where is the old code that runs before?

onnx/test/optimizer_test.py
```diff
@@ -194,12 +199,17 @@ def check_pad(node):  # type: (NodeProto) -> None
...
-            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (2, 3))],
-            [helper.make_tensor_value_info("Y", TensorProto.FLOAT, (2, 4))])
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (2, 3)),
+             helper.make_tensor_value_info("Pads", TensorProto.INT64, (4,))],
+            [helper.make_tensor_value_info("Y", TensorProto.FLOAT, (2, 4))],
+            [helper.make_tensor("Pads", TensorProto.INT64,
+                dims=(4,),
+                vals=np.array([0, 1, 0, 0]).astype(np.int64).tobytes(),
+                raw=True)])
         optimized_model = self._optimized(graph, ["eliminate_nop_pad"])
```
wschin: I guess this `Pad` won't be removed. If yes, could you rename this test to `not_eliminate_pad`?

onnx/test/optimizer_test.py
```diff
@@ -766,42 +776,48 @@ def test_fuse_matmul_add_bias_into_gemm_multiple_use_no_fuse(self):  # type: ()
     def test_fuse_pad_into_conv(self):  # type: () -> None
```
wschin: Is it possible to keep the test for old Pad?

onnx/defs/tensor/defs.cc
hariharans29: I prefer to keep it `int64` to be consistent with other usage - 

1) Previously the attributes were `int64` type

2) Other very similar semantics use `int64`. For example, ConstantOfShape - the tensor type for the shape is `int64` (where `int32` would actually do)

3) There is not much value add in making it `int32` (not much space savings) when it is safer to keep it `int64`.

onnx/defs/tensor_util.cc
hariharans29: There is no such code needed to run before. Previously, `pads` was an attribute and hence there were semantics already in place to parse the data within the attribute. Now `pads` is an input tensor and we need this code to parse the data within the struct now. Hence, this code is needed and used within the optimizer code.

onnx/test/optimizer_test.py
hariharans29: Good point. will do.

docs/Operators.md
hariharans29: Sure. Will do.

onnx/test/optimizer_test.py
hariharans29: I tried. But I am not sure we can create a node for an older opset. I notive that there is something within the `circleci` CI that covers the old optimizer logic because when I removed the old logic, it broke the CI and when the optimizer supported both versions - the CI passed....

docs/Operators.md
hariharans29: Fixed.

onnx/test/optimizer_test.py
hariharans29: Done. thanks for catching this.

hariharans29: by the way - the line your comment is pointing to is the optimizer pass name - that cannot be modified. but the test name can be modified and I modified that.

onnx/defs/tensor_util.cc
wschin: Can they be merged into one function?

onnx/defs/tensor/defs.cc
wschin: Make senses. Thanks.

onnx/test/optimizer_test.py
wschin: Ok. I think old optimization is implicitly tested somewhere. Could you please one an issue to discuss if we should add more tests for old optimizations? Then, we can close this comment.

onnx/defs/tensor_util.cc
hariharans29: I feel this is maybe not a good idea to merge ALL the typed parsing related logic into a single function because -

1) It's really hard to merge them all into one function and honestly it spoils readability of the code as there will be plenty of type checking both while parsing raw data or the parsing the buffer corresponding to the type. This way the code is much cleaner and easily readable as it is templated. 

2) This template will not be instantiated a lot of times -atmost a few times for the types that we need it for. So will not bloat up size too much. (Code readability vs size trade-off)

3) I would like to keep it consistent with tensor_proto_util.cc which has similar cleaner templated implementation

onnx/test/optimizer_test.py
hariharans29: sure, will do.

hariharans29: Raised this issue here - https://github.com/onnx/onnx/issues/2313. Resolving comment as prescribed. Thanks!

docs/Changelog.md
```diff
@@ -11461,6 +11461,81 @@ This version of the operator has been available since version 11 of the default
...
+  1) `constant`(default) - pads with a given constant value as specified by `value` (which defaults to 0)
+  
+  2) `reflect` - pads with the reflection of the vector mirrored on the first and last values of the vector along each axis
+  
+  3) `edge` - pads with the edge values of array
+  
+  Example (`constant` mode):
+    Insert 0 pads to the beginning of the second dimension.
+  
+    data =
```
ebarsoum: Add example for other modes.

docs/Changelog.md
```diff
@@ -11461,6 +11461,81 @@ This version of the operator has been available since version 11 of the default
...
+</dl>
+
+#### Inputs (2 - 3)
+
+<dl>
+<dt><tt>data</tt> : T</dt>
+<dd>Input tensor.</dd>
+<dt><tt>pads</tt> : tensor(int64)</dt>
+<dd>Tensor of integers indicating the number of padding elements to add or remove (if negative) at the beginning and end of each axis. For 2D input tensor, it is the number of pixels. `pads` should be a 1D tensor of shape [2 * input_rank]. `pads` format should be: [x1_begin, x2_begin,...,x1_end, x2_end,...], where xi_begin is the number of pad values added at the beginning of axis `i` and xi_end, the number of pad values added at the end of axis `i`.</dd>
+<dt><tt>value</tt> (optional) : T</dt>
```
ebarsoum: rename it to constant_value

hariharans29: Makes sense. Will rename. Thanks. 

(Updates PR change description to reflect this)

docs/Changelog.md
hariharans29: Will add

hariharans29: Add example for 'reflect' and 'edge' modes. Resolved.

docs/Changelog.md
hariharans29: Renamed to 'constant_value'. Resolved.

onnx/test/shape_inference_test.py
```diff
@@ -615,20 +615,6 @@ def test_slice_variable_input_types(self):  # type: () -> None
                          make_tensor('axes', TensorProto.INT32, (2,), (0, 1))])
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.DOUBLE, (2, 2))])
 
-    def test_pad(self):  # type: () -> None
```
hariharans29: Moved down and renamed as "..._opset10"

onnx/test/optimizer_test.py
hariharans29: I preserved the old Pad tests. Added some infrastructure code for it. Will resolve the issue after PR merge.

onnx/test/shape_inference_test.py
hariharans29: Resolved. (FYI comment)

Exlsunshine:
Hi @hariharans29, glad to see your change has been merged into master, but I still got the following error:
```
OP=Pad
Name=some/path/to/op/Pad
Inputs:
        some/path/to/op/concat:0=Concat, [-1, -1], 1
        some/path/to/op/Pad/paddings_Concat__113:0=Concat, [2, 2], 6
Outpus:
        some/path/to/op/Pad:0=[-1, -1], 1
  File "D:\Python\Python36\lib\site-packages\tf2onnx\tfonnx.py", line 352, in tensorflow_onnx_mapping
    func(g, node, **kwargs)
  File "D:\Python\Python36\lib\site-packages\tf2onnx\onnx_opset\nn.py", line 401, in version_1
    paddings = np.array(node.inputs[1].get_tensor_value()).transpose().flatten()
  File "D:\Python\Python36\lib\site-packages\tf2onnx\graph.py", line 256, in get_tensor_value
    raise ValueError("get tensor value: {} must be Const".format(self.name))
ValueError: get tensor value: some/path/to/op/Pad/paddings_Concat__113 must be Const
2019-11-21 15:52:39,788 - VERBOSE - tf2onnx.tfonnx: Mapping TF node to ONNX node(s)
2019-11-21 15:52:40,098 - ERROR - tf2onnx.tfonnx: Failed to convert node some/path/to/op/Fill


OP=ConstantOfShape
Name=some/path/to/op/Fill
Inputs:
        some/path/to/op/Fill__851:0=Cast, [1], 7
        some/path/to/op/strided_slice_21__842:0=Cast, [], 6
Outpus:
        some/path/to/op/Fill:0=[-1], 6
  File "D:\Python\Python36\lib\site-packages\tf2onnx\tfonnx.py", line 352, in tensorflow_onnx_mapping
    func(g, node, **kwargs)
  File "D:\Python\Python36\lib\site-packages\tf2onnx\onnx_opset\generator.py", line 100, in version_9
    value = np.array([node.inputs[1].get_tensor_value()]).astype(utils.map_onnx_to_numpy_type(dtype))
  File "D:\Python\Python36\lib\site-packages\tf2onnx\graph.py", line 256, in get_tensor_value
    raise ValueError("get tensor value: {} must be Const".format(self.name))
ValueError: get tensor value: some/path/to/op/strided_slice_21__842 must be Const
```

I am using:
```
python -m tf2onnx.convert --input a.pb --inputs in_0:0,in_1:0 --outputs o_0:0,o_1:0 --output b.onnx --opset 11 --fold_const  --continue_on_error --verbose

2019-11-21 15:52:31,785 - INFO - tf2onnx.tfonnx: Using tensorflow=1.13.1, onnx=1.6.0, tf2onnx=1.5.3/7b598d
2019-11-21 15:52:31,785 - INFO - tf2onnx.tfonnx: Using opset <onnx, 11>

```

Do you have any advice about this error? Thanks!

PR #2041, opened 2019-05-22
Add shape inference logic for Expand op

hariharans29:
cc: @stevenlix


onnx/defs/math/defs.cc
```diff
@@ -1158,7 +1158,78 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            if (shape_initializer_shape.dim_size() != 1 ||
+                shape_initializer->data_type() != TensorProto::INT64)
+              fail_shape_inference(
+                  "'shape' input must be 1D tensor of type INT64");
+
+            const auto& input_shape =
+                ctx.getInputType(0)->tensor_type().shape();
+            const auto& shape_data = ParseData<int64_t>(shape_initializer);
+
+            size_t max_dimension_count = std::max(
```
linkerzhang: is this following the same broadcasting as https://github.com/onnx/onnx/blob/master/onnx/defs/shape_inference.h#L435 please?

hariharans29: @linkerzhang - indeed it is. Thanks for pointing me to these helper functions. Resolved.

PR #2048, opened 2019-05-23
Clarify the axis/size in pads, strides, dilations

daquexian:
I was confusing that the pads, strides and dilations are only for the spatial axes or all axes. After studying the models in model zoo, and also the code for shape inference, I figured out they are only for the spatial axes, however, the doc says they are for "each axis". I updated the docs to clarify it.


daquexian:
CircleCI fails due to an unrelated error:

> ________________ ERROR collecting test/onnx/test_onnx_opset.py _________________
> ImportError while importing test module '/tmp/pytorch/test/onnx/test_onnx_opset.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> test/onnx/test_onnx_opset.py:11: in <module>
>     from torch.onnx.symbolic_helper import _export_onnx_opset_version
> E   ImportError: No module named symbolic_helper

linkerzhang:
@postrational

PR #2054, opened 2019-05-23
Fix torchvision installation

bddppq:


bddppq:
https://github.com/pytorch/pytorch/pull/20890

PR #2055, opened 2019-05-24
Fix inconsistency while performing shape inference that involves broadcasting

hariharans29:
We must stop at just incrementing rank for a specific index of output shape when we see a mixture of symbolic dims and dim values in the input shapes for that specific index (currently, the seen dim value from the input shapes will be propagated to the output if and only if > 1 - which doesn't seem to make sense)

When we see both dim values and symbolic dims (one or multiple), we don't have enough information to compute value (or symbolic dim) that will get set for the output index. So, it is safer to stop after we increment rank to account for the new dimension in the output shape


onnx/test/shape_inference_test.py
```diff
@@ -154,8 +154,8 @@ def test_matmul_allow_unknown(self):  # type: () -> None
         self._make_matmul_test_allow_unknown((4, 2), (2, "a"), (4, "a"))
         self._make_matmul_test_allow_unknown((4, None), (2, "a"), (4, "a"))
         self._make_matmul_test_allow_unknown((4, None), (None, "a"), (4, "a"))
-        self._make_matmul_test_allow_unknown((1, 4, 2), ("a", 2, 5), ("a", 4, 5))
```
hariharans29: the output "a" seems arbitrary - the input dims have 1 and "a" - so ideally, we need to stop at incrementing rank and not setting any value (actual value or symbolic)

hariharans29:
@linkerzhang and @askhade - could you please review this and give your feedback ?

gramalingam:
Hi: the logic here is that broadcasting happens only if one of the dims is 1 and the other dim is > 1. So, if you see any of the input-dims has a value > 1, we can be sure that the corresponding dim will have that value (otherwise the broadcast will fail with a runtime error). If the input-dim has a value 1, then the corresponding output-dim may be > 1 due to broadcasting. So, I believe that the existing logic for shape-inference is correct.

hariharans29:
> Hi: the logic here is that broadcasting happens only if one of the dims is 1 and the other dim is > 1. So, if you see any of the input-dims has a value > 1, we can be sure that the corresponding dim will have that value (otherwise the broadcast will fail with a runtime error). If the input-dim has a value 1, then the corresponding output-dim may be > 1 due to broadcasting. So, I believe that the existing logic for shape-inference is correct.

Hi @gramalingam - You are right. Sorry - I didn't mean to say the broadcasting the loigc was incorrect. What I meant was there seems to be some arbitrary decision making when the final dim value is assigned to the output shape.

In the existing logic, when computed dim value is greater than 1 (after it passes throuh the afore-mentioned logic), even if it sees symbolic dims in other inputs, it seems to assign the seen dim value. My argument is that the dim value should not be assigned - rather it should just stop with rank inference.

gramalingam:
Hi @hariharans29 : the idea is to return the most precise information possible (which, in turn, may enable more optimizations). A numeric dimension (e.g., 2) is more precise than a symbolic dimension (e.g., "a"), which is more precise than None.  In the first case, both "a" and None are correct (but 1 would be wrong), and we return "a" as it is more precise. In the second case, 2 and None are both correct, but 2 is the most precise and we return that. Same with the third case.

hariharans29:
> Hi @hariharans29 : the idea is to return the most precise information possible (which, in turn, may enable more optimizations). A numeric dimension (e.g., 2) is more precise than a symbolic dimension (e.g., "a"), which is more precise than None. In the first case, both "a" and None are correct (but 1 would be wrong), and we return "a" as it is more precise. In the second case, 2 and None are both correct, but 2 is the most precise and we return that. Same with the third case.

Thanks for the detailed explanation @gramalingam. That makes sense. Closing the PR.

Issue #1774, opened 2019-01-29
Differentiate modes in Upsample operator


vandanavk:
Currently, [Upsample operator](https://github.com/onnx/onnx/blob/master/docs/Changelog.md#upsample-9) in ONNX specifies mode as 'linear' for linear, bilinear, trilinear etc. Changing this to exactly specify which type of linear mode it is, will help importing in different frameworks.

Or is there any other way to distinguish between linear, bilinear from ONNX Upsample?

linkerzhang:
Thank you. Yep. the spec is not clear. We'd clarify it by either having linear meaning bilinear or adding bilinear/trilinear to make the mode clear.

Do you want to send out a PR for this please?

jcwchen:
Upsample has been deprecated and the doc for similar op Resize has been clarified about what these modes are. Close it now. Thanks!

PR #2057, opened 2019-05-26
Update resize op

daquexian:
1. Add "cubic" interpolation mode, and clarify the "linear", "bilinear" and "trilinear" (https://github.com/onnx/onnx/issues/1774)
2. Update the tests: The current tests of resize linear follows the old behavior of TF, which is changed now. Related issue: https://github.com/tensorflow/tensorflow/issues/6720, https://github.com/onnx/onnx/issues/2070
3. Add "coordinate_transformation_mode" attribute including "half_pixel", "align_corners", "asymmetric", "tf_crop_and_resize" and so on
4. Add a new input 'sizes' as requested in #2062.
5. Add two new attrs 'cubic_coeff_a' and 'exclude_outside' for the compatibility with TensorFlow. There is no standard implementation of cubic interpolation. [TensorFlow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/resize_bicubic_op.cc#L60) (and also [matlab](https://theailearner.com/2018/12/29/image-processing-bicubic-interpolation/)) sets a=-0.5 (In the legacy version of TF, a=-0.75) and [exclude_outside=True](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/resize_bicubic_op.cc#L95), while PyTorch (and OpenCV) sets a=-0.75 and exclude_outside=False

Both cubic interpolation mode and align_corners attribute are supported by [tensorflow](https://www.tensorflow.org/api_docs/python/tf/image/resize_bicubic), [pytorch](https://pytorch.org/docs/stable/nn.html#torch.nn.functional.interpolate) and so on.

Note: 
* Some frameworks set `output_dimension` to `round(input_dimension * scale)` instead of `floor(input_dimension * scale)`, some frameworks re-assign the `scale` to `output_dimension / input_dimension`. I think it is not good to add two extra attributes in resize op for these cases. This op has many attributes already, and the converters in frameworks have the ability and responsibility to make the `scale` compatible with the behavior of this op.

* This proposal is fully compatible with TensorFlow and PyTorch. Please refer to [the checking script](https://gist.github.com/daquexian/4bab98a7e706ad95616567433c1234b8 ) for the attribute values


daquexian:
About the existing example/test:

It is following the behavior of tensorflow, which has been considered as a bug and fixed in TensorFlow nightly (https://github.com/tensorflow/tensorflow/issues/6720)

```python
import tensorflow as tf
import numpy as np

print(tf.__version__)
a=np.array([[1,2],[3,4]])
input=tf.constant(a)
input4d=tf.reshape(input, [1,2,2,1])
resize=tf.compat.v2.image.resize(input4d,[4,4])
sess=tf.Session()
r1 = sess.run(resize)
print(r1)
```
gives
```python
1.14.1-dev20190526
[[[[1.  ]
   [1.25]
   [1.75]
   [2.  ]]

  [[1.5 ]
   [1.75]
   [2.25]
   [2.5 ]]

  [[2.5 ]
   [2.75]
   [3.25]
   [3.5 ]]

  [[3.  ]
   [3.25]
   [3.75]
   [4.  ]]]]
```

daquexian:
I have updated the tests :)

linkerzhang:
@postrational

onnx/defs/tensor/defs.cc
```diff
@@ -1355,9 +1355,24 @@ ONNX_OPERATOR_SET_SCHEMA(
     OpSchema()
```
linkerzhang: please bump the op version. You may refer to https://github.com/onnx/onnx/blob/master/docs/Versioning.md#operator-versioning to check onnx versioning design.

onnx/backend/test/case/node/resize.py
```diff
@@ -104,8 +132,170 @@ def export_downsample_linear():  # type: () -> None
...
+            [ 1,  2,  3,  4],
+            [ 5,  6,  7,  8],
+            [ 9, 10, 11, 12],
+            [13, 14, 15, 16],
+        ]]], dtype=np.float32)
+
+        scales = np.array([1.0, 1.0, 2.0, 2.0], dtype=np.float32)
+
+        output = np.array([[[
+            [ 0.47265625,  0.76953125,  1.24609375,  1.87500000,  2.28125000,
```
wschin: Where do these numbers come from? It's not straightforward to write code to implement this behavior. Would you consider adding more `math` details and a lightweight numpy implementation here?

linkerzhang:
please also fix the ci failure due to flake8 errors.

like, 
+flake8
./onnx/backend/test/case/node/resize.py:176:14: E201 whitespace after '['
./onnx/backend/test/case/node/resize.py:176:17: E241 multiple spaces after ','
./onnx/backend/test/case/node/resize.py:176:21: E241 multiple spaces after ','
.....

onnx/backend/test/case/node/resize.py
daquexian: @wschin 

These numbers come from pytorch and opencv. The result of opencv is the same as pytorch with align_corners=False.

I have implemented linear and cubic interpolation [here](https://gist.github.com/daquexian/18b3fda2c7d2a6ad236a0a5ded4fe845) :)

daquexian:
@linkerzhang I have upgraded the op version and fixed flake8 errors. :)

daquexian:
I have figured out that I should add the upgraded operator in `onnx/defs/operator_sets.h`. Now the docs look right.

onnx/backend/test/case/node/resize.py
wschin: Could you please have a Python version of your implementation (only depending on numpy) and compute the test outputs using that Python implementation? This operator is important and also very sophisticated, so a reference implementation is desired. Please take a look at #2066 for more reasons.

daquexian: All right, thanks for your suggestion :) I'll do it soon.

daquexian:
I will resolve the conflict of Changelog.md soon.

I added a new input 'sizes' as requested in https://github.com/onnx/onnx/issues/2062. 

Furthermore, I added two new attrs, 'cubic_coeff_a' and 'exclude_outside', for the compatibility with TensorFlow. TensorFlow sets [A=-0.5](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/resize_bicubic_op.cc#L60) (In its legacy version, A=-0.75) and [exclude_outside=True](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/resize_bicubic_op.cc#L95), while PyTorch (and OpenCV) sets A=-0.75 and exclude_outside=False

daquexian:
@linkerzhang @wschin I have added the implementation in the tests. It is ready to review. I'll implement the shape inference after https://github.com/onnx/onnx/pull/2085 is merged.

lara-hdr:
Is this ready? We are also waiting for this PR, @ebarsoum @houseroad

daquexian:
@lara-hdr I'll update the shape inference soon, and then this PR will be ready to review :)

daquexian:
@ebarsoum @linkerzhang @houseroad @wschin I have updated shape inference. This PR is ready to review :)

daquexian:
@linkerzhang I have checked the proposed RoiCropAndResize op. I found that it is actually unable to compose it by Slice and Resize ops. The coordinates of rois can be non-integer, however, the Slice op can only generate an output tensor with integer height and width. Maybe we can combine RoiCrpAndResize op with Resize op by adding "crop" and "extrapolation_value" attribute?

jiafatom:
> @linkerzhang I have checked the proposed RoiCropAndResize op. I found that it is actually unable to compose it by Slice and Resize ops. The coordinates of rois can be non-integer, however, the Slice op can only generate an output tensor with integer height and width. Maybe we can combine RoiCrpAndResize op with Resize op by adding "crop" and "extrapolation_value" attribute?
> 
> @jiafatom @ebarsoum @BowenBao What's your opinion? :)
What do you mean "combine RoiCrpAndResize op with Resize op by adding "crop" and "extrapolation_value" attribute?" 
RoiCropAndResize op has extrapolation_value attribute. Where do you want to add "crop" and "extrapolation_value"? To Resize op?

daquexian:
> > @linkerzhang I have checked the proposed RoiCropAndResize op. I found that it is actually unable to compose it by Slice and Resize ops. The coordinates of rois can be non-integer, however, the Slice op can only generate an output tensor with integer height and width. Maybe we can combine RoiCrpAndResize op with Resize op by adding "crop" and "extrapolation_value" attribute?
> > @jiafatom @ebarsoum @BowenBao What's your opinion? :)
> 
> What do you mean "combine RoiCrpAndResize op with Resize op by adding "crop" and "extrapolation_value" attribute?"
> RoiCropAndResize op has extrapolation_value attribute. Where do you want to add "crop" and "extrapolation_value"? To Resize op?

Yes, added crop and extrapolation_value attributes to resize op.

jiafatom:
Resize is a special case of RoiCropAndResize, where the roi for Resize is [0, 0, 1, 1] (normalized coordinates). On the other hand, the current RoiCropAndResize is used only for 2d case, there is no use case to extend it to n-d.

So I feel it not good to add crop and extrapolation_value on resize op, because these two are only used in RoiCropAndResize.

daquexian:
@jiafatom thanks for your reply. Do you have any ideas on how to define roicropandresize as a function? I think the problem is the non-integer rois (as said in https://github.com/onnx/onnx/pull/2057#issuecomment-510722131)

jiafatom:
To define roicropandresize as a function, we need enhance resize op to make it like a crop_and_resize for a single ROI. Then like you said, we need add one input as crop, the other attributes as extrapolation_value. This way we make resize op as a general roicropandresize. Actually I have concerns about this because there is no use case for n-d roicropandresize other than 2d, it is over-killing the problem.

Another thing: Why non-integer roi is a problem? The input roi range is from 0 to 1 because it is normalized. if it is [0, 0, 1, 1], then it is actually original resize. No matter it is an integer or not, we need find the corresponding pixel in the input tensor (which is also a floating number), and then apply specific modes such as bilinear, nearest (it will use the integer coordinates in the original input).

daquexian:
> Why non-integer roi is a problem? The input roi range is from 0 to 1 because it is normalized. if it is [0, 0, 1, 1], then it is actually original resize. No matter it is an integer or not, we need find the corresponding pixel in the input tensor (which is also a floating number), and then apply specific modes such as bilinear, nearest (it will use the integer coordinates in the original input).

What I mean is, we cannot define cropandresize as a function based on slice op and the existing resize op as https://github.com/onnx/onnx/pull/2142#issuecomment-507898382 since the output of slice has integer size but the roi has float size. As a result, we have to expand resize op by adding a bulit-in crop attribute.

daquexian:
> Actually I have concerns about this because there is no use case for n-d roicropandresize other than 2d, it is over-killing the problem.
@jiafatom I think it is necessary to add this "crop" input. There is no other way to handle the float roi coordinates. Please correct me if I am wrong.

jiafatom:
> @jiafatom I think it is necessary to add this "crop" input. There is no other way to handle the float roi coordinates. Please correct me if I am wrong.

I agree that we need add "crop" input to support roi coordinates :)

onnx/defs/tensor/defs.cc
```diff
@@ -1358,35 +1358,75 @@ ONNX_OPERATOR_SET_SCHEMA(
         })
 );
 
-static const char* Resize_ver10_doc = R"DOC(
+static const char* Resize_ver11_doc = R"DOC(
 Resize the input tensor.
 Each dimension value of the output tensor is:
   output_dimension = floor(input_dimension * scale).
```
BowenBao: add "if input 'sizes' is not specified"

onnx/defs/tensor/defs.cc
```diff
@@ -1358,35 +1358,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 ONNX_OPERATOR_SET_SCHEMA(
     Resize,
-    10,
+    11,
     OpSchema()
         .Attr(
             "mode",
-            "Two interpolation modes: nearest (default), and linear (including bilinear, trilinear, etc)",
+            "Three interpolation modes: nearest (default), linear and cubic. "
+            "The \"linear\" mode includes linear interpolation for 1D tensor and N-D linear interpolation for N-D tensor (For example, trilinear interpolation for 2D tensor). "
```
BowenBao: > "The \"linear\" mode includes linear interpolation for 1D tensor and N-D linear interpolation for N-D tensor (For example, trilinear interpolation for 2D tensor). "

should it be "_bilinear_ interpolation for 2D tensor" ?

onnx/defs/tensor/defs.cc
```diff
@@ -1358,35 +1358,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "The \"cubic\" mode includes cubic interpolation for 1D tensor and N-D cubic interpolation for N-D tensor (For example, tricubic interpolation for 2D tensor)."
```
BowenBao: same above, bicubic?

winnietsang:
@daquexian which Tensorflow operator will create the same result as your unit test. There are tf.image.resize_images and tf.compat.v2.image.resize.
tf.image.resize_images has align_corners as one of the args but tf.compat.v2.image.resize don't.
In your earlier comment you mention that you have try this on Tensorflow, may you please share with me which tf operator are you using? Thanks

daquexian:
> @daquexian which Tensorflow operator will create the same result as your unit test. There are tf.image.resize_images and tf.compat.v2.image.resize.
> tf.image.resize_images has align_corners as one of the args but tf.compat.v2.image.resize don't.
> In your earlier comment you mention that you have try this on Tensorflow, may you please share with me which tf operator are you using? Thanks

I tested with tf.compat.v2.image.resize. I'm not familiar with tensorflow. It seems that the result of tf.image.resize_images and tf.compat.v2.image.resize are different.

spandantiwari:
@daquexian - could you please confirm if all the feedback has been addressed. This is an important change and we  are waiting for this PR to merge in ONNX.

daquexian:
@spandantiwari I'll update this PR soon. Thanks :)

onnx/defs/tensor/defs.cc
daquexian: Thanks! I'll edit them soon.

spandantiwari:
@daquexian - thanks a lot!

daquexian:
@winnietsang tf.v1 behavior with half_pixel_centers=False is supported by the new 'tf_legacy_scaler` attribute

daquexian:
@jiafatom I have updated this PR and added input "roi" and attr "extrapolation_value". I'll check whether the output is consistent with your PR tomorrow.

onnx/defs/tensor/defs.cc
daquexian: Thanks! It is added.

onnx/defs/tensor/defs.cc
daquexian: Thanks! It is fixed.

onnx/defs/tensor/defs.cc
daquexian: Thanks! It is fixed.

onnx/test/shape_inference_test.py
```diff
@@ -305,7 +305,19 @@ def test_expand_raw_data(self):  # type: () -> None
             graph,
             [make_tensor_value_info('y', TensorProto.INT32, (3, 4))])
 
-    def test_resize(self):  # type: () -> None
+    def test_resize_size(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.INT32, (2, 4, 3, 5)),
+             ('scales', TensorProto.FLOAT, (4,)),
+             ('sizes', TensorProto.INT64, (4,))],
```
jiafatom: add roi here?

onnx/backend/test/case/node/resize.py
```diff
@@ -8,12 +8,221 @@
...
+
+def linear_coeffs(ratio):           # type: (float) -> np.ndarray
+    return np.array([1 - ratio, ratio])
+
+
+def nearest_coeffs(ratio):          # type: (float) -> np.ndarray
+    return np.array([1])
 
 
 class Resize(Base):
```
jiafatom: add roi in all the test case?

docs/Operators.md
```diff
@@ -11663,43 +11663,161 @@ for test_name, shape in test_cases.items():
...
-#### Inputs
+#### Inputs (2 - 3)
 
 <dl>
-<dt><tt>X</tt> : T</dt>
+<dt><tt>X</tt> : T1</dt>
 <dd>N-D tensor</dd>
-<dt><tt>scales</tt> : tensor(float)</dt>
-<dd>The scale array along each dimension. It takes value greater than 0. If it's less than 1, it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should be the same as the rank of input 'X'.</dd>
+<dt><tt>scales</tt> : T2</dt>
```
jiafatom: roi missing?

daquexian: Thanks! I have not re-generate the doc yet.

onnx/backend/test/case/node/resize.py
daquexian: Thanks! Will add.

onnx/test/shape_inference_test.py
daquexian: Added. Thanks!

onnx/backend/test/case/node/resize.py
daquexian: Added. Thanks!

docs/Operators.md
daquexian: Done. Thanks!

daquexian:
@jiafatom I have updated this PR. The output of my implementation is the same as yours. :)

daquexian:
@ebarsoum @linkerzhang @houseroad @wschin @jiafatom @BowenBao @spandantiwari It is ready to review :)

winnietsang:
@daquexian I'm in the process to verify all of your test cases using Tensorflow ops, I will update you shortly once I have some preliminary result.

daquexian:
> @daquexian I'm in the process to verify all of your test cases using Tensorflow ops, I will update you shortly once I have some preliminary result.

Please set scaler="tf_legacy" for the case half_pixel_centers=False, and set scaler="half_pixel", cubic_coeff_a=-0.5, exclude_outside=True for the case half_pixel_center=True

winnietsang:
> > @daquexian I'm in the process to verify all of your test cases using Tensorflow ops, I will update you shortly once I have some preliminary result.
> 
> Thanks! Please set scaler="tf_legacy" for the case half_pixel_centers=align_corners=False, and set scaler="half_pixel", cubic_coeff_a=-0.5, exclude_outside=True for the case half_pixel_center=True

@daquexian I try the 4 testcases for mode=nearest and here are my findings:

1. upsample_scales_nearest: able to get the same values as yours with 
tf.compat.v2.image.resize, 
tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=False) and tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True)
 
2. downsample_scales_nearest: can only get the same values as yours with     tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=False)
but got different values with 
tf.compat.v2.image.resize and    
tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True)

3. upsample_size_nearest: can only get the same values as yours with 
tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=False)  
but got different values with 
tf.compat.v2.image.resize and    
tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True)

4. downsample_size_nearest: got different values with 
tf.compat.v2.image.resize, 
tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=False) and tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True)

Based on above results, I think the implementations of half_pixel_centers in onnx and tensorflow are different. Can you possibly look into this?

I also changed the initial values for case 1, scale = [1.0, 1.0, 2.5, 3.0], and saw the results not match with the expected values in tf.compat.v2.image.resize or tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True). So this seems to again indicate the half_pixel_centers has significant impact to the expected results.

daquexian:
> > > @daquexian I'm in the process to verify all of your test cases using Tensorflow ops, I will update you shortly once I have some preliminary result.
> > 
> > 
> > Thanks! Please set scaler="tf_legacy" for the case half_pixel_centers=align_corners=False, and set scaler="half_pixel", cubic_coeff_a=-0.5, exclude_outside=True for the case half_pixel_center=True
> 
> @daquexian I try the 4 testcases for mode=nearest and here are my findings:
> 
> 1. upsample_scales_nearest: able to get the same values as yours with
>    tf.compat.v2.image.resize,
>    tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=False) and tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True)
> 2. downsample_scales_nearest: can only get the same values as yours with     tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=False)
>    but got different values with
>    tf.compat.v2.image.resize and
>    tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True)
> 3. upsample_size_nearest: can only get the same values as yours with
>    tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=False)
>    but got different values with
>    tf.compat.v2.image.resize and
>    tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True)
> 4. downsample_size_nearest: got different values with
>    tf.compat.v2.image.resize,
>    tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=False) and tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True)
> 
> Based on above results, I think the implementations of half_pixel_centers in onnx and tensorflow are different. Can you possibly look into this?
> 
> I also changed the initial values for case 1, scale = [1.0, 1.0, 2.5, 3.0], and saw the results not match with the expected values in tf.compat.v2.image.resize or tf.compat.v1.image.resize_nearest_neighbor(half_pixel_centers=True). So this seems to again indicate the half_pixel_centers has significant impact to the expected results.

Thanks! I found there are some differences in the implementation of nearest resize. I'll add a new attribute for this case soon.

daquexian:
@winnietsang I have checked the result of this PR vs tensorflow on linear and cubic interpolation.

Here is the full code:

https://gist.github.com/daquexian/4bab98a7e706ad95616567433c1234b8

 The resize implementation of tf.v2 are the same with those of tf.v1 with half_pixel_centers=True, please check out [tf.v1](https://github.com/tensorflow/tensorflow/blob/v1.14.0/tensorflow/python/ops/image_ops_impl.py#L1102) and [tf.v2](https://github.com/tensorflow/tensorflow/blob/v1.14.0/tensorflow/python/ops/image_ops_impl.py#L1186) to validate it.

winnietsang:
@daquexian I try to verify the testcase for linear and cubic today, I got a lot of cases that the expected values doesn't match with the values returned from Tensorflow. 
Then I switch to look at the test code you refer https://gist.github.com/daquexian/4bab98a7e706ad95616567433c1234b8 yesterday. I try to run downsample_scales_linear in that code by changing h=1, w=2 and resize='linear' and I got matching result from expected and tensorflow values. I also try to run upsample_sizes_cubic and downsample_sizes_cubic cases in your test code, both get matching values. 
Then I go back to the onnx unit test for resize in this PR, I notice that not all cubic testcase with scaler=half_pixel have the exclude_outside attribute set. And the ﻿coeffs value are different in the onnx unit test and your test code. May you please look into this and validate the unit test in this PR match with the test code you provide yesterday? Thanks

daquexian:
> @daquexian I try to verify the testcase for linear and cubic today, I got a lot of cases that the expected values doesn't match with the values returned from Tensorflow.
> Then I switch to look at the test code you refer https://gist.github.com/daquexian/4bab98a7e706ad95616567433c1234b8 yesterday. I try to run downsample_scales_linear in that code by changing h=1, w=2 and resize='linear' and I got matching result from expected and tensorflow values. I also try to run upsample_sizes_cubic and downsample_sizes_cubic cases in your test code, both get matching values.
> Then I go back to the onnx unit test for resize in this PR, I notice that not all cubic testcase with scaler=half_pixel have the exclude_outside attribute set. And the ﻿coeffs value are different in the onnx unit test and your test code. May you please look into this and validate the unit test in this PR match with the test code you provide yesterday? Thanks

Sorry for not making myself clear. 

There are many frameworks and they are different implementations on resize op. To be compatible with every framework, some attributes like "exclude_outside" are introduced. TensorFlow expects exclude_outside to be True while PyTorch expects it to be false (when scaler='half_pixel'). I have to cover all these cases in tests. So only some of the tests are for TensorFlow, while others are for other frameworks like PyTorch. As a result, when testing this PR against TensorFlow, not every test can be used directly, one should always set attributes according to what TensorFlow expects.

daquexian:
@winnietsang I have updated this PR and also my gist https://gist.github.com/daquexian/4bab98a7e706ad95616567433c1234b8, now it is fully compatible with TensorFlow.

winnietsang:
> @winnietsang I have updated this PR and also my gist https://gist.github.com/daquexian/4bab98a7e706ad95616567433c1234b8, now it is fully compatible with TensorFlow (and also PyTorch)

@daquexian, thanks for the explanation of the unit test.

daquexian:
@winnietsang Thanks for your careful check! I have updated them

daquexian:
@houseroad @linkerzhang @wschin Is this PR ok to merge? I think I have addressed all comments :)

onnx/defs/tensor/defs.cc
```diff
@@ -1361,41 +1361,103 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(
+            2,
             "scales",
             "The scale array along each dimension. It takes value greater than 0. If it's less than 1,"
             " it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should"
-            " be the same as the rank of input 'X'.",
-            "tensor(float)")
-        .Output(0, "Y", "N-D tensor after resizing", "T")
+            " be the same as the rank of input 'X'. It will be ignored if 'sizes' is specified.",
+            "T3")
```
linkerzhang: put "tensor(float)" here directly.

onnx/defs/tensor/defs.cc
```diff
@@ -1361,41 +1361,103 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(
+            3,
+            "sizes",
+            "The size of the output tensor. The number of elements of 'sizes' should be the same as the"
+            " rank of input 'X'. 'scales' will be ignored if 'sizes' is specified.",
+            "T4",
```
linkerzhang: put "tensor(int64)" here directly.

onnx/defs/tensor/defs.cc
daquexian: Done :)

onnx/defs/tensor/defs.cc
daquexian: Done :)

daquexian:
@linkerzhang Thanks. It has been updated

onnx/defs/tensor/defs.cc
```diff
@@ -1395,40 +1395,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "Three interpolation modes: nearest (default), linear and cubic. "
+            "The \"linear\" mode includes linear interpolation for 1D tensor and N-D linear interpolation for N-D tensor (For example, bilinear interpolation for 2D tensor). "
+            "The \"cubic\" mode includes cubic interpolation for 1D tensor and N-D cubic interpolation for N-D tensor (For example, bicubic interpolation for 2D tensor)."
+            ,
             AttributeProto::STRING,
             std::string("nearest"))
-        .Input(0, "X", "N-D tensor", "T")
+        .Attr("cubic_coeff_a",
+            "The coefficient 'a' used in cubic interpolation. Two common choice are -0.5 (in TensorFlow) and -0.75"
+            " (in PyTorch). Check out https://ieeexplore.ieee.org/document/1163711 for the details.",
```
wschin: Do you mean `a` in the Equation (4) in the reference paper?

onnx/defs/tensor/defs.cc
```diff
@@ -1395,40 +1395,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
     Resize,
-    10,
+    11,
     OpSchema()
         .Attr(
             "mode",
-            "Two interpolation modes: nearest (default), and linear (including bilinear, trilinear, etc)",
+            "Three interpolation modes: nearest (default), linear and cubic. "
+            "The \"linear\" mode includes linear interpolation for 1D tensor and N-D linear interpolation for N-D tensor (For example, bilinear interpolation for 2D tensor). "
+            "The \"cubic\" mode includes cubic interpolation for 1D tensor and N-D cubic interpolation for N-D tensor (For example, bicubic interpolation for 2D tensor)."
```
wschin: I can't figure out how to implement this operator after reading this spec. Are you using this equation (copied from your reference paper)?
image

onnx/defs/tensor/defs.cc
```diff
@@ -1395,40 +1395,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             AttributeProto::STRING,
             std::string("nearest"))
-        .Input(0, "X", "N-D tensor", "T")
+        .Attr("cubic_coeff_a",
+            "The coefficient 'a' used in cubic interpolation. Two common choice are -0.5 (in TensorFlow) and -0.75"
+            " (in PyTorch). Check out https://ieeexplore.ieee.org/document/1163711 for the details.",
+            AttributeProto::FLOAT,
+            static_cast<float>(-0.75))
+        .Attr("exclude_outside",
+            "If set to 1, the weight of sampling locations outside the tensor will be set to 0"
```
wschin: What is `weight`? Without a equation like 
image
, it's not easy to explain those attributes in a way that the reader can implement this operator.

onnx/defs/tensor/defs.cc
```diff
@@ -1395,40 +1395,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            " and the weight will be renormalized so that their sum is 1.0. The default value is 0.",
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
+        .Attr("scaler",
+            "The scaler used in interpolation. "
+            "For each dimension, denote x_resized as the coordinate in the resized tensor, x_original as the corresponding coordinate in the original tensor, length_original as the length of the original tensor in the specific dimension, length_resized as the length of the resized tensor in the specific dimension, roi_x = (start, end) of the corresponding dimension in input \"roi\", scale = length_resized / length_original"
```
wschin: length_original as the length of the original tensor in the `specific dimension`? If it's for a specific dimension, would it be clearer to use length_original_0 for axis 0 and so on?

onnx/defs/tensor/defs.cc
daquexian: Yes :)

onnx/defs/tensor/defs.cc
daquexian: N-D (2-D is a special case where N=2) interpolation is implemented by performing 1-D interpolation recursively. For N=2, assume we want to get the value at `(1.6, 2.8)` in resized 2-D HxW tensor, we firstly calculate H values at `(1.6)` by performing 1-D interpolation on H rows (a.k.a. H 1-D tensors) respectively, and then calculate the value at `(2.8)` of the generate line (a.k.a. 1-D tensor) consisting of H above values. Similarly, 3-D interpolation is 2-D interpolation on every layer (a.k.a. every 2-D tensor) and 1-D interpolation on the generated line (a.k.a. 1-D tensor).

onnx/defs/tensor/defs.cc
```diff
@@ -1395,40 +1395,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "if scaler is \"half_pixel\", "
+            "x_original = (x_resized + 0.5) / scale - 0.5, "
+            "if scaler is \"pytorch_half_pixel\", "
```
wschin: Could we break this long sentence into smaller ones?

onnx/defs/tensor/defs.cc
```diff
@@ -1395,40 +1395,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "The coefficient 'a' used in cubic interpolation. Two common choice are -0.5 (in TensorFlow) and -0.75"
+            " (in PyTorch). Check out https://ieeexplore.ieee.org/document/1163711 for the details.",
+            AttributeProto::FLOAT,
+            static_cast<float>(-0.75))
+        .Attr("exclude_outside",
+            "If set to 1, the weight of sampling locations outside the tensor will be set to 0"
+            " and the weight will be renormalized so that their sum is 1.0. The default value is 0.",
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
+        .Attr("scaler",
```
```suggestion
        .Attr("coordinate_transformation_mode",
```
maybe?

onnx/defs/tensor/defs.cc
daquexian: Given the context that N-D interpolation is indeed a combination of 1-D interpolation, symbols here can cover all axes. I'll try to make this paragraph more clear.

onnx/defs/tensor/defs.cc
```diff
@@ -1395,40 +1395,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "roi",
+            "1-D tensor given as [start1, ..., startN, end1, ..., endN], where N is the rank of X. The RoIs' coordinates are normalized in the coordinate system of the input image. It only takes effect when scaler is \"tf_crop_and_resize\"",
+            "T2")
+        .Input(
+            2,
             "scales",
             "The scale array along each dimension. It takes value greater than 0. If it's less than 1,"
             " it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should"
-            " be the same as the rank of input 'X'.",
+            " be the same as the rank of input 'X'. It will be ignored if 'sizes' is specified.",
```
```suggestion
            " be the same as the rank of input 'X'. Only one of 'scales' and 'sizes' can be specified. If 'size' is needed, the user can use an empty string as the name of 'scales' in this operator's input list.",
```

onnx/defs/tensor/defs.cc
```diff
@@ -1395,40 +1395,94 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "tensor(float)")
-        .Output(0, "Y", "N-D tensor after resizing", "T")
+        .Input(
+            3,
+            "sizes",
+            "The size of the output tensor. The number of elements of 'sizes' should be the same as the"
+            " rank of input 'X'. 'scales' will be ignored if 'sizes' is specified.",
```
```suggestion
            " rank of input 'X'. Only one of 'scales' and 'sizes' can be specified.",
```

onnx/defs/tensor/defs.cc
daquexian: Thanks! I will do it

onnx/backend/test/case/node/resize.py
```diff
@@ -8,15 +8,233 @@
...
+                          x,                         # type: float
+                          get_coeffs,                # type: Callable[[float], np.ndarray]
+                          roi=None,                  # type: np.ndarray
+                          extrapolation_value=0.0,   # type: float
+                          scaler='half_pixel',       # type: Text
+                          exclude_outside=False,     # type: bool
+                          ):                         # type: (...) -> np.ndarray
+    def get_neighbor_idxes(x, n, limit):  # type: (float, int, int) -> np.ndarray
+        """
+        Return the n nearest indexes, prefer the indexes smaller than x
```
wschin:
```suggestion
        Return the n nearest indexes to x among [0, limit), prefer the indexes smaller than x.
```

onnx/defs/tensor/defs.cc
daquexian: Looks much better than the old one. I'll update it.

onnx/backend/test/case/node/resize.py
```diff
@@ -8,15 +8,233 @@
...
+        idxes = sorted(range(limit), key=lambda idx: (abs(x - idx), idx))[:n]
+        idxes = sorted(idxes)
+        return np.array(idxes)
+
+    def get_neighbor(x, n, data):  # type: (float, int, np.ndarray) -> np.ndarray
+        """
+        Pad `data` in 'edge' mode, and get n nearest elements in the padded array and their indexes in the original
+        array
+        :param x:
+        :param n:  the number of the wanted elements
```
wschin:
```suggestion
        :param n: the expected element number in the padded x.
```

onnx/backend/test/case/node/resize.py
```diff
@@ -8,15 +8,233 @@
...
+        """
+        idxes = sorted(range(limit), key=lambda idx: (abs(x - idx), idx))[:n]
+        idxes = sorted(idxes)
+        return np.array(idxes)
+
+    def get_neighbor(x, n, data):  # type: (float, int, np.ndarray) -> np.ndarray
+        """
+        Pad `data` in 'edge' mode, and get n nearest elements in the padded array and their indexes in the original
+        array
+        :param x:
```
wschin:
```suggestion
        :param x:
```
```suggestion
        :param x: vector to be padded.
```

onnx/backend/test/case/node/resize.py
```diff
@@ -8,15 +8,233 @@
...
+        :param n:  the number of the wanted elements
+        :param data: the array
+        :return: A tuple containing the indexes of neighbor elements (the index can be smaller than 0 or higher than
+        len(data)) and the value of these elements
+        """
+        pad_width = np.ceil(n / 2).astype(np.int)
+        padded = np.pad(data, pad_width, mode='edge')
+        x += pad_width
+
+        idxes = get_neighbor_idxes(x, n, len(padded))
```
wschin: What if `x > len(padded)` or `x + n/2 > len(padded)`?

onnx/backend/test/case/node/resize.py
```diff
@@ -8,15 +8,233 @@
...
+
+    idxes, points = get_neighbor(x_ori, n, data)
+
+    if exclude_outside:
+        for i, idx in enumerate(idxes):
+            if idx < 0 or idx >= input_width:
+                coeffs[i] = 0
+        coeffs /= sum(coeffs)
+
+    return np.dot(coeffs, points).item()
```
wschin: Your code looks very inspiring. It basically decompose Resize into several abstractive steps:

- roi selection
- coordinate transformation
- neighbor selection
- dot product (interpolation)
It's not a blocking comment but I am wondering if we can further decompose Resize in the future.

onnx/defs/tensor/defs.cc
wschin: Sounds good. Let's me more specific,
```suggestion
            " (in PyTorch). Check out Equation (4) in https://ieeexplore.ieee.org/document/1163711 for the details.",
```

onnx/backend/test/case/node/resize.py
daquexian: it will not happen, after `x += pad_width`, x is in [n/2, len(data) - 1 + n/2], while len(padded) is len(data) + n

onnx/defs/tensor/defs.cc
daquexian: I have added some description in 1c87d37. Is it enough? Thanks!

onnx/defs/tensor/defs.cc
daquexian: I have added some description in ff74643

daquexian:
@wschin thanks for your valuable comments. I have updated this pr accordingly

onnx/defs/tensor/defs.cc
```diff
@@ -1602,40 +1602,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-  output_dimension = floor(input_dimension * scale).
+  output_dimension = floor(input_dimension * scale) if input \"sizes\" is not specified.
 )DOC";
 
+static const char* Resize_attr_coordinate_transformation_mode_doc = R"DOC(
+This attribute describes how to transform the coordinate in the resized tensor to the coordinate in the original tensor. <br/>
+
+The coordinate of each dimension is transformed individually. Denote x_resized as the coordinate of a specific dimension in the resized tensor, x_original as the coordinate of this dimension in the original tensor, length_original as the length of the original tensor in this dimension, length_resized as the length of the resized tensor in this dimension, roi_x = (start, end) of the this dimension in input "roi", scale = length_resized / length_original, <br/>
+
+if scaler is "half_pixel", <br/>
```
```suggestion
if coordinate_transformation_mode is "half_pixel", <br/>
```
and also for the following places.

onnx/defs/tensor/defs.cc
```diff
@@ -1602,40 +1602,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+Resize the input tensor. In general, it calculates every value in the output tensor as a weighted average of neighborhood (a.k.a. sampling locations) in the input tensor.
 Each dimension value of the output tensor is:
-  output_dimension = floor(input_dimension * scale).
+  output_dimension = floor(input_dimension * scale) if input \"sizes\" is not specified.
 )DOC";
 
+static const char* Resize_attr_coordinate_transformation_mode_doc = R"DOC(
+This attribute describes how to transform the coordinate in the resized tensor to the coordinate in the original tensor. <br/>
+
+The coordinate of each dimension is transformed individually. Denote x_resized as the coordinate of a specific dimension in the resized tensor, x_original as the coordinate of this dimension in the original tensor, length_original as the length of the original tensor in this dimension, length_resized as the length of the resized tensor in this dimension, roi_x = (start, end) of the this dimension in input "roi", scale = length_resized / length_original, <br/>
```
```suggestion
The coordinate of each dimension is transformed individually. Let's describe a case using axis x as an example. 
Denote x_resized as the coordinate of a specific dimension in the resized tensor, x_original as the coordinate of this dimension in the original tensor, length_original as the length of the original tensor in this dimension, length_resized as the length of the resized tensor in this dimension, roi_x = (startx, endx) of the this dimension in input "roi", scale = length_resized / length_original, <br/>
```

docs/Changelog.md
```diff
@@ -10746,6 +10746,83 @@ This version of the operator has been available since version 11 of the default
...
+x_original = x_resized * (length_original - 1) / (length_resized - 1), <br/>
+
+if scaler is "asymmetric", <br/>
+x_original = x_resized / scale, <br/>
+
+if scaler is "tf_half_pixel_for_nn", <br/>
+x_original = (x_resized + 0.5) / scale, <br/>
+
+if scaler is "tf_crop_and_resize", <br/>
+x_original = length_resized > 1 ? roi_x[0] * (length_original - 1) + x_resized * (roi_x[1] - roi_x[0]) * (length_original - 1) / (length_resized - 1) : 0.5 * (roi_x[0] + roi_x[1]) * (length_original - 1).</dd>
```
```suggestion
x_original = length_resized > 1 ? roi_x[0] * (length_original - 1) + x_resized * (endx[1] - startx[0]) * (length_original - 1) / (length_resized - 1) : 0.5 * (endx[0] + startx[1]) * (length_original - 1).</dd>
```

onnx/defs/tensor/defs.cc
```diff
@@ -1602,40 +1602,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+x_original = x_resized * (length_original - 1) / (length_resized - 1), <br/>
+
+if scaler is "asymmetric", <br/>
+x_original = x_resized / scale, <br/>
+
+if scaler is "tf_half_pixel_for_nn", <br/>
+x_original = (x_resized + 0.5) / scale, <br/>
+
+if scaler is "tf_crop_and_resize", <br/>
+x_original = length_resized > 1 ? roi_x[0] * (length_original - 1) + x_resized * (roi_x[1] - roi_x[0]) * (length_original - 1) / (length_resized - 1) : 0.5 * (roi_x[0] + roi_x[1]) * (length_original - 1).)DOC";
```
```suggestion
x_original = length_resized > 1 ? roi_x[0] * (length_original - 1) + x_resized * (endx[1] - startx[0]) * (length_original - 1) / (length_resized - 1) : 0.5 * (endx[0] + startx[1]) * (length_original - 1).)DOC";
```

onnx/defs/tensor/defs.cc
```diff
@@ -1602,40 +1602,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         .SetDoc(Upsample_ver10_doc)
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
-          resizeShapeInference(ctx);
+          resizeShapeInference(ctx, false);
         }));
 
-static const char* Resize_ver10_doc = R"DOC(
-Resize the input tensor.
+static const char* Resize_ver11_doc = R"DOC(
+Resize the input tensor. In general, it calculates every value in the output tensor as a weighted average of neighborhood (a.k.a. sampling locations) in the input tensor.
```
```suggestion
Resize the input tensor. In general, it calculates every value in the output tensor as a weighted average of neighborhood (a.k.a. sampled locations) in the input tensor.
```

onnx/defs/tensor/defs.cc
wschin: I really like your detailed documentation. It truly standardize something.

onnx/defs/tensor/defs.cc
```diff
@@ -1602,40 +1602,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 ONNX_OPERATOR_SET_SCHEMA(
     Resize,
-    10,
+    11,
     OpSchema()
         .Attr(
             "mode",
-            "Two interpolation modes: nearest (default), and linear (including bilinear, trilinear, etc)",
+            "Three interpolation modes: nearest (default), linear and cubic. "
+            "The \"linear\" mode includes linear interpolation for 1D tensor and N-D linear interpolation for N-D tensor (For example, bilinear interpolation for 2D tensor). "
```
```suggestion
            "The \"linear\" mode includes linear interpolation for 1D tensor and N-D linear interpolation for N-D tensor (for example, bilinear interpolation for 2D tensor). "
```

onnx/defs/tensor/defs.cc
```diff
@@ -1602,40 +1602,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "The \"cubic\" mode includes cubic interpolation for 1D tensor and N-D cubic interpolation for N-D tensor (For example, bicubic interpolation for 2D tensor)."
```
```suggestion
            "The \"cubic\" mode includes cubic interpolation for 1D tensor and N-D cubic interpolation for N-D tensor (for example, bicubic interpolation for 2D tensor)."
```

onnx/defs/tensor/defs.cc
wschin: Looks good to me. Thank you.

onnx/defs/tensor/defs.cc
```diff
@@ -1602,40 +1602,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Attr("coordinate_transformation_mode",
+            Resize_attr_coordinate_transformation_mode_doc,
+            AttributeProto::STRING,
+            std::string("half_pixel"))
+        .Attr("nearest_mode",
+            "Four modes: round_prefer_floor (default), round_prefer_ceil, floor, ceil. Only used by nearest interpolation. It indicates how to get \"nearest\" pixel in input tensor from x_original.",
+            AttributeProto::STRING,
+            std::string("round_prefer_floor"))
+        .Attr("extrapolation_value",
+            "When scaler is \"tf_crop_and_resize\" and x_original is outside the range [0, length_original - 1], this value is used as the corresponding output value. Default is 0.0f.",
```
```suggestion
            "When scaler is \"tf_crop_and_resize\" and x_original is outside the range [0, length_original - 1], this value is used as the corresponding output value. Default is 0.0f. This attribute is valid only if \"exclude_outside\" is 0.",
```
Is my understanding correct? If so, could you please add a check in your shape inference code? Thank you.

onnx/defs/tensor/defs.cc
```diff
@@ -1602,40 +1602,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "If set to 1, the weight of sampling locations outside the tensor will be set to 0"
+            " and the weight will be renormalized so that their sum is 1.0. The default value is 0.",
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
+        .Attr("coordinate_transformation_mode",
+            Resize_attr_coordinate_transformation_mode_doc,
+            AttributeProto::STRING,
+            std::string("half_pixel"))
+        .Attr("nearest_mode",
+            "Four modes: round_prefer_floor (default), round_prefer_ceil, floor, ceil. Only used by nearest interpolation. It indicates how to get \"nearest\" pixel in input tensor from x_original.",
```
```suggestion
            "Four modes: round_prefer_floor (default), round_prefer_ceil, floor, ceil. Only used by nearest interpolation. It indicates how to get \"nearest\" pixel in input tensor from x_original, so this attribute is valid only if \"mode\" is \"nearest\".",
```

onnx/defs/tensor/defs.cc
```suggestion
            "When \"coordinate_transform_mode\" is \"tf_crop_and_resize\" and x_original is outside the range [0, length_original - 1], this value is used as the corresponding output value. Default is 0.0f.",
```

lara-hdr:
Thank you @daquexian for the updates.
The CI is failing with the error "Unrecognized attribute: scaler for operator Resize" when checking the models, since the attribute was renamed.
Could you regenerate them to fix the issue?

onnx/defs/tensor/defs.cc
daquexian: Great, your help is precious!

onnx/defs/tensor/defs.cc
daquexian: "sampling locations" is from https://github.com/tensorflow/tensorflow/blob/3ae2c6691b7c6e0986d97b150c9283e5cc52c15f/tensorflow/core/kernels/resize_bicubic_op.cc#L98. Do you think "sampled locations" is natural?

onnx/defs/tensor/defs.cc
daquexian: It is not so. "exclude_outside" is about the neighbors outside of the input tensor in the weighted average, "extrapolation_value" is about the "x_original".

onnx/defs/tensor/defs.cc
daquexian: Thanks! I have updated it.

onnx/defs/tensor/defs.cc
daquexian: Thanks! They are now replaced in 66d2003

onnx/defs/tensor/defs.cc
daquexian: Thanks! It is now updated in b067855

docs/Changelog.md
daquexian: Updated in b067855, thanks!

daquexian:
@wschin @lara-hdr I have updated this PR :) Travis CI failed due to network error

PR #2060, opened 2019-05-28
Revamped test cases for Gemm

JamesAllingham:
Hi all,

I'm currently working on an ONNX importer for the [neural network's framework](https://reference.wolfram.com/language/guide/NeuralNetworks.html) in Wolfram Mathematica. While developing I've found it useful to add more test cases to make sure that I'm following the ONNX spec properly, so I figured I should contribute them back upstream so that others can also use them. I'll definitely be adding more in the near future, I just wanted to get my hands dirty with this to make sure I'm doing things correctly. So assuming that these changes are welcome I'll be adding more soon. :)

This is my first time contributing to the project, so please do let me know if I'm doing anything wrong! After adding the test cases, I re-installed ONNX, ran the test suite, as well as the type checker, and didn't see any issues. I also made linted the code to make sure it was all looking good. 

Here is what the PR changes: 

1. Increases the number of tests for Gemm from 2 to 8.

2. The tests now each test a specific attribute or input case.

3. Adds test data.

4. Updates the operators doc.

5. Updates the test coverage.

Regarding point 2, I wanted to make it easier for myself, and other developers, to incrementally add support for the attributes and other features of the Gemm operator. By splitting the tests up so that each one only covers a particular attribute or input case I think this goal is accomplished. It also makes it immediately clear what has broken when a test case regression occurs. 

One thing I was a little worried about when making this change was that it would not be allowed to remove existing test cases. If this is the case I'll add the old cases back -- I simply removed them because they were a bit redundant. I also imagine there might be some feedback about how I've named the test cases.


linkerzhang:
@JamesAllingham welcome and thank you for the contribution to ONNX. It's ok to remove some test cases as long as the scenarios/cases are also covered by new ones.

Please fix the CI failures:

+flake8
./onnx/backend/test/case/node/gemm.py:24:24: E231 missing whitespace after ','
./onnx/backend/test/case/node/gemm.py:72:29: E261 at least two spaces before inline comment
./onnx/backend/test/case/node/gemm.py:81:24: E231 missing whitespace after ','
./onnx/backend/test/case/node/gemm.py:87:29: E261 at least two spaces before inline comment
./onnx/backend/test/case/node/gemm.py:96:24: E231 missing whitespace after ','
./onnx/backend/test/case/node/gemm.py:111:24: E231 missing whitespace after ','

JamesAllingham:
@linkerzhang Thanks for the welcome. I'm very keen to be getting involved with this project. And thanks for the feedback. 

What my test cases don't cover (that the old ones did) is whether the `transA`, `transB`, `alpha` and `beta` parameters work as expected when used together. However, that seems like a bit of an edge case, and to properly test all interactions like that for every op would require a huge number of tests. That said, I suppose there is nothing wrong with having a few of these sorts of test cases. Let me know what you think. 

BTW, how did you see the reason for the CI failures? When I clicked on the "Details" link for the failing build I couldn't find that information. I think I've fixed the issue now, but for future, it would be useful to know how to get that information. I assume I am missing something obvious here.

JamesAllingham:
Any feedback on this? Would be great to get this closed so that I can go ahead with some more PRs (would be nice to know I'm doing things correctly before doing more 😃)

JamesAllingham:
Hey @linkerzhang any feedback on this? I also have another [PR](https://github.com/onnx/onnx/pull/2127) which could use some love. I have quite a few more tests I'd like to submit PRs for but I really would like to get some feedback on what I'm doing before submitting!

JamesAllingham:
@postrational I've updated the branch. Let me know if I need to do anything else to get this merged.

docs/Operators.md
```diff
@@ -5566,46 +5566,164 @@ Other versions of this operator: <a href="Changelog.md#Gemm-1">Gemm-1</a>, <a hr
...
+node = onnx.helper.make_node(
+    'Gemm',
+    inputs=['a', 'b', 'c'],
+    outputs=['y'],
+    alpha=0.5
+)
+a = np.random.ranf([3, 5]).astype(np.float32)
+b = np.random.ranf([5, 4]).astype(np.float32)
+c = np.zeros([1, 4]).astype(np.float32)
+y = 0.5 * np.dot(a, b) + c
```
wschin: [nit] This is good enough for tests. On the other hand, it'd be nice if you can create a reference implementation of Gemm using numpy and use it in all tests. For example,
```
y = your_gemm_reference_impl(a, b, c, ...)
```

docs/Operators.md
```diff
@@ -5566,46 +5566,164 @@ Other versions of this operator: <a href="Changelog.md#Gemm-1">Gemm-1</a>, <a hr
...
+node = onnx.helper.make_node(
+    'Gemm',
+    inputs=['a', 'b', 'c'],
+    outputs=['y']
+)
+a = np.random.ranf([3, 5]).astype(np.float32)
+b = np.random.ranf([5, 4]).astype(np.float32)
+c = np.zeros([1, 4]).astype(np.float32)
+y = np.dot(a, b) + c
+expect(node, inputs=[a, b, c], outputs=[y],
```
wschin: No bias? Do you mean that `c` will be removed?

docs/Operators.md
```diff
@@ -5566,46 +5566,164 @@ Other versions of this operator: <a href="Changelog.md#Gemm-1">Gemm-1</a>, <a hr
...
+
+```python
+node = onnx.helper.make_node(
+    'Gemm',
+    inputs=['a', 'b', 'c'],
+    outputs=['y']
+)
+a = np.random.ranf([2, 3]).astype(np.float32)
+b = np.random.ranf([3, 4]).astype(np.float32)
+c = np.random.ranf([1]).astype(np.float32)
```
wschin: `[1]` is not a scalar. It's a 1-element vector. Maybe we can use `np.array(1)` whose shape is `()`. As mentioned in my comment above, it'd even nicer if we have a reference implementation.

docs/Operators.md
```diff
@@ -5566,46 +5566,164 @@ Other versions of this operator: <a href="Changelog.md#Gemm-1">Gemm-1</a>, <a hr
...
+
+<details>
+<summary>transposeA</summary>
 
 ```python
 node = onnx.helper.make_node(
     'Gemm',
     inputs=['a', 'b', 'c'],
     outputs=['y'],
-    alpha=0.5,
```
wschin: May we keep this one? It represents a case where all attributes are non-default.

docs/Operators.md
JamesAllingham: What I meant was that the bias term has no impact on the result. I think this is a useful test case because in many frameworks the bias term of the fully-connected layer is optional with a default of 0. I'll rename the test to make this more clear. 

With that in mind, do you think it is worth changing the spec for `Gemm` so that the bias is optional and has a default of 0? I'm happy to do so, and I don't think it is too much work or too big a change. (Then I can add a test which legitimately has no bias term).

docs/Operators.md
JamesAllingham: Agreed and done 👍

docs/Operators.md
JamesAllingham: You are totally correct. I've fixed this. And I am going to do a reference implementation.

docs/Operators.md
JamesAllingham: I agree, it will be good to have a reference implementation, I'll add one. 👍

docs/Operators.md
JamesAllingham: (I renamed this test, it is now called 'test_gemm_all_attributes')

JamesAllingham:
> 
> 
> This PR looks very good. Could you please take a look at my minor comments and see if we can address them? Thank you! To me more clear, I think this PR is almost ready to be merged.

Cool, I think I've addressed all of your comments, please let me know if I misunderstood anything. Also, let me know what you think about allowing the bias input (`'C'`) to be optional with a default of 0.

docs/Operators.md
wschin: It totally makes sense to me. Could we do it in another PR? I will try to get it reviewed (and then hopefully merged) before the deadline (9/19 is the targeted release date)?

docs/Operators.md
```diff
@@ -5680,40 +5704,40 @@ expect(node, inputs=[a, b, c], outputs=[y],
...
     'Gemm',
     inputs=['a', 'b', 'c'],
     outputs=['y']
 )
-a = np.random.ranf([3, 5]).astype(np.float32)
-b = np.random.ranf([5, 4]).astype(np.float32)
-c = np.zeros([1, 4]).astype(np.float32)
+a = np.random.ranf([2, 3]).astype(np.float32)
+b = np.random.ranf([3, 4]).astype(np.float32)
+c = np.random.ranf(1).astype(np.float32)
```
wschin: This is still not a scalar. 
```
>>> np.random.ranf(1).shape
(1,)
```
As you already have a reference implementation, could we do
```
c= 3.14
```
?

docs/TestCoverage.md
```diff
@@ -3098,38 +3120,38 @@ expect(node, inputs=[a, b, c], outputs=[y],
...
     'Gemm',
     inputs=['a', 'b', 'c'],
     outputs=['y']
 )
-a = np.random.ranf([3, 5]).astype(np.float32)
-b = np.random.ranf([5, 4]).astype(np.float32)
-c = np.zeros([1, 4]).astype(np.float32)
+a = np.random.ranf([2, 3]).astype(np.float32)
+b = np.random.ranf([3, 4]).astype(np.float32)
+c = np.random.ranf(1).astype(np.float32)
```
wschin: Humm, it's not a scalar.

onnx/backend/test/case/node/gemm.py
```diff
@@ -10,38 +10,162 @@
 from . import expect
 
 
+def gemm_reference_implementation(A, B, C, alpha=1., beta=1., transA=0,
+                                  transB=0):  # type: (np.ndarray, np.ndarray, np.ndarray, float, float, int, int) -> np.ndarray
+    A = A if transA == 0 else A.T
+    B = B if transB == 0 else B.T
+
+    Y = alpha * np.dot(A, B) + beta * C
```
wschin: Nice!

docs/Operators.md
JamesAllingham: Sure, I'll make another PR for it. I'm afraid it will have to be about 11 hours from now though!

docs/Operators.md
JamesAllingham: Whoops! Let me fix that!

JamesAllingham: Okay, I think I've fixed this for real now, sorry about that :)

JamesAllingham:
> 
> 
> Overall looks good and it improves the document considerablely. Just last few comments to address.

One unfortunate thing about the documentation is that the reference implementation is not shown in `Operators.md`, I don't suppose there is an easy fix for this?

PR #2077, opened 2019-06-03
Add shape inference for LinearClassifier

hariharans29:
cc: @stevenlix


onnx/defs/traditionalml/defs.cc
```diff
@@ -505,15 +515,74 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
             std::string("NONE"))
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           std::vector<std::string> label_strs;
-          auto result =
+          std::vector<int64_t> label_ints;
```
hariharans29: Actual change begins here

onnx/defs/traditionalml/defs.cc
```diff
@@ -505,15 +515,74 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+              // if input_rank is 1, batch_size is interpreted to be 1
+              output_shape_0->mutable_dim((int)0)->set_dim_value(1);
+              output_shape_1->mutable_dim((int)0)->set_dim_value(1);
+            } else if (input_rank == 2) {
+              *output_shape_0->mutable_dim((int)0) = input_shape.dim((int)0);
+              *output_shape_1->mutable_dim((int)0) = input_shape.dim((int)0);
+            } else {
+              fail_shape_inference("Input's shape should be 1D or 2D");
+            }
+          }
```
hariharans29: Actual change ends here

onnx/defs/traditionalml/defs.cc
```diff
@@ -505,15 +515,72 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+          std::vector<float> intercepts;
+          getRepeatedAttribute(ctx, "intercepts", intercepts);
+          int class_count = static_cast<int>(intercepts.size());
+          if (intercepts.size() == 1 &&
+              ((using_strings && label_strs.size() == 2) ||
+               (!using_strings && label_ints.size() == 2))) {
+            class_count = 2;
+          }
+
+          auto* output_shape_0 =
```
gramalingam: This may be a subjective/personal preference, but I feel the following lines could be replaced as below. I perfonally find this a bit easier to read and check:
```cpp
    TensorShapeProto::Dimension batch_size_dim, class_count_dim;
    class_count_dim->set_dim_value(class_count);
    if (hasNInputShapes(ctx, 1)) {
        const auto& input_shape = ctx.getInputType(0)->tensor_type().shape();
        const auto input_rank = input_shape.dim_size();
        if (input_rank == 1) {
          // if input_rank is 1, batch_size is interpreted to be 1
          batch_size_dim->set_dim_value(1);
        } else if (input_rank == 2) {
          batch_size_dim = input_shape.dim((int)0);
        } else {
          fail_shape_inference("Input's shape should be 1D or 2D");
        }
    }
    updateOutputShape(ctx, 0, {batch_size_dim});
    updateOutputShape(ctx, 1, {batch_size_dim, class_count});
}
```

hariharans29: Hi @gramalingam - that does look better. Thanks for your feedback. I will incorporate it.

onnx/test/shape_inference_test.py
```diff
@@ -1920,6 +1920,30 @@ def test_tile_rank_inference(self):  # type: () -> None
             [])
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (None, None, None))])  # type: ignore
 
+    def test_linearclassifier_1D_input(self):  # type: () -> None
+        onnx_ml = os.environ.get('ONNX_ML')  # type: ignore
```
hariharans29: is this logic to determine whether to run the test or not okay with you @KeDengMS @houseroad @gramalingam ?

(We need to skip this test for ONNX_ML=0 builds as this test is for a traditional ML op)

houseroad: A better way to check it is creating a helper method using the #ifdef ONNX_ML to generate the body of the helper function, which returns whether it's build with ONNX_ML or not.

hariharans29: Can you please point me to which file you think this helper can go into ? A similar example method will be useful...

hariharans29: For now, I prefer to keep this logic simple and as it is. This is not too dissimilar to the logic in `gen_doc.py` where the environment variable is queried to determine which file to generate he documentation for. If this is okay for now, can I merge this PR ? @linkerzhang @houseroad

onnx/defs/traditionalml/defs.cc
```diff
@@ -424,19 +430,20 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
-              ctx, "values_floats", values_floats);
+          bool values_floats_result =
+              getRepeatedAttribute(ctx, "values_floats", values_floats);
 
           // Check if only one values_* attribute is set.
           if (static_cast<int>(values_strings_result) +
                   static_cast<int>(values_int64s_result) +
-                  static_cast<int>(values_floats_result) != 1)
+                  static_cast<int>(values_floats_result) !=
+              1)
```
linkerzhang: this is weird format. did you run "clang-format ..."

hariharans29: Yes, I did. That's why we see a lot of formatting changes in this file.

linkerzhang:
The pytorch failure should not be related with this change. Will merge this PR end of day today if no concern. @houseroad @spandantiwari  Thanks!

PR #2080, opened 2019-06-05
Enforce non-negative lambd attribute values of the Shrink operator

tomdol:
I'd like to propose a slight change of accepted values for the `lambd` attribute of the Shrink operator.

While implementing this op I started creating test cases and came up with a set of attributes for which I could not produce non-ambiguous values of Shrink.

```
 L = lambd
 B = bias
 input = [0;5]

       {x < -L} => x + B 
      /
shrink - 0
      \ 
       {x > L}  => x - B

 L B | 0  1 2 3 4 5
 ------------------
 2 2 | 0  0 0 1 2 3
 0 2 | 0 -1 0 1 2 3
-1 2 | ?  ? 0 1 2 3
```

For positive lambd values, the range against which input values should be checked is `[-lambd;lambd]` for example `1 => [-1;1]`. For negative values, this range becomes inverted `[1;-1]` and some input values fall into multiple branches of the `if`.

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -12598,7 +12598,7 @@ This version of the operator has been available since version 9 of the default O
 <dt><tt>bias</tt> : float (default is 0.0)</dt>
 <dd>The bias value added to output. Default is 0.</dd>
 <dt><tt>lambd</tt> : float (default is 0.5)</dt>
-<dd>The lambd value for the Shrink formulation. Default is 0.5.</dd>
+<dd>The non-negative lambd value for the Shrink formulation. Default is 0.5.</dd>
 </dl>
 
 #### Inputs


wschin:
Maybe we can check if `lambda >= 0` in its shape inference function.

linkerzhang:
Thank you! This is a nice clarification!

Please don't edit the Operators.md directly. This file is auto-generated. See, https://github.com/onnx/onnx/blob/master/docs/CONTRIBUTING.md#generated-operator-documentation.

To update the doc, the related defs.cc should be changed and doc-generation should be run to refresh the docs.

Here's a doc on how to add a new op, https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md, but it also tells how to update the doc, I think.

tomdol:
Thanks, I will rever this commit and add the changes it in the proper location.

wschin:
@tomdol, could you please sync this PR with the master branch?

tomdol:
@wschin since I introduced changes in a totally wrong location, I think I would rather close this PR and open a new one when I have changes ready. Unfortunately I didn't have time to take care of it before.

Would this be ok?

wschin:
It's ok. You can do it in this branch and then do `git push -f`.

tomdol:
Closing for now. Will get back to this later and implement it properly.

PR #2084, opened 2019-06-06
Fix shape inference error in Resize op

lara-hdr:
(scales->float_data_size() == input_shape.dim_size()) is not giving right results, and failing when it should not.

Reverting the code to match Upsample when accessing the data in "scales".

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -1430,28 +1430,30 @@ ONNX_OPERATOR_SET_SCHEMA(
           auto scales = ctx.getInputData(1);
           if (nullptr != scales) {
             // Infer output shape's dimension value if 'scales' is known.
-            if (scales->data_type() == TensorProto::FLOAT &&
-                scales->float_data_size() == input_shape.dim_size()) {
-              for (int i = 0; i < input_shape.dim_size(); ++i) {
-                int64_t dim_value = static_cast<int64_t>(std::floor(
-                    static_cast<float>(input_shape.dim(i).dim_value()) *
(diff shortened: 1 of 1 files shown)


hariharans29:
CC: @raymondxyang who was also working on this

onnx/defs/tensor/defs.cc
```diff
@@ -1430,28 +1430,30 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-                          dim_value,
-                          ") is not equal to the existing dim value (",
-                          output_shape->dim(i).dim_value(),
-                          ").");
+            if (scales->data_type() == TensorProto::FLOAT){
+              const auto& data = ParseData<float>(scales);
+              if(data.size() == input_shape.dim_size()){
+                for (int i = 0; i < input_shape.dim_size(); ++i) {
+                  int64_t dim_value = static_cast<int64_t>(std::floor(
+                      static_cast<float>(input_shape.dim(i).dim_value()) *
```
hariharans29: Hi @lara-hdr - can you please wait for @raymondxyang 's fix for this ? We have some other fixes like handling symbolic dims (this code doesn't have a check `has_dim_value()`) and while we are at it - we can fix `Upsample` as well which is very similar to this....

lara-hdr: Sounds good. Thanks.

raymondxyang:
Yes I am fixing/unifying the shape inference for upsample/resize.. theres also another problem with the dim_value when handling symbolic shapes. will send a PR likely by today. Thanks @hariharans29 for the help!

PR #2085, opened 2019-06-07
Fix Resize/Upsample Shape inference function

raymondxyang:
* Unify and fix upsample/resize shape inference to handle symbolic cases
* Fix resize's shape inference to take raw_data
* Adding some testcases.


raymondxyang:
@lara-hdr @hariharans29 would you mind helping review it since you are also working on this part? Thanks!

onnx/defs/tensor/defs.cc
```diff
@@ -1325,6 +1325,69 @@ ONNX_OPERATOR_SET_SCHEMA(
           return;
         }));
 
+void resizeShapeInference(InferenceContext& ctx) {
+  if (!hasNInputShapes(ctx, 2)) {
```
hariharans29: I think we need to only check for the first input shape. Even if the second input shape is missing, we can still proceed further and atleast trigger rank inference section. What do you think about this ?

onnx/defs/tensor/old.cc
```diff
@@ -426,40 +426,63 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain input 'X' and output 'Y' to all tensor types.")
         .SetDoc(Upsample_ver9_doc)
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
-          if (!hasNInputShapes(ctx, 1)) {
+          if (!hasNInputShapes(ctx, 2)) {
```
hariharans29: same comment as previously - I think first input shape is good enough to proceed further....

onnx/defs/tensor/old.cc
```diff
@@ -426,40 +426,63 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             return;
           }
           propagateElemTypeFromInputToOutput(ctx, 0, 0);
           auto& input_shape = getInputShape(ctx, 0);
           auto* output_shape = getOutputShape(ctx, 0);
-          output_shape->clear_dim();
           auto scales = ctx.getInputData(1);
+
```
hariharans29: If this is more or less just a copy of the above method - maybe we can just shift it to a util.h under`..defs\tensor\` and reference that method in all places. Any concerns with that approach ?

onnx/defs/tensor/old.cc
```diff
@@ -426,40 +426,63 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          if (output_shape->dim_size() > 0) {
+            if (output_shape->dim_size() != input_shape.dim_size()) {
+              fail_shape_inference(
```
hariharans29: Since we support opset 7 too - I guess this fix needs to be applied to `Upsample (7)` as well. Since Upsample (7) uses attributes to store 'scales ' (scales is not an input), I would make the function handle attribute 'scales' as well and just reference that function in a ll 4 places - Upsample (7), Upsample (9), Upsample (10), Resize (10).

onnx/defs/tensor/old.cc
```diff
@@ -426,40 +426,63 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                "Ranks inferred (",
+                input_shape.dim_size(),
+                ") is not equal to the existing rank value (",
+                output_shape->dim_size(),
+                ").");
```
hariharans29: I would add a few simple tests for Upsample (7), Upsample (9) as well.

hariharans29:
CC: @linkerzhang - we need this fix to unblock a few failing models.

onnx/defs/tensor/old.cc
raymondxyang: Good point. Switch to do so

onnx/defs/tensor/utils.cc
```diff
@@ -0,0 +1,124 @@
...
+    } // input_shape.dim(i).has_dim_value()
+  }
+}
+
+void upsampleShapeInferenceV7(InferenceContext& ctx) {
+  if (!hasNInputShapes(ctx, 1)) {
+    return;
+  }
+  propagateElemTypeFromInputToOutput(ctx, 0, 0);
+  auto& input_shape = getInputShape(ctx, 0);
```
hariharans29: nit: const auto&

onnx/defs/tensor/utils.cc
```diff
@@ -0,0 +1,124 @@
...
+  auto* output_shape = getOutputShape(ctx, 0);
+  auto* scales = ctx.getAttribute("scales");
```
hariharans29: nit: const auto*

onnx/defs/tensor/utils.cc
```diff
@@ -0,0 +1,124 @@
...
+      "Attribute 'scales' is required.");
+  } // nullptr != scales
+}
+
+void resizeShapeInference(InferenceContext& ctx) {
+  if (!hasNInputShapes(ctx, 1)) {
+    return;
+  }
+  propagateElemTypeFromInputToOutput(ctx, 0, 0);
+  auto& input_shape = getInputShape(ctx, 0);
```
hariharans29: same nits as above (optional)

take-cheeze:
Great PR!
I needed this workaround to make it infer: https://github.com/chainer/onnx-chainer/blob/d428f4def37e32ac4af4213c2fc69e72fa75f87d/onnx_chainer/export.py#L48-L52

onnx/defs/tensor/utils.h
```diff
@@ -0,0 +1,14 @@
...
+#pragma once
+
+#include "onnx/defs/schema.h"
+#include "onnx/defs/tensor_proto_util.h"
+
+#include <cmath>
+
+namespace ONNX_NAMESPACE {
+  void resizeShapeInference(InferenceContext& ctx);
+  void upsampleShapeInferenceV7(InferenceContext& ctx);
```
linkerzhang: is this one "upsampleShapeInferenceV7" also needed to be generalized and put in this common file? I'm seeing it's only used once. if it's not needed, let's not put it here,

PR #2042, opened 2019-05-22
Support Unique operator in ONNX

hariharans29:
Copy of #2002

PR #2087, opened 2019-06-07
Support Range operator in ONNX

hariharans29:
This change is to support Range operator in ONNX.


jiafatom:
Looks good to me.

wschin:
At the first glance of seeing the proposed signature, the first thing comes to my mind is
```c
    vector<int> indexes;
    int start = 1;
    int end = 8;
    int step = 2;
    for (int i = start; i < end; i += step)
        indexes.push_back(i);
```
Is it something we can compose using `Loop`?

hariharans29:
> At the first glance of seeing the proposed signature, the first thing comes to my mind is
> 
> ```c
>     vector<int> indexes;
>     int start = 1;
>     int end = 8;
>     int step = 2;
>     for (int i = start; i < end; i += step)
>         indexes.push_back(i);
> ```
> 
> Is it something we can compose using `Loop`?

@jiafatom - do you think this can be constructed using Loop and will there be perf implications if so ?

jiafatom:
> At the first glance of seeing the proposed signature, the first thing comes to my mind is
> 
> ```c
>     vector<int> indexes;
>     int start = 1;
>     int end = 8;
>     int step = 2;
>     for (int i = start; i < end; i += step)
>         indexes.push_back(i);
> ```
> 
> Is it something we can compose using `Loop`?

@wschin It can be composed using `Loop`, but feels not convenient, because we need construct the body graph for `Loop` operator using multiple nodes and logic. This `Range` is already a contrib op in onnxruntime, it is straightforward picking it up in onnx.

jiafatom:
@wschin I also feel that if it is not easily readable if we use `Loop` to compose `Range`. If we just use `Range`, it is easy to understand its meaning. But if we use `Loop`, the actual function is encapsulated in the graph structure of the `Loop`, which is hard to understand if we look at the model graph.

wschin:
@jeffbloo, for both of convince and readability, I prefer to have `Function` instead of an new operator. This is very important to avoid a exponential grow of ONNX operator set. In an ONNX model, we can just use that `Function` as if we have a `Range` operator but runtime can decide if that `Function` will be executed as a sub-graph or an internal magic implementation of that `Function`.

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +530,177 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+  return n;
+}
+
+const std::vector<NodeProto> build_nodes_range_op() {
+    // body for 'Loop node'
+    GraphProto loop_sub_graph;
+
+    // 'Loop' node 'body' attribute's graph inputs
+    auto* input_value_info_proto_0 = loop_sub_graph.add_input();
```
hariharans29: @wschin - any thoughts on this please ?

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,185 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    auto* output_value_info_proto_1 = loop_sub_graph.add_output();
+    output_value_info_proto_1->set_name("current");
+
+    auto* output_value_info_proto_2 = loop_sub_graph.add_output();
+    output_value_info_proto_2->set_name("range");
+
+    return FunctionBodyHelper::BuildNodes({
+      // nodes: {outputs, op, inputs, attributes}
+      {{"sub_result"}, "Sub", {"limit", "start"}},
+      {{"delta_cast"}, "Cast", {"delta"}, {{"to", static_cast<int64_t>(1)}}},
```
hariharans29: The only issue is "delta" is an optional input. But I don't know how to deal with optional inputs in the 'Functions' world. Any pointers @linkerzhang please ?

hariharans29:
CC: @linkerzhang @houseroad for feedback

liqunfu:
Thanks for this new op and for making it a function op. Range op is needed by PyTorch as well.

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,185 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+inline int compute_output_dim_for_range(
+    const TensorProto* start,
+    const TensorProto* limit,
+    const TensorProto* delta) {
+  const auto& start_data = ParseData<T>(start);
+  const auto& limit_data = ParseData<T>(limit);
+  const auto& delta_data =
+      delta != nullptr ? ParseData<T>(delta) : std::vector<T>(1, 1);
+
+  if (start_data.size() != 1 || limit_data.size() != 1 ||
```
wschin: Scalar in ONNX has empty shape. This basically means that we support both of scalar and tensor with a single element (shape can be `[1]`, `[1, 1, 1, 1, 1]`, and so on. Is this semantic expected?

onnx/defs/generator/defs.cc
wschin: We can make it non-optional.

onnx/defs/generator/defs.cc
wschin: You can use empty string, `set_name("")`, if this variable is not used.

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,185 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    auto* node_proto_2 = loop_sub_graph.add_node();
+    node_proto_2->set_op_type("Identity");
+    node_proto_2->add_input();
+    node_proto_2->set_input(0, "prev");
+    node_proto_2->add_output();
+    node_proto_2->set_output(0, "range");
+
+    // 'Loop' node 'body' attribute's graph inputs
+    auto* output_value_info_proto_0 = loop_sub_graph.add_output();
+    output_value_info_proto_0->set_name("cond_out");
```
wschin: Why do you need to output `cond_out`?

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,185 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+    return FunctionBodyHelper::BuildNodes({
+      // nodes: {outputs, op, inputs, attributes}
+      {{"sub_result"}, "Sub", {"limit", "start"}},
+      {{"delta_cast"}, "Cast", {"delta"}, {{"to", static_cast<int64_t>(1)}}},
+      {{"div_result"}, "Div", {"sub_result", "delta_cast"}},
+      {{"ceil_result"}, "Ceil", {"div_result"}},
+      {{"ceil_cast_int"}, "Cast", {"ceil_result"}, {{"to", static_cast<int64_t>(7)}}},
+      {{"ceil_cast_bool"}, "Cast", {"ceil_cast_int"}, {{"to", static_cast<int64_t>(9)}}},
+      {{"variadic_output", "output"}, "Loop", {"ceil_cast_int", "ceil_cast_bool", "start"}, {MakeAttribute("body", loop_sub_graph)}}
```
wschin: What are `Loop` node's outputs? I guess you can add `{"range"}` right before `MakeAttribute(...)`?

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,185 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+    auto* input_value_info_proto_1 = loop_sub_graph.add_input();
+    input_value_info_proto_1->set_name("cond");
+
+    auto* input_value_info_proto_2 = loop_sub_graph.add_input();
+    input_value_info_proto_2->set_name("prev");
+
+    // 'Loop' node 'body' attribute's graph nodes
+    auto* node_proto_0 = loop_sub_graph.add_node();
+    node_proto_0->set_op_type("Identity");
```
wschin: Is this `Identity` necessary?

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,185 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+    auto* output_value_info_proto_2 = loop_sub_graph.add_output();
+    output_value_info_proto_2->set_name("range");
+
+    return FunctionBodyHelper::BuildNodes({
+      // nodes: {outputs, op, inputs, attributes}
+      {{"sub_result"}, "Sub", {"limit", "start"}},
+      {{"delta_cast"}, "Cast", {"delta"}, {{"to", static_cast<int64_t>(1)}}},
+      {{"div_result"}, "Div", {"sub_result", "delta_cast"}},
+      {{"ceil_result"}, "Ceil", {"div_result"}},
```
wschin: If `Ceil` is used, could the `output` contain elements larger than `limit`?

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,185 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    node_proto_0->set_output(0, "cond_out");
+
+    auto* node_proto_1 = loop_sub_graph.add_node();
+    node_proto_1->set_op_type("Add");
+    node_proto_1->add_input();
+    node_proto_1->set_input(0, "prev");
+    node_proto_1->add_input();    
+    node_proto_1->set_input(1, "delta");
+    node_proto_1->add_output();
+    node_proto_1->set_output(0, "current");
```
wschin: Is it `current` or `prev`?

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,185 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+    // 'Loop' node 'body' attribute's graph nodes
+    auto* node_proto_0 = loop_sub_graph.add_node();
+    node_proto_0->set_op_type("Identity");
+    node_proto_0->add_input();
+    node_proto_0->set_input(0, "cond");
+    node_proto_0->add_output();
+    node_proto_0->set_output(0, "cond_out");
+
+    auto* node_proto_1 = loop_sub_graph.add_node();
```
gramalingam: I recommend using the builder convenience functions to keep this compact: e.g., see: https://github.com/onnx/onnx/blob/806aa863020fa180e57f576cb032ec44ce8ddcca/onnx/defs/nn/defs.cc#L2157 or https://github.com/onnx/onnx/blob/b29e78a4efb8e5d8995f576bbf19a846807829b6/onnx/defs/function.h#L69 .
For example, something like:
```
    // current = Add(prev, delta)
    { {"current"}, "Add", {"prev", "delta"}} 
```

hariharans29: Will do. Thanks for the suggestion.

onnx/defs/generator/defs.cc
hariharans29: That looks a bit ugly as usually delta is 1 and thus would be cleaner to make it optional. If the way to support optional inputs in functions is cumbersome, then we can make it optional. Any thoughts ?

onnx/defs/generator/defs.cc
hariharans29: I don't think so. 

The number of elements for range is computed as: 

Max(ceil((limit - start) / delta), 0)

like here:

https://github.com/microsoft/onnxruntime/blob/bf6a9f9c27e14cf334e6cf30e8a91c39f2456728/onnxruntime/contrib_ops/cpu/range.cc#L38

onnx/defs/generator/defs.cc
hariharans29: I was only asking for your review :) Thanks!

onnx/defs/generator/defs.cc
hariharans29: Not sure what you mean. 

In this `if` condition, I only check to make sure the initializers have a single elements (i.e.) scalar - I have in fact not checked the shape and I believe this loose check is fine.

onnx/defs/generator/defs.cc
hariharans29: If my understanding of the `Loop` operator is correct,I believe this has to be `current`. `current` is then mapped to the loop carried dependency `prev` below.

onnx/defs/generator/defs.cc
hariharans29: I believe this `Identity` is necessary. This `Idenetity` maps `cond` to `cond_out` which in turn becomes `cond` in the next iteration.

onnx/defs/generator/defs.cc
wschin: I'd prefer strongly-typed setting. If one variable is a scalar, we should use the scalar definition in ONNX. Could you please throw if the shape is not an empty list?

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,187 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  const auto& delta_data =
+      delta != nullptr ? ParseData<T>(delta) : std::vector<T>(1, 1);
+
+  if (start_data.size() != 1 || limit_data.size() != 1 ||
+      delta_data.size() != 1) {
+    fail_shape_inference(
+        "Input to 'Range' op should be scalars (Tensor with only one element)");
+  }
+
+  int n = static_cast<int>(
```
wschin: Please use int64, the largest allowed type in the spec below. Also, can we access the data during shape inference? I through this shape inference happens before we load the data.

onnx/defs/generator/defs.cc
wschin: @gramalingam, do we need to connect `cond` and `cond_out`? I thought `cout_out` is just an alias of `cond` so `cond_out` automatically store the latest value of `cond`.

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,187 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  if (n < 0)
+    n = 0;
+
+  return n;
+}
+
+const std::vector<NodeProto> build_nodes_range_op() {
+    // body for 'Loop node'
+    GraphProto loop_sub_graph;
+    loop_sub_graph.set_name("loop body attribute");
```
gramalingam: Please use a name like loop_body_attribute (a valid C identifier). See: https://github.com/onnx/onnx/blob/master/docs/IR.md#names-within-a-graph … while this check is probably not done (because of some models that violate this), it will help us in the future when we start enforcing these requirements.

onnx/defs/generator/defs.cc
gramalingam: The identity assignment is needed if they have different names. (I wish I could see the loop-body compactly ... our existing syntax for building this is so verbose I can't see it all in one place.)

onnx/defs/generator/defs.cc
wschin: For optional inputs, we use `empty string` as their names.

onnx/defs/generator/defs.cc
wschin: I just some manual calculation. You're right. Thanks.

onnx/defs/generator/defs.cc
wschin: You're probably right. My question comes from an statement in Loop's spec.
```
Any variables which you wish to make available in the enclosing scope (i.e. the variables b and keepgoing) must be declared as either loop-carried dependencies (_both_ at the op inputs and output and at the body net input and output) or scan_outputs.
```
which makes me wonder if we really replace `current` with `prev` in your code below
```cpp
    auto* output_value_info_proto_1 = loop_sub_graph.add_output(); 
    output_value_info_proto_1->set_name("current");
```
.

onnx/defs/generator/defs.cc
wschin: Are those names always different? If the same name can be used in both of input and output lists, SSA would be broken.

onnx/defs/generator/defs.cc
hariharans29: I did try that, but the function validator failed. I need to check if the validator has a bug.

onnx/defs/generator/defs.cc
hariharans29: Done

onnx/defs/generator/defs.cc
hariharans29: Yes, we can access data at shape inference if it is available as an initializer. There is a check on line 678 to make sure the necessary data is available as an initializer before invoking this method.

hariharans29: And I don't see why it should be made int64. The value that is being returned is of `int` type (this is a dim value which is int type). So after applying the formule - ceil(....), it needs to be casted back to 'int' as per requirement. This looks valid to me.

onnx/defs/generator/defs.cc
hariharans29: Ok, done.

onnx/defs/generator/defs.cc
hariharans29: This is a goo point. Honestly, I do not know if we ca make "current" -> "prev" in the specific line that you mention.  From the spec, I understood that the subgraph's outputs are: condition, loop carried dependencies (final values), and scan outputs. Since "cond" is made "cond_out" at the output, I guess it doesn't hurt to make "prev" into "current" at the output.  @gramalingam  - thoughts please?

P.S. - I didn't have a lot to go by when I wrote this implementation. I did this from what I could understand from the spec. Please suggest any simplifications if you can in the loop body...

wschin: Yes. Let's me add more details to make sure we are on the same page. The loop-carried dependencies are matched and assigned by position in the body graph's input and output lists. If we have iteration count `T`, loop-terminating condition `cond`, and loop-carried dependencies `[X, Y, Z]`, the input lists of the body graph would be `[T, cond, X, Y, Z]`, and its output list would be `[cond_new, X_new, Y_new, Z_new]`. The variable `cond_new`/`X_new`/`Y_new`/`Z_new` would be assigned to `cond`/`X`/`Y`/`Z` at the beginning of the next iteration.

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,188 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          const auto* delta_initializer =
+              num_inputs == 3 ? ctx.getInputData(2) : nullptr;
+
+          // Output is always 1-D
+          auto* output_dim = ctx.getOutputType(0)
+                                 ->mutable_tensor_type()
+                                 ->mutable_shape()
+                                 ->add_dim();
+
+          // all three inputs (if num_inputs == 3) (or)
```
wschin: Nice comment.

onnx/defs/generator/defs.cc
wschin: Now, I think we should use non-optional. Here is the reason.
The invoker of function needs to be aware of optional inputs of functions. If an optional input is empty, the runtime has to look into its spec and generate the right optional value. It means --- runtime will have an implementation for each function with some optional inputs, and therefore breaks the idea of having function.

onnx/defs/generator/defs.cc
wschin: Can't `limit_data` be int64? I thought it can according to the allowed types of `limit_data`. If the dim type must be int32, please throw if overflow happens. You can do something like
```
int64_t n = ...
if (n > INT_MAX)
  throw ...
```

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,188 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                (delta_initializer != nullptr &&
+                 start_initializer->data_type() !=
+                     delta_initializer->data_type())) {
+              fail_shape_inference(
+                  "All inputs to 'Range' op must be of the same type");
+            }
+
+            int output_dim_value = -1;
+
+            if (start_initializer->data_type() == TensorProto::FLOAT) {
```
wschin:
```suggestion
            // Explicitly compute the output dimension if Range's inputs are stored in initializer list.
            if (start_initializer->data_type() == TensorProto::FLOAT) {
```

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +527,188 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            if (start_initializer->data_type() !=
+                    limit_initializer->data_type() ||
+                (delta_initializer != nullptr &&
+                 start_initializer->data_type() !=
+                     delta_initializer->data_type())) {
+              fail_shape_inference(
+                  "All inputs to 'Range' op must be of the same type");
+            }
+
+            int output_dim_value = -1;
```
wschin:
```suggestion
            // If any of Range's inputs are not initializers, the output dimension would be unknown. 
            int output_dim_value = -1;
```

docs/Changelog.md
```diff
@@ -10648,6 +10648,44 @@ This version of the operator has been available since version 11 of the default
...
+
+#### Version
+
+This version of the operator has been available since version 11 of the default ONNX operator set.
+
+#### Inputs (2 - 3)
+
+<dl>
+<dt><tt>start</tt> : T</dt>
+<dd>Tensor(scalar, or dims=[1]). First entry in the range.</dd>
```
wschin: I think those inputs should just be scalars as described in your def.cc.

onnx/defs/generator/defs.cc
hariharans29: That is what I get from the spec too. So, "current" will be fed in as "prev" in the second iteration onwards. For the initial iteration, "start" will be fed in as "prev". Based, on "cond" in the first iteration, "start" will be included in the result (cond is true) or may not be (cond is false)

onnx/defs/generator/defs.cc
hariharans29: Add the comment - thanks.

onnx/defs/generator/defs.cc
hariharans29: Thanks for this comment. But I felt this comment is more suitable at the place we do the check in line 680 (the if... there does this check). The comment is added there. let me know what you think of this.

docs/Changelog.md
hariharans29: Right - I made the change in the def but didn't generate the doc and check-it in. Did that now. #Resolved.

onnx/defs/generator/defs.cc
hariharans29: ell I think 'int' may be either int64_t or int32_t based on the platform. So dim type can be either int32 or int64. Correct ? 

But I will add this check just to be safe when it is running on 32-bit platform...

onnx/defs/generator/defs.cc
hariharans29: Ok - making it non-optional is ugly, but I see no other choice. making it non-optional.

onnx/defs/generator/defs.cc
wschin: Both work to me. If you move it to line 680, please do
```suggestion
            // If any of Range's inputs are not initializers, the output dimension would be unknown (denoted by -1 below). 
            int output_dim_value = -1;
```

hariharans29: Moved already. Resolved.

onnx/test/shape_inference_test.py
```diff
@@ -1970,6 +1970,29 @@ def test_linearclassifier_2D_input(self):  # type: () -> None
                                           make_tensor_value_info('z', TensorProto.FLOAT, (4, 3))],
                                           opset_imports=[make_opsetid('ai.onnx.ml', 1), make_opsetid('', 11)])
 
+    def test_range(self):  # type: () -> None
+        graph = self._make_graph(
+            [('start', TensorProto.FLOAT, ()),
```
wschin: Need another test for `int` type.

wschin (approved):
LGTM. Thank you! My last minor comment is adding a test for integer type.

onnx/defs/generator/defs.cc
gramalingam: This is essentially a limitation of the function construct. We have been considering extending the function construct so that, instead of registering a fixed constant graph as its body, we register a lambda, similar to the type-and-shape-inference lambda, which takes a "context" (which will give us access to the types of inputs, information about attribute values, and which inputs/outputs are present/required), and returns a subgraph. With such an extension, it should be easy to support optional inputs like delta.

onnx/defs/generator/defs.cc
gramalingam: Just to make sure: why initialize this to -1 here? Please do not set_dim_value to -1: if it is unknown, please leave dim_value not set. It doesn't look like this -1 will reach the set_dim_value, but I just want to make sure. This is a slight problem because ONNX uses a convention slightly different from several frameworks where a single integer is used to represent both known and unknown dimensions, and this seems to confuse people.

onnx/defs/generator/defs.cc
gramalingam: Dim value is int64 (see: https://github.com/onnx/onnx/blob/3ba2e31a4954b4c151dfcf7deb141f244a43fecf/onnx/onnx.in.proto#L451 ). I think we should use int64 here.

gramalingam: I mean the return value, and the corresponding variable in the calling function should all be int64.

onnx/defs/generator/defs.cc
hariharans29: Thanks. That would be great. Looking forward to it. 

For now, resolving his comment as there is no action needed.

onnx/defs/generator/defs.cc
wschin: Cool, so the only change we need is to use int64 in the entire line.

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +528,193 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    auto* output_value_info_proto_2 = loop_sub_graph.add_output();
+    output_value_info_proto_2->set_name("range");
+
+    return FunctionBodyHelper::BuildNodes({
+      // nodes: {outputs, op, inputs, attributes}
+      {{"sub_result"}, "Sub", {"limit", "start"}},
+      {{"delta_cast"}, "Cast", {"delta"}, {{"to", static_cast<int64_t>(1)}}},
+      {{"div_result"}, "Div", {"sub_result", "delta_cast"}},
+      {{"ceil_result"}, "Ceil", {"div_result"}},
+      {{"ceil_cast_int"}, "Cast", {"ceil_result"}, {{"to", static_cast<int64_t>(7)}}},
```
gramalingam: I think that the "7" denotes the repeated ints type, and should be "2" here?

onnx/defs/generator/defs.cc
```diff
@@ -523,4 +528,193 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    output_value_info_proto_1->set_name("current");
+
+    auto* output_value_info_proto_2 = loop_sub_graph.add_output();
+    output_value_info_proto_2->set_name("range");
+
+    return FunctionBodyHelper::BuildNodes({
+      // nodes: {outputs, op, inputs, attributes}
+      {{"sub_result"}, "Sub", {"limit", "start"}},
+      {{"delta_cast"}, "Cast", {"delta"}, {{"to", static_cast<int64_t>(1)}}},
+      {{"div_result"}, "Div", {"sub_result", "delta_cast"}},
```
gramalingam: If I remember right, "Div" expects both inputs to be of same type. So, this would require a cast of "sub_result" also.

hariharans29: Fixed, thanks a lot!

hariharans29:
@gramalingam - Just to confirm, I got notified about this comment but don't find it here - 

"I think that the "7" denotes the repeated ints type, and should be "2" here?"

I assume you deleted it, correct ? I think "7" does correspond to int64. So it should be right. But just confirming that this needs no action.

gramalingam:
@hariharans29 : yes, I deleted that comment. My mistake, I was looking at the wrong enumeration-type.

onnx/test/shape_inference_test.py
hariharans29: This test is for shape inference. I assume you wanted another test for the actual op in `range.py` ? Anyway, I made one of the 2 tests in shape inference pertaining to INT as well.

hariharans29:
> @hariharans29 : yes, I deleted that comment. My mistake, I was looking at the wrong enumeration-type.

Thanks for confirming.

onnx/test/shape_inference_test.py
hariharans29: Added a test in 'range.py' pertaining to a node test for int32 and one for shape inference for int32. Resoving this.

hariharans29:
@linkerzhang @liqunfu @wschin @gramalingam @ebarsoum  - All comments should now be addressed. Can you guys please go a final check of the main op part (composed using a function) in the defs.cc and this should be ready for merge? Thanks a lot!

docs/Operators.md
```diff
@@ -10237,6 +10238,91 @@ This version of the operator has been available since version 1 of the default O
 </dl>
 
 
+### <a name="Range"></a><a name="range">**Range**</a>
+
+  Generate a tensor containing a sequence of numbers that begin at `start` and extends by increments of `delta`
```
ebarsoum: Can you add a pseudocode or a simple example in the doc?

ebarsoum (changes requested):

docs/Operators.md
hariharans29: Added pseudocode and some examples.

ebarsoum (approved):

hariharans29:
Closed in favor of #2042

PR #2088, opened 2019-06-07
Clarify shape inference requirements for new operators

hariharans29:
Clarify shape inference requirements for new operators

CC: @linkerzhang

### DIFF
--- a/docs/AddNewOp.md
+++ b/docs/AddNewOp.md
@@ -35,6 +35,9 @@ ambiguity, adding the links to refs if necessary
 6. type constraints about input and output tensors
 7. shape inference function Example:
 https://github.com/onnx/onnx/blob/master/onnx/defs/nn/defs.cc#L1183
+All operators MUST implement a shape inference function before it is accepted into ONNX.
+In cases where shape inference is not possible, it must have logic to perform 
+rank inference at the very least (adding right amount of dimensions to the output shape)
 
 Usually, if we can find similar functions in Numpy, we will try to align with
 numpy.
--- a/docs/AddNewOp.md
+++ b/docs/AddNewOp.md
@@ -33,11 +33,13 @@ ambiguity, adding the links to refs if necessary
 4. outputs,
 5. attributes,
 6. type constraints about input and output tensors
-7. shape inference function Example:
-https://github.com/onnx/onnx/blob/master/onnx/defs/nn/defs.cc#L1183
-All operators MUST implement a shape inference function before it is accepted into ONNX.
+7. shape inference function 
+Please provide a shape inference function in cases where it is meaningful and applicable.
 In cases where shape inference is not possible, it must have logic to perform 
 rank inference at the very least (adding right amount of dimensions to the output shape)
+Shape inference functions must be accompanied by unit tests in the file `onnx\test\shape_inference_test.py`.
+You can refer to the shape inference function for the `TopK` operator while implementing your own function -
+https://github.com/onnx/onnx/blob/master/onnx/defs/math/defs.cc#L943
 
 Usually, if we can find similar functions in Numpy, we will try to align with
 numpy.


docs/AddNewOp.md
```diff
@@ -35,6 +35,9 @@ ambiguity, adding the links to refs if necessary
 6. type constraints about input and output tensors
 7. shape inference function Example:
 https://github.com/onnx/onnx/blob/master/onnx/defs/nn/defs.cc#L1183
+All operators MUST implement a shape inference function before it is accepted into ONNX.
+In cases where shape inference is not possible, it must have logic to perform 
+rank inference at the very least (adding right amount of dimensions to the output shape)
```
pranavsharma: Can we also add that if the shape cannot be determined for whatever reasons, we shouldn’t throw an exception?

docs/AddNewOp.md
```diff
@@ -35,6 +35,9 @@ ambiguity, adding the links to refs if necessary
 6. type constraints about input and output tensors
 7. shape inference function Example:
 https://github.com/onnx/onnx/blob/master/onnx/defs/nn/defs.cc#L1183
+All operators MUST implement a shape inference function before it is accepted into ONNX.
```
pranavsharma: Do we want to enforce this? If yes, can we enforce this in the code rather than documentation (can happen in a separate PR)?

docs/AddNewOp.md
hariharans29: As discussed offline - (FYI for others) - I would refrain from discouraging contributors to not throw exceptions in case of bad inputs/attribute data. For legitimate cases, exceptions should be thrown, isn't it ?

Also this seems like a detail - something that should be caught in PR reviews (should a user throw an exception here (is this bad input?) or should the user just return back gracefully with a no-op (not enough data to proceed?). Such things can be clarified at code-review time.)

docs/AddNewOp.md
hariharans29: For now - I vote that the key reviewers for ONNX ops (not a lot of people) have a common agenda - refrain from checking-in ops without shape inference and encourage contributors to implement them as part of code review process.

If we need to enforce such a thing, that obviously means we need to plug gaps (missing shape inference functions) along with the PR that enforces it...

houseroad: Asking shape inference for all the nodes, is it too much?

Also mention where to add the tests for shape inference? And provide a good example for shape inference?

docs/AddNewOp.md
pranavsharma: Sure.

docs/AddNewOp.md
linkerzhang: @pranavsharma this is a document on how to add a new op, so that I think it's ok to ask contributors to add shape inference function for new ops.

@houseroad @hariharans29 we may rephrase it a little bit to not a "MUST", but provide shape inference function when it's applicable and meaningful.

hariharans29: Updated

linkerzhang: great. the updated one look good to me. @houseroad  sounds good to you?

gramalingam:
Hi: In general, I see that existing shape-inference code can be simplified quite a bit if (a) We lift common patterns into utility functions so it can be shared, and (b) We use these utility functions appropriately. In the context of the current PR, it would help to have some "model shape inference functions" that we can point to so that the new code follows recommended styles. But may be that requires some work first, so perhaps we can do it as a separate PR. I am happy to help with that. (For example, the documentation here points to TopK. I think it is valuable to extract lines 967 to 976 as a utility function "getScalarConstant(...)", which is likely to be needed for all ops which take a scalar input which may be a constant.)

hariharans29:
> Hi: In general, I see that existing shape-inference code can be simplified quite a bit if (a) We lift common patterns into utility functions so it can be shared, and (b) We use these utility functions appropriately. In the context of the current PR, it would help to have some "model shape inference functions" that we can point to so that the new code follows recommended styles. But may be that requires some work first, so perhaps we can do it as a separate PR. I am happy to help with that. (For example, the documentation here points to TopK. I think it is valuable to extract lines 967 to 976 as a utility function "getScalarConstant(...)", which is likely to be needed for all ops which take a scalar input which may be a constant.)

This is a valuable suggestion @gramalingam. As you said, let's do this across multiple small PR(s) that are easy to review.

PR #2089, opened 2019-06-08
Clarify RNN op spec regarding sequence lengths.

edgchen1:


gramalingam:
Hi, it might be worth double-checking if this is a clarification or change in spec: the previous spec left it ambiguous whether the values beyond the sequence-length will be zero or potentially undefined/garbage. There could be some implementations that choose the latter interpretation.

gramalingam:
Just to clarify (@edgchen1 and I talked about it offline): I was wondering whether the version-number of the op needs to be bumped or not? @Ke, @ebarsoum, any thoughts?

edgchen1:
@linkerzhang, @ebarsoum, @gramalingam - another question: in the (nonsensical?) case where sequence_lens specifies a zero-length sequence, with this change, Y should contain zeros for that sequence. What about Y_h and Y_c (given possibly non-zero initial_h and initial_c)?

gramalingam:
I am okay with this change if we need it: but I am just curious why we need to specify that the outputs will be zeroed out beyond the sequence length (given that in many cases they may never be used). This may have a performance impact (e.g., in a cpu setting as opposed to gpu). Is it really needed?

edgchen1:
@gramalingam One motivation for avoiding possibly uninitialized memory in the output is to allow consumers of the output to freely read it without risking undefined behavior (admittedly a C/C++-specific concern). Alternatively, the consumers need to be aware that they had better not try to access the possibly uninitialized segments of the output.

gramalingam:
What is the risk? It is not as if we access an invalid memory location. These are valid memory locations with an undefined value. Is this zeroing standard in other existing frameworks? If so, that is a good reason to adopt it.

docs/Changelog.md
```diff
@@ -6728,7 +6728,7 @@ This version of the operator has been available since version 7 of the default O
 
 <dl>
 <dt><tt>Y</tt> (optional) : T</dt>
-<dd>A tensor that concats all the intermediate output values of the hidden. It has shape `[seq_length, num_directions, batch_size, hidden_size]`. </dd>
+<dd>A tensor that concats all the intermediate output values of the hidden. If a sequence with length less than `seq_length` is specified via sequence_lens, values past the end of that sequence will be set to zero. This tensor has shape `[seq_length, num_directions, batch_size, hidden_size]`.</dd>
```
ebarsoum: The end of the sequence shouldn't be set to zero.

ebarsoum: Even in C++, you have the length of the sequence, so the remaining part shouldn't be read.

edgchen1: Ok. That was the main point of this PR, so I'll just close it.

PR #2093, opened 2019-06-11
Add string equal operator

waterknows:
Add onnx support for string equal comparison.


wschin:
Why does this PR change so many model files?

gramalingam:
Equals still does not support string inputs. It makes sense to make this extension, but it seems better to close this PR, and open a new PR if someone wants to.

PR #2096, opened 2019-06-11
Update Clip in opset 11 to support min/max as inputs instead of attributes

BowenBao:
Needed to improve Faster/Mask R-CNN for supporting variable image sizes. 
Detection boxes can be clipped based on runtime image size.


onnx/defs/math/defs.cc
```diff
@@ -608,33 +608,44 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            std::numeric_limits<float>::max())
-        .Input(0, "input", "Input tensor whose elements to be clipped", "T")
+            "M",
+            OpSchema::Optional)
         .Output(0, "output", "Output tensor with clipped input elements", "T")
         .TypeConstraint(
             "T",
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to float tensors.")
+        .TypeConstraint(
```
linkerzhang: Thanks for the refactoring! The 3 inputs (input, min, max) should have same type. Let's use same "T" for them then.

onnx/defs/math/defs.cc
```diff
@@ -608,33 +608,44 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .SetDoc(Clip_ver11_doc)
+        .Input(
+            0,
+            "input",
+            "Input tensor whose elements to be clipped",
+            "T")
+        .Input(
+            1,
             "min",
             "Minimum value, under which element is replaced by min",
```
linkerzhang: this min should be a scaler. please make it clear in the description.

linkerzhang: same for max, I think.

onnx/backend/test/case/node/clip.py
```diff
@@ -16,70 +16,74 @@ class Clip(Base):
...
+            inputs=['x', 'min', 'max'],
             outputs=['y'],
-            min=-1.0,
-            max=1.0
         )
 
         x = np.array([-2, 0, 2]).astype(np.float32)
-        y = np.clip(x, -1, 1)  # expected output [-1., 0., 1.]
-        expect(node, inputs=[x], outputs=[y],
+        min_val = np.array(-1).astype(np.float32)
```
linkerzhang: this one should be a scaler, instead of an array. min_val = np.float32(-1)?

onnx/defs/math/defs.cc
BowenBao: Thanks for reviewing. The reason I separated out is to try mention the shape constraint on min and max in TypeConstraint, I know this also brings in the side effect that `T` and `M` now can actually be different float types. I'll use the same "T" and move the shape constraint to input description.

onnx/backend/test/case/node/clip.py
BowenBao: I think we should support both scalar and 1-d tensor. That seems to be the consensus for the current onnx spec of the definition scalar. 

For example
https://github.com/onnx/onnx/blob/master/docs/Operators.md#size

> Size
Takes a tensor as input and outputs a int64 **scalar** that equals to the total number of elements of the input tensor.
...
```
node = onnx.helper.make_node(
    'Size',
    inputs=['x'],
    outputs=['y'],
)
x = np.array([
    [1, 2, 3],
    [4, 5, 6],
]).astype(np.float32)
y = np.array(6).astype(np.int64)
expect(node, inputs=[x], outputs=[y],
       name='test_size_example')
```

https://github.com/onnx/onnx/blob/master/docs/Operators.md#nonmaxsuppression

> score_threshold (optional) : tensor(float)
Float representing the threshold for deciding when to remove boxes based on score. It is a scalar
```
iou_threshold = np.array([0.5]).astype(np.float32)
score_threshold = np.array([0.0]).astype(np.float32)
selected_indices = np.array([[0, 0, 3], [0, 0, 0], [0, 0, 5]]).astype(np.int64)
```

onnx/defs/math/defs.cc
```diff
@@ -608,33 +608,44 @@ ONNX_OPERATOR_SET_SCHEMA(
     8,
     OpSchema().FillUsing(ElementwiseMultiOpDocGenerator("mean")));
 
-static const char* Clip_ver6_doc = R"DOC(
+static const char* Clip_ver11_doc = R"DOC(
 Clip operator limits the given input within an interval. The interval is
 specified with arguments 'min' and 'max'. They default to
```
wschin:
```suggestion
specified by the inputs 'min' and 'max'. They default to
```

onnx/defs/math/defs.cc
```diff
@@ -608,33 +608,44 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 numeric_limits::lowest() and numeric_limits::max() respectively.
```
wschin:
```suggestion
numeric_limits::lowest() and numeric_limits::max(), respectively.
```

onnx/defs/math/defs.cc
```diff
@@ -608,33 +608,44 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "Input tensor whose elements to be clipped",
+            "T")
+        .Input(
+            1,
             "min",
             "Minimum value, under which element is replaced by min",
-            AttributeProto::FLOAT,
-            std::numeric_limits<float>::lowest())
-        .Attr(
+            "M",
```
wschin:
```suggestion
            "T",
```
Could we make it type-safer?

onnx/defs/math/defs.cc
```diff
@@ -608,33 +608,44 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         .Output(0, "output", "Output tensor with clipped input elements", "T")
         .TypeConstraint(
             "T",
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to float tensors.")
+        .TypeConstraint(
+            "M",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Tensor of float that contains only 1 element. "
+            "It can be either a scalar(rank = 0), or a 1-D tensor.")
```
wschin: It should be just a scalar.

onnx/backend/test/case/node/clip.py
wschin: If you allow tensors, it means you need to check the shape in runtime because ONNX can have dynamic shapes. Whenever possiblre, I prefer strongly-typed way, so I'd suggest to follow Ke's comment. 

Previously, we didn't do a good job to emphasize the difference between scalar and 1-element tensor, but it doesn't mean we can't starting improving it from now on.

:)

onnx/backend/test/case/node/clip.py
```diff
@@ -16,70 +16,74 @@ class Clip(Base):
...
+        max_val = np.array(1).astype(np.float32)
+        y = np.clip(x, min_val, max_val)  # expected output [-1., 0., 1.]
+        expect(node, inputs=[x, min_val, max_val], outputs=[y],
                name='test_clip_example')
 
         x = np.random.randn(3, 4, 5).astype(np.float32)
-        y = np.clip(x, -1.0, 1.0)
-        expect(node, inputs=[x], outputs=[y],
+        y = np.clip(x, min_val, max_val)
```
wschin: Could you add expected output in a comment (just like what you did right above)?

onnx/backend/test/case/node/clip.py
```diff
@@ -16,70 +16,74 @@ class Clip(Base):
...
-            inputs=['x'],
+            inputs=['x', 'min'],
             outputs=['y'],
-            min=0.0
         )
+        min_val = np.array(0).astype(np.float32)
         x = np.random.randn(3, 4, 5).astype(np.float32)
-        y = np.clip(x, 0.0, np.inf)
-        expect(node, inputs=[x], outputs=[y],
+        y = np.clip(x, min_val, np.inf)
```
wschin: Could you print out the expected output in a comment? Foe example,
```
# expected output of print(y)
#  [0.2, 0.2]
```

onnx/backend/test/case/node/clip.py
```diff
@@ -16,70 +16,74 @@ class Clip(Base):
...
+            inputs=['x', no_min, 'max'],
             outputs=['y'],
-            max=0.0
         )
+        max_val = np.array(0).astype(np.float32)
         x = np.random.randn(3, 4, 5).astype(np.float32)
-        y = np.clip(x, -np.inf, 0.0)
-        expect(node, inputs=[x], outputs=[y],
+        y = np.clip(x, -np.inf, max_val)
+        expect(node, inputs=[x, max_val], outputs=[y],
```
wschin:
```suggestion
        expect(node, inputs=[x, "", max_val], outputs=[y],
```
maybe?

onnx/backend/test/case/node/clip.py
BowenBao: good point for known inputs, bug how to add expected output for inputs that are randomly generated?

onnx/backend/test/case/node/clip.py
BowenBao: This is correct. Please check the implementation of `expect`
https://github.com/onnx/onnx/blob/47acb06a969c3fd899f7a13270588ddb0cccd1c1/onnx/backend/test/case/node/__init__.py#L107
and another use case of optional input with operator `Scan`
https://github.com/onnx/onnx/blob/47acb06a969c3fd899f7a13270588ddb0cccd1c1/onnx/backend/test/case/node/scan.py#L43

onnx/backend/test/case/node/clip.py
BowenBao: I agree that runtime will need to check the shape, but that is introduced by changing min/max from attributes to tensors. It has nothing to do with whether it is a scalar(0-rank tensor) or a 1-element tensor.

scalar vs 1-elem tensor is a bigger issue and should be discussed separately. It is not immediately clear to me which one is more preferable. That being said, I think we should use 'tensor' as description in the spec, avoid using the ambiguous 'scalar'.

onnx/defs/math/defs.cc
BowenBao: thanks, updated.

onnx/defs/math/defs.cc
BowenBao: thanks, updated.

onnx/backend/test/case/node/clip.py
wschin: In ONNX, the definition of scalar is a tensor with empty shape. Since `min` and `max` attributes are scalars (they are float attribute instead of floats attribute), I don't feel we should use 1-element tensor to store them.

docs/Changelog.md
```diff
@@ -10332,6 +10332,41 @@ This version of the operator has been available since version 11 of the default
 <dd>Constrain input and output types to integer tensors.</dd>
 </dl>
 
+### <a name="Clip-11"></a>**Clip-11**</a>
+
+  Clip operator limits the given input within an interval. The interval is
+  specified by the inputs 'min' and 'max'. They default to
+  numeric_limits::lowest() and numeric_limits::max(), respectively.
```
wschin: Write out values explicitly.
```
3.343 x 1e304 (for example, numeric_limits::max() in c++)
```

onnx/backend/test/case/node/clip.py
gramalingam: I agree with @wschin and @linkerzhang : it is better to consistently use a tensor with an empty shape as a scalar, instead of a 1D tensor with 1 element. We have not been doing this consistently, but this will potentially create problems later on. In this case, since we are specifying a unit-test-case, we should use zero-dimensional tensors to represent scalars in the test-case. Otherwise, implementations will start supporting both zero-dimensional and 1D tensors (to get past the test-case) and then when somebody wants to do code-generation they will be unable to represent the value using a primitive value in the generated-code … they will be forced to support a boxed representation, making things inefficient.

gramalingam: Note that people can always use a Reshape to coerce any tensor with 1 element into a scalar. The idea is that people should pay the cost of this kind of representation-conversion only when needed. Strong typing will enable zero-cost when typing guarantees the representation is already the correct one.

gramalingam: As long as the test-cases here represent the corresponding scalars with empty shape (is this correct?), we should be fine here.

BowenBao: Thanks for the review @wschin @gramalingam @linkerzhang ! I agree we should have a clear and strict definition of scalar.

BowenBao:
@linkerzhang could you take a look at the updates to the requested changes?

spandantiwari:
Thanks @ebarsoum.

PR #2098, opened 2019-06-12
Avoid unnecessary copies of names by checker

skottmckay:
Currently LexicalScopeContext used by the checker copies the current set of names when descending into nodes. I believe this is only necessary if the node contains a GraphProto attribute, in order to keep the names in the GraphProto attribute from polluting the enclosing scope.

To avoid that, add the ability to nest LexicalScopeContext instances so that when descending into a GraphProto attribute the names in the parent graph scope are kept separate. 

Reduces load time of SSD MobileNet model by over 20x.

NOTE: The changed files were formatted using the ONNX .clang-format file, which introduced quite a few whitespace diffs. Is that expected? I can manually undo those diffs if preferred.


onnx/checker.cc
```diff
@@ -451,34 +463,41 @@ void check_graph(
       if (output.empty()) {
         continue;
       }
-      if (output_names.count(output)) {
+
+      // TODO: should this check parent graph/s as well?
+      // Doing so could break existing models though
+      if (lex_ctx.this_graph_has(output)) {
```
gramalingam: I think "lex_ctx.this_or_parent_graph_has(output)" would preserve the existing behavior, since output_names at this point, includes outer-scope names.

onnx/checker.h
```diff
@@ -80,8 +80,32 @@ class CheckerContext final {
...
+
+  void add(const std::string& name) {
+    output_names.insert(name);
+  }
+
+  bool this_graph_has(const std::string& name) const {
+    return output_names.find(name) != output_names.cend();
+  }
+
+  bool this_or_parent_graph_has(const std::string& name) const {
```
gramalingam: minor nit: perhaps "this_or_ancestor_graph_has" would be more accurate.

gramalingam:
Looks good to me, thanks.

onnx/checker.cc
```diff
@@ -389,31 +401,33 @@ void check_graph(
...
-  std::unordered_set<std::string> output_names{};
-  // Inherit values avaiailable in outer scope
+  // Inherit values available in outer scope
   // Note that we do not allow shadowing, so the presence of an already-defined
   // name is always an error.
+  LexicalScopeContext lex_ctx{parent_lex};
+
   for (const auto& value_info : graph.input()) {
-    if (output_names.count(value_info.name())) {
+    // TODO: If shadowing isn't allowed, this should maybe use
```
houseroad: check line 405, we don't allow shadowing

skottmckay: Doesn't that happens after the graph (or subgraph) inputs are checked? So in the case of a subgraph, a value in the parent graph could be shadowed by a subgraph input.

gramalingam: @houseroad : taking a step back: the existing version of the checker allows shadowing in the case of graph-inputs (as Scott mentions). However, it does look slightly odd, considering that shadowing is not allowed for node outputs inside the subgraph. I believe Scott's current version preserves the existing behavior, but if it seems appropriate we could change the behavior to forbid shadowing for the subgraph inputs also. (The concern is whether it will break some existing model; @skottmckay : I assume that the related bug in ORT implies that no model currently using ORT could have used this feature successfully?)

skottmckay: Regarding the bug in ORT, the model could have run but it would not have been correct as it would have been using the incorrect value (outer scope value instead of the subgraph input value).

onnx/checker.h
```diff
@@ -80,8 +80,32 @@ class CheckerContext final {
...
+    return this_graph_has(name) ||
+        (parent_context_ && parent_context_->this_or_ancestor_graph_has(name));
+  }
+
+  // public for backwards compatibility. please prefer the public interface of
+  // this class over directly changing output_names
   std::unordered_set<std::string> output_names;
+
+ private:
+  const LexicalScopeContext* parent_context_{nullptr};
```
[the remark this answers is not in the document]
skottmckay: It's a raw pointer so the ownership is outside of this class. As the parent context would be null for the top level graph a reference can't be used. 

The LexicalScopeContext instance is just a temporary local variable when the checker is traversing the graph so the parent instance should always be available. Added comments to clarify.

linkerzhang:
@houseroad do you have any more comments/concern please? Thank you!

Issue #2100, opened 2019-06-13
Add a mode to ROIAlign for implementing tf.crop_and_resize?


jiafatom:
In tensorflow, many object detection models use tf.crop_and_resize to extract ROI features. One similar op in onnx doing this is ROIAlign. But they are now behaving differently in three aspects:
(1) the order of input image tensor differs, ROIAlign [x1,y1,x2,y2], tf.crop_and_resize [y1,x1,y2,x2]
(2) ROIAlign rois value is integer, tf.crop_and_resize is normalized by roi size
(3) The sampling technique is different. 

(3) is the most important because we cannot use ROIAlign to implement crop_and_resize, then the conversion of those pretrained object model cannot get correct result. Since these two ops are doing similar things, I want to propose **adding a new mode for ROIAlign** which uses the same sampling technique as crop_and_resize. Then ROIAlign has two modes, one is crop_and_resize, the other is the existing algorithm.

Thank you.

wschin:
[crop_and_resize](https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize) is a combination of two steps: (1) Crop and (2) Resize. Couldn't we use `Crop` and `Resize` from ONNX to compose it?

fuzzyBatman:
@jiafatom Are you adding the support for this op (CropAndResize) to ONNX?

jiafatom:
> @jiafatom Are you adding the support for this op (CropAndResize) to ONNX?

yes

jiafatom:
> [crop_and_resize](https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize) is a combination of two steps: (1) Crop and (2) Resize. Couldn't we use `Crop` and `Resize` from ONNX to compose it?

synced offline, we agree on adding a mode for `ROIAlign`.

fuzzyBatman:
> > @jiafatom Are you adding the support for this op (CropAndResize) to ONNX?
> 
> yes

Thank you! I was thinking of working on it.
However, you seem to have a better idea and a head start.
Waiting for the op support :)

jiafatom:
With offline discussion, the current solution is to add CropAndResize op rather than adding a mode.
See [PR](https://github.com/onnx/onnx/pull/2142)

PR #2102, opened 2019-06-14
Definition of RoiAlign to support crop_and_resize and RoiAlign

jiafatom:
This PR handles the issue https://github.com/onnx/onnx/issues/2100
We want to add a mode for RoiAlign op to support both crop_and_resize and RoiAlign.
Let's agree on the interface first. Will add tests (maybe in another PR).

### DIFF
--- a/onnx/defs/object_detection/defs.cc
+++ b/onnx/defs/object_detection/defs.cc
@@ -18,19 +18,30 @@ quantizations while converting from original image into feature
 map and from feature map into RoI feature; in each ROI bin,
 the value of the sampled locations are computed directly
 through bilinear interpolation.
+
+The operator [crop_and_resize](https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize)
+in tensorflow behaves similarly as RoiAlign, where
+the major difference is the coordinate location and the mode chosen to compute the output tensor.
+This RoiAlign supports both crop_and_resize and RoiAlign.
 )DOC";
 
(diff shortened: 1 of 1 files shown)


jiafatom:
Does anyone agree (or not) the interface change? @wschin

linkerzhang:
please follow https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md to update an operator.

wschin:
LGTM. Thanks.

jiafatom:
@ebarsoum do you agree the interface? thanks.

onnx/defs/object_detection/defs.cc
```diff
@@ -18,19 +18,30 @@ quantizations while converting from original image into feature
...
 )DOC";
 
 ONNX_OPERATOR_SET_SCHEMA(
     RoiAlign,
-    10,
+    11,
     OpSchema()
       .SetDoc(RoiAlign_ver1_doc)
+      .Attr(
+          "method",
```
ebarsoum: Most other OP use `mode` instead of `method`

jiafatom: > Most other OP use mode instead of method

We already have an attribute called `mode`, used to distinguish the mode in `roi_align`. If we change `method` to `mode`, then what should the old `mode` be called?

jiafatom:
We decide to add an op rather than adding a mode, so I am going to use [PR](https://github.com/onnx/onnx/pull/2142) instead. Close this one.

PR #2106, opened 2019-06-18
Support GatherND operator in ONNX

hariharans29:
This change is to support GatherND operator in ONNX.


gramalingam:
TF's gather-nd seems to have an extra batch_dims attribute for batched gather-nd … is that required or useful?

hariharans29:
> Also could you provide some brief summary about the PR in the PR's description?

Well sure - but the description is same as the title - to support GatherND op in ONNX. Do you mean you want a justification to include this by any chance ?

hariharans29:
> TF's gather-nd seems to have an extra batch_dims attribute for batched gather-nd … is that required or useful?

Not sure about this - let me take a look. Thanks!

hariharans29:
> > TF's gather-nd seems to have an extra batch_dims attribute for batched gather-nd … is that required or useful?
> 
> Not sure about this - let me take a look. Thanks!

@gramalingam - I took a look at the `batch_dims` concept in the tf Gather_nd op and like you said, it is to support batched inputs ('params' and 'indices' in tf op) and does seem like a good extension to the op's abstraction. However, I see no immediate use-case for us to justify including that. If need be, we can version this op later to support `batch_dims`. Or I can make the spec adjustment now and modify the contrib implementation of this op in ORT. Any thoughts?

CC: @wschin @linkerzhang @spandantiwari

gramalingam:
I personally think it is okay to leave it for future extension if we don't see a use-case for it.

hariharans29:
> I personally think it is okay to leave it for future extension if we don't see a use-case for it.

I think this is a good idea too. We can always version the op later.

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  const auto& indices_shape = ctx.getInputType(1)->tensor_type().shape();
+  const auto indices_rank = indices_shape.dim_size();
+
+  if (data_rank < 1 || indices_rank < 1)
+    fail_shape_inference(
+        "Both `data` and `indices` tensors need to have rank larger than 0.");
+
+  // cannot ascertain if the input shapes are valid if shape of `indices` 
+  // is missing last dimension value so return at this point
+  if (!indices_shape.dim(indices_rank - 1).has_dim_value())
```
spandantiwari: nit: Based on the code format style followed in this file (at least) I would suggest putting the bidy for for/if in braces, even for single-statement body. Similar to what you are doing above on line 1836.

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    return;
+
+  const auto last_indice_dimension =
+      indices_shape.dim(indices_rank - 1).dim_value();
+
+  if (last_indice_dimension > data_rank)
+    fail_shape_inference(
```
spandantiwari: nit: same, braces.

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  if (data_rank < 1 || indices_rank < 1)
+    fail_shape_inference(
+        "Both `data` and `indices` tensors need to have rank larger than 0.");
+
+  // cannot ascertain if the input shapes are valid if shape of `indices` 
+  // is missing last dimension value so return at this point
+  if (!indices_shape.dim(indices_rank - 1).has_dim_value())
+    return;
+
+  const auto last_indice_dimension =
```
spandantiwari: Some people might see the name `last_indice_dimension` as a typo. Maybe rename this to `last_index`dimension`

spandantiwari:
@hariharans29 - Overall, this looks OK. I don't have any specific blocking issues on this. Could you please take a look at the recently approved guideline document for adding new ops and update (if needed) based on that. 
https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md 
Some suggestions in line with the guidelines:
1. Add more details in the op definition; could we include some equations?
2. Add more test examples.
3. Add a reference implementation in the test script.

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    .SetDoc(GatherND_ver11_doc)
+    .Input(0, "data", "Tensor of rank r >= 1.", "T")
+    .Input(1, "indices", "Tensor of rank q >= 1.", "tensor(int64)")
+    .Output(0, "output", "Tensor of rank q + r - indices_shape[-1] - 1.", "T")
+    .TypeConstraint(
+        "T",
+        OpSchema::all_tensor_types(),
+        "Constrain input and output types to any tensor type.")
+    .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+  // Type inference
```
BowenBao: nit: indent seems off starting from this line

docs/Changelog.md
```diff
@@ -10332,6 +10332,54 @@ This version of the operator has been available since version 11 of the default
 <dd>Constrain input and output types to integer tensors.</dd>
 </dl>
 
+### <a name="GatherND-11"></a>**GatherND-11**</a>
+
+  Given `data` tensor of rank r >= 1, and `indices` tensor of rank q >= 1, gathers
+  slices of `data` into an output tensor of rank q + r - indices_shape[-1] - 1.
```
wschin: Please describe the indexing mechanism to compute output values. It'd be very helpful if we have something like
```
output[i][j] = data[indices[i]][indices[j], :]
```

docs/TestCoverage.md
```diff
@@ -2533,6 +2533,28 @@ expect(node, inputs=[data, indices.astype(np.int64)], outputs=[y],
...
+```python
+node = onnx.helper.make_node(
+    'GatherND',
+    inputs=['data', 'indices'],
+    outputs=['output'],
+)
+
+data = np.array([[[0, 1], [2, 3]], [[4, 5], [6, 7]]], dtype=np.float32)
+indices = np.array([[[0, 1]], [[1, 0]]], dtype=np.int64)
+output = np.array([[[2, 3]], [[4, 5]]], dtype=np.float32)
```
wschin: Could you compute output using a reference implementation? The actual value can be kept in a comment such as
```
output = your_implementation(data, index) # output = [[2, 3], [4, 5]]

File: docs/Changelog.md [hunk 7]
```diff
@@ -10332,6 +10332,54 @@ This version of the operator has been available since version 11 of the default
 <dd>Constrain input and output types to integer tensors.</dd>
 </dl>
 
+### <a name="GatherND-11"></a>**GatherND-11**</a>
+
+  Given `data` tensor of rank r >= 1, and `indices` tensor of rank q >= 1, gathers
```
wschin: Is it possible to have `q>r`?

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+  const auto& data_shape = ctx.getInputType(0)->tensor_type().shape();
+  const auto data_rank = data_shape.dim_size();
+
+  const auto& indices_shape = ctx.getInputType(1)->tensor_type().shape();
+  const auto indices_rank = indices_shape.dim_size();
+
+  if (data_rank < 1 || indices_rank < 1)
+    fail_shape_inference(
+        "Both `data` and `indices` tensors need to have rank larger than 0.");
```
```suggestion
        "Both `data` and `indices` tensors in GatherND need to have rank larger than 0.");
```

onnx/defs/tensor/defs.cc
wschin: Could you please explain why this check is required?

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  // is missing last dimension value so return at this point
+  if (!indices_shape.dim(indices_rank - 1).has_dim_value())
+    return;
+
+  const auto last_indice_dimension =
+      indices_shape.dim(indices_rank - 1).dim_value();
+
+  if (last_indice_dimension > data_rank)
+    fail_shape_inference(
+        "Last dimension of `indices` tensor must not be larger than the rank of `data` tensor");
```
```suggestion
        "Last dimension of `indices` tensor in GatherND must not be larger than the rank of `data` tensor");
```

docs/TestCoverage.md
hariharans29: I don't understand why this is needed, sorry. To implement this in Python might not be needed in the context of this file. We can add equations - I feel that is good enough. This file is mostly to generate the node test case that will be tested by a backend - not for humadn understanding....

onnx/defs/tensor/defs.cc
hariharans29: Just as the comment above it says - if the indices_shape is missing a value in the last dimension - it is not possible to validate the tensor shape requirement for this op. So we stop at this point with the shape inferencing as we are not able to validate the input shapes....

docs/TestCoverage.md
wschin: Yes, reference implementation is a requirement for new operators per Op WG's conclusion. :)

onnx/defs/tensor/defs.cc
wschin: I understnad every word in this sentence, but why do you choose `!indices_shape.dim(indices_rank - 1).has_dim_value()`? Can I do `!indices_shape.dim(i).has_dim_value()` for `i=0,...,rank-1`?

hariharans29: Fixed, thanks

onnx/defs/tensor/defs.cc
hariharans29: Fixed

onnx/defs/tensor/defs.cc
hariharans29: Should be okay now

docs/Changelog.md
hariharans29: It is very hard to write a single equation for this op - in fact the equation written by tf gather_nd is not even clear to me. Instead, I have a detailed section on how to compute the output values in my latest commit and I have linked the explanation to an illustrative example. I believe this drives home the usage of the operator better than some equation that may be hard to parse.

docs/Changelog.md
hariharans29: This is a very good question - thanks. I have addressed all this in the doc for this op in my latest commits.

onnx/defs/tensor/defs.cc
hariharans29: Fixed

onnx/defs/tensor/defs.cc
hariharans29: No we don't have to check for every dimension. Please look at the next check in the next line or two. We need to perform a validation against `data_rank` for which we need this dim_value. Without this dim_value, we cannot perform this validation and hence we stop at that point.

docs/TestCoverage.md
hariharans29: Added reference implementation as suggested.

hariharans29:
@wschin @spandantiwari @BowenBao - Thanks for all the comments. I have addressed them all. Can you please take a look again ? Thanks!

docs/Changelog.md
```diff
@@ -10720,6 +10720,99 @@ This version of the operator has been available since version 11 of the default
...
+   
+  1) If `indices_shape[-1] > r` => error condition
+  
+  2) If `indices_shape[-1] == r`, since the rank of `indices` is `q`, `indices` can be thought of as a `(q-1)`-dimensional tensor
+     containing 1-D tensors of dimension `r`. Let us think of each such `r` ranked tensor as `indices_slice`. 
+     Each *scalar value* corresponding to `data[indices_slice]` is filled into the corresponding location of the `(q-1)`-dimensional tensor 
+     to form the `output` tensor (Example 1 below)
+  
+  3) If `indices_shape[-1] < r`, since the rank of `indices` is `q`, `indices` can be thought of as a `(q-1)`-dimensional tensor
+     containing 1-D tensors of dimension `< r`. Let us think of each such tensors as `indices_slice`.
```
wschin: >1-D tensors [](start = 16, length = 11)

1-D tensors?

hariharans29: yes - plural. 1-D tensors => as there are many 1-D tensors within the (q-1)-dimensional tensor

onnx/defs/tensor/defs.cc
```diff
@@ -2206,77 +2214,238 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+where each element defines a slice of `data`
+
+Some salient points about the inputs' rank and shape:
+ 
+1) r >= 1 and q >= 1 are to be honored. There is no dependency condition to be met between ranks `r` and `q`.
+
+2) The `indices_shape[-1]` should have a value between 1 (inclusive) and rank `r` (inclusive) 
+
+3) The value of `indices` need to be non-negative (i.e.) >= 0. The upper bound for each value in `indices` 
+   is the `data_shape` value along the corresponding axis (i.e.) `0 <= indices[:, :, i] < data_shape[i]`
```
gramalingam: Replace "indices[:,:,i]" by "indices[…,i]" ?

onnx/defs/tensor/defs.cc
```diff
@@ -2206,77 +2214,238 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             }
           }
-      }));
+        }));
+
+static const char* GatherND_ver11_doc = R"DOC(
+Given `data` tensor of rank `r` >= 1, and `indices` tensor of rank `q` >= 1, this operator gathers 
+slices of `data` into an output tensor of rank `q + r - indices_shape[-1] - 1`.
+
+`indices` is an q-dimensional integer tensor, best thought of as a `(q-1)`-dimensional tensor of indices into `data`,
```
gramalingam: May be change "tensor of indices" to "tensor of index-tuples"?

onnx/defs/tensor/defs.cc
```diff
@@ -2206,77 +2214,238 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+1) r >= 1 and q >= 1 are to be honored. There is no dependency condition to be met between ranks `r` and `q`.
+
+2) The `indices_shape[-1]` should have a value between 1 (inclusive) and rank `r` (inclusive) 
+
+3) The value of `indices` need to be non-negative (i.e.) >= 0. The upper bound for each value in `indices` 
+   is the `data_shape` value along the corresponding axis (i.e.) `0 <= indices[:, :, i] < data_shape[i]`
+
+The output is computed as follows:
+
+The output tensor can be thought of as filling in the `indices` tensor with the appropriate slices of the input `data`.
```
gramalingam: Change to "The output tensor is obtained by mapping each index-tuple in the  `indices` tensor to the corresponding slice of the input `data`"

onnx/defs/tensor/defs.cc
hariharans29: sure

hariharans29:
Let me also add - This operator is the inverse of ScatterND ,since ScatterND has a similar line in it

gramalingam:
By the way: there was a discussion about supporting negative indices in Gather (since pytorch supports that). Do we want to extend this to allow negative indices?

hariharans29:
> By the way: there was a discussion about supporting negative indices in Gather (since pytorch supports that). Do we want to extend this to allow negative indices?

This is a good question - in this spec version, I explicitly did not allow it and pretty much followed the spec clarification in `Gather`. Any comments @spandantiwari @BowenBao  ? Btw - Does `GatherElemets`, `ScatterElements`, and `ScatterND` support negative indices ?

hariharans29:
Spoke offline to @spandantiwari and @BowenBao - We will support negative indexing in this op after a consensus on it is reached and we will add negative indexing support in a more streamlined manner across similar ops.

For now, it makes sense to stay consistent with `Gather` which doesn't support negative indexing.

PR #2108, opened 2019-06-18
Fix some minor issues in IR.md and Versioning.md

edgchen1:


docs/IR.md
```diff
@@ -55,7 +55,7 @@ The IR specification uses simple monotonically increasing numbers for its versio
   IR_VERSION = 0x0000000000000003;
 ```
 
-Operator sets use a simple version number. Each operator set version represents the combination of the most recent version of each operator.
+Operator sets use a simple version number. Each operator set version represents the combination of the most recent version of each operator at a particular point in time.
```
edgchen1: **each** operator set version representing the combination of the **most recent version** of each operator was confusing to me. hopefully this matches the intended meaning.

docs/IR.md
```diff
@@ -115,7 +115,7 @@ Name|Type|Description
 magic|string|The value ‘ONNXOPSET’
 ir_version|int32|The ONNX version corresponding to the operators.
 ir_version_prerelease|string|The prerelease component of the SemVer of the IR.
-ir_build_metadata|string|The symbolic identifier of the operator to invoke.
+ir_build_metadata|string|TODO come up with wording here
```
edgchen1: the original looks like a copy-paste error. need some help with the actual wording

docs/IR.md
linkerzhang: I'm not a good English editor/writer, so am not sure whether this editing change improves the readability and clarifies more on semantics of operator set versioning. @prasanthpul  @gramalingam @houseroad @ebarsoum 

The operator set version may be considered as a time point (normally one per release). Each operator may have its specific semantics since one time point (one operator set version). That's why operator does not have "version" itself, but "sinceversion".

gramalingam: I think "Each operator set version represents a combination of a specific version of some set of operators." would be simpler. I think adding "The operator set version may considered as a time point. Each version of an operator belongs to some contiguous set of operator set versions. Normally, there is one operator set version per release of ONNX (if and only if there is a change in the set of operators or the specification of any existing operator)." would also be helpful.

docs/IR.md
gramalingam: The build metadata of this version of the operator set?

edgchen1: thanks for the suggestion.
since the field is named `ir_build_metadata`, is it the metadata of the operator set or the IR?
if the former, would a better name for this field be `opset_build_metadata`?

docs/IR.md
edgchen1: updated and added an example to versioning.md.

docs/IR.md
gramalingam: Changing field names is tough, right? May be @linkerzhang can comment on it, but it may not be worth changing field names. (My suggestion was based on the documentation in the proto file … I don't really know what this is used for.)

docs/Versioning.md
```diff
@@ -114,18 +139,17 @@ Model authors and applications/systems MAY elect to ignore the model versioning
 
     * Breaking changes to the semantics of an input or output (e.g., changing the required contents of an input tensor from a color image to a black and white image).
     * Changing the declared type of an input or output to an incompatible type (e.g., `tensor(int)->tensor(string)`).
-    * Adding a new input for which there is no meaningful or specified default value. For graph inputs, those values are provided by a same-named value in GraphProto. initializer.
```
linkerzhang: “ For graph inputs, those values are provided by a same-named value in GraphProto. initializer.", it's better to keep this. This is telling "how to specify the default value of a graph input".

docs/Versioning.md
```diff
@@ -85,23 +85,48 @@ operator id MUST be greater than any extant `op_version` for the
...
 
-How nodes bind to operator declarations is strictly defined, and are designed to increase model compatibility across ONNX implementations (appealing to the conservative clause of the robustness principle). 
+How nodes bind to operator declarations is strictly defined, and are designed to increase model compatibility across ONNX implementations (appealing to the conservative clause of the robustness principle).
 
 How ONNX implementations bind an operator declaration to specific implementation is outside the scope of this specification. Implementations of ONNX MAY elect to introduce more sophisticated operator declaration/implementation binding modes to appeal to the liberal clause of the robustness principle.
 
+### Operator Set and Operator Versioning Example
```
linkerzhang: Thank you so much for adding this table to show a very clear view of operator set version evolvement.

docs/IR.md
```diff
@@ -351,7 +351,7 @@ Dimension variables are scoped to the declaration that they appear in. For graph
 
 For example, a graph that performs matrix cross-product may be defined as taking two inputs of shape [K,M] and [M,N], and producing an output of shape [K,N].
 
-The emptry string "", when used as a dimension name, denotes a single dimension of any cardinality. The string "*", when used as a dimension name, denotes zero or more dimensions of unknown cardinality.
+The empty string "", when used as a dimension name, denotes a single dimension of any cardinality. The string "*", when used as a dimension name, denotes zero or more dimensions of unknown cardinality.
```
gramalingam: Hmmm … this (the original version itself, I don't mean this PR's change) is inaccurate. There were proposals for both these extensions, but I don't believe either is supported. I believe the statement should be entirely dropped.

docs/IR.md
```diff
@@ -351,7 +351,7 @@ Dimension variables are scoped to the declaration that they appear in. For graph
```
gramalingam: Even the preceding paragraph seems inaccurate. This could be a subject for design discussion, but I believe currently the dimension variables are globally scoped.

docs/Versioning.md
edgchen1: that's also specified here, right?
https://github.com/edgchen1/onnx/blame/edgchen1/ir_doc_fixes/docs/IR.md#L159

would it be better to specify it once? that information also seemed out of place here.

docs/IR.md
edgchen1: ok, i can remove it

docs/IR.md
edgchen1: how should we follow up?

gramalingam: Let's merge in what you have. I can take a shot at clarifying this.

docs/Versioning.md
gramalingam: Well, no harm in reminding the reader that the default value is specified by the initializer. Most people are familiar with the term "initializer", but may not immediately correlate that to a "default value". How about adding back "Recall that default-values for inputs are specified in the initializer list."

linkerzhang: I agree to add some words back for more clarification.

PR #2115, opened 2019-06-19
Fix inconsistency in describing graph's initializer. 

xykong58:
This PR is to ensure the consistency on relations between graph's input and initializer.
Essentially, a graph's initializer is not required to be subset of the graph's input, and
any initializer not part of graph's input are considered constant definition in the graph.


xykong58:
Do I need to do anything to push the change to the master ?

PR #2122, opened 2019-06-20
Update gen_doc script to validate proto3 files

raymondxyang:
Add a step in gen_doc.py to validate the generated proto3 files by using them to generate the .cc .h code.
This step will only be triggered when **ONNX_VERIFY_PROTO3** is set to true in CI env.


raymondxyang:
I added this into windows CI already. Any changes breaks the protoc compilation will be blocked

houseroad:
@raymondxyang could you sign the CLA?

gramalingam:
Raymond left Microsoft. So, we may need to duplicate this, if we want to merge this in.

PR #2126, opened 2019-06-24
Liqun/topk sort

liqunfu:


liqunfu:
@houseroad @linkerzhang 
This PR updates TopK to take optional attributes for direction and output sorting. It aligns with TF and PyTorch better than the current version.

docs/Changelog.md
```diff
@@ -10529,3 +10529,61 @@ This version of the operator has been available since version 11 of the default
...
+  Retrieve the top-K largest or smallest elements along a specified axis. Given an input tensor of
+  shape [a_1, a_2, ..., a_n, r] and integer argument k, return two outputs:
+    -Value tensor of shape [a_1, a_2, ..., a_{axis-1}, k, a_{axis+1}, ... a_n]
+      which contains the values of the top k elements along the specified axis
+    -Index tensor of shape [a_1, a_2, ..., a_{axis-1}, k, a_{axis+1}, ... a_n] which
+     contains the indices of the top k elements (original indices from the input
+     tensor).
+  
+  If "largest" is 1 (the default value) then the k largest elements are returned.
+  If "sorted" is 1 (the default value) then the resulting k elements will be sorted.
```
wschin: Adding `sorted` implies the need of another `Sort` operator. Would you consider split the new top-k into two operators?

docs/Changelog.md
```diff
@@ -10529,3 +10529,61 @@ This version of the operator has been available since version 11 of the default
...
+
+  Retrieve the top-K largest or smallest elements along a specified axis. Given an input tensor of
+  shape [a_1, a_2, ..., a_n, r] and integer argument k, return two outputs:
+    -Value tensor of shape [a_1, a_2, ..., a_{axis-1}, k, a_{axis+1}, ... a_n]
+      which contains the values of the top k elements along the specified axis
+    -Index tensor of shape [a_1, a_2, ..., a_{axis-1}, k, a_{axis+1}, ... a_n] which
+     contains the indices of the top k elements (original indices from the input
+     tensor).
+  
+  If "largest" is 1 (the default value) then the k largest elements are returned.
```
wschin: I am not sure if `largest` is necessary. If a user wants to have a graph:
```
X ---> TopK(largest=0) ---> Y, IndexY
```
He can do this
```
X ---> Neg ---> NegX ---> TopK ---> NegY, IndexY
NegY ---> Neg ---> Y
```
Another possibility is to propose `BottomK` as a `FunctionProto` which mimics the graph right above.

wschin:
`docs/TestCoverage.md` got removed? Is it expected?

docs/Changelog.md
liqunfu: good question on Sort operator! Actually sort op could be defined as a function op using TopK with 'sorted' attribute. This is one of reasons of this PR.  Hope it clears the question of "split the new top-k into two operators".

docs/Changelog.md
liqunfu: I think it is better to avoid extra cost of Negs. 
Regarding BottomK, I think "TopK" is a more acceptable term. It aligns with PyTorch spec.

docs/Changelog.md
wschin: I feel we start preferring smaller primitives when possible, so I still likely weight `Sort+TopK` more over modifying `TopK`.

docs/Changelog.md
wschin: This is not extra cost. How a FunctionProto is evaluated is totally up to runtime. Even if you have a very complicated FunctionProto, a runtime can still map that FunctionProto to a single custom operator.

onnx/defs/math/defs.cc
```diff
@@ -890,24 +890,27 @@ ONNX_OPERATOR_SET_SCHEMA(
...
   -Value tensor of shape [a_1, a_2, ..., a_{axis-1}, k, a_{axis+1}, ... a_n]
     which contains the values of the top k elements along the specified axis
   -Index tensor of shape [a_1, a_2, ..., a_{axis-1}, k, a_{axis+1}, ... a_n] which
    contains the indices of the top k elements (original indices from the input
    tensor).
-   
-Given two equivalent values, this operator uses the indices along the axis  as
+
+If "mode" is 1 (the default value) then the k largest elements are returned.
+If "sorted" is 1 (the default value) then the resulting k elements will be sorted.
```
wschin: I'd suggest to use `order` as the attribute name and that attribute is an enum of `"ascent"` and `"descent"`.

onnx/defs/math/old.cc
```diff
@@ -1294,5 +1295,116 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .TypeConstraint(
+            "I",
+            {"tensor(int64)"},
+            "Constrain index tensor to int64")
+        .Attr(
+            "axis",
+            "Dimension on which to do the sort.",
+            AttributeProto::INT,
+            static_cast<int64_t>(-1))
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
```
wschin: Do you have shape inference tests?

onnx/defs/math/defs.cc
gramalingam: I'd prefer changing the attribute-name to "largest" (or something like that). "Order/descent/ascent" is still slightly ambiguous whether it is talking about which elements are selected or about the order in which the selected elements appear in the output.

gramalingam: It may also be useful to say that the attribute value must be 0 or 1 (to avoid dealing with ambiguous cases like a value like 2).

onnx/defs/math/old.cc
gramalingam: One comment/question about the shape-inference logic (I understand it was pre-existing shape-inference logic, not part of this PR): this checks that the "axis" input is a 1-dimensional tensor. However, it is usually recommended that scalars be represented as 0-dimensional tensor. This will likely cause some unexpected error later on. I think it is better that we allow 0-dimensional tensors also here.

onnx/defs/math/defs.cc
wschin: @gramalingam, this operator can return the largest `K` or the smallest `K` elements depending on the attribute `mode`, so `largest` looks ambiguous.

gramalingam: "largest" is a Boolean flag. So, if "largest=1" return K largest. If "largest=0" return K smallest. Or, use a string-valued "mode" with possible values as "largest" or "smallest" … but I think Boolean-flags are simpler than string-valued flags.

wschin: Your `largest` attribute is the `mode` attribute in this PR. This PR's `sorted` attribute indicates if the returned values should be sorted.

docs/Operators.md
```diff
@@ -15008,6 +15015,44 @@ expect(node, inputs=[X, K], outputs=[values_ref, indices_ref],
...
+    [4, 5, 6],
+    [8, 9, 10],
+], dtype=np.float32)
+indices_ref = np.array([
+    [0, 1, 2],
+    [0, 1, 2],
+    [3, 2, 1],
+], dtype=np.int64)
+
+expect(node, inputs=[X, K], outputs=[values_ref, indices_ref],
```
wschin: This place needs an reference implementation and the output should be computed using the reference implementation.

onnx/backend/test/case/node/topk.py
```diff
@@ -38,3 +38,35 @@ def export_top_k():  # type: () -> None
 
         expect(node, inputs=[X, K], outputs=[values_ref, indices_ref],
                name='test_top_k')
+
+    @staticmethod
+    def export_top_k_smallest():  # type: () -> None
+        mode = 0
+        sorted = 1
+
+        node = onnx.helper.make_node(
```
wschin: This place needs an reference implementation and the output should be computed using the reference implementation.

onnx/defs/math/old.cc
```diff
@@ -1294,5 +1295,116 @@ ONNX_OPERATOR_SET_SCHEMA(
           result_shape.mutable_dim(static_cast<int>(axis))->set_dim_value(k);
           updateOutputShape(ctx, 0, result_shape);
           updateOutputShape(ctx, 1, result_shape);
-        }));		
+        }));
+
+static const char* TopK_ver10_doc = R"DOC(
+Retrieve the top-K elements along a specified axis. Given an input tensor of
```
wschin: Is the sorting algorithm stable?

onnx/defs/math/defs.cc
gramalingam: Okay, it looks like we were talking about two different attributes. I was suggesting renaming the "mode" attribute to "largest" (because mode=1 or 0 doesn't tell me what it means). It looks like you are talking about changing the "sorted" attribute. Sorry about the confusion.

onnx/defs/math/old.cc
liqunfu: yes it is stable sort, i.e. to use axis indices as a tiebreaker for equal values.

onnx/defs/math/old.cc
liqunfu: Yes it is the preexisting shape inference code. Spec update in this PR does not change shape logic.
Regarding axis type, because it is an attribute instead of an input, it shall be a int64_t.

onnx/defs/math/defs.cc
liqunfu: I changed attribute name back from 'mode' to 'largest'. 
Regarding attribute name 'sort', it is redundant to name it with ascending or descending because order is already implied by the 'largest' attribute. I updated description for sort to make it clearer:
'If "sorted" is 0, order of returned 'Values' and 'Indices' are undefined.'

liqunfu:
@gramalingam , @linkerzhang, @ebarsoum please confirm if this PR can be merged.

onnx/defs/math/defs.cc
gramalingam: How about extending "resulting k elements will be sorted." to "resulting k elements will be sorted, in descending order if largest=1 and in ascending order if largest=0."?

ebarsoum:
Can you resolve the conflict?

PR #2127, opened 2019-06-25
Updated test cases for reshape

JamesAllingham:
* Increased the number of test cases from 5 to 7.

    * The new tests cover the 0 dim case

* Changed the existing cases so that, where possible, the 'batch' dim is unchanged.

    * Some frameworks do not allow for the first dim to change, so this makes these tests more generally applicable (developers of ONNX importers and exporters can try and replicate the results from these test cases).

* Generated the new test data, updated the operators doc, and updated the test coverage doc.


JamesAllingham:
Hmmm, I can't seem to figure out why travis-ci is failing. There doesn't seem to be a reason given. Just `The command "./.travis/script.sh" exited with 1.`. Does anyone have any idea what the problem might be?

JamesAllingham:
@linkerzhang any idea why the travis-ci build is failing here? Also, the circlecle build has started to fail but I don't think it is a problem on my side since it used to pass and I haven't made any changes since it last passed. The error message suggests a problem with the venv: 

```
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

postrational:
@JamesAllingham I think Travis CI is failing, because your PR ends up modifying these files:
```
	modified:   docs/Operators.md
	modified:   docs/TestCoverage.md
```
These files are generated from other sources.
If you want to change them, you need to change their sources, then re-generate them and add to your PR.

JamesAllingham:
@postrational Thanks, re-generating the files has fixed the issue.

docs/Operators.md
```diff
@@ -13066,11 +13066,13 @@ Other versions of this operator: <a href="Changelog.md#Reshape-1">Reshape-1</a>
 ```python
 original_shape = [2, 3, 4]
 test_cases = {
-    'reordered_dims': np.array([4, 2, 3], dtype=np.int64),
-    'reduced_dims': np.array([3, 8], dtype=np.int64),
-    'extended_dims': np.array([3, 2, 2, 2], dtype=np.int64),
+    'reordered_dims': np.array([2, 4, 3], dtype=np.int64),
```
wschin: It seems that we should still have a test to test if the first dimension can be changed correctly. It doesn't sound right if we change all tests not to touch the first dimension. Those are tests for ONNX operators, not specific framework. What do you think?

docs/Operators.md
```diff
@@ -13081,7 +13083,13 @@ for test_name, shape in test_cases.items():
...
     )
 
-    reshaped = np.reshape(data, shape)
+    # replace zeros with corresponding dim size
+    # we need to do this because np.reshape doesn't support 0
+    new_shape = np.copy(shape)
+    zero_index = np.where(shape == 0)
+    new_shape[zero_index] = np.array(original_shape)[zero_index]
+
+    reshaped = np.reshape(data, new_shape)
```
wschin: It means that we should have a reference implementation of `ONNX Reshape`. I am thinking something like
```
reshaped = onnx_reshape_impl(…)
```

docs/Operators.md
JamesAllingham: Totally agree, I did try to make sure that there were still tests touching the first dim, but I didn't think to specifically test reordering of the first dim with another. I'll add one back in.

docs/Operators.md
JamesAllingham: Sure, I'll  do that 👍

PR #2129, opened 2019-06-26
Support make_attribute empty string

shjwudp:
onnx.helper.make_attribute(key, "") would got a FLOATS attribute, fix it


lucienwang1009:
Empty list also falls into this issue.

PR #2132, opened 2019-06-26
Fix shape inference logic for Tensor and Math op families

hariharans29:
This change strives to address:

1) Bug fixes for existing shape inference logic
2) Fill implementation gaps - either partial (support for shape inference/ rank inference for missing cases) or full support (entire implementation)
3) Improve error messages
4) More tests for ops in this family


hariharans29:
CC: @linkerzhang @houseroad

onnx/defs/tensor/defs.cc
```diff
@@ -252,14 +243,12 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-                ->add_dim()
-                ->set_dim_value(
-                    ctx.getInputType(0)->tensor_type().shape().dim_size());
-          }
+          ctx.getOutputType(0)
+              ->mutable_tensor_type()
+              ->mutable_shape()
+              ->add_dim()
+              ->set_dim_value(
+                  ctx.getInputType(0)->tensor_type().shape().dim_size());
```
linkerzhang: what will be set if "ctx.getInputType(0)->tensor_type().has_shape() is false"? "0"? if that's the case it's wrong. you may always add a dim, but the value should not be set in that case (ctx.getInputType(0)->tensor_type().has_shape() is false).

hariharans29: Well I just removed the duplicate check - in the line previous to this - the if condition invokes `hasNInputShapes()` which invokes `hasInputShape()` which basically checks if `has_shape()` is true along with a bunch of other stuff. So this if check is redundant and so I removed it.

One thing that I can improve is actually to add the dimension even before the first if check (rank inference) because output is always 1D. I can fill the value with input rank if shape is known.

gramalingam:
@hariharans29 : do you know if this PR is still relevant? Or, can it be closed?

PR #2141, opened 2019-06-28
Unique

liqunfu:
this is a continuation of https://github.com/onnx/onnx/pull/2042


onnx/defs/tensor/defs.cc
```diff
@@ -1797,62 +1797,115 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              auto* dim = yTensorProto->mutable_tensor_type()->mutable_shape()->add_dim();
+              if (i != axis)
+                dim->set_dim_value(input_shape.dim(i).dim_value());
+            }
 
-          return;
-        }));
+            // 'indices', 'inverse_indices', and 'counts' are 1-D tensors of unknown dimension.
+            if (indicesTensorProto)
+              indicesTensorProto->mutable_tensor_type()->mutable_shape()->add_dim();
```
gramalingam: According to the description earlier, indices has same shape as input.

onnx/defs/tensor/defs.cc
```diff
@@ -1797,62 +1797,115 @@ ONNX_OPERATOR_SET_SCHEMA(
...
   input_x = [2, 1, 1, 3, 4, 3]
   output_uniques = [2, 1, 3, 4]
   output_idx = [0, 1, 1, 2, 3, 2]
   output_counts = [1, 2, 2, 1]
+
+Example 2:
+  input_x = [[1, 3], [2, 3]]
+  output_uniques = [1, 2, 3]
+  output_idx = [[0, 2], [1, 2]]
+  output_counts = [1, 2, 2, 1]
```
gramalingam: This doesn't seem consistent with the description … shouldn't this have the same number of elements as "uniques"?

onnx/defs/tensor/defs.cc
```diff
@@ -1797,62 +1797,115 @@ ONNX_OPERATOR_SET_SCHEMA(
 
 static const char* Unique_ver11_doc = R"DOC(
 Finds all the unique values (deduped list) present in the given input tensor. This operator returns 3 outputs. 
-The first output tensor 'uniques' contains all of the unique elements of the input, sorted in the same order that they occur in the input.
+The first output tensor 'uniques' contains all of the unique elements of the input, 
+optionally sorted in ascending or the same order that they occur in the input.
```
gramalingam: Please add description of "inverse_indices" output also.

gramalingam:
What happens in the "axis" case when the number of unique elements is not the same at every position? I guess the dimension will be the maximum, but what happens to the undefined/padding positions? How will the users know which are padding values?

onnx/defs/tensor/defs.cc
gramalingam: Similar comments for the other outputs as well. I assume that they will not be flattened if axis is specified?

liqunfu: Updated the spec to take output shapes of numpy. indices, inverse_indices, and counts are all 1-D tensors.

onnx/defs/tensor/defs.cc
liqunfu: spec updated. when axis is not provided, input is flattened. output are all 1-D tensors. When axis is provided, output Y has the same shape of "X" except in dimension "axis" which is unknown.

onnx/defs/tensor/defs.cc
liqunfu: updated comment to fix this.

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
           propagateShapeFromInputToOutput(ctx, 0, 0);
         }));
+
+static const char* Unique_ver11_doc = R"DOC(
+Find the unique elements of a tensor. When an optional attribute 'axis' is provided, unique subtensors along the 'axis' are returned. 
+Otherwise the input tensor is flattened and unique values of the flattened tensor are returned. 
+
+This operator returns the unique elements of the input tensor and three optional outputs. 
+The first output tensor 'Y' contains all unique values or subtensors of the input. 
+The second optional output tensor 'indices' contains indices of the the unique elements in input.
```
gramalingam: How about "The second optional output tensor 'indices': indices[j] contains the index of the first occurrence of Y[j] in input X."

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         }));
+
+static const char* Unique_ver11_doc = R"DOC(
+Find the unique elements of a tensor. When an optional attribute 'axis' is provided, unique subtensors along the 'axis' are returned. 
+Otherwise the input tensor is flattened and unique values of the flattened tensor are returned. 
+
+This operator returns the unique elements of the input tensor and three optional outputs. 
+The first output tensor 'Y' contains all unique values or subtensors of the input. 
+The second optional output tensor 'indices' contains indices of the the unique elements in input. 
+The third optional output tensor 'inverse_indices' contains indices of input elements in 'Y'.
```
gramalingam: The third optional output tensor 'inverse_indices': inverted_indices[j] contains the index of the occurrence of X[j] in the output Y.

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            // 'axis' is not provided. Input 'X' is flattened.
+            // 'Y' is a 1-D tensor of unknown dimension.
+            yTensorProto->mutable_tensor_type()->mutable_shape()->add_dim();
+          }
+          else 
+          {
+            // 'axis' is provided.
+            int axis = static_cast<int>(axisAttr->i());
+
+            // 'Y' has the same shape as 'X' except in the 'axis' dimension which is unknown.
```
gramalingam: When I read the scipy documentation, it says that all the other dimensions (other than axis) are flattened ... is that correct, or does that flattening not happen for the output?

liqunfu: I think it means that the rest dimensions are flattened after the 'axis' dimension is moved to the front for sorting. However the final output is transposed back to the original input form. Test cases also conform this.

onnx/defs/tensor/defs.cc
liqunfu: Good one! Updated the doc.

onnx/defs/tensor/defs.cc
liqunfu: updated the doc according to the comments.

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         }));
+
+static const char* Unique_ver11_doc = R"DOC(
+Find the unique elements of a tensor. When an optional attribute 'axis' is provided, unique subtensors along the 'axis' are returned. 
+Otherwise the input tensor is flattened and unique values of the flattened tensor are returned. 
+
+This operator returns the unique elements of the input tensor and three optional outputs. 
+The first output tensor 'Y' contains all unique values or subtensors of the input. 
+The second optional output tensor 'indices': indices[j] contains the index of the first occurrence of Y[j] in input X. 
+The third optional output tensor 'inverse_indices': inverted_indices[j] contains the index of the occurrence of X[j] in the output Y.
```
gramalingam: "inverted_indices[j]" => "inverse_indices[j]"

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+static const char* Unique_ver11_doc = R"DOC(
+Find the unique elements of a tensor. When an optional attribute 'axis' is provided, unique subtensors along the 'axis' are returned. 
+Otherwise the input tensor is flattened and unique values of the flattened tensor are returned. 
+
+This operator returns the unique elements of the input tensor and three optional outputs. 
+The first output tensor 'Y' contains all unique values or subtensors of the input. 
+The second optional output tensor 'indices': indices[j] contains the index of the first occurrence of Y[j] in input X. 
+The third optional output tensor 'inverse_indices': inverted_indices[j] contains the index of the occurrence of X[j] in the output Y. 
+The forth optional output tensor 'counts' contains the count of each element of 'Y' in the input.
```
gramalingam: "forth" => "fourth"

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            // 'axis' is provided.
+            int axis = static_cast<int>(axisAttr->i());
+
+            // 'Y' has the same shape as 'X' except in the 'axis' dimension which is unknown.
+            const TensorShapeProto& input_shape = xTensorProto->tensor_type().shape();
+
+            for (int i = 0; i < input_shape.dim_size(); i++) {
+              auto* dim = yTensorProto->mutable_tensor_type()->mutable_shape()->add_dim();
+              if (i != axis){
+                if (input_shape.dim(i).has_dim_value())
```
gramalingam: Can replace these 4 lines by "*dim = input_shape.dim(I)", I think.

gramalingam: That is a lower-case eye ... auto-correct capitalized it :-(

wschin:
What are the targeted frameworks of this operator? TF, numpy, and pytorch?

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .SetDoc(Unique_ver11_doc)
+        .Attr(
+            "sorted",
+            "(Optional) Whether to sort the unique elements in ascending order before returning as output. "
+            "Must be one of 0, or 1 (default).",
+            AttributeProto::INT,
+            static_cast<int64_t>(1))
+        .Attr(
+            "axis",
+            "(Optional) The dimension to apply unique. If None, the unique elements of the flattened input are returned.",
```
wschin: This is optional, so what's the default value?

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            AttributeProto::INT,
+            OPTIONAL)
+        .Input(0, "X", "A N-D input tensor that is to be processed.", "T")
+        .Output(0, "Y", "A 1-D tensor of the same type as 'X' "
+                        "containing all the unique values in 'X', either sorted "
+                        "or maintained in the same order they occur in the input 'X'", "T")
+        .Output(1, "indices", "A 1-D INT64 tensor " 
+                          "containing corresponding input tensor indices for elements in 'Y'",
```
wschin: Do you mean linear index? In general, index of a tensor element is a tuple such as `(i,j,k)` when the tensor is 3-D.

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+This operator returns the unique elements of the input tensor and three optional outputs. 
+The first output tensor 'Y' contains all unique values or subtensors of the input. 
+The second optional output tensor 'indices': indices[j] contains the index of the first occurrence of Y[j] in input X. 
+The third optional output tensor 'inverse_indices': inverted_indices[j] contains the index of the occurrence of X[j] in the output Y. 
+The forth optional output tensor 'counts' contains the count of each element of 'Y' in the input. 
+
+Outputs are either sorted in ascending order or optionally in the order they occur in the input. 
+
+https://docs.scipy.org/doc/numpy/reference/generated/numpy.unique.html
```
wschin: Can this signature also handle TF's unique?

onnx/defs/tensor/defs.cc
wschin: Is `X[j]` also a correct expression when outputing sub-tensors?

onnx/defs/tensor/defs.cc
```diff
@@ -1794,4 +1794,132 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            AttributeProto::INT,
+            OPTIONAL)
+        .Input(0, "X", "A N-D input tensor that is to be processed.", "T")
+        .Output(0, "Y", "A 1-D tensor of the same type as 'X' "
+                        "containing all the unique values in 'X', either sorted "
+                        "or maintained in the same order they occur in the input 'X'", "T")
+        .Output(1, "indices", "A 1-D INT64 tensor " 
+                          "containing corresponding input tensor indices for elements in 'Y'",
+                          "tensor(int64)")
+        .Output(2, "inverse_indices", "A 1-D INT64 tensor "
```
wschin: I saw you mentioned that output 1 to output 3 are all optional above. Could you also add `Optional` into their doc strings here?

onnx/test/shape_inference_test.py
```diff
@@ -1915,6 +1915,26 @@ def test_reversesequence(self):  # type: () -> None
             [])
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (4, 5, 6))])
 
+    def test_unique_without_axis(self):  # type: () -> None
```
wschin: It'd be nice if you can add more complicated cases (such as 2x3x2 tensor) and some edge cases such as scalar, one-element tensor, etc.

docs/Operators.md
```diff
@@ -15093,6 +15094,144 @@ expect(node, inputs=[data], outputs=[transposed],
...
+  The first output tensor 'Y' contains all unique values or subtensors of the input. 
+  The second optional output tensor 'indices': indices[j] contains the index of the first occurrence of Y[j] in input X. 
+  The third optional output tensor 'inverse_indices': inverted_indices[j] contains the index of the occurrence of X[j] in the output Y. 
+  The forth optional output tensor 'counts' contains the count of each element of 'Y' in the input. 
+  
+  Outputs are either sorted in ascending order or optionally in the order they occur in the input. 
+  
+  https://docs.scipy.org/doc/numpy/reference/generated/numpy.unique.html
+  
+  Example 1:
```
hobei: Should we have examples in the description and in the 'examples' section?

onnx/defs/tensor/defs.cc
liqunfu: it is to support both TF and PyTorch.

onnx/defs/tensor/defs.cc
```diff
@@ -1801,4 +1801,129 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "Must be one of 0, or 1 (default).",
+            AttributeProto::INT,
+            static_cast<int64_t>(1))
+        .Attr(
+            "axis",
+            "(Optional) The dimension to apply unique. If not specified, the unique elements of the flattened input are returned.",
+            AttributeProto::INT,
+            OPTIONAL)
+        .Input(0, "X", "A N-D input tensor that is to be processed.", "T")
+        .Output(0, "Y", "A 1-D tensor of the same type as 'X' "
```
gramalingam: Whether the output is 1-D depends on whether the axis attribute is specified.  I suggest dropping the "1-D" part above.

gramalingam:
A general comment: it looks like this op will not work well in the batched case. Batching requires a different treatment of axes. For example, if the outermost axis is a batch axis, then with an input of [ [0,0,1], [1, 2, 2] ], the user would want an output of [ [0,1], [1,2] ]. The current definition does not allow us to get this with or without the axis attribute. (I think we should think about what ONNX's philosophy and approach is for batch-support in general. But this op may be a good example to drive the general approach we want, so mentioning it here.)

onnx/defs/tensor/defs.cc
```diff
@@ -1801,4 +1801,129 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+Find the unique elements of a tensor. When an optional attribute 'axis' is provided, unique subtensors along the 'axis' are returned. 
+Otherwise the input tensor is flattened and unique values of the flattened tensor are returned. 
+
+This operator returns the unique elements of the input tensor and three optional outputs. 
+The first output tensor 'Y' contains all unique values or subtensors of the input. 
+The second optional output tensor 'indices': indices[j] contains the index of the first occurrence of Y[j] in input X. 
+The third optional output tensor 'inverse_indices': inverse_indices[j] contains the index of the occurrence of X[j] in the output Y. 
+The fourth optional output tensor 'counts' contains the count of each element of 'Y' in the input. 
+
+Outputs are either sorted in ascending order or optionally in the order they occur in the input.
```
gramalingam: "in the order they occur in the input" => "in the order of the first occurrence of the values in the input" ?

onnx/backend/test/case/node/unique.py
```diff
@@ -0,0 +1,54 @@
...
+        y, indices, inverse_indices, counts = np.unique(x, True, True, True)
+        expect(node_sorted, inputs=[x], outputs=[y, indices, inverse_indices, counts], name='test_unique_sorted_without_axis')
+
+        node_not_sorted = onnx.helper.make_node(
+            'Unique',
+            inputs=['X'],
+            outputs=['Y', 'indices', 'inverse_indices', 'counts'],
+            sorted=0
+        )
+        # numpy unique does not retain original order (it sorts the output unique values)
```
gramalingam: Is it possible to add a python implementation for this case? E.g., can we use some sort-routine on inverse-indices to permute all the outputs to the required order?

onnx/defs/tensor/defs.cc
gramalingam: From what I observed when experimenting with np.unique, it looks like the result in 1D. When axis is not specified, it looks like a linear index. Actually, the whole thing (that is, the behavior of inverted indices as well) looks like the input is flattened, and then the basic "unique" is invoked on the flattened tensor. When axis is specified, index is naturally 1D since it is an index into only one axis.

liqunfu: Indices are always 1D of int64_t. It either points to a flattened input, when axis is not provided, or a slice along the given axis. Updated the comments to clear this confusion.

liqunfu:
@hariharans29 please update your cla

docs/TestCoverage.md
```diff
@@ -8300,6 +8300,116 @@ expect(node, inputs=[data], outputs=[transposed],
...
+    inputs=['X'],
+    outputs=['Y', 'indices', 'inverse_indices', 'counts'],
+    sorted=0
+)
+# numpy unique does not retain original order (it sorts the output unique values)
+# https://github.com/numpy/numpy/issues/8621
+# we need to recover unsorted output and indices
+y, indices, inverse_indices, counts = np.unique(x, True, True, True, axis=0)
+
+# prepare index mapping from sorted to unsorded
```
wschin:
```suggestion
# prepare index mapping from sorted values to unsorted values.
```

docs/TestCoverage.md
```diff
@@ -8300,6 +8300,116 @@ expect(node, inputs=[data], outputs=[transposed],
...
+argsorted_indices = np.argsort(indices)
+indices = np.sort(indices)
+inverse_indices_map = {i: si for i, si in zip(argsorted_indices, np.arange(len(argsorted_indices)))}
+
+y = np.asarray([x[indices[i]] for i in range(len(indices))], dtype=np.float32)
```
wschin: You can do
```python
y = y[argsorted_indices]
indices = y[argsorted_indices]
counts=y[argsorted_indices]

```

onnx/backend/test/case/node/unique.py
```diff
@@ -0,0 +1,107 @@
...
+
+        node_sorted = onnx.helper.make_node(
+            'Unique',
+            inputs=['X'],
+            outputs=['Y', 'indices', 'inverse_indices', 'counts']
+        )
+        y, indices, inverse_indices, counts = np.unique(x, True, True, True)
+        expect(node_sorted, inputs=[x], outputs=[y, indices, inverse_indices, counts], name='test_unique_sorted_without_axis')
+
+        node_not_sorted = onnx.helper.make_node(
```
wschin: Could you split this test into smaller pieces? One test function would be better to include only one node.

docs/TestCoverage.md
wschin: Can this reference implementation handle N-D tensor?

PR #2142, opened 2019-06-28
Add new op RoiCropAndResize to opset 11

jiafatom:
CropAndResize is very widely used op in object detection. 
We raised an [issue](https://github.com/onnx/onnx/issues/2100) to add this op. With offline discussion, we decide to add this op to onnx.

We already have an implementation [here](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/contrib_ops/cpu/crop_and_resize.cc) in onnxruntime as contrib op.


prasanthpul:
Can you explain why this should not be a function?

jiafatom:
> Can you explain why this should not be a function?

The algorithm logic for CropAndResize is so complicated that it is very challenging to use existing ops to construct it.

jiafatom:
cc @wschin @linkerzhang @ebarsoum

BowenBao:
Can you explain what is the difference between existing RoIAlign, and the reason why choosing not to add this as a mode for RoIAlign?

jiafatom:
> Can you explain what is the difference between existing RoIAlign, and the reason why choosing not to add this as a mode for RoIAlign?

For CropAndResize and RoiAlign, the algorithm to calculate the output is totally different. (1) They choose different locations in original input. (2) CropAndResize uses bilinear or nearest interpolation, but RoiAlign uses weighted sampling. Therefore, the output result is different. 

The reason not uses mode: (1) the implementation is totally different. So this is basically two algorithm. (2) the interface is similar, but we still see some attributes can only be uses one or the other: extrapolation_value only used in CropAndResize, sampling_ratio, spatial_scale only uses in RoiAlign.

BowenBao:
@jiafatom Thanks for the explanation. 

> > Can you explain what is the difference between existing RoIAlign, and the reason why choosing not to add this as a mode for RoIAlign?
> 
> For CropAndResize and RoiAlign, the algorithm to calculate the output is totally different. (1) They choose different locations in original input. (2) CropAndResize uses bilinear or nearest interpolation, but RoiAlign uses weighted sampling. Therefore, the output result is different.

For (1) What are the location differences? The crop_and_resize uses [y1,x1,y2,x2], and is normalized by image size v.s. RoiAlign [x1,y1,x2,y2], integer + optional spatial_scale. This part can be handled with transpose and mul/div. Are there other differences that I'm missing?

(2) Is there a way to represent the resizing in crop_and_resize/RoiAlign using existing onnx ops? Like Resize/AveragePool?

> The reason not uses mode: (1) the implementation is totally different. So this is basically two algorithm. (2) the interface is similar, but we still see some attributes can only be uses one or the other: extrapolation_value only used in CropAndResize, sampling_ratio, spatial_scale only uses in RoiAlign.

It is confusing to add another op with a totally different name that does almost the same thing if ignoring the implementation details. Can we re-design both to be functions? 
```
# pre-process locations
for (size_t i = 0; i < num_of_rois; ++i) {
    cat(output, resize(crop(images[batch_indices[i]], rois[i]), ...), axis=0)
}
# post-process
```

jiafatom:
> @jiafatom Thanks for the explanation.
> 
> 
> Can you explain what is the difference between existing RoIAlign, and the reason why choosing not to add this as a mode for RoIAlign?
> 
> For CropAndResize and RoiAlign, the algorithm to calculate the output is totally different. (1) They choose different locations in original input. (2) CropAndResize uses bilinear or nearest interpolation, but RoiAlign uses weighted sampling. Therefore, the output result is different.
> 
> For (1) What are the location differences? The crop_and_resize uses [y1,x1,y2,x2], and is normalized by image size v.s. RoiAlign [x1,y1,x2,y2], integer + optional spatial_scale. This part can be handled with transpose and mul/div. Are there other differences that I'm missing?
> (2) Is there a way to represent the resizing in crop_and_resize/RoiAlign using existing onnx ops? Like Resize/AveragePool?
> 
> The reason not uses mode: (1) the implementation is totally different. So this is basically two algorithm. (2) the interface is similar, but we still see some attributes can only be uses one or the other: extrapolation_value only used in CropAndResize, sampling_ratio, spatial_scale only uses in RoiAlign.
> 
> It is confusing to add another op with a totally different name that does almost the same thing if ignoring the implementation details. Can we re-design both to be functions?
> # pre-process locations
> for (size_t i = 0; i < num_of_rois; ++i) {
>     cat(output, resize(crop(images[batch_indices[i]], rois[i]), ...), axis=0)
> }
> # post-process

For location, it means the location of the input tensor corresponded to the output tensor, not the roi location mentioned here. Actually it is that the algorithm is totally different, so the code is actually different and cannot be reused.

BowenBao:
> For location, it means the location of the input tensor corresponded to the output tensor, not the roi location mentioned here. Actually it is that the algorithm is totally different, so the code is actually different and cannot be reused.

After offline discussion, the main difference above is from the resize step. 

If the resize step in crop_and_resize is equivalent to onnx::resize, then we should have enough operators to represent crop_and_resize. We don't need to add operator crop, onnx::slice should be enough. The algorithm might be too complicated for the current function design, as it requires roughly more than 10 operators. 

I'd suggest renaming the proposed op CropAndResize to something closer to RoIAlign.

jiafatom:
@wschin @linkerzhang @ebarsoum can we proceed this PR? any comments?

ebarsoum:
@jiafatom  what the difference between the resize in resize_and_crop and onnx:resize?  Also, I agree with bowen the name should change (i.e. RoiCropAndResize).

onnx/backend/test/case/node/cropandresize.py
```diff
@@ -0,0 +1,78 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
```
ebarsoum: Why one test case? What about with different extrapolation_value?

onnx/backend/test/case/node/cropandresize.py
```diff
@@ -0,0 +1,78 @@
...
+
+class CropAndResize(Base):
+    @staticmethod
```
ebarsoum: Can you add a reference implementation?

docs/Changelog.md
```diff
@@ -10332,6 +10332,57 @@ This version of the operator has been available since version 11 of the default
 <dd>Constrain input and output types to integer tensors.</dd>
 </dl>
 
+### <a name="CropAndResize-11"></a>**CropAndResize-11**</a>
+
+  Extracts crops from the input image tensor and resizes them using bilinear sampling or nearest neighbor sampling
+  (possibly with aspect ratio change) to a common output size specified by crop_height and crop_width.
+  Returns a tensor with crops from the input image at positions defined at the bounding box locations in boxes.
```
ebarsoum: Add link to the equivalent TensorFlow version.

jiafatom:
> @jiafatom  what the difference between the resize in resize_and_crop and onnx:resize?  Also, I agree with bowen the name should change (i.e. RoiCropAndResize).

onnx:resize treats the input tensor as a whole (no crop there).  crop_and_resize "crops" the input tensor based on input rois first, and then do resize. In fact, crop_and_resize does not need generate the cropped image, but calculated the image coordinate instead.

docs/Changelog.md
```diff
@@ -10332,6 +10332,57 @@ This version of the operator has been available since version 11 of the default
...
+  The cropped boxes are all resized (with bilinear or nearest neighbor interpolation) to
+  a fixed size = [crop_height, crop_width]. The result is a 4-D tensor [num_boxes, crop_height, crop_width, depth].
+  The resizing is corner aligned.
+
```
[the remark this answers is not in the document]
jiafatom: added.

docs/Changelog.md
```diff
@@ -10332,6 +10332,57 @@ This version of the operator has been available since version 11 of the default
...
+#### Version
+
+This version of the operator has been available since version 11 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>extrapolation_value</tt> : float (default is 0.0)</dt>
+<dd>Value used for extrapolation, when applicable. Default is 0.0f. </dd>
```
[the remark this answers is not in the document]
jiafatom: For each element in the output tensor, crop_and_resize finds its corresponding coordinates in the original input tensor (image). If this coordinate is out of bound (for example, < 0), the extrapolation value is applied. For each element Y[m, c, ch, cw] in the output tensor, the image coordinate in the original X, (ch_x, cw_x), is calculated at first, based on ch, cw, R[m, :], and B[m]. Then Y[m, c, ch, cw] = mode(X[B[m], c, ch_x, cw_x], extrapolation_value).

docs/Changelog.md
jiafatom: added

onnx/backend/test/case/node/cropandresize.py
jiafatom: CropAndResize is now a contrib op in onnxruntime, see the C++ implementation here:
https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/contrib_ops/cpu/crop_and_resize.cc
Do you mean adding a python implementation? The code is a bit long and this node cannot be easily composed by existing onnx ops. Take existing onnx RoiAlign op for example, there is no reference implementation either because code is long. 

This is a unit test, so the test example is already good to guard the result. In addition, I provide the link for tensorflow implementation. Then it should be enough.

jiafatom: Will add

onnx/backend/test/case/node/cropandresize.py
jiafatom: Will add

docs/Changelog.md
```diff
@@ -10549,6 +10549,62 @@ This version of the operator has been available since version 11 of the default
...
```
ebarsoum: Can you describe `extrapolation_value` in more details?

docs/Changelog.md
```diff
@@ -10549,6 +10549,62 @@ This version of the operator has been available since version 11 of the default
...
 
+### <a name="RoiCropAndResize-11"></a>**RoiCropAndResize-11**</a>
+
+  Extracts crops from the input image tensor and resizes them using bilinear sampling or nearest neighbor sampling
+  (possibly with aspect ratio change) to a common output size specified by crop_height and crop_width.
+  Returns a tensor with crops from the input image at positions defined at the bounding box locations in boxes.
+  The cropped boxes are all resized (with bilinear or nearest neighbor interpolation) to
+  a fixed size = [crop_height, crop_width]. The result is a 4-D tensor [num_boxes, crop_height, crop_width, depth].
+  The resizing is corner aligned. Suppose the input tensor X of shape (N, C, H, W), and the input rois tensor R of shape (M, 4),
+  and the batch indices tensor B of shape (M, ). The shape of output tensor Y is (M, C, CH, CW) where CH and CW is the crop_height
```
ebarsoum: Add equation or pseudocode?

jiafatom: I have the python implementation [here](https://github.com/onnx/onnx/pull/2142/files/4f82d4f9f37e8980e29320fc20cc1f2a7026e412#diff-bd6ef8521750dfb968d157511646c345), which is very lengthy. Here I already mentioned `Y[m, c, ch, cw] = mode(X[B[m], c, ch_x, cw_x], extrapolation_value)`, it is clear now.

docs/Changelog.md
jiafatom: This description here is same as tensorflow there. I can add "when the corresponding input coordinate is not available" (because of out of tensor bound) there.

ebarsoum: I know that the description match TF one, but it is lacking. Can you add more details?

jiafatom: This [PR](https://github.com/onnx/onnx/pull/2057) will add crop to Resize op, so the extrapolation_value will be used there. The plan is that RoiCropAndResize will be a function of Resize. Does it make sense? I will hold this PR until Resize is finalized.

postrational:
Since the update to Resize op #2057 was merged, would it now be possible to express this logic as a `Function`?

Regardless of how complex the `Function` may be, if it is possible to express this in terms of simpler ops, we should always provide this as a form of reference implementation.

If it's not possible, what specifically is blocking it?

postrational:
Since #2057 was merged, it should now be possible to implement this as an ONNX `Function`.

Please refactor the PR to do this and we can add it to version 1.7 of ONNX.

jiafatom:
Had offline discussion on this. There is debating discussion on whether to implement it as a function. We have a custom op RoiAlign to make mask-rcnn model work. So we hold this implementation.

PR #2143, opened 2019-06-28
Add GatherElements Op and Rename ScatterElements

lara-hdr:
ONNX Gather supports a particular semantic used in certain frameworks (e.g. Caffe), but there are other “flavors” of Gather-like operators that differ from ONNX Gather and that cannot be supported today.

Also, Gather and Scatter operators are supposed to be inverses of each other, but the semantics of ONNX Scatter and ONNX Gather are misaligned (they are not inverse pairs).

This PR introduces a new operator GatherElements which is the inverse of the current ONNX Scatter, and aligns with the PyTorch Gather (https://pytorch.org/docs/stable/torch.html?highlight=gather#torch.gather).

This PR also deprecates ONNX Scatter by changing its name to ScatterElements to emphasize that these two operators are inverse pairs.


lara-hdr:
@gramalingam @linkerzhang could you please take a look at this PR?

onnx/defs/tensor/defs.cc
```diff
@@ -735,7 +735,8 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
         }));
 
-static const char* Scatter_ver9_doc = R"DOC(
+static const char* Scatter_ver11_doc = R"DOC(
+This operator is deprecated and renamed ScatterElements.
```
gramalingam: Minor rewording suggestion: "This operator is deprecated. Please use ScatterElements, which provides the same functionality."

onnx/defs/tensor/defs.cc
```diff
@@ -915,6 +991,82 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
         }));
 
+static const char* GatherElements_ver11_doc = R"DOC(
```
gramalingam: Here is a suggested rewording of the documentation:

GatherElements takes two inputs `data` and `indices` of the same rank r >= 1 and an optional attribute axis that identifies an axis of `data` (by default, the outer-most axis, that is axis 0). It is an indexing operation that produces its output by indexing into the input `data` tensor at index positions determined by elements of the `indices` tensor. Its output shape is the same as the shape of `indices` and consists of one value (gathered from the `data`) for each element in `indices`.

For instance, in the 3-D case (r = 3), the output produced is determined by the following equations:
```
  out[i][j][k] = input[index[i][j][k]][j][k]  # if dim == 0
  out[i][j][k] = input[i][index[i][j][k]][k]  # if dim == 1
  out[i][j][k] = input[i][j][index[i][j][k]]  # if dim == 2
```
This operator is the inverse of ScatterElements.

Example 1:
```
  data = [
      [1, 2],
      [3, 4],
  ]
  indices = [
      [0, 0],
      [1, 0],
  ]
  axis = 1
  output = [
      [
        [1, 1],
        [4, 3],
      ],
  ]
```

Example 2:
```
  data = [
      [1, 2, 3],
      [4, 5, 6],
      [7, 8, 9],
  ]
  indices = [
      [1, 2, 0],
      [2, 0, 0],
  ]
  axis = 0
  output = [
      [
        [4, 8, 3],
        [7, 2, 3],
      ],
  ]
  ```

Comment (gramalingam) [reply onnx/defs/tensor/defs.cc]: Please use ``` for blocks above. Dot dot dot does not seem to have the same effect.

Comment (gramalingam) [reply onnx/defs/tensor/defs.cc]: Also suggest adding "This operation is similar to Torch's gather operation."

File: onnx/defs/tensor/defs.cc [hunk 3]
```diff
@@ -773,9 +774,84 @@ Example 2:
...
+            "Constrain indices to integer types")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          if (hasNInputShapes(ctx, 1)) {
+            propagateShapeFromInputToOutput(ctx, 0, 0);
+          }
+        }));
+
+static const char* ScatterElements_ver11_doc = R"DOC(
+Given `data`, `updates` and `indices` input tensors of rank r >= 1, write the values provided by `updates`
```
gramalingam: Suggested rewording:

ScatterElements takes three inputs `data`, `updates`, and `indices` of the same rank r >= 1 and an optional attribute axis that identifies an axis of `data` (by default, the outer-most axis, that is axis 0). The output of the operation is produced by creating a copy of the input `data`, and then updating its value to values specified by `updates` at specific index positions specified by `indices`. Its output shape is the same as the shape of `data`.

For each entry in `updates`, the target index in `data` is obtained by combining the corresponding entry in `indices` with the index of the entry itself: the index-value for dimension = axis is obtained from the value of the corresponding entry in `indices` and the index-value for dimension != axis is obtained from the index of the entry itself.

For instance, in a 2-D tensor case, the update corresponding to the [i][j] entry is performed as below:
```
   output[indices[i][j]][j] = updates[i][j] if axis = 0, 
   output[i][indices[i][j]] = updates[i][j] if axis = 1,
```
This operator is the inverse of GatherElements. It is similar to Torch's Scatter operation.

Example 1:
```
  data = [
      [0.0, 0.0, 0.0],
      [0.0, 0.0, 0.0],
      [0.0, 0.0, 0.0],
  ]
  indices = [
      [1, 0, 2],
      [0, 2, 1],
  ]
  updates = [
      [1.0, 1.1, 1.2],
      [2.0, 2.1, 2.2],
  ]
  output = [
      [2.0, 1.1, 0.0]
      [1.0, 0.0, 2.2]
      [0.0, 2.1, 1.2]
  ]
```

Example 2:
```
  data = [[1.0, 2.0, 3.0, 4.0, 5.0]]
  indices = [[1, 3]]
  updates = [[1.1, 2.1]]
  axis = 1
  output = [[1.0, 1.1, 3.0, 2.1, 5.0]]
  ```

File: onnx/defs/tensor/defs.cc [hunk 4]
```diff
@@ -773,9 +775,99 @@ Example 2:
...
+is performed as below:
+```
+  output[indices[i][j]][j] = updates[i][j] if axis = 0, 
+  output[i][indices[i][j]] = updates[i][j] if axis = 1,
+```
+
+This operator is the inverse of GatherElements. It is similar to Torch's Scatter operation.
+
+Example 1:
+  data = [
```
gramalingam: Minor nit: Adding block quote ``` before the "data = [" line and closing it at the end of the example will help format the example. Same comment for all examples.

onnx/defs/tensor/defs.cc
```diff
@@ -735,7 +735,9 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
         }));
 
-static const char* Scatter_ver9_doc = R"DOC(
+static const char* Scatter_ver11_doc = R"DOC(
+This operator is deprecated. Please use ScatterElements, which provides the same functionality.
+
```
gramalingam: I think it would help to share the updated documentation used for ScatterElements for Scatter also. It will help make it clear they have the same definition.

onnx/defs/tensor/defs.cc
lara-hdr: I initially left it as is to be consistent with the other examples in this file.
But just updated all the gather/scatter ops to have the quotes.

onnx/defs/tensor/defs.cc
lara-hdr: you mean replace scatter's definition by the new definition, of just mention that the definition is updated in ScatterElements?

gramalingam: I mean the first (replace the old documentation with the new documentation), with the deprecation line alone being extra.

lara-hdr:
@linkerzhang could you take a look?
I'll rebase and resolve the conflict once reviewed to avoid rebasing continuously :)

spandantiwari:
@postrational - could you please help review this PR. We are looking to merge this soon as we are blocked by it. We are waiting to get two approvals. Thanks.

onnx/backend/test/case/node/scatterelements.py
```diff
@@ -0,0 +1,82 @@
...
+            'ScatterElements',
+            inputs=['data', 'indices', 'updates'],
+            outputs=['y'],
+            axis=axis,
+        )
+        data = np.array([[1.0, 2.0, 3.0, 4.0, 5.0]], dtype=np.float32)
+        indices = np.array([[1, 3]], dtype=np.int64)
+        updates = np.array([[1.1, 2.1]], dtype=np.float32)
+
+        y = scatter_elements(data, indices, updates, axis)
```
wschin: Please add a comment to describe the value of `y`, for example,
```suggestion
        y = scatter_elements(data, indices, updates, axis)
        // print(y) produces
        // [...
```

spandantiwari:
Thanks @gramalingam for all the discussion about Gather/Scatter ops and for reviewing and merging this. 

Thanks @wschin for your review.

PR #2146, opened 2019-07-02
typing is not needed since Python 3.5 as it's built-in

yan12125:


yan12125:
Tests timed out on xcode9.3 image with `PYTHON_VERSION=python3 ONNX_ML=0`. I think it's due to instability of Travis CI workers and not relevant to this change.

linkerzhang:
PR #2199 made similar fix and merged. Thank you! @yan12125

yan12125:
> PR #2199 made similar fix and merged.

Thanks for the info! That's a good news :)

PR #2148, opened 2019-07-02
Add helper function update_inputs_outputs_dims to tools

BowenBao:
The function updates the dimension of the inputs and outputs of the model, to the provided values in the parameter. These values can be either static or dynamic. This can be useful for extending models with fixed input size to support dynamic input sizes.


docs/PythonAPIOverview.md
```diff
@@ -214,3 +214,18 @@ import onnx.utils
 model = onnx.load('path/to/the/model.onnx')
 polished_model = onnx.utils.polish_model(model)
 ```
+
+### Updating Model's Inputs Outputs Dimension Sizes with Variable Length
```
houseroad: props for the doc

onnx/tools/update_model_dims.py
```diff
@@ -0,0 +1,59 @@
...
+
+def update_inputs_outputs_dims(model, input_dims, output_dims):  # type: (ModelProto, List[List[Any]], List[List[Any]]) -> ModelProto
+    """
+        This function updates the dimension sizes of the model's inputs and outputs to the values
+        provided in input_dims and output_dims. if the dim value provided is negative, a unique dim_param
+        will be set for that dimension.
+
+        Example. if we have the following shape for inputs and outputs:
+                shape(input_1) = ('b', 3, 'w', 'h')
+                and shape(input_2) = ('b', 4)
```
wschin:
```suggestion
                shape(input_2) = ('b', 4)
```

onnx/tools/update_model_dims.py
```diff
@@ -0,0 +1,59 @@
...
+                and shape(output)  = ('b', 'd', 5)
+
+            The parameters can be provided as:
+                input_dims = [
+                    ['b', 3, 'w', 'h'],
```
wschin: nit: do you think using dictionary for `input_dims` and `output_dims` would be better? It could reduce the likelihood of making mistakes.

onnx/tools/update_model_dims.py
```diff
@@ -0,0 +1,59 @@
...
+                    ['b', -1, 5]
+                ]
+
+            Putting it together:
+                model = onnx.load('model.onnx')
+                updated_model = update_inputs_outputs_dims(model, input_dims, output_dims)
+                onnx.save(updated_model, 'model.onnx')
+    """
+    def update_dim(tensor, dim, i, j, dim_param_prefix):  # type: (ValueInfoProto, Any, int, int, Text) -> None
+        dim_proto = tensor.type.tensor_type.shape.dim[j]
```
wschin: Would you check if the existing shape and the provided shape are compatible? Changing `[-1, 3]` to `[2, 3]` is reasonable but I am not sure if you want to allow `[1, 2, 3]` to `[4, 5, 6]`.

prasanthpul:
Please sign the updated CLA

onnx/tools/update_model_dims.py
```diff
@@ -0,0 +1,59 @@
...
+        if isinstance(dim, int):
+            if dim >= 0:
+                dim_proto.dim_value = dim
+            else:
+                dim_proto.dim_param = dim_param_prefix + str(i) + '_' + str(j)
```
gramalingam: Hi: this can cause potential problems if the generated symbolic name happens to be the same as the name of a symbolic dimension used elsewhere. I suggest either using "dim_proto.ClearField("dim_param"); dim_proto.ClearField("dim_value")" to clear out both fields or raising an exception here.

onnx/tools/update_model_dims.py
gramalingam: I agree. A dictionary would be better.

onnx/tools/update_model_dims.py
BowenBao: good catch, added.

onnx/tools/update_model_dims.py
BowenBao: indeed, updated to dictionary

onnx/tools/update_model_dims.py
BowenBao: Thanks, I have updated to raising an exception here.

onnx/tools/update_model_dims.py
```diff
@@ -22,38 +22,62 @@ def update_inputs_outputs_dims(model, input_dims, output_dims):  # type: (ModelP
...
+
+    init_dim_param_set(dim_param_set, model.graph.input)
+    init_dim_param_set(dim_param_set, model.graph.output)
+    init_dim_param_set(dim_param_set, model.graph.value_info)
+
+    def update_dim(tensor, dim, j, name):  # type: (ValueInfoProto, Any, int, Text) -> None
         dim_proto = tensor.type.tensor_type.shape.dim[j]
         if isinstance(dim, int):
             if dim >= 0:
+                import pdb; pdb.set_trace()
```
wschin: pdb? Is it needed?

BowenBao: guess not ...

PR #2155, opened 2019-07-08
Typo fix in Python API Example

LeicongLi:
In the API example, the code shows how to convert Numpy Array to TensorProto, and then convert back to Numpy Array.
For the step converting TensorProto back to Numpy Array, the example prints the original "numpy_array" which I think its more reasonable to print "new_array", which is created to store the converted Numpy Array.

Thanks.


linkerzhang:
@LeicongLi  would you mind to sign the CLA?

llc-intel:
@linkerzhang Thanks! CLA has been signed.

Issue #2020, opened 2019-05-15
GroupNormalization Op


hobei:
Add support for group normalization as defined in https://arxiv.org/abs/1803.08494. 
The new operator will be similar to the existing BatchNormalization & InstanceNormalization

wschin:
According to the reference's Figure 3, this op can be composed by primitives without using any control flow op. Therefore, I'd recommend to propose it as a `FunctionProto` if we really need it.

PR #2160, opened 2019-07-10
Add group normalization op

hobei:
PR for #2020.

As suggested I have defined group normalization as a operation with a registered function. I would appreciate feedback on
1. Converting the GroupNormalization attributes (eps & num_groups) into tensors that can be used by the FunctionBody. I have added a new MakeRefAttribute function to create reference attributes that differ by name and added conversions from 'float' to 'tensor(float)'. Glad to hear alternative approaches.
2. I have modified the test/runner/__init__.py to allow the devices under test to be overridden.

Still to be done 
1. Add to opset 12


hobei:
I also wonder if it would make sense to change the MeanVarianceNormalization function op to output the mean & var, then I could use that from the GroupNormalization function op?

onnx/defs/nn/defs.cc
```diff
@@ -1757,6 +1757,104 @@ ONNX_OPERATOR_SET_SCHEMA(
           propagateShapeAndTypeFromFirstInput(ctx);
         }));
 
+
+
+static const char* GroupNormalization_ver11_doc = R"DOC(
+Carries out group normalization as described in the paper
+https://arxiv.org/abs/1803.08494.
```
wschin: Could you please add some equations to describe this operator's behavior? Copying-and-pasting from the paper is fine as long as all symbols are well-defined.

onnx/defs/nn/defs.cc
```diff
@@ -1757,6 +1757,104 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(2, "B", "The input 1-dimensional bias tensor of size C.", "T")
+        .Output(0, "output", "The output tensor of the same shape as input", "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)"},
+            // "tensor(double)" - Can not hve tensor(double) because that is not supported by ConstOfShape
+            // But the VerifyTypeConstraint test assumes that the input type is used on the op, which is incorrect
+            "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateShapeAndTypeFromFirstInput(ctx);
```
wschin: Do you have shape inference test?

docs/Operators.md
```diff
@@ -5125,6 +5126,166 @@ expect(node, inputs=[x, y], outputs=[z],
...
+```python
+
+# A 4 dimension
+x = np.random.rand(2, 4, 3, 6, 4, 2).astype(np.float32)
+
+b = np.array([0, 0, 0, 0]).astype(np.float32)
+s = np.array([1, 1, 1, 1]).astype(np.float32)
+
+num_groups = 2
+eps = 1e-05
```
postrational: Let's expand `eps` to `epsilon` for code clarity.

onnx/defs/reduction/defs.cc
```diff
@@ -28,6 +28,13 @@ False instead of True.)DOC";
         "Keep the reduced dimension or not, default 1 mean keep reduced dimension.",
         AttributeProto::INT,
         static_cast<int64_t>(1));
+    if (version == 11) {
+      schema.Attr(
+          "exclude_axes",
+          "Keep the reduced dimension or not, default 1 mean keep reduced dimension.",
```
postrational: The description does not explain how this is different from `keep_dims`

hobei:
Thank you @wschin I will try that suggestion and you are right that it will not require a change to the ReduceMean, I had originally developed it with 4D inputs.

onnx/defs/reduction/defs.cc
hobei: Have removed this attribute from the reduction in the latest pull request

docs/Operators.md
hobei: Done

onnx/defs/nn/defs.cc
hobei: Yes, I have added a shape inference test

onnx/defs/nn/defs.cc
hobei: I am not certain what information from the paper you are requesting.

PR #2163, opened 2019-07-11
depth to space shuffle order

neginraoof:
Added the "mode" attribute to be able to set the order for DepthToSpace shuffle.
DCR (default) is for depth-column-row order.
Using CRD mode, depthToSpace would be an equivalent of pytorch pixel_shuffle, and could be used for export.


gramalingam:
Hi, can you please update the op documentation to clarify what the shuffle order means? Ideally, it would be great to add an equation of the form below, filling in what goes into "..."
```
    output[i,j,k,m] = input[...]
```

gramalingam:
Thanks for adding the description. I think that the operator documentation needs to be regenerated and committed. Please see: https://github.com/onnx/onnx/blob/master/docs/CONTRIBUTING.md#generated-operator-documentation

onnx/defs/tensor/defs.cc
```diff
@@ -1185,22 +1185,115 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 
-static const char* DepthToSpace_ver1_doc =
+static const char* DepthToSpace_ver11_doc =
     R"DOC(DepthToSpace rearranges (permutes) data from depth into blocks of spatial data.
 This is the reverse transformation of SpaceToDepth. More specifically, this op outputs a copy of
 the input tensor where values from the depth dimension are moved in spatial blocks to the height
-and width dimensions.
+and width dimensions. By default, `mode` = `DCR`. So elements along the depth dimension from the input tensor are
+rearranged in the following order: depth, column, and then row in the output.
+
```
gramalingam: Hi, is this addition showing up in the operators.md? I can't quite see it. I think that for the documentation, we could describe the logic as below:
In the default DCR mode, the output y is computed from the input x as below:
```python
   b, c, h, w = x.shape 
   tmp = np.reshape(x, [b, blocksize, blocksize, c // (blocksize**2), h, w])
   tmp = np.transpose(tmp, [0, 3, 4, 1, 5, 2])
   y = np.reshape(tmp, [b, c // (blocksize**2), h * blocksize, w * blocksize])
```

In the CRD mode, the output y is computed from the input x as below:
```python
   b, c, h, w = x.shape 
   tmp = np.reshape(x, [b, c // (blocksize ** 2), blocksize, blocksize, h, w])
   tmp = np.transpose(tmp, [0, 1, 4, 2, 5, 3])
   y = np.reshape(tmp, [b, c // (blocksize ** 2), h * blocksize, w * blocksize])
```

gramalingam: If the example is useful, that can be added separately, following the above as "For example, for an input value of …, the output valus is …"

neginraoof: Sounds good. I modified the description. You should be able to see it on operators.md now

neginraoof:
@ebarsoum could you please review this PR?

gramalingam:
@houseroad : any comments/feedback on this op extension? Thanks.

neginraoof:
@gramalingam @houseroad can we merge this?

PR #2133, opened 2019-06-26
Add more shape inference tests for Logical operators to improve coverage

hariharans29:
This change strives to add more shape inference tests for `Logical` ONNX operators

PR #2167, opened 2019-07-12
Fix errors in RoiAlign shape inference code

gramalingam:
Fix errors in RoiAlign shape inference. The existing version has the following issues:
(a) It assumes that the attributes for height and width are present, but they are optional. This will cause a null-dereference.
(b) It assumes that the input-shape dimensions are all known values, but they could be unknown or symbolic dimensions. This will cause output dimensions to be inferred as zero incorrectly.

This PR also adds shape-inference-tests to catch above two issues.

This PR also adds some utility functions to help simplify writing shape-inference.

### DIFF
--- a/onnx/defs/object_detection/defs.cc
+++ b/onnx/defs/object_detection/defs.cc
@@ -24,109 +24,107 @@ ONNX_OPERATOR_SET_SCHEMA(
     RoiAlign,
     10,
     OpSchema()
-      .SetDoc(RoiAlign_ver1_doc)
-      .Attr(
-          "spatial_scale",
-          "Multiplicative spatial scale factor to translate ROI coordinates "
-          "from their input spatial scale to the scale used when pooling, "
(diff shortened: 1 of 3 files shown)


gramalingam:
Close and reopen to trigger CI

gramalingam:
I hadn't noticed that @hariharans29 had already created a PR for shape-inference for the same op at https://github.com/onnx/onnx/pull/2133 : I will talk with him to resolve these PRs.

gramalingam:
I had an offline discussion with Hari. The purpose of the current PR is to introduce unification utilities to help simplify our shape-inference code. It uses RoiAlign to illustrate the usage. Hari offered to reconcile these changes with his own PR. @hariharans29 : please take a look at this PR, thanks.

hariharans29:
Looks good. The simple helpers make it easy to review the overall logic :). I will try and adopt these in my PRs and contribute more helpers if need be. Thanks!

PR #2168, opened 2019-07-14
[WIP][Training] Draft of Gradient Operator

wschin:
**PR #2314 is a single place for reviewing the whole training story.**

In Tensorflow and Pytorch, most models are trained using backpropagation and gradient-based optimization algorithms. This operator provides a way to compute gradient given an inference graph.

- [x] Signature and documentation.
- [ ] Shape inference code
- [ ] Reference implementation
- [x] Tests
- [ ] Verification script

[Here](https://pytorch.org/docs/stable/_modules/torch/autograd.html#grad) is a similar operator in Pytroch.


onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+inputs specified by "xs" and outputs containing "y"'s content and do a backward
+pass (or other auto-differentiation techniques whenever needed).
+
+Since X is not trainable tensor, the user can avoid the creation of dY/dX by
+assigning an empty string to the 1st output name of that Gradient. In other
+words, all Gradient's outputs are optional. Note that the concept of optional
+outputs can also be found in ONNX's RNN, GRU, and LSTM.
+
+Gradient operator can compute derivative against intermediate tensors. For
+example, the gradient of Y with respect to W can be done in
```
wschin:
```suggestion
example, the gradient of Y with respect to H can be done in
```

onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+// Licensed under the MIT license.
+
+#include <algorithm>
+#include <cmath>
+#include "onnx/defs/function.h"
+#include "onnx/defs/schema.h"
+
+namespace ONNX_NAMESPACE {
+static const char* Gradient_ver12_doc = R"DOC(
+Gradient operator computes the derivatives of some tensors with respect to
```
gramalingam: change to "the partial derivatives of a specified tensor with respect to some other tensors"? I think of "dy/dx" as the derivative of y with respect to x.

onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+```
+W --> Gemm --> Y --> Loss --> O
+|      ^              ^
+|      |              |
+|      X              L
+|      |
+|      |
++------+--> Gradient(xs=["X", "W"], y=["Y"]) ---> dO/dX (1st output of Gradient)
+|      |      |
+|      |      '---> dO/dW (2nd output of Gradient)
```
gramalingam: I suggest replacing "dO/dW" by a valid identifier, like "dO_dW" or "first_order_gradient".

onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+|      v
+'---> Gradient(xs=["X", "W"], y=["dO/dW"]) ---> d(dO/dW)dX (1st output of
+       |                                          Gradient)
+       |
+       |
+       '---> d^2O/dW^2 (2nd output of Gradient)
```
gramalingam: Suggest using a valid identifier (as above).

onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+           v
+W_1 --> Gradient(xs=["H", "Z"], y=["Y"]) ---> dY/dX when Y = Y_1
+         |   |
+         |   '-----------------------------------> dY/dW (2nd output of Gradient)
+         |
+         '---------------------------------------> dY/dZ (3rd output of Gradient)
+```
+
+This graph also implies that one can identify a gradient graph once and reuse
+it multiple times with different input values.
```
gramalingam: How? We don't have a way to give a name to the gradient-graph and reuse it. Do you mean that we can have multiple gradient nodes with the same attribute values?

onnx/defs/training/defs.cc
gramalingam: It might be worth adding an explanatory note somewhere, something like "The gradient operator is slightly different from the standard operators. It is conceptually a higher-order operator that takes a computation-graph as an attribute, but we abuse the attribute mechanism to describe the computation-sub-graph implicitly by listing the inputs and output of the computation-sub-graph in the current graph.'

onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+'---> Gradient(xs=["X", "W"], y=["dO/dW"]) ---> d(dO/dW)dX (1st output of
+       |                                          Gradient)
+       |
+       |
+       '---> d^2O/dW^2 (2nd output of Gradient)
+```
+
+As mentioned above, the attributes "xs" and "y" are used to identify a graph,
+and we can feed different tensors to the identified graph. For,
+example, one can compute the gradient of Y with respect to H by substituting
```
gramalingam: Are there typos here? I am a bit confused, because I am not sure what is Y_1, H_1, and the picture below shows W_1 and Z_1. Are you trying to say that it is not necessary for the inputs to be identical to the list xs specified in the attributes?

gramalingam: If the above is the case, it would be helpful to add that "The implementation can be optimized in the common case where the inputs are identical to the attributes." … because if the inputs are not identical to the attributes, the implementation has to do extra work to create a duplicate of the forward-graph for the actual-inputs.

onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+            "Input tensor names of the differentiated sub-graph. It "
+            "contains and only contains the necessary inputs of a "
+            "(sub-)graph. Variables (usually called intermediate "
+            "variables) which can be generated by inputs cannot be "
+            "included in this attribute.",
+            AttributeProto::STRINGS)
+        .Attr(
+            "y",
+            "The differentiated tensor. It is viewed as a function of "
+            "the attribute \"xs\".",
```
gramalingam: change "the attribute" to "the tensors named in the attribute"

onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+            "V",
+            OpSchema::Variadic,
+            false)
+        .Attr(
+            "xs",
+            "Input tensor names of the differentiated sub-graph. It "
+            "contains and only contains the necessary inputs of a "
+            "(sub-)graph. Variables (usually called intermediate "
+            "variables) which can be generated by inputs cannot be "
+            "included in this attribute.",
```
gramalingam: Is it correct to say the following? If xs=[X_1, ..., X_m], then no X_i can depend on X_j (i <> j): that is, there cannot be a dependence-path in the graph from X_j to X_i? Furthermore, do we need to say the list in xs must be complete? If we have Y = X_1 + X_2, is it valid or invalid to specify xs=["X_1"]?

onnx/defs/training/defs.cc
```diff
@@ -0,0 +1,171 @@
...
+```
+W --> Conv --> H --> Gemm --> Y
+|      ^              ^
+|      |              |
+|      X              Z
+|      |              |
+|      |   .----------'
+|      |   |  (X/W/Z is the 1st/2nd/3rd input of Gradient as shown in "xs")
+|      v   v
+'---> Gradient(xs=["X", "W", "Z"], y=["Y"]) ---> dY/dX (1st output of Gradient)
```
gramalingam: Why is "y" shown as a list here? It is a single string according to the definition below.

onnx/defs/training/defs.cc
wschin: No problem.

onnx/defs/training/defs.cc
wschin: Humm, but `d(dO/dW)dX` would be changed to something like `d_dO_dW_dX` and become less readable.

onnx/defs/training/defs.cc
wschin: If you put `Gradient` in a loop, it can receive different values from loop-carried dependencies.

onnx/defs/training/defs.cc
wschin: Do you think we need to allow
```
 if the inputs are not identical to the attributes, the implementation has to do extra work to create a duplicate of the forward-graph for the actual-inputs
```
? Looks like it's not necessary.

onnx/defs/training/defs.cc
wschin: Ok.

onnx/defs/training/defs.cc
wschin: Yes. The inputs named in `xs` must be independent variables; that is, they cannot be upstream variable of each other.

The list must be complete. Otherwise, we cannot finish a forward path due to missing input (in your example, we cannot compute `Y` without `X_2`).

onnx/defs/training/defs.cc
wschin: Fixed. Thanks a lot for poiniting this out.

onnx/defs/training/defs.cc
gramalingam: If we don't support it, that's okay with me (since I can't immediately think of use-cases where we would want to use it). But, then, the documentation should make this explicit (and implementations should check that the input-list and attribute-list match).

onnx/defs/training/defs.cc
gramalingam: I see your point. But my concern is that some users will copy this blindly and use "dO/dW" as a tensor name. This can lead to potential problems downstream. Technically, the ONNX documentation says that tensor names should be valid C identifiers. If the ONNX checker were to enforce this, that would be somewhat helpful. But it does not seem to do that. (I remember Niklas was trying to add such a check ... but it doesn't seem to exist today … not sure why. Possibly because a number of models were violating this then.) We should probably sort this out one way or the other in the ONNX standard. Until then, I think it would be better to try to stick to the identifier restriction. 

We could use simply use names like D1 and D2 to denote the first and second derivative in the example (and in the text we could say D1 denotes dO/dW, for example).

docs/Operators.md
```diff
@@ -5044,6 +5045,158 @@ expect(node, inputs=[x], outputs=[y], name='test_globalmaxpool_precomputed')
...
+  '---> Gradient(xs=["X", "W", "Z"], y="Y") ---> dY/dX (1st output of Gradient)
+         |   |
+         |   '-----------------------------------> dY/dW (2nd output of Gradient)
+         |
+         '---------------------------------------> dY/dZ (3rd output of Gradient)
+  ```
+  
+  , where the content inside the braces of Gradient are attributes. The attribute
+  "xs" specifies the necessary inputs to compute the variable specified by
+  attribute "y". To compute gradient, the runtime should identify a sub-graph with
```
newling: This feels to me like an implementation detail, not a definition of what the Gradient Operator does.

wschin:
The operator `Gradient` is now a part of #2314 so this PR can be closed.

docs/Operators.md
wschin: Removed. Thanks!

PR #2171, opened 2019-07-16
pin virtualenv to old version

houseroad:
new version caused issues with --user.


.circleci/build.sh
```diff
@@ -19,6 +19,8 @@ if hash sccache 2>/dev/null; then
 fi
 
 # setup virtualenv
+pip uninstall virtualenv
```
bddppq: no need to uninstall, your next line will force the version.

PR #2176, opened 2019-07-17
add more types support for Equal op

linkerzhang:
Equal should support all numeric types.

### DIFF
--- a/onnx/defs/logical/defs.cc
+++ b/onnx/defs/logical/defs.cc
@@ -6,35 +6,35 @@
 namespace ONNX_NAMESPACE {
 
 inline void unaryLogicalOpInference(InferenceContext& ctx) {
-    updateOutputElemType(ctx, 0, TensorProto::BOOL);
-    if (hasInputShape(ctx, 0)) {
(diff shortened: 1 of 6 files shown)


onnx/defs/logical/defs.cc
```diff
@@ -109,13 +109,13 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-    7,
+    11,
     OpSchema()
         .FillUsing(BinaryLogicDocGenerator("equal"))
         .TypeConstraint(
             "T",
-            {"tensor(bool)", "tensor(int32)", "tensor(int64)"},
-            "Constrains input to integral tensors.")
+            OpSchema::all_numeric_types(),
+            "Constrains input types to all numeric tensors.")
```
wschin: Could you add shape inference and output tests for new types?

linkerzhang: I would like to add it when we have Alibaba's auto test-gen tool moved into ONNX org later, otherwise, it's duplicating work.

Issue #842, opened 2018-04-27
Negative Axis Support


houseroad:
Shall we support negative axis in ONNX spec?

This is unclear, and worth more discussion.

bddppq:
Yes we should. It's necessary to represent "from the end" in some cases with dynamic shape (e.g. Slice). And to make it consistent, we should allow negative axis everywhere. This also aligns with numpy semantics (which the Test and Compliance working group agreed on in last workshop).

linkerzhang:
does this mean we only want to have "-1" to represent sth, instead of having any negative value, right? btw, for unknown type, we have symbolic shape parameter for the purpose.

gramalingam:
It seems well-defined and useful. I am fine with supporting it (though it is a generalization of the current spec so we may want to add it to a suitable version of ONNX.) @linkerzhang , this seems orthogonal to the use of symbolic dimensions (or negative values) to represent statically unknown dimension values. This is allowing attributes that specify an axis to specify it from the left end or right end.

askhade:
Support for negative axis has been added. If there are any gaps then open a new issues to track it.

PR #2177, opened 2019-07-18
Handle negative 'axis' value in Split type and shape inferencing

skottmckay:
Handle a negative 'axis' value when running the type and shape inferencing for Split


hobei:
A test case (example) with a negative axis value would also be helpful and could be verified with the backend tests.

skottmckay:
> A test case (example) with a negative axis value would also be helpful and could be verified with the backend tests.

Added a test to the shape inferencing tests. Let me know if that's sufficient.

gramalingam:
Wouldn't it be better to bump the op version (opset) number? That would make sense if we treat this as an extension of the op's spec. If we treat this as a bug-fix, that's not needed, but it seems to me that treating it as an extension is somewhat better.

skottmckay:
> Wouldn't it be better to bump the op version (opset) number? That would make sense if we treat this as an extension of the op's spec. If we treat this as a bug-fix, that's not needed, but it seems to me that treating it as an extension is somewhat better.

Currently the ONNX Runtime implementation of Split handles a negative axis so fwiw this is just bringing the shape inferencing into line with that (not that an implementation gets to force changes of the spec though). There are also a number of other ops in the ONNX Runtime implementation where a negative axis is handled even though the spec for the op doesn't explicitly state whether that is allowed (Softmax/LogSoftmax/TopK/Scan/Reduce*), so it looks like this is another gray area of the ONNX spec which has been around a while. #842. 

There are only two places I could see where using a negative value for an axis was explicitly disallowed (Squeeze/Unsqueeze), so it seems like they are in general supported. Maybe the ONNX spec should say that is the case, unless explicitly disallowed, in order to clarify things. Not sure if that change requires an opset bump though given it's just being more flexible (assuming no ops change to explicitly disallow of course).

gramalingam:
Okay, thanks. Given the comments elsewhere that negative axis should be uniformly supported everywhere, it seems reasonable to treat this as a bug-fix.

onnx/defs/tensor/defs.cc
```diff
@@ -397,20 +397,25 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            int r = shape.dim_size();
+            int axis = static_cast<int>(getAttribute(ctx, "axis", 0));
+            if (axis < -r || axis >= r) {
+              fail_type_inference(
+                  "Invalid value of attribute 'axis'. Rank=",
+                  r,
+                  " Value=",
+                  axis);
+            }
+            if (axis < 0) {
```
linkerzhang: Thank you very much for the change! Please also add more clarification for "axis" attribute in the op definition, so that ONNX users do not need to find the negative axis handling logic from shape inference codes.

PR #2187, opened 2019-07-23
Add function to fuse dynamic quantization graph into 1 node

askhade:
During dynamic quantization we have to add around 12 nodes to calculate scale, zero point and then quantize the FP32 data to 8bit data. This incurs considerable cost during inference. Hence adding a function to fuse this sub graph into 1 node.


askhade:
FunctionVerification.VerifyFunctionOps test is failing because of a bug in the test itself. I will add a fix for this soon. In the meanwhile please proceed to review the code.

onnx/defs/quantization/defs.cc
```diff
@@ -101,4 +102,49 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          {"tensor(uint8)"},
+          "Constrain 'Y_Zero_Point' and 'Y' to 8-bit unsigned integer tensor.")
+      .FunctionBody(FunctionBodyHelper::BuildNodes(
+          {// nodes: {outputs, op, inputs, attributes}
+           FunctionBodyHelper::Const<float>("Q_Min", 0.0f),
+           FunctionBodyHelper::Const<float>("Q_Max", 255.f),
+           {{"X_Min"}, "ReduceMin", {"X"}},
+           {{"X_Min_Adjusted"}, "Min", {"X_Min", "Q_Min"}},
+           {{"X_Max"}, "ReduceMax", {"X"}},
+           {{"X_Max_Adjusted"}, "Max", {"X_Max", "Q_Min"}},
```
gramalingam: Is that supposed to be Q_Min or Q_Max?

onnx/defs/quantization/defs.cc
```diff
@@ -101,4 +102,49 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    11,
+    OpSchema()
+        .SetDoc(FusedQuantizeLinear_ver11_doc)
+        .Input(0, "X", "Input tensor", "T1")
+        .Output(0, "Y", "Quantized output tensor", "T2")
+        .Output(1, "Y_Scale", "Output Scale. It's a scalar or a 1D tensor with size 1.", "tensor(float)")
+        .Output(2, "Y_Zero_Point", "Output Zero point. It's a scalar or a 1D tensor of size 1.", "T2")
+        .Attr(
+          "to",
+          "The data type to which the elements of the input tensor are quantized to. Strictly must be one of the types from DataType enum in TensorProto",
```
gramalingam: I assume that this is not being used currently and that T2 is restricted to one type (until Functions are extended to simplify coding this up)? It may be worth mentioning this in the documentation.

onnx/defs/quantization/defs.cc
```diff
@@ -101,4 +102,49 @@ ONNX_OPERATOR_SET_SCHEMA(
               updateOutputShape(ctx, 0, input_shape);
             }));
 
+static const char* FusedQuantizeLinear_ver11_doc = R"DOC(
+A Function to fuse calculation for Scale, Zero Point and FP32->8Bit convertion of FP32 Input data.
+Outputs Scale, ZeroPoint and Quantized Input for a given FP32 Input)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    FusedQuantizeLinear,
```
gramalingam: Minor nit: The "Fused" part is more of an implementation-detail or concern. It would be better to give it a name that describes what the function is intended to do. But I can see naming it is a bit challenging since the name "QuantizeLinear" is already used up. Would "DynamicLinearQuantization" or some name like that work?

onnx/defs/quantization/defs.cc
askhade: It is supposed to be Q_Min... What we are doing here is making sure 0 is included in the input range to guarantee that zero can be uniquely represented in the quantized range.
Example: if input range is [-3, -1] then this would adjust max to 0 and the new range will be [-3, 0]

onnx/defs/quantization/defs.cc
askhade: Yes, I will add it in the doc

onnx/defs/quantization/defs.cc
askhade: Agree... I was struggling with the right name for this so chose to go with Fused... I will change it to DYamicLinearQuantization

askhade:
> FunctionVerification.VerifyFunctionOps test is failing because of a bug in the test itself. I will add a fix for this soon. In the meanwhile please proceed to review the code.

I will add the fix for this test as a separate PR... I dont want to mix these 2 changes in the same PR. To work around this issue I changed the Input and Output names in the schema from "X" and "Y" to "x" and "y" this resolves the test issue.

onnx/defs/quantization/defs.cc
```diff
@@ -101,4 +102,50 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    OpSchema()
+        .SetDoc(DynamicQuantizeLinear_ver11_doc)
+        .Input(0, "x", "Input tensor", "T1")
+        .Output(0, "y", "Quantized output tensor", "T2")
+        .Output(1, "y_scale", "Output Scale. It's a scalar or a 1D tensor with size 1.", "tensor(float)")
+        .Output(2, "y_zero_point", "Output Zero point. It's a scalar or a 1D tensor of size 1.", "T2")
+        .Attr(
+          "to",
+          "The data type to which the elements of the input tensor are quantized to. Strictly must be one of the types from DataType enum in TensorProto."
+          "Currently this is not used since we only allow uint8 as output data type.",
```
gramalingam: Sorry, just noticed that it is used (in the Cast op below), but it is restricted to be uint8? It would be better to say "Currently, this is required to be uint8 (value 2)."

askhade: My bad ... I also missed it...

onnx/defs/quantization/defs.cc
```diff
@@ -101,4 +102,50 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+ONNX_OPERATOR_SET_SCHEMA(
+    DynamicQuantizeLinear,
+    11,
+    OpSchema()
+        .SetDoc(DynamicQuantizeLinear_ver11_doc)
+        .Input(0, "x", "Input tensor", "T1")
+        .Output(0, "y", "Quantized output tensor", "T2")
+        .Output(1, "y_scale", "Output Scale. It's a scalar or a 1D tensor with size 1.", "tensor(float)")
+        .Output(2, "y_zero_point", "Output Zero point. It's a scalar or a 1D tensor of size 1.", "T2")
```
gramalingam: Ideally, we should say "scalar, that is a 0D tensor" in the preceding two lines. Is there any reason we need to say "or 1D tensor"? Let us try to uniformly use tensors with rank 0 for scalars.

askhade: I added 1D tensor of size 1 to relax the requirement for y_scale and y_zero_point to be "Scalars" in onnx terms... the requirement is really that they be of size 1 .i.e 1 scale and zero point per layer.

onnx/defs/quantization/defs.cc
```diff
@@ -101,4 +102,50 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    OpSchema()
+        .SetDoc(DynamicQuantizeLinear_ver11_doc)
+        .Input(0, "x", "Input tensor", "T1")
+        .Output(0, "y", "Quantized output tensor", "T2")
+        .Output(1, "y_scale", "Output scale. It's a scalar, which means a per-tensor/layer quantization.", "tensor(float)")
+        .Output(2, "y_zero_point", "Output zero point. It's a scalar, which means a per-tensor/layer quantization.", "T2")
+        .Attr(
+          "to",
+          "The data type to which the elements of the input tensor are quantized to. Strictly must be one of the types from DataType enum in TensorProto. "
+          "Currently this is required to be uint8 .i.e. value 2.",
```
linkerzhang: this attribute is actually specifying the value of "T2", so, does that mean only one is needed, otherwise, is there a way to verify to ensure the two have no conflict with each other?

gramalingam: The role of this attribute is the same as that in the cast op: it specifies what type to cast to. We need both, since both type-inference and the runtime will choose the final type based on the attribute value. (Of course, it is not needed right now since only one type is supported, but it will be needed if and when multiple output-types are supported.) But you are right about checking/verifying: we could add a shape-and-type-inference function to check the right values are specified … is there some other way to add a simple check?

gramalingam: Or, we could omit this for now, since we are not using it currently.

askhade: No we cannot omit this ... since we need a way to translate T2 to an attribute which Cast can use in the functionBody... plus can we always rely the type info to be present? 

I am in favor of adding a shape inference for this function instead

gramalingam: No, we won't translate T2 to the attribute. (Usually, the computation order is the other way: the attribute will be used to infer the value of T2.) However, in this case, you can set the attribute value of Cast to be the constant "2", since that is the only case being supported right now, right?

gramalingam: To add more context: I think that once we have the ability to specify the FunctionBody using a lambda, that function will be the right place to do the check: since we may actually generate different bodies based on the value of this attribute. So, it seems like we could postpone this issue to that extension in whatever is the most convenient way.

askhade: Eventually when we enable this for multiple data types we will need this attribute... Right now if you both think this may cause confusion then I am ok to remove this and make cast attr as a const.

However I expect when a runtime implements kernel for this function it should verify that T2 and attr values dont conflict. This check does not necessarily have to be in ONNX.

linkerzhang: My suggestion please: if there's an existing way to do the check, it's ok to keep the attribute, otherwise, we should remove it right now and think of adding it as well as a way to do verification.

askhade: I will remove it right now... let's add it when it is necessary.

askhade:
@houseroad , @ebarsoum please let me know if you have any questions or concerns
Thanks!

docs/Changelog.md
```diff
@@ -10495,6 +10495,46 @@ This version of the operator has been available since version 11 of the default
 <dd>Constrain input and output types to all tensor types.</dd>
 </dl>
 
+### <a name="DynamicQuantizeLinear-11"></a>**DynamicQuantizeLinear-11**</a>
+
+  A Function to fuse calculation for Scale, Zero Point and FP32->8Bit convertion of FP32 Input data.
+  Outputs Scale, ZeroPoint and Quantized Input for a given FP32 Input
```
ebarsoum: Can you provide the equation or pseudocode of the fused function?

onnx/backend/test/case/node/dynamicquantizelinear.py
```diff
@@ -0,0 +1,51 @@
+from __future__ import absolute_import
```
ebarsoum: Can you add a reference implementation?

PR #2194, opened 2019-07-23
specifying python 3.6 since python 3.7 does not work well with typing

linkerzhang:
CI for some spec update PR is failing with error 

" .....
   File "/Users/travis/virtualenv/lib/python3.7/site-packages/typing.py", line 1005, in __new__
      self._abc_registry = extra._abc_registry
  AttributeError: type object 'Callable' has no attribute '_abc_registry'"
  ......
"

this should be a very specific issue in python 3.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ matrix:
     - os: osx
       osx_image: xcode9.3
       env: PYTHON_VERSION=python3 ONNX_ML=0
+      python: "3.6"
     - os: linux
       sudo: required
       env: PYTHON_VERSION=python2
@@ -34,6 +35,7 @@ matrix:
     - os: osx
       osx_image: xcode9.3
       env: PYTHON_VERSION=python3
+      python: "3.6"
     - os: linux
       sudo: required
       env: PYTHON_VERSION=python2 LITE=1


linkerzhang:
looks like specifying python 3.6 in travis ymal file does not work as expected. It's still using python 3.7

.travis.yml
```diff
@@ -18,6 +18,7 @@ matrix:
     - os: osx
       osx_image: xcode9.3
       env: PYTHON_VERSION=python3 ONNX_ML=0
+      python: "3.6"
```
askhade: this works for Linux but is ignored on macOS or Windows
Check this: https://docs.travis-ci.com/user/languages/python/#running-python-tests-on-multiple-operating-systems

askhade: I think this is where you need to make the change :
https://github.com/onnx/onnx/blob/ba76e4508eecf08797c2c5b7b9ca196fabbeab0f/.travis/before_install.sh#L31

prasanthpul:
@anderspapitto can you comment? looks like you worked on the mac travis most recently

linkerzhang:
https://github.com/onnx/onnx/pull/2199 is fixing the issue :). Let's see.

linkerzhang:
issue fixed by PR #2199

PR #2200, opened 2019-07-26
More unsqueeze tests

JamesAllingham:
There are now 3 test cases for Unsqueeze:

- Unsqueeze 1 axis, which is really 3 tests for unsqueezing each axis of a 3d array
- Unsqueeze 2 axes, which tests unsqueezing both the 1st and 4th axes of a 3d array
- Unsqueeze 3 axes, which tests unsqueezing the 2nd, 4th and 5th axes of a 3d array

One weird thing is that the Operators.md and TestCoverage.md that were generated show the two and three axes test out of order, I'm not sure why that happened or if this is something to care about... it is just a little strange to read in that order.


docs/Operators.md
```diff
@@ -15137,20 +15137,67 @@ This version of the operator has been available since version 1 of the default O
 #### Examples
 
 <details>
-<summary>unsqueeze</summary>
+<summary>unsqueeze_one_axis</summary>
 
 ```python
+x = np.random.randn(3, 4, 5).astype(np.float32)
+
+for i in range(len(x) + 1):
```
```suggestion
for i in range(x.ndim):
```
maybe?
```
print(np.expand_dims(x, axis=len(x) + 1).shape)
```
produces
```
(3,4,5,1)
```
which is an undefined behavior in ONNX Unsqueeze.

docs/Operators.md
```diff
@@ -15137,20 +15137,67 @@ This version of the operator has been available since version 1 of the default O
...
 node = onnx.helper.make_node(
     'Unsqueeze',
     inputs=['x'],
     outputs=['y'],
-    axes=[0],
+    axes=[2, 4, 5],
 )
+y = np.expand_dims(x, axis=2)
+y = np.expand_dims(y, axis=4)
+y = np.expand_dims(y, axis=5)
```
wschin: I am not sure if this is the valid test. The original tensor is 3-D, so the min axis you can expand is 0 and the max one is 3. Did I miss something?

docs/Operators.md
```diff
@@ -15137,20 +15137,67 @@ This version of the operator has been available since version 1 of the default O
...
+
+```python
 x = np.random.randn(3, 4, 5).astype(np.float32)
-y = np.expand_dims(x, axis=0)
+
+node = onnx.helper.make_node(
+    'Unsqueeze',
+    inputs=['x'],
+    outputs=['y'],
+    axes=[1, 4],
```
```suggestion
    axes=[1, 3],
```

wschin:
Overall this PR looks good to me. I have one comment for changing an axis bound in your test.

docs/Operators.md
JamesAllingham: I'll make this `x.ndim + 1` if that is okay? Based on your other comment it seems like you are happy that this is defined behaviour.

docs/Operators.md
```diff
@@ -15137,20 +15137,67 @@ This version of the operator has been available since version 1 of the default O
 #### Examples
 
 <details>
-<summary>unsqueeze</summary>
+<summary>unsqueeze_one_axis</summary>
```
askhade: Looks like this file was manually updated... Please use gen_doc.py to create Operators.md and Changelog.md and stat_coverage.py to update TestCoverage.md
(make sure you set environment variable ONNX_ML=0 before building onnx and then run these scripts)

Issue #2179, opened 2019-07-18
Intended behavior for Gather indices out of bounds?


kahmed10:
In [tensorflow](https://www.tensorflow.org/api_docs/python/tf/gather), Gather throws an exception on the CPU side if any value within _indices_ is out of bounds. On the GPU side, the output values are set to 0 for that index. Is this is the same behavior in onnx?

askhade:
All index values are expected to be within bounds. It is an error if any of the index values are out of bounds. Runtimes will handle such errors in a runtime-specific fashion

PR #2202, opened 2019-07-26
Clarify ambiguity in gather spec regarding indices out of bounds

askhade:
ONNX spec for gather operator does not clearly specify expectation from indices attribute. Adding this PR to clarify this ambiguity referenced in issue : https://github.com/onnx/onnx/issues/2179


fdwr:
@askhade : I agree with the message (particularly for CPU), but I'd also point out that on some execution providers (particularly for GPU), it's already too late to throw any kind of error from deep inside the shader execution where the indices are read. So a sane fallback in that case is to either clamp the index or use index 0.

Issue #1993, opened 2019-05-03
Squeeze gives no error when axis selected shape entry is not 1.


vuzelac-cadence:
Editing `test_squeeze` in `tests/shape_inference_test.py`to remove axis 5 only (is not 1). Testing that the axis is removed passes.
```
def test_squeeze(self):  # type: () -> None
        graph = self._make_graph(
            [('x', TensorProto.FLOAT, (1, 3, 1, 1, 2, 2))],
            [make_node('Squeeze', 'x', 'y', axes=[5])],
            [])
        self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (1, 3, 1, 1, 2))])
```

Per the documentation, this should raise error ?

askhade:
Right this check has now been added to Squeeze operator's shape inference function. 
Closing this issue. @vuzelac-cadence  : please reopen it or create a new one if you have any concerns.

PR #2204, opened 2019-07-27
Add check to disallow squeezing input axes which are not 1

askhade:
Per the Squeeze op spec : "If an axis is selected with shape entry not equal to one, an error is raised." This check is not done in squeeze shape inference. Adding this PR to address this gap.

This issue is reported here: https://github.com/onnx/onnx/issues/1993

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -957,7 +957,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
           const auto& input_shape = ctx.getInputType(0)->tensor_type().shape();
 
-          for (int i = 0, j = 0; input_shape.dim_size(); ++i) {
+          for (int i = 0, j = 0; i < input_shape.dim_size(); ++i) {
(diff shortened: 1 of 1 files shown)


gramalingam:
Thanks for the fix. Just want to note, as a general comment, that (a) The runtime MUST catch all such errors, (b) Shape-inference catches such errors on a best-effort basis. In a number of cases, the required information to catch the error is not available statically, as shape-inference time. I should probably be posting this comment in the discussion in the other issues since there seems to be a confusion about the role of shape-inference and runtime in catching errors.

onnx/defs/tensor/defs.cc
```diff
@@ -955,18 +955,24 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
 
           ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+          const auto& input_shape = ctx.getInputType(0)->tensor_type().shape();
 
-          for (int i = 0, j = 0;
-               i < ctx.getInputType(0)->tensor_type().shape().dim_size();
-               ++i) {
+          for (int i = 0, j = 0; input_shape.dim_size(); ++i) {
```
gramalingam: The loop termination condition is missing the "<" comparison.

Issue #814, opened 2018-04-24
pytorch-converted/test_softmax_lastdim has test with negative axis


skottmckay:
https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/pytorch-converted/test_softmax_lastdim has a test with a negative axis as input. 

The ONNX spec does not say whether a negative axis is allowed or not. https://github.com/onnx/onnx/blob/master/docs/Operators.md#Softmax

Should the spec be clarified or the test changed?

houseroad:
Actually, this is a good topic for discussion. If we don't support negative axis, it's probably hard for us to specify the last dimension conveniently.

linkerzhang:
@houseroad , we may use symbolic dimension in this case, right? using negative value is confusing somehow given we have symbolic support in onnx.

houseroad:
@linkerzhang I didn't follow. Could you explain how to use symbolic to solve this case (expressing the last dimension in axis) ?

gramalingam:
@houseroad , so a negative axis k corresponds to an axis of "rank+k", is that right? So, "-1" corresponds to the last dimension?

houseroad:
@gramalingam Yes, it works like python negative index.

houseroad:
Let's move the discussion to https://github.com/onnx/onnx/issues/842

PR #2207, opened 2019-07-30
Clarify negative axis support in some more ops

skottmckay:
Clarify the documentation for some more ops where a negative axis value was already supported. 
Add check for invalid negative axis in a couple of places in the shape inferencing. 
Update Concat to allow a negative axis for consistency.


wschin:
What do you mean by `already` supported? If the negative axis was not explicitly mentioned, I'd feel it was not officially supported. If my feeling is valid, I'd suggest to bump their operator versions and consider `negative` axis as a new feature.

skottmckay:
> What do you mean by `already` supported? If the negative axis was not explicitly mentioned, I'd feel it was not officially supported. If my feeling is valid, I'd suggest to bump their operator versions and consider `negative` axis as a new feature.

The shape inferencing code already supported a negative axis, so this PR is largely (with the exception of Concat where support is being added) making the documentation reflect the implemented behavior. 

See https://github.com/onnx/onnx/pull/2177 for some additional discussion. 

FWIW, recently Gather and Split have been updated to add support for negative axis values without operator version changes. Personally this sort of change feels more like a bug fix for consistency.

gramalingam:
@wschin : this is a gray area (depending on whether we see this as fixing an incompleteness-bug in the specification/documentation or extending the specification), etc.. As discussed in the thread Scott referenced above, it appears there was a decision to uniformly support negative axes quite a while back. Overall, it seems keeping the same version is simpler for almost all cases. For concat, bumping the op version may be desirable if there is an implementation out there that doesn't support negative axes. I guess if someone has a concern in this regard, we could bump it for concat.

wschin:
Are we sure that all other runtimes follow the same rule?

skottmckay:
> Are we sure that all other runtimes follow the same rule?

Which runtimes are you concerned about? 

CNTK seems to have negative axis support. https://docs.microsoft.com/en-us/python/api/cntk/cntk.axis.axis?view=cntk-py-2.7
I believe pytorch does as the model that started the whole negative axis discussion was converted from pytorch: https://github.com/onnx/onnx/issues/814
TensorFlow supports it: e.g. https://www.tensorflow.org/api_docs/python/tf/math/argmax

wschin:
I mean Intel's and NVIDIA's previous run-time engines.

postrational:
We should merge this change to clear up the negative-values question, which was unclear in the ONNX spec. Only question is, whether to bump operator versions or not.

There are definitely runtimes (like Intel's nGraph), which don't support this yet. If we bump versions, then theoretically we would only allow negative values starting from the next operator version. As many runtimes already support this feature implicitly for current versions, this may be more confusing than helpful. Perhaps we don't need to bump it.

I would suggest adding a set of simple unit-tests to this PR, which check the negative `axis` values on all ops which should support them.

skottmckay:
Given there's an ONNX release coming up, we should choose between the previously accepted changes for Gather and Split where no op version increment was done, or increment the op version for all the ops where negative axis support is being explicitly added. I don't mind which, but we should decide soon. @linkerzhang what are your thoughts?

gramalingam:
I personally lean towards bumping the op version. The drawback is the "noise" and confusion this may generate (and possibly a negative perception about the stability of op specs in ONNX). If people think that these disadvantages outweigh the advantages, I am okay with not bumping it. (There was a proposal in todays operator SIG meeting to bring this up in tomorrow's ONNX workshop discussion.)

skottmckay:
Let's see if it gets discussed and resolved at the workshop tomorrow. I can add unit tests and bump op versions as needed.

wschin:
Per my personal feeling, we will increase the version number. Making something from undefined to defined sounds a breaking change.

wschin:
This PR is merged into #2260 and #2281, so please let me close it for now. Feel free to reopen if I miss something. Thank you very much!

PR #2220, opened 2019-08-06
Support ScatterND operator in ONNX

BowenBao:
This is the inverse of GatherND https://github.com/onnx/onnx/pull/2106

This will support the exporting of `torch.masked_scatter`, `torch.index_put`, `torch.copy_`, `tf.scatter_nd_update`, etc.


BowenBao:
cc @spandantiwari @ebarsoum @linkerzhang @wschin @gramalingam @postrational

gramalingam:
A question: I believe indices should not have a repeated value, is that correct? That is, we cannot have two or more updates for the same index-location. If so, let us add that to the documentation. (I guess this is common with scatter too, I presume.)

onnx/defs/tensor/defs.cc
```diff
@@ -836,6 +836,71 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         }));
 
+static const char* ScatterND_ver11_doc = R"DOC(
+ScatterND takes three inputs `data` tensor of rank r >= 1, `indices` tensor of rank q >= 1,
+and `updates` tensor of rank q + r - indices.shape[-1] - 1. The output of the operation
+is produced by creating a copy of the input `data`, and then updating its value to values
+specified by `updates` at specific index positions specified by `indices`. Its output shape
+is the same as the shape of `data`.
+
+`indices` is an integer tensor. The last dimension of `indices` can be at most the rank of
```
gramalingam: Suggest extending this to
```
`indices` is an integer tensor. Let k denote indices.shape[-1], the last dimension in the shape of `indices`. `indices' is treated as a (q-1)-dimensional tensor of k-tuples, where each k-tuple is a partial-index into `data`. Hence, k  can a value at most the rank of `data`. When k equals rank(data), each update entry specifies an update to a single element of the tensor. When k is less than rank(data) each update entry specifies an update to a slice of the tensor.
```

onnx/defs/tensor/defs.cc
```diff
@@ -836,6 +836,71 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+`data`:
+
+    indices.shape[-1] <= rank(data)
+
+`updates` is a tensor with shape
```
gramalingam: Suggest extending this to
```
`updates` is treated as a (q-1)-dimensional tensor of replacement-slice-values. Thus, the first (q-1) dimensions of updates.shape must match the first (q-1) dimensions of indices.shape. The remaining dimensions of `updates` correspond to the dimensions of the replacement-slice-values. Each replacement-slice-value is a (r-k) dimensional tensor, corresponding to the trailing (r-k) dimensions of `data`.  Thus, the shape of `updates` must equal indices.shape[0:q-1] ++ data.shape[k:r-1], where ++ denotes the concatenation of shapes.

File: onnx/defs/tensor/defs.cc [hunk 3]
```diff
@@ -836,6 +836,71 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+    indices.shape[:-1] + data.shape[indices.shape[-1]:]
+
+The `output` is calculated via the following equation:
+
+    output = data
+    output[indices[(i_0, ..., i_{q-2})]] = updates[(i_0, ..., i_{q-2})]
```
gramalingam: Can you add an iteration loop around the update equation (the second line)? I think we could use numpy's ndindex to express the iteration. Something like below
```python
   update_indices = indices.shape[:-1]
   for idx in np.ndindex(update_indices):
      output[indices[idx]] = updates[idx]
```
You may have to adjust a few things above to make it run, but as pseudo-code, hopefully it will be clear.

BowenBao:
@gramalingam Thanks for your detailed suggestions! I have updated the doc with clarification. You are right duplicate entries of indices is not supported.

docs/Operators.md
```diff
@@ -13975,6 +13976,124 @@ expect(node, inputs=[data, indices, updates], outputs=[y],
...
+  must equal indices.shape[0:q-1] ++ data.shape[k:r-1], where ++ denotes the concatenation
+  of shapes.
+  
+  The `output` is calculated via the following equation:
+  
+      output = np.copy(data)
+      update_indices = indices.shape[:-1]
+      for idx in np.ndindex(update_indices):
+          output[indices[idx]] = updates[idx]
+
```
gramalingam: Suggest adding: "The order of iteration in the above loop is not specified. In particular, indices should not have duplicate entries: that is, if idx1 != idx2, then indices[idx1] != indices[idx2]. This ensures that the output value does not depend on the iteration order."

BowenBao: Thanks! added as well to the sample implementation in backend test scatternd.py.

spandantiwari:
@gramalingam - thanks for the review!

spandantiwari:
@bddppq - for his review.

BowenBao:
@prasanthpul could you include this in ONNX 1.6 milestone as well? Thanks.

docs/Operators.md
```diff
@@ -13975,6 +13976,128 @@ expect(node, inputs=[data, indices, updates], outputs=[y],
...
+  ScatterND takes three inputs `data` tensor of rank r >= 1, `indices` tensor of rank q >= 1,
+  and `updates` tensor of rank q + r - indices.shape[-1] - 1. The output of the operation
+  is produced by creating a copy of the input `data`, and then updating its value to values
+  specified by `updates` at specific index positions specified by `indices`. Its output shape
+  is the same as the shape of `data`. Note that `indices` should not have duplicate entries.
+  That is, two or more `updates` for the same index-location is not supported.
+  
+  `indices` is an integer tensor. Let k denote indices.shape[-1], the last dimension in the shape of `indices`.
+   `indices' is treated as a (q-1)-dimensional tensor of k-tuples, where each k-tuple is a partial-index into `data`.
```
fdwr: `indices' <- closing quote should be backtick

BowenBao: Thanks!

PR #2233, opened 2019-08-09
Add Det to ONNX

BowenBao:
Det calculates determinant of a square matrix or batches of square matrices.
  Det takes one input tensor of shape `[*, M, M]`, where `*` is zero or more batch dimensions,
  and the inner-most 2 dimensions form square matrices.
  The output is a tensor of shape `[*]`, containing the determinants of all input submatrices.

PyTorch: https://pytorch.org/docs/master/torch.html?highlight=det#torch.det
TF: https://www.tensorflow.org/api_docs/python/tf/linalg/det


ajaysg:
i added the following functions in symbolic_opset9.py for `det` and `logdet`
`def logdet(g, self):
    return g.op("ATen", self)


@parse_args('v')
def det(g, self):
    return g.op("Aten", self)`

then onnx is able to export

docs/Changelog.md
```diff
@@ -10495,6 +10495,38 @@ This version of the operator has been available since version 11 of the default
...
+<dl>
+<dt><tt>Y</tt> : T</dt>
+<dd>Output tensor</dd>
+</dl>
+
+#### Type Constraints
+
+<dl>
+<dt><tt>T</tt> : tensor(float16), tensor(float), tensor(double)</dt>
+<dd>Constrain input and output types to float tensors.</dd>
```
spandantiwari: nit: consider saying "floating-point tensors" for clarity.

onnx/defs/math/defs.cc
```diff
@@ -1586,4 +1586,45 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          // Shape inference
+          if (hasInputShape(ctx, 0)) {
+            const TensorShapeProto& input_shape =
+                ctx.getInputType(0)->tensor_type().shape();
+            TensorShapeProto* output_shape =
+                ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+            const int rank = static_cast<int>(input_shape.dim_size());
+
+            if (rank < 2) {
+              fail_shape_inference("Input rank must be >= 2.")
```
spandantiwari: Are we specifying behavior when rank >3? If not, would it make sense to check for that here?

docs/Operators.md
```diff
@@ -3637,6 +3638,79 @@ expect(node, inputs=[x, x_scale, x_zero_point], outputs=[y],
...
+
+<dl>
+<dt><tt>Y</tt> : T</dt>
+<dd>Output tensor</dd>
+</dl>
+
+#### Type Constraints
+
+<dl>
+<dt><tt>T</tt> : tensor(float16), tensor(float), tensor(double)</dt>
```
spandantiwari: Do we want to support complex types (like TF) at this point?

onnx/defs/math/defs.cc
BowenBao: The shape inferencing covers for all rank >= 2 cases, and errors out when rank < 2. I do forgot to add checks for sizes of the inner-most 2 dimensions. Thanks.

docs/Operators.md
BowenBao: Currently we are not seeing model use cases for complex types. We can extend dtype support in future opsets when necessary.

onnx/defs/math/defs.cc
```diff
@@ -1586,4 +1586,53 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            }
+
+            const auto mat_w = input_shape.dim(rank - 1);
+            const auto mat_h = input_shape.dim(rank - 2);
+            if (mat_w.has_dim_value() &&
+                mat_h.has_dim_value() &&
+                (mat_w.dim_value() != mat_h.dim_value())) {
+              fail_shape_inference("The inner-most 2 dimensions must have the same size.");
```
wschin: Please print out the values of `mat_w` and `mat_h`.

onnx/defs/math/defs.cc
```diff
@@ -1586,4 +1586,53 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            }
+
+            for (int i=0; i < rank - 2; ++i) {
```
wschin: Very nice design which can correctly produce scalar shapes.

onnx/defs/math/defs.cc
```diff
@@ -1586,4 +1586,53 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain input and output types to float tensors.")
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
+static const char* Det_ver11_doc = R"DOC(
+Det calculates determinant of a square matrix or batches of square matrices.
+Det takes one input tensor of shape `[*, M, M]`, where `*` is zero or more batch dimensions,
+and the inner-most 2 dimensions form square matrices.
+The output is a tensor of shape `[*]`, containing the determinants of all input submatrices.
```
wschin: Please add the output shape (i.e., scalar) when the input is 2-D.

onnx/defs/math/defs.cc
BowenBao: Added, thanks!

spandantiwari:
cc: @postrational for review.

spandantiwari:
cc: @bddppq

BowenBao:
@prasanthpul could you include this in ONNX 1.6 milestone as well? Thanks.

docs/Operators.md
```diff
@@ -3637,6 +3638,80 @@ expect(node, inputs=[x, x_scale, x_zero_point], outputs=[y],
...
+
+```python
+node = onnx.helper.make_node(
+    'Det',
+    inputs=['x'],
+    outputs=['y'],
+)
+
+x = np.arange(4).reshape(2, 2).astype(np.float32)
+y = det_impl(x)  # expect -2
```
postrational: `det_impl` should be replaced with `np.linalg.det` in these examples to explain what it does.

BowenBao: done, thanks!

PR #2235, opened 2019-08-13
Support undefined type in TypesWrapper

hariharans29:
Fill in a gap in TypesWrapper (undefined is a valid type)


hariharans29:
CC: @gramalingam @ebarsoum

gramalingam:
I am not sure about this. "undefined" is not a valid type (in most contexts). If there is some context where we need to use the "undefined" type, it would be better to clarify the situations where this is permitted and used: the use-cases/scenarios where we need this.

hariharans29:
@gramalingam - Well undefined is still valid within the spec of ONNX isn't it ?

skottmckay:
Is there something that says TypesWrapper should only have values that are always valid in a model? The change is to add a value that is defined in the proto which matches the comment:
  // DataType strings. These should match the DataTypes defined in onnx.proto	

tensor<undefined> seems valid as a placeholder that's treated as 'must be replaced by type inferencing prior to running'. Only real use-case I'm aware of is for a subgraph input, as that needs some type info like a rank, but not necessarily the element type as that can be inferred from the parent graph.

hariharans29:
> Is there something that says TypesWrapper should only have values that are always valid in a model? The change is to add a value that is defined in the proto which matches the comment:
> // DataType strings. These should match the DataTypes defined in onnx.proto
> 
> tensor seems valid as a placeholder that's treated as 'must be replaced by type inferencing prior to running'. Only real use-case I'm aware of is for a subgraph input, as that needs some type info like a rank, but not necessarily the element type as that can be inferred from the parent graph.

I agree. The intention of this map only seems to be holding the data type strings for the data types in onnx.proto. If that is indeed the actual intention, this map's contents shouldn't define which types are valid in a model and which are not.

gramalingam:
Doesn't this generalization now allow users to define a op that takes as input a "tensor(undefined)"? If we are introducing this generalization, it seems worth clarifying where it is legal and where it is not. We can start conservatively, allowing it in special cases where it seems necessary and useful.

hariharans29:
> Doesn't this generalization now allow users to define a op that takes as input a "tensor(undefined)"? If we are introducing this generalization, it seems worth clarifying where it is legal and where it is not. We can start conservatively, allowing it in special cases where it seems necessary and useful.

Thanks Rama. That is a good suggestion. I have a question though - I went over the IR.md file and couldn't find a mention of "undefined" type being prohibited anywhere for graph inputs. Is is=t implicitly stated somewhere ? (Sorry if I missed it)

PR #2244, opened 2019-08-16
move map and sequence types to onnx domain,

linkerzhang:
this is the first step of merging onnx-ml and onnx types.


linkerzhang:
Merging ONNX and ONNX-ML IR was agreed in infra SIG meeting on 7/1/2019 as https://github.com/onnx/sigs/blob/master/infra/meetings/001-20190701.md. Especially for types, the difference between DNN and traditional ML can/will be shown by different op domains. there's no need having two different IR formats for the differentiation.

linkerzhang:
@bddppq IR version will be bumped before the 1.6 release, since there will be more changes to IR before 1.6 - merging onnx-ml types into onnx. @gramalingam is also trying to make changes to SparseTensor type.

Of course, I can bump the IR version now if you want.

onnx/onnx.proto
```diff
@@ -510,19 +510,49 @@ message TypeProto {
     optional TensorShapeProto shape = 2;
   }
 
+  // repeated T
+  message Sequence {
+    // The type and optional shape of each element of the sequence.
+    // This field MUST be present for this version of the IR.
+    optional TypeProto elem_type = 1;
```
spandantiwari: I am wondering if we still need this comment here? If this can be optional indeed, we can use Sequence to export heterogeneous tensors of different datatypes also by keeping `elem_type` unspecified.

onnx/onnx.in.proto
```diff
@@ -564,11 +562,13 @@ message TypeProto {
 
     // The type of a map.
     Map map_type = 5;
+    
+    SparseTensor sparse_tensor_type = 8;
+
+// #if ONNX-ML
```
wschin: Can you remove this line?

onnx/onnx.in.proto
```diff
@@ -535,6 +525,16 @@ message TypeProto {
...
+  
+  message SparseTensor { 
+    // This field MUST NOT have the value of UNDEFINED 
+    // This field MUST have a valid TensorProto.DataType value
+    // This field MUST be present for this version of the IR. 
+    optional int32 elem_type = 1;
+    optional TensorShapeProto shape = 2;
+  }
+
+// #if ONNX-ML
```
wschin: This line looks redundant.

onnx/onnx.proto
wschin: Supporting heterogeneous tensors can probably by adding another type `Tuple`. Sequence stills sounds like a homogeneous type like `IEnumerable<ElementType>` in C#.

onnx/onnx.proto
```diff
@@ -510,19 +510,49 @@ message TypeProto {
...
-  
+
 
   oneof value {
     // The type of a tensor.
     Tensor tensor_type = 1;
 
+    // NOTE:  DNN-only implementations of ONNX MAY elect to not support non-tensor values
+    //        as input and output to graphs and nodes. These types are needed to naturally
+    //        support classical ML operators.  DNN operators SHOULD restrict their input
```
wschin:
```suggestion
    //        support classical ML operators. DNN operators SHOULD restrict their input
```

onnx/onnx.proto
```diff
@@ -510,19 +510,49 @@ message TypeProto {
...
+    //        and output types to tensors.
+
+    // The type of a sequence.
+    Sequence sequence_type = 4;
+
+    // The type of a map.
+    Map map_type = 5;
+
```
wschin:
```suggestion

    // The type of sparse tensor. This type is different than tensor.
```

onnx/onnx.proto3
```diff
@@ -510,19 +510,49 @@ message TypeProto {
...
-  
+
 
   oneof value {
     // The type of a tensor.
     Tensor tensor_type = 1;
 
+    // NOTE:  DNN-only implementations of ONNX MAY elect to not support non-tensor values
+    //        as input and output to graphs and nodes. These types are needed to naturally
+    //        support classical ML operators.  DNN operators SHOULD restrict their input
```
wschin:
```suggestion
    //        support classical ML operators. DNN operators SHOULD restrict their input
```

onnx/onnx.proto
gramalingam: I believe Spandan is talking about an unbounded list/sequence, but where the elements don't have the same type. I think there are some details concerning such a type worth sorting out first. E.g., do we just want a list of tensors of different types, or do we want truly heterogeneous lists (a list where some elements are tensors, other elements are lists of tensors, other elements are maps, etc.)? I guess it should be the later to be truly general-purpose. More importantly, we would need support for some form of dynamic-cast from untyped values to make use of such a type. I think it would be helpful to have a more comprehensive proposal (e.g., take a simple example with heterogeneous tensors and see what's required to support it).

gramalingam: In short, I think we can do that generalization separately, so that this PR can be kept simple.

spandantiwari: Yes, I think hetergeneous lists is a use case that we need to target. But I agree that it can designed separately to keep this PR simple.

onnx/onnx.in.proto
linkerzhang: no, this can't be removed.

onnx/onnx.in.proto
linkerzhang: no, it's not. see line 520 before.

onnx/onnx-ml.proto
```diff
@@ -591,8 +591,4 @@ message OperatorSetIdProto {
   // The version of the operator set being identified.
   // This field MUST be present in this version of the IR.
   optional int64 version = 2;
-}
-
-// For using protobuf-lite
-option optimize_for = LITE_RUNTIME;
```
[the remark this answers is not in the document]
linkerzhang: these files were auto-generated.... :) let me double check.

linkerzhang:
@prasanthpul given this PR is not merging ONNX-ML with ONNX, I'm not updating the docs to remove ONNX-ML concept.

onnx/onnx.proto
wschin: If we need heterogeneous types in one single variable, I'd personally consider a new type `Tuple` instead of `Sequence`. `Sequence` is more like a collection type where elements are not stored in consecutive memory blocks.

Issue #2198, opened 2019-07-25
ONNX optimzer crashes for initializers not listed as graph-input


thilow:
Initializers are not required being listed in graph inputs (To my understanding this requirement was lifted in context of issue #1449).

**Problem**: The ONNX optimizer fails for models that have initializers which are not listed in graph inputs. 

Below there is a sample script demonstrating the issue. For the first version of the model (model_1) the optimizer passes while for the second it fails.
See further below for script output 

```python
#%%
import onnx
from onnx import helper, optimizer
from onnx import AttributeProto, TensorProto, GraphProto

# Creating a simple model that adds 42 to the input-value.
# The initializer is not added to he graph-inputs
def create_simpel_model():
    # Create one input (ValueInfoProto)
    a = helper.make_tensor_value_info('a', TensorProto.FLOAT, [1])

    # Create an initialzer with value 42
    b = helper.make_tensor("b",TensorProto.FLOAT,[1], [42])

    # Create one output (ValueInfoProto)
    Y = helper.make_tensor_value_info('Y', TensorProto.FLOAT, [1])

    # Create a node (NodeProto)
    node_def = helper.make_node(
        'Add', # node name
        ['a', 'b'], # inputs
        ['Y'], # outputs
    )

    # Create the graph (GraphProto)
    graph_def = helper.make_graph(
        [node_def],
        'test-model',
        [a],
        [Y],
        [b]
    )

    # Create the model (ModelProto)
    return helper.make_model(graph_def, producer_name='onnx-example')


print("ONNX version: ", onnx.__version__)
#%%
model_1 = create_simpel_model()

# Postprocessing the model by adding the initializer 'b' to the input.
# This prevents the optimizer from crashing.
new_input = helper.make_tensor_value_info('b', TensorProto.FLOAT,[1])
model_1.graph.input.append(new_input)

#%% check model_1
onnx.checker.check_model(model_1)
print('model_1 is checked and ok!')

#%%
# Optimize the model
passes = [
    "eliminate_deadend",
    "eliminate_identity",
    "eliminate_nop_dropout",
    "eliminate_nop_monotone_argmax",
    "eliminate_nop_pad",
    "eliminate_nop_transpose",
    "extract_constant_to_initializer",
    "fuse_consecutive_concats",
    "fuse_consecutive_log_softmax",
    "fuse_consecutive_reduce_unsqueeze",
    "fuse_consecutive_squeezes",
    "fuse_consecutive_transposes"]

[...]

PR #2247, opened 2019-08-20
Fix: ONNX optimzer crashes for initializers not listed as graph-input #2198

wutiantong:
I described my original intention at #2246

### DIFF
--- a/onnx/common/ir_pb_converter.cc
+++ b/onnx/common/ir_pb_converter.cc
@@ -198,8 +198,8 @@ std::unique_ptr<Graph> graphProtoToGraph(
   // outputs them. Therefore we initialize the Nodes and Values in
   // several stages.
   //
-  // 0) add all initializers as input Values
   // 1) add all input (to the graph) Values, owned by the sentinel Param node
+  // 1) add all initializers (maybe) as input Values
   // 2) add all Nodes and their output Values, but don't intialize inputs
   // 3) initialize inputs of all Nodes
   // 4) initialize inputs of the Return sentinel node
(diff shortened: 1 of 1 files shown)


onnx/common/ir_pb_converter.cc
```diff
@@ -236,6 +237,16 @@ std::unique_ptr<Graph> graphProtoToGraph(
     value_by_name_of[vip.name()] = v;
   }
 
+  for (int i = 0; i < gp.initializer_size(); i++) {
+    auto init = tensorProtoToTensor(gp.initializer(i));
+    if (value_by_name_of.count(init.name()))
```
wschin:
```suggestion
    if (value_by_name_of.count(init.name()) == 0)
```

wutiantong:
This PR fix #2198

wschin:
Thanks a lot for this fix. Could you please add a test to avoid the problem?

wutiantong:
> Thanks a lot for this fix. Could you please add a test to avoid the problem?
> 
> [Update] An initializer is not necessarily an input. Why do you add that initializer as an input?

It seems to be unavoidable for current ir.h : 
`struct Node` need to invoke `addInput()` with a `Value *` for each initializer,
and `Value *` must be held by some node's `outputs_` ( by invoking `addOutput()` ),
however, graph's `input_` node is the only choice.

I suppose, when exporting graph, trace back inputs made with this way and drop them.
Or maybe we can extend `struct Node` with a new member to hold initializers' `Value *`, obviously, some more adjustments will be involved.

spandantiwari:
> It seems to be unavoidable for current ir.h :
> `struct Node` need to invoke `addInput()` with a `Value *` for each initializer,
> and `Value *` must be held by some node's `outputs_` ( by invoking `addOutput()` ),
> however, graph's `input_` node is the only choice.
> 

Yes, this is a larger issue. That is, the graph IR here is mostly following ONNX IR v3, where it assumes each initializer MUST be an input. That is no longer true in IR v4 and therefore the graph IR might not be able to handle such situations. 
The solution is to fix this graph IR grounds up. It may take a larger fix as you are alluding to.

Zehaos:
Any update?

PR #2249, opened 2019-08-22
Sequence related ops

BowenBao:
Following https://github.com/onnx/onnx/pull/2244. This PR adds the first few ops that supports basic sequence functionality in ONNX.

- [x] Complete sequence ops spec.
- [x] Update related ops with variadic inputs/outputs (concat, split, etc) spec.
- [x] Update shape inference for sequences.
- [x] Update onnx.helper for sequences.
- [x] Add test cases for new ops.


onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,235 @@
...
+        .TypeConstraint(
+            "T",
+            OpSchema::all_tensor_types(),
+            "Constrain input types to any tensor type.")
+        .TypeConstraint(
+            "S",
+            OpSchema::all_tensor_sequence_types(),
+            "Constrain output types to any tensor type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
```
gramalingam: I don't think propagateElemTypeFromInputToOutput will work … that assumes that the output type is a Tensor, which is not the case here. We need another utility function for sequence here.

onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,235 @@
...
+            1,
+            "tensor",
+            "Input tensor to be inserted into the input sequence.",
+            "T")
+        .Input(
+            2,
+            "position",
+            "Position in the sequence where the new tensor is inserted. "
+            "It is optional and default is to insert to the back of the sequence. "
+            "It must be a scalar(tensor of empty shape).",
```
gramalingam: Are indices zero-based? Is the position required to be <= current length?

onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,235 @@
...
+        .TypeConstraint(
+            "S",
+            OpSchema::all_tensor_sequence_types(),
+            "Constrain to any tensor type.")
+        .TypeConstraint(
+            "I",
+            {"tensor(int32)", "tensor(int64)"},
+            "Constrain position to integral tensor. It must be a scalar(tensor of empty shape).")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
```
gramalingam: As above. We need a different utility function for sequences, which are different from tensors.

gramalingam:
Re. updating ops like concat/split: I feel it may be better to introduce new ops (e.g., ConcatSequence) allowing both the older form and newer form. Once implementations stabilize (e.g, if it becomes clear that there is no cost/overhead between the two versions), if necessary, we can deprecate the older form.

onnx/defs/sequence/defs.cc
BowenBao: Updated definition for position.

spandantiwari:
> Re. updating ops like concat/split: I feel it may be better to introduce new ops (e.g., ConcatSequence) allowing both the older form and newer form. Once implementations stabilize (e.g, if it becomes clear that there is no cost/overhead between the two versions), if necessary, we can deprecate the older form.

I agree with this.

spandantiwari:
This is relevant to this PR and maybe even https://github.com/onnx/onnx/pull/2244 which brings in Sequence and Map types in ONNX. Do we want to rethink the name `Sequence` for the datatype and these new ops, given that it is also commonly used for data "sequences" for RNN models? This may cause confusion, if only initially, between the two because the `Sequence` type used here is not really what will be used for sequences for RNN models.

BowenBao:
> This is relevant to this PR and maybe even #2244 which brings in Sequence and Map types in ONNX. Do we want to rethink the name `Sequence` for the datatype and these new ops, given that it is also commonly used for data "sequences" for RNN models? This may cause confusion, if only initially, between the two because the `Sequence` type used here is not really what will be used for sequences for RNN models.

I agree. There is also an existing operator named `ReverseSequence`, which is unrelated to the `Sequence` type.

docs/Operators.md
```diff
@@ -2349,6 +2357,45 @@ for test_case, values_ in test_cases.items():
 </details>
 
 
+### <a name="ConcatFromSequence"></a><a name="concatfromsequence">**ConcatFromSequence**</a>
+
+  Concatenate a sequence of tensors into a single tensor.
```
shinh: IIUC, this corresponds to `np.concatenate`. I wonder if it makes sense to support `np.stack` as well. `np.stack` could be implemented by `Loop`-ing for the input sequence, `Unsqueeze` in each tensor in the sequence, and `ConcatFromSequence`, but I think this is a bit tedious. I think one solution is adding an attribute like `new_axis=false`. If this is true, this operator behaves like `np.stack`

docs/Operators.md
```diff
@@ -15291,6 +15552,57 @@ expect(node, inputs=[input], outputs=[y for y in expected_outputs], name='test_s
 </details>
 
 
+### <a name="SplitToSequence"></a><a name="splittosequence">**SplitToSequence**</a>
+
+  Split a tensor into a sequence of tensors, along the specified
```
shinh: Similar to my comment on concat, I think it'd be verbose to realize `list(a_tensor)` with this op. IIUC, we need to loop over the output sequence and apply Squeeze.

docs/Changelog.md
```diff
@@ -11208,6 +11246,264 @@ This version of the operator has been available since version 11 of the default
...
+#### Version
+
+This version of the operator has been available since version 11 of the default ONNX operator set.
+
+#### Inputs
+
+<dl>
+<dt><tt>input_sequence</tt> : S</dt>
+<dd>Input sequence.</dd>
+<dt><tt>position</tt> : I</dt>
```
shinh: I think we'll want to take `start`, `end`, and `step` like `Slice` op?

docs/Operators.md
BowenBao: Thanks! That makes sense, adding `new_axis` as attribute default to `0` to support both `concat` and `stack`.

docs/Operators.md
BowenBao: Added attribute `keepdims` default to `1`. Also making input `split` optional, so that if not provided the split size is 1. `keepdims` will be ignored if `split` is provided, since the only case we need `keepdims` is when split size is 1.

docs/Changelog.md
BowenBao: That will probably be a separate operator(or function). This operator returns `Tensor`, while the `Slice` equivalent returns `Sequence`.

BowenBao:
> Let me leave one more comment

Thanks for your comments! These are very helpful suggestions.

docs/Changelog.md
shinh: Ah yeah, that makes sense!

onnx/defs/shape_inference.h
```diff
@@ -252,6 +252,62 @@ inline void propagateElemTypeFromInputToSequenceOutput(
...
+        ->mutable_tensor_type()
+        ->set_elem_type(input_elem_type);
+  } else {
+    // This is not expected to happen
+    fail_type_inference(
+        "Output ", outputIndex, " expected to have tensor type");
+  }
+}
+
+inline void propagateElemTypeFromSequenceInputToSequenceOutput(
```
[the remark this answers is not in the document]
BowenBao: makes sense, updated.

onnx/shape_inference/implementation.cc
```diff
@@ -58,6 +58,29 @@ void checkShapesAndTypes(
...
+}
+
+void checkShapesAndTypes(
+    const TypeProto& inferredType,
+    const TypeProto& existingType) {
+  if (inferredType.has_tensor_type() && existingType.has_tensor_type()) {
+    checkShapesAndTypes(inferredType.tensor_type(), existingType.tensor_type());
+  } else if (inferredType.has_sequence_type() && existingType.has_sequence_type()) {
+    checkShapesAndTypes(inferredType.sequence_type(), existingType.sequence_type());
+  }
```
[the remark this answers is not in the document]
BowenBao: checks added

BowenBao:
@prasanthpul - can we include this in 1.6?

@linkerzhang @houseroad @postrational - for review

onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,630 @@
...
+          auto *output_tensor_type =
+              ctx.getOutputType(0)
+                  ->mutable_sequence_type()
+                  ->mutable_elem_type()
+                  ->mutable_tensor_type();
+
+          *(output_tensor_type->mutable_shape()) = 
+              ctx.getInputType(0)->sequence_type().elem_type().tensor_type().shape();
+        
+          if (!hasInputShape(ctx, 1)) {
```
gramalingam: I think this check has to be moved up before copying the shape from input 0.

BowenBao:
cc @wschin for review as well.

docs/Changelog.md
```diff
@@ -10370,6 +10370,49 @@ This version of the operator has been available since version 11 of the default
 <dd>Constrain input and output types to float tensors.</dd>
 </dl>
 
+### <a name="ConcatFromSequence-11"></a>**ConcatFromSequence-11**</a>
+
+  Concatenate a sequence of tensors into a single tensor.
```
spandantiwari: Nit: Theoretically it is possible to have a `Sequence` of other types, such as `Map` or `Sequence`. It may be helpful to add a simple clarification that this op is intended for `Sequence` of `Tensor` only (it is implied from the phrasing, but I suggest being explicit).

docs/Changelog.md
```diff
@@ -11424,6 +11467,266 @@ This version of the operator has been available since version 11 of the default
...
+<dt><tt>I</tt> : tensor(int64)</dt>
+<dd>Constrain output to integral tensor. It must be a scalar(tensor of empty shape).</dd>
+</dl>
+
+### <a name="SplitToSequence-11"></a>**SplitToSequence-11**</a>
+
+  Split a tensor into a sequence of tensors, along the specified
+  'axis'. Lengths of the parts can be specified using argument 'split'.
+  'split' must contain only positive numbers.
+  'split' is either a scalar(tensor of empty shape), or a 1-D tensor.
```
spandantiwari: nit: space needed after 'scalar', i.e. 'scalar ('

docs/Changelog.md
```diff
@@ -11424,6 +11467,266 @@ This version of the operator has been available since version 11 of the default
...
+
+This version of the operator has been available since version 11 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>axis</tt> : int (default is 0)</dt>
+<dd>Which axis to split on. A negative value means counting dimensions from the back. Accepted range is [-rank, rank-1].</dd>
+<dt><tt>keepdims</tt> : int (default is 1)</dt>
+<dd>Keep the split dimension or not, default 1 mean keep split dimension. If input 'split' is specified this attribute is ignored.</dd>
```
spandantiwari: nit: phrasing. 'Keep the split dimension or not. Default 1, which means we keep the split dimension. If input 'split' is specified, this attribute is ignored.'

docs/Changelog.md
BowenBao: thanks, fixed.

docs/Changelog.md
BowenBao: Yes we need to be explicit on types. The input type constraint shows `sequence<tensor<elem_type>>`, that should be enough ..

docs/Operators.md
```diff
@@ -14522,6 +14574,220 @@ expect(node, inputs=[x], outputs=[y],
...
+#### Version
+
+This version of the operator has been available since version 11 of the default ONNX operator set.
+
+#### Inputs
+
+<dl>
+<dt><tt>input_sequence</tt> : S</dt>
+<dd>Input sequence.</dd>
+<dt><tt>position</tt> : I</dt>
```
shinh: How about making this an optional input for consistency with `SequenceInsert`? I think it should behave like `SequenceErase(seq, -1)`.

onnx/backend/test/case/model/sequence.py
```diff
@@ -0,0 +1,226 @@
...
+        z = np.ones(z_shape, dtype=np.float32) * 2
+
+        pos = onnx.helper.make_tensor('pos', TensorProto.INT64, (), (1, ))
+        pos_at = onnx.helper.make_tensor('pos_at', TensorProto.INT64, (), (2, ))
+
+        graph = make_graph(
+            [seq_empty_node, seq_insert_node, seq_insert_node2, seq_insert_node3, seq_at_node],
+            [x_shape, y_shape, z_shape, [], []],  # type: ignore
+            [out_shape],  # type: ignore
+            ['X', 'Y', 'Z', 'pos', 'pos_at'],
```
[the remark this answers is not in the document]
BowenBao: Good catch, this is indeed a bug. Thanks @shinh

docs/Operators.md
BowenBao: makes sense, updated.

onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,655 @@
...
+
+#include "onnx/defs/function.h"
+#include "onnx/defs/schema.h"
+
+#include <numeric>
+#include <algorithm>
+
+namespace ONNX_NAMESPACE {
+
+void UnionShapeInfoToSequence(
```
gramalingam: Thanks for renaming it. I still think that this function has nothing specific to "Sequence". It deals only with Tensor type and its shape. So, I think that the "ToSequence" part of the name can be dropped.

onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,655 @@
...
+            OpSchema::all_tensor_types(),
+            "Constrain input types to any tensor type.")
+        .TypeConstraint(
+            "S",
+            OpSchema::all_tensor_sequence_types(),
+            "Constrain output types to any tensor type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          size_t numInputs = ctx.getNumInputs();
+          if (numInputs < 1) {
+            return;
```
gramalingam: If we have zero inputs, we will have a problem, since we cannot infer the output type (which is why "SequenceEmpty" needs a "dtype" attribute). The default interpretation for a variadic input is 1-or-more inputs. So, this case should not happen for a valid model. May be a comment that this cannot happen or fail_type_inference may be helpful.

BowenBao:
Thanks for your review @gramalingam !

docs/Changelog.md
```diff
@@ -10531,6 +10531,49 @@ This version of the operator has been available since version 11 of the default
 <dd>Constrain output types to any tensor type.</dd>
 </dl>
 
+### <a name="ConcatFromSequence-11"></a>**ConcatFromSequence-11**</a>
+
+  Concatenate a sequence of tensors into a single tensor.
+  All input tensors must have the same rank.
```
wschin: Can I stack [3, 2]-tensor and [5, 4]-tensor? Do you have any constraints in sequence elements' shapes?

docs/Changelog.md
```diff
@@ -10531,6 +10531,49 @@ This version of the operator has been available since version 11 of the default
...
+
+<dl>
+<dt><tt>concat_result</tt> : T</dt>
+<dd>Concatenated tensor</dd>
+</dl>
+
+#### Type Constraints
+
+<dl>
+<dt><tt>S</tt> : seq(tensor(uint8)), seq(tensor(uint16)), seq(tensor(uint32)), seq(tensor(uint64)), seq(tensor(int8)), seq(tensor(int16)), seq(tensor(int32)), seq(tensor(int64)), seq(tensor(float16)), seq(tensor(float)), seq(tensor(double)), seq(tensor(string)), seq(tensor(bool)), seq(tensor(complex64)), seq(tensor(complex128))</dt>
```
wschin: [nit] I don't think we need this many tensors, especially complex64 and complex 128.

docs/Changelog.md
```diff
@@ -12479,6 +12522,215 @@ This version of the operator has been available since version 11 of the default
...
+
+This version of the operator has been available since version 11 of the default ONNX operator set.
+
+#### Inputs
+
+<dl>
+<dt><tt>input_sequence</tt> : S</dt>
+<dd>Input sequence.</dd>
+<dt><tt>position</tt> : I</dt>
+<dd>Position of the tensor in the array. Negative value means counting positions from the back. Accepted range in `[-n, n - 1]`, where `n` is the number of tensors in 'input_sequence'. It is an error if any of the index values are out of bounds. It must be a scalar(tensor of empty shape).</dd>
```
```suggestion
<dd>Position of the tensor in the sequence. Negative value means counting positions from the back. Accepted range in `[-n, n - 1]`, where `n` is the number of tensors in 'input_sequence'. It is an error if any of the index values are out of bounds. It must be a scalar(tensor of empty shape).</dd>
```

docs/Changelog.md
```diff
@@ -12641,6 +12893,59 @@ This version of the operator has been available since version 11 of the default
...
+### <a name="SplitToSequence-11"></a>**SplitToSequence-11**</a>
+
+  Split a tensor into a sequence of tensors, along the specified
+  'axis'. Lengths of the parts can be specified using argument 'split'.
+  'split' must contain only positive numbers.
+  'split' is either a scalar (tensor of empty shape), or a 1-D tensor.
+  If 'split' is a scalar, then 'input' will be split into equally sized chunks(if possible).
+  Last chunk will be smaller if the 'input' size along the given axis 'axis' is not divisible
+  by 'split'.
+  Otherwise, the tensor is split into 'size(split)' chunks, with lengths of the parts on 'axis'
```
wschin: Very nice description.

onnx/backend/test/case/model/sequence.py
```diff
@@ -0,0 +1,240 @@
...
+            [onnx.TensorProto.FLOAT] * 3 + [onnx.TensorProto.INT64] * 2,  # type: ignore
+            [onnx.TensorProto.FLOAT],
+            [pos, pos_at])
+        model = onnx.helper.make_model(graph, producer_name='backend-test')
+        expect(model, inputs=[x, y, z], outputs=[y], name="test_sequence_model1")
+
+        #2nd testcase - erase and at.
+        # 1. SequenceConstruct(x, y, z):    -> [x, y, z]
+        # 2. SequenceErase(1):              -> [x, z]
+        # 4. SequenceAt(1):                 -> z
```
wschin:
```suggestion
        # 3. SequenceAt(1):                 -> z
```

onnx/backend/test/case/model/sequence.py
```diff
@@ -0,0 +1,240 @@
+from __future__ import absolute_import
```
wschin: Future task: we need for single operator. There are serveral tests at grahp level but we need single-node test for fine-level tests. The type of sequence can be Python list, `[x, y, z]`, where `x`, `y`, `z` are numpy tensors.

onnx/helper.py
```diff
@@ -358,6 +358,30 @@ def make_tensor_value_info(
     return value_info_proto
 
 
+def make_sequence_value_info(
+        name,  # type: Text
+        elem_type,  # type: int
+        shape,  # type: Optional[Sequence[Union[Text, int]]]
+        doc_string="",  # type: Text
+        shape_denotation=None,  # type: Optional[List[Text]]
```
wschin:
```suggestion
        elem_shape_denotation=None,  # type: Optional[List[Text]]
```

onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,665 @@
...
+ONNX_OPERATOR_SET_SCHEMA(
+    SequenceEmpty,
+    11,
+    OpSchema()
+        .SetDoc(SequenceEmpty_ver11_doc)
+        .Attr(
+            "dtype",
+            "(Optional) The data type of the tensors in the output sequence. "
+            "The default type is 'float'.",
+            AttributeProto::INT,
```
wschin: It should be a TypeProto but I don't know if the current framework allows it.

onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,665 @@
...
+        .TypeConstraint(
+            "T",
+            OpSchema::all_tensor_types(),
+            "Constrain input types to any tensor type.")
+        .TypeConstraint(
+            "S",
+            OpSchema::all_tensor_sequence_types(),
+            "Constrain output types to any tensor type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          size_t numInputs = ctx.getNumInputs();
```
wschin:
```suggestion
          const size_t numInputs = ctx.getNumInputs();
```

onnx/defs/sequence/defs.cc
```diff
@@ -0,0 +1,665 @@
...
+
+ONNX_OPERATOR_SET_SCHEMA(
+    SequenceConstruct,
+    11,
+    OpSchema()
+        .SetDoc(SequenceConstruct_ver11_doc)
+        .Input(
+            0,
+            "inputs",
+            "List of tensors.",
```
wschin:
```suggestion
            "Tensors.",
```

onnx/defs/sequence/defs.cc
gramalingam: Agree. We don't have any existing examples for doing this. We could use a string-based language (as in the specification of input/output types of operators) for specifying this.

docs/Changelog.md
BowenBao: No, I believe the constraints are the same with np.concatenate. This is taken from the Concat spec. I'll update it then.

onnx/defs/sequence/defs.cc
BowenBao: The current proto doesn't support it. It's good to add it once we are extending the functionalities further to support nested sequences.

Issue #2049, opened 2019-05-23
Troubles I met when trying to update docs


daquexian:
I ran [tools/update_doc.sh](tools/update_doc.sh) to update the docs.

I found

1. **(solved by myself in #2478)** The version of ONNX schema version [has been updated to 11](https://github.com/onnx/onnx/blob/master/onnx/defs/schema.h#L652), however, the onnx model has not been uploaded. For example, https://s3.amazonaws.com/download.onnx/models/opset_11/bvlc_alexnet.tar.gz is an invaild url.

   When running `tools/update_doc.sh`, the urls in `data.json` are edited to opset 11 ([link](https://github.com/onnx/onnx/blob/master/onnx/backend/test/cmd_tools.py#L36) and [link](https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/model/__init__.py#L61)), and are loaded in [link](https://github.com/onnx/onnx/blob/master/onnx/backend/test/loader/__init__.py#L45). **As a result, the process of updating docs cannot succeed.**

   The download will be skipped if there are already models of the same name on the disk. This could be the reason why the problem got ignored. However, **this problem prevents new contributors from contributing to ONNX.** 

1. **(Solved by myself in #2050)** The script echoes "Please run this script in the ONNX root folder" when "onnx/defs/gen_doc.py" _exists_.

1. **(Solved by myself in #2050)** The script runs `python setup.py develop` without `--user`. So the script can only be run as a root user. It's not the best practice. What's more, the script will create or edit many files, and they will be all owned by root. As a result, the regular user cannot delete, edit or git add them anymore without running chown manually. 

1. **(Unsolved)** I got the following error when running onnx/defs/gen_doc.py.

   > Traceback (most recent call last):
   >   File "/home/daquexian/repos/onnx/onnx/defs/gen_doc.py", line 350, in <module>
   >     main(Args)
   >   File "/home/daquexian/repos/onnx/onnx/defs/gen_doc.py", line 240, in main
   >     s += display_schema(schema, [schema])
   >   File "/home/daquexian/repos/onnx/onnx/defs/gen_doc.py", line 196, in display_schema
   >     type_constraint.type_param_str, allowedTypeStr)
   > UnboundLocalError: local variable 'allowedTypeStr' referenced before assignment

   After digging into it, I found a strange thing

   ```python
   Python 3.7.3 (default, Mar 26 2019, 21:43:19)
   [GCC 8.2.1 20181127] on linux
[...]

PR #2257, opened 2019-08-25
Update the description of nearest_mode of resize op

daquexian:
It is reported that "round_prefer_floor" and "round_prefer_ceil" are not clear. "round half down" and "round half up" are commonly used terms according to wikipedia.


daquexian:
I cannot re-generate the docs because of https://github.com/onnx/onnx/issues/2049 :|

daquexian:
I have re-generated the docs

PR #2262, opened 2019-08-26
[WIP] move most operator docs into separate folder

prasanthpul:
This is the first step to move static files into an operators folder. Next step would be to update the doc gen script


houseroad:
Fix the conflict?

wschin:
The broadcasting file is referenced in many operators' signatures, for example, https://github.com/onnx/onnx/blob/master/docs/Operators.md#add. I am not sure if we also need to update a C string somewhere in def.cc.

prasanthpul:
@wschin yes, it is referenced as a relative link by the doc gen srcipt. If we do the step 2 change (update generation of operators.md) then it will be a non-issue. Alternatively need to update it now to reference the subdirectory and then update it again later to remove the subdir from the path.

My preference is to wait on merging this for now and try to get the phase 2 done too. I've updated title to WIP to reflect this shouldn't be merged yet

PR #2269, opened 2019-08-29
Shape inference for NMS

hariharans29:
Implement shape inference for NMS op and add corresponding test.

CC: @linkerzhang @gramalingam @stevenlix 

One open question I had was since NMS has multiple optional scalar inputs that do not determine output shape, do we still need to validate their shape (if they are present) or let the runtime deal with faulty inputs ? There also seems to be a debate as to whether a "scalar" should strictly be a one-element tensor with 0 dims or whether it can be a one-element tensor with shape [1]. Strictly checking for something here might cause ShapeInference exceptions to be thrown for any existing working model and hence I did not validate the optional input shapes as they do not affect output shape anyway.


hariharans29:
@ebarsoum @gramalingam  - @stevenlix said he wanted this logic. So could you kindly take a look when you have a couple of minutes please ?

PR #2279, opened 2019-09-02
More recurrent tests

JamesAllingham:
This PR greatly expanded the test suite for the various recurrent operators (RNN, GRU, and LSTM). The following kinds of tests have been added (the `Operators.md` and `TestCoverage.md` have been regenerated to show these):

1. Tests for the reverse direction.
2. Tests for bidirectional.
3. Tests for the initial hidden state input.
4. Tests for the intermediate hidden states output.
5. A test for the initial cell state of the LSTM.
6. A test for the final cell state of the LSTM.
7. Tests for multiple outputs (e.g. final hidden state and final cell state).
8. Tests for clipping.
9. Tests for specifying the activation functions.

In order to add these tests I had to expand the support in the `RNN_Helper`, `GRU_Helper`, and `LSTM_Helper` classes. These helper classes now take both inputs and attributes as parameters of the constructor (as opposed to just the inputs). Note that the implementations for activation functions don't cover the optional activations and therefore don't cover the `activation_alpha` and `activation_beta` attributes.

I also renamed some variables for consistency between the RNN, GRU, and LSTM test cases as well as the `Operators.md` docs. For example, I renamed the output of the existing RNN, GRU, and LSTM test cases to `Y_h` (from `Y`) since this matches the naming scheme in `Operators.md`.

I've run the test suite and the type checker and have no errors with either. I've also compared the implementations of the helper classes against the implementations of the RNN, GRU, and LSTM in Mathematica for *most* of the tests cases to confirm that the behaviour is correct. The only ones I have not yet tested are the clipping and activation implementations. This is because we do not yet support such options in Mathematica. 

**TODO:**

1. Test the clipping implementation against another framework. **Does anyone have a suggestion for how to do this?** Neither Pytorch nor TensorFlow seems to support this option.

2. Tests the activation functions implementation. I'll use Pytorch for this. 

3. Test the `linear_before_reset` implementation. This was already implemented, so perhaps it was already tested, however, there were/are no test cases for it. **Does anyone know a framework which implements this attribute?** (So that I can test against it.)

Let me know if there is any feedback (particularly on coding style) while I am still working on the above, and please let me know about frameworks which I can test the implementation of the helper classes against.


JamesAllingham:
@MatteoSal

PR #2281, opened 2019-09-03
Supporting negative axes for all existing onnx ops

neginraoof:
I modified the ops documentations as well as tests, to add support for negative axes

Issue #2159, opened 2019-07-10
RNN ops input and output shapes are inconsistent with other ops


JamesAllingham:
Recurrent operators (`RNN`, `GRU`, `LSTM`) currently have the following input and output shapes:

```
X : T
    The input sequences packed (and potentially padded) into one 3-D tensor with the shape of `[seq_length, batch_size, input_size]`

initial_h (optional) : T
    Optional initial value of the hidden. If not specified - assumed to be 0. It has shape `[num_directions, batch_size, hidden_size]`.


Y (optional) : T
    A tensor that concats all the intermediate output values of the hidden. It has shape `[seq_length, num_directions, batch_size, hidden_size]`. 

Y_h (optional) : T
    The last output value of the hidden. It has shape `[num_directions, batch_size, hidden_size]`. 
```

Notice that the shapes of these inputs and outputs do not start with the batch size. On the other hand, *many* other ops, for example `Conv`, `ConvTranspose`, `AveragePool`, `MaxPool`, `SpaceToDepth`, `DepthToSpace`, `Multinomial`, `RoiAlign`, `BatchNorm`, etc., have inputs and outputs whose shapes start with the batch size. 

This is not ideal because it requires that extra `Transpose` operations be inserted when using the recurrent layers in conjunction with *batch first* layers. For example when exporting this simple net from Pytorch:

```
class Net(nn.Module):

    def __init__(self):
        super(Net, self).__init__()
        self.rnn1 = nn.RNN(10,10)
        self.fc1 = nn.Linear(10, 10)

    def forward(self, x):
        x, _ = self.rnn1(x)
        x = self.fc1(x)
        return x
```
we get an ONNX graph which contains an extra transpose:

```
%22 : Tensor, %23 : Float(1, 3, 10) = onnx::RNN[activations=["Tanh"], hidden_size=10](%input.1, %19, %20, %21, %17, %16), scope: Net/RNN[rnn1]
%24 : Float(10, 3, 10) = onnx::Squeeze[axes=[1]](%22), scope: Net/RNN[rnn1]
%25 : Float(10!, 10!) = onnx::Transpose[perm=[1, 0]](%fc1.weight), scope: Net/Linear[fc1]
%26 : Float(10, 3, 10) = onnx::MatMul(%24, %25), scope: Net/Linear[fc1]
```

The transpose would be unnecessary if the `RNN` output shape was batch first. 

[...]

PR #2284, opened 2019-09-05
Allow recurrent operations to be batchwise 

JamesAllingham:
This PR addresses issue #2159. The gist of the issue is that the recurrent ops (GRU, LSTM, and RNN) currently do not support batchwise implementations which exist in several popular deep learning frameworks, such as [PyTorch](https://pytorch.org/docs/stable/nn.html?highlight=rnn#torch.nn.RNN) and [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/keras/layers/RNN) (see the `batch_first` and `time_major` args respectively). This means that, when exporting batchwise operations from these frameworks to ONNX, transpose operations need to be inserted. 

Furthermore, many high-level frameworks, such as Keras, MATLAB's Deep Learning Toolbox, and Mathematica's Neural Networks library (for which I am a developer), abstract away the notion of batchsize from the construction of a network. In these cases, it would be convenient if ONNX supported batchwise recurrent operations.

To support batchwise operations, I have made two changes - one minor change and one breaking change:

1. I have added a `time_major` attribute to the recurrent ops. This attribute defaults to a value of 1, which means that the recurrent op is timewise (the existing behaviour). Any other value means that the recurrent op is batchwise. The name of this attribute was taken from the TensorFlow [docs](https://www.tensorflow.org/api_docs/python/tf/keras/layers/RNN). Alternative names are `batch_first` (from the PyTorch [docs](https://pytorch.org/docs/stable/nn.html?highlight=rnn#torch.nn.RNN)), `batchwise`, and `timewise`. 

2. (**Breaking change**) I have change the default order of the dimensions for all inputs and outputs so that the `num_directions` dim comes last (where applicable). 

The result of these two changes is as follows. By default, the dimensions of the inputs and outputs of the recurrent ops are:

- `X` - `[seq_length, batch_size, input_size]`
- `W` - `[3*hidden_size, input_size, num_directions]`
- `R` - `[3*hidden_size, hidden_size, num_directions]`
- `B` - `[6*hidden_size, num_directions]`
- `sequence_lens` - `[batch_size]`
- `initial_h` - `[batch_size, hidden_size, num_directions]`
- `initial_c` (**LSTM only**) - `[batch_size, hidden_size, num_directions]`
- `P` (**LSTM only**) - `[3*hidde_size, num_directions]`
- `Y` - `[seq_length, batch_size, hidden_size, num_directions]`
- `Y_h` - `[batch_size, hidden_size, num_directions]`
- `Y_c` (**LSTM only**) - `[batch_size, hidden_size, num_directions]`

Then, with `time_major != 1`, we have the following differences:

- `X` - `[batch_size, seq_length, input_size]`
- `Y` - `[batch_size, seq_length, hidden_size, num_directions]`

This means that:

1. If there is a `num_directions` dim, it always comes last. I think this is pretty nice since it makes dealing with multiple directions easier. Note that before it was usually first except for the output `Y` where it was second.
2. With `time_major == 1`, if there is a `seq_length` dim, it always comes first (as before).
3. With `time_major != 1`, if there is a `batch_size` dim, it always comes first. This means that we can easily support batchwise recurrent ops! 🥳

While, of course, I'd like feedback on anything and everything in this PR, there are a few things in particular that I wasn't sure about when making these changes:

1. The name `time_major`. My main thought for choosing the name was to see what other frameworks used, so both `time_major` and `batch_first` were options. I went with `time_major` because I wanted the attribute to match our current behaviour by default.

2. How to update the `op_version`s for the recurrent ops. I updated the versions to 11, since it seemed like this was what other recent PRs (e.g. #2281) had done. However, [the versioning document](https://github.com/fdwr/onnx/blob/master/docs/Versioning.md#operator-versioning) made me think that perhaps I should have gone to 12 and also incremented the maximum version in `DomainToVersionRange`. What was I supposed to do here?

Finally, here is a summary of all of the changes this PR makes:

1. Copies the old definitions of the recurrent ops from `rnn/defs.cc` to `rnn/old.cc`.
2. Updates the definitions in `defs.cc` to include the `time_major` attribute and the new default ordering of the dims (with `num_directions` last).
    * This includes updating the `RNNShapeInference` function.
3. Adds the new version numbers to `operator_sets.h`.
4. Adds tests for the `time_major != 1` case for each of the recurrent ops.
    * This required extending the implementation of `RNN_Helper` and friends.
5. Adds a test for the `time_major != 1` case to the dedicated shape inferences tests in `shape_inference_tests.py`.
6. Updates all exisiting recurrent op test cases for the new default dimension ordering. This includes the dedicated shape inference tests.
7. Re-generates the test data.
8. Re-generates the `Changelog.md`, `Operators.md`, and `TestCoverage.md` docs.


JamesAllingham:
Should I also be adding a converter to upgrade recurrent ops from version 7 to 11? (Like those [here](https://github.com/onnx/onnx/tree/master/onnx/version_converter/adapters)? I assume so, but I haven't seen any docs referring to writing converters, so any advice would be much appreciated.

wschin:
> Should I also be adding a converter to upgrade recurrent ops from version 7 to 11? (Like those here?) I assume so, but I haven't seen any docs referring to writing converters, so any advice would be much appreciated.

Maybe https://github.com/onnx/onnx/blob/57372f39631fddc720287a4703040c0551d4ee48/docs/VersionConverter.md?

JamesAllingham:
> 
> 
> > Should I also be adding a converter to upgrade recurrent ops from version 7 to 11? (Like those here?) I assume so, but I haven't seen any docs referring to writing converters, so any advice would be much appreciated.
> 
> Maybe https://github.com/onnx/onnx/blob/57372f39631fddc720287a4703040c0551d4ee48/docs/VersionConverter.md? I will take a closer to your changes soon. They are not small, so we need to be careful.

Thanks for pointing to that doc, I missed it somehow. I'll take a look at writing a converter. I think that it would be great if the doc outlined the steps required for implementing a converter & linked to a good example of a PR adding a coverter, so I'll also try and add that.

I do understand that these are not insignificant changes, so I am more than happy for you to take your time reviewing this. I'm also more than happy to discuss any issues that come up and make changes as necessary.

JamesAllingham:
@matteosal

gramalingam:
A couple of comments:
(a) The breaking change for num_directions is a potential concern. Why not make this also controlled by an attribute so that, by default, the old behavior is preserved?
(b) The scan operation (https://github.com/onnx/onnx/blob/master/docs/Operators.md#scan ) allows users to specify any axis value for the sequence axis (instead of just axis 0 or axis 1). Would that be helpful in some contexts, or is that an unnecessary generalization/complication?

postrational:
Superseded by: https://github.com/onnx/onnx/pull/2922

PR #2288, opened 2019-09-05
Improve compatiblity with proto3 and enable reading attributes

yuslepukhin:
Proto3 does not output primitive types into a stream if they match type default.
This presents a problem when C# protobuf generated code creates a model where type default are omitted. Our proto files are compatible with proto3 and in some cases special provisions exist such as AttributeType enum which actually indicate whether the Attribute was really set by the user.
Thus we remove has_i(), has_f() and has_s() checks if the corresponding enum is set. 
This allows proto2 code to read models serialized by proto3 code.


gramalingam:
It would help other users if we could document our guidelines for ensuring proto2 and proto3 compatibility (as well as compatibility across different languages in case it makes a difference). My understanding (for this PR) is the following, is this correct? DO NOT use the has_foo() method for any field "foo" of primitive or enum types; it is treated as being present with a default value (if absent in the serialized representation). However, for all the non-primitive types (which are always declared optional in ONNX) as well as oneof fields, it is safe to use the "has_foo" method. Is this correct?

yuslepukhin:
> 
> 
> It would help other users if we could document our guidelines for ensuring proto2 and proto3 compatibility (as well as compatibility across different languages in case it makes a difference). My understanding (for this PR) is the following, is this correct? DO NOT use the has_foo() method for any field "foo" of primitive or enum types; it is treated as being present with a default value (if absent in the serialized representation). However, for all the non-primitive types (which are always declared optional in ONNX) as well as oneof fields, it is safe to use the "has_foo" method. Is this correct?

To be a little more specific. The reason we are able not to use `has_*()` on the primitives is because when we assign a value to them we also set the enum. So if enum is not set then it is `UNDEFINED` (0)  by default, otherwise simply read the primitive type because we know the value is correct. This works when primitives are intentionally paired with enums and 0 means `UNDEFINED`. There are comments in the proto files to that effect.

We also do not want to use `has_*()` on `oneof` fields because in that case proto compiler automatically generates a corresponding enum. That enum value can tell us which field is set if any.

Otherwise, we can use `has_*()` providing we are built with `proto2` library. `proto3` compiler does not generate `has_*()` methods, which would become an issue if we migrated to `proto3`.

yuslepukhin:
@houseroad Can you, please, be a bit more specific as to what you mean to change existing CI to proto3? Current code will not compile with proto3 interfaces, mostly because of the extensive use of `has_*()` methods. So that would call for refactoring. However, changing to proto3 would require even more discussions. I am trying alleviate a specific pain point where compatibility is required.

PR #2291, opened 2019-09-06
Updated docs for strides and dilations attributes 

JamesAllingham:
The docs for the `strides` and `dilations` attributes of various convolutional and pooling layers did not mention the defaults. I copied the way the defaults were described for the `pads` attribute.


onnx/defs/nn/defs.cc
```diff
@@ -876,17 +882,24 @@ computes the output.)DOC";
...
         std::string("NOTSET"));
-    schema.Attr("pads", pads_doc, AttributeProto::INTS, OPTIONAL);
+    schema.Attr(
+        "pads",
+        pads_doc,
+        AttributeProto::INTS,
+        OPTIONAL);
     schema.Attr(
         "group",
         "number of groups input channels and output channels are divided into.",
```
wschin: What's the default value of `group`?

wschin:
Do you mean that you make some `undefined` behaviors to `defined`? If yes, please bump their version numbers.

JamesAllingham:
> 
> 
> Do you mean that you make some `undefined` behaviors to `defined`? If yes, please bump their version numbers.

So the docs didn't mention the

onnx/defs/nn/defs.cc
JamesAllingham: Good catch! The default value is 1. I'll add that.

JamesAllingham:
> 
> 
> Do you mean that you make some `undefined` behaviors to `defined`? If yes, please bump their version numbers.

Yes, I suppose I did... these defaults were mentioned in [conv.py](https://github.com/onnx/onnx/blob/57372f39631fddc720287a4703040c0551d4ee48/onnx/backend/test/case/node/conv.py) but I suppose that isn't a proper way to define things. I'll bump the version numbers.

wschin:
Could you merge the conflicts? I think we can just keep both of your and master's changes and rerun document/test update.

PR #2294, opened 2019-09-10
Fix resize shape inference issue in opset10

BowenBao:
The issue is found in https://github.com/microsoft/onnxruntime/pull/1756, that the updated opset 10 resize shape inference in defs/tensor/old.cc is failing Yolov3 https://github.com/onnx/models/tree/master/vision/object_detection_segmentation/yolov3.

This is a simple fix to bring back and use the version of shape inference functions before opset11. 

@daquexian @hariharans29 @yuslepukhin


onnx/defs/tensor/old.cc
```diff
@@ -485,52 +485,7 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            }
-          } else {
-            if (0 == output_shape->dim_size()) {
-              // Infer output shape's rank in any case.
-              for (int i = 0; i < input_shape.dim_size(); ++i) {
-                output_shape->add_dim();
-              }
-            }
-          }
+          resizeShapeInference_opset7(ctx);
```
wschin: The version of this `Resize` is 10. Is it compatible with 7's shape inference code?

BowenBao: Yes, it was added in https://github.com/onnx/onnx/pull/2085/. Ops between opset7 and opset10 were calling it. I feel naming it with lowest supported opset version makes more sense then highest. If you think it's not too verbose i could also change it to `resizeShapeInference_opset7_to_10` ...

wschin: It'd be better to call it `*7_to_10` because you already saw a stupid question from me. :)

onnx/defs/tensor/utils.cc
```diff
@@ -98,4 +98,73 @@ void resizeShapeInference(InferenceContext& ctx, bool is_resize_op) {
...
+}
+
+void resizeShapeInference_opset7(InferenceContext& ctx) {
+  if (!hasNInputShapes(ctx, 1)) {
+    return;
+  }
+  propagateElemTypeFromInputToOutput(ctx, 0, 0);
+  const auto& input_shape = getInputShape(ctx, 0);
+  auto* output_shape = getOutputShape(ctx, 0);
+  const auto scales = ctx.getInputData(1);
```
hariharans29: nit: const auto* maybe?

onnx/defs/tensor/utils.cc
```diff
@@ -98,4 +98,73 @@ void resizeShapeInference(InferenceContext& ctx, bool is_resize_op) {
...
+      } // dim->has_dim_value()
+    } // input_shape.dim(i).has_dim_value()
+  }
+}
+
+void resizeShapeInference_opset7(InferenceContext& ctx) {
+  if (!hasNInputShapes(ctx, 1)) {
+    return;
+  }
+  propagateElemTypeFromInputToOutput(ctx, 0, 0);
```
hariharans29: move this line before the 'if' condition ? We can do TypeInference even if we can't proceed with ShapeInference

BowenBao: sure, that makes sense.
Note this PR is just reverting back the shape inference function for these ops < opset 11. I would prefer keeping the Upsample-7 the way it was to reduce the risk of another regression.

hariharans29: sure Bowen :)

PR #2295, opened 2019-09-10
Delete setup.sh

96KingH96:


prasanthpul:
Was this PR intentional? What is the purpose?

PR #2296, opened 2019-09-10
Fix shapeinference function

jignparm:
Updated the shape inference function for CumSum.


wschin:
Could you please add a test into [this file](https://github.com/onnx/onnx/blob/7988d8360b11e6003560076e9b1d4aa426db3244/onnx/test/shape_inference_test.py#L2097)?

jignparm:
> Could you please add a test into [this file](https://github.com/onnx/onnx/blob/7988d8360b11e6003560076e9b1d4aa426db3244/onnx/test/shape_inference_test.py#L2097)?

Done.

wschin:
@ebarsoum, how does this PR look to you?

PR #2304, opened 2019-09-11
Change incorrect use of ValueError to TypeError

autoih:


onnx/numpy_helper.py
```diff
@@ -33,7 +33,7 @@ def to_array(tensor):  # type: (TensorProto) -> np.ndarray[Any]
         raise ValueError(
             "Currently not supporting loading segments.")
     if tensor.data_type == TensorProto.UNDEFINED:
-        raise ValueError("The data type is not defined.")
+        raise TypeError("The data type is not defined.")
```
wschin:
```suggestion
        raise TypeError("The element type in the input tensor is not defined.")
```

autoih: Thanks @wschin for the revision. I've updated it.

Issue #2286, opened 2019-09-05
Relaxing the requirement for types of inputs/outputs of subgraphs


gramalingam:
GraphProto (as originally designed) required that the types MUST be specified for the inputs and outputs. Control-flow ops added later use a GraphProto as an attribute to specify the "body" of a loop or if-then-else statement. However, the types are not strictly necessary for such sub-graphs as they can typically be inferred by the corresponding control-flow-op's type-inference logic.

This is a proposal to relax the type-requirement for subgraphs (graphs specified as attributes of higher-order ops). (If a specific op requires this information, the op's spec can indicate so. In other words, whether the type is required is up to the op's specification.)

It looks like this relaxation has already been done in the checker, though neither the comments in the proto file nor the documentation in IR.md reflect this. Hence, this issue to get feedback about this relaxation.

gramalingam:
The function implementation of the Range op (see https://github.com/onnx/onnx/pull/2287 ) is an example use of this relaxation.

gramalingam:
Here is the original PR that relaxed the checker: https://github.com/onnx/onnx/pull/2009 .

Ideally, we should be using the IR-version to decide whether to do the stronger check (as earlier) or the weaker check (as currently). Any thoughts about this?

skottmckay:
Makes sense to me to not require type info for subgraphs inputs/outputs given it can be inferred. 

As for the IR version I agree the most correct thing would be to base the check on it. Given the original change is from May one could argue it's not necessary to increment the IR verison in this case given we're relaxing the rules and five months have passed since the checker change without (afaik) issue. I don't have a good feel for whether frequent IR version changes cause issues/pain for implementers to say whether it's worth avoiding an IR version update though.

PR #2310, opened 2019-09-13
Update documentation about required input output types

gramalingam:
The types of inputs and outputs are required only for the top-level graph and not for subgraphs. Update the documentation to clarify this. (Also clarify documentation about variadic inputs and outputs.)


gramalingam:
@houseroad : see Issue https://github.com/onnx/onnx/issues/2286 and the earlier PR https://github.com/onnx/onnx/pull/2009

PR #2317, opened 2019-09-16
Update IR doc to clarify initializers can be node inputs

TMVector:
Since IR v4, the restriction that initializers must be graph inputs has been removed. This just makes it clear in the modified table.

PR #2320, opened 2019-09-16
Update IR doc to clarify initializers are permitted as node inputs

gramalingam:
Minor fix in IR documentation to keep up with the relaxation that initializers need not be graph inputs. (Duplicate of the PR https://github.com/onnx/onnx/pull/2317 which was abandoned.)

### DIFF
--- a/docs/IR.md
+++ b/docs/IR.md
@@ -208,7 +208,7 @@ They have the following properties:
 Name|Type|Description
 |---|---|---|
 name|string|An optional name of the node, used for diagnostic purposes only.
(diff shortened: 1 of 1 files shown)


snnn:
I think it should say: 
when IR_VERSION < 4, .... , else .... .

Because a model with IR_VERSION ==3 is still legit.

gramalingam:
@snnn : the way it has been worded, it is valid for all IR_VERSIONs.

snnn:
It's valid but not clear.

Issue #2318, opened 2019-09-16
External data is undocumented


TMVector:
Support for storing large tensor values in external files was introduced in #678, but AFAICT is undocumented.

This is a pretty important feature, functionally, but it's also important for end users who may not realise that they need to move around more than just the `*.onnx` file.

I would suggest it should be documented in IR.md, and perhaps there are other locations from which it could be signposted? I would draft changes myself but I don't think I can contribute during work hours due to the CLA.

cc @postrational

PR #2323, opened 2019-09-17
Document external tensor data

TMVector:
This is a *draft* documentation proposal; discussion needed.

---

Per #2318, support for tensor data to be stored in external files was added in #678, but I couldn't find documentation for it aside from in a comment the proto definition.

I'm not sure how exactly this should be documented, since it's not clear which parts of the proto definition should be specified in IR.md and which are syntactical details.

However, IMO, external data files need to be clearly documented _somewhere_* because it could trip up users who aren't interested in the lower-level details of the IR. I'm not sure my suggestion is massively more user friendly -- please help :innocent:.

\*Arguably it _is_ documented in the proto, but diving into proto definitions and python testing code feels not very user friendly.

---

cc @postrational @houseroad


docs/IR.md
```diff
@@ -282,6 +282,9 @@ There are two ways to leave an optional input or output unspecified: the first,
 
 Each node referring to an operator with optional outputs MUST provide a name for each output that is computed and MUST NOT provide names for outputs that are not computed.
 
+#### External Tensor Data
+
+The raw data for large constant tensors, such as initializers, MAY be serialised in a separate file. In such a case, the tensor MUST provide the filename relative to the model file and MUST NOT use the value fields. It MAY provide a byte offset and length within that file. It MAY also specify a SHA1 digest of the file. One file MAY contain the data for multiple tensors.
```
snnn: However, this feature is experimental, it may break shape inference.  And we're not going to fix it.

TMVector: I cannot find anything saying external raw data is experimental in code/docs/PR comments, so if that's true then documenting its experimental status should be high priority.  
AFAICT, external data is supported at least by the ONNX Python API (onnx/onnx#678) and ONNX Runtime (microsoft/onnxruntime#520 e.g. test file `opset9/tf_resnet_v2_50/model.onnx`).

I wouldn't expect it to break type inference, since the type/shape is stored in the [`data_type`/`dims` fields](https://github.com/onnx/onnx/blob/master/onnx/onnx.in.proto#L351-L356) on TensorProto, not the `raw_data` field.

Apologies if I misunderstood something.

wschin: It can breaks shape inference, for example, when `Reshape`'s shape is stored in external data.

TMVector: Okay, that makes sense.

Maybe a dumb question, but why are tensors in external files considered inaccessible for shape inference?

snnn: It could, but it need a big effort.

linkerzhang:
I'm viewing this clarification is still correct and good, as part of the spec (docs).

Shape inference (functionality wise) is not a MUST. I'm viewing that as a helper tool. It should/can be either fixed in this repo or users fix them in their own repo (env).

PR #2326, opened 2019-09-17
Fix bug in get-attribute

gramalingam:
Replace "is" by ==. This bug is breaking gen_doc.py.


wschin:
It's be nice if you can add a test.

gramalingam:
```
   It's be nice if you can add a test.
```
Agree. I am not quite sure when this manifests itself. I understand that "is" is a pointer-equality comparison (same object comparison), but here we are comparing primitives. In simple test-cases, "is" seems to work like "==" for primitive type … but may be it fails when object values are serialized from C++ and received in python? However, after this change, gen_doc.py works fine for me.

PR #1855, opened 2019-03-11
Infer output shape of Expand op

shinh:

PR #2328, opened 2019-09-18
Fix shape inference for Split with split attribute

### LINKED ISSUE #1735: Shape inference fails with a Split node with an split attribute
When a `split` attribute is set to a `Split` node, `onnx.shape_inference.infer_shapes` fails to infer its output shapes.

```
import onnx
import onnx.shape_inference

op = onnx.helper.make_node("Split", inputs=["a"], outputs=["b", "c", "d"], axis=0)
i = onnx.helper.make_tensor_value_info(name="a", elem_type=onnx.TensorProto.FLOAT, shape=[6])
model = onnx.helper.make_model(onnx.helper.make_graph([op], "graph", [i], []))
print(onnx.shape_inference.infer_shapes(model)) # contains proper value_info

op = onnx.helper.make_node("Split", inputs=["a"], outputs=["b", "c", "d"], axis=0, split=[1, 2, 3])
i = onnx.helper.make_tensor_value_info(name="a", elem_type=onnx.TensorProto.FLOAT, shape=[6])
model = onnx.helper.make_model(onnx.helper.make_graph([op], "graph", [i], []))
print(onnx.shape_inference.infer_shapes(model)) # the shapes of b, c, and d are not inferred 
```

While the former example outputs the correct shapes (three `[2]`), the resulting `value_info` of the latter example do not contain dimension information.
This will happen even if the tensor `a` is divided evenly (by setting `split=[2, 2, 2]`).
I'm using ONNX 1.3.0.

shinh:
This fixes #1735


onnx/defs/tensor/defs.cc
```diff
@@ -403,46 +403,59 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                "Invalid value of attribute 'axis'. Rank=",
+                rank,
+                " Value=",
+                axis);
+          }
+          if (axis < 0) {
+            axis += rank;
+          }
+          const auto& splitDim = shape.dim(axis);
+          if (!splitDim.has_dim_value()) {
```
hariharans29: we can still do rank inference as the ranks of the outputs will be the same as the rank of the input before returning.

onnx/defs/tensor/defs.cc
```diff
@@ -403,46 +403,59 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            if (axis < 0) {
-              axis += rank;
+            int64_t totalDim = 0;
+            for (int64_t d : split) {
+              totalDim += d;
             }
-            const auto& splitDim = shape.dim(axis);
-            if (!splitDim.has_dim_value()) {
+            if (totalDim != splitDimValue) {
               return;
```
hariharans29: should this be an error ?

onnx/defs/tensor/defs.cc
```diff
@@ -403,46 +403,59 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             }
-            int splitDimValue = static_cast<int>(splitDim.dim_value());
+          } else {
             int chunkSize =
```
hariharans29: Per spec, it says if 'split' is not provided, split it into equal sized parts, so should the case where splitDimValue % ctx.getNumOutputs() != 0 be an error condition ?

hariharans29: Otherwise, we might have to arbitrarily deal with it like the logic in 444-448. The logic is not prescribed per spec. So I think this should be an error condition as well. To me there are 3 possible error conditions in this op-

1) When 'split' is present and its count does not match number of outputs
2) When 'split' is present and its sum does not match input shape's corresponding value in the 'axis' dimension
3) When 'split' is not present and  input shape's corresponding value in the 'axis' dimension cannot be split evenly to the number of outputs

onnx/defs/tensor/defs.cc
shinh: Done

onnx/defs/tensor/defs.cc
shinh: Done

onnx/defs/tensor/defs.cc
shinh: Done

shinh:
Thanks for your review! I think I handled all of your comments. Please take a look again?

onnx/test/shape_inference_test.py
```diff
@@ -1097,6 +1097,36 @@ def test_split_negative_axis(self):  # type: () -> None
...
+
+    def test_split_with_split_attribute_unknown_split_dim(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, (2, 'a', 'b'))],
+            [make_node('Split', ['x'], ['y', 'z'], axis=1, split=[3, 1])],
+            [])
+        self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (2, None, 'b')),  # type: ignore
+                                      make_tensor_value_info('z', TensorProto.FLOAT, (2, None, 'b'))])  # type: ignore
+
+    def test_split_fail_with_invalid_split_attribute(self):  # type: () -> None
```
hariharans29: Nit: I don't think we explicitly add tests for failure scenarios. Seems like these 2 may not be needed.

onnx/test/shape_inference_test.py
```diff
@@ -1097,6 +1097,36 @@ def test_split_negative_axis(self):  # type: () -> None
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (2, 2)),
                                       make_tensor_value_info('z', TensorProto.FLOAT, (2, 2))])
 
+    def test_split_with_split_attribute(self):  # type: () -> None
```
hariharans29: can you please add 2 more tests - 

1) One with split attribute is missing and the default 'equal' split occurs for each output
2) One with (valid) negative axis

onnx/defs/tensor/defs.cc
```diff
@@ -403,46 +403,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                  ")");
+            }
+          } else {
+            int numOutputs = static_cast<int>(ctx.getNumOutputs());
+            if (splitDimValue % numOutputs != 0) {
+              fail_shape_inference("The input is not evenly splittable");
+            }
+            int chunkSize = splitDimValue / numOutputs;
+            for (int i = 0; i < static_cast<int>(ctx.getNumOutputs()); i++) {
+              split.push_back(chunkSize);
```
hariharans29: nit 1: It seems a little ugly to store the same value 'ChunkSize' in the vector for each output. it would be great if the single value 'chunkSize' can just be handled below when you actually assign the dim value.

nit 2: 'chunkSize' doesn't seem c++ style. You can consider using 'chunk_size'

hariharans29:
> Thanks for your review! I think I handled all of your comments. Please take a look again?

Just some minor comments to be addressed. Looks good otherwise. :)

onnx/test/shape_inference_test.py
shinh: Isn't the existing test case, `test_split_negative_axis` enough to cover these two cases? Do you want me to split it into two tests?

onnx/defs/tensor/defs.cc
shinh: As for 2, fixed.

As for 1, let me double-check if you like the change, since I feel current code is better. If I follow your suggestion, I think we'll have two copies of the for loop which assigns `dim_value` like

```
if (getRepeatedAttribute(ctx, "split", split)) {
   ...
   for (size_t i = 0; i < ctx.getNumOutputs(); i++) {
       *ctx.getOutputType(i)->mutable_tensor_type()->mutable_shape() =
            shape;
        ctx.getOutputType(i)
            ->mutable_tensor_type()
             ->mutable_shape()
             ->mutable_dim(axis)
             ->set_dim_value(split[i]);
    }
} else {
  ...
  int chunk_size = split_dim_value / num_outputs;
   for (size_t i = 0; i < ctx.getNumOutputs(); i++) {
       *ctx.getOutputType(i)->mutable_tensor_type()->mutable_shape() =
            shape;
        ctx.getOutputType(i)
            ->mutable_tensor_type()
             ->mutable_shape()
             ->mutable_dim(axis)
             ->set_dim_value(chunk_size);
    }
}
```
Is this what you want?

onnx/test/shape_inference_test.py
shinh: Removed

shinh:
Thanks for your review! Let me check a couple of things before I actually change them.

shinh:
The test was randomly failing so I pushed it several times to trigger the rebuild.

shinh:
@hariharans29 Ping?

hariharans29:
Sorry - I missed this. Will take a look tomorrow. Thanks for the ping.

onnx/test/shape_inference_test.py
hariharans29: You are right. :) I did not notice the other test.

onnx/defs/tensor/defs.cc
hariharans29: That's fine. We can just keep it the way it is for now.

shinh:
@wschin: Friendly reminder :)

hariharans29:
@gramalingam @ebarsoum - pinging more people for reviews 

@shinh - Don't worry. This PR has been tagged for release 1.7. 1.6 just got done last week. So it will be picked up when the PRs start getting merged for the next release (maybe sooner).

shinh:
@hariharans29: Got it, thanks! I'm worried another person makes a similar patch and there'll be duplicated effort. It happened once in https://github.com/onnx/onnx/pull/1855 (and you were the author of https://github.com/onnx/onnx/pull/2041 :).

hariharans29:
@shinh - I see. Sorry, I should have checked if there was already a PR for Expand shape inference. Someone on my team needed it urgently, and I just wrote it without checking for any existing PRs.

PR #2329, opened 2019-09-18
apply unittest assertion method

autoih:


autoih:
Looks, can't use an instance to run the assertion.

PR #2330, opened 2019-09-18
Gemm optional bias

JamesAllingham:
As discussed (briefly) in #2060 it would be useful for the bias term (the `'C'` input) of the `'Gemm'` operation to be optional. This PR adds this functionality. 

Specifically, if the bias is not specified it is assumed to be 0. This is a common behaviour among deep learning frameworks.

This PR:
1. Updates the definition (in `math/defs.cc`) to make the `'C'` input optional. The version for the `Gemm` operator is bumped to 11, and the old definition is moved to `math/old.cc`. Similarly, `operator_sets.h` is updated to reflect the new version number.

2. Updates the `gemm_reference_implementation` to allow for `'C'` being optional. 

3. Adds a test case for no bias term.


wschin:
+ some people from SIG operator @ebarsoum, @spandantiwari, @gramalingam for visibility. I personally consider this change is reasonable for some reasons below.

1. We have `MatMul` and `Add` but still introduced `Gemm` for this opportunity of further computation optimization. Because 2-D matrix multiplication is even more fundamental than Gemm, I feel it's reasonable to allow it in Gemm.
2. Most frameworks can easily implement it given their existing `Gemm` code.
3. From user's perspective, it's sometimes annoying to use `Gemm` because we need to create zero initializer only for matching the spec.

spandantiwari:
@wschin @JamesAllingham Just curious is this is being tracked for 1.6 release. If not, then we may have to update the opset version number in the PR from 11 to 12.

docs/Operators.md
```diff
@@ -5572,13 +5572,13 @@ expect(node, inputs=[data, indices], outputs=[output],
   input tensor B has shape (K, N) or (N, K), input tensor C is broadcastable to shape (M, N),
   and output tensor Y has shape (M, N). A will be transposed before doing the
   computation if attribute transA is non-zero, same for B and transB.
-  This operator supports **unidirectional broadcasting** (tensor C should be unidirectional broadcastable to tensor A * B); for more details please check [the doc](Broadcasting.md).
+  This operator supports **unidirectional broadcasting** (tensor C should be unidirectional broadcastable to tensor A * B); for more details please check [the doc](Broadcasting.md).This operator has **optional** inputs/outputs. See [the doc](IR.md) for more details about the representation of optional arguments. An empty string may be used in the place of an actual argument's name to indicate a missing argument. Trailing optional arguments (those not followed by an argument that is present) may also be simply omitted.
```
```suggestion
  This operator supports **unidirectional broadcasting** (tensor C should be unidirectional broadcastable to tensor A * B); for more details please check [the doc](Broadcasting.md). This operator has **optional** inputs/outputs. See [the doc](IR.md) for more details about the representation of optional arguments. An empty string may be used in the place of an actual argument's name to indicate a missing argument. Trailing optional arguments (those not followed by an argument that is present) may also be simply omitted.
```

docs/Operators.md
```diff
@@ -5593,15 +5593,15 @@ Other versions of this operator: <a href="Changelog.md#Gemm-1">Gemm-1</a>, <a hr
...
 
 <dl>
 <dt><tt>A</tt> : T</dt>
 <dd>Input tensor A. The shape of A should be (M, K) if transA is 0, or (K, M) if transA is non-zero.</dd>
 <dt><tt>B</tt> : T</dt>
 <dd>Input tensor B. The shape of B should be (K, N) if transB is 0, or (N, K) if transB is non-zero.</dd>
-<dt><tt>C</tt> : T</dt>
-<dd>Input tensor C. The shape of C should be unidirectional broadcastable to (M, N).</dd>
+<dt><tt>C</tt> (optional) : T</dt>
+<dd>Optional input tensor C. If not specified - assumed to be 0. The shape of C should be unidirectional broadcastable to (M, N).</dd>
```
```suggestion
<dd>Optional input tensor C. If not specified, the computation is done as if C is a scalar 0. The shape of C should be unidirectional broadcastable to (M, N).</dd>
```

onnx/backend/test/case/node/gemm.py
```diff
@@ -4,16 +4,18 @@
...
 from . import expect
 
 
-def gemm_reference_implementation(A, B, C, alpha=1., beta=1., transA=0,
-                                  transB=0):  # type: (np.ndarray, np.ndarray, np.ndarray, float, float, int, int) -> np.ndarray
+def gemm_reference_implementation(A, B, C=None, alpha=1., beta=1., transA=0,
+                                  transB=0):  # type: (np.ndarray, np.ndarray, Optional[np.ndarray], float, float, int, int) -> np.ndarray
     A = A if transA == 0 else A.T
     B = B if transB == 0 else B.T
+    C = C if C is not None else np.array(0)
```
wschin: Very nice. It matches the spec perfectly.

onnx/defs/math/defs.cc
```diff
@@ -742,11 +742,12 @@ computation if attribute transA is non-zero, same for B and transB.
 
 ONNX_OPERATOR_SET_SCHEMA(
     Gemm,
-    9,
+    11,
     OpSchema()
         .SetDoc(
             Gemm_ver9_doc +
```
wschin: ver9 --> ver11.

wschin:
> @wschin @JamesAllingham Just curious if this is being tracked for 1.6 release. If not, then we may have to update the opset version number in the PR from 11 to 12.

It's a nice to have thing in 1.6. If we can't make it on time, we can bump the opset version if you want.

wschin:
@JamesAllingham, we need a shape inference test as well.

onnx/defs/math/defs.cc
JamesAllingham: Fixed :)

docs/Operators.md
JamesAllingham: Good change, I've implemented it.

docs/Operators.md
JamesAllingham: I ended up adding a new line rather than a space, let me know if you feel strongly about the space and I'll change it.

JamesAllingham:
> 
> 
> @JamesAllingham, we need a shape inference test as well.

I've added a test but I wasn't 100% sure what you wanted here so let me know if you wanted something else.

docs/Operators.md
wschin: New line sounds good. Thanks.

wschin:
> @JamesAllingham, we need a shape inference test as well.
> 
> I've added a test but I wasn't 100% sure what you wanted here so let me know if you wanted something else.

Ah, my bad. I missed the last file.

JamesAllingham:
> 
> 
> > @JamesAllingham, we need a shape inference test as well.
> > I've added a test but I wasn't 100% sure what you wanted here so let me know if you wanted something else.
> 
> Ah, my bad. I missed the last file.

No, I only just added it, you didn't miss anything!

onnx/defs/math/old.cc
```diff
@@ -1126,6 +1126,100 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            auto transAAttr = ctx.getAttribute("transA");
+            bool transA =
+                transAAttr ? static_cast<int>(transAAttr->i()) != 0 : false;
+            auto transBAttr = ctx.getAttribute("transB");
+            bool transB =
+                transBAttr ? static_cast<int>(transBAttr->i()) != 0 : false;
+            auto& first_input_shape = getInputShape(ctx, 0);
+            auto& second_input_shape = getInputShape(ctx, 1);
+            if (first_input_shape.dim_size() != 2)
+              fail_shape_inference("First input does not have rank 2");
```
spandantiwari: nit: Minor point about code style - consider adding braces even for single line scopes to be consistent with coding style in this file.

JamesAllingham: Will do 👍

PR #2332, opened 2019-09-18
external_data: Don't move typed value fields to external data

TMVector:
The current implementation of `convert_model_to_external_data` sets all tensors to use external data even if they store their data in the typed value fields, not the raw_data field. That leads to an error when then calling `write_external_data_tensors`:

> ValueError: raw_data field doesn't exist.

This change just silently ignores them, leaving them inside the ModelProto. That might not be the behaviour desired in all cases, but I might suggest that would be better addressed by adding a `convert_model_tensors_to_raw_data` function for users to call before `convert_model_to_external_data`?


askhade:
@TMVector : The change looks good to me can you please sign the DCO and add a test case for this. Thanks!

gramalingam:
Thanks @TMVector for this PR. I am trying to understand the intention of the original code. I wonder if this requires some more cleanup. Considering ```convert_model_to_external_data``` : it is unclear how it is supposed to save multiple tensors in the same file when it is not using different offsets for different tensors. Further, I guess it is supposed to be used in conjunction with ```write_external_data_tensors```; it seems to me it would be better to expose one method that combines both together. I realize that this issue is not due to the current PR. Adding @yufenglee @postrational for comments.

onnx/external_data_helper.py
```diff
@@ -96,7 +98,7 @@ def set_external_data(tensor,  # type: TensorProto
 def convert_model_to_external_data(model, all_tensors_to_one_file=True, location=None):
     # type: (ModelProto, bool, Optional[Text]) -> None
     """
-    call to set all tensors as external data. save_model saves all the tensors data as external data after calling this function.
+    Call to set all tensors as external data. save_model saves all the tensors data as external data after calling this function.
```
gramalingam, line 101: Would help to clarify in the documentation that only raw_data tensors will be transformed, not typed tensors.

onnx/external_data_helper.py
```diff
@@ -78,6 +78,8 @@ def set_external_data(tensor,  # type: TensorProto
                       checksum=None,  # type: Optional[Text]
                       basepath=None  # type: Optional[Text]
                       ):  # type: (...) -> None
+    if not tensor.HasField("raw_data"):
```
gramalingam, line 81: Unclear if this change is required here. It seems to be a minor utility; may be better to not even expose this externally. But if it is used externally, the correctness depends on the usage, I suppose. But I am okay with adding this as extra check for now.

onnx/external_data_helper.py
```diff
@@ -109,15 +111,17 @@ def convert_model_to_external_data(model, all_tensors_to_one_file=True, location
...
             file_name = location
         for tensor in _get_all_tensors(model):
-            set_external_data(tensor, file_name)
+            if tensor.HasField("raw_data"):
+                set_external_data(tensor, file_name)
     else:
         for tensor in _get_all_tensors(model):
-            set_external_data(tensor, tensor.name)
+            if tensor.HasField("raw_data"):
+                set_external_data(tensor, tensor.name)
```
jcwchen, line 119: Other than tensor without 'raw_data', do all other kinds of tensors need to be converted to external data? (The original method looks like brute-force to me and large model will be separated into a lot of files in this case) If yes, ORT would encounter some issues from some tensors with external data. https://github.com/onnx/onnx/issues/3157 Tensor checking and unpacking with external data needs to be fixed.

Once the decision is made, it would be great if ONNX can clarify the usage of external data here: https://github.com/onnx/onnx/blob/master/docs/IR.md#external-tensor-data.

askhade:
save_external_data creates a new file for a tensor if it already does not exist (see line 149). It uses the name of the tensor as filename. There should be a check here to validate filename and if it is not valid then it should generate a random name for this file. For example tensor names can include characters like ':' , ',' , ';' which are invalid in a filename.

jcwchen:
Close it because another similar PR has been merged. Please reopen it if you have other thoughts. Thank you!

PR #2335, opened 2019-09-18
Fix some backend tests 

hariharans29:
1) Minor self-explanatory bug in a Gather test case
2) incorrect input type for a Gemm test case
3) Update the description of Softmax family to indicate the inputs need not be 2D
4) add shape inference test cases for the Softmax family


onnx/defs/math/defs.cc
```diff
@@ -83,29 +82,7 @@ will throw errors.
...
-    schema.TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
-      propagateElemTypeFromInputToOutput(ctx, 0, 0);
-      if (!hasNInputShapes(ctx, 1)) {
-        return;
-      }
-      propagateShapeFromInputToOutput(ctx, 0, 0);
-      const TensorShapeProto& input_shape =
-        ctx.getInputType(0)->tensor_type().shape();
-      int r = input_shape.dim_size();
-      if (r != 2) {
```
hariharans29: This is incorrect as the input need not be 2D. This was a comment in the PR #2281  

Effectively once we know that the output shape is the same as the input shape, there is no point in any other validation. This is based on Rama's comment in #2132 - 

"I feel that validation and generating output shape are distinct sub-problems. We do the best we can for both. Even if we cannot do validation here, we can still generate information..."

So I just reverted the shape inference function altogether

skottmckay: If we're doing the best we can we can still validate the axis so I think that should remain.

gramalingam: Just to clarify the context and the point of my comment: there were cases where the inference function was "returning" while there was still useful steps it could perform (e.g., setting the output rank). My point was that we should DO it. This case is a bit different, since it is a "fail_shape_inference". If the rank is not required to be 2, it seems erroneous to enforce it, so dropping it seems the right thing. For the axis check: as Scott says, there seems nothing wrong with retaining it.

onnx/defs/math/defs.cc
```diff
@@ -54,7 +52,8 @@ case where axis=1, this means the input tensor will be coerced into a 2D tensor
 of dimensions [a_0, a_1 * ... * a_{n-1}], where a_0 is often the batch size.
 In this situation, we must have a_0 = N and a_1 * ... * a_{n-1} = D.
 Each of these dimensions must be matched correctly, or else the operator
-will throw errors.
+will throw errors. The output tensor has the same shape
+and contains the {name} values of the corresponding input.
```
wschin: What does `{name}` mean?

gramalingam: It is replaced by the value of the parameter name, as in two lines below.

hariharans29: I moved this line from the top to the bottom where I thought it would be meaningful. Will fix this.

hariharans29: This line is from above. Please check.

hariharans29: It should be okay to leave as it is. This doc generator is shared between Softmax, Hardmax, LogSoftmax and the name is replaced in the generated doc.

onnx/defs/math/defs.cc
hariharans29: Retained the validation logic. Thanks for clarifying. :)

PR #2337, opened 2019-09-19
Update doc loop op

gramalingam:
Change example in Loop's documentation to clarify things. (Redo of earlier PR https://github.com/onnx/onnx/pull/2144 to incorporate various suggestions.)


docs/Operators.md
```diff
@@ -7510,14 +7510,14 @@ expect(node, inputs=[x], outputs=[y],
...
         %i[INT32, scalar]
-        %keepgoing[BOOL, scalar]
-        %b[INT32, scalar]
+        %keepgoing_in[BOOL, scalar]
+        %b_in[INT32, scalar]
       ) {
-        %my_local = Add(%a, %b)
-        %b_out = Sub(%a, %b)
+        %my_local = Add(%a, %b_in)
+        %b_out = Sub(%a, %b_in)
```
```suggestion
        // The value of b_out is used as the value of b_in in the next iteration. Note that it doesn't affect the value of the b_in passed into this operator.
        %b_out = Sub(%a, %b_in)
```

docs/Operators.md
```diff
@@ -7532,23 +7532,35 @@ expect(node, inputs=[x], outputs=[y],
...
-        for (int i=0; i < max_trip_count && keepgoing; ++i) {
+        /* initialize loop-carried variables and scan-output variables */
+        bool keepgoing_out = keepgoing
+        int b_out = b
+  
+        for (int i=0; i < max_trip_count && keepgoing_out; ++i) {
+          /* Implicitly-defined code: bind actual parameter values
+             to formal parameter variables of loop-body */
+          keepgoing_in = keepgoing_out; 
+          b_in = b_out;
```
```suggestion
          // The variable b_in is passed into the block by value, so the original value of b_in won't be changed.
          b_in = b_out;
```

docs/Operators.md
gramalingam: I added a few more comments to address the two points you mention.

docs/Operators.md
```diff
@@ -7509,15 +7509,15 @@ expect(node, inputs=[x], outputs=[y],
...
+        %keepgoing_in[BOOL, scalar] // incoming loop-termination-condition; not used
+        %b_in[INT32, scalar]        // incoming value of loop-carried-dependency b
       ) {
-        %my_local = Add(%a, %b)
-        %b_out = Sub(%a, %b)
-        %keepgoing_out = Greater(%my_local, %b_out)
-        %user_defined_vals = Add(%b, %b)
-        return %keepgoing_out, %b_out, %user_defined_vals
+        %my_local = Add(%a, %b_in)
+        %b_out = Sub(%a, %b_in) // outgoing value of loop-carried-dependency b
```
wschin: There is no `b` anymore. This statement doesn't explicitly say that `the last value of b_out in one iteration will be the value of b_in in the next iteration`.

docs/Operators.md
```diff
@@ -7509,15 +7509,15 @@ expect(node, inputs=[x], outputs=[y],
...
+        %keepgoing_out = Greater(%my_local, %b_out) // outgoing loop-termination-condition
+        %user_defined_val = Add(%b_in, %b_in) // scan-output value to be accumulated
+        return %keepgoing_out, %b_out, %user_defined_val
```
wschin: Given the `return …` here, it looks like the Loop will terminate at the end of the first iteration.

gramalingam: I think it is more than 1 iteration (in the whole example, you mean?), though it may terminate in 2 or 3 iterations (haven't checked fully). I just reused the existing loop logic, thought it is an artificial one.

docs/Operators.md
gramalingam: That's what the pseudo-code is for, to show how b_out is copied into b_in that assignment statement. Explanatory comments may be more useful outside, in the description. I will see if I can add any more to clarify things there.

docs/Changelog.md
```diff
@@ -11613,29 +11613,49 @@ This version of the operator has been available since version 11 of the default
...
+          user_defined_val = b_in + b_in; // b_in and b_out are different variables
           /* End user-defined code */
+  
+          /* Implicitly defined-code */
+          user_defined_vals[i] = user_defined_val // accumulate scan-output values
         }
-        // my_local = 123; // Can't do this. my_local was defined in the the body
+        // int t = my_local; // Can't do this. my_local is not accessible here.
   
         // These below values are live-out from the loop and therefore accessible
```
wschin: [nit]
```suggestion
        // These below values are bound to the outputs of the loop and therefore accessible
```

PR #2343, opened 2019-09-20
Clarify behavior in ConvTranspose

### LINKED ISSUE #2333: ConvTranspose's output_padding attribute requires clarification
ConvTranspose's output_padding attribute is defined as follows:

> The **zero-padding** added to one side of the output. This is also called adjs/adjustment in some frameworks.

(Emphasis mine.) The most likely interpretation of this is that a positive output_padding would add rows/columns of zeros to one side of the output tensor. This is misleading, as the apparent intention of this attribute is to match PyTorch's definition, which is different. PyTorch instead defines it as follows:

https://pytorch.org/docs/stable/nn.html#convtranspose2d
 
> The padding argument effectively adds dilation * (kernel_size - 1) - padding amount of zero padding to both sizes of the input. This is set so that when a Conv2d and a ConvTranspose2d are initialized with same parameters, they are inverses of each other in regard to the input and output shapes. However, when stride > 1, Conv2d maps multiple input shapes to the same output shape. output_padding is provided to resolve this ambiguity by effectively increasing the calculated output shape on one side. **Note that output_padding is only used to find output shape, but does not actually add zero-padding to output.**

(Emphasis mine.) In other words in PyTorch, output_padding affects the output shape (just like it does in ONNX) but the added rows/columns *are not zero'd* - they participate in the ConvTranspose just like the other elements of the output tensor.

The definition of output_padding in ConvTranspose should therefore probably be clarified to say something closer to how PyTorch defines it - that is, that it participates in shape inference of the output tensor, but doesn't actually zero out those added rows/columns.

(This was discussed earlier in an offline thread with @spandantiwari and @pranavsharma)

wschin:
This PR makes the wording for `output_padding` attribute in ConvTranspose clearer. Since the version number of ConvTranspose in ONNX-1.5 is still 1, this PR doesn't touch ConvTranspose's version number. That is, the ConvTranspose is still in opset 11. Note that since Tensorflow doesn't use `output_padding`, this attribute is mainly for matching Pytorch's behavior.

[Pytorch reference](https://pytorch.org/docs/stable/nn.html#convtranspose2d). [Tensorflow reference](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d_transpose).

This PR should fix #2333.


docs/Changelog.md
```diff
@@ -10762,7 +10762,7 @@ This version of the operator has been available since version 11 of the default
 <dt><tt>kernel_shape</tt> : list of ints</dt>
 <dd>The shape of the convolution kernel. If not present, should be inferred from input W.</dd>
 <dt><tt>output_padding</tt> : list of ints</dt>
-<dd>The zero-padding added to one side of the output. This is also called adjs/adjustment in some frameworks.</dd>
+<dd>Additinal elements added to the side with higher coordinate indexes in the output. Each element in "output_padding" cannot be larger than the stride along the associated axis. By default, this attribute is a zero vector. Note that this attribute doesn't directly affect the computed output values. It only controls the selection of the computed values. This is also called adjs or adjustment in some frameworks.</dd>
```
fdwr: Additional

onnx/defs/nn/defs.cc
```diff
@@ -1299,8 +1299,12 @@ output_shape can also be explicitly specified in which case pads values are auto
         OPTIONAL);
     schema.Attr(
         "output_padding",
-        "The zero-padding added to one side of the output."
-        " This is also called adjs/adjustment in some frameworks.",
+        "Additinal elements added to the side with higher coordinate indexes in the output. "
```
adtsai: Typo - "additional"

docs/Operators.md
```diff
@@ -2948,7 +2948,7 @@ Other versions of this operator: <a href="Changelog.md#ConvTranspose-1">ConvTran
 <dt><tt>kernel_shape</tt> : list of ints</dt>
 <dd>The shape of the convolution kernel. If not present, should be inferred from input W.</dd>
 <dt><tt>output_padding</tt> : list of ints</dt>
-<dd>The zero-padding added to one side of the output. This is also called adjs/adjustment in some frameworks.</dd>
+<dd>Additinal elements added to the side with higher coordinate indexes in the output. Each element in "output_padding" cannot be larger than the stride along the associated axis. By default, this attribute is a zero vector. Note that this attribute doesn't directly affect the computed output values. It only controls the selection of the computed values. This is also called adjs or adjustment in some frameworks.</dd>
```
adtsai: Q: is "selection of the computed values" equivalent to saying that it affects only the shape of the output?

wschin: Yes.

docs/Changelog.md
wschin: Thanks!

onnx/defs/nn/defs.cc
wschin: Will fix it.

docs/Changelog.md
fdwr: How about:
"Additional elements added to the output tensor shape, extending the side with higher coordinates. Each padding value in `output_padding` must be less than or equal to the corresponding stride value. By default, this attribute is a zero vector of size equal to the output tensor rank. Note that this attribute doesn't directly affect the computed output values. It only controls the shape of the computed values. This is also called `adjs` or `adjustment` in some frameworks."

wschin: I will merge your ideas into the description.

PR #2345, opened 2019-09-23
Relax IF's shape inference rule

wschin:
This PR allows If to produce different shapes in `then` and `else` branches. It makes the uses of `If` easier. It will also fix a shape inference problem in SSD objection detection model.


onnx/defs/controlflow/defs.cc
```diff
@@ -244,8 +244,8 @@ void IfInferenceFunction(InferenceContext& ctx) {
           else_output->value_case());
     }
 
-    auto* if_output = ctx.getOutputType(i);
-    *if_output = *then_output;
+		auto* if_output = ctx.getOutputType(i);
```
wschin:
```suggestion
    auto* if_output = ctx.getOutputType(i);
```

onnx/defs/controlflow/defs.cc
```diff
@@ -245,7 +245,7 @@ void IfInferenceFunction(InferenceContext& ctx) {
     }
 
     auto* if_output = ctx.getOutputType(i);
-    *if_output = *then_output;
+		*if_output = *then_output;
```
wschin:
```suggestion
    *if_output = *then_output;
```

onnx/defs/controlflow/defs.cc
```diff
@@ -261,10 +261,16 @@ void IfInferenceFunction(InferenceContext& ctx) {
             else_elem_type);
       }
 
-      // merge the 'else' shape information to check it's consistent and
-      // augment the 'if' output if possible
-      mergeInShapeInfo(
-          else_output->tensor_type(), *if_output->mutable_tensor_type());
+			if (checkShapeCompatibility(
```
hariharans29: nit: some minor formatting issue

docs/Changelog.md
```diff
@@ -11480,6 +11480,46 @@ This version of the operator has been available since version 11 of the default
...
+<dl>
+<dt><tt>cond</tt> : B</dt>
+<dd>Condition for the if</dd>
+</dl>
+
+#### Outputs (1 - &#8734;)
+
+<dl>
+<dt><tt>outputs</tt> (variadic, heterogeneous) : V</dt>
+<dd>Values that are live-out to the enclosing scope. The return values in the `then_branch` and `else_branch` must be of the same same data type. If `then_branch` and `else_branch` produce tensors with different shapes, conflict shapes would be unknown.</dd>
```
kevinch-nv: Typo: `same same data type`. Also the wording of the last sentence is awkward. What is a `conflict shape`?

docs/Operators.md
```diff
@@ -6437,7 +6439,7 @@ This version of the operator has been available since version 1 of the default O
 
 <dl>
 <dt><tt>outputs</tt> (variadic, heterogeneous) : V</dt>
-<dd>Values that are live-out to the enclosing scope. The return values in the `then_branch` and `else_branch` must be of the same shape and same data type.</dd>
+<dd>Values that are live-out to the enclosing scope. The return values in the `then_branch` and `else_branch` must be of the same same data type. If `then_branch` and `else_branch` produce tensors with different shapes, conflict shapes would be unknown.</dd>
```
kevinch-nv: Same typo as above

docs/Changelog.md
hariharans29: Is it better to just be explicit ? 

"The `then_branch` and `else_branch` may produce tensors with different shapes"

onnx/defs/controlflow/defs.cc
```diff
@@ -382,16 +388,17 @@ void LoopInferenceFunction(InferenceContext& ctx) {
...
     OpSchema()
         .SetDoc("If conditional")
         .Input(0, "cond", "Condition for the if", "B")
         .Output(
             0,
             "outputs",
             "Values that are live-out to the enclosing scope. The return values in "
-            "the `then_branch` and `else_branch` must be of the same shape and same "
-            "data type.",
+            "the `then_branch` and `else_branch` must be of the same same data type. "
```
kevinch-nv: Same as above

onnx/test/shape_inference_test.py
```diff
@@ -1667,6 +1702,38 @@ def test_if(self):  # type: () -> None
 
         self._assert_inferred(graph, [make_tensor_value_info('if_output', TensorProto.FLOAT, (1,))])
 
+    def test_if_with_different_shapes_in_then_else_branches(self):  # type: () -> None
+
+        # Create a simple If node where the 'then' subgraph adds to the current value, and the 'else' subgraph
+        # subtracts.
+        # can't use self._make_graph for the subgraphs as that add more inputs for the Reshape operations it inserts.
```
kevinch-nv: Wording is awkward here, perhaps `avoid calling self._make_graph here for subgraph generation as extra reshape nodes will be created`

docs/Changelog.md
wschin: Sure.

wschin: I will add a explicit example.

onnx/defs/controlflow/defs.cc
gramalingam: "same same" to "same"

onnx/defs/controlflow/defs.cc
```diff
@@ -382,16 +388,17 @@ void LoopInferenceFunction(InferenceContext& ctx) {
...
+            "If `then_branch` and `else_branch` produce tensors with different "
+            "shapes, conflict shapes would be unknown.",
```
gramalingam: It is unclear what this ("conflict shapes would be unknown") means to an user interested in creating a model. Suggest changing this to "Corresponding tensors produced by the then-branch and else-branch are not required to have the same or compatible (static) shapes.". We could talk about how shape-inference will handle it, but I don't think that is part of the operator semantics.

onnx/defs/controlflow/defs.cc
```diff
@@ -261,10 +261,16 @@ void IfInferenceFunction(InferenceContext& ctx) {
...
-      // merge the 'else' shape information to check it's consistent and
-      // augment the 'if' output if possible
-      mergeInShapeInfo(
-          else_output->tensor_type(), *if_output->mutable_tensor_type());
+			if (checkShapeCompatibility(
+              then_output->tensor_type().shape(),
+              else_output->tensor_type().shape())) {
+        // merge the 'else' shape information to check it's consistent and
+        // augment the 'if' output if possible
+        mergeInShapeInfo(
```
gramalingam: There are a few issues with this logic. What we need to do is set the resulting shape as the "union" of the shapes produced by the THEN and ELSE branch. I recommend using this function: https://github.com/onnx/onnx/blob/c5af774aa950f6a5cf1a7b5f37d014fb2fbd82bc/onnx/defs/sequence/defs.cc#L21

onnx/defs/controlflow/old.cc
```diff
@@ -989,4 +989,119 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        " but subgraphs produce ",
+        num_then_outputs);
+  }
+
+  for (size_t i = 0, end = then_output_types.size(); i < end; ++i) {
+    auto then_output = then_output_types[i];
+    auto else_output = else_output_types[i];
+
+    if (then_output->value_case() != else_output->value_case()) {
+      fail_type_inference(
```
yuslepukhin: Should we output separate message if  any of these are not set (NOT_SET) value?

onnx/defs/controlflow/old.cc
```diff
@@ -989,4 +989,119 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          " then=",
+          then_output->value_case(),
+          " else=",
+          else_output->value_case());
+    }
+
+    auto* if_output = ctx.getOutputType(i);
+    *if_output = *then_output;
+
+    if (then_output->has_tensor_type()) {
```
yuslepukhin: Suggest to avoide has_type() function. Simply type() would return a default instance which would have value_case() NOT_SET

onnx/defs/shape_inference.h
```diff
@@ -613,6 +613,48 @@ inline void mergeInShapeInfo(
     mergeInShapeInfo(source.shape(), target);
 }
 
+// If two dimensions have different numerical values, return
+// false. Otherwise, true is returned.
+inline bool checkDimensionCompatibility(
+    const TensorShapeProto_Dimension& source_dim,
+    const TensorShapeProto_Dimension& target_dim) {
+  if (source_dim.has_dim_value() && target_dim.has_dim_value()) {
```
yuslepukhin: Use switch statement instead of has_*() functions which are absent in proto3. You want this to run in ML.NET.

gramalingam: I believe it is not a problem with "one of" fields … the has_* functions are generated even in proto3 for "one of", if I am not mistaken

gramalingam: See https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#oneof-numeric-fields and https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#oneof-embedded-message-fields … both proto2 and proto3 support "has_*" for one of fields. It is only for primitive-type fields that are not "one of" we have the proto2/proto3 issue (IIUC).

onnx/defs/controlflow/defs.cc
gramalingam: Suggestion: "If corresponding outputs from the then-branch and the else-branch have static shapes S1 and S2, then the shape of the corresponding output variable of the if-node (if present) must be compatible with both S1 and S2 as it represents the union of both possible shapes." If we want to add a more detailed explanation, suggest adding to the operator description or IR.md (for the notion of "compatible").

onnx/defs/controlflow/old.cc
hariharans29: I think @wschin just copied this over to old.cc. This is the previous shape inference logic. This is now legacy. I think it is safer not to "enhance" this further unless there is a bug to be patched. We could be introducing new regressions in existing models using If-1.

wschin: Yes!

onnx/defs/controlflow/defs.cc
wschin: Added. Thank you!

onnx/defs/controlflow/defs.cc
```diff
@@ -261,10 +261,16 @@ void IfInferenceFunction(InferenceContext& ctx) {
...
-      // merge the 'else' shape information to check it's consistent and
-      // augment the 'if' output if possible
-      mergeInShapeInfo(
-          else_output->tensor_type(), *if_output->mutable_tensor_type());
+      if (checkShapeCompatibility(
+              then_output->tensor_type().shape(),
+              else_output->tensor_type().shape())) {
+        // merge the 'else' shape information to check it's consistent and
+        // augment the 'if' output if possible
+        UnionShapeInfo(
```
gramalingam: Just the "UnionShapeInfo(…)" should be sufficient, I think. No need for if-condition or the else-branch.

onnx/defs/controlflow/defs.cc
```diff
@@ -382,16 +388,26 @@ void LoopInferenceFunction(InferenceContext& ctx) {
...
+            "The `then_branch` and `else_branch` may produce tensors with the same "
+            "element type and different shapes. "
+            "If corresponding outputs from the then-branch and the else-branch have "
+            "static shapes S1 and S2, then the shape of the corresponding output "
+            "variable of the if-node (if present) must be compatible with both S1 "
+            "and S2 as it represents the union of both possible shapes."
+            "For example, if in a model file, the the first "
+            "output of `then_branch` is typed float tensor with shape [2] and the "
+            "first output of `else_branch` is another float tensor with shape [3], "
+            "If's first output should be a float tensor with an unset shape field. "
```
gramalingam: Actually, it is valid for the first output to (a) Have no shape set, or (b) A shape of rank 1 set, with neither dim-value nor dim-param set, or (c) A shape of rank 1 with a unique dim-param.

onnx/defs/controlflow/defs.cc
```diff
@@ -382,16 +388,26 @@ void LoopInferenceFunction(InferenceContext& ctx) {
...
+            "In contrast, shapes [2, 3] and [2, 4] are not compatible.",
```
gramalingam: How about changing it to "In contrast, the first output cannot have the shape [2] since [2] and [3] are not compatible."

onnx/defs/controlflow/defs.cc
wschin: You're right.

onnx/defs/controlflow/defs.cc
wschin: Sure.

onnx/defs/controlflow/defs.cc
wschin: Ok. I have replaced part of my description with your idea.

onnx/test/shape_inference_test.py
wschin: I feel the current description is more precise. From this description, adding Reshape itself is not a problem, but adding input required by Reshape causes the problem.

onnx/defs/controlflow/defs.cc
wschin: Sure.

onnx/defs/controlflow/old.cc
wschin: I feel it's a bigger problem in the whole codebase. If you agree, I'd like to open another issue for addressing it globally.

onnx/defs/controlflow/defs.cc
wschin: Using UnionShapeInfo produces wrong shapes in my test. Can I roll back to my original function?

wschin: Fixed.

docs/Operators.md
```diff
@@ -6437,7 +6439,7 @@ This version of the operator has been available since version 1 of the default O
 
 <dl>
 <dt><tt>outputs</tt> (variadic, heterogeneous) : V</dt>
-<dd>Values that are live-out to the enclosing scope. The return values in the `then_branch` and `else_branch` must be of the same shape and same data type.</dd>
+<dd>Values that are live-out to the enclosing scope. The return values in the `then_branch` and `else_branch` must be of the same data type. The `then_branch` and `else_branch` may produce tensors with the same element type and different shapes. If corresponding outputs from the then-branch and the else-branch have static shapes S1 and S2, then the shape of the corresponding output variable of the if-node (if present) must be compatible with both S1 and S2 as it represents the union of both possible shapes.For example, if in a model file, the the first output of `then_branch` is typed float tensor with shape [2] and the first output of `else_branch` is another float tensor with shape [3], If's first output should have (a) no type set, (b) a shape of rank 1 with neither `dim_value` nor `dim_param` set, or (c) a shape of rank 1 with a unique `dim_param`. In contrast, the first output cannot have the shape [2] since [2] and [3] are not compatible.</dd>
```
gramalingam: "(a) no type set" => "(a) no shape set"

wschin: Nice catch! Thank you.

PR #2347, opened 2019-09-24
Fix spec and shape inference for Unsqueeze op

hariharans29:
1) Unsqueeze op is unclear on how it deals with values in `axes` and what are the acceptable values for `axes`

2) Fix shape inference logic. In current state, it cannot handle negative axes properly (the way it is intended to)


hariharans29:
CC: @wschin @gramalingam

onnx/test/shape_inference_test.py
```diff
@@ -455,13 +455,27 @@ def test_squeeze(self):  # type: () -> None
...
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (1, 1, 3, 1, 2, 1))])
 
+    def test_unsqueeze_2(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, (3, 4, 5))],
+            [make_node('Unsqueeze', 'x', 'y', axes=[0, 4])],
+            [])
+        self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (1, 3, 4, 5, 1))])
+
+    def test_unsqueeze_negative_axes(self):  # type: () -> None
```
hariharans29: won't work with existing logic.

onnx/defs/tensor/defs.cc
```diff
@@ -1301,11 +1301,17 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-  Given a tensor such that tensor with shape [3, 4, 5], then
-  Unsqueeze(tensor, axes=[0, 4]) has shape [1, 3, 4, 5, 1]
+Insert single-dimensional entries to the shape of an input tensor (`data`).
+Takes one required argument `axes` - which contains a list of dimension indices and this operator will insert a dimension of value `1` into the corresponding index of the output tensor (`expanded`).
+
+For example:
+  Given an input tensor (`data`) of shape [3, 4, 5], then
+  Unsqueeze(data, axes=[0, 4]) outputs a tensor (`expanded`) containing same data as `data` but with shape [1, 3, 4, 5, 1].
+
+The attribute `axes` should not contain any duplicate entries. It is an error if it contains duplicates.
```
spandantiwari: +1 for mentioning about duplicate entries.

onnx/defs/tensor/defs.cc
```diff
@@ -1301,11 +1301,17 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-Takes one required argument `axes`, a list of dimensions that will be inserted.
-Dimension indices in `axes` are as seen in the output tensor. For example:
-  Given a tensor such that tensor with shape [3, 4, 5], then
-  Unsqueeze(tensor, axes=[0, 4]) has shape [1, 3, 4, 5, 1]
+Insert single-dimensional entries to the shape of an input tensor (`data`).
+Takes one required argument `axes` - which contains a list of dimension indices and this operator will insert a dimension of value `1` into the corresponding index of the output tensor (`expanded`).
+
+For example:
+  Given an input tensor (`data`) of shape [3, 4, 5], then
+  Unsqueeze(data, axes=[0, 4]) outputs a tensor (`expanded`) containing same data as `data` but with shape [1, 3, 4, 5, 1].
```
spandantiwari: As discussed, we should consider adding a note clarifying that the order of axes specified in the `axes` attribute does not matter, and that they should be sorted before being applied.

wschin:
Do we have tests for unsorted cases?

onnx/defs/tensor/defs.cc
hariharans29: I added this line - 

The order of values in `axes` does not matter and can come in any order.

I think this is good for a high level spec. Details like the operator will sort them after correcting negative values are superfluous, I feel. What do you think ?

hariharans29:
> Do we have tests for unsorted cases?

I will add shape inference test for now.

hariharans29:
@spandantiwari @wschin - added backend test for unsorted axes too

onnx/backend/test/case/node/unsqueeze.py
```diff
@@ -61,6 +61,23 @@ def export_unsqueeze_three_axes():  # type: () -> None
...
 
+    @staticmethod
+    def export_unsqueeze_unsorted_axes():  # type: () -> None
+        x = np.random.randn(3, 4, 5).astype(np.float32)
+
+        node = onnx.helper.make_node(
+            'Unsqueeze',
+            inputs=['x'],
+            outputs=['y'],
+            axes=[5, 4, 2],
```
hariharans29: axes values in jumbled order

PR #2348, opened 2019-09-24
Bump NMS version for avoiding regression in existing models

wschin:
In opset 10, NMS doesn't produce shape. With the newly added shape inference function, some existing models can be broken due to the interaction between NSM and IF nodes (IF's then branch and else branch produce different shapes). Thus, I feel it'd be better to bump the version of NMS so that old code can still work with old models.

Because shape inference produces side information for a model, if we consider side information a part of the spec, we also need to bump operator's version each time we add a shape inference function.

### DIFF
--- a/onnx/defs/object_detection/old.cc
+++ b/onnx/defs/object_detection/old.cc
@@ -61,6 +61,13 @@ ONNX_OPERATOR_SET_SCHEMA(
             "1 - the box data is supplied as [x_center, y_center, width, height]. Mostly used for Pytorch models.",
             AttributeProto::INT,
             static_cast<int64_t>(0))
-        .SetDoc(NonMaxSuppression_doc));
+        .SetDoc(NonMaxSuppression_doc)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          auto selected_indices_type =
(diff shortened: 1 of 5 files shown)


onnx/defs/object_detection/old.cc
```diff
@@ -0,0 +1,66 @@
...
+            "tensor(int64)")
+        .Attr(
+            "center_point_box",
+            "Integer indicate the format of the box data. The default is 0. "
+            "0 - the box data is supplied as [y1, x1, y2, x2] where (y1, x1) and (y2, x2) are the coordinates of any diagonal pair of box corners "
+            "and the coordinates can be provided as normalized (i.e., lying in the interval [0, 1]) or absolute. Mostly used for TF models. "
+            "1 - the box data is supplied as [x_center, y_center, width, height]. Mostly used for Pytorch models.",
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
+        .SetDoc(NonMaxSuppression_doc));
```
hariharans29: Shouldn't this have TypeAndShapeInference function as before ? Please see #2269 for how the function looked before the change. This PR effectively removes type inferencing which was done previously and may lead to new regressions.

wschin: You are right. Thanks.

PR #2351, opened 2019-09-24
Cherry pick final fixes

wschin:
As title. Those fixes are the final expected efforts to ONNX-1.6.0.


prasanthpul:
Is the table in https://github.com/onnx/onnx/blob/master/docs/Versioning.md updated?

kevinch-nv:
@prasanthpul @houseroad I'll make one final commit after this one bumping the Version number in both those places, and that commit will be the one that is tagged as v1.6.0. I'll then cherry pick the version number update back into master once everything is done.

PR #2355, opened 2019-09-25
Improve precision of shape analysis for OneHot nodes

wintersteiger:
I'm not sure whether this is strictly necessary, but it was helpful to get the shape of onehot outputs inferred at least for simple cases.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -1926,6 +1926,8 @@ ONNX_OPERATOR_SET_SCHEMA(
           // and 1 element vector for now. In future when version update for
           // this op is done we should only allow scalar or chage the spec to
           // allow both.
+          int64_t depth = 0;
+          std::string depth_name = "";
           if (hasInputShape(ctx, 1)) {
             auto& depth_shape = getInputShape(ctx, 1);
             if (depth_shape.dim_size() != 0 && depth_shape.dim_size() != 1) {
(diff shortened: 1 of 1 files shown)


onnx/defs/tensor/defs.cc
```diff
@@ -1938,6 +1940,17 @@ ONNX_OPERATOR_SET_SCHEMA(
               fail_type_inference(
                   "Input 'depth' must have exactly one element.");
             }
+            const TensorProto* depth_initializer = ctx.getInputData(1);
+            if (depth_initializer != nullptr) {
+              if (depth_initializer->has_raw_data()) {
```
skottmckay: Should this also check int64_data and int32_data if the value isn't specified in raw_data?

onnx/defs/tensor/defs.cc
```diff
@@ -1979,6 +1992,11 @@ ONNX_OPERATOR_SET_SCHEMA(
                 } else if (indices_shape.dim(i).has_dim_param()) {
                   dim->set_dim_param(indices_shape.dim(i).dim_param());
                 }
+              } else if (i == axis) {
+                if (depth_name != "")
+                  dim->set_dim_param(depth_name);
+                else
+                  dim->set_dim_value(depth);
```
skottmckay: This looks off in two ways.

If we were able to read an actual 'depth' value shouldn't that be preferred?

If we didn't read a value for 'depth' from an initializer, I would have expected we shouldn't call set_dim_value to the default of 0. I'm assuming 'depth' doesn't necessarily come from an initializer (so the getInputData call may return nullptr) so the actual value may not be known until runtime.

gramalingam: Agree with Scott. Furthermore, we should not reuse a tensor's name as a dimension-variable (they belong to different namespaces and this could introduce an unintended clash if the model used the same name for another dimension-variable). We should leave the dim with neither param nor value set if we don't know the depth value.

PR #2356, opened 2019-09-25
Fix Small Error In DynamicQuantizeLinear Desc

CodesWithWolves:
The DynamicQuantizeLinear description was modified to reflect the computation that is actually performed. Namely, the zero-point should be the difference between the minimum value of the data and the minimum of the quantization range scaled using 'y_scale'; and not scaled using the width of the quantization range. This can also be seen from the example where it is done correctly.

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -4073,9 +4073,9 @@ expect(node, inputs=[x], outputs=[y],
   ```
   Zero point is calculated as:
   ```
-  intermediate_zero_point = (qmin - min(x))/(qmax - qmin)
+  intermediate_zero_point = (qmin - min(x))/y_scale
   y_zero_point = cast(round(saturate(itermediate_zero_point)))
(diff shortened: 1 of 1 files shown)


postrational:
Please don't modify the Operators.md file directly. This is a generated file.
You should modify the source file `onnx/defs/quantization/defs.cc` instead.

justinchuby:
@jcwchen Do you think this is still needed?

jcwchen (commented):
> @jcwchen Do you think this is still needed?

Seems good to have -- The latest sentence is:
```
intermediate_zero_point = qmin - min(x)/y_scale
```
with
```
where qmax and qmin are max and min values for quantization range .i.e [0, 255] in case of uint8
```
The formula does not use qmax anymore and probably we should remove qmax related wordings to prevent confusion.

justinchuby:
I think the doc is clear

PR #2366, opened 2019-09-27
Create StaticShape4DynShapeOpsProposal.md

goldgeisser:
This proposal is being scheduled to discussed during the next onnx/Operators SIG meeting.
I am creating the first draft of the proposal so the members of the SIG can get familiar with this and comment on it.


dashesy:
Shouldn't this be part of an optimization? Export everything dynamic, then optimize away shape nodes to static/const values.

PR #2367, opened 2019-09-27
Add type check for node inputs

RandySheriffH:
This PR extends the model checker to 
(a) Call shape-inference to do the extra-checking performed by the type-and-shape-inference methods of ops, as well as
(b) Check that the typing constraints specified by the op schema are satisfied, and
(c) Infer output types of nodes from the typing constraints specified by the op schema (if not specified or inferred otherwise)


onnx/shape_inference/implementation.cc
```diff
@@ -214,6 +214,7 @@ static void InferShapesImpl(
 	}
 
     try {
+      schema->Verify(n, &valueTypesByName);
```
gramalingam: Can we separate out the basic-verification done by Verify previously from the new checks you added into separate functions? I believe that the idea was that in TypeAndShapeInferenceFunction one could assume that the basic-verification has happened successfully … so it doesn't need to check if a required-attribute is missing, etc. The new checks you added looks like it has to happen after the TypeAndShapeInferenceFunction is called. So, we may need to split it.

RandySheriffH: Good idea - like this?

onnx/defs/schema.cc
```diff
@@ -94,7 +94,69 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
...
 }
 
-void OpSchema::Verify(const NodeProto& node, std::unordered_map<std::string, TypeProto*>* type_map) const {
+void OpSchema::CheckInputOutputType(const NodeProto& node, std::unordered_map<std::string, TypeProto*>* type_map) const {
+
+  std::unordered_map<std::string, DataType> type_constraints;
+  for (int in_idx = 0; in_idx < node.input_size() && in_idx < static_cast<int>(inputs_.size()); ++in_idx) {
+    const auto& all_supported_types = inputs_[in_idx].GetTypes();
+    const auto& type_iter = type_map->find(node.input(in_idx));
+    if (type_iter != type_map->end()) {
```
gramalingam: I think we should fail_check if the type doesn't exist. (Same for outputs.)

onnx/shape_inference/implementation.cc
gramalingam: Yes, looks good, thanks! 

Can we also extend this checker:https://github.com/onnx/onnx/blob/c963586d0f8dd5740777b2fd06f04ec60816de9f/onnx/checker.py#L87 

Extend it to do the basic checking (as currently) and then call shape-inference for the type-inference and checking? May be it can be optionally invoked (controlled by an optional parameter) so that we preserve the existing behavior if invoked as currently?

onnx/defs/schema.cc
RandySheriffH: Now using inference context as suggested.
The context handles the nullptr situation by setting corresponding type as nullptr.

onnx/shape_inference/implementation.cc
RandySheriffH: CR updated according with new checking routine, let's see.

onnx/defs/schema.cc
```diff
@@ -94,6 +94,71 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
...
+    }
+  }//for inputs
+  //check all output types
+  for (size_t out_idx = 0; out_idx < ctx.getNumOutputs() && out_idx < outputs_.size(); ++out_idx) {
+    const auto& param      = outputs_[out_idx];
+    const auto& type_str   = param.GetTypeStr();
+    const auto& param_type = ctx.getOutputType(out_idx);
+    const auto& all_types  = param.GetTypes();
+    bool output_type_found = true;
+    //infer type if necessary
```
gramalingam: I think it would be better to do:
```C++
if (param_type->value_case() == VALUE_NOT_SET) {
   if (all_types.size() == 1 || type_constraints.find(type_str) != type_constraints.end()) {
      *param_type = Utils::DataTypeUtils::ToTypeProto(all_types.size() == 1 ? *all_types.begin() : type_constraints[type_str]);
   }
   continue;
}
```

gramalingam: It helps handle other type-cases (ONNXML has more types than ONNX: ONNX has only tensor-type and sequence-type).

RandySheriffH: Good suggestion, thanks!

RandySheriffH: Done.

onnx/defs/schema.cc
```diff
@@ -94,6 +94,68 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
...
+    if (nullptr == param_type || param_type->value_case() == TypeProto::VALUE_NOT_SET) {
+      if (all_types.size() == 1 || type_constraints.find(type_str) != type_constraints.end()) {
+        const auto& input_type_proto = Utils::DataTypeUtils::ToTypeProto(all_types.size() == 1 ? *all_types.begin() : type_constraints[type_str]);
+        if (input_type_proto.has_tensor_type()) {
+          param_type->mutable_tensor_type()->set_elem_type(input_type_proto.tensor_type().elem_type());
+        } else if (input_type_proto.has_sequence_type()) {
+          param_type->mutable_sequence_type()->mutable_elem_type()->mutable_tensor_type()->set_elem_type(input_type_proto.sequence_type().elem_type().tensor_type().elem_type());
+        } else {
+          output_type_found = false;
+        }
```
gramalingam: Why not replace the above lines with "*param_type = input_type_proto"? I think that is simpler and more general too.

onnx/defs/schema.cc
```diff
@@ -94,6 +94,68 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
...
+  }//for inputs
+  //check all output types
+  for (size_t out_idx = 0; out_idx < ctx.getNumOutputs() && out_idx < outputs_.size(); ++out_idx) {
+    const auto& param      = outputs_[out_idx];
+    const auto& type_str   = param.GetTypeStr();
+    const auto& param_type = ctx.getOutputType(out_idx);
+    const auto& all_types  = param.GetTypes();
+    bool output_type_found = true;
+    //infer type if necessary
+    if (nullptr == param_type || param_type->value_case() == TypeProto::VALUE_NOT_SET) {
```
gramalingam: I don't think we need to handle the "nullptr == param_type" case. The branch-statements assume it is non-null anyway.

RandySheriffH: No - will remove soon.

onnx/defs/schema.cc
RandySheriffH: Good idea, will do.

onnx/defs/schema.cc
```diff
@@ -94,26 +94,39 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
...
+  std::stringstream ss;
+  for (const auto& dt: dt_set) {
+    ss << dt << ",";
+  }
+  return ss.str();
+}
+
 void OpSchema::CheckInputOutputType(struct InferenceContext& ctx) const {
-  std::unordered_map<std::string, DataType> type_constraints;
+  std::unordered_map<std::string, std::string> type_constraints;
```
gramalingam: DataType and the related utils are used to use a pointer (to a string) guaranteed to be unique for a type so that pointer-comparison is sufficient to check for type-equality. Is there a need to use string instead of DataType here?

gramalingam:
Some of the shape-inference test-cases seem to end with a segmentation-fault in the CI.

onnx/defs/schema.cc
RandySheriffH: No - it is just for debugging failed cases, will go back with pointers shortly.

RandySheriffH: done.

onnx/defs/schema.cc
```diff
@@ -94,6 +94,63 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
...
+    if (param_type->value_case() == TypeProto::VALUE_NOT_SET) {
+      if (all_types.size() == 1) {
+        *param_type = Utils::DataTypeUtils::ToTypeProto(*all_types.begin());
+      } else if (type_constraints.find(type_str) != type_constraints.end()) {
+        auto data_type = Utils::DataTypeUtils::ToType(type_constraints[type_str]);
+        *param_type = Utils::DataTypeUtils::ToTypeProto(data_type);
+      } else {
+        output_type_found = false;
+      } 
+    }
```
gramalingam: The logic looks correct. But I think it can be simplified a bit since if we take the THEN branch above (VALUE_NOT_SET), we set the param_type (to a correct type) and so we don't really need to do the checks below. So, I think I would just make the checking logic below the ELSE part. Does that make sense? The THEN part does inference, and the ELSE part checks correctness of pre-specified type.

RandySheriffH: Yes that's sensible approach - the reason I am not making if an "ELSE" is because if sometime in the future we place some new logic there inferring the type (apart from using those we already have like allowed types and types constraints), we might well wish the inferred type get scrutinized.

PR #2380, opened 2019-10-01
Add reference impl for sequence ops

BowenBao:


onnx/backend/test/case/model/sequence.py
```diff
@@ -13,6 +13,64 @@
...
+    return sequence
+
+
+def SequenceLengthImpl(sequence):
+    # type: (List[np.ndarray]) -> np.int64
+    return np.int64(len(sequence))
+
+
+def SplitToSequenceImpl(tensor, split=None, axis=0, keepdims=1):
+    # type: (np.ndarray, Optional[Union[int, List[int]]], int, int) -> List[np.ndarray]
```
wschin:
```suggestion
    # type: (np.ndarray, Optional[Union[int, List[int]]], int, int) -> List[np.ndarray]
```
Can we use numpy type such as `np.int64` as the `int` here?

onnx/backend/test/case/model/sequence.py
```diff
@@ -13,6 +13,64 @@
...
+        split = 1
+        split_indices = [i * split + 1 for i in range(dim_size) if i * split + 1 < dim_size]
+        if not keepdims:
+            results = np.array_split(tensor, split_indices, axis)
+            return [np.squeeze(res, axis) for res in results]
+    if np.isscalar(split):
+        split_indices = [i * split + 1 for i in range(dim_size) if i * split + 1 < dim_size]  # type: ignore
+    else:
+        split_indices = np.cumsum(split) + 1
+    return np.array_split(tensor, split_indices, axis)  # type: ignore
```
wschin: Why do you ignore the type?

onnx/backend/test/case/model/sequence.py
```diff
@@ -13,6 +13,64 @@
 from typing import List, Optional, Text, Union
 
 
+def SequenceEmptyImpl():  # type: () -> List[Optional[np.ndarray]]
```
wschin: Can we reuse those implementations to create examples in `Operators.md`?

onnx/backend/test/case/model/sequence.py
BowenBao: Everywhere else also uses int. I'd prefer using int here as well. In addition as below, it seems some numpy methods don't work well with mypy.

onnx/backend/test/case/model/sequence.py
BowenBao: some numpy methods don't work well with mypy

onnx/backend/test/case/model/sequence.py
BowenBao: Yes. But to create examples in Operators.md, we need to update the proto to enable Sequence Tensor Proto (in order to save test data in sequence format). That should be a separate PR of its own.

PR #2381, opened 2019-10-01
Fix Unique unit test

skottmckay:
Expected data for Y was incorrectly sorted.


skottmckay:
> Update doc and pass CI please

Done

Issue #2239, opened 2019-08-14
Experimental operator debug spew to std::cerr


fdwr:
When an older experimental operator is encountered, `check_node` in `onnx/checker.cc` writes directly to `std::cerr` (rather than to some more proper logging mechanism) which shows up as direct spew to the user and looks rather unprofessional. For console apps, when loading a model, you see a long paragraph of messages. For a GUI app, this message isn't even seen anyway and isn't useful. Note the operators were already removed in a previous ONNX release, and thus the transitional warning isn't so important.

In the context of ONNX runtime, which continues to support these experimental operators for backwards compatibility of older models, the diagnostic message adds no value and is confusing because they are still supported. If/when the shape inference logic is removed from core ONNX, ONNX runtime can adopt the schema and shape inference functions into itself for compat.

## A list of older models affected:
coreml_MNIST.onnx
winmlperf_coreml_FNS-Candy
coreml_VGG16_ImageNet.onnx
cs_generalfeaturizer.onnx
keras2coreml_LSTM_ImageNet.onnx
onnxzoo_winmlperf_tiny_yolov2.onnx
winmlperf_coreml_Inceptionv3
winmlperf_coreml_MobileNet
winmlperf_coreml_SqueezeNet

## Operators affected:
ImageScaler, Affine, Crop

## Desired action:
Just remove the message.

Alternately: Support a CMakeList define, and #ifdef the block.

## Pertinent Code

    void check_node(
        const NodeProto& node,
        const CheckerContext& ctx,
        const LexicalScopeContext& lex_ctx
        )
    {
    ...
        if (experimental_ops.count(node.op_type()))
        {
            std::cerr << "Warning: " << node.op_type() << " was a removed "
                      << "experimental op. In the future, we may directly "
                      << "reject this operator. Please update your model as soon "
                      << "as possible." << std::endl;

fdwr:
FYI @linkerzhang, @faxu.

linkerzhang:
this error message could be removed given exp ops has been removed since 1.5. Let's remove it in 1.6.

askhade:
It was decided to stop the checker if it encounters a model with experimental op : PR https://github.com/onnx/onnx/pull/2783 
FYI: @jcwchen

askhade:
Lets fix this in onnx 1.10

askhade:
Closing this as a duplicate of #3696.

PR #2384, opened 2019-10-02
Update Versioning.md to include 1.6 opset value and add deprecated operators table

fdwr:
The 1.6 release missed adding a row for opset 11 to the table.

Many experimental operators were removed from ONNX, including dozens used in client models for published applications, which now cause warnings to be written to the console in `check_node`. e.g. "*Warning: Affine was a removed experimental op. In the future, we may directly reject this operator...*". These warnings provide no guidance on what to do though, and they're confusing to the user given frameworks like ONNX Runtime continue to support these operators anyway for backwards compatibility. Since this is related to versioning, I'm proposing to document the guidance here (note we can't it in Operators.md because these operators don't exist anymore, and we can't include it in ChangeLog.md either because all records of these operators were removed).

Part of #2239.


fdwr:
@linkerzhang Here is the documentation aspect of the issue.

docs/Versioning.md
```diff
@@ -127,6 +127,24 @@ Notes:
 - Values that are new or updated from a previous OpSet version are in **bold**.
 - \*: the operator is deprecated.
 
+### Operator Deprecations
+
+The following deprecated operators should be removed from models, substituted with newer operators, or decomposed:
```
prasanthpul: I think these are all experimental operators, right? You should clarify that these are experimental ops that have been deprecated. The concept of experimental ops is also deprecated now.

Non-experimental operators will remain in documentation even after deprecation. Example is Scatter op which was deprecated in opset 11

prasanthpul:
can you fix the conflict?

docs/Versioning.md
```diff
@@ -127,6 +127,24 @@ Notes:
 - Values that are new or updated from a previous OpSet version are in **bold**.
 - \*: the operator is deprecated.
 
+### Operator Deprecations
```
prasanthpul: Consider adding this info to https://github.com/onnx/onnx/blob/5ca0a09e6846fe70c9dda8b5034f4a1454c6d130/docs/ManagingExperimentalOps.md and linking to it instead. Experimental ops do not have any versioning and now the concept itself is deprecated, so it's odd to list them in the mainline documentation.

fdwr:
Fixed PR here, with correct CLA-associated commits: https://github.com/onnx/onnx/pull/2555

PR #2389, opened 2019-10-04
using expm1

autoih:
using superior expm1 to calculate `exp(x) - 1`


autoih:
Hi @prasanthpul, can you help review the PR when you are convenient?

autoih:
Thanks @prasanthpul. I think it may be good if we have `expm1p` for the calculation `exp(x) - 1` to provide greater precision, according the [ref](https://en.wikipedia.org/wiki/Loss_of_significance), It only makes sense to use `expm1` when the computation requires like `exp(x) - 1`, and this is the exact cases of this PR.

autoih:
Thanks @prasanthpul, would you please guide me how to fix the error raised by travis-ci? I am not sure why the build has been terminated. Thank you!

postrational:
It seems the `Operators.md` will have to be regenerated or conflicts resolved otherwise.

PR #2390, opened 2019-10-07
Fix bug in function body verifier

gramalingam:
The partial-checker for function-body has a bug where it uses the formal-parameter's name instead of actual-parameter's name. This PR fixes this bug.


onnx/test/cpp/function_verify_test.cc
```diff
@@ -13,33 +13,29 @@ using namespace checker;
 using TENSOR_TYPES_MAP =
     std::unordered_map<std::string, std::vector<std::string>>;
 
+// Unfortunately, the different data-structures use different int types to
+// represent a parameter-index.
+using parameter_index = unsigned int;
```
ebarsoum: Why not use size_t and cast based on that?

onnx/test/cpp/function_verify_test.cc
```diff
@@ -49,51 +45,53 @@ void VerifyTypeConstraint(
...
-        if (allowed_types.find(type) == allowed_types.end()) {
-          fail_check(
-              "Input type " + type + " defined in " + schema->Name() +
-              "'s function body is not allowed in node " + op_type);
+    // Check inputs of node:
+    parameter_index num_formal_inputs =
+        static_cast<parameter_index>(schema->inputs().size());
+    parameter_index num_actual_inputs = node.input_size();
+    // The above may not be same due to optional inputs
+    auto num_checked_inputs = std::min(num_formal_inputs, num_actual_inputs);
```
ebarsoum: Is there a case where actual inputs > formal input?

gramalingam: No. I can add a check and throw an error in this case, if that's what you mean.

ebarsoum: What I mean why not just use actual inputs?

gramalingam: Actually: sorry, I was wrong. This is possible with variadic parameters. I will modify the logic to do the right thing.

Issue #2396, opened 2019-10-11
Please update SequenceToSplit spec with default value of 'split'.


pranavsharma:
The input 'split' is mentioned as optional in the spec without specifying a default value. I chatted with @BowenBao and he mentioned that it should be 1. Please update the spec accordingly. Thanks!

PR #2397, opened 2019-10-11
Clarify SplitToSequence default value for Input split

BowenBao:
Fixing #2396. This logic was missing in the spec, and was only present in shape inferencing/test cases.


pranavsharma:
Doesn't this require a version upgrade?

gramalingam:
It looks like this clarification is still appropriate and worth making. Not sure if it is easier to create a new PR or update this.

PR #2402, opened 2019-10-18
Add shape existence check in GatherElements shape inference logic

hariharans29:
The shape inference logic basically propagated the shape of the `indices` input to the output shape. This logic is correct but we need to add a guard to make sure the `indices` input has a shape associated with it.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -1227,7 +1227,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain indices to integer types")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           propagateElemTypeFromInputToOutput(ctx, 0, 0);
-          // propogate indices' shape to output if it exists
+          // propagate indices' shape to output if it exists
           if (hasInputShape(ctx, 1)) {
             propagateShapeFromInputToOutput(ctx, 1, 0);
           }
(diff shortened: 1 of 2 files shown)


onnx/test/shape_inference_test.py
```diff
@@ -2707,6 +2707,22 @@ def test_pad(self):  # type: () -> None
...
+            [])
+        self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (2,))])
+
+    def test_gatherelements_noshape(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, (6,)),
+             ('indices', TensorProto.INT64, ())],
+            [make_node('GatherElements', ['x', 'indices'], ['y'])],
+            [])
+        self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, ())])
```
skottmckay: Should this be
('indices', TensorProto.INT64, None)],
and
make_tensor_value_info('y', TensorProto.FLOAT, None)?

onnx/defs/tensor/defs.cc
```diff
@@ -1227,7 +1227,10 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain indices to integer types")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           propagateElemTypeFromInputToOutput(ctx, 0, 0);
-          propagateShapeFromInputToOutput(ctx, 1, 0);
+          // propogate indices' shape to output if it exists
```
skottmckay: nit: propagate

onnx/test/shape_inference_test.py
hariharans29: You are right. Fixed, thanks.

onnx/defs/tensor/defs.cc
hariharans29: Fixed, thanks

PR #2408, opened 2019-10-22
Add TensorProtos to represent a list of Tensors

yinghai:
Sometimes we want to serialized/deserialize a list of tensors for testing purposes (inputs/outputs). A repeated TensorProto message makes this convenient.

### DIFF
--- a/onnx/onnx.in.proto
+++ b/onnx/onnx.in.proto
@@ -456,6 +456,11 @@ message TensorProto {
   repeated uint64 uint64_data = 11 [packed = true];
 }
 
+// TensorProtos stores multiple TensorProto objects in one single proto.
+message TensorProtos {
+  repeated TensorProto protos = 1;
+}
+
 // A serialized sparse-tensor value
 message SparseTensorProto {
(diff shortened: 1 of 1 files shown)


onnx/onnx.in.proto
```diff
@@ -456,6 +456,11 @@ message TensorProto {
   repeated uint64 uint64_data = 11 [packed = true];
 }
 
+// TensorProtos stores multiple TensorProto objects in one single proto.
+message TensorProtos {
```
houseroad: Wanna try sequence: https://github.com/onnx/onnx/blob/master/onnx/onnx.proto#L515

houseroad:
Btw, there is no good way to represent sequence and map type weights in initializer list. We have created an issue: https://github.com/onnx/onnx/issues/2410

onnx/onnx.in.proto
yinghai: How to use this?

yinghai:
What's the status about this?

PR #2418, opened 2019-10-25
Keep symbolic dims in Concat with a single input

shinh:
When there is only a single input, we need to discard the
symbolic dimension in the specified axis.


onnx/defs/tensor/defs.cc
```diff
@@ -317,6 +317,10 @@ ONNX_OPERATOR_SET_SCHEMA(
               !hasNInputShapes(ctx, static_cast<int>(numInputs))) {
             return;
           }
+          if (numInputs == 1) {
+            propagateShapeFromInputToOutput(ctx, 0, 0);
```
skottmckay: Would be better to move this down so it's after the axis is validated.

shinh: Sorry for the latency, done!

PR #2422, opened 2019-10-29
Print value case of TypeProto more friendly

take-cheeze:


onnx/shape_inference/implementation.cc
```diff
@@ -6,6 +6,19 @@ namespace ONNX_NAMESPACE {
 namespace shape_inference {
 namespace {
 
+std::string getValueCaseString(const TypeProto& type) {
```
skottmckay: Why not explicitly handle all currently defined values? 

  enum ValueCase {
    kTensorType = 1,
    kSequenceType = 4,
    kMapType = 5,
    kSparseTensorType = 8,
    kOpaqueType = 7,
    VALUE_NOT_SET = 0,
  };

PR #2425, opened 2019-10-29
fix test bugs for resize op version 11

askhade:
Some of the tests have bugs where the model is not given the expected inputs.

For example test : resize_upsample_sizes_nearest_ceil_half_pixel
Expects the results with rounding mode as ceil however the model is not given this as input therefore during inferencing backends end up using the default rounding mode and the test fails.


askhade:
@daquexian : I noticed the formula for coordinate transformation which is specified in the spec https://github.com/onnx/onnx/blob/3ea3b0e04cc0cb0f2fd250cf40b7e284d8bb945c/onnx/defs/tensor/defs.cc#L1678 is different than the one in the interpolate_1d method used in resize.py 
https://github.com/onnx/onnx/blob/3ea3b0e04cc0cb0f2fd250cf40b7e284d8bb945c/onnx/backend/test/case/node/resize.py#L133 
https://github.com/onnx/onnx/blob/3ea3b0e04cc0cb0f2fd250cf40b7e284d8bb945c/onnx/backend/test/case/node/resize.py#L134
Can you clarify why it is different?


Either ways we need to change one so that they both match... If the change is needed in resize.py then I can add it to this PR... if it is required in defs.cc then we will need an opset bump

daquexian:
Nice catch. I think the changes is needed in defs.cc. The implementation in resize.py has been verified to be compatible with PyTorch [here](https://gist.github.com/daquexian/4bab98a7e706ad95616567433c1234b8)

askhade:
> Nice catch. I think the changes is needed in defs.cc. The implementation in resize.py has been verified to be compatible with PyTorch [here](https://gist.github.com/daquexian/4bab98a7e706ad95616567433c1234b8)

This means we need to update the opset version for this change... I will include it as part of another PR.

There are more changes required in defs.cc for example : 

Here scales is meant to be optional but it is not declared as optional... "OpSchema::Optional" needs to be explicitly specified for any optional inputs. 
https://github.com/onnx/onnx/blob/3ea3b0e04cc0cb0f2fd250cf40b7e284d8bb945c/onnx/defs/tensor/defs.cc#L1742


I am planning to make these changes as part of the next PR and bump up the opset version

PR #2426, opened 2019-10-30
Update model creation example due to change in Pad spec in latest opset

### LINKED ISSUE #2423: PythonAPIOverview.md: the exampe from "Creating an ONNX Model Using Helper Functions" fails
https://github.com/onnx/onnx/blob/master/docs/PythonAPIOverview.md

```
  File "create-onnx-model.py", line 38, in <module>
    onnx.checker.check_model(model_def)
  File "/usr/local/lib/python3.6/site-packages/onnx/checker.py", line 91, in check_model
    C.check_model(model.SerializeToString())
onnx.onnx_cpp2py_export.checker.ValidationError: Node () has input size 1 not in range [min=2, max=3].

==> Context: Bad node spec: input: "X" output: "Y" op_type: "Pad" attribute { name: "mode" s: "constant" type: STRING } attribute { name: "pads" ints: 0 ints: 1 ints: 0 ints: 1 type: INTS } attribute { name: "value" f: 1.5 type: FLOAT }
```

skottmckay:
Fix #2423 

An alternative would be to explicitly pick an older opset so the original example works, however this example with the latest Pad spec includes showing how initializers can be added so it's slightly more comprehensive.

### DIFF
--- a/docs/PythonAPIOverview.md
+++ b/docs/PythonAPIOverview.md
@@ -64,8 +64,12 @@ from onnx import AttributeProto, TensorProto, GraphProto
 # https://github.com/onnx/onnx/blob/master/onnx/onnx.proto
 
 
-# Create one input (ValueInfoProto)
+# Create inputs (ValueInfoProto)
 X = helper.make_tensor_value_info('X', TensorProto.FLOAT, [1, 2])
+
+# Create initializer with static pads and constant value
+pads = helper.make_tensor('pads', TensorProto.INT64, [4], [0, 1, 0, 1])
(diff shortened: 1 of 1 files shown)


aviaisr:
Can you please check this issue? It shows a similar error to issue 2423. 
I'm working with the latest master 1.6.0

https://github.com/onnx/onnx/issues/2459#issue-524251228

aviaisr:
Isn't it the same padding error you've fixed?
https://github.com/onnx/onnx/issues/2423#issue-514020094

jcwchen:
This should be covered by this: https://github.com/onnx/onnx/pull/2491. Close this one now.

PR #2437, opened 2019-11-07
Introduce POSIX_DATETIME primitive type represented by int64.

yuslepukhin:


gramalingam:
LGTM. It would be good to have some toy unit test case for the added code. But, off the top of my head, I am not sure whether we can do anything simple without adding an op or two that uses "datetime". Can we create some dummy op that uses datetime, but is not part of the standard onnx opset?

yuslepukhin:
> 
> 
> LGTM. It would be good to have some toy unit test case for the added code. But, off the top of my head, I am not sure whether we can do anything simple without adding an op or two that uses "datetime". Can we create some dummy op that uses datetime, but is not part of the standard onnx opset?

It looks like ops that are not defined in ONNX are ignored: https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/__init__.py#L76

gramalingam:
@houseroad : any feedback on this?

onnx/mapping.py
```diff
@@ -21,7 +21,7 @@
     int(TensorProto.COMPLEX64): np.dtype('complex64'),
     int(TensorProto.COMPLEX128): np.dtype('complex128'),
     int(TensorProto.UINT32): np.dtype('uint32'),
-    int(TensorProto.UINT64): np.dtype('uint64'),
+    int(TensorProto.POSIX_DATETIME): np.dtype('uint64'),
```
houseroad: why remove UNIT64 case?

onnx/onnx-ml.proto
```diff
@@ -592,8 +596,4 @@ message OperatorSetIdProto {
   // The version of the operator set being identified.
   // This field MUST be present in this version of the IR.
   optional int64 version = 2;
-}
-
-// For using protobuf-lite
-option optimize_for = LITE_RUNTIME;
```
houseroad: Why remove this?

onnx/onnx-ml.proto
```diff
@@ -592,8 +596,4 @@ message OperatorSetIdProto {
...
-
+}
```
houseroad: endline charactor?

onnx/onnxifi.h
```diff
@@ -38,13 +38,13 @@ extern "C" {
 #endif
```
houseroad: please don't mix too much formatting in one PR.

onnx/mapping.py
yuslepukhin: Oversight

---
In reply to: [344909641](https://github.com/onnx/onnx/pull/2437#discussion_r344909641) [](ancestors = 344909641)

onnx/onnx-ml.proto
yuslepukhin: This was copied as one of the build products. I will re-do change manually.

---
In reply to: [344909908](https://github.com/onnx/onnx/pull/2437#discussion_r344909908) [](ancestors = 344909908)

onnx/onnxifi.h
yuslepukhin: Will do the change outside VS which often applies formatting automatically

---
In reply to: [344910141](https://github.com/onnx/onnx/pull/2437#discussion_r344910141) [](ancestors = 344910141)

yuslepukhin:
Build fails bc Conda protobuf package has DLL in it not a static library and therefore it can not link.

yuslepukhin:
@houseroad Can you, please, check if the changes are addressed?

gramalingam:
@houseroad : do the updates look fine to you? Thanks!

PR #2452, opened 2019-11-13
Fix windows conda build

askhade:
A lot of ONNX users hit issues when building ONNX on windows in a Anaconda environment. 
This PR adds :
1. instructions to build onnx on windows the right way (by building protobuf)
2. an alternative way to successfully build ONNX in Anaconda environment (with a few caveats)


CMakeLists.txt
```diff
@@ -445,6 +457,13 @@ if(BUILD_ONNX_PYTHON)
...
                                            # uf\wire_format_lite.h
                                    ${EXTRA_FLAGS})
+    if(ONNX_USE_PROTOBUF_SHARED_LIBS)
+      target_compile_options(onnx_cpp2py_export 
+                             PRIVATE /wd4251 # 'identifier' : class 'type1' needs to 
+                                             # have dll-interface to be used by 
+                                             # clients of class 'type2'
+                            )
+    endif()
     target_compile_options(onnx_cpp2py_export PRIVATE /MT)
```
snnn: No. It won't work.
See: line 467 in your change

snnn: "/MT" causes the application to use the multithread, static version of the run-time library. Defines _MT and causes the compiler to place the library name LIBCMT.lib into the .obj file so that the linker will use LIBCMT.lib to resolve external symbols.

In this case, it must use "/MD"

README.md
```diff
@@ -91,6 +91,56 @@ sudo apt-get install protobuf-compiler libprotoc-dev
...
 
+### Windows
+When building on Windows it is highly recommended that you also build protobuf locally as a static library. The version distributed with conda-forge is a DLL and this is a conflict as ONNX expects it to be a static lib.
+
+#### Instructions to build protobuf and ONNX on windows
+Step 1 : Build protobuf locally
+```
+git clone https://github.com/protocolbuffers/protobuf.git
+git checkout 3.9.x
+cd cmake
```
skottmckay: 'cd protobuf' between clone and checkout?

README.md
```diff
@@ -91,6 +91,56 @@ sudo apt-get install protobuf-compiler libprotoc-dev
...
+Step 2: Build ONNX
+```
+# Get ONNX
+git clone https://github.com/onnx/onnx.git
+cd onnx
+git submodule update --init --recursive
+
+# Set environment variables to find protobuf
+set PATH=<protobuf_install_dir>\bin;%PATH%
+
```
skottmckay: I would suggest adding this path to the user environment variables so it's more permanent. Otherwise it's easily forgotten and the next time they try to rebuild they can get errors and not remember that a custom protobuf directory needed to be added to the path.

CMakeLists.txt
yuslepukhin: We did not want to change the default but the env var provides an option for /MD[d]

---
In reply to: [345602405](https://github.com/onnx/onnx/pull/2452#discussion_r345602405) [](ancestors = 345602405)

cmake/Utils.cmake
```diff
@@ -7,6 +7,12 @@ function(add_msvc_runtime_flag lib)
     else()
       target_compile_options(${lib} PRIVATE /MT)
     endif()
+  else()
+    if(${CMAKE_BUILD_TYPE} MATCHES "Debug")
```
snnn: See #1995 

It interferes in creating VS Solutions that support multiple configurations.

snnn: See #1563

snnn: https://stackoverflow.com/questions/24460486/cmake-build-type-is-not-being-used-in-cmakelists-txt

CMakeLists.txt
```diff
@@ -44,6 +45,17 @@ if(NOT DEFINED ONNX_VERIFY_PROTO3)
   option(ONNX_VERIFY_PROTO3 "Generate code by proto3" ${PROTO3_ENABLED})
 endif()
 
+ if (ONNX_USE_PROTOBUF_SHARED_LIBS)
+  if(MSVC)
+    add_definitions(-DPROTOBUF_USE_DLLS)
+  else()
+    message(
```
snnn: Would be better if you can check if ONNX_USE_MSVC_STATIC_RUNTIME is OFF.

cmake/Utils.cmake
askhade: Looks like we can get rid of this else which I added since the that is the default anyways... earlier 
-DONNX_USE_MSVC_STATIC_RUNTIME was always set to ON state... now we have an option to set this to OFF...

appveyor.yml
```diff
@@ -41,8 +41,15 @@ before_build:
 - >
   IF "%_arch%"=="x64"
   (
+  IF "%CONDA_PREFIX%"=="C:\Miniconda35-x64"
+  (
+  conda install -y -c conda-forge libprotobuf=3.5.2 numpy
```
snnn: Why the version number is missed?

CMakeLists.txt
askhade: I dont quite understand... what purpose will this check serve?

setup.py
```diff
@@ -41,6 +41,9 @@
 # Global variables for controlling the build variant
 ################################################################################
 
+# Default value is set to TRUE\1 to keep the settings same as the current ones.
+# However going forward the recomemded way to is to set this to False\0
+USE_MSVC_STATIC_RUNTIME = bool(os.getenv('USE_MSVC_STATIC_RUNTIME', '1') == '1')
```
yuslepukhin: >USE_MSVC_STATIC_RUNTIME  [](start = 0, length = 24)

This is would be false by default here?

cmake/Utils.cmake
yuslepukhin: Yes, just need to come up with a way without if/else clause

---
In reply to: [346084942](https://github.com/onnx/onnx/pull/2452#discussion_r346084942) [](ancestors = 346084942)

setup.py
askhade: USE_MSVC_STATIC_RUTIME will be tru by default which means we will use /MT option (current option) by default

---
In reply to: [346494736](https://github.com/onnx/onnx/pull/2452#discussion_r346494736) [](ancestors = 346494736)

CMakeLists.txt
askhade: yes added the check... For current versions of protobuf from conda-forge which are built as a shared lib "ONNX_USE_MSVC_STATIC_RUNTIME" needs to be turned OFF

appveyor.yml
```diff
@@ -37,22 +37,27 @@ install:
...
+  IF "%CONDA_PREFIX%"=="C:\Miniconda35-x64"
+  (
+  conda install -y -c conda-forge libprotobuf=3.5.2 numpy  
   )
   ELSE
   (
-  IF "%CONDA_PREFIX%"=="C:\Miniconda36-x64"
-  (
-  conda install -y -c conda-forge libprotobuf=3.5.2 numpy=1.17.0
+  conda install -y -c conda-forge libprotobuf=3.5.2=he51fdeb_1 numpy=1.17.0
```
snnn: Why don't use the latest libprotobuf?

askhade: Because latest one is built as a shared lib... Current change to appveyor simply forces it use the package version which it was using earlier and not take the latest update...

The ideal change to this CI pipeline should be to use our recommended way .i.e. build protobuf locally instead of pulling from conda-forge... but this change is a bigger one and will need all important stakeholders to agree which means it will take longer... Since right now all the PRs are blocked I want to push this change and then open discussion for the above mentioned change

snnn: But the 3.5.x version has a serious bug on Windows: https://github.com/protocolbuffers/protobuf/pull/4777 
https://github.com/protocolbuffers/protobuf/issues/4773

askhade: Well appveyor CI was already using 3.5.2 version... as I mentioned the aim of this PR for appveyor CI is to just force it its old behavior. Its good that you pointed this out... I am starting another thread to get a consensus on the right fix for this scenario. From my perspective the right fix is to simply build protobuf locally so that we have a better control... relying on conda-forge for protobuf has proven to be very painful so far,

snnn: >  simply build protobuf locally
But in conda, you shouldn't do that. Because people can install protobuf from conda, and they can't safely build onnx if onnx has a local protobuf lib, in such case you don't know which protobuf lib/dll/header files will be used.

README.md
```diff
@@ -91,6 +91,61 @@ sudo apt-get install protobuf-compiler libprotoc-dev
...
+```
+
+If you do not want to build protobuf and instead want to use protobuf from conda forge then follow these instructions. 
+However please note : This method is just added as a convenience for users and there is very limited support from ONNX team when using this method.
+
+#### Instructions to build ONNX on windows in anaconda environment
+
+```
+# Use conda-forge protobuf
+conda install -c conda-forge protobuf numpy
```
gramalingam: Don't you need to specify the version for protobuf? I use protobuf=3.5.1 in a conda install.

README.md
```diff
@@ -91,6 +91,61 @@ sudo apt-get install protobuf-compiler libprotoc-dev
...
+# even better option is to add it to user\system PATH so this step can be performed only once.
+# For more details check https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=vs-2017
+set PATH=<protobuf_install_dir>\bin;%PATH%
+set USE_MSVC_STATIC_RUNTIME=0
+
+# Build ONNX
+python setup.py install
+```
+
+If you do not want to build protobuf and instead want to use protobuf from conda forge then follow these instructions.
```
gramalingam: Are there any requirements or recommendations for python version?

README.md
```diff
@@ -91,6 +91,61 @@ sudo apt-get install protobuf-compiler libprotoc-dev
 pip install onnx
 ```
 
+### Windows
```
gramalingam: Is the preceding explanation (under "Source", which overlaps the anaconda instructions below) for Linux? If so, please add a sub-section title "Linux" or as appropriate.

gramalingam: A clarification about the environment variable ONNX_ML would be helpful here (in the "Windows" section).

README.md
```diff
@@ -91,6 +91,61 @@ sudo apt-get install protobuf-compiler libprotoc-dev
...
+cd protobuf
+git checkout 3.9.x
+cd cmake
+# Explicitly set -Dprotobuf_MSVC_STATIC_RUNTIME=OFF to make sure protobuf does not statically link to runtime library
+cmake -G "Visual Studio 15 2017 Win64" -Dprotobuf_MSVC_STATIC_RUNTIME=OFF -Dprotobuf_BUILD_TESTS=OFF -Dprotobuf_BUILD_EXAMPLES=OFF -DCMAKE_INSTALL_PREFIX=<protobuf_install_dir>
+msbuild protobuf.sln /m /p:Configuration=Release
+msbuild INSTALL.vcxproj /p:Configuration=Release
+```
+
+Step 2: Build ONNX
```
gramalingam: Any instructions for a clean uninstall & rebuild? I know I had to do it some times, but I was using conda environments (not sure about outside).

snnn: Indeed cmake doesn't support uninstall, neither Windows nor Linux.

snnn: You may install the files into a separated folder and delete the whole folder afterwards.

README.md
askhade: Do we have an official supported version for protobuf which is common for all platforms or perhaps one for each platform ?
Windows CI uses 3.5.2 : Changming pointed out an existing issue with this version  protocolbuffers/protobuf#4777 protocolbuffers/protobuf#4773
Linux and MacOS use 3.10.0

I will test to see if 3.10.0 works for windows too... but am curious do we use these versions right now just because they work or is there any reasoning behind it?

README.md
askhade: In CIs - for Linux we test using python 2.7 and 3.6 , on Mac we test using 3.7 and 3.7 whereas on in windows ci we test with python 3.5, 3.6, 3.7

Right now I do not want to add any recommendations without understanding what is our official stance. There is no other documentation which lists the officially supported versions...

README.md
gramalingam: I believe there are some cleanup instructions for conda (removing the onnx folder from site-packages) which helps sometimes.

gramalingam: For example, one mysterious problem I have run into a few times is the following: I add a new op, run setup. I then run the document-generator, but it seems to access the old schema registry, not the updated one with the new op. I never figured out what was going on.

askhade: Yes there are a bunch of issues in this area too... We have a script update_docs... do you see the same problem when you run that script instead of independently running gen_doc? If yes then we should just fix that script to do the right thing... The existing doc for adding a new operator mentions this script.

snnn:
Should it be: Building ONNX with Protobuf as shared lib is only available for MSVC compilers and that too when USE_MSVC_STATIC_RUNTIME environment variable is set to **0**.?

PR #2461, opened 2019-11-18
Update Argmin/Argmax

lara-hdr:
Add param"select_last_index" to Argmin/Argmax in opset version 12, to specify whether to select the last index or the first index if the min/max appears in multiple indices.
This is needed to fully support both PyTorch and Tensorflow (PyTorch chooses the last index, Tensorflow the first index).


onnx/defs/reduction/defs.cc
```diff
@@ -149,6 +151,11 @@ The type of the output tensor is integer.)DOC";
         "Keep the reduced dimension or not, default 1 mean keep reduced dimension.",
         AttributeProto::INT,
         static_cast<int64_t>(1));
+    schema.Attr(
+        "select_last_index",
+        "Wether to select the last index or the first index if the {name} appears in multiple indices, default is False (first index).",
```
gramalingam: Wether => Whether

onnx/defs/reduction/defs.cc
```diff
@@ -134,8 +134,10 @@ std::function<void(OpSchema&)> ArgReduceDocGenerator(const char* name) {
   return [=](OpSchema& schema) {
     std::string doc = R"DOC(
 Computes the indices of the {name} elements of the input tensor's element along the 
-provided axis. The resulted tensor has the same rank as the input if keepdims equal 1.
+provided axis. The resulted tensor has the same rank as the input if keepdims equal 1. 
 If keepdims equal 0, then the resulted tensor have the reduced dimension pruned. 
+If select_last_index is True (default False), the index of the last occurence of the {name} 
+is selected if the {name} appears more than once in the input.
```
spandantiwari: We should also add explicitly what is the behavior when `select_last_index=False` (the default) in this blurb as well.

onnx/defs/reduction/defs.cc
```diff
@@ -134,8 +134,10 @@ std::function<void(OpSchema&)> ArgReduceDocGenerator(const char* name) {
   return [=](OpSchema& schema) {
     std::string doc = R"DOC(
 Computes the indices of the {name} elements of the input tensor's element along the 
-provided axis. The resulted tensor has the same rank as the input if keepdims equal 1.
+provided axis. The resulted tensor has the same rank as the input if keepdims equal 1.
```
spandantiwari: nit: Consider replacing "resulted tensor" with "resulting tensor" or just "output tensor" everywhere in these two ops. I think that is grammatically more correct.

lara-hdr:
@gramalingam can we merge this?
The new CI failure is unrelated.

PR #2464, opened 2019-11-18
Add clarification for bias quantization in QlinearConv Op spec

askhade:
Add clarification for bias quantization. 

This change should not need version update because for QlinearConv op to produce correct result there is only one way to quantize bias. Current PR only add clear documentation on how bias should be quantized.


askhade:
@linkerzhang , @fdwr : Add the clarification to the spec as discussed

PR #2465, opened 2019-11-19
Document error and resolution ImportError: cannot import name 'ONNX_ML'

yuslepukhin:


vinitra:
Re-running `pip install -e .` in the onnx directory also helped solved this error for me.

PR #2478, opened 2019-11-29
Fix wrong opset version in model downloading

daquexian:
As what has been repeatedly pointed in https://github.com/onnx/onnx/issues/2049, the opset version in the URLs of models should not be `onnx.des.onnx_opset_version()`. After waiting for a fix for a long time, I found it is easy to fix by maintaining a separate opset version.

The opset version should have been updated to 11 (the opset version of the latest stable release of ONNX), but URLs like https://s3.amazonaws.com/download.onnx/models/opset_11/vgg19.tar.gz are still invalid :| They seems ignored by ONNX maintainers. In order to keep fresh contributors from feeling confused and frustrated as soon as possible, I set the opset version to 10 for now.


daquexian:
The CI errors are not related to this PR

snnn:
https://github.com/onnx/models/issues/141 
CC: @vinitra .  We should restore the model pipelines and refresh these models.

PR #2494, opened 2019-12-07
[WIP] Fix build error with cmake 3.16

snnn:
1. Fix build error with cmake 3.16

See the discussion at the end of: https://gitlab.kitware.com/cmake/cmake/merge_requests/3555

2. Make ONNX_WERROR work on Windows


CMakeLists.txt
```diff
@@ -45,21 +45,28 @@ if(NOT DEFINED ONNX_VERIFY_PROTO3)
...
-    message(
-      FATAL_ERROR 
-        "Building ONNX with Protobuf as shared lib is only available for MSVC compilers and that too when USE_MSVC_STATIC_RUNTIME environment variable is set to 1. Please read the build instructions for more details."
-    )
+if(ONNX_USE_PROTOBUF_SHARED_LIBS)
+  set(Protobuf_USE_STATIC_LIBS OFF FORCE)
+else()
+  set(Protobuf_USE_STATIC_LIBS ON FORCE)
+  #only needed for cmake version < 3.16
+  if (ONNX_USE_PROTOBUF_SHARED_LIBS)
```
houseroad: Here ONNX_USE_PROTOBUF_SHARED_LIBS is always false, so we never enter this branch.

Shall we check the cmake version instead?

snnn: So, ONNX_USE_PROTOBUF_SHARED_LIBS  is always false, I didn't see any place it get set. 

The Protobuf_USE_STATIC_LIBS  is introduced in cmake 3.9. 
From cmake 3.9 to 3.15, the variable "Protobuf_USE_STATIC_LIBS" only affects Linux builds.  If it is OFF, it won't matter. If it's ON, then libprotobuf.so will be ignored during search because the option forces using protobuf as a static lib. As you pointed, it may not be a good idea.  Let me think it again.

PR #2500, opened 2019-12-12
Correct the order of arguments of InferShapes

shinh:
The order of arguments in implementation.cc mismatches with
the header. Since `check_type` is an optional argument, it
should appear after `opset_imports`.


shinh:
@RandySheriffH and @gramalingam: Could you review this PR? I think we cannot use 
 InferShapes(Graph) now and this issue was introduced in https://github.com/onnx/onnx/pull/2367

PR #2504, opened 2019-12-13
Add Einsum op

neginraoof:
The Einsum operator evaluates algebraic tensor operations on the operands, using the Einstein summation convention.
It takes the Einsum equation (string) and operand tensors as inputs, calculates the equation, and returns the output tensor.
PyTorch: https://pytorch.org/docs/stable/torch.html?highlight=einsum#torch.einsum
TensorFlow: https://www.tensorflow.org/api_docs/python/tf/einsum

For shape inference, we can only infer the output rank from the equation string. Finding output size requires calculating the algebraic operations.


neginraoof:
@BowenBao @lara-hdr please review
cc @ebarsoum @linkerzhang @spandantiwari

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1665,126 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+static const char* Einsum_ver11_doc = R"DOC(
+The Einsum operator evaluates algebraic tensor operations on the operands, using the Einstein summation convention.).
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    Einsum,
+    12,
+    OpSchema()
+        .SetDoc(Einsum_ver11_doc)
+        .Input(0, "Equation", "A 0-D tensor holding Einsum expression in UTF-8 String", "tensor(string)")
```
lara-hdr: can we have this as an attribute instead or is there a reason to have it as an input?

docs/Changelog.md
```diff
@@ -13965,3 +13965,34 @@ This version of the operator has been available since version 12 of the default
 <dd>Constrain input and output types to all numeric tensors.</dd>
 </dl>
 
+### <a name="Einsum-12"></a>**Einsum-12**</a>
+
+  The Einsum operator evaluates algebraic tensor operations on the operands, using the Einstein summation convention.).
```
lara-hdr: ")." -> typo?

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1665,126 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    for (size_t index = 0; index < number_of_letters; index++) {
+      if (num_letter_occurrences[index] == 1) {
+        output_shape->add_dim();
+      }
+    }
+  }
+}
+
+static const char* Einsum_ver11_doc = R"DOC(
+The Einsum operator evaluates algebraic tensor operations on the operands, using the Einstein summation convention.).
```
lara-hdr: ")." -> typo?

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1665,126 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    Einsum,
+    12,
+    OpSchema()
+        .SetDoc(Einsum_ver11_doc)
```
lara-hdr: rename "Einsum_ver11_doc" to "Einsum_ver12_doc"

onnx/defs/math/defs.cc
lara-hdr: mhmm. .  actually tensor(string) are not supported in PyTorch, how would we support this as an input in the PyTorch exporter?

gramalingam: Yes, I think we should make it an attribute.

onnx/defs/math/defs.cc
gramalingam: I think it would be helpful to add a precise, complete, and unambiguous description of this language. From the description I have seen in other frameworks, there are some choices and ambiguities in what Einstein summation notation means and allows.

gramalingam: It would be good to have a simple grammar for the language too. For example, are white-spaces allowed in the string? It is a good idea to allow spaces in the string and skip them. When I look at the rank-inference logic, the use term.size() makes me wonder if this assumes that there are no white-spaces in the string.

onnx/defs/math/defs.cc
neginraoof: Yes, that makes sense

onnx/defs/math/defs.cc
neginraoof: Thanks @gramalingam, I've updated the doc with more details about equation string.

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1666,130 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            OpSchema::all_tensor_types(),
+            "Constrain input and output types to all tensor types.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          // Type inference
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          std::string equation = getAttribute(ctx, "equation", "");
+          if (equation.compare("") == 0) {
+            return;
+          }
+					einsumRankInference(ctx,equation);
```
lara-hdr: nit: too many spaces

neginraoof: Thanks!

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1666,130 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+  const size_t numInputs = ctx.getNumInputs();
+  if (numInputs < 1 || !hasNInputShapes(ctx, static_cast<int>(numInputs))) {
+    return;
+  }
+
+  auto* output_shape = getOutputShape(ctx, 0);
+  std::string  left_equation;
+
+  equation.erase(std::remove(equation.begin(), equation.end(), ' '), equation.end()); // Remove whitespaces
```
gramalingam: Whitespace generally includes other characters like tab and newline. I think it may be better to either use "isspace" function or change the documentation if only blank characters are permitted. I see that the numpy implementation is used as the reference implementation ... we could go with whatever numpy supports and document it.

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1666,130 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+  std::string term;
+  size_t num_operands = 0;
+  size_t num_ellipsis = 0;
+  size_t num_ellipsis_indices = 0;
+
+  // Parse the left-hand side
+  std::stringstream str(left_equation);
+  while(std::getline(str, term, ',')) {
+    auto ellipsis_index = term.find("...");
```
gramalingam: A check to handle the case where number of operands here exceeds numInputs would be helpful.

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1666,130 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+static const char* Einsum_ver12_doc = R"DOC(
+The Einsum operator evaluates algebraic tensor operations on a sequence of tensors, using the Einstein summation
+convention. The equation string contains a comma-separated sequence of lower case letters. Each term corresponds to
+an operand tensor, and the characters within the terms correspond to operands dimensions.
+This sequence may be followed by a '->' to separate the left and right hand side of the equation.
+If the equation contains '->' followed by the right-hand side, the explicit (not classical) form of Einstein summation
+is performed, and the right-hand side indices indicate output tensor dimensions. In other cases,
+output indices are (implicitly) set to the alphabetically sorted sequence of indices appearing exactly once in the
+equation.
+The equation may contain ellipsis ('...') to indicate a fixed number of dimensions. The right-hand
```
gramalingam: Can you clarify what happens for the ellipsis? E.g., all occurrences of ellipsis on the left-hand-side must denote the same number of dimensions. Is there broadcasting or not?

gramalingam: When an index variable is repeated on the left-hand-side, are the corresponding dimensions required to be equal?

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1666,130 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+      if (num_letter_occurrences[index] == 1) {
+        output_shape->add_dim();
+      }
+    }
+  }
+}
+
+static const char* Einsum_ver12_doc = R"DOC(
+The Einsum operator evaluates algebraic tensor operations on a sequence of tensors, using the Einstein summation
+convention. The equation string contains a comma-separated sequence of lower case letters. Each term corresponds to
```
gramalingam: Adding an explanation as below (please correct it as appropriate) right at the beginning may be helpful:

An einsum of the form `term1, term2, term3 -> output-term` produces an output tensor using the following equation:
```
   output[output-term] = implicit-sum( input1[term1] * input2[term] * input3[term3])
```
where the implicit-sum performs a summation over all the indices occurring in in the input terms (term1, term2, and term3) that do not occur in the output-term.

onnx/defs/math/defs.cc
```diff
@@ -1750,17 +1750,22 @@ void einsumRankInference(
 }
 
 static const char* Einsum_ver12_doc = R"DOC(
+An einsum of the form '''term1, term2, term3 -> output-term ''' produces an output tensor using the following equation
+
+'''output[i][j][k] = explicit-sum( input1[term1] * input2[term] * input3[term3] )'''
```
gramalingam: Isn't the number of indices i, j, k and their order determined by output-term? So, it may be better to write "[output-term]" instead of "[i][j][k]", using the same notation as in "[term1]" etc. Alternatively, we should explain the relationship between "[i][j][k]" and output-term: this explanation is shared by "term1]" etc..

neginraoof: It think now it looks fine. Except that the matrix multiplication in the equation is not very clear. In some cases, depending on the shape of the operands, this could be expand + multiply.

onnx/defs/math/defs.cc
neginraoof: This is in line 1714

onnx/defs/math/defs.cc
neginraoof: Addressed this.

neginraoof:
@gramalingam Can we close this?

docs/Changelog.md
```diff
@@ -13965,6 +13965,63 @@ This version of the operator has been available since version 12 of the default
...
+
+<dl>
+<dt><tt>Output</tt> : T</dt>
+<dd>Output tensor</dd>
+</dl>
+
+#### Type Constraints
+
+<dl>
+<dt><tt>T</tt> : tensor(uint8), tensor(uint16), tensor(uint32), tensor(uint64), tensor(int8), tensor(int16), tensor(int32), tensor(int64), tensor(float16), tensor(float), tensor(double), tensor(string), tensor(bool), tensor(complex64), tensor(complex128)</dt>
```
spandantiwari: We should remove `tensor(string)` from supported types and keep only numeric types.

neginraoof: Thanks a lot for the note. I've updated the docs.

neginraoof:
cc @gramalingam for review.
Thanks.

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1666,141 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(0,
+            "Inputs",
+            "Operands",
+            "T",
+            OpSchema::Variadic)
+        .Output(0, "Output", "Output tensor", "T")
+        .TypeConstraint(
+            "T",
+            OpSchema::all_numeric_types(),
+            "Constrain input and output types to all tensor types.")
```
TMVector:
```suggestion
            "Constrain input and output types to all numerical tensor types.")
```

onnx/defs/math/defs.cc
gramalingam: The problem is what happens in line 1701 with "ctx.getInputType(num_operands)->tensor_type()"? This may crash before we get to line 1714 if num_operands > numInputs, right?

gramalingam: May be a test-case with num_operands > num_inputs would also help.

onnx/defs/math/defs.cc
gramalingam: Thanks. I still think that it is better to say "output[output-term]" instead of "output" in the equation.

onnx/defs/math/defs.cc
neginraoof: Sure. Does it make sense to have a negative shape inference test?

gramalingam: I think so. Test cases to check that the behavior is acceptable when preconditions are not satisfied are useful. E.g., something like in: https://github.com/onnx/onnx/blob/dfa4384c3f8fe9d804032f992e6f818fc54152b0/onnx/test/shape_inference_test.py#L109

gramalingam:
@houseroad : any feedback?

neginraoof:
cc @postrational @houseroad 
Would you please help review this PR for adding Einsum op

spandantiwari:
Thanks @houseroad, @gramalingam, and @lara-hdr for your review.

PR #2510, opened 2019-12-17
add 8 bit support to maxpool op

askhade:
Adding 8 bit data-type support for MaxPool operator.


askhade:
@linkerzhang @ebarsoum : Please take a look
@fdwr , @tracysh, @nickfeeney :  FYI

onnx/defs/nn/defs.cc
```diff
@@ -308,12 +318,14 @@ ONNX_OPERATOR_SET_SCHEMA(
...
     MaxPool,
-    11,
+    12,
     OpSchema()
         .FillUsing(PoolOpSchemaGenerator(
             "MaxPool",
             "max",
-            "The output of each pooling window is maximum number of elements exclude pad.",
+            "The output of each pooling window is maximum number of elements exclude pad. "
+            "In case of 8 bit data types the output and input scales and zero points need to be same.",
```
linkerzhang: this line is not needed. In this case, MaxPool is just doing its regular job of finding out the max number, it does not really care the semantics behind each inputs.

PR #2516, opened 2019-12-18
add 8 bit support to reducemin and reducemax ops

askhade:
Adding 8 bit support to reducemin and reducemax ops


askhade:
@linkerzhang @ebarsoum : Please take a look
@fdwr , @tracysh, @nickfeeney : FYI

PR #2314, opened 2019-09-16
Training Proposal: Spec Changes and Gradient Operator

wschin:
~~This PR includes all necessary changes for enabling training. It currently includes #2013, #1955, #1970, #1959, #2168, #1939. Per Training WG's discussion, we want to put everything into a single place for convenience and having a global view to our status.~~

~~Please see #2038 and those sub-PRs' messages for details.~~

To speedup the code review process and properly distribute works, this PR now only contains content from #2517 by @tbennun, #2013, and #2168,. They are the minimal requirement to express training graph. Losses and optimizers will be added separately.

This PR introduces several major changes.
1. Add a protobuf message, `TrainingInfoProto` originally designed in #2013, to store training information.
2. In `TrainingInfoProto`, the user can store training algorithm in `algorithm` field as a `GraphProto`.
3. The user can also store initialization algorithm for resetting the model in `TrainingInfoProto.initialization` (proposed by @tbennun in #2517 and agreed by Training WG).
4. `ModelProto.graph` is callable inside `TrainingInfoProto.algorithm`. `ModelProto.graph.initializer` are visible to nodes in `TrainingInfoProto.algorithm.node`.
5. This PR also introduces a `Gradient` operator to differentiate a function represented by a (sub-)graph. This idea is from #2168. The domain of `Gradient` is `ai.onnx.training`.

PR #2517, opened 2019-12-18
Graph initializers for tensors

tbennun:
ONNX now supports dense and sparse tensors as initializers. This PR adds missing IR documentation for sparse tensor initializers, and adds an initializer type that is based on a computational graph. 

The reasoning behind the distinction between an initializer graph and simply adding nodes to the ONNX model is to support the strict separation between how a model is computed and the way to initialize (some of) its tensors. For example, pre-trained models can use tensor initializers to store weights, whereas untrained models can provide schemes for the initialization of weights (e.g., a `RandomNormal` operator followed by offset and scaling).

Our use case for this is reproducible training workloads (such as in MLPerf). Instead of defining a reference model as a PyTorch/TensorFlow script, with this PR one could simply upload an .onnx file that competitors can use. Beyond reducing the file size from hundreds of megabytes to a something that can be pushed to git, the initializers also act as "recipes" to what the fair starting conditions of training should be (as that differs from model to model).


onnx/onnx.in.proto
```diff
@@ -287,6 +287,9 @@ message GraphProto {
   // Initializers (see above) stored in sparse format.
   repeated SparseTensorProto sparse_initializer = 15;
 
+  // Initializers for tensor values in the form of graphs.
+  repeated GraphProto graph_initializer = 16;
```
linkerzhang: don't really understand the scenario of having a "Graph" as initializers of a model. Do you mind to elaborate that in our gitter room. https://gitter.im/onnx/Infra

gramalingam:
It looks like this might also be related to the training-proposal to extend the IR to support training: please see: https://github.com/onnx/onnx/pull/2314 ... training needs more support, since it also needs to describe how to update the weights after processing a batch.

tbennun:
@linkerzhang Thanks! Posted more details on the gitter.

@gramalingam Yes, this PR is related to #2314 but currently orthogonal to it, as the initializers there are still defined as Tensors.

linkerzhang:
@wschin

linkerzhang:
Comments based on chatting over gitter room - https://gitter.im/onnx/Infra:
1. Both IR.md and *.proto files should be commented (clarifying the 1:1 name mapping is still there between the only-one output of these graphs and initializers (weights).
2. Comments from training work group folks if any @wschin @SherlockNoMad @chinhuang007 

One more question, is it necessary to have one graph (with one output) mapping to one weight? how about just using one graph to cover all weights?

tbennun:
There is no functional difference between having one graph to initialize all tensors or having a graph per tensor, but there are two main reasons why we thought it was best to use a graph per tensor:
* In both cases, in order to create different weights for each tensor, you would have to replicate the graph nodes.
* Using the name field of a GraphProto is more straightforward to implement (both saving and loading) rather than iterating over nodes in the singular initializer graph and updating the output names. This is also more equivalent to how tensor and sparse tensor initializers are defined now, so the implementations would remain similar.

onnx/onnx.in.proto
```diff
@@ -287,6 +287,12 @@ message GraphProto {
   // Initializers (see above) stored in sparse format.
   repeated SparseTensorProto sparse_initializer = 15;
 
+  // A list of named graphs, used to specify how model inputs should be 
+  // initialized. Each GraphProto must have a name (distinct from any other
+  // initializer) and one output, which is of the same type and shape as the
+  // corresponding model input.
```
TMVector: How about this?
> Initializers (see above) described by graphs. Each GraphProto must have a name (which acts as the initializer name), zero inputs, and one output (which acts as the initializer tensor).

TMVector: I think the type/shape constraint is implied by the spec already, but if you think it is worth being explicit, then IMO it should be on the `initializer` field since it applied to all initializers.

tbennun: I agree, this is why I didn't mention the type/shape constraint in the first iteration. The zero inputs requirement is also important. Thanks!

docs/IR.md
```diff
@@ -157,6 +157,8 @@ Graphs have the following properties:
 name|string|The name of the model graph.
 node|Node[]|A list of nodes, forming a partially ordered computation graph based on input/output data dependencies.
 initializer|Tensor[]|A list of named tensor values. When an initializer has the same name as a graph input, it specifies a default value for that input. When an initializer has a name different from all graph inputs, it specifies a constant value.
+sparse_initializer|SparseTensor[]|A list of initializer tensor values that are represented by sparse tensors.
+graph_initializer|Graph[]|A list of initializers that are represented by graphs, which describe how to compute the initializer tensor. Each initializer graph has zero inputs and one output, which acts as the initializer tensor. The initializer graph name acts as the initializer name.
```
wschin: I don't quite sure the reasons behind graph_initializers. If an initializer is input-dependent, it should be computed from the input in the main computation graph. If an initializer is input-independent, it sounds a constant and the current initializer support is sufficient. Are there cases we must compute initializers from separated graphs? Or this new field is mainly for convenience?

ONNX training spec will support some ways to **compute** the initializers and possibly have a similar semantic here. It's also strange if we have a main graph, training graph, and initialization graph --- should we execute initialization graph each time we execute the training graph?

tbennun: I see. I created the PR after we saw that the training spec does not include an initialization graph. Training is the main purpose for this change.

IMO there may still be other reasons to keep the initialization separate from training. For instance, if for inference some weights are represented by large tensors that can be simply computed, it would be better to initialize tensors once. Currently, if they are added as constant initializers, the file size would be large. If they are added as nodes to the graph, they will be recomputed every time, which might not be efficient. 

It would probably be fine if this change will only be applied in `TrainingInfoProto`. What do you think?

wschin: If we want to use graph to compute large constant tensor, we can specify that logic in the main inference graph (that is, we insert **nodes** into inference graph). Is it necessary to define another field for that logic?

For combining this proposal and training, we need to first list out the scenarios we want to support. For example, if we only want to express `initialization + training` in the training graph, we can merge `initialization` and `training` into one single graph and extra fields are not necessary.

linkerzhang: @wschin I am understanding this PR is trying to add a field to show the initialization logic of weights in training scenario. @tbennum correct me if I'm wrong. This is why I think this should align with training proposal.

tbennun: @linkerzhang yes, that is the goal.

wschin: Training's initialization is a very specific process. It should be executed for training from scratch but it can't be executed for transferring learning. If we put some a layer's weight into `graph_initializer`, it means we don't necessarily have a placeholder for the tensor, which can cause some problems when training algorithm updates its value. For example, if I define `W` as a graph initializer, where should I save its value produced by the training algorithm?

The signature of `graph_initializer` looks fine. We just need to refine its semantic. It should be an optional step to update `existing` initializers rather than defining new initializers.

onnx/onnx-ml.proto
```diff
@@ -290,6 +290,12 @@ message GraphProto {
   // Initializers (see above) stored in sparse format.
   repeated SparseTensorProto sparse_initializer = 15;
 
+  // A list of named graphs, used to specify how model inputs should be 
+  // initialized. Each GraphProto must have a name (distinct from any other
```
wschin:
```suggestion
  // initialized. Each GraphProto must have a name (the same to one initializer
```

onnx/onnx-ml.proto
```diff
@@ -290,6 +290,12 @@ message GraphProto {
...
+  // initializer) and one output, which is of the same type and shape as the
+  // corresponding model input.
+  repeated GraphProto graph_initializer = 16;
```
wschin: Should we put the entire initialization stage into a single graph? I feel it's cleaner than having a list of graphs.

linkerzhang:
With more thoughts of supporting training (plus the training proposal #2314 ),

I'd suggest,
1. move this initialization field as part of training proposal (TrainingInfoProto).
2. if my understanding is correct, this initialization for training should be run only once (for training reproducibility), in this way, I'd still suggest to have only one GraphProto to initialize all trainable weights.
3. The initialization graph's outputs will have exactly the same name as the "initializers" in inference graph. Say, in "initializers" there's a tensor named "a", if this "a" is trainable, there should be an output tensor named "a" in initialization graph.

Thoughts?

tbennun:
@linkerzhang I agree with everything. We also had a discussion in the training WG that supports this decision. As @wschin said, adding graph initializers to `TrainingInfoProto` has the additional benefit of being able to restart training on a pre-trained ONNX file, and the behavior of saving/loading models is more well defined.

I will now close this PR and create one targeted at the branch in #2314, so that it will appear as part of that PR. I will incorporate notes (2) and (3) from the message above. 

Thank you all for the important feedback!

Issue #2442, opened 2019-11-08
Operator or function CDist


xadupre:
I'd like to add cdist function. We made some experiments and having a dedicated operator is ten times faster than its onnx implementation with Scan operator (see graph below). What is the best way to add it? As a function or as an operator? In case of a function, should I wait for #1939 to be merged?

![image](https://user-images.githubusercontent.com/22452781/68491559-5d56e500-024b-11ea-9051-36349818039a.png)

xadupre:
More figures about this. I trained a KNearestClassifier on the digits datasets. First graph is without cdist operator, second graph is with the cdist oprator:

* black: scikit-learn (same for both graph)
* blue: onnxruntime
* yellow: onnx runtime implemented with numpy + cdist function from scipy in the second graph

It shows the runtime is 10 times faster with the use of cdist operator.

![image](https://user-images.githubusercontent.com/22452781/69899152-9a9e2680-1362-11ea-9dd6-dfc6df7a532c.png)

xadupre:
It is still relevant. However the list of distances to support can be huge.

justinchuby:
Marked for discussion for 1.16

justinchuby:
From discussion this operator is marked low priority. Please comment or reopen if there is further support to include it in the spec.

PR #2518, opened 2019-12-19
New Operator CDist (#2442)

xadupre:
The operator is needed to reach scipy cdist function performance (see #2442).

Open questions:

*Operator or function?* the operator can be composed with other ONNX operators but the graph depends on the parameter metric.

*User function?* Should we support custom user function?


appveyor.yml
```diff
@@ -17,19 +17,19 @@ environment:
...
-  - CONDA_PREFIX: C:\Miniconda36-x64
-    ONNX_VERIFY_PROTO3: 1
+  #- CONDA_PREFIX: C:\Miniconda36-x64
+  #  ONNX_VERIFY_PROTO3: 1
 
-  - CONDA_PREFIX: C:\Miniconda36
+  #- CONDA_PREFIX: C:\Miniconda36
 
-  - CONDA_PREFIX: C:\Miniconda35
+  #- CONDA_PREFIX: C:\Miniconda35
```
askhade: Why are all these test configurations being removed? The CI is failing right now and the ideal fix should be to fix the CI ad not remove the failing configs?

xadupre: Sorry for that, I just wanted to have faster iterations, I forgot to uncomment them before creating the PR.

xadupre: Python 3.5 is failing with this message: *IPython 7.10+ supports Python 3.6 and above*.

xadupre: I fixed the appveyor build (force the use of IPython 7.9 for python 3.5). This fix would unblock many PR too.

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1665,54 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    OpSchema()
+        .SetDoc(CDist_ver12_doc)
+        .Input(0, "X", "Input tensor X. Shape of X should be (M, K).", "T")
+        .Input(1, "Y", "Input tensor B. Shape of Y should be (N, K).", "T")
+        .Output(0, "Z", "Output tensor of shape (M, N).", "T")
+        .TypeConstraint(
+            "T", {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .Attr("metric",
+              "Metric to use to compute distances, it can be euclidean, "
```
snnn: Scipy supports much more metrics. Do we need any of these in the future? If yes, we'd better to add them here now.

xadupre: I kept the list to what is commonly used. Here is the whole list:
```
'seuclidean' 'cosine' 'correlation' 'hamming' 'jaccard'
'chebyshev' 'canberra' 'braycurtis' 
'yule' 'dice' 'kulsinski' 'rogerstanimoto' 'russellrao' 'sokalmichener' 'sokalsneath'
'mahalanobis' 'wminkowski' 'haversine'
```
See [cdist](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.cdist.html), and [DistanceMetric](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.DistanceMetric.html?highlight=distancemetric#sklearn.neighbors.DistanceMetric).

So you think I should do all of them?

EmmaNingMS: @linkerzhang @ebarsoum could you help review this op and give some guidance here?

snnn: I suggest we put all of them here, to avoid bumping up the op version in the future.

xadupre: Before I add them, what do you think of the rest of the PR? Is this the only required change?

ebarsoum:
@xadupre can you provide more details on: "Operator or function? the operator can be composed with other ONNX operators but the graph depends on the parameter metric."

xadupre:
CDist computes all pairs $\forall i,j, dist(X_i, Y_j)$, broadcast takes care of looping on j, operator Scan handles the loop on i: it execute N times the same distance broadcasted on N examples. Each distance is different and ends up being converted into ONNX with a different graph. There are >= 20 different distances for function cdist in scipy, it means >= 20 ONNX graphs to implement to implement CDist as a function. If it's an operator, there is no need to specify these graphs in ONNX. So implementing CDist as a function with all possible distances would take quite some time.

snnn:
This operator is often used in conjunction with[GaussianProcessRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html). Users are free to choose any one of the distance function as a part of the kernel function. So it's hard to say which distance function is less important than the others. I suggest we should add all of them(as @xadupre listed)  at one time, instead of frequently changing the spec.

PR #2519, opened 2019-12-20
Include <ostream> in common/status.h

CaseyCarter:
Before defining operator<<(std::ostream&, const Status&).

This fixes a build break with a future version of MSVC for which `<string>` doesn't transitively include `<ostream>`.


CaseyCarter:
FYI, it looks like AppVeyor CI is broken due to python version mismatch.

onnx/common/status.h
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <memory>
+#include <ostream>
```
linkerzhang: it's better to add this within a ifdef? (a macro of new msvc version?)

CaseyCarter: I think not. This header *always* needs the contents of `<ostream>`, adding the `#include` here when it's unnecessary due to transitive inclusion (e.g. on older MSVC versions) has nearly no cost due to include guards and `#pragma once`.

PR #2526, opened 2020-01-02
Fix Slice op's shape inference logic

hariharans29:
Slice shape inference has logic to copy over `dim_value` and `dim_param` from the input to the output prior to processing each dimension specified by the `axes` input. In case, we cannot proceed with processing a particular dimension, any copied over `dim_param` (if any in the copy step) must be cleared to leave it truly "empty".

Resolve https://github.com/microsoft/onnxruntime/issues/2756

CC: @skottmckay


onnx/defs/tensor/defs.cc
```diff
@@ -645,8 +645,16 @@ ONNX_OPERATOR_SET_SCHEMA(
 
             // input dim value is missing - cannot perform shape inference for
             // this axis
-            if (!input_dim.has_dim_value())
+            if (!input_dim.has_dim_value()) {
+              // Clear any previously propagated dim_param and leave this dimension "empty",
+              // before moving on to the next dimension
+              ctx.getOutputType(0)
```
hariharans29: `clear_dim_param()` checks if `dim_param` is present before attempting to clear it.

onnx/test/shape_inference_test.py
```diff
@@ -493,6 +493,17 @@ def test_slice_with_input_shape(self):  # type: () -> None
...
+    def test_slice_with_input_shape_containing_dim_params(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, (1, 'a', 1)),
+             ('starts', TensorProto.INT64, (3,)),
+             ('ends', TensorProto.INT64, (3,))],
+            [make_node('Slice', ['x', 'starts', 'ends'], ['y'])],
+            [],
+            initializer=[make_tensor('starts', TensorProto.INT64, (3,), (0, 0, 0)),
+                            make_tensor('ends', TensorProto.INT64, (3,), (1, 1, 1))])
+        self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (1, None, 1))])
```
hariharans29: The output with the previous logic would have been (1, 'a', 1) which is incorrect as the output shape need not have the particular dimension to have the same value corresponding to the input dimension's `dim_param`

skottmckay:
LGTM

gramalingam:
The logic looks correct. But it seems a bit redundant: why does it copy the input dim_value/dim_param in the first place, and then clear it or overwrite it? Can't we simply delete the lines the copy the dim_value/dim_param in the earlier loop? Or, did I misunderstand the logic?

PR #2527, opened 2020-01-02
Update CircleCI job to use Python3.6

bddppq:


.circleci/build.sh
```diff
@@ -39,9 +39,11 @@ cp -r "$PWD" "$ONNX_DIR"
 # install ninja to speedup the build
 pip install ninja
 
-# install everything
+# install pytorch
 cd $PYTORCH_DIR
 ./scripts/onnx/install-develop.sh
+# install onnxruntime
+pip install -q --user -i https://test.pypi.org/simple/ ort-nightly==1.0.0.dev1123
```
snnn: Why do you choose a nightly build instead of a release build?
Why do you fix the ort version to this specific number?

.circleci/config.yml
```diff
@@ -69,13 +69,13 @@ linux_default: &linux_default
         docker exec -e CIRCLE_JOB=${CIRCLE_JOB} -u jenkins $(cat .docker_pid) .circleci/test.sh
 
 jobs:
-  py2-gcc7-ubuntu16.04:
+  py3.6-clang7-ubuntu16.04:
```
snnn: So, we are dropping support for python 2.x?

bddppq: For this pull request, we are only changing the CI job that verifies the integration between ONNX and PyTorch and onnxruntime backends to use Python3. This makes senses because both backends have dropped py2 support already.
There are still py2 other CI jobs (running on Travis) that checks ONNX is py2 compatible. Given the developers of the Python language started dropping support for py2 from 01/02/2020, I think there should be discussions in the corresponding SIG to decide on the timeline of dropping py2 support universally in ONNX.

.circleci/build.sh
bddppq: This version was copied from the PyTorch CI. Maybe @spandantiwari @houseroad could answer why this version specifically.

houseroad: Since we keep adding new translation logic into the pytorch master branch, we would like to test it end to end. Some ops are only supported in nightly dev build. So we prefer dev build in the CI.

It will make things simpler if we make these two versions consistent.

bddppq:
waiting for https://github.com/pytorch/pytorch/pull/31827

.circleci/build.sh
snnn: But, does it mean only pytorch team can maintain it and change it?

houseroad: @spandantiwari and his team help us maintain the version.

Issue #2528, opened 2020-01-02
Integer support for more operators


TMVector:
I scanned the operator list for operators that should work with integer tensors but are currently only defined for floats. (Had to work around a couple of these for one model).

**Straighforward operators:**
- [x] Clip (#2450) (PR #2532)
- [ ] GlobalMaxPool
- [ ] Hardmax
- [x] Max (PR https://github.com/onnx/onnx/pull/2608)
- [ ] MaxPool
- [ ] MaxRoiPool
- [ ] MaxUnpool
- [x] Min (PR https://github.com/onnx/onnx/pull/2608)
- [x] Relu (PR #3141)
- [ ] ThresholdedRelu

**Operators that could overflow or require larger intermediate types:**
- [ ] Det
- [ ] Pow

I plan to have a go at the non-pooling simple ones, as I think they should also be trivial to add support for in ONNX Runtime. The pooling ones look like they might be okay but I haven't tried it.

linkerzhang:
Thanks! PRs are welcome to add more necessary types' support for existing ops.

pbamotra:
```
  File "/classifier/src/utils/deployer_lib.py", line 297, in deploy
    self.to_triton_onnx(dataloader, model)
  File "/classifier/src/utils/deployer_lib.py", line 376, in to_triton_onnx
    onnx.checker.check_model(converted_model)
  File "/opt/conda/lib/python3.6/site-packages/onnx/checker.py", line 91, in check_model
    C.check_model(model.SerializeToString())
onnx.onnx_cpp2py_export.checker.ValidationError: Unrecognized attribute: min for operator Clip

==> Context: Bad node spec: input: "2161" output: "2162" name: "Clip_1383" op_type: "Clip" attribute { name: "min" f: 1e-12 type: FLOAT }
```

Onnx version 1.6

TMVector:
Hi @pbamotra that looks like an issue with the PyTorch exporter -- it's producing a `Clip` node which doesn't match the [Clip operator definition in opset 11](https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Clip-11). The operator [changed to take min/max as inputs](https://github.com/onnx/onnx/pull/2096) instead of attributes in that operator set version.

Can you please make sure you are using the most recent version of PyTorch and if it still fails, raise it as an issue on the [PyTorch repo](https://github.com/pytorch/pytorch) (check for existing issues first ofc 🙂). You could also try specifying a different opset version.

[...]

PR #2532, opened 2020-01-04
Add integer support to Clip

### LINKED ISSUE #2450: Clip support for integer tensors
[Clip](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Clip) inputs in constrained to floating points. I have tensors that are indices, and thus `int64` and would like to `Clip` them to valid ranges.
In general, what is the rationale for limiting operators input types? does it really affect binary sizes?

TMVector:
Expand Clip operator to support integer types too. Fixes #2450 and is the first step in #2528.

## Tests
It looks like tests are to be self-contained for inclusion in the docs, but it also looks like data type support is not well tested in general, and adding one for each type would be verbose and noisy. I have modified the `clip_default` node test to loop over the supported dtypes and generate the same test case for each one (the constants are all unsigned and small in that test). That increases the number of tests generated from `clip_default` from 3 to 33. **Feedback would be appreciated.** If we want to expand dtype tests for other operators then it might be worth thinking about how to do that without making all the doc examples noisy.

## ~This PR is not ready to merge -- backend test data needs generating. [help wanted]~
~When I ran `tools/update_doc.sh` all of the `test/data/` ONNX files had small changes in -- do I need a specific version of protobuf or something? I'm using protobuf 3.11.2 installed with conda.~ Turns out it was from the ir/opset version being updated.


linkerzhang:
please fix ci failure - "
+flake8
2326
./onnx/backend/test/case/node/clip.py:68:24: E222 multiple spaces after operator"

TMVector:
Okay, I actually ran some diffs on the text form of the protobufs, and it turns out the differences are all from ir/opset version updates. I will take it that means it's okay to only update tests actually changed by this PR.

What is the convention with regard to backend tests -- just overwrite the tests for old ir/opset versions?

Is it okay to change the test to loop over dtypes like I have? (I'm thinking of the docs mainly).

@linkerzhang feedback on testing and above questions would be really helpful :)

TMVector:
Looks like circleci failing is due to ORT not producing Python 2.7 builds anymore

TMVector:
And the failures now look to be nbval/nbformat related

askhade:
Looks like you are adding a new test for every combination of test case + data type. For example 
test_clip_default_inbounds_float32
test_clip_default_inbounds_int32
test_clip_default_max_float32
test_clip_default_max_int32

This is adding too many files which in my opinion are unnecessary... I would suggest you to pick a few out of these.... I do not believe we need a test case for each and every data type

TMVector:
There is now only one additional test (clip_default_int8) which adds 3 test files.

I would love to see a full backend test suite which tested support for data types, but I guess that's outside the scope of this PR anyway.

linkerzhang:
@TMVector can you fix the conflicts so that the PR can be merged?

linkerzhang:
merging #2608 introduces new conflicts .....

@TMVector  can you update it again?

PR #2544, opened 2020-01-13
Clarify split supports zero length splits

neginraoof:
This case is supported by numpy and PyTorch:

x = np.array([]).astype(np.float32)
np.split(x, [0,0,0])

output will be:
[array([], dtype=float32), array([], dtype=float32), array([], dtype=float32), array([], dtype=float32)]


neginraoof:
cc @postrational 
Could you please help review this PR?

neginraoof:
cc @BowenBao @spandantiwari @houseroad for review

BowenBao:
Should we update SplitToSequence as well?

gramalingam:
Is this just a clarification of the specification of the existing op or it is extending the specification of the existing op? I think it is the first case: if so, I suggest changing the title of the PR (e.g., to "Clarify that split supports zero length splits").

gramalingam:
Just to clarify: any changes to the op spec require bumping the op's version number.

neginraoof:
@gramalingam Thanks! I think this is a clarification for split spec, that's why I didn't bump the version. Please let me know if you think otherwise.

docs/Operators.md
```diff
@@ -17210,7 +17210,7 @@ This version of the operator has been available since version 11 of the default
 <dt><tt>input</tt> : T</dt>
 <dd>The tensor to split</dd>
 <dt><tt>split</tt> (optional) : I</dt>
-<dd>Length of each output. It can be either a scalar(tensor of empty shape), or a 1-D tensor. All values must be positive. </dd>
+<dd>Length of each output. It can be either a scalar(tensor of empty shape), or a 1-D tensor. Values should be >= 0. All values must be positive. </dd>
```
gramalingam: Why both "Values should be >= 0." and "All values must be positive." ? May be better to drop the second sentence, to avoid confusion.

neginraoof:
Thanks @gramalingam. Let me know if we can merge this.

gramalingam:
LGTM. But do you know why some of the CI is failing?

neginraoof:
cc @gramalingam I think this PR is ready for merge. Thanks.

PR #2549, opened 2020-01-15
Fix shape inference of Split-2

shinh:
https://github.com/onnx/onnx/pull/2328/ did not fix the shape inference of the old Split.

for #1735

### DIFF
--- a/onnx/defs/tensor/old.cc
+++ b/onnx/defs/tensor/old.cc
@@ -3591,12 +3591,15 @@ ONNX_OPERATOR_SET_SCHEMA(
           const auto& shape = ctx.getInputType(0)->tensor_type().shape();
           int rank = shape.dim_size();
           int axis = static_cast<int>(getAttribute(ctx, "axis", 0));
-          if (axis < 0 || axis >= rank) {
+          if (axis < -rank || axis >= rank) {
             fail_type_inference(
                 "Invalid value of attribute 'axis'. Rank=",
                 rank,
                 " Value=",
                 axis);
+          }
+          if (axis < 0) {
+            axis += rank;
           }
           const auto& split_dim = shape.dim(axis);
           if (!split_dim.has_dim_value()) {
--- a/onnx/defs/tensor/old.cc
+++ b/onnx/defs/tensor/old.cc
@@ -3587,33 +3587,19 @@ ONNX_OPERATOR_SET_SCHEMA(
           if (!hasNInputShapes(ctx, 1)) {
             return;
           }
-          auto axisAttr = ctx.getAttribute("axis");
-          int axis = axisAttr ? static_cast<int>(axisAttr->i()) : 0;
-          if (axis < 0) {
-            return;
-          }
-          std::vector<int64_t> split;
-          if (!getRepeatedAttribute(ctx, "split", split)) {
-            if (!ctx.getInputType(0)->tensor_type().has_shape()) {
-              return;
-            }
-            const auto& shape = ctx.getInputType(0)->tensor_type().shape();
-            if (axis >= shape.dim_size()) {
-              fail_type_inference("Invalid value of attribute 'axis'");
-            }
-            const auto& splitDim = shape.dim(axis);
-            if (!splitDim.has_dim_value()) {
-              return;
-            }
-            int splitDimValue = static_cast<int>(splitDim.dim_value());
-            int chunkSize =
-                splitDimValue / static_cast<int>(ctx.getNumOutputs());
-            int leftOver = splitDimValue -
-                (chunkSize * static_cast<int>(ctx.getNumOutputs()));
-            for (int i = 0; i < static_cast<int>(ctx.getNumOutputs()); i++) {
-              split.push_back(i < leftOver ? chunkSize + 1 : chunkSize);
-            }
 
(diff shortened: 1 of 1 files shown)


onnx/defs/tensor/utils.cc
```diff
@@ -167,4 +167,83 @@ void resizeShapeInference_opset7_to_10(InferenceContext& ctx) {
...
+  int rank = shape.dim_size();
+  int axis = static_cast<int>(getAttribute(ctx, "axis", 0));
+  if (axis < (allow_negative_axis ? -rank : 0) || axis >= rank) {
+    fail_type_inference(
+        "Invalid value of attribute 'axis'. Rank=",
+        rank,
+        " Value=",
+        axis);
+  }
+  if (allow_negative_axis && axis < 0) {
```
linkerzhang: no need checking allow_negative_axis here.

shinh: Done, sorry for the large latency!

jcwchen:
Hi @shinh,
Thank you so much for proposing this! I think it can help fix the shape inference bug for Split... Could you resolve the merge conflict? Thanks.

shinh:
Unlike previous approach which factored out a function to share code among Split-11 and Split-2, this time, I just copied the shape inference of Split-11 to Split-2. This way seems to be more consistent with other operator definitions.

At first, I tried to remove the support for negative axis from the one of Split-2 since Split-2 did not mention negative axis. However, it made some tests fail so I decided to handle negative axis even in Split-2.

This is quite an old PR. I don't have the model fixed by this PR anymore. If this patch does not look good and you still want to fix Split-2, I'd suggest you create another PR. I'll close this PR then.

jcwchen (approved):
LGTM. I have confirmed it can fix shape inference bug with Split-2. For instance: https://github.com/onnx/models/blob/master/vision/classification/shufflenet/model/shufflenet-v2-10.onnx

jcwchen:
@linkerzhang Could you help to merge this? Thank you.

onnx/defs/tensor/old.cc
```diff
@@ -3587,42 +3587,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            int chunkSize =
-                splitDimValue / static_cast<int>(ctx.getNumOutputs());
-            int leftOver = splitDimValue -
-                (chunkSize * static_cast<int>(ctx.getNumOutputs()));
-            for (int i = 0; i < static_cast<int>(ctx.getNumOutputs()); i++) {
-              split.push_back(i < leftOver ? chunkSize + 1 : chunkSize);
-            }
-
+          const auto& split_dim = shape.dim(axis);
+          if (!split_dim.has_dim_value()) {
```
gramalingam, line 3605: A minor point: if the "split" attribute is specified, then we can infer the output dimensions, even if the input-shape does not have a dim-value for the split-axis, isn't that right?

gramalingam:
@shinh : thanks for updating the fix. Regarding
```
At first, I tried to remove the support for negative axis from the one of Split-2 since Split-2
did not mention negative axis. However, it made some tests fail so I decided to handle negative
axis even in Split-2.
```
that is troubling. I guess there is some mismatch between the spec (which doesn't mention negative
axes) and the test-cases (which use negative axes)? Were there many test-cases that failed? I wonder if we should fix the test-cases. If it is complicated, it is worth at least adding a comment to the shape-inference code to mention this reason, to help future maintenance.

jcwchen:
@gramalingam Makes sense. https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/pytorch-converted/test_GLU This model consists the split node with negative which might break the test previously.

onnx/defs/tensor/old.cc
```diff
@@ -3587,42 +3587,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          int rank = shape.dim_size();
+          int axis = static_cast<int>(getAttribute(ctx, "axis", 0));
+          if (axis < -rank || axis >= rank) {
+            fail_type_inference(
+                "Invalid value of attribute 'axis'. Rank=",
+                rank,
+                " Value=",
+                axis);
+          }
           if (axis < 0) {
```
jcwchen, line 3601: Could you add some comments here for explanation? Something like:
```suggestion
          // Previously Split-2 does not mention how to deal with negative axis
          // However, there is an existing test onnx/backend/test/data/pytorch-converted/test_GLU
          // using Split-2 with negative axis and it is hard to be regenerated. 
          // To compromise, handle negative axis for Split-2 here. 
          if (axis < 0) {
```

jcwchen (changes requested):

onnx/defs/tensor/old.cc
shinh: I agree. Do you want to do this in this PR? I'm afraid it'd take time for me to update this PR if we do this in this PR.

gramalingam: I think it's fine to do that as a separate PR.

gramalingam (approved):

gramalingam:
LGTM. I am afraid it needs your signoff. We can merge it once we have the signoff.

shinh:
Done!

jcwchen (approved):

PR #2551, opened 2020-01-15
Add NegativeLogLikelihood(NllLoss) op

liqunfu:
add NegativeLogLikelihood(NllLoss) op. See the op's doc for more details. The op is a function op so it's default implementation is given. One thing special with NllLoss op is that it's computation graph depends on it's optional weight input and reduction attribute. For this, the op's function body (FunctionProto) is provided at load time according to the op's NodeProto instance. 

A comparison between the proposed operator with its Pytorch equivalence can be found [here](https://github.com/onnx/onnx/pull/2551#discussion_r374969916).


onnx/defs/schema.h
```diff
@@ -25,6 +25,8 @@
 #include "onnx/onnx-operators_pb.h"
 namespace ONNX_NAMESPACE {
 
+using FunctionBodyQueryFunction = std::function<bool(InferenceContext&)>;
```
gramalingam: Hi, I think it will be better if we use a single "std::function<std::vector<NodeProto>(InferenceContext&)>" instead of a vector of pairs of (condition, body). In particular, this will be useful when we want to use the value of an attribute to create a specific ONNX node. Would it be possible to generalize this way?

gramalingam: Sorry, I think some confusion happened in the HTML formatting with special characters. I mean a function that takes an input of type "InferenceContext&" and returns the body (a vector<NodeProto>).

onnx/defs/schema.cc
```diff
@@ -724,19 +742,33 @@ OpSchema& OpSchema::FillUsing(const std::function<void(OpSchema&)>& populator) {
...
+}
+
+void OpSchema::BuildFunction(FunctionProto& function_body) {
+  function_body.set_name(this->name_);
+  function_body.set_doc_string(this->doc_);
+  function_body.set_since_version(this->since_version_);
+  function_body.set_status(OperatorStatus(1 - (int)this->support_));
   for (auto& i : inputs_) {
-    function_body_.add_input(i.GetName());
+    if (FilterFunctionBodyInputByName(i.GetName(), function_body))
```
gramalingam: I am not sure if this filtering is a good idea, because the input signature of the function should match parameters passed at call-site. This filtering could break things. It seems okay if a function body has an unused input.

lara-hdr:
NLLLoss in PyTorch has an additional attribute ignore_index, which when specified ignores the specified target value (and the loss is then averaged over the non-ignore targets only).
https://pytorch.org/docs/stable/nn.html?highlight=nllloss#torch.nn.NLLLoss

I am thinking of a scenario where we have a NLLLoss with mean reduction, with an ignore_index. 

In that case, with the current introduced ONNX::NLLLoss, we have to:
1 - Create a ONNX::NLLLoss with no reduction.
2 - Identify the indices in 'target' that are equal to the ignore_index.
3 - Set the values of the identified indices to 0 in the output of ONNX::NLLLoss.
4 - Create a ONNX::ReduceSum on the last output.
5 - Divide the reduced sum by (N - Number_of_ignored_indices), to compute the mean over the non-ignored targets.

Given all these steps, would it be preferable to add ignore_index as a parameter?
LMK if you think of a better way to support this.

onnx/defs/math/defs.cc
```diff
@@ -1665,4 +1666,299 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+)DOC";
+
+TensorProto ToDimensionOneTensor(int32_t value) {
+  auto t = ToTensor(std::vector<int32_t>({value}));
+  t.add_dims(1);
+  return t;
+}
+
+ONNX_OPERATOR_SET_SCHEMA(
+    NllLoss,
```
wschin: Maybe
```suggestion
    NegativeLogLikekihood,
```
?

wschin:
Does it cover both of Tensorflow and Pytorch scenarios?

liqunfu:
@postrational @houseroad , please review this PR. thanks!

docs/Changelog.md
```diff
@@ -14102,6 +14102,127 @@ This version of the operator has been available since version 12 of the default
 <dd>Constrain index tensor to int64</dd>
 </dl>
 
+### <a name="NllLoss-12"></a>**NllLoss-12**</a>
```
TMVector: The generated docs don't say this is operator has a function, because [gen_doc.py](https://github.com/onnx/onnx/blob/master/onnx/defs/gen_doc.py#L204-L207) only checks for `has_function`. It also isn't listed in the table of contents as one of the operators with a function registered.

docs/Changelog.md
```diff
@@ -14102,6 +14102,127 @@ This version of the operator has been available since version 12 of the default
...
+
+  A NllLoss op computes (weighted) negative log likelihood loss.
```
```suggestion
  A NegativeLogLikelihood op computes (weighted) negative log likelihood.
```

docs/Changelog.md
```diff
@@ -14102,6 +14102,127 @@ This version of the operator has been available since version 12 of the default
...
+  Its "input" tensor has the shape of (N, C, d1, d2, ..., dk) where k >= 0. The "input" tensor contains log-probabilities for samples being in a class of [0, C).
```
```suggestion
  Its "input" tensor has the shape of (N, C, d0, d1, ..., d_{k-1}) where k >= 0. The "input" tensor contains log-probabilities for samples being in a class of [0, C).
```
but the `k` here starts with 0.

wschin: What is the log-probabilities for a sample? Is it `input[i, :, k, l, m, ...]`?

docs/Changelog.md
```diff
@@ -14102,6 +14102,127 @@ This version of the operator has been available since version 12 of the default
...
+  The op's "target" input tensor has the shape of (N, d1, d2, ..., dk). It contains classifications (one of C classes) for N x d1 x d2 x ... x dk samples.
```
```suggestion
  The operator's "target" input tensor has the shape of (N, d1, d2, ..., dk). It contains classifications (one of C classes) for N x d1 x d2 x ... x dk samples. 
```
Let's use operator when we can.

wschin: cla
```suggestion
  The op's "target" input tensor has the shape of (N, d1, d2, ..., dk). It encodes class labels (one of C classes) for N x d1 x d2 x ... x dk samples. 
```

docs/Changelog.md
```diff
@@ -14102,6 +14102,127 @@ This version of the operator has been available since version 12 of the default
 <dd>Constrain index tensor to int64</dd>
 </dl>
 
+### <a name="NllLoss-12"></a>**NllLoss-12**</a>
+
+  A NllLoss op computes (weighted) negative log likelihood loss.
+  Its "input" tensor has the shape of (N, C, d1, d2, ..., dk) where k >= 0. The "input" tensor contains log-probabilities for samples being in a class of [0, C).
+  The op's "target" input tensor has the shape of (N, d1, d2, ..., dk). It contains classifications (one of C classes) for N x d1 x d2 x ... x dk samples. 
+  The loss for a sample at n, d_1, d_2,...d_k being classified as class c = target[n][d_1][d_2]...[d_k] is computed as:
```
```suggestion
  The loss for a sample at n, d_1, d_2,...d_k being classified as class c = target[n][d_1][d_2]...[d_k] is computed as:
```
`a sample at n, d_1, d_2,...d_k` is not a formal description. Do you mean `input[n, d_1, d_2, ..., d_k]`?

docs/Changelog.md
```diff
@@ -14102,6 +14102,127 @@ This version of the operator has been available since version 12 of the default
...
+  
+      loss[n][d_1][d_2]...[d_k] = -input[n][c][d_1][d_2]...[d_k].
+  
+  When an optional "weight" is provided, the sample loss is calculated as:
+  
+      loss[n][d_1][d_2]...[d_k] = -input[n][c][d_1][d_2]...[d_k] * weight[c].
+  
+  If "reduction" attribute is set to "none", the op's output will be the above loss with shape (N, d1, d2, ..., dk).
```
```suggestion
  If "reduction" attribute is set to "none", the operator's output will be the above loss with shape (N, d1, d2, ..., dk).
```

onnx/backend/test/case/node/nllloss.py
```diff
@@ -0,0 +1,191 @@
...
+from __future__ import unicode_literals
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+def compute_nll_loss(input, target, weight=None, reduction='mean'):  # type: ignore
```
wschin:
```suggestion
def compute_nll(input, target, weight=None, reduction='mean'):  # type: ignore
```

wschin: Does it produce the same outputs as Pytorch's NLLLoss?

wschin:
Are there some shape inference tests?

docs/Changelog.md
```diff
@@ -14102,6 +14102,127 @@ This version of the operator has been available since version 12 of the default
...
+
+#### Version
+
+This version of the operator has been available since version 12 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>reduction</tt> : string (default is mean)</dt>
+<dd>Type of reduction to apply to loss: none, sum, mean(default). 'none': the output is the loss for each sample in the batch.'sum': the output will be summed. 'mean': the sum of the output will be divided by the batch_size.</dd>
```
```suggestion
<dd>Type of reduction to apply to loss: none, sum, mean (default). 'none': the output is the loss for each sample in the batch. 'sum': the output will be summed into a scalar. 'mean': similar to the sum mode, but further divided by the batch_size.</dd>
```
Also, what is `batch_size`? I can't find its definition in the doc.

liqunfu: added shape tests

onnx/backend/test/case/node/nllloss.py
liqunfu: # this is the test script to verify equivalency between pytorch and nllloss.py test data. 
def compute_nll_loss(input, target, weight=None, reduction='mean'):  # type: ignore
    ''' Compute nll_loss '''
    input_shape = input.shape

    # GatherElement(-input, target)
    if len(input_shape) == 2:
        N, C = input_shape
        for i in range(N):
            neg_gather_element_input[i] = -input[i][target[i]]
    else:
        N, C, dim1, dim2 = input_shape
        neg_gather_element_input = np.zeros((N, dim1, dim2), dtype=np.float32)
        for i in range(N):
            for d1 in range(dim1):
                for d2 in range(dim2):
                    neg_gather_element_input[i][d1][d2] = -input[i][target[i][d1][d2]][d1][d2]

    if weight is not None:
        # Gather(input=weight, index=target)
        gather_weight = np.take(weight, target)

        loss = gather_weight * loss
        if reduction == 'mean':
            return loss.sum() / gather_weight.sum()

    if reduction == 'none':
        return loss
    elif reduction == 'mean':
        return np.mean(loss, keepdims=False)
    elif reduction == 'sum':
        return np.sum(loss, keepdims=False)

import numpy as np
from itertools import product
import torch
import torch.nn as nn
from numpy.testing import assert_allclose, assert_array_equal

# options 
reduction_options=['none', 'mean', 'sum']
N, C, height, width = 2, 3, 4, 5

sample_shape_option = [(), (height, width)]
weight_options=[None, np.random.randn(C).astype(np.float)]

for config in list(product(sample_shape_option, reduction_options, weight_options)):
    sample_shape, reduction, weight = config

    if weight is None:
        weight_pt = None
    else:
        weight_pt = torch.from_numpy(weight).float()

    loss = nn.NLLLoss(weight=weight_pt, reduction=reduction)
    log_softmax_ = nn.LogSoftmax(dim=1)

    data_ = torch.randn(N, C, *sample_shape)
    input_ = log_softmax_(data_)
    target = torch.empty((N, *sample_shape), dtype=torch.long).random_(0, C)

    input_.shape, target.shape

    output_pt = loss(input_, target)
    output_pt.shape, output_pt

    my_loss = compute_nll_loss(input_, target, weight=weight, reduction=reduction)
    assert_allclose(my_loss, output_pt.numpy(), rtol=1e-05)
    my_loss - output_pt.numpy()
    print ("pass: ", sample_shape, reduction, weight, my_loss - output_pt.numpy())

onnx/defs/math/defs.cc
```diff
@@ -1666,6 +1667,301 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            std::string("mean"))
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input, weight, and output types to floating-point tensors.")
+        .TypeConstraint(
+            "Tind",
+            {"tensor(int32)", "tensor(int64)"},
+            "Constrain target to integer types")
+        .AddQueriedFunctionBody([](FunctionBodyQueryContext& ctx) { // no weight, reduction is "none"
```
SherlockNoMad: >AddQueriedFunctionBody [](start = 9, length = 22)

general speaking, I feel that this approach is not extensible... 

In this op, there is 6 combinations....
What about other ops if the options are combinatorial.... 

Is this the designed way to handle conditional function body?

wschin: @gramalingam, @linkerzhang might have some opinions.

docs/Changelog.md
```diff
@@ -14102,6 +14102,132 @@ This version of the operator has been available since version 12 of the default
...
+
+#### Version
+
+This version of the operator has been available since version 12 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>reduction</tt> : string (default is mean)</dt>
+<dd>Type of reduction to apply to loss: none, sum, mean(default). 'none': the output is the loss for each sample.'sum': the output will be summed. 'mean': the sum of the output will be divided by the sum of applied weights.</dd>
```
```suggestion
<dd>Type of reduction to apply to loss: none, sum, mean (default). 'none': the output is the loss for each sample.'sum': the output will be summed. 'mean': the sum of the output will be divided by the sum of applied weights.</dd>
```

docs/Changelog.md
```diff
@@ -14102,6 +14102,132 @@ This version of the operator has been available since version 12 of the default
 <dd>Constrain index tensor to int64</dd>
 </dl>
 
+### <a name="NegativeLogLikelihoodLoss-12"></a>**NegativeLogLikelihoodLoss-12**</a>
+
+  A NegativeLogLikelihoodLoss operator computes (weighted) negative log likelihood loss.
+  Its "input" tensor has the shape of (N, C, d1, d2, ..., dk) where k >= 0.
+  The "input" tensor contains log-probabilities for input[n, :, d_1, d_2,...d_k] being in a class of [0, C).
```
```suggestion
  The "input" tensor contains log-probabilities for input[n, :, d_1, d_2,..., d_k] being in a class of [0, C).
```

docs/Changelog.md
```diff
@@ -14102,6 +14102,132 @@ This version of the operator has been available since version 12 of the default
...
+  The operator's "target" input tensor has the shape of (N, d1, d2, ..., dk). It encode class labels (one of C classes) for N x d1 x d2 x ... x dk samples.
+  The loss value for input[n, :, d_1, d_2,...d_k] being classified as class c = target[n][d_1][d_2]...[d_k] is computed as:
```
```suggestion
  The loss value for input[n, :, d_1, d_2,..., d_k] being classified as class c = target[n][d_1][d_2]...[d_k] is computed as:
```
Copy-and-paste? :)

docs/Changelog.md
```diff
@@ -14102,6 +14102,132 @@ This version of the operator has been available since version 12 of the default
 <dd>Constrain index tensor to int64</dd>
 </dl>
 
+### <a name="NegativeLogLikelihoodLoss-12"></a>**NegativeLogLikelihoodLoss-12**</a>
+
+  A NegativeLogLikelihoodLoss operator computes (weighted) negative log likelihood loss.
+  Its "input" tensor has the shape of (N, C, d1, d2, ..., dk) where k >= 0.
+  The "input" tensor contains log-probabilities for input[n, :, d_1, d_2,...d_k] being in a class of [0, C).
+  The operator's "target" input tensor has the shape of (N, d1, d2, ..., dk). It encode class labels (one of C classes) for N x d1 x d2 x ... x dk samples.
```
```suggestion
  The operator's "target" input tensor has the shape of (N, d1, d2, ..., dk). It encodes class labels (one of C classes) for N x d1 x d2 x ... x dk samples.
```

onnx/backend/test/case/node/nllloss.py
wschin: Yeah, compute_negative_log_likelihood_loss is better.

onnx/defs/math/defs.cc
gramalingam: My suggestion (which is up earlier) is to register a single function that takes the Context as input and returns the function-body.

docs/Changelog.md
liqunfu: fixed

onnx/defs/math/defs.cc
liqunfu: If the options are combinatorial, then there will be combinatorial number of graph bodies. I do not see how it can be reduced in the current framework. Maybe it will be solved with MLIR dialect?

gramalingam: @liqunfu : if we define a single function that takes the Context as input and returns the function-body, it will look like below:
```
   NodeSet nodes;
   nodes.add(common_prefix_nodes);
   if (condition1(context)) nodes.add(variant1_A); else nodes.add(variant1_B);
   nodes.add(common_middle_nodes);
   if (condition2(context)) nodes.add(variant2_A); else nodes.add(variant2_B);
   nodes.add(common_suffix);
```
If this is converted to the existing approach, we will have to map every execution-path in the above code to a different (condition,body) pair, leading to the combinatorial explosion that @SherlockNoMad is talking about.

onnx/defs/schema.cc
```diff
@@ -724,19 +742,33 @@ OpSchema& OpSchema::FillUsing(const std::function<void(OpSchema&)>& populator) {
...
```
TMVector: ONNX compilers and runtimes are required to correctly handle operators with [optional inputs and outputs](https://github.com/onnx/onnx/blob/master/docs/IR.md#optional-inputs-and-outputs). Although functions are somewhat underdocumented, I think it is more consistent that runtimes should be required (and able) to correctly handle function bodies which don't use those optional inputs/outputs. IMO this would be easier to implement too.

Also, this filtering seems like it could be a big source of bugs in the future, and it already has at least two that I can see: it will remove inputs that aren't optional if they aren't used by the function body, and it will remove inputs that are used within a subgraph but not by a node in the top-level function graph.

---

For those reasons, I would strongly recommend this be changed to not filter inputs.

onnx/defs/schema.h
```diff
@@ -25,6 +25,40 @@
...
 namespace ONNX_NAMESPACE {
 
+struct FunctionBodyQueryContext {
+  virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
+  virtual size_t getNumInputs() const = 0;
+  virtual size_t getNumOutputs() const = 0;
+  virtual ~FunctionBodyQueryContext() {}
+};
+
+struct FunctionBodyQueryContextImpl : public FunctionBodyQueryContext {
```
linkerzhang: are you going to have different implementation in ORT?

onnx/defs/schema.h
```diff
@@ -25,6 +25,40 @@
 #include "onnx/onnx-operators_pb.h"
 namespace ONNX_NAMESPACE {
 
+struct FunctionBodyQueryContext {
+  virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
+  virtual size_t getNumInputs() const = 0;
```
linkerzhang: do you allow that an optional input exists but the other optional input before it does not exist?

onnx/defs/schema.h
```diff
@@ -25,6 +25,40 @@
...
+  virtual size_t getNumOutputs() const = 0;
```
linkerzhang: is this needed?

onnx/defs/math/defs.cc
liqunfu: @gramalingam this sounds very good. Let me give it a try.

onnx/defs/schema.cc
liqunfu: discussed with @gramalingam offline. The op signature need to be kept otherwise the provider kernel will fail. I will update this according to reviewers' comment.

onnx/defs/schema.cc
liqunfu: totally understand and agree now.

onnx/defs/schema.h
liqunfu: Yes. with @gramalingam and @TMVector 's comment implemented, the functionProto will keep the amount of inputs (required and optional). Thus optional input may be in any combination.

onnx/defs/schema.h
liqunfu: Not in this nllloss case. But I would assume cases where the function graph will be built depending on its output option.

onnx/defs/schema.h
liqunfu: yes.

onnx/defs/math/defs.cc
liqunfu: updated the PR according to @gramalingam 's suggestion.

liqunfu:
@TMVector , thank you so much for your review and suggestions. It did help to avoid introducing a bug by this PR. Will you please take time to check if this PR is at its current state satisfactory? If so, would you sign of it? Thanks.

onnx/defs/schema.h
gramalingam: I think @linkerzhang may be suggesting that we enrich this interface by explicitly adding interface method like:
```
   virtual bool hasInput(size_t index);
```
I think that is useful because intermediate optional inputs may be absent. Similarly, I think we will need
```
```
to allow type-dependent specialization.

(However, I am okay if we add the above as a separate PR also, to get this PR in first.)

onnx/test/shape_inference_test.py
```diff
@@ -2850,6 +2850,104 @@ def test_einsum_incorrect_num_inputs(self):  # type: () -> None
             [])
         self.assertRaises(checker.ValidationError, self._inferred, graph)
 
+    def test_negative_log_likehood_shape_is_NCdd(self):  # type: () -> None
+        N, C = 3, 4
+        graph = self._make_graph(
+            [('input', TensorProto.FLOAT, (N, C)),
+             ('target', TensorProto.INT64, (N,))],
+            [make_node('NegativeLogLikelihoodLoss', ['input', 'target'], ['loss'], reduction='none')],
```
TMVector: Optional inputs can also be omitted by providing an empty string. I'm not sure if that case would currently be correctly handled or not?
```python
> node = onnx.helper.make_node(op_type="Nll", inputs=["a", "b", ""], outputs=["z"])
> len(node.input)
3
```

onnx/defs/math/defs.cc
```diff
@@ -1666,6 +1667,240 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  t.add_dims(1);
+  return t;
+}
+
+bool BuildContextDependentFunctionBody(const FunctionBodyQueryContext& ctx, const OpSchema& schema, FunctionProto& functionProto) {
+  std::vector<FunctionBodyHelper::NodeDef> body;
+  body.push_back({{"input_shape"}, "Shape", {"input"}});
+  body.push_back({{"zeros"}, "ConstantOfShape", {"input_shape"}, {MakeAttribute("value", ToDimensionOneTensor(0))}});
+  body.push_back({{"expanded_target"}, "Unsqueeze", {"target"}, {MakeAttribute("axes", std::vector<int64_t>({1}))}});
+  body.push_back({{"expanded_broadcasted_target"}, "Add", {"expanded_target", "zeros"}});
```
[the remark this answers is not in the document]
liqunfu: your first option works. Thanks!

onnx/test/shape_inference_test.py
liqunfu: replaced numInput/numOutput with hasInput(i)/hasOutput(i).

onnx/defs/schema.h
linkerzhang: YES, Rama is correct!

If allowing the case that I mentioned, then current interface is not enough.

PR #2552, opened 2020-01-16
Update Reshape op: Add 'allowzero' attribute

jignparm:
This PR retains the behavior of old Reshape op, but introduces 'allowzero' attribute, which (if set) allows tensors to be reshaped with zeros in target shape, similar to NumPy.

Previous behavior was :
* **input**
data =  tensor of shape [1, 0]
shape = [0, 1]
* **output**
result  = tensor of shape [1, 1]

The previous behavior is unchanged, unless 'allowzero' is set. If set, the behavior is as below:
* **input**
data =  tensor of shape [1, 0]
shape = [0, 1]
allowzero = 1
* **output**
result  = tensor of shape [0, 1]


jignparm:
N.B. PR  #2552 and  PR #2553 deal address the same issue in 2 different ways. Only **one** of these PRs should be merged.

onnx/defs/tensor/defs.cc
```diff
@@ -86,19 +86,28 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 ONNX_OPERATOR_SET_SCHEMA(
     Reshape,
-    5,
+    12,
     OpSchema()
-        .SetDoc(Reshape_ver5_doc)
+        .SetDoc(Reshape_ver12_doc)
+        .Attr(
+            "allowzero",
+            "(Optional) By default, when any output dimension is equal to zero "
```
gramalingam: Change "output dimension" to "value in the 'shape' input"

onnx/defs/tensor/defs.cc
```diff
@@ -86,19 +86,28 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "the desired dimension value is copied from the input tensor dynamically. "
```
gramalingam: change "desired" to "corresponding"

onnx/defs/tensor/defs.cc
```diff
@@ -86,19 +86,28 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "allowzero=1 indicates that if any output dimension value is set to zero, "
```
gramalingam: change "output dimension" to "value in the 'shape' input"

gramalingam:
LGTM. I prefer this to the other option (2553), since this maintains backward compatibility. Any opinions @linkerzhang @houseroad ?

onnx/defs/tensor/defs.cc
jignparm: Done!

onnx/defs/tensor/defs.cc
jignparm: Done!

onnx/defs/tensor/defs.cc
jignparm: Done!

neginraoof:
Can we keep the same behavior using -1 dim value, and don't add "allow zero"?

jignparm:
> Can we keep the same behavior using -1 dim value, and don't add "allow zero"?

Currently '0' means 'copy from input'. Are you suggesting that '-1' should be updated so that it means 'copy from input'? This seems like an odd change -- most frameworks use '-1' to mean 'infer from remaining dimensions'. 

 
There is also PR #2553 that removes the special meaning of zero (so special values can start from -1 downwards). Perhaps this is what you meant instead?

BowenBao:
ping: any updates on this? I think after several releases of ONNX, the version needs to be opset 14.

gramalingam:
Closing this as this was recreated as https://github.com/onnx/onnx/pull/3113 .

PR #2553, opened 2020-01-16
Update Reshape op: Remove special meaning of 0 dims

jignparm:
Make Reshape op work like NumPy. Zero dimension does not mean 'copy from input'.

Previous versions of Reshape treated 'zero' dim values as special flag values, to be imputed from the input tensor. This PR eliminates that behavior, and treats 'zero' as 'zero', as shown in the example below.

Previous behavior was :
* **input**
data =  tensor of shape [1, 0]
shape = [0, 1]
* **output**
result  = tensor of shape [1, 1]

This PR updates Reshape to behave as below:
* **input**
data =  tensor of shape [1, 0]
shape = [0, 1]
* **output**
result  = tensor of shape [0, 1]


jignparm:
N.B. PR  #2552 and  PR #2553 deal address the same issue in 2 different ways. Only **one** of these PRs should be merged.

diyessi:
PaddlePaddle seems to be the source of the 0 behavior. Using a value that cannot appear in a shape, such as -2, would still allow the copy behavior, but not be confusable with a valid length in a shape. The PaddlePaddle shape is specified with an attribute, so it will have a constant value and it should be easy to convert a 0 to a -2 in an ONNX conversion.

gramalingam:
Closing this. Please see https://github.com/onnx/onnx/pull/3113 for the follow-up.

PR #2555, opened 2020-01-16
Update ManagingExperimentalOps.md to include deprecated operators table

fdwr:
The 1.6 release missed adding a row for opset 11 to the table.

Many experimental operators were removed from ONNX, including dozens used in client models for published applications, which now cause warnings to be written to the console in check_node. e.g. "Warning: Affine was a removed experimental op. In the future, we may directly reject this operator...". These warnings provide no guidance on what to do though, and they're confusing to the user given frameworks like ONNX Runtime continue to support these operators anyway for backwards compatibility. Since this is related to versioning, I'm proposing to document the guidance here (note we can't it in Operators.md because these operators don't exist anymore, and we can't include it in ChangeLog.md either because all records of these operators were removed).

Part of issue #2239.

### DIFF
--- a/docs/ManagingExperimentalOps.md
+++ b/docs/ManagingExperimentalOps.md
@@ -1,4 +1,22 @@
-## [Deprecated - as of v1.5 experimental ops are no longer supported]
+## Deprecated Experimental Operators
+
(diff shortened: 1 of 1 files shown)


fdwr:
FYI @prasanthpul.

PR #2565, opened 2020-01-22
Fix circleci build and drop unsupported Windows builds

wschin:
For circleci, `nijna` cannot be installed with `pip>=20`, so we switch to `pip=19`. Anaconda doesn't have packages for `protobuf=3.9.2`  when `python=3.5`, so we drop python 3.5 Windows builds. I don't know if this is a long-term solution, but at least I can unblock myself with them.

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -26,7 +26,7 @@ if [[ "${CIRCLE_JOB}" =~ py((2|3)\.?[0-9]?\.?[0-9]?) ]]; then
 fi
 $PYTHON -m virtualenv "$VENV_DIR"
 source "$VENV_DIR/bin/activate"
-pip install -U pip=19 setuptools
+pip install -U pip==19 setuptools
 
 # setup onnx as the submodule of pytorch
 PYTORCH_DIR=/tmp/pytorch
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -38,26 +33,14 @@ before_build:
...
   (
   set "PATH=%PATH%;%CONDA_PREFIX%\Library\bin"
   )
 
-  IF "%_prefix%"=="C:\Miniconda35"
-  (
-  pip install --quiet "ipython<7.10"
-  )
-
 - cmd: pip install --quiet pytest nbval numpy
```
vinitra: numpy can be removed from this command? It might be double installed.

.circleci/build.sh
```diff
@@ -26,7 +26,7 @@ if [[ "${CIRCLE_JOB}" =~ py((2|3)\.?[0-9]?\.?[0-9]?) ]]; then
 fi
 $PYTHON -m virtualenv "$VENV_DIR"
 source "$VENV_DIR/bin/activate"
-pip install -U pip setuptools
+pip install -U pip=19 setuptools
```
vinitra: pip==19

appveyor.yml
wschin: Removed at the protobuf installation line. I feel protobuf itself is difficult enough to occupy one line.

gramalingam:
Hope this works! Thanks for the fix.

neginraoof:
@wschin Can we merge this please?

PR #2567, opened 2020-01-23
Clarify spec for constant of shape with dim_n = 0

neginraoof:
Both Numpy and PyTorch do support constructing a tensor of shape zero.
This PR clarifies the spec for ONNX constantofshape.


docs/Changelog.md
```diff
@@ -8184,7 +8184,7 @@ This version of the operator has been available since version 9 of the default O
 
 <dl>
 <dt><tt>input</tt> : T1</dt>
-<dd>1D tensor. The shape of the expected output tensor. If empty tensor is given, the output would be a scalar.</dd>
+<dd>1D tensor. The shape of the expected output tensor. If empty tensor is given, the output would be a scalar. All values must be >= 0.</dd>
```
wschin: Must be `>0`? I believe the `=` case corresponds to a scalar and we already have `empty tensor` for that.

neginraoof: Yes, but we can a multidimensional tensor with one dim = 0.
Like:
numpy.full((2, 0), 4)
numpy.random.randn(2, 0)
or
torch.randn([0, 2, 3])

neginraoof:
@wschin 
An example model that needs this is a torchvision model.
In case of no detection, the model creates an empty tensor of size (0, B, ...)

docs/Changelog.md
gramalingam: Just to clarify: the output is expected to be an empty tensor, so the constant-value used is immaterial. Is that right?

neginraoof: Yes, that's correct.

neginraoof:
cc @postrational @houseroad. Could you please help review this PR? Thanks.

PR #2568, opened 2020-01-23
Update Dropout and  BatchNorm to be Training Friendly

lara-hdr:
Initial PR: https://github.com/onnx/onnx/pull/1887
This PR updates Dropout and BatchNormalization to be training friendly.

For Dropout: 
The "ratio" is now an input rather than an attribute of the operator, and a new attribute "seed" is introduced.

For BatchNormalization: 
The operator already have support for the mean and var as inputs/outputs and saved_mean and saved_var as output of the model.  The input mean and var will be the running values if we are in training mode, otherwise they will be the estimated values. The optional outputs mean/var/saved_mean/saved_var will only be used in training mode.
A new optional input "training_mode" is introduced and defaults to False since in most cases we would export ONNX models in inference mode; this is an input rather than an attribute since we need to modify it during the runtime.
The new attribute "training_mode" would allow to state explicitly if the operator is in training or inference mode. However the backend engine running the ONNX model could infer this information from the ONNX model, so we could potentially remove this attribute completely, and let the engine decide how to compute the output of this operator.


lara-hdr:
@SherlockNoMad @wschin for review

onnx/defs/nn/defs.cc
```diff
@@ -1534,37 +1534,49 @@ ONNX_OPERATOR_SET_SCHEMA(
             "var",
             "running (training) or estimated (testing) variance tensor of shape (C).",
             "T")
+        .Input(
+            5,
+            "training_mode",
+            "If set to true, run spatial batch normalization in training mode, default is false.",
+            "T1",
+            OpSchema::Optional)
```
wschin: If we make it an attribute, we can do static shape inference for both of training and inference. If we make it an input, I am not sure how ONNX shape inference would work.

onnx/defs/nn/defs.cc
```diff
@@ -1534,37 +1534,49 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         .Output(0, "Y", "The output tensor of the same shape as X", "T")
         .Output(
             1,
             "mean",
-            "The running mean after the BatchNormalization operator.",
+            "The running mean after the BatchNormalization operator."
+            "This output is populated only when training_mode is set to true.",
             "T",
```
wschin: Those outputs (mean, var, saved_mean, saved_var) cannot be consumed by any other operators. Otherwise, SSA assumption will be broken. Could we make it explicitly here by saying that `this output cannot be an input of any other operator`.

onnx/defs/nn/defs.cc
```diff
@@ -1534,37 +1534,49 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             OpSchema::Optional)
         .TypeConstraint(
             "T",
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to float tensors.")
+        .TypeConstraint(
+            "T1",
+            {"tensor(bool)"},
+            "Constrain input 'training_mode' types to boolean tensors.")
```
wschin: If we really make `mode` an input, we need to check if it's a Boolean scalar in the shape inference below.

onnx/defs/nn/defs.cc
```diff
@@ -1660,25 +1672,29 @@ the training phase, so during testing nothing needs to be done.
...
-        .SetDoc(Dropout_ver10_doc + GenerateOptionalArgumentsDoc())
-        .Attr(
-            "ratio",
-            "The ratio of random dropout",
-            AttributeProto::FLOAT,
-            0.5f)
+        .SetDoc(Dropout_ver12_doc + GenerateOptionalArgumentsDoc())
+        .Attr("seed", "(Optional) Seed to the random generator, if not specified we will auto generate one.", AttributeProto::INT, OPTIONAL)
         .Input(0, "data", "The input data as Tensor.", "T")
+        .Input(1, "ratio", "The ratio of random dropout, with value in [0, 1). If this input was not set, "
```
wschin: >ratio [](start = 19, length = 5)

Can we check if this is an scalar inside the shape inference below?

onnx/defs/nn/defs.cc
lara-hdr: If we make it an attribute, then training_mode would be fixed once the model is exported.
So if we export the model in training_mode, we would not be able to run the model in inference mode since batch_norm will have a fixed attribute value of training_mode.
I am thinking that this should not be an issue, is that correct? In that case, I will change it to be an attribute instead.

onnx/defs/nn/defs.cc
lara-hdr: makes sense; I will add it explicitly in the doc.

docs/Operators.md
```diff
@@ -4208,30 +4252,32 @@ expect(node, inputs=[x, y], outputs=[z],
...
 
-#### Inputs
+#### Inputs (1 - 2)
 
 <dl>
 <dt><tt>data</tt> : T</dt>
 <dd>The input data as Tensor.</dd>
+<dt><tt>ratio</tt> (optional) : T1</dt>
+<dd>The ratio of random dropout, with value in [0, 1). If this input was not set, or if it was set to 0, the output would be a simple copy of the input. If it's non-zero, output will be a random dropout of input, which is typically the case during training.</dd>
 </dl>
```
SherlockNoMad: There is a scaling involved when ratio is non-zero 

const float scale = 1.f / (1.f - ratio);

EigenMap<float>(Y) = scale * EigenMap<float>(X).cwiseProduct(EigenMap<bool>(mask).cast<float>());

docs/Operators.md
```diff
@@ -4240,8 +4286,10 @@ Other versions of this operator: <a href="Changelog.md#Dropout-1">Dropout-1</a>,
 <dl>
 <dt><tt>T</tt> : tensor(float16), tensor(float), tensor(double)</dt>
 <dd>Constrain input and output types to float tensors.</dd>
-<dt><tt>T1</tt> : tensor(bool)</dt>
-<dd>Constrain output mask types to boolean tensors.</dd>
+<dt><tt>T1</tt> : tensor(float)</dt>
```
SherlockNoMad: >tensor(float [](start = 18, length = 12)

ratio can be all float type 
including fp16

docs/Operators.md
```diff
@@ -4240,8 +4286,10 @@ Other versions of this operator: <a href="Changelog.md#Dropout-1">Dropout-1</a>,
 <dl>
 <dt><tt>T</tt> : tensor(float16), tensor(float), tensor(double)</dt>
```
SherlockNoMad: >: tensor(float16), tensor(float), tensor(double) [](start = 15, length = 48)

T should be all type 

including all ints

docs/Operators.md
```diff
@@ -1848,6 +1848,8 @@ Other versions of this operator: <a href="Changelog.md#BatchNormalization-1">Bat
 <dd>running (training) or estimated (testing) mean tensor of shape (C).</dd>
 <dt><tt>var</tt> : T</dt>
 <dd>running (training) or estimated (testing) variance tensor of shape (C).</dd>
+<dt><tt>training_mode</tt> (optional) : T1</dt>
```
SherlockNoMad: >training_mode [](start = 8, length = 13)

great! This flag has to be an input, instead of an attribute, since we need to modify it during the runtime. 

Please add this to the PR description.

onnx/defs/nn/defs.cc
wschin: My original concern is  its dynamic output schema which depends on an input flag. Can we make them optional (like LSTM's outputs)? It can probably make ONNX's static shape inference working again.

onnx/defs/nn/defs.cc
wschin: Let's just make `mean` and `var` inaccessible. The other two will be accessed by training operators and they don't break SSA.

onnx/defs/nn/defs.cc
lara-hdr: This needs to be an an input following the comment: https://github.com/onnx/onnx/pull/2568#discussion_r371459340 .
You mean make the output mean/var optional? It is already the case.

wschin: But you mentioned that if `training_mode=True`, those outputs won't be populated. This doesn't match the behavior of LSTM --- A optional output will be always populated as long as the user specifies that in the output list.

docs/Changelog.md
```diff
@@ -13965,6 +13965,122 @@ This version of the operator has been available since version 12 of the default
 <dd>Constrain input and output types to all numeric tensors.</dd>
 </dl>
 
+### <a name="BatchNormalization-12"></a>**BatchNormalization-12**</a>
+
+  Carries out batch normalization as described in the paper
+  https://arxiv.org/abs/1502.03167. Depending on the mode it is being run,
+  there are multiple cases for the number of outputs, which we list below:
+  
+  Output case #1: Y, mean, var, saved_mean, saved_var (training mode)
```
wschin: Could you please add `equations` to formally describe these two cases' behaviors?

docs/Operators.md
wschin: @lara-hdr, could you include Sherlock's equation in the spec if we haven't done that yet?

onnx/defs/nn/defs.cc
lara-hdr: got it. I updated the doc for them to be optional

docs/Changelog.md
```diff
@@ -13967,12 +13967,18 @@ This version of the operator has been available since version 12 of the default
...
+  There is one required output 'Y' and four optional outputs : 'mean', 'var', 'saved_mean', 'saved_var' used for training.
   
-  Output case #1: Y, mean, var, saved_mean, saved_var (training mode)
-  Output case #2: Y (test mode, where other outputs will not be populated)
+  The statistics are updated as follows:
+  ```
+  mean = running_mean * momentum + saved_mean * (1 - momentum)
+  var = running_var * momentum + saved_var * (1 - momentum)
+  ```
+  where 'saved_mean' and 'saved_var' are the observed mean and var per channel of the input X.
```
wschin: From the description here, it looks like `saved_mean = ReduceMean(X, axis=channel_index)`. Is it correct?

docs/Changelog.md
```diff
@@ -14034,11 +14040,17 @@ This version of the operator has been available since version 12 of the default
...
-  output (floating tensor) and mask (`Tensor<bool>`). Depending on whether it is
-  in test mode or not, the output Y will either be a random dropout, or a simple
-  copy of the input. Note that our implementation of Dropout does scaling in
-  the training phase, so during testing nothing needs to be done.
+  Dropout takes an input floating tensor and an input ratio (float scalar), and produces two tensor outputs,
+  output (floating tensor) and mask (`Tensor<bool>`). The output Y will be a random dropout;
+  Note that our implementation of Dropout does scaling in
+  the training phase, so during testing nothing needs to be done. The output is computed as :
+  ```
+  output_i = scale * input_i * mask_i,
```
wschin: What is `_i`? Could you please give it a formal definition? The input name below is `data`. Could you please change `input_i` to `data_i` for consistency? Thanks.

docs/Changelog.md
```diff
@@ -14034,11 +14040,17 @@ This version of the operator has been available since version 12 of the default
...
+  ```
+  where
+  ```
+  scale = 1. / (1. - ratio).
```
wschin: Nice!

docs/Changelog.md
```diff
@@ -13965,6 +13965,134 @@ This version of the operator has been available since version 12 of the default
...
+
+<dl>
+<dt><tt>Y</tt> : T</dt>
+<dd>The output tensor of the same shape as X</dd>
+<dt><tt>mean</tt> (optional) : T</dt>
+<dd>The running mean after the BatchNormalization operator.Note that this output cannot be an input of any other operator.</dd>
+<dt><tt>var</tt> (optional) : T</dt>
+<dd>The running variance after the BatchNormalization operator.Note that this output cannot be an input of any other operator.</dd>
+<dt><tt>saved_mean</tt> (optional) : T</dt>
+<dd>Saved mean used during training to speed up gradient computation.Note that this output cannot be an input of any non-training operator.</dd>
```
```suggestion
<dd>Saved mean used during training to speed up gradient computation.</dd>
```
Because there is no notion of `training` and `non-training` operators, we probably need to drop the new description. It's fine for existing operators to consume `saved_*` as long as it doesn't break SSA.

docs/Changelog.md
lara-hdr: actually saved_mean=reducemean(X, axis=all_except_channel_index)

for example : if the input size is (1, 2, 1, 3), the nb channel is at index 1, and nb_channel = 2. 
So saved_mean is reduced over indices (0,2,3), and saved_mean should be of size (2).

wschin: Thanks a lot for the nice explanation. May we add it to the spec? As you have seen, I am a poor user who misinterpret this sentence. For a mathematical operator, I feel it's always better to write done things in a mathematical way. :)

docs/Changelog.md
```diff
@@ -13965,6 +13965,134 @@ This version of the operator has been available since version 12 of the default
...
 
+### <a name="BatchNormalization-12"></a>**BatchNormalization-12**</a>
+
+  Carries out batch normalization as described in the paper https://arxiv.org/abs/1502.03167.
+  There is three required inputs 'X', 'mean' and 'var', in addition to one optional input 'training_mode'.
+  Note that 'mean' and 'var' are expected to be the estimated statistics in inference mode (training_mode=False, default),
+  and the running statistics in training mode (traning_mode=True).
+  There is one required output 'Y' and four optional outputs : 'mean', 'var', 'saved_mean', 'saved_var' used for training.
+  
+  The statistics are updated as follows:
```
```suggestion
  The statistics are updated as follows:
```
Would the update happen in inference mode? If the computation rules depend on `training_mode`, we need to have the equations listed for both cases (`training_mode`=True/False).

docs/Changelog.md
```diff
@@ -13965,6 +13965,134 @@ This version of the operator has been available since version 12 of the default
...
+  ```
+  mean = mean * momentum + saved_mean * (1 - momentum)
+  var = var * momentum + saved_var * (1 - momentum)
+  ```
+  where 'saved_mean' and 'saved_var' are the observed mean and var per channel of the input X.
+  
+  For previous (depreciated) non-spatial cases, implementors are suggested
+  to flatten the input shape to (N x C*D1*D2 ..*Dn) before a BatchNormalization Op.
```
```suggestion
  to flatten the input shape to (N x C*D1*D2 ..*Dn) before a BatchNormalization operator.
```

docs/Changelog.md
```diff
@@ -13965,6 +13965,134 @@ This version of the operator has been available since version 12 of the default
...
+
+This version of the operator has been available since version 12 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>epsilon</tt> : float (default is 1e-05)</dt>
+<dd>The epsilon value to use to avoid division by zero.</dd>
+<dt><tt>momentum</tt> : float (default is 0.9)</dt>
+<dd>Factor used in computing the running mean and variance.e.g., running_mean = running_mean * momentum + saved_mean * (1 - momentum).</dd>
```
wschin: `running_mean` --> `mean`?

docs/Changelog.md
```diff
@@ -13965,6 +13965,134 @@ This version of the operator has been available since version 12 of the default
...
+
+<dl>
+<dt><tt>X</tt> : T</dt>
+<dd>Input data tensor from the previous operator; dimensions are in the form of (N x C x D1 x D2 ... Dn), where N is the batch size, C is the number of channels. Statistics are computed for every channel of C over N and D1 to Dn dimensions. For image data, input dimensions become (N x C x H x W). The op also accepts single dimension input of size N in which case C is assumed to be 1</dd>
+<dt><tt>scale</tt> : T</dt>
+<dd>Scale tensor of shape (C).</dd>
+<dt><tt>B</tt> : T</dt>
+<dd>Bias tensor of shape (C).</dd>
+<dt><tt>mean</tt> : T</dt>
+<dd>running (training) or estimated (testing) mean tensor of shape (C).</dd>
```
wschin: If `mean` is not updated in inference mode, we can do
```suggestion
<dd>running (training) or fixed (testing) mean tensor of shape (C).</dd>
```

docs/Changelog.md
```diff
@@ -13965,6 +13965,134 @@ This version of the operator has been available since version 12 of the default
...
+<dl>
+<dt><tt>T</tt> : tensor(float16), tensor(float), tensor(double)</dt>
+<dd>Constrain input and output types to float tensors.</dd>
+<dt><tt>T1</tt> : tensor(bool)</dt>
+<dd>Constrain input 'training_mode' types to boolean tensors.</dd>
+</dl>
+
+### <a name="Dropout-12"></a>**Dropout-12**</a>
+
+  Dropout takes an input floating tensor and an input ratio (float scalar), and produces two tensor outputs,
```
wschin: `float -> floating-point` also `floating -> `floating-point`.

docs/Changelog.md
```diff
@@ -13965,6 +13965,134 @@ This version of the operator has been available since version 12 of the default
...
+  output (floating tensor) and mask (`Tensor<bool>`). The output Y will be a random dropout;
+  Note that our implementation of Dropout does scaling in
+  the training phase, so during testing nothing needs to be done. The output is computed as :
```
```suggestion
  the training phase, so during testing nothing needs to be done. Under training mode, the output is computed as :
```

docs/Changelog.md
```diff
@@ -13965,6 +13965,134 @@ This version of the operator has been available since version 12 of the default
...
+  ```
+  output = scale * data * mask,
+  ```
+  where
+  ```
+  scale = 1. / (1. - ratio).
```
```suggestion
  scale = 1. / (1. - ratio) if training_mode=true else 1.
```

lara-hdr: Dropout does not have a training_mode attribute. 
Do you think we should introduce it?

docs/Changelog.md
lara-hdr: thanks! I updated the doc

docs/Changelog.md
```diff
@@ -13965,6 +13965,145 @@ This version of the operator has been available since version 12 of the default
...
+<dd>Constrain input and output types to float tensors.</dd>
+<dt><tt>T1</tt> : tensor(bool)</dt>
+<dd>Constrain input 'training_mode' types to boolean tensors.</dd>
+</dl>
+
+### <a name="Dropout-12"></a>**Dropout-12**</a>
+
+  Dropout takes an input floating-point tensor and an input ratio (floating-point scalar), and produces two tensor outputs,
+  output (floating-point tensor) and mask (`Tensor<bool>`). The output Y will be a random dropout;
+  Note that our implementation of Dropout does scaling in
```
```suggestion
  Note that this Dropout scales the masked input data by the following equation, so to convert the trained model into inference mode, the user can simply replace this Dropout with an Identity operator.
```

docs/Changelog.md
```diff
@@ -13965,6 +13965,145 @@ This version of the operator has been available since version 12 of the default
...
+  the training phase, so during testing nothing needs to be done. Under training mode, the output is computed as :
```
```suggestion
```

docs/Changelog.md
```diff
@@ -13965,6 +13965,145 @@ This version of the operator has been available since version 12 of the default
...
+  ```
+  output = scale * data * mask,
+  ```
+  where
+  ```
+  scale = 1. / (1. - ratio) if in training mode else 1.
```
```suggestion
  scale = 1. / (1. - ratio).
```

wschin: Sorry making this back and forth.

onnx/defs/nn/defs.cc
```diff
@@ -1650,37 +1726,54 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             {"tensor(bool)"},
-            "Constrain output mask types to boolean tensors.")
+            "Constrain output 'mask' types to boolean tensors.")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           propagateShapeAndTypeFromFirstInput(ctx);
+
+          if (ctx.getNumInputs() > 1) {
+            auto& ratio_input_shape = getInputShape(ctx, 1);
+            if (ratio_input_shape.dim_size() != 0) {
+                fail_shape_inference("Ratio is not a scalar.");
```
wschin:
```suggestion
                fail_shape_inference("Ratio of Dropout must be a scalar.");
```

onnx/defs/nn/defs.cc
```diff
@@ -1483,24 +1483,41 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+mean = mean * momentum + saved_mean * (1 - momentum)
+var = var * momentum + saved_var * (1 - momentum)
+```
+
+When training_mode=False:
+```
+saved_mean = ReducedMean(X, axis=all_except_channel_index)
+saved_var =  ReducedVar(X, axis=all_except_channel_index)
+
+mean = mean
```
wschin:
```suggestion
mean = saved_mean
```
?

onnx/defs/nn/defs.cc
```diff
@@ -1483,24 +1483,41 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+var = var
```
wschin:
```suggestion
var = saved_var
```
?

onnx/defs/nn/defs.cc
```diff
@@ -1527,48 +1544,107 @@ ONNX_OPERATOR_SET_SCHEMA(
         .Input(
             3,
             "mean",
-            "running (training) or estimated (testing) mean tensor of shape (C).",
+            "running (training) or fixed (testing) mean tensor of shape (C).",
```
wschin:
```suggestion
            "running (training) or estimated (testing) mean tensor of shape (C).",
```
I was wrong. The word `estimated` is more accurate. Sorry.

onnx/defs/nn/defs.cc
```diff
@@ -1527,48 +1544,107 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "T")
         .Input(
             4,
             "var",
-            "running (training) or estimated (testing) variance tensor of shape (C).",
+            "running (training) or fixed (testing) variance tensor of shape (C).",
```
wschin:
```suggestion
            "running (training) or estimated (testing) variance tensor of shape (C).",
```

docs/Changelog.md
lara-hdr: you are suggesting to remove the whole sentence "Note that ... computed as:" ?

lara-hdr: nvm just noticed the other comment.

docs/Changelog.md
wschin: No. I add another comment for deleting my stupid idea.

lara-hdr:
Thank you @wschin!

@SherlockNoMad lmk if there are any additional comments, so we can merge this as soon as possible :)

wschin:
@lara-hdr, we need another approval from operator SIG to merge it.

lara-hdr:
@gramalingam for review

onnx/defs/nn/defs.cc
gramalingam: I didn't understand the last two comments. For output values that will not be used in training_mode=false, the spec should not force the computation of those values.

onnx/defs/nn/defs.cc
gramalingam: I think we just need to (locally) rename the outputs of the operator to be different from the inputs of the operator for the sake of clarity: e.g., use, output_mean and output_var instead of mean and var. I don't think there needs to be any constraints like "this output cannot be an input of any other operator".

gramalingam: The names used here are "formal parameter names", and have no connection to the "actual parameter names" used in an actual node containing a call to this op.

onnx/defs/nn/defs.cc
gramalingam: More specifically: when training_mode=false, which output-values need to be computed?

onnx/defs/nn/defs.cc
```diff
@@ -1534,41 +1555,100 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "Constrain input and output types to float tensors.")
+        .TypeConstraint(
+            "T1",
+            {"tensor(bool)"},
+            "Constrain input 'training_mode' types to boolean tensors.")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           propagateShapeAndTypeFromFirstInput(ctx);
-          // TODO in training mode, it may be possible to infer some of
-          // the other outputs as well.
+
```
gramalingam: Before we use the i-th inputShape, we should check if an i-th inputShape is available (using either https://github.com/onnx/onnx/blob/121114107c38ec49bab4955c13cb82b7675bfc47/onnx/defs/shape_inference.h#L269 or https://github.com/onnx/onnx/blob/121114107c38ec49bab4955c13cb82b7675bfc47/onnx/defs/shape_inference.h#L274 ). It is possible to have input N, but have no shape for it (if shape is unknown).

onnx/defs/nn/defs.cc
```diff
@@ -1650,37 +1730,54 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .TypeConstraint(
+            "T2",
             {"tensor(bool)"},
-            "Constrain output mask types to boolean tensors.")
+            "Constrain output 'mask' types to boolean tensors.")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           propagateShapeAndTypeFromFirstInput(ctx);
+
+          if (ctx.getNumInputs() > 1) {
+            auto& ratio_input_shape = getInputShape(ctx, 1);
```
gramalingam: Needs a check that the inputShape is available first.

onnx/defs/nn/defs.cc
lara-hdr: 'mean', 'var', 'saved_mean', and 'saved_var' are only needed for training.
But they are optional parameters (so they have static shapes and we can do the shape inference), so even if training_mode=False and the user specify them as output, they will be outputs of the op

onnx/defs/nn/defs.cc
lara-hdr: I updated the shape inference to include the check, thanks!

lara-hdr:
cc @houseroad @postrational

lara-hdr:
@ebarsoum CI is green. Thanks!

PR #2570, opened 2020-01-24
Introduce MeanSquaredError as Loss Function

KsenijaS:
This ONNX function calculates loss using the equation for mean square error.
As inputs it takes tensors of arbitrary shapes and, depending on the reduction type, it outputs a scalar (if reduction is sum or mean) or the tensor of same shape as input (if reduction is none)

PyTorch: https://pytorch.org/docs/stable/nn.html#torch.nn.MSELoss
TensorFlow: https://www.tensorflow.org/api_docs/python/tf/keras/losses/MeanSquaredError


onnx/backend/test/case/node/meansquareerror.py
```diff
@@ -0,0 +1,35 @@
...
+from ..base import Base
+from . import expect
+
+
+class MeanSquaredError(Base):
+
+    @staticmethod
+    def export_mean_square_error():  # type: () -> None
+        # Define operator attributes.
+        reduction = 'mean'
```
SherlockNoMad: >mean [](start = 21, length = 4)

A few more examples for "sum and none"

and one example for with weight

onnx/backend/test/case/node/meansquareerror.py
```diff
@@ -0,0 +1,35 @@
...
+
+        # Create operator.
+        node = onnx.helper.make_node('MeanSquaredError',
+                                     inputs=['R', 'T'],
+                                     outputs=['X'],
+                                     reduction=reduction
+                                     )
+
+        # Define operator inputs
+        r = np.array([1.2, 2.5], dtype=np.float32)
```
SherlockNoMad: >[1.2, 2.5] [](start = 21, length = 10)

Some example for 3D inpu

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,67 @@
...
+            0,
+            "output",
+            "Weighted loss float Tensor. If reduction is none, this has the "
+            "shape of [batch_size]; otherwise, it is scalar.",
+            "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {}));
```
lara-hdr: We could add the shape and type inference here for the output.
We canpropagate the type and shape when the reduction in none, otherwise it's a tensor of shape 1

lara-hdr: If you can add tests for the shape inference here as well https://github.com/onnx/onnx/blob/master/onnx/test/shape_inference_test.py (maybe one with reduction and one wihout).

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,67 @@
...
+
+Finally, L is reduced:
+L = ReduceSum(L), if reduction = 'sum';
+    ReduceMean(L), if reduction = 'mean';
+    ReduceMean(L, axes=[0]), if reduction = 'none';
+
+.)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    MeanSquaredError,
```
wschin:
```suggestion
    MeanSquaredDistance,
```

docs/Changelog.md
```diff
@@ -14102,6 +14102,63 @@ This version of the operator has been available since version 12 of the default
...
 
+### <a name="MeanSquaredError-12"></a>**MeanSquaredError-12**</a>
+
+  Loss function that measures the
+  mean squared error (squared L2 norm) between each element in the 'scores'
+  and 'labels'.
+  
+  The loss can be described as:
+      L = (l_1, l_2, ..., l_N), l_n = (score_n - label_n)^2
+  , where N is the batch size.
```
wschin: Does it mean `score` and `label` must be N-element vectors?

docs/Changelog.md
```diff
@@ -14102,6 +14102,63 @@ This version of the operator has been available since version 12 of the default
 <dd>Constrain index tensor to int64</dd>
 </dl>
 
+### <a name="MeanSquaredError-12"></a>**MeanSquaredError-12**</a>
+
+  Loss function that measures the
+  mean squared error (squared L2 norm) between each element in the 'scores'
+  and 'labels'.
+  
+  The loss can be described as:
```
wschin: Before introducing this equation (output of loss function), we need to describe the shapes of `score` and `label` (inputs of loss function).

SherlockNoMad:
Looks good to me. Wei-Sheng will help you further review this in details.

onnx/backend/test/case/node/meansquareerror.py
```diff
@@ -0,0 +1,143 @@
...
+        expect(node, inputs=[r, t], outputs=[mse], name='test_mean_square_error_mean')
+
+
+    @staticmethod
+    def export_mean_square_error_mean_do_not_keepdims():  # type: () -> None
+        # Define operator attributes.
+        reduction = 'mean'
+        shape = [3, 2, 2]
+        axes = [1]
+        keepdims = 0
```
lara-hdr: The MSE op does not have any keepdims/axes attribute, so it should not be part of the tests.

onnx/backend/test/case/node/meansquareerror.py
```diff
@@ -0,0 +1,143 @@
+from __future__ import absolute_import
```
lara-hdr: since some of the logic repeats in the tests you could add a function mse(input, target, reduction='mean', w=None) to call in each of the test. Something like :

def mse(input, target, reduction='mean', w=None):
  out = np.square(input - target)
  if w is not None:
    out = out * w
  if reduction == 'mean':
    out = out.mean()
  elif reduction == 'sum':
    out = out.sum()
 return out

docs/Changelog.md
lara-hdr: yes, @KsenijaS could you update the documentation with the input shape?

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,73 @@
...
+            "output",
+            "Weighted loss float Tensor. If reduction is none, this has the "
+            "shape of [batch_size]; otherwise, it is scalar.",
+            "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+	    propagateElemTypeFromInputToOutput(ctx, 0, 0);
```
lara-hdr: The output shape is the same only when reduction==none.
so if reduction is none we want to propagateElemTypeFromInputToOutput(ctx, 0, 0),
otherwise the output is a scalar (output shape with 0 dimension)

onnx/backend/test/case/node/meansquareerror.py
```diff
@@ -0,0 +1,143 @@
...
+        r = np.array([1.2, 2.5], dtype=np.float32)
+        t = np.array([1.1, 2.6], dtype=np.float32)
+
+        # Compute Mean Square Error
+        l = np.square(r - t)
+        l = np.multiply(weights, l)
+        mse = l.mean(axis=0)
+
+        # Check results
+        expect(node, inputs=[r, t], outputs=[mse], name='test_mean_square_error_none_weights')
```
lara-hdr: make sure to add the files generated for mse with ./update_docs.sh (only the ones you added).
(it should generate some .onnx and .pb files for these tests)

wschin:
This PR introduces a new operator. Please make sure this PR follows the [rules](https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md) agreed by Operator SIG

docs/Changelog.md
```diff
@@ -14102,6 +14102,63 @@ This version of the operator has been available since version 12 of the default
...
+  , where N is the batch size.
+  
+  If 'weights' is provided, it should be broadcastable to shape of 'scores'.
+      L = Mul(weights, L)
+  , where Mul is element-wise binary multiplication with Numpy-style broadcasting support.
+  
+  Finally, L is reduced:
+  L = ReduceSum(L), if reduction = 'sum';
+      ReduceMean(L), if reduction = 'mean';
+      ReduceMean(L, axes=[0]), if reduction = 'none';
```
wschin: Why is `none` mapped to `axes=[0]`?

KsenijaS: When it's along axis 0 no reduction is happening

KsenijaS:
This PR is dependent upon Liqun's PR (https://github.com/onnx/onnx/pull/2551) that introduced property HasQueriedFunction and that's why some tests are failing.

onnx/defs/loss/defs.cc
```diff
@@ -63,7 +63,66 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
             "Constrain input and output types to float tensors.")
-        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+	.AddQueriedFunctionBody([](FunctionBodyQueryContext& ctx) { // no weight, reduction is "none"
+	      return ctx.getNumInputs() == 2 && ctx.getAttribute("reduction")->s() == "none"; },
+	    FunctionBodyHelper::BuildNodes({
+	        // nodes: {outputs, op, inputs, attributes}
+	        FunctionBodyHelper::Const<int>("Q_Pow", 2), 
+	        {{"X_Sub"}, "Sub", {"scores", "labels"}},
+                {{"output"}, "Pow", {"X_Sub", "Q_Pow"}}
```
lara-hdr: nit: a couple of lines are not aligned in this file

onnx/defs/loss/defs.cc
```diff
@@ -63,7 +63,66 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                }))
+
+	.AddQueriedFunctionBody([](FunctionBodyQueryContext& ctx) { // weight, reduction is "none"
+              return ctx.getNumInputs() == 2 && ctx.getAttribute("reduction")->s() == "none"; },                                                                                                                 FunctionBodyHelper::BuildNodes({
+                // nodes: {outputs, op, inputs, attributes}
+		FunctionBodyHelper::Const<int>("Q_Pow", 2),                                                                                                           
+		{{"X_Sub"}, "Sub", {"scores", "labels"}},                                                                                                                                                 	     {{"X_Pow"}, "Pow", {"X_Sub", "Q_Pow"}},
+                {{"output"}, "Mul", {"weights", "X_Pow"}}
+                }))
+
```
lara-hdr: missing X_Pow here?

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,134 @@
...
+                {{"X_Pow"}, "Pow", {"X_Sub", "Q_Pow"}},
+                {{"X_Mul"}, "Mul", {"weights", "X_Pow"}},
+		{{"output"}, "ReduceSum", {"X_Mul"}}
+                }))
+ 
+	.TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+	    propagateElemTypeFromInputToOutput(ctx, 0, 0);
+	    std::string reduction = getAttribute(ctx, "reduction", "mean");
+	    if (reduction.compare("none") == 0) {
+	    	propagateShapeFromInputToOutput(ctx, 0, 0);
```
lara-hdr: can you check that inputShape is available as explained in https://github.com/onnx/onnx/pull/2568#discussion_r373230828

onnx/defs/loss/defs.cc
spandantiwari: I agree. Renaming it to `MeanSquaredDistance` and keeping it in the default domain would be my recommendation.

KsenijaS:
Please review @houseroad @postrational

docs/Changelog.md
```diff
@@ -14102,6 +14102,63 @@ This version of the operator has been available since version 12 of the default
...
+      L = (l_1, l_2, ..., l_N), l_n = (score_n - label_n)^2
+  , where N is the batch size.
+  
+  If 'weights' is provided, it should be broadcastable to shape of 'scores'.
+      L = Mul(weights, L)
+  , where Mul is element-wise binary multiplication with Numpy-style broadcasting support.
+  
+  Finally, L is reduced:
+  L = ReduceSum(L), if reduction = 'sum';
+      ReduceMean(L), if reduction = 'mean';
```
gramalingam: Is weights allowed for 'mean'? If it is allowed, does it mean compute a weighted-mean (which would required dividing by the sum-of-weights, and not the batch-size)?

spandantiwari:
Support for weighted mean has been removed. Does that mean there's no immediate use case for it?

onnx/backend/test/case/node/meansquaredistance.py
```diff
@@ -0,0 +1,132 @@
...
+from __future__ import unicode_literals
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+class MeanSquaredDistance(Base):
```
lara-hdr: It would be good to have a reference implementation function.
As explained in the comment in the deleted file : https://github.com/onnx/onnx/pull/2570#discussion_r372130505

docs/Changelog.md
KsenijaS: weights are used only in tensorflow and are allowed for mean, and it is divided by batch-size

KsenijaS:
> Support for weighted mean has been removed. Does that mean there's no immediate use case for it?

No it hasn't, tensorflow is using weighted mean.

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,118 @@
...
+  body.push_back(FunctionBodyHelper::Const<int>("Q_Pow", 2));
+  body.push_back({{"X_Sub"}, "Sub", {"scores", "labels"}});
+
+  if (ctx.hasInput(2)) {
+    if (ctx.getAttribute("reduction")->s() == "none") {
+      body.push_back({{"output"}, "Pow", {"X_Sub", "Q_Pow"}});
+    } else {
+      body.push_back({{"X_Pow"}, "Pow", {"X_Sub", "Q_Pow"}});
+      if (ctx.getAttribute("reduction")->s() == "mean") {
+        body.push_back({{"output"}, "ReduceMean", {"X_Pow"}});
```
SherlockNoMad: >ReduceMean [](start = 37, length = 10)

default behavior is reducing a tensor to a scalar

[M, N, K] -> [1, 1, 1]

Is this the expected bahavior?

onnx/defs/operator_sets.h
```diff
@@ -731,10 +732,12 @@ class OpSet_Onnx_ver12 {
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, MaxPool)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, ReduceMax)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, ReduceMin)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, MeanSquaredDistance)>());
```
SherlockNoMad: >fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, MeanSquaredDistance)>()); [](start = 4, length = 86)

duplicated?

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,118 @@
...
+            0,
+            "output",
+            "Weighted loss float Tensor. If reduction is none, this has the "
+            "shape of [batch_size]; otherwise, it is scalar.",
+            "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+	.SetContextDependentFunctionBodyBuilder(BuildContextDependentFunctionBodyMSD)
```
SherlockNoMad: >	 [](start = 0, length = 1)

tab to space

onnx/defs/loss/defs.cc
KsenijaS: default behavior is when reduction is mean and output is scalar

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,118 @@
...
+
+static const char* MSD_ver12_doc = R"DOC(Loss function that measures the
+mean squared distance (squared L2 norm) between each element in the 'scores'
+and 'labels'.
+
+The loss can be described as:
+    L = (l_1, l_2, ..., l_N), l_n = (score_n - label_n)^2
+, N is the batch size.
+
+score and label are vectors of arbitrary shapes with total of N elements each.
```
gramalingam: (a) I guess these can be multi-dimensional? If so, suggest changing "vector" to "tensor".
(b) It seems to me that either we should require score and label to have same shape or allow them to be broadcastable. Which is it?

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,118 @@
...
+
+If 'weights' is provided, it should be broadcastable to shape of 'scores'.
```
gramalingam: change to "broadcastable to the shape of L" (which may differ from the shape of scores if we allow broadcast between scores and label)

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,118 @@
...
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .SetContextDependentFunctionBodyBuilder(BuildContextDependentFunctionBodyMSD)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+            propagateElemTypeFromInputToOutput(ctx, 0, 0);
+            std::string reduction = getAttribute(ctx, "reduction", "mean");
+            if (reduction.compare("none") == 0 && hasInputShape(ctx, 0)) {
+                propagateShapeFromInputToOutput(ctx, 0, 0);
+            }
```
gramalingam: else we should propagate a scalar-shape, which is the empty list of dimensions, I think. May be something like "updateOutputShape(ctx, 0, TensorShapeProto());"?

gramalingam: I mean only in the case where reduction is "none"

gramalingam: In the then-branch, we may need to do broadcast, if broadcast is allowed.

docs/Changelog.md
```diff
@@ -14292,6 +14292,69 @@ This version of the operator has been available since version 12 of the default
...
 
+### <a name="MeanSquaredDistance-12"></a>**MeanSquaredDistance-12**</a>
+
+  Loss function that measures the
+  mean squared distance (squared L2 norm) between each element in the 'scores'
+  and 'labels'.
+  
+  The loss can be described as:
+      L = (l_1, l_2, ..., l_N), l_n = (score_n - label_n)^2
+  , N is the batch size.
```
wschin: Line 14301-14305 can be shortened to

This operator first computes a element-wise squared loss tensor `L` via
```
L = Pow(Sub(score, label), 2)
```

docs/Changelog.md
```diff
@@ -14292,6 +14292,69 @@ This version of the operator has been available since version 12 of the default
...
+  
+  score and label are vectors of arbitrary shapes with total of N elements each.
+  
+  If 'weights' is provided, it should be broadcastable to shape of 'scores'.
+      L = Mul(weights, L)
+  , where Mul is element-wise binary multiplication with Numpy-style broadcasting support.
+  
+  Finally, L is reduced:
```
```suggestion
  Finally, L is optionally reduced:
```

docs/Changelog.md
```diff
@@ -14292,6 +14292,69 @@ This version of the operator has been available since version 12 of the default
...
+
+#### Version
+
+This version of the operator has been available since version 12 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>reduction</tt> : string (default is mean)</dt>
+<dd>Type of reduction to apply to loss: none, sum, mean(default). 'none': the output is the loss for each sample in the batch.'sum': the output will be summed. 'mean': the sum of the output will be divided by the batch_size.</dd>
```
```suggestion
<dd>Type of reduction to apply to loss: none, sum, mean(default). 'none': the output is the loss for each sample in the batch.'sum': the output will be summed into a scalar. 'mean': the sum of the output will be divided by the batch_size.</dd>
```

onnx/backend/test/case/node/meansquaredistance.py
```diff
@@ -0,0 +1,136 @@
...
+from __future__ import unicode_literals
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+def mean_squared_distance(input, target, reduction='mean', w=None):
```
wschin: Any numerical comparison between this reference implementation and the pytorch (or tensorflow) corresponding one?

onnx/test/shape_inference_test.py
```diff
@@ -2993,6 +2993,22 @@ def test_negative_log_likehood_input_weight_shape_mismatch(self):  # type: () ->
...
+        graph = self._make_graph(
+            [("x", TensorProto.FLOAT, (2, 3)),
+             ("y", TensorProto.FLOAT, (2, 3))],
+            [make_node('MeanSquaredDistance', ['x', 'y'], ['z'], reduction='none')],
+            [],)
+        self._assert_inferred(graph, [make_tensor_value_info('z', TensorProto.FLOAT, (2, 3))])  # type: ignore
+
+    def test_mean_square_mean(self):  # type: () -> None
+        graph = self._make_graph(
+            [("x", TensorProto.FLOAT, (2, 3)),
```
wschin: Please also add 1-D and 4-D cases. Thank you.

docs/Changelog.md
KsenijaS: I think that makes it less readable

onnx/test/shape_inference_test.py
KsenijaS: there is already 1-D, I will add 4-D

onnx/backend/test/case/node/meansquaredistance.py
KsenijaS: import torch
import torch.nn as nn
import numpy as np
from numpy.testing import assert_allclose

def mean_squared_distance(input, target, reduction='mean', w=None):
    out = np.square(input - target)
    if w is not None:
        out = np.multiply(out, w)
    if reduction == 'mean':
        out = np.mean(out)
    elif reduction == 'sum':
        out = np.sum(out)
    return out

loss = nn.MSELoss()
input = torch.randn(3, 5)
target = torch.randn(3, 5)
output = loss(input, target)

my_loss = mean_squared_distance(input.numpy(), target.numpy())
assert_allclose(my_loss, output.numpy(), rtol=1e-05)

KsenijaS:
```
import torch
import torch.nn as nn
import numpy as np
from numpy.testing import assert_allclose

def mean_squared_distance(input, target, reduction='mean', w=None):
    out = np.square(input - target)
    if w is not None:
        out = np.multiply(out, w)
    if reduction == 'mean':
        out = np.mean(out)
    elif reduction == 'sum':
        out = np.sum(out)
    return out

loss = nn.MSELoss()
input = torch.randn(3, 5)
target = torch.randn(3, 5)
output = loss(input, target)

my_loss = mean_squared_distance(input.numpy(), target.numpy())
assert_allclose(my_loss, output.numpy(), rtol=1e-05)
```

onnx/defs/math/defs.cc
```diff
@@ -2038,4 +2038,117 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .SetContextDependentFunctionBodyBuilder(BuildContextDependentFunctionBodyMSD)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+            propagateElemTypeFromInputToOutput(ctx, 0, 0);
+            std::string reduction = getAttribute(ctx, "reduction", "mean");
+            if (reduction.compare("none") == 0 && hasInputShape(ctx, 0)) {
+                propagateShapeFromInputToOutput(ctx, 0, 0);
+            } else {
```
gramalingam: change to
```C++
   if (reduction.compare("none") == 0) {
      if (hasInputShape(ctx, 0)) {
      }
   } else {
      updateOutputShape(ctx, 0, TensorShapeProto());
   }
```

PR #2573, opened 2020-01-29
Introduce SoftmaxCrossentropy as a loss function

KsenijaS:
This ONNX function calculates loss using softmax and cross entropy functions.
Inputs can be K-dimensional tensor of the same shape, where K >= 1.
Output is dependent on reduction attribute. If reduction is none, output is scalar and if not, output is the same shape as input.

Pytorch: https://pytorch.org/docs/stable/nn.html#crossentropyloss


wschin:
We have a [guideline](https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md) for adding new operator. Could you please take a look and make sure all requirements are met? Many thanks.

onnx/backend/test/case/node/crossentropy.py
```diff
@@ -0,0 +1,118 @@
...
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+def softmax_2d(x):  # type: (np.ndarray) -> np.ndarray
+            max_x = np.max(x, axis=1).reshape((-1, 1))
```
lara-hdr: nit: to many spaces

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,72 @@
...
+            "output",
+            "Weighted loss float Tensor. If reduction is 'none', this has the "
+            "shape of [batch_size], or [batch_size, d1, d2, ..., dk] in case of "
+            "K-dimensional loss. Otherwise, it is a scalar.",
+            "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {}));
```
lara-hdr: could you add the shape inference

KsenijaS:
this PR is dependent upon Liqun's PR (https://github.com/onnx/onnx/pull/2551) where AddQueriedFunctionBody was introduced. And that's why some of the tests are failing.

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,141 @@
...
+        .Input(
+            0,
+            "scores",
+            "The predicted outputs with shape [batch_size, class_size], or "
+            "[batch_size, class_size, d1, d2 , ..., dk], where K is the number of dimensions.",
+            "T")
+        .Input(
+            1,
+            "labels",
+            "The ground truth output tensor, same dimensions as 'scores'. "
```
lara-hdr: can you verify that 'scores' and 'label' have the same shape in the shape inference?

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,141 @@
...
+		{{"output"}, "Mul", {"labels", "X_Log"}}
+		}))
+	.AddQueriedFunctionBody([](FunctionBodyQueryContext& ctx) { // no weight, reduction is "sum"
+              return ctx.getNumInputs() == 2 && ctx.getAttribute("reduction")->s() == "sum"; },
+            FunctionBodyHelper::BuildNodes({
+                // nodes: {outputs, op, inputs, attributes}
+                {{"X_SM"}, "Softmax", {"scores"}},
+                {{"X_Log"}, "Log", {"X_SM"}},
+                {{"X_Mul"}, "Mul", {"labels", "X_Log"}},
+		{{"output"}, "ReduceSum", {"X_Mul"}}
```
lara-hdr: nit: align for readability (same comment for lines 101 -110-119-129)

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,146 @@
...
+                {{"output"}, "Mul", {"weights", "X_Mul"}}
+                }))
+        .AddQueriedFunctionBody([](FunctionBodyQueryContext& ctx) { // weight, reduction is "sum"
+              return ctx.getNumInputs() > 2 && ctx.getAttribute("reduction")->s() == "sum"; },
+            FunctionBodyHelper::BuildNodes({
+                // nodes: {outputs, op, inputs, attributes}
+                {{"X_SM"}, "Softmax", {"scores"}},
+                {{"X_Log"}, "Log", {"X_SM"}},
+                {{"X_Mul"}, "Mul", {"labels", "X_Log"}},
+		{"X_Mul2"}, "Mul", {"weights", "X_Mul"}},
```
SherlockNoMad: >		 [](start = 0, length = 2)

tab to space 

in all source files

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,146 @@
...
+                }))
+        .AddQueriedFunctionBody([](FunctionBodyQueryContext& ctx) { // no weight, reduction is "mean"
+              return ctx.getNumInputs() > 2 && ctx.getAttribute("reduction")->s() == "mean"; },
+            FunctionBodyHelper::BuildNodes({
+                // nodes: {outputs, op, inputs, attributes}
+                {{"X_SM"}, "Softmax", {"scores"}},
+                {{"X_Log"}, "Log", {"X_SM"}},
+                {{"X_Mul"}, "Mul", {"labels", "X_Log"}},
+		{{"X_Mul2"}, "Mul", {"weights", "X_Mul"}},
+                {{"output"}, "ReduceMean", {"X_Mul2"}}
```
SherlockNoMad: >ReduceMean [](start = 30, length = 10)

do we need to set axes here ? 
Does default value works ?

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,146 @@
...
+            "output",
+            "Weighted loss float Tensor. If reduction is 'none', this has the "
+            "shape of [batch_size], or [batch_size, d1, d2, ..., dk] in case of "
+            "K-dimensional loss. Otherwise, it is a scalar.",
+            "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+	.AddQueriedFunctionBody([](FunctionBodyQueryContext& ctx) { // no weight, reduction is "none"
```
SherlockNoMad: >AddQueriedFunctionBody [](start = 2, length = 22)

general speaking, I feel that this function is not extensible... 

In this op, there is 6 combinations....
What about other ops if the options are combinatorial....

onnx/defs/loss/defs.cc
lara-hdr: no need to specify axes here.

If the reduction is specified, ReduceMean will be computed over the whole ouput (axes=None, and output will be scalar).
For more specific cases requiring computing the mean over specific axes, there is the possibility of using a subgraph with ONNX::SoftmaxCrossEntropy with 'none' reduction followed ONNX::ReduceMean with the desired axes.

How does that sound?

KsenijaS:
Please review @houseroad @postrational

onnx/backend/test/case/node/crossentropy.py
```diff
@@ -0,0 +1,118 @@
...
+    @staticmethod
+    def export_crossentropy_none():
+        # Define operator attributes.
+        reduction = 'none'
+
+        # Create operator.
+        node = onnx.helper.make_node('SoftmaxCrossEntropy',
+                                     inputs=['x', 'y'],
+                                     outputs=['z'],
+                                     reduction=reduction
```
wschin:
```suggestion
                                     reduction=reduction)
```

onnx/backend/test/case/node/crossentropy.py
```diff
@@ -0,0 +1,118 @@
...
+                                     )
+
+        # Define operator inputs.
+        x = np.array([[0, 1, 2, 3], [10000, 10001, 10002, 10003]], dtype=np.float32)
+        labels = np.array([[0.0320586, 0.08714432, 0.23688284, 0.64391428], [0.0320586, 0.08714432, 0.23688284, 0.64391428]], dtype=np.float32)
+        
+        # Compute SoftmaxCrossEntropy
+        p = softmax_2d(x)
```
wschin: May we do
```
l = SoftmaxCrossEntropy(x, labels)
```
?

onnx/backend/test/case/node/crossentropy.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @staticmethod
+    def export_crossentropy_none_weights():
+        # Define operator attributes.
+        reduction = 'none'
+        weights = [[0.9, 0.8, 0.9, 0.8], [0.9, 0.7, 0.8, 0.9]]
+
+        # Create operator.
+        node = onnx.helper.make_node('SoftmaxCrossEntropy',
+                                     inputs=['x', 'y'],
```
wschin: Where is `weights` in this node? Also, can we make `weights` an attribute?

onnx/backend/test/case/node/crossentropy.py
```diff
@@ -0,0 +1,118 @@
...
+                                     outputs=['z'],
+                                     reduction=reduction
+                                     )
+
+        # Define operator inputs.
+        x = np.array([[0, 1, 2, 3], [10000, 10001, 10002, 10003]], dtype=np.float32)
+        labels = np.array([[0.0320586, 0.08714432, 0.23688284, 0.64391428], [0.0320586, 0.08714432, 0.23688284, 0.64391428]], dtype=np.float32)
+
+        # Compute SoftmaxCrossEntropy
+        p = softmax_2d(x)
```
wschin: These three line clearly indicate that this operator can be composed... I don't insist this should be proposed as a function though. Can you merge the three lines into a single function and called it in every test? It's not very clear if we have different numpy reference implementations for different `reduce` and `input` configurations....

onnx/backend/test/case/node/crossentropy.py
```diff
@@ -0,0 +1,118 @@
+from __future__ import absolute_import
```
wschin: `crossentropy.py`? Isn't it `SoftmaxCrossEntropyLoss`?

onnx/backend/test/case/node/crossentropy.py
```diff
@@ -0,0 +1,118 @@
...
+from __future__ import unicode_literals
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+def softmax_2d(x):  # type: (np.ndarray) -> np.ndarray
```
wschin: Could we make this a full reference implementation
```
def softmax_cross_entropy(x, y, w=None):
  # Softmax
  ...
```
? It'd be even better if you have a Pytorch script showing that this reference implementation matches whatever Pytorch is doing. Please also confirm that Tensorflow can also use this operator.

docs/Operators.md
```diff
@@ -17030,6 +17031,187 @@ expect(node, inputs=[x], outputs=[y],
...
+</dl>
+
+#### Inputs (2 - 3)
+
+<dl>
+<dt><tt>scores</tt> : T</dt>
+<dd>The predicted outputs with shape [batch_size, class_size], or [batch_size, class_size, d1, d2 , ..., dk], where K is the number of dimensions.</dd>
+<dt><tt>labels</tt> : T</dt>
+<dd>The ground truth output tensor, same dimensions as 'scores'. Usualy, it's a one-hot representation of groud-truth class.</dd>
+<dt><tt>weights</tt> (optional) : T</dt>
```
wschin: I feel it should be an attribute.

docs/Operators.md
```diff
@@ -17030,6 +17031,187 @@ expect(node, inputs=[x], outputs=[y],
 </details>
 
 
+### <a name="SoftmaxCrossEntropy"></a><a name="softmaxcrossentropy">**SoftmaxCrossEntropy**</a>
```
```suggestion
### <a name="SoftmaxCrossEntropyLoss"></a><a name="softmaxcrossentropy">**SoftmaxCrossEntropy**</a>
```
SoftmaxCrossEntropy ---> SoftmaxCrossEntropyLoss

The first function also outputs normalization coefficient but the second term doesn't.

onnx/test/shape_inference_test.py
```diff
@@ -2792,6 +2792,22 @@ def test_einsum_incorrect_num_inputs(self):  # type: () -> None
...
+            [("x", TensorProto.FLOAT, (2, 3)),
+             ("y", TensorProto.FLOAT, (2, 3))],
+            [make_node('SoftmaxCrossEntropy', ['x', 'y'], ['z'], reduction='none')],
+            [],)
+        self._assert_inferred(graph, [make_tensor_value_info('z', TensorProto.FLOAT, (2, 3))])  # type: ignore
+
+    def test_cross_entropy_mean(self):  # type: () -> None
+        graph = self._make_graph(
+            [("x", TensorProto.FLOAT, (2, 3)),
+             ("y", TensorProto.FLOAT, (2, 3))],
```
wschin: Please add a case with the present of `weight`, if we decide not to move `weight` to the attribute list.

docs/Changelog.md
```diff
@@ -14188,3 +14188,61 @@ This version of the operator has been available since version 12 of the default
...
 </dl>
 
+### <a name="SoftmaxCrossEntropy-12"></a>**SoftmaxCrossEntropy-12**</a>
+
+  Loss function that measures the softmax cross entropy
+  between 'scores' and 'labels'.
+  The loss can be described as:
+      L = (l_1, l_2, ..., l_N), where N is the batch_size
+  
+  scores: (N, C) where C is the number of classes, or (N, C, d1, d2,..., dk),
```
gramalingam: Change "scores" to "shape(scores)"?

docs/Changelog.md
```diff
@@ -14188,3 +14188,61 @@ This version of the operator has been available since version 12 of the default
...
+  	with K >= 1 in case of K-dimensional loss.
+  labels: (N) where each value is 0 <= labels[i] <= C-1, or (N, d1, d2,..., dk),
```
gramalingam: change "labels" to "shape(labels)"?

docs/Changelog.md
```diff
@@ -14188,3 +14188,61 @@ This version of the operator has been available since version 12 of the default
...
+  The loss for one sample, l_n, can caculated as follows
+      let p = Softmax(scores)
+      l_n = -sum(label_i * log(p_i)), where i is the index of classes.
+  or
+      l_n = -sum(weight_i * label_i * log(p_i)), if 'weights' is provided.
+  Finally, L is reduced:
+  L = ReduceSum(L), if reduction = 'sum';
+      ReduceMean(L), if reduction = 'mean';
+      L, if reduction = 'none'
+
```
gramalingam: What about the K-dimensional loss? I don't quite understand how the above equations apply in that case.

docs/Changelog.md
```diff
@@ -14188,3 +14188,61 @@ This version of the operator has been available since version 12 of the default
...
+<dd>Type of reduction to apply to loss: none, sum, mean(default). 'none': no reduction will be applied, 'sum': the output will be summed. 'mean': the sum of the output will be divided by the number of elements in the output.</dd>
+</dl>
+
+#### Inputs (2 - 3)
+
+<dl>
+<dt><tt>scores</tt> : T</dt>
+<dd>The predicted outputs with shape [batch_size, class_size], or [batch_size, class_size, d1, d2 , ..., dk], where K is the number of dimensions.</dd>
+<dt><tt>labels</tt> : T</dt>
+<dd>The ground truth output tensor, same dimensions as 'scores'. Usualy, it's a one-hot representation of groud-truth class.</dd>
```
gramalingam: It isn't the same dimension as "scores", right? May be better to say [batch_size] or etc.

docs/Changelog.md
gramalingam: Is the same computation applied to each of of the k-dimensional points separately?

docs/Operators.md
KsenijaS: in order to be consistent with MSELoss weight should be an input.

onnx/defs/loss/defs.cc
KsenijaS: scores and label don't have the same shape.

onnx/backend/test/case/node/crossentropy.py
KsenijaS: weights will stay an input to be consistent with other loss functions

docs/Changelog.md
KsenijaS: It's the same as in NllLoss, SoftmaxCrossEntropy = Log(Softmax(scores)) + NllLoss(scores, labels).
L is the same shape as labels and if there is reduction, L is flattened and then sum or mean is performed and output is scalar.

docs/Operators.md
```diff
@@ -136,6 +136,7 @@
   * <a href="#Size">Size</a>
   * <a href="#Slice">Slice</a>
   * <a href="#Softmax">Softmax</a>
+  * <a href="#SoftmaxCrossEntropyLoss">SoftmaxCrossEntropyLoss</a>
```
SherlockNoMad: >SoftmaxCrossEntropyLoss [](start = 14, length = 23)

 nit. 
how about just call it SoftmaxCrossEntropy

SherlockNoMad: ah... I see WeiSheng's comment ... 

---
In reply to: [377937804](https://github.com/onnx/onnx/pull/2573#discussion_r377937804) [](ancestors = 377937804)

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,113 @@
...
+    "elements in the output.";
+
+static const char* SoftmaxCrossEntropyLoss_ver12_doc =
+    R"DOC(Loss function that measures the softmax cross entropy
+between 'scores' and 'labels'.
+The loss can be described as:
+    L = (l_1, l_2, ..., l_N), where N is the batch_size
+
+shape(scores): (N, C) where C is the number of classes, or (N, C, d1, d2,..., dk),
+	with K >= 1 in case of K-dimensional loss.
```
SherlockNoMad: >	 [](start = 0, length = 1)

tab to space 
Please also set it in the IDE to only use space

docs/Operators.md
SherlockNoMad: If it's already universally called *Loss, I am ok with the current naming

---
In reply to: [377938280](https://github.com/onnx/onnx/pull/2573#discussion_r377938280) [](ancestors = 377938280,377937804)

SherlockNoMad:
>   **Operators with function registered:**

Does this op belong here ?

---
Refers to: docs/Operators.md:163 in 9ca0b54. [](commit_id = 9ca0b5496cb06594eef68fb94738eb18e034c011, deletion_comment = False)

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,113 @@
...
+    ReduceMean(L), if reduction = 'mean'; if "weight" is provided, output is averaged by sum of weights.
+    L, if reduction = 'none'
+)DOC";
+
+bool BuildContextDependentFunctionBodySCE(const FunctionBodyBuildContext& ctx, const OpSchema& schema, FunctionProto& functionProto) {
+  std::vector<FunctionBodyHelper::NodeDef> body;
+  body.push_back({{"X_SM"}, "Softmax", {"scores"}});
+  body.push_back({{"X_Log"}, "Log", {"X_SM"}});
+  if (ctx.hasInput(2)) {
+    body.push_back({{"output"}, "NegativeLogLikelihoodLoss", {"X_Log", "labels"}});
```
SherlockNoMad: >NegativeLogLikelihoodLoss [](start = 33, length = 25)

How do we propagate attribute to the function body?

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,113 @@
...
+The loss can be described as:
+    L = (l_1, l_2, ..., l_N), where N is the batch_size
+
+shape(scores): (N, C) where C is the number of classes, or (N, C, d1, d2,..., dk),
+        with K >= 1 in case of K-dimensional loss.
+shape(labels): (N) where each value is 0 <= labels[i] <= C-1, or (N, d1, d2,..., dk),
+        with K >= 1 in case of K-dimensional loss.
+
+The loss for one sample, l_i, can caculated as follows:
+    l_i = -y[i][c][d1][d2]..[dk], where i is the index of classes.
```
gramalingam: is the LHS supposed to be l[i][d1][d2]...[dk] ?

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,113 @@
...
+            "scores",
+            "The predicted outputs with shape [batch_size, class_size], or "
+            "[batch_size, class_size, d1, d2 , ..., dk], where K is the number of dimensions.",
+            "T")
+        .Input(
+            1,
+            "labels",
+            "The ground truth output tensor, with shape [batch_size], or "
+            "[batch_size, d1, d2 , ..., dk], where K is the number of dimensions."
+            "Usualy, it's a one-hot representation of groud-truth class.",
```
gramalingam: "groud" => "ground"

onnx/defs/loss/defs.cc
```diff
@@ -0,0 +1,113 @@
...
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .SetContextDependentFunctionBodyBuilder(BuildContextDependentFunctionBodySCE)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+            propagateElemTypeFromInputToOutput(ctx, 0, 0);
+            std::string reduction = getAttribute(ctx, "reduction", "mean");
+            if (reduction.compare("none") == 0 && hasInputShape(ctx, 1)) {
+                propagateShapeFromInputToOutput(ctx, 1, 0);
+            }
```
gramalingam: if reduction is none, the output is a scalar, right? we can set the output shape for that case also.

KsenijaS: If reduction is none, output is the same shape as labels. If reduction is not none then output is scalar. Do you want me to add that case?

onnx/defs/loss/defs.cc
KsenijaS: yes, l_i is of shape l[i][d1][d2]...[dk]

onnx/defs/loss/defs.cc
gramalingam: Yes, sorry that's what I meant. Please do add it.

onnx/defs/loss/defs.cc
gramalingam: I think it is better to change the LHS to l[i][d1]...[dk]. Otherwise, it is unclear where the d1 to dk in the RHS come from.

docs/Changelog.md
```diff
@@ -14504,3 +14504,66 @@ This version of the operator has been available since version 12 of the default
...
+  
+  shape(scores): (N, C) where C is the number of classes, or (N, C, d1, d2,..., dk),
+  	with K >= 1 in case of K-dimensional loss.
+  shape(labels): (N) where each value is 0 <= labels[i] <= C-1, or (N, d1, d2,..., dk),
+  	with K >= 1 in case of K-dimensional loss.
+  
+  The loss for one sample, l_i, can caculated as follows:
+      l_i = -y[i][c][d1][d2]..[dk], where i is the index of classes.
+  or
+      l_i = -y[i][c][d1][d2]..[dk]*weights[c], if 'weights' is provided.
```
```suggestion
      l_i = -y[i][c][d1][d2]..[dk] * weights[c], if 'weights' is provided.
```

docs/Changelog.md
```diff
@@ -14504,3 +14504,66 @@ This version of the operator has been available since version 12 of the default
...
+  
+  where:
+      p = Softmax(scores)
```
wschin: I check the document of ONNX Softmax. If the input is N-D tensor with shape (N, C, D1, ..., Dk), it will be flattened into a 2-D tensor with shape (N, C * D1 * ... * Dk). I don't quite sure this matches the Softmax needed in Pytorch SoftmaxCrossEntropy.

KsenijaS: If there is reduction that means output is a scalar, so it doesn't matter. Only when reduction is none the shape matters, I could try to use Reshape op for that

KsenijaS: Also I think we should use LogSoftmax instead of Log and Softmax

KsenijaS: In Pytorch output of Softmax is of the same dimension and shape as input.

wschin: To compose it, we probably only need `Sum(Exp(X), axis=1)`. Please double-check.

docs/Changelog.md
```diff
@@ -14504,3 +14504,66 @@ This version of the operator has been available since version 12 of the default
 <dd>Constrain input and output types to high-precision and 8 bit numeric tensors.</dd>
 </dl>
 
+### <a name="SoftmaxCrossEntropyLoss-12"></a>**SoftmaxCrossEntropyLoss-12**</a>
+
+  Loss function that measures the softmax cross entropy
+  between 'scores' and 'labels'.
+  The loss can be described as:
+      L = (l_1, l_2, ..., l_N), where N is the batch_size
```
```suggestion
This operator first computes a loss tensor whose shape is identical to the labels input.
If the input is 2-D with shape (N, C), the loss tensor may be a N-element vector L = (l_1, l_2, ..., l_N).
If the input is N-D tensor with shape (N, C, d1, d2, ..., dk), the loss tensor L may have (N, d1, d2, ..., dk) as its shape and L[i,][j_1][j_2]...[j_k] denotes a scalar element in L.
```

docs/Changelog.md
```diff
@@ -14504,3 +14504,66 @@ This version of the operator has been available since version 12 of the default
...
+  
+  shape(scores): (N, C) where C is the number of classes, or (N, C, d1, d2,..., dk),
```
```suggestion
  shape(scores): (N, C) where C is the number of classes, or (N, C, D1, D2,..., Dk),
```
because we have `y[i][c][d1][d2]..[dk]` below.

docs/Changelog.md
```diff
@@ -14504,3 +14504,66 @@ This version of the operator has been available since version 12 of the default
...
+      l_i = -y[i][c][d1][d2]..[dk], where i is the index of classes.
+  or
+      l_i = -y[i][c][d1][d2]..[dk]*weights[c], if 'weights' is provided.
+  
+  where:
+      p = Softmax(scores)
+      y = log(p)
+      c = labels[i][d1][d2]...[dk]
+  
+  Finally, L is reduced:
```
```suggestion
  Finally, L is optionally reduced:
```

docs/Changelog.md
```diff
@@ -14504,3 +14504,66 @@ This version of the operator has been available since version 12 of the default
...
+  L = ReduceSum(L), if reduction = 'sum';
+      ReduceMean(L), if reduction = 'mean'; if "weight" is provided, output is averaged by sum of weights.
```
wschin: I think `weight` also applies to the `sum` and `none` mode. The description here seems saying `weight` only applies to `mean`. In addition, the equation to compute weighted loss is unclear. Is it L_weighted = Mul(L, weight)?

docs/Changelog.md
```diff
@@ -14504,3 +14504,66 @@ This version of the operator has been available since version 12 of the default
...
+<dl>
+<dt><tt>reduction</tt> : string (default is mean)</dt>
+<dd>Type of reduction to apply to loss: none, sum, mean(default). 'none': no reduction will be applied, 'sum': the output will be summed. 'mean': the sum of the output will be divided by the number of elements in the output.</dd>
+</dl>
+
+#### Inputs (2 - 3)
+
+<dl>
+<dt><tt>scores</tt> : T</dt>
+<dd>The predicted outputs with shape [batch_size, class_size], or [batch_size, class_size, d1, d2 , ..., dk], where K is the number of dimensions.</dd>
```
```suggestion
<dd>The predicted outputs with shape [batch_size, class_size], or [batch_size, class_size, D1, D2 , ..., Dk], where K is the number of dimensions.</dd>
```
because we already use `d1`, ..., `dk` as indexes above.

onnx/test/shape_inference_test.py
```diff
@@ -2992,6 +2992,22 @@ def test_negative_log_likehood_input_weight_shape_mismatch(self):  # type: () ->
             [make_node('NegativeLogLikelihoodLoss', ['input', 'target', 'weight'], ['loss'], reduction='none')],
             [])
         self.assertRaises(checker.ValidationError, self._inferred, graph)
+   
+   def test_cross_entropy_none(self):  # type: () -> None
+        graph = self._make_graph(
```
wschin: We need some tests with N-D input.

docs/Changelog.md
KsenijaS: I think it is Div(Exp(X - Max(X)), ReduceSum(Exp(X-Max(X)), axis=1, keepdims=True))

docs/Changelog.md
KsenijaS: yes, but when there is a weighted mean, then mean is not averaged by the number of elements in the output but by the sum of weights

KsenijaS: above there is a description that applies to all reduction types and in case of reduction=mean it has to be averaged by sum of weights as well

onnx/test/shape_inference_test.py
KsenijaS: import torch
import torch.nn as nn
import numpy as np
from numpy.testing import assert_allclose

def softmaxcrossentropy(x, target, weight=None, reduction='mean'):  # type: (np.ndarray) -> np.ndarray
    max_x = np.max(x, axis=1, keepdims=True)
    exp_x = np.exp(x - max_x)
    p = exp_x / np.sum(exp_x, axis=1, keepdims=True)
    inp = np.log(p)
    input_shape = inp.shape
    if len(input_shape) == 2:
        N, C = input_shape
        for i in range(N):
            neg_gather_element_input[i] = -inp[i][target[i]]
    if len(input_shape) == 3:
        N, C, D = input_shape
        neg_gather_element_input = np.zeros((N, D), dtype=np.float32)
        for i in range(N):
            for d in range(D):
                neg_gather_element_input[i][d] = -inp[i][target[i][d]][d]

    if weight is not None:
        gather_weight = np.take(weight, target)
        loss = gather_weight * loss
        if reduction == 'mean':
            return loss.sum() / gather_weight.sum()

    if reduction == 'mean':
        loss = np.mean(loss)
    if reduction == 'sum':
        loss = np.sum(loss)

    return loss

loss = nn.CrossEntropyLoss()
input = torch.randn(3, 5)
target = torch.empty(3, dtype=torch.long).random_(5)
output = loss(input, target)

my_loss = softmaxcrossentropy(input.numpy(), target.numpy())
assert_allclose(my_loss, output.numpy(), rtol=1e-05)

KsenijaS: Test numerical values:
```
import torch
import torch.nn as nn
import numpy as np
from numpy.testing import assert_allclose

def softmaxcrossentropy(x, target, weight=None, reduction='mean'):  # type: (np.ndarray) -> np.ndarray
    max_x = np.max(x, axis=1, keepdims=True)
    exp_x = np.exp(x - max_x)
    p = exp_x / np.sum(exp_x, axis=1, keepdims=True)
    inp = np.log(p)
    input_shape = inp.shape
    if len(input_shape) == 2:
        N, C = input_shape
        for i in range(N):
            neg_gather_element_input[i] = -inp[i][target[i]]
    if len(input_shape) == 3:
        N, C, D = input_shape
        neg_gather_element_input = np.zeros((N, D), dtype=np.float32)
        for i in range(N):
            for d in range(D):
                neg_gather_element_input[i][d] = -inp[i][target[i][d]][d]

    if weight is not None:
        gather_weight = np.take(weight, target)
        loss = gather_weight * loss
        if reduction == 'mean':
            return loss.sum() / gather_weight.sum()

    if reduction == 'mean':
        loss = np.mean(loss)
    if reduction == 'sum':
        loss = np.sum(loss)

    return loss

loss = nn.CrossEntropyLoss()
input = torch.randn(3, 5)
target = torch.empty(3, dtype=torch.long).random_(5)
output = loss(input, target)

my_loss = softmaxcrossentropy(input.numpy(), target.numpy())
assert_allclose(my_loss, output.numpy(), rtol=1e-05)
```

docs/Changelog.md
wschin: Sorry. So there are two behaviors? What are equations of them?

KsenijaS: For mean, yes. If weights is provided it's divided by sum of weights otherwise by the number of elements in output.

onnx/defs/math/defs.cc
```diff
@@ -2038,4 +2038,121 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    l[i][d1][d2]...[dk] = -y[i][c][d1][d2]..[dk] * weights[c], if 'weights' is provided.
+
+where:
+    p = Softmax(scores)
+    y = log(p)
+    c = labels[i][d1][d2]...[dk]
+
+Finally, L is optionally reduced:
+L = ReduceSum(L), if reduction = 'sum';
+    ReduceMean(L), if reduction = 'mean'; if "weight" is provided, output is averaged by sum of weights.
```
gramalingam: May be useful to change "output is averaged by sum of weights" to "output is weighted-mean".

SherlockNoMad:
I just realized that the second output is missing in this spec. 
The second optional output should be log_prob = logsoftmax(input), this will output will be used for speeding up the backward computation. 

See https://aiinfra.visualstudio.com/Lotus/_git/onnxruntime/pullrequest/5592 for details

linkerzhang:
@ebarsoum @postrational any more comments on this PR?

PR #2574, opened 2020-01-30
Copy sizes in some optimizers to remain shape information

daquexian:
In some optimizers, the shape information is discarded if the `Value` is destroyed. That makes the model broken if the destroyed value is an output of the graph, as the new output doesn't have corresponding `ValueInfoProto` in pb.

For example, applying "eliminate_identity" optimizer on the model https://drive.google.com/open?id=1EEencrovhayXTuqVL-zzV5eD_NoBEpfm makes `onnx.checker.check_model` fail.

This patch fixes it.


onnx/optimizer/passes/eliminate_identity.h
```diff
@@ -24,6 +24,10 @@ struct EliminateIdentity final : public PredicateBasedPass {
   }
   bool runTransform(Node* node, Graph&, NodeDestroyType& destroy_current)
       override {
+
+    if (node->output()->has_sizes()) {
```
wschin: Do we have tests for the new logic?

onnx/test/optimizer_test.py
```diff
@@ -122,6 +122,21 @@ def check_identity(node):  # type: (NodeProto) -> None
...
+            [add, identity],
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (5,)),
+             helper.make_tensor_value_info("Y", TensorProto.FLOAT, (5,))],
+            [helper.make_tensor_value_info("B", TensorProto.FLOAT, (5,))])
+        optimized_model = self._optimized(graph, ["eliminate_identity"])
+
+        for node in optimized_model.graph.node:
+            assert node.op_type != "Identity"
+        assert len(optimized_model.graph.node) == 1
```
wschin: Would it be suitable to check if output shape is valid here?

onnx/test/optimizer_test.py
```diff
@@ -122,6 +122,21 @@ def check_identity(node):  # type: (NodeProto) -> None
         assert len(optimized_model.graph.node[2].attribute[0].g.output) == 2
         assert optimized_model.graph.node[2].attribute[0].g.output[1].name == "_Y"
 
+    def test_eliminate_identity_graph_output(self):  # type: () -> None
+        add = helper.make_node("Add", ["X", "Y"], ["A"])
+        identity = helper.make_node("Identity", ["A"], ["B"])
```
wschin: This is a great test for `Identity`. May we have tests for `Dropout`, `Pad`, and `Transpose` as well? Thanks.

onnx/optimizer/passes/eliminate_identity.h
daquexian: Could you elaborate, please? What should be tested against? :)

onnx/test/optimizer_test.py
daquexian: > Does the current master branch break this test?

The current master branch will fail on [this method](https://github.com/onnx/onnx/pull/2574/files#diff-0653f936835022eecf61d5127741c081R134), which checks the optimized model and find the output shape is missing.

onnx/test/optimizer_test.py
daquexian: Sure :) Thanks.

daquexian: Added

Issue #1646, opened 2018-11-25
New Operators Request List


houseroad:
We are maintaining a list to track the requests for new operators. Please post your request here.

- [ ] Less or equal than (https://github.com/onnx/onnx/issues/870)
- [ ] Greater or equal than (https://github.com/onnx/onnx/issues/870)
- [ ] CELU (https://github.com/onnx/onnx/issues/1121)
- [ ] FFT (https://github.com/onnx/onnx/issues/298)
- [ ] RFFT (https://github.com/onnx/onnx/issues/298)
- [ ] IFFT (https://github.com/onnx/onnx/issues/298)
- [ ] IRFFT (https://github.com/onnx/onnx/issues/298)
- [ ] DCT (https://github.com/onnx/onnx/issues/298)
- [ ] IDCT (https://github.com/onnx/onnx/issues/298)

michael200892458:
BBoxTransform
RoIAlign
BBoxTransform
BoxWithNMSLimit
GenerateProposals
CopyCPUToGPU
CopyGPUToCPU
ResizeNearest

the above operators is not support

prasanthpul:
RoiAlign is here: https://github.com/onnx/onnx/pull/1869

linkerzhang:
ResizeNearest is here: https://github.com/onnx/onnx/blob/master/docs/Operators.md#Resize.

linkerzhang:
I'll put a question mark on whether we need CopyCPUToGPU and CopyGPUToCPU, which may be taken as a runtime implementation detail, instead of a model representation requirement.

gramalingam:
Agree, we shouldn't have CopyCPUtoGPU or CopyGPUToCPU. They are runtime implementation concerns.

fuzzyBatman:
[CropAndResize](https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize) (tensorflow)

Current work
- Issue #2100 
- PR #2102

fuzzyBatman:
[InvertPermutation](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/invert-permutation) (tensorflow)

jeremycochoy:
Is anyone working on FFT/IFFT/RFFT/IRFFT? I'd be happy to add them otherwise.

garymm:
Fourier transform (signal) ops are tracked by #3573.
All the other ops from the original list were added at some point. I don't think this issue is useful anymore so closing.

PR #2575, opened 2020-01-30
Implement CELU node as a Function

jeremycochoy:
I had a look at your guidelines and tutorial for adding missing op according to https://github.com/onnx/onnx/issues/1121#issuecomment-519319483 .

# Description:

The CELU operator has been required at this issue https://github.com/onnx/onnx/issues/1121 and is now part of the new operator request list https://github.com/onnx/onnx/issues/1646 .

First introduced in [Continuously Differentiable Exponential Linear Units](https://arxiv.org/abs/1704.07483) the CELU is similar to the ELU operation.

Given the attribute α, CELU is a pointwise application of the following formula:
```
CELU(x)=max(0,x)+min(0,α*(exp(x/α)−1))
```
and allow `leakage` of the gradient in the negative values, while having a differential remaining continuous for any value of alpha (which is not the case of ELU).

It is [implemented in Pytorch](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/Activation.cpp) based on the Pytorch-ELU operation:
```
Tensor celu(const Tensor & self, Scalar alpha) {
  double inv_alpha = 1. / alpha.to<double>();
  return at::elu(self, alpha, Scalar(1.0), Scalar(inv_alpha));
}
```
A similar approach for ONNX-ELU is `alpha * ELU(x / alpha, alpha=1)`.

An alternative implementation in numpy, not requiring the Pytorch-ELU operator is given in the tests:
```
import numpy as np

input_data = np.random.randn(1, 2, 3)
alpha = 2

positive_input = np.maximum(0, input_data)
negative_input = np.minimum(0, alpha * (np.exp(input_data / alpha) - 1))
output_data = positive_input + negative_input
```

A first implementation was intended in https://github.com/onnx/onnx/pull/1676 but never merged.

# Graph
The CELU function is implemented using the expression `alpha * ELU(x / alpha, alpha=1)`. This make the graph smaller (and easier to read) than using each individual functions (sum, dub, div, exp, mult) present in the expression of the operator. It also leverage good supports of `Elu in most onnx backend implementations.

# Tests
A unit test and a shape inference test are available following the tests of `MeanVarianceNormalization` function.

### DIFF
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2216,11 +2216,14 @@ ONNX_OPERATOR_SET_SCHEMA(
            "Constrain input and output types to floating-point tensors.")
        .FunctionBody(FunctionBodyHelper::BuildNodes(
            {// nodes: {outputs, op, inputs, attributes}
+            //FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {{"value", 1.f}}},
+	    //FunctionBodyHelper::Const<float>("alpha", 1.0f),
+            //FunctionBodyHelper::Const<float>("Y", 1.0f),
+	    FunctionBodyHelper::NodeDef{{"Y"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}},
             {{"X_alpha"},
              "Div",
-             {"X"},
-             {MakeRefAttribute("alpha", AttributeProto::FLOAT)}
-	    },
-            {{"Y"}, "Elu", {"X_alpha"}}})));
+             {"X", "alpha"}
+            },
+            {{"U"}, "Elu", {"X_alpha"}}})));
 
 } // namespace ONNX_NAMESPACE
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2192,7 +2192,7 @@ ONNX_OPERATOR_SET_SCHEMA(
 static const char* celu_ver12_doc = R"DOC(
        Continuously Differentiable Exponential Linear Units:
        Perform the linear unit element-wise on the input tensor X
-       using formula: <br/> ``` max(0,x) + min(0,α*(exp(x/α)−1)) ```
+       using formula: <br/> ``` max(0,x) + min(0,alpha*(exp(x/alpha)−1)) ```
 )DOC";
 
 static float celu_default_alpha = 1.0;
@@ -2216,14 +2216,19 @@ ONNX_OPERATOR_SET_SCHEMA(
            "Constrain input and output types to floating-point tensors.")
        .FunctionBody(FunctionBodyHelper::BuildNodes(
            {// nodes: {outputs, op, inputs, attributes}
-            //FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {{"value", 1.f}}},
-	    //FunctionBodyHelper::Const<float>("alpha", 1.0f),
-            //FunctionBodyHelper::Const<float>("Y", 1.0f),
-	    FunctionBodyHelper::NodeDef{{"Y"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}},
-            {{"X_alpha"},
-             "Div",
-             {"X", "alpha"}
-            },
-            {{"U"}, "Elu", {"X_alpha"}}})));
+	    // How can we create a constant node `alpha` containing the value of
+            // the attribute `alpha` as a Tensor and not a scalar?
+            // The following (commented) line do not work:
+	    // FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}},
+	    FunctionBodyHelper::Const<float>("alpha", 2.f),
+	    FunctionBodyHelper::Const<float>("zero", 0),
+            FunctionBodyHelper::Const<float>("one", 1.f),
+	    {{"Positive"}, "Max", {"zero", "X"}},
+	    {{"X_alpha"}, "Div", {"X", "alpha"}},
+	    {{"Exponential"}, "Exp", {"X_alpha"}},
+	    {{"A"}, "Sub", {"Exponential", "one"}},
+	    {{"B"}, "Mul", {"A", "alpha"}},
+	    {{"Negative"}, "Max", {"zero", "B"}},
+	    {{"Y"}, "Add", {"Negative", "Positive"}}})));
 
 } // namespace ONNX_NAMESPACE
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2218,7 +2218,7 @@ ONNX_OPERATOR_SET_SCHEMA(
            {// nodes: {outputs, op, inputs, attributes}
             {{"X_alpha"},
              "Div",
-             {"X", "alpha"},
+             {"X", "alpha"}
 	    },
             {{"Y"}, "Elu", {"X_alpha"}}})));
 
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2216,19 +2216,9 @@ ONNX_OPERATOR_SET_SCHEMA(
            "Constrain input and output types to floating-point tensors.")
        .FunctionBody(FunctionBodyHelper::BuildNodes(
            {// nodes: {outputs, op, inputs, attributes}
-	    // How can we create a constant node `alpha` containing the value of
-            // the attribute `alpha` as a Tensor and not a scalar?
-            // The following (commented) line do not work:
-	    // FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}},
-	    FunctionBodyHelper::Const<float>("alpha", 2.f),
-	    FunctionBodyHelper::Const<float>("zero", 0),
-            FunctionBodyHelper::Const<float>("one", 1.f),
-	    {{"Positive"}, "Max", {"zero", "X"}},
-	    {{"X_alpha"}, "Div", {"X", "alpha"}},
-	    {{"Exponential"}, "Exp", {"X_alpha"}},
-	    {{"A"}, "Sub", {"Exponential", "one"}},
-	    {{"B"}, "Mul", {"A", "alpha"}},
-	    {{"Negative"}, "Max", {"zero", "B"}},
-	    {{"Y"}, "Add", {"Negative", "Positive"}}})));
+            FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {MakeRefAttribute("value_float", "alpha", AttributeProto::FLOAT)}},
+            {{"X_alpha"}, "Div", {"X", "alpha"}},
+            {{"Elu_Result"}, "Elu", {"X_alpha"}, {{"alpha", 1.f}}},
+            {{"Y"}, "Mul", {"alpha", "Elu_Result"}}})));
 
 } // namespace ONNX_NAMESPACE
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2189,4 +2189,37 @@ ONNX_OPERATOR_SET_SCHEMA(
              {{"Processed_STD"}, "Add", {"STD", "Epsilon"}},
              {{"Y"}, "Div", {"X_variance", "Processed_STD"}}})));
 
+static const char* celu_ver11_doc = R"DOC(
+       Continuously Differentiable Exponential Linear Units:
+       Perform the linear unit element-wise on the input tensor X
+       using formula: <br/> ``` max(0,x) + min(0,α*(exp(x/α)−1)) ```
+)DOC";
+
+static float celu_default_alpha = 1.0;
+
+ONNX_OPERATOR_SET_SCHEMA(
+    Celu,
+    11,
+    OpSchema()
+       .SetDoc(celu_ver11_doc)
+       .Input(0, "X", "Input tensor", "T")
+       .Output(0, "Y", "Output tensor", "T")
+       .Attr(
+           "alpha",
+           "The Alpha value in Celu formula which control the shape of "
+           "the unit. The default value is 1.0.",
+           AttributeProto::FLOAT,
+           celu_default_alpha)
+       .TypeConstraint(
+           "T",
+           {"tensor(float16)", "tensor(float)", "tensor(double)"},
+           "Constrain input and output types to all numeric tensors.")
+       .FunctionBody(FunctionBodyHelper::BuildNodes(
+           {// nodes: {outputs, op, inputs, attributes}
+            {{"X_alpha"},
+             "Div",
+             {"X", "alpha"},
+	    },
+            {{"Y"}, "Elu", {"X_alpha"}}})));
+
 } // namespace ONNX_NAMESPACE
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2189,36 +2189,4 @@ ONNX_OPERATOR_SET_SCHEMA(
              {{"Processed_STD"}, "Add", {"STD", "Epsilon"}},
              {{"Y"}, "Div", {"X_variance", "Processed_STD"}}})));
 
-static const char* celu_ver12_doc = R"DOC(
-       Continuously Differentiable Exponential Linear Units:
-       Perform the linear unit element-wise on the input tensor X
-       using formula: <br/> ``` max(0,x) + min(0,alpha*(exp(x/alpha)−1)) ```
-)DOC";
-
-static float celu_default_alpha = 1.0;
-
-ONNX_OPERATOR_SET_SCHEMA(
-    Celu,
-    12,
-    OpSchema()
-       .SetDoc(celu_ver12_doc)
-       .Input(0, "X", "Input tensor", "T")
-       .Output(0, "Y", "Output tensor", "T")
-       .Attr(
-           "alpha",
-           "The Alpha value in Celu formula which control the shape of "
-           "the unit. The default value is 1.0.",
-           AttributeProto::FLOAT,
-           celu_default_alpha)
-       .TypeConstraint(
-           "T",
-           {"tensor(float16)", "tensor(float)", "tensor(double)"},
-           "Constrain input and output types to floating-point tensors.")
-       .FunctionBody(FunctionBodyHelper::BuildNodes(
-           {// nodes: {outputs, op, inputs, attributes}
-            FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {MakeRefAttribute("value_float", "alpha", AttributeProto::FLOAT)}},
-            {{"X_alpha"}, "Div", {"X", "alpha"}},
-            {{"Elu_Result"}, "Elu", {"X_alpha"}, {{"alpha", 1.f}}},
-            {{"Y"}, "Mul", {"alpha", "Elu_Result"}}})));
-
 } // namespace ONNX_NAMESPACE
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2189,7 +2189,7 @@ ONNX_OPERATOR_SET_SCHEMA(
              {{"Processed_STD"}, "Add", {"STD", "Epsilon"}},
              {{"Y"}, "Div", {"X_variance", "Processed_STD"}}})));
 
-static const char* celu_ver11_doc = R"DOC(
+static const char* celu_ver12_doc = R"DOC(
        Continuously Differentiable Exponential Linear Units:
        Perform the linear unit element-wise on the input tensor X
        using formula: <br/> ``` max(0,x) + min(0,α*(exp(x/α)−1)) ```
@@ -2199,9 +2199,9 @@ static float celu_default_alpha = 1.0;
 
 ONNX_OPERATOR_SET_SCHEMA(
     Celu,
-    11,
+    12,
     OpSchema()
-       .SetDoc(celu_ver11_doc)
+       .SetDoc(celu_ver12_doc)
        .Input(0, "X", "Input tensor", "T")
        .Output(0, "Y", "Output tensor", "T")
        .Attr(
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2213,7 +2213,7 @@ ONNX_OPERATOR_SET_SCHEMA(
        .TypeConstraint(
            "T",
            {"tensor(float16)", "tensor(float)", "tensor(double)"},
-           "Constrain input and output types to all numeric tensors.")
+           "Constrain input and output types to floating-point tensors.")
        .FunctionBody(FunctionBodyHelper::BuildNodes(
            {// nodes: {outputs, op, inputs, attributes}
             {{"X_alpha"},
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -2218,7 +2218,8 @@ ONNX_OPERATOR_SET_SCHEMA(
            {// nodes: {outputs, op, inputs, attributes}
             {{"X_alpha"},
              "Div",
-             {"X", "alpha"}
+             {"X"},
+             {MakeRefAttribute("alpha", AttributeProto::FLOAT)}
 	    },
             {{"Y"}, "Elu", {"X_alpha"}}})));
 
--- a/onnx/defs/operator_sets.h
+++ b/onnx/defs/operator_sets.h
@@ -635,6 +635,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Pad);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Gemm);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, If);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, NonMaxSuppression);
+class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Celu);
 
 // Iterate over schema from ai.onnx version 11
 class OpSet_Onnx_ver11 {
@@ -706,6 +707,7 @@ class OpSet_Onnx_ver11 {
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Gemm)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, If)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, NonMaxSuppression)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Celu)>());
   }
 };
 
--- a/onnx/defs/operator_sets.h
+++ b/onnx/defs/operator_sets.h
@@ -635,7 +635,6 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Pad);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Gemm);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, If);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, NonMaxSuppression);
-class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Celu);
 
 // Iterate over schema from ai.onnx version 11
 class OpSet_Onnx_ver11 {
@@ -707,7 +706,6 @@ class OpSet_Onnx_ver11 {
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Gemm)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, If)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, NonMaxSuppression)>());
-    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Celu)>());
   }
 };
 
@@ -722,6 +720,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, NegativeLogLikelihoodLoss);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, Dropout);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, BatchNormalization);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, Constant);
+class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, Celu);
 
 // Iterate over schema from ai.onnx version 12
 class OpSet_Onnx_ver12 {
@@ -737,6 +736,7 @@ class OpSet_Onnx_ver12 {
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, Dropout)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, BatchNormalization)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, Constant)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, Celu)>());
   }
 };
 
--- a/onnx/shape_inference/implementation.cc
+++ b/onnx/shape_inference/implementation.cc
@@ -336,7 +336,9 @@ void InferShapeForFunctionNode(
     for (auto attr : n.attribute()) {
       if (attr.has_ref_attr_name()) {
         if (attr_map.count(attr.ref_attr_name())) {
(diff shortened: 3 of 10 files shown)


onnx/defs/nn/defs.cc
```diff
@@ -2074,4 +2074,37 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+           celu_default_alpha)
+       .TypeConstraint(
+           "T",
+           {"tensor(float16)", "tensor(float)", "tensor(double)"},
+           "Constrain input and output types to all numeric tensors.")
+       .FunctionBody(FunctionBodyHelper::BuildNodes(
+           {// nodes: {outputs, op, inputs, attributes}
+            {{"X_alpha"},
+             "Div",
+             {"X", "alpha"},
```
jeremycochoy: I can't figure out how to reference the attribute of the `Celu` instruction as an argument of the node `Div`.

I had a look at the helpers `FunctionBodyHelper::BuildNodes`, `FunctionBodyHelper::Const` and `MakeRefAttribute` without success.

Could you show me some documentation / example of this usage?

onnx/defs/nn/defs.cc
```diff
@@ -2074,4 +2074,37 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+           celu_default_alpha)
+       .TypeConstraint(
+           "T",
+           {"tensor(float16)", "tensor(float)", "tensor(double)"},
+           "Constrain input and output types to all numeric tensors.")
+       .FunctionBody(FunctionBodyHelper::BuildNodes(
+           {// nodes: {outputs, op, inputs, attributes}
+            {{"X_alpha"},
+             "Div",
+             {"X", "alpha"}
```
jeremycochoy: I can't figure out how to reference the attribute of the `Celu` instruction as an argument of the node `Div`.

I had a look at the helpers `FunctionBodyHelper::BuildNodes`, `FunctionBodyHelper::Const` and `MakeRefAttribute` without success.

Could you show me some documentation / example of this usage?

TMVector: There is a comment about this on `FunctionBodyHelper::BuildNodes` [here](https://github.com/onnx/onnx/blob/121114107c38ec49bab4955c13cb82b7675bfc47/onnx/defs/function.h#L79-L80), and it's used in the `MeanVarianceNormalization` operator function [here](https://github.com/onnx/onnx/blob/121114107c38ec49bab4955c13cb82b7675bfc47/onnx/defs/nn/defs.cc#L2057-L2075). Shout if you have trouble :)

onnx/defs/operator_sets.h
```diff
@@ -635,6 +635,7 @@ class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Pad);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Gemm);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, If);
 class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, NonMaxSuppression);
+class ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 11, Celu);
```
TMVector: This should be in version 12, as the [latest released version](https://github.com/onnx/onnx/releases/tag/v1.6.0) is 11 :slightly_smiling_face:

onnx/defs/nn/defs.cc
```diff
@@ -2074,4 +2074,37 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+       .Attr(
+           "alpha",
+           "The Alpha value in Celu formula which control the shape of "
+           "the unit. The default value is 1.0.",
+           AttributeProto::FLOAT,
+           celu_default_alpha)
+       .TypeConstraint(
+           "T",
+           {"tensor(float16)", "tensor(float)", "tensor(double)"},
+           "Constrain input and output types to all numeric tensors.")
```
TMVector: This doesn't include integer tensors.
```suggestion
           "Constrain input and output types to floating-point tensors.")
```

onnx/test/shape_inference_test.py
```diff
@@ -2792,6 +2792,14 @@ def test_einsum_incorrect_num_inputs(self):  # type: () -> None
             [])
         self.assertRaises(checker.ValidationError, self._inferred, graph)
 
+    def test_celu_function_output_shape(self):  # type: () -> None
+        graph = self._make_graph(
+            [('X', TensorProto.FLOAT, (25, 48, 16, 16))],
+            [make_node('Celu', 'X', 'Y', alpha=2.0)],
```
TMVector:
```suggestion
            [make_node('Celu', ['X'], ['Y'], alpha=2.0)],
```

onnx/defs/nn/defs.cc
jeremycochoy: Thanks you for your answer. :)

Unfortunately I did read this line and the usage in MeanVarianceNormalization but I am still confused. I tried different syntax that compile, but since the shape inference test fail it is probably not right graph.

As I understand, I can create a graph equivalent to `Div(X, alpha=alpha)` using
```
           {{"X_alpha"},
             "Div",
             {"X"},
             {MakeRefAttribute("alpha", AttributeProto::FLOAT)}
            },
```
But it doesn't seams to be what I am lloking for, probably because `Div` have two input and 0 attribute.

How can write the equivalent of `Div(X, alpha)` (i.e. use this reference as the second argument of Div)? I would like to write
```
           {{"X_alpha"},
             "Div",
             {"X", MakeRefAttribute("alpha", AttributeProto::FLOAT)}
            }
```
but obviously this is not possible since `std::String != AttributeProtoWrapper` 😅

In the MeanVarianceNormalization it seams all the usage of axis simply forward the attribute to the underlying Ops, right ?

jeremycochoy: Thanks you for your answer. :)

Unfortunately I did read this line and the usage in MeanVarianceNormalization but I am still confused. I tried different syntax that compile, but since the shape inference test fail it is probably not right graph.

As I understand, I can create a graph equivalent to `Div(X, alpha=alpha)` using
```
           {{"X_alpha"},
             "Div",
             {"X"},
             {MakeRefAttribute("alpha", AttributeProto::FLOAT)}
            },
```
But it doesn't seams to be what I am lloking for, probably because `Div` have two input and 0 attribute.

How can write the equivalent of `Div(X, alpha)` (i.e. use this reference as the second argument of Div)? I would like to write
```
           {{"X_alpha"},
             "Div",
             {"X", MakeRefAttribute("alpha", AttributeProto::FLOAT)}
            }
```
but obviously this is not possible since `std::String != AttributeProtoWrapper` 😅

In the MeanVarianceNormalization it seams all the usage of axis simply forward the attribute to the underlying Ops, right ?

TMVector: Ah okay, I see, you are quite right. I think you should be able to move from a reference to a value by adding a `Constant` node.
```
FunctionBodyHelper::BuildNodes(
           {// nodes: {outputs, op, inputs, attributes}
            {{"alpha"}, "Constant", {}, {MakeRefAttribute("alpha", AttributeProto::FLOAT)}},
            {{"X_alpha"}, "Div", {"X", "alpha"}, {}},
            {{"Y"}, "Elu", {"X_alpha"}}})
```

onnx/defs/nn/defs.cc
```diff
@@ -2074,4 +2074,38 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+           "Constrain input and output types to floating-point tensors.")
+       .FunctionBody(FunctionBodyHelper::BuildNodes(
+           {// nodes: {outputs, op, inputs, attributes}
+            {{"X_alpha"},
+             "Div",
+             {"X"},
+             {MakeRefAttribute("alpha", AttributeProto::FLOAT)}
+	    },
+            {{"Y"}, "Elu", {"X_alpha"}}})));
+
```
linkerzhang: This function body is NOT a correct "sub-graph" representing the formula you described. Function body is actually a graph to represent the math formula you mentioned with other ops, in this case, it should be "Constant", "Div", "Elu".

jeremycochoy:
Thanks @TMVector and @linkerzhang for your feedback.

 * Regarding the Tensor/Scalar issue raised by TMVector, I can say that the following code pass the shape inference test. But I don't know if it is enough to say if everything is fine if the second argument is a scalar and not a tensor (I don't know if this 1.f isn't converted to a tensor implicitly).
```
          {// nodes: {outputs, op, inputs, attributes}                                                           
            FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {{"value", 1.f}}},                                        
            {{"X_alpha"},
             "Div",
             {"X", "alpha"}
            },
            {{"Y"}, "Elu", {"X_alpha"}}})));
```

 * Regarding the second problem (using the actual attribute):

I made some attempt to create a constant node that recover the `alpha` attribute from the `Celu` operator using AtributeProto. Although the code compile, the shape inference test is a huge failure. In order to understand what is happening, I simplified the body of the function.

If I run the shape inference test with the following body, I get a nice "Y" of empty shape.
```
FunctionBodyHelper::Const<float>("Y", 1.0f)
```
```
E             name: "Y"
E             type {
E               tensor_type {
E                 elem_type: 1
E                 shape {
E                 }
E               }
E             }
```

but if I try to run the shape inference test with the attribute (see code below), then no "Y" is inferred at all.
```
FunctionBodyHelper::NodeDef{{"Y"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}}
```
```
E       AssertionError: ({'X', 'Y'}, {'X'})
E       assert {'X', 'Y'} == {'X'}
E         Extra items in the left set:
E         'Y'
E         Use -v to get the full diff
```

On my side I am stuck. Looking at `Const` and `ToVector` implementation didn't gave me any new idea to test. Do you have any idea of what is happening? Is it related to this Tensor/Scalar problem? 🙃

onnx/defs/nn/defs.cc
```diff
@@ -2074,4 +2074,41 @@ ONNX_OPERATOR_SET_SCHEMA(
              {{"Processed_STD"}, "Add", {"STD", "Epsilon"}},
              {{"Y"}, "Div", {"X_variance", "Processed_STD"}}})));
 
+static const char* celu_ver12_doc = R"DOC(
+       Continuously Differentiable Exponential Linear Units:
+       Perform the linear unit element-wise on the input tensor X
+       using formula: <br/> ``` max(0,x) + min(0,α*(exp(x/α)−1)) ```
```
wschin: Very nice `alpha` symbol but the attribute uses plain text name `alpha`... I am also afraid of that this cool symbol can't be displayed in my VIM.

onnx/defs/nn/defs.cc
```diff
@@ -2074,4 +2074,41 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+           {// nodes: {outputs, op, inputs, attributes}
+            //FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {{"value", 1.f}}},
+	    //FunctionBodyHelper::Const<float>("alpha", 1.0f),
+            //FunctionBodyHelper::Const<float>("Y", 1.0f),
+	    FunctionBodyHelper::NodeDef{{"Y"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}},
+            {{"X_alpha"},
+             "Div",
+             {"X", "alpha"}
+            },
+            {{"U"}, "Elu", {"X_alpha"}}})));
```
wschin: From Pytorch, CELU equation is
```
CELU(x)=max(0,x)+min(0,α∗(exp(x/α)−1))
```
while ELU uses
```
ELU(x)=max(0,x)+min(0,α∗(exp(x)−1))
```
In addition, here the function body is doing
```
ELU(x/α)=max(0,x/α)+min(0,(exp(x/α)−1))
```
which doesn't exactly match Pytorch CELU. Is this expected? Or I miss something?

onnx/defs/operator_sets.h
```diff
@@ -727,6 +728,7 @@ class OpSet_Onnx_ver12 {
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, MaxPool)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, ReduceMax)>());
     fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, ReduceMin)>());
+    fn(GetOpSchema<ONNX_OPERATOR_SET_SCHEMA_CLASS_NAME(Onnx, 12, Celu)>());
```
wschin: Please split this line following his friends above.

onnx/defs/nn/defs.cc
```diff
@@ -2074,4 +2074,41 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+           "the unit. The default value is 1.0.",
+           AttributeProto::FLOAT,
+           celu_default_alpha)
+       .TypeConstraint(
+           "T",
+           {"tensor(float16)", "tensor(float)", "tensor(double)"},
+           "Constrain input and output types to floating-point tensors.")
+       .FunctionBody(FunctionBodyHelper::BuildNodes(
+           {// nodes: {outputs, op, inputs, attributes}
+            //FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {{"value", 1.f}}},
```
wschin: Are the comments here left intendedly?

wschin:
> Thanks @TMVector and @linkerzhang for your feedback.
> 
> * Regarding the Tensor/Scalar issue raised by TMVector, I can say that the following code pass the shape inference test. But I don't know if it is enough to say if everything is fine if the second argument is a scalar and not a tensor (I don't know if this 1.f isn't converted to a tensor implicitly).
> 
> ```
>           {// nodes: {outputs, op, inputs, attributes}                                                           
>             FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {{"value", 1.f}}},                                        
>             {{"X_alpha"},
>              "Div",
>              {"X", "alpha"}
>             },
>             {{"Y"}, "Elu", {"X_alpha"}}})));
> ```
> 
> * Regarding the second problem (using the actual attribute):
> 
> I made some attempt to create a constant node that recover the `alpha` attribute from the `Celu` operator using AtributeProto. Although the code compile, the shape inference test is a huge failure. In order to understand what is happening, I simplified the body of the function.
> 
> If I run the shape inference test with the following body, I get a nice "Y" of empty shape.
> 
> ```
> FunctionBodyHelper::Const<float>("Y", 1.0f)
> ```
> 
> ```
> E             name: "Y"
> E             type {
> E               tensor_type {
> E                 elem_type: 1
> E                 shape {
> E                 }
> E               }
> E             }
> ```
> 
> but if I try to run the shape inference test with the attribute (see code below), then no "Y" is inferred at all.
> 
> ```
> FunctionBodyHelper::NodeDef{{"Y"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}}
> ```
> 
> ```
> E       AssertionError: ({'X', 'Y'}, {'X'})
> E       assert {'X', 'Y'} == {'X'}
> E         Extra items in the left set:
> E         'Y'
> E         Use -v to get the full diff
> ```
> 
> On my side I am stuck. Looking at `Const` and `ToVector` implementation didn't gave me any new idea to test. Do you have any idea of what is happening? Is it related to this Tensor/Scalar problem? 🙃

As described [here](https://github.com/onnx/onnx/blob/master/docs/Operators.md#attributes-9), the `value` attribute should be an `tensor`, not a `float`.

onnx/defs/nn/defs.cc
jeremycochoy: 😱

You are completely right, the ELU implementation of Pytorch is different from ONNX Elu, and it is not possible to express CELU from ONNX's ELU. Thanks you for noticing it, I am working on a fix.

onnx/defs/nn/defs.cc
jeremycochoy: Yes, I need some information on how to convert the `Scalar` Attribute into a Constant `Tensor`. Do you know how to accomplish this?

onnx/defs/operator_sets.h
jeremycochoy: Could you develop? I don't understand which friends. 🙂

jeremycochoy:
> > Thanks @TMVector and @linkerzhang for your feedback.
> > 
> > * Regarding the Tensor/Scalar issue raised by TMVector, I can say that the following code pass the shape inference test. But I don't know if it is enough to say if everything is fine if the second argument is a scalar and not a tensor (I don't know if this 1.f isn't converted to a tensor implicitly).
> > 
> > ```
> >           {// nodes: {outputs, op, inputs, attributes}                                                           
> >             FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {{"value", 1.f}}},                                        
> >             {{"X_alpha"},
> >              "Div",
> >              {"X", "alpha"}
> >             },
> >             {{"Y"}, "Elu", {"X_alpha"}}})));
> > ```
> > 
> > 
> > 
> > * Regarding the second problem (using the actual attribute):
> > 
> > I made some attempt to create a constant node that recover the `alpha` attribute from the `Celu` operator using AtributeProto. Although the code compile, the shape inference test is a huge failure. In order to understand what is happening, I simplified the body of the function.
> > If I run the shape inference test with the following body, I get a nice "Y" of empty shape.
> > ```
> > FunctionBodyHelper::Const<float>("Y", 1.0f)
> > ```
> > 
> > 
> > ```
> > E             name: "Y"
> > E             type {
> > E               tensor_type {
> > E                 elem_type: 1
> > E                 shape {
> > E                 }
> > E               }
> > E             }
> > ```
> > 
> > 
> > but if I try to run the shape inference test with the attribute (see code below), then no "Y" is inferred at all.
> > ```
> > FunctionBodyHelper::NodeDef{{"Y"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}}
> > ```
> > 
> > 
> > ```
> > E       AssertionError: ({'X', 'Y'}, {'X'})
> > E       assert {'X', 'Y'} == {'X'}
> > E         Extra items in the left set:
> > E         'Y'
> > E         Use -v to get the full diff
> > ```
> > 
> > 
> > On my side I am stuck. Looking at `Const` and `ToVector` implementation didn't gave me any new idea to test. Do you have any idea of what is happening? Is it related to this Tensor/Scalar problem? 🙃
> 
> As described [here](https://github.com/onnx/onnx/blob/master/docs/Operators.md#attributes-9), the `value` attribute should be an `tensor`, not a `float`.

Unfortunately, after hours digging documentation and code, I can't figure a way to convert a scalar (from `MakeRefAttribute("alpha", Attribute\
Proto::FLOAT)`) to a tensor constant.


PS: If this is not possible, then maybe there is still a way to cheat with the `Gemm` instruction (it is the only instruction I found which take a scalar attribute and do his product with a tensor). But I would need some help to create the 1x1 tensor input matrices.

onnx/defs/nn/defs.cc
jeremycochoy: Here is a code testing the differences between Pytorch CELU and the implementation (with corrected parenthesis) I provided.

```
import numpy as np
import torch

def onnx_celu(input_data, alpha=1.0):
    positive_input = np.maximum(0, input_data)
    negative_input = np.minimum(0, alpha * (np.exp(input_data / alpha) - 1))
    output_data = positive_input + negative_input
    return output_data

def torch_celu(input_data, alpha=1.0):
    return torch.nn.CELU(alpha=alpha)(torch.Tensor(input_data)).numpy()

input_data = np.random.randn(1, 2, 3).astype('float32')
alpha = 2

assert (onnx_celu(input_data, alpha) == torch_celu(input_data, alpha)).all()

```

onnx/defs/operator_sets.h
jeremycochoy: On Github the lines appear split but in the actual file (from current master), each op is on a sigle line. Or did I missed something?

wschin:
> > > Thanks @TMVector and @linkerzhang for your feedback.
> > > 
> > > * Regarding the Tensor/Scalar issue raised by TMVector, I can say that the following code pass the shape inference test. But I don't know if it is enough to say if everything is fine if the second argument is a scalar and not a tensor (I don't know if this 1.f isn't converted to a tensor implicitly).
> > > 
> > > ```
> > >           {// nodes: {outputs, op, inputs, attributes}                                                           
> > >             FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {{"value", 1.f}}},                                        
> > >             {{"X_alpha"},
> > >              "Div",
> > >              {"X", "alpha"}
> > >             },
> > >             {{"Y"}, "Elu", {"X_alpha"}}})));
> > > ```
> > > 
> > > 
> > > 
> > > * Regarding the second problem (using the actual attribute):
> > > 
> > > I made some attempt to create a constant node that recover the `alpha` attribute from the `Celu` operator using AtributeProto. Although the code compile, the shape inference test is a huge failure. In order to understand what is happening, I simplified the body of the function.
> > > If I run the shape inference test with the following body, I get a nice "Y" of empty shape.
> > > ```
> > > FunctionBodyHelper::Const<float>("Y", 1.0f)
> > > ```
> > > 
> > > 
> > > ```
> > > E             name: "Y"
> > > E             type {
> > > E               tensor_type {
> > > E                 elem_type: 1
> > > E                 shape {
> > > E                 }
> > > E               }
> > > E             }
> > > ```
> > > 
> > > 
> > > but if I try to run the shape inference test with the attribute (see code below), then no "Y" is inferred at all.
> > > ```
> > > FunctionBodyHelper::NodeDef{{"Y"}, "Constant", {}, {MakeRefAttribute("value", AttributeProto::FLOAT, "alpha")}}
> > > ```
> > > 
> > > 
> > > ```
> > > E       AssertionError: ({'X', 'Y'}, {'X'})
> > > E       assert {'X', 'Y'} == {'X'}
> > > E         Extra items in the left set:
> > > E         'Y'
> > > E         Use -v to get the full diff
> > > ```
> > > 
> > > 
> > > On my side I am stuck. Looking at `Const` and `ToVector` implementation didn't gave me any new idea to test. Do you have any idea of what is happening? Is it related to this Tensor/Scalar problem? 🙃
> > 
> > 
> > As described [here](https://github.com/onnx/onnx/blob/master/docs/Operators.md#attributes-9), the `value` attribute should be an `tensor`, not a `float`.
> 
> Unfortunately, after hours digging documentation and code, I can't figure a way to convert a scalar (from `MakeRefAttribute("alpha", Attribute\ Proto::FLOAT)`) to a tensor constant.
> I left a comment pointing the the problematic line in the body of the function.
> 
> PS: If this is not possible, then maybe there is still a way to cheat with the `Gemm` instruction (it is the only instruction I found which take a scalar attribute and do his product with a tensor). But I would need some help to create the 1x1 tensor input matrices.

I will try something on my side. In the meanwhile, what do you think if we make `alpha` an input?

linkerzhang:
I think "I can't figure a way to convert a scalar (from MakeRefAttribute("alpha", Attribute\ Proto::FLOAT)) to a tensor constant" needs to be fixed. Logically, the body graph is referring an attribute outside (which should be an AttributeProto) and the "Constant" OP will use the attribute and output a Tensor.

jeremycochoy:
@wschin Personally, I really don't mind moving alpha as an input. But it may be very confusing for both developper of user if CELU and ELU have completely different interface. If this approach get merged, it also means supporting it for a long time. 😅

@linkerzhang Would be awesome. I think anyone who implement a new function op that do not directly forward its arguments will end up having the exact same problem, and a clean way to move the scalars into the graph would solve this. Do you have any idea on how this could be archived?

linkerzhang:
@jeremycochoy The AttributeProto itself was designed to support this kind of reference already, though the utility function is missing, I guess.

PR #2583 should resolve it.

MakeRefAttribute("value", "alpha", AttributeProto::FLOAT) for the Constant node in the function body.

TMVector:
@linkerzhang I think that will work if the `alpha` attribute is a tensor, but ideally it would be a naked float. Maybe `Constant` should be changed to promote non-tensor values to scalar tensors?

jeremycochoy:
> @jeremycochoy The AttributeProto itself was designed to support this kind of reference already, though the utility function is missing, I guess.
> 
> PR #2583 should resolve it.
> 
> MakeRefAttribute("value", "alpha", AttributeProto::FLOAT) for the Constant node in the function body.

@linkerzhang 
Isn't essentially the same thing of https://github.com/onnx/onnx/pull/2575/files#diff-8073bde925403bcdfa7d23c68d914d97 present in the current PR?

wschin:
> @linkerzhang I think that will work if the `alpha` attribute is a tensor, but ideally it would be a naked float. Maybe `Constant` should be changed to promote non-tensor values to scalar tensors?

We might need to support `floats` in addition to `float`. The fundamental here is that `Attribute` and `Graph` use different numerical type systems. `Attribute` has `float`, `floats`, and `tensor`. `Graph` only has `tensor`. I think it will be a nice and small change to bridge these two systems.

jeremycochoy:
> > @linkerzhang I think that will work if the `alpha` attribute is a tensor, but ideally it would be a naked float. Maybe `Constant` should be changed to promote non-tensor values to scalar tensors?
> 
> We might need to support `floats` in addition to `float`. The fundamental cause here is that `Attribute` and `Graph` use different numerical type systems. `Attribute` has `float`, `floats`, and `tensor`. `Graph` only has `tensor`. I think changing `Constant` will be a nice and small change to bridge these two systems -- because `Attribute` is always a constant in graphs.
> 
> @TMVector, @jeremycochoy, any comments?

To me it sounds the best solution, and it definitively makes sense to convert both float and floats to their corresponding tensors.

linkerzhang:
@jeremycochoy yep, it's the same as the one in current PR (I missed the part).

One more solution is removing AttributeProto and having TensorProto be used to store Attribute, to unify the two type systems (one tensor type system and one attribute type system).

onnx/defs/attr_proto_util.cc
```diff
@@ -58,4 +58,15 @@ AttributeProto MakeRefAttribute(
   return a;
```
linkerzhang: change this function to call the overridden one added below please.

jeremycochoy: Oh, I think you can just merge your PR and I can rebase this branch on top of it. I remember you introduced documentation, your ordering of arguments sounds more natural to me, and I have nothing against splitting PRs in smaller piece.

linkerzhang: Aha, I abandoned my PR this morning (realized it's duplicate with changes in your PR). Let me get it back and merge it in this way :)

wschin:
> @jeremycochoy yep, it's the same as the one in current PR (I missed the part).
> 
> One more solution is removing AttributeProto and having TensorProto be used to store Attribute, to unify the two type systems (one tensor type system and one attribute type system).
> 
> AttributeProto and its type system were designed at the very beginning, to introduce a simpler way of having "scalar" attribute data. However, it introduces many troubles when specifying operator spec. For example, it's really hard to specify when attribute needs to share the same type as input or output (now most cases are using "Tensor" attribute type).
> 
> This PR reminds me again that the benefit of AttributeProto and its type system is not that much, while many troubles introduced.
> 
> I'd suggest to remove them and have only one type system in ONNX.
> 
> @gramalingam @wschin @jeremycochoy @TMVector What do you think please?

If we did this in the beginning, that would be perfect. However, I am not sure how the removal of `scalar` attributes could impact existing models.

linkerzhang:
Changing "Constant" op is not a bad idea, however, it also means that multiple attribute types should be allowed when defining an op, which is not supported right now.

The other way is to add utility functions to make AttributeProto and TensorProto transferable from each other, given that they're indeed transferrable.

linkerzhang:
@wschin about backward compatibility, it depends on how to remove the AttributeProto. if we're NOT removing it from the .proto file but are not going to use it in any updated operator definitions, then the IR itself is still backward compatible (old models can still be loaded and run successfully) and operators of course are always backward compatible (as we're having them versioned). Make sense?

jeremycochoy:
I do not understand clearly the implication of a such choice, both regarding backward compatibility, impact on signatures of new and existing operators, and the work it represent of backend implementors.

So I will let you take this hard decision. 😅

gramalingam:
@linkerzhang : re. changing the Constant op: the solution adopted in some ops is to define multiple attributes (foo_int of type int and foo_float of type float) with the requirement that exactly one of these attributes should be defined. We could do this. This would be more compatibility-preserving. If we need this in multiple places, a utility like PR #2584 would be helpful ... but, as mentioned, it is more disruptive in terms of compatibility.

linkerzhang:
@gramalingam I checked the repo and there're only "Constant“ and "ConstantOfShape" use AttributeType::TENSOR, so that your suggestion of adding more attributes to them may be OK.

But again, please guys help to think it twice on PR #2584 whether it's better :).

To unblock this PR, @jeremycochoy do you mind to send a PR per @gramalingam suggested above (adding "value_int", "value_ints", "value_float", "value_floats", "value_string", "value_strings" attributes to Constant op). Thank you!

jeremycochoy:
I will have a look today :)

onnx/shape_inference/implementation.cc
```diff
@@ -336,7 +336,9 @@ void InferShapeForFunctionNode(
     for (auto attr : n.attribute()) {
       if (attr.has_ref_attr_name()) {
         if (attr_map.count(attr.ref_attr_name())) {
-          copy_n.add_attribute()->CopyFrom(*attr_map[attr.ref_attr_name()]);
+          auto copy_attr = *attr_map[attr.ref_attr_name()];
+          copy_attr.set_name(attr.name());
```
jeremycochoy: We really need to rename attribute when `ref_name != name`.

linkerzhang: Good catch!

linkerzhang:
CI needs to be fixed and passed. :)

jeremycochoy:
> CI needs to be fixed and passed. :)

Yes, working on it (i also want to run a graph with this instruction for testing). I will Re-request review when I am done. :)

linkerzhang:
@spandantiwari @houseroad there's pytorch ci failure. You may take a look at it.

jeremycochoy:
> @spandantiwari @houseroad there's pytorch ci failure. You may take a look at it.

From what I understand the CI is having trouble downloading some models (HTTP 403) :( .
```

self = <urllib.request.HTTPDefaultErrorHandler object at 0x7f3137d84978>
req = <urllib.request.Request object at 0x7f3108b8e518>
fp = <http.client.HTTPResponse object at 0x7f3108b8e358>, code = 403
msg = 'Forbidden', hdrs = <http.client.HTTPMessage object at 0x7f3108b8e0b8>

    def http_error_default(self, req, fp, code, msg, hdrs):
>       raise HTTPError(req.full_url, code, msg, hdrs, fp)
E       urllib.error.HTTPError: HTTP Error 403: Forbidden

/usr/lib/python3.6/urllib/request.py:650: HTTPError
```

onnx/defs/nn/defs.cc
```diff
@@ -2189,4 +2189,36 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+       .TypeConstraint(
+           "T",
+           {"tensor(float16)", "tensor(float)", "tensor(double)"},
+           "Constrain input and output types to floating-point tensors.")
+       .FunctionBody(FunctionBodyHelper::BuildNodes(
+           {// nodes: {outputs, op, inputs, attributes}
+            FunctionBodyHelper::NodeDef{{"alpha"}, "Constant", {}, {MakeRefAttribute("value_float", "alpha", AttributeProto::FLOAT)}},
+            {{"X_alpha"}, "Div", {"X", "alpha"}},
+            {{"Elu_Result"}, "Elu", {"X_alpha"}, {{"alpha", 1.f}}},
+            {{"Y"}, "Mul", {"alpha", "Elu_Result"}}})));
```
TMVector: Wouldn't it be equivalent to pass Celu.alpha to Elu.alpha instead of setting Elu.alpha=1 and then multiplying by Celu.alpha?

jeremycochoy: It is not. 😅 (Because the alpha is applied only on the second member of the + operator in the Elu equation).

TMVector: Ah, quite right.

Btw should Celu be defined in `onnx/defs/math/defs.cc`? -- that's where Relu, Elu, etc. are.

jeremycochoy: Right, it would make more sense to place it net to relu / elu / leaky relu. I will change it this evening.

fdwr:
@jeremycochoy : This is an excellent description for a new operator (explaining why it's being added, where it came from, the actual equation used, and even an alternate Python implementation), and I'll point to it as a good example in the future. 👍

PallHaraldsson:
Thanks, I had no idea about the useful Celu. Is there a reason to use `min`? I mean `max` only (for ReLU) is natural, implying one test, and while having both imlplies two, and I wouldn't trust a compiler to know one needed, and that it can be 583 times slower:

```
julia> x = 1.1; α = 1.0

julia> @btime ONNX_CELU($x, $α);
  13.999 ns (0 allocations: 0 bytes)

julia> ONNX_CELU(x, α)=if x >= zero(x) x/α else exp(x/α)-1 end
ONNX_CELU (generic function with 2 methods)

julia> @btime ONNX_CELU($x, $α);
  0.024 ns (0 allocations: 0 bytes)

julia> x = -1.1

julia> @btime ONNX_CELU($x, $α);
  11.627 ns (0 allocations: 0 bytes)
```
Yes, only 20% faster on the other side, and maybe with values all over the place (ca. 50-50% split?) it's not a useful optimization as I would think?

jeremycochoy:
Hi @PallHaraldsson 

To be honest, I am efraid that if you want to have something really optimized you'd need the backend to provide a specific implementation for celu that replace this graph function (this is 100% possible and any backend can chose th implementation that fits it's specific targeted hardware).

PallHaraldsson:
Yes, and FYI, I found an even better activation function (in part since it's also continuously differentiable, why better than ReLU):

Mish: A Self Regularized Non-Monotonic Neural Activation Function
https://arxiv.org/pdf/1908.08681v1.pdf

>In Tensorflow[11], the function definition of Mish can be written as x * tf.math.tanh(tf.softplus(x)) while in Torch[12]it is x * torch.tanh(F.softplus(x)). For improved results over ReLU, it is advised to use a slightly lower learning rate for Mish.

It's also compared to ELU and some variant (while not CELU).

Also interesting, and supposed advantages contrary to those supposed above (such as bounded at low):

PLU: The Piecewise Linear Unit Activation Function
https://arxiv.org/abs/1809.09534


And if you're interested, a very simple idea here (using two "opposite", but similar, I wander if you could do similar for two dissimilar, e.g. those above):

https://arxiv.org/pdf/1709.04054.pdf

>We propose a simple extension to the ReLU-family of activation functions thatallows them to shift the mean activation across a layer towards zero. Combinedwith proper weight initialization, this alleviates the need for normalization layers.We explore the training of deep vanilla recurrent neural networks (RNNs) with up to 144 layers, and show that bipolar activation functions help learning in this setting.  On the Penn Treebank and Text8 language modeling tasks we obtain competitive results, improving on the best reported results for non-gated networks.

PR #2550, opened 2020-01-15
Do not infer shape when input shape is unknown

shinh:
Without this check, a scalar shape is wrongly inferred when
the input has an unknown shape. A test was added for ScatterND
since this op is sufferred by this issue.

PR #2577, opened 2020-01-30
Fix a bug in ScatterND shape inference

BowenBao:
Check if input has shape before propagating.

cc @spandantiwari, @lara-hdr I think this is the root cause for the issue we seen with torch vision.


linkerzhang:
@spandantiwari @houseroad please check the pytorch test failure.

shinh:
I tried to fix the same issue (https://github.com/onnx/onnx/pull/2550/files) but in a different way. I thought this check should be done in `propagateShapeFromInputToOutput` rather than in the callers side because I found no valid usage to call this function without check the shape is available. To show my fix can solve issues in other ops, I've just created another testcase for EyeLike (https://github.com/shinh/onnx/commit/39850769779dcdcf3fe3d4724c97c8eedaae8443). That said, I admit my change was not consistent with existing code and we should probably remove redundant `hasNInputShapes` calls if we get my PR merged. What do you think?

BowenBao:
@shinh Agreed. What you proposed is a cleaner solution. However it may be backward breaking. Let's merge these two PRs, and then remove redundant `hasNInputShapes` `hasInputShape`.

spandantiwari:
cc: @houseroad @postrational

PR #2578, opened 2020-01-31
Adding Inverse Op

neginraoof:
Adding inverse op.
Similar to torch.inverse and numpy.linalg.inv(a), this op only supports square shape tensors, or matrices of shape [*, M, M].


neginraoof:
@gramalingam 
cc @houseroad @postrational
Could you please help review this PR?
Thanks

docs/Operators.md
```diff
@@ -6946,6 +6947,79 @@ expect(node, inputs=[x, s, bias], outputs=[y],
...
+
+<dl>
+<dt><tt>Y</tt> : T</dt>
+<dd>Output tensor</dd>
+</dl>
+
+#### Type Constraints
+
+<dl>
+<dt><tt>T</tt> : tensor(uint8), tensor(uint16), tensor(uint32), tensor(uint64), tensor(int8), tensor(int16), tensor(int32), tensor(int64), tensor(float16), tensor(float), tensor(double)</dt>
```
linkerzhang: for 8-bit integers, do you mean quantized integers or just regular 8 bit integer? if it's quantized, same scale/offset used for input and output?

Input and output can be always the same type?

neginraoof: Thanks! This PR is not targeting quantized tensors. 
I added the type constraint for output (same type as input) in the spec.

docs/Operators.md
```diff
@@ -6946,6 +6947,79 @@ expect(node, inputs=[x, s, bias], outputs=[y],
 </details>
 
 
+### <a name="Inverse"></a><a name="inverse">**Inverse**</a>
+
+  Calculates inverse of a square matrix or batches of square matrices.
+  Inverse takes one input tensor of shape `[*, M, M]`, where `*` is zero or more batch dimensions,
+  and the inner-most 2 dimensions form square matrices.
+  The output is a tensor of shape `[*, M, M]`, containing the individual inverses of all input submatrices.
```
spandantiwari: What happens when the matrix is rank-deficient, or more generally, ill-conditioned with a large condition number? We should clarify what the expectation is in that case.
I would recommend that we keep it flexible for backends and treat this as an undefined behavior. Any backend can choose detect the condition and throw exception, or simply return garbage result (possibly with a warning).

neginraoof: Thanks for pointing this out. I agree that we should expect an exception from backend. Should I mention that rank-deficient matrices are not supported?

neginraoof: I updated the spec noting that the op only supports fill-rank tensors.

onnx/defs/math/defs.cc
```diff
@@ -2038,4 +2038,57 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                (mat_w.dim_value() != mat_h.dim_value())) {
+              fail_shape_inference(
+                  "The inner-most 2 dimensions must have the same size (mat_w:",
+                  mat_w.dim_value(),
+                  " != mat_h:",
+                  mat_h.dim_value(),
+                  ").");
+            }
+
+            for (int i=0; i < rank ; ++i) {
```
gramalingam: Can replace this loop by "propagateShapeFromInputToOutput(ctx, 0, 0);".

neginraoof: Thanks. Replaced it.

neginraoof:
cc @gramalingam @linkerzhang could you please help merge this PR?
Thanks

PR #2581, opened 2020-02-01
Serialization for Sequence and Map data types

vinitra:
Purpose: Add serialization for inputs and outputs of Sequence and Map data types so relevant Sequence and Map operator unit tests can be enabled.

- Proposing new onnx-data proto file with sequences and maps
- Defining SequenceProto
- Defining MapProto
- related functions in cmd_tools, numpy_helper, and helper
- Cleaning comments for whitespace
- Adding example of Sequence test with SequenceInsert unit test

Follow on from: https://github.com/onnx/onnx/pull/2249


vinitra:
Very much an initial attempt, currently. Comments on the proto were sparse, so feedback is much appreciated.

linkerzhang:
Please also show the model example/usage of sequence data serialization in the PR.

onnx/onnx.in.proto
```diff
@@ -497,6 +497,50 @@ message TensorShapeProto {
...
+      optional MapProto map_value = 3;
+    };
+  };
+  repeated SequenceElement values = 1;
+
+  // shape of the underlying sequence
+  repeated int64 dims = 2;
+
+  // The data type of the sequence.
+  optional TypeProto elem_type = 3;
```
linkerzhang: is this needed? will it be conflict with the type in "TensorProto"?

gramalingam:
I see that the newly added message-types are not used anywhere else. So, I am not sure what the usage scenario is. Is the goal to extend "AttributeProto"? Or, is it just to support input/output values of this type (which doesn't really require a serializatio-format)?

onnx/onnx.in.proto
```diff
@@ -497,6 +497,50 @@ message TensorShapeProto {
   repeated Dimension dim = 1;
 }
 
+// Sequences
+//
+// Defines a dense, ordered, collection of elements that are of homogeneous types.
+message SequenceProto {
+  // Sequences are made of TypeProto elements, which can either be tensors,
+  // maps, or sequences themselves.
+  message SequenceElement {
```
gramalingam: I think this and MapElement below should be the same (shared) message-type. (We could probably copy the style used in the definition of TypeProto, in case the mutual recursive dependence causes a problem.) I think we should allow a sparse-tensor here too.

onnx/onnx.in.proto
```diff
@@ -497,6 +497,50 @@ message TensorShapeProto {
...
+  message MapElement {
+    oneof value {
+      TensorProto tensor_value = 1;
+      SequenceProto sequence_value = 2;
+      MapProto map_value = 3;
+    };
+  };
+
+  optional TypeProto value_type = 3;
+  repeated MapElement value = 4;
```
gramalingam: We need repeated KeyValuePairs, consisting of a key and a value.

onnx/onnx.in.proto
gramalingam: Yes, I was wondering the same. Why do we need the "dims" field also? (I think the edge-case where we have an empty-sequence may need it, depending on the use-case. But ideally we should not need this.)

vinitra: Yes, I added dims originally for the empty-sequence case and for ease of use, but it's perhaps overengineered. I'm removing both elem_type and dims.

onnx/onnx.in.proto
vinitra: Added KeyValuePair message type. Thanks for the suggestion.

onnx/onnx.in.proto
```diff
@@ -497,6 +497,45 @@ message TensorShapeProto {
...
+    oneof value {
+      TensorProto tensor_value = 1;
+      SequenceProto sequence_value = 2;
+      MapProto map_value = 3;
+    };
+  };
+
+  // Key-Value pair element
+  message KeyValuePair {
+    required int32 key = 1;
```
TMVector: The key needs to support [u]int types and string.

`TypeProto.Map.key_type` says:
```
    // This field MUST have a valid TensorProto.DataType value
    // This field MUST refer to an integral type ([U]INT{8|16|32|64}) or STRING
    optional int32 key_type = 1;
```

onnx/onnx.in.proto
vinitra: Created shared SequenceMapElement and added sparse tensor support.

vinitra:
> I see that the newly added message-types are not used anywhere else. So, I am not sure what the usage scenario is. Is the goal to extend "AttributeProto"? Or, is it just to support input/output values of this type (which doesn't really require a serializatio-format)?

Discussed with @gramalingam in person. The purpose is to enable unit tests for the operators of Sequence and Map data types that require serialization.

gramalingam:
Thanks @vinitra for explaining the context. I think enabling unit tests with sequence/map data will require further extensions:

(a) The unit-test-case exporter seems to assume that inputs/outputs are tensors: see https://github.com/onnx/onnx/blob/8b3f7e2e7a0f2aba0e629e23d89f07c7fc0e6a5e/onnx/backend/test/case/node/__init__.py#L91 and https://github.com/onnx/onnx/blob/8b3f7e2e7a0f2aba0e629e23d89f07c7fc0e6a5e/onnx/backend/test/cmd_tools.py#L49 

(b) Extending helper.py to convert from a python representation of values to protobuf (e.g., as in make_tensor: see https://github.com/onnx/onnx/blob/8b3f7e2e7a0f2aba0e629e23d89f07c7fc0e6a5e/onnx/helper.py#L144 ) may be helpful too.

spandantiwari:
@vinitra - I agree with @gramalingam on both his points about adding necessary helper/support functionality so that a unit test involving Sequences/Dict can be written reasonably easily by an author. 

Thanks for this welcome change.

BowenBao:
@vinitra Thanks for working on this PR, this is a good start to complete the support for Sequence and Map data types. 
As others have suggested on showing an example usage of this serialization, I'd recommend adding one unit test for the Sequence/Map related operator, e.g. SequenceInsert, to demonstrate the end-to-end usage with support of serialization.

sveta-levitan:
Should this get into ONNX 1.7? If yes, please finalize very soon. Thank you!

vinitra:
> Should this get into ONNX 1.7? If yes, please finalize very soon. Thank you!

ONNX 1.8 is fine! Sorry for the delay on updates. Should be forthcoming shortly.

onnx/onnx.in.proto
vinitra: Addressed in recent update, but unsure if this is the right way to do it.

vinitra: Rewritten as per to Bowen's suggestion of including datatype enum field and storing data in separate key fields similarly to TensorProto format in KeyValuePair.

onnx/backend/test/cmd_tools.py
```diff
@@ -46,12 +46,16 @@ def prepare_dir(path):  # type: (Text) -> None
                     output_dir, 'test_data_set_{}'.format(i))
                 prepare_dir(data_set_dir)
                 for j, input_np in enumerate(inputs):
+                    if input_np.dtype == 'object':
+                        raise TypeError('Sequences not yet supported.')
```
BowenBao, line 50: Seems we need to extend numpy_helper to add conversion from `list[np.array]` to `SequenceProto` and vice versa. And by using `SequenceProto::SerializeToString()` we should be able to store the values to file.

vinitra: Working on this currently. Should have a PR update out soon.

onnx/onnx.in.proto
```diff
@@ -626,6 +626,89 @@ message TensorShapeProto {
...
+  optional TypeProto elem_type = 1;
+  repeated SequenceMapElement values = 2;
+}
+
+// Maps
+//
+// Specifies an associative table, defined by a key type and a value type.
+message MapProto {
+  // Key-Value pair element
+  message KeyValuePair {
```
vinitra, line 655: Should KeyValuePair be moved out of MapProto?

onnx/backend/test/cmd_tools.py
vinitra: Updated! See numpy_helper for the new functions. Also added support for Maps.

onnx/backend/test/cmd_tools.py
```diff
@@ -46,17 +46,31 @@ def prepare_dir(path):  # type: (Text) -> None
                     output_dir, 'test_data_set_{}'.format(i))
                 prepare_dir(data_set_dir)
                 for j, input_np in enumerate(inputs):
-                    tensor = numpy_helper.from_array(
-                        input_np, case.model.graph.input[j].name)
+                    if isinstance(input_np, dict):
+                        arr = numpy_helper.from_dict_to_map(
+                            input_np, case.model.graph.input[j].name)
+                    elif input_np.dtype == 'object':
```
BowenBao, line 52: this seems to be picking `test_strnorm_model_monday_casesensintive_nochangecase` as sequence as well. The input for the above test case is `array(['monday', 'tuesday', 'wednesday', 'thursday'], dtype=object)`.

vinitra: Ah, yes, just noticed strings are treated as dtype=object in np. Will try to identify ideas for how to better specify sequences.

gramalingam: Why not use python dictionaries and lists as the representation for ONNX's map and sequence? Won't that be more natural in Python?

vinitra: I had initially implemented it with dicts and lists, but ran into a significant number of errors in the testing infrastructure as everything was tailored to np.array formats.

vinitra: However, I do think it is the more natural data structure within Python. I'll look into it a bit more.

onnx/backend/test/case/node/sequenceinsert.py
```diff
@@ -0,0 +1,57 @@
...
+
+class SequenceInsert(Base):
+
+    @staticmethod
+    def export():  # type: () -> None
+        test_cases = {
+            'at_back': [np.array([10, 11, 12])],
+            'at_front': [np.array([-2, -1, 0]), np.array([0])]
+        }
+        sequence = np.array([np.array([1, 2, 3, 4]), np.array([5, 6, 7]), np.array([8, 9])])
```
gramalingam, line 34: It would be great if we could just write sequence = [ np.array(...), np.array(...), ...] here. It also clarifies that it is a sequence of tensors.

vinitra: Addressed. Thanks!

gramalingam: I am a bit surprised that this did not require other changes (as discussed elsewhere above about whether we should use arrays or lists for representing sequences)

vinitra: Likewise. This ended up being a point in test implementation, not test infrastructure as I had previously thought. I'll test with ORT and see if it works. Thanks!

vinitra: The changes I was talking about earlier were in the helper functions (i.e. from_array_to_sequence, etc.) Will submit an update.

onnx/backend/test/case/node/__init__.py
```diff
@@ -102,6 +102,13 @@ def function_testcase_helper(node, name):  # type: (NodeProto, Text) -> List[Nod
 
 
 def _extract_value_info(arr, name):  # type: (np.ndarray, Text) -> onnx.ValueInfoProto
+    if arr.dtype == "object":
```
gramalingam, line 105: What happens here if "arr" is a list, instead of an array? (As an aside, the type-annotations above will no longer be valid. Unfortunately, it looks like the python type-checker is not part of the CI ... this is a separate issue, but it does complicate our task here since we can't rely on the type-checker.) I don't understand how things can work without changes here.

onnx/backend/test/case/node/sequenceinsert.py
gramalingam: I think changes might be required in _extract_value_info (as I have commented there) also to make this work.

onnx/backend/test/cmd_tools.py
vinitra: See recent updates.

onnx/backend/test/case/node/__init__.py
vinitra: I've just addressed this change as part of the recent updates for arr->list, but some work might need to be done in sequence_value_info as well. It seems like it's only used in the shape-inference-tests: https://github.com/onnx/onnx/blob/8e9c7d57f7c5aa6f6eb7ee7abb0ba2a243781933/onnx/test/shape_inference_test.py#L2522

onnx/backend/test/cmd_tools.py
```diff
@@ -45,18 +45,40 @@ def prepare_dir(path):  # type: (Text) -> None
                 data_set_dir = os.path.join(
                     output_dir, 'test_data_set_{}'.format(i))
                 prepare_dir(data_set_dir)
-                for j, input_np in enumerate(inputs):
-                    tensor = numpy_helper.from_array(
-                        input_np, case.model.graph.input[j].name)
+                for j, input in enumerate(inputs):
+                    if isinstance(input, dict):
```
gramalingam, line 49: A minor suggestion: it may be more general to use "collections.abc.Mapping" instead of "dict". Similarly, "collections.abc.iterable" or "collections.abc.Sequence" might be better than "list" below. However, this would require us to move the test for ndarray up first, since an array might also be an iterable/sequence object. (I haven't tried it out, so don't know if anything will break, so do test it out if you try this. https://stackoverflow.com/questions/35690572/what-does-isinstance-with-a-dictionary-and-abc-mapping-from-collections-doing has more info on this. (I am not fully familiar with pythonic idioms.)

vinitra: After a quick test (shown below) -- I've updated the listings with the abc mappings accordingly. Highlight: lists are abc.Sequence but arrays are not.

>>> import numpy as np
>>> from collections import abc
>>> arr = np.array([1, 2, 3])
>>> lst = [4, 5, 6]
>>> isinstance(arr, abc.Sequence)
False
>>> isinstance(lst, abc.Sequence)
True
>>> isinstance(arr, np.ndarray)
True
>>> isinstance(arr, abc.Iterable)
True
>>> isinstance(lst, abc.Iterable)
True

gramalingam:
@linkerzhang : since this PR involves updates to the proto definition, it would be great to have your feedback on the proposed changes. Thanks.

onnx/backend/test/cmd_tools.py
vinitra: Ultimately decided to not use abc functionality as collections.abc is not included in Python 2.7

onnx/onnx.in.proto
```diff
@@ -626,6 +626,90 @@ message TensorShapeProto {
...
+  // required to store the content in the repeated bytes string_key field.
+  //
+  // When this raw_key field is used to store the key, elements MUST
+  // be stored in as fixed-width, little-endian order.
+  //
+  // Note: the advantage of specific field rather than the raw_key field is
+  // that in some cases (e.g. int data), protobuf does a better packing via
+  // variable length storage, and may lead to smaller binary footprint.
+  // When this field is present, the key_type field MUST NOT be STRING or UNDEFINED
+  optional bytes raw_data = 7;
```
linkerzhang, line 700: can we just use a "Tensor" as a key?

Though this design does not have any function miss, it looks complex...

onnx/onnx.in.proto
```diff
@@ -626,6 +626,90 @@ message TensorShapeProto {
...
+  }
+  // The data type of the element.
+  // This field MUST have a valid SequenceMapElement.DataType value
+  optional int32 elem_type = 1;
+
+  optional TensorProto tensor_value = 2;
+  optional SparseTensorProto sparse_tensor_value = 3;
+  optional SequenceProto sequence_value = 4;
+  optional MapProto map_value = 5;
+}
```
linkerzhang, line 646: I'm seeing this as trying to make a complete set of data values representing any kind of sequences/maps.

Can you give some real use cases please? say, some traditional ml models. I guess, not all the compositions (also recursive) of different values (sequences/maps based on tensors/sequences/maps again) are that useful.

onnx/onnx.in.proto
vinitra: I agree, it is a bit complex. The reason for this spec was to mirror the data types in TensorProto due to the following spec for sequences in TypeProto. raw_data can be removed.
```
    // This field MUST have a valid TensorProto.DataType value
    // This field MUST refer to an integral type ([U]INT{8|16|32|64}) or STRING
    optional int32 key_type = 1;
```

onnx/onnx.in.proto
vinitra: > I'm seeing this as trying to make a complete set of data values representing any kind of sequences/maps.

Yes, this was the intention as detailed in the spec requirements for sequences and maps. I suppose the use case would be multimodal data input, specifically using a data structure analogous to a pythonic list (sequence) to contain tensor data of different shapes and dictionary data (maps) as a single input to a model.

i.e. [ [ np.array([1, 2, 3]), np.array([2, 3, 4, 5]) ], {1:A, 2:B, 3:C} ]  as model input.

I agree that there's not a lot of use cases for it, but I did think it was the best way to represent the written spec. (Unless we determine that sequences / maps cannot recursively contain themselves).

vinitra: > I'm seeing this as trying to make a complete set of data values representing any kind of sequences/maps.

Yes, defining the data type's complete possibilities is the thought process here. I don't see real-life use cases of these recursively-defined data values. There are comparative usefulness levels for multimodal data, i.e. a map with individual maps as it's own values seems less relevant than a sequence containing a sequence. Any more than 2 dimensions of recursively defining themselves don't seem to be useful as input for the current state-of-the-art in NN models.

vinitra: However, the other side of the coin is that some frameworks do allow input of this form to be expressed, even if they're not being currently used. The alternative in onnx.in.proto would be to create a similar format to key representation in MapProto or data representation in TensorProto with a data_type variable and optional input entries (Tensor, SparseTensor, Map, Sequence).

onnx/onnx.in.proto
vinitra: If we use a Tensor, is there any good way to type check that it only contains one element and it is not a float value? If the strict adherence to the TypeProto spec is not necessary to be enforced, then I think Tensor is the right way to go.

BowenBao:
I think `docs/TestCoverage.md` is also updated and needs to be included in this PR.

vinitra:
> I think `docs/TestCoverage.md` is also updated and needs to be included in this PR.

Thanks for the catch! Added now.

onnx/onnx.in.proto
gramalingam: The main concern/question I have is that this is used only to represent test-data and it is not used in a model itself. Adding this to onnx.proto means bumping the IR_VERSION, which is somewhat heavy-weight since it has significant implications (e.g., an ONNX backend might reject the model as being an unsupported version, even though the model itself has no changes). How about we define a different onnxdata.proto for this, used only by test-data generator and testing-utilities?

onnx/onnx.in.proto
gramalingam: Yes, it is complete. But the completeness does not cost us anything (it doesn't add to the complexity in any way, right?). This seems preferable to alternative restricted versions that allow only 1 or 2 levels of nesting, because sooner or later we might find we need to use 3 levels of nesting, etc. As I mention above, the drawback/complexity is only in bumping IR_VERSION, so I think may be defining a different proto file will solve that problem.

BowenBao:
Please also include the following updated files:

```
	onnx/onnx-ml.proto
	onnx/onnx-ml.proto3
	onnx/onnx.proto
	onnx/onnx.proto3
```

There seems to be another test that is failing in the CI, but I'm not sure which test it is by just looking at the log.
```
  File "/home/vsts/.conda/envs/py3.7/bin/backend-test-tools", line 11, in <module>
    load_entry_point('onnx==1.7.0', 'console_scripts', 'backend-test-tools')()
  File "/home/vsts/.conda/envs/py3.7/lib/python3.7/site-packages/onnx-1.7.0-py3.7-linux-x86_64.egg/onnx/backend/test/cmd_tools.py", line 99, in main
    args.func(args)
  File "/home/vsts/.conda/envs/py3.7/lib/python3.7/site-packages/onnx-1.7.0-py3.7-linux-x86_64.egg/onnx/backend/test/cmd_tools.py", line 79, in generate_data
    "and cannot be processed accordingly.", output)
TypeError: ('Your output is not a sequence (list), dictionary (map), or tensor (array) ', 'and cannot be processed accordingly.', 3.0)
```

onnx/onnx.in.proto
vinitra: I agree with this -- I don't think that updating the IR version is strictly necessary for these updates. I will define a separate proto file.

BowenBao:
Please rebase with master again

BowenBao:
You might need to include the following 2 files in the PR as well. They have uncaptured changes that are being reported by the CI. This might be the root cause for the `import not found` errors.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	onnx/onnx-data-ml.proto
	onnx/onnx-data-ml.proto3

onnx/gen_proto.py
```diff
@@ -214,7 +214,7 @@ def main():  # type: () -> None
     parser.add_argument('--protoc_path',
                         default='',
                         help='path to protoc for proto3 file validation')
-    parser.add_argument('stems', nargs='*', default=['onnx', 'onnx-operators'],
+    parser.add_argument('stems', nargs='*', default=['onnx', 'onnx-operators', 'onnx-data'],
```
BowenBao, line 217: i wasn't aware that the onnx-data was something new added in this PR. Why create a separate proto file?

BowenBao:
BTW the import issue is still there. I don't think we should create a separate onnx-data.proto. The `TensorProto` in original onnx.in.proto is also defining data/serialization. These should be defined in separate files.

To solve the import issue, you need to update `CmakeLists.txt` and add

```
relative_protobuf_generate_cpp(gen_onnx_data_proto
                               __tmp_srcs
                               __tmp_hdrs
                               ${ONNX_ROOT}
                               gen_onnx_proto
                               onnx/onnx-data.in.proto)
list(APPEND ONNX_PROTO_SRCS ${__tmp_srcs})
list(APPEND ONNX_PROTO_HDRS ${__tmp_hdrs})
```

then in `onnx/__init__.py` add
```
from .onnx_data_pb import * # noqa
```

at last in `onnx/helper.py` and `onnx/numpy_helper.py`, update import to be
```
from onnx import SequenceProto, MapProto, SequenceMapElement, KeyValuePair
```

Also be sure to run `sh tools/update_doc.sh` locally to run tests and generate updated docs. The above steps fix the import issue, but still encounter some type error that exists previously.

```
===> regenerate test data from node test
  File "onnx/backend/test/cmd_tools.py", line 103, in <module>
    main()
  File "onnx/backend/test/cmd_tools.py", line 99, in main
    args.func(args)
  File "onnx/backend/test/cmd_tools.py", line 79, in generate_data
    "and cannot be processed accordingly.", output)
TypeError: ('Your output is not a sequence (list), dictionary (map), or tensor (array) ', 'and cannot be processed accordingly.', 3.0)
```

onnx/gen_proto.py
gramalingam: This extension does not affect the IR representation used for ONNX models ... it impacts only the representation of test-data. The main goal is to avoid bumping the IR_VERSION unnecessarily. Keeping the proto files separate is helpful as it clarifies that the data-serialization does not impact the model IR or IR_VERSION. Do you see reasons for keeping them together in one proto (other than the challenges discussed below for importing etc.)?

BowenBao: My concern is that the `TensorProto` which is also data-serialization is kept in the original proto. We also see references of it in `initializers` and `attributes`. So the data serialization is impacting the model IR right now (although for sequence and map it is currently not the case, but then we must declare that these types will not supported as initializers or attributes).

vinitra: Discussed offline, consensus is to keep current architecture.

onnx/backend/test/case/node/sequenceinsert.py
```diff
@@ -0,0 +1,56 @@
...
+from __future__ import unicode_literals
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+def sequence_insert_reference_implementation(sequence, tensor, position=None):  # type: (np.ndarray, np.ndarray, np.ndarray) -> np.ndarray
```
BowenBao, line 13: need to be careful with python objects: the list is passed by reference, so `sequence` gets updated as well. Can use `copy` to create a new sequence to be udpated.

onnx/backend/test/case/node/sequenceinsert.py
```diff
@@ -0,0 +1,56 @@
...
+            tensor = test_inputs[0]
+
+            if len(test_inputs) > 1:
+                node = onnx.helper.make_node(
+                    'SequenceInsert',
+                    inputs=['sequence', 'tensor', 'position'],
+                    outputs=['output_sequence']
+                )
+                position = test_inputs[1]
+                inserted = sequence_insert_reference_implementation(sequence, tensor, position)
```
BowenBao, line 45: see above comment, i think that's probably the reason why the generated test data file is not expected. The `sequence` at this point is already updated by the other branch.

onnx/backend/test/case/node/sequenceinsert.py
```diff
@@ -0,0 +1,56 @@
...
+def sequence_insert_reference_implementation(sequence, tensor, position=None):  # type: (np.ndarray, np.ndarray, np.ndarray) -> np.ndarray
+    if position:
+        # In these cases, insert_position will be between [-len(sequence), len(sequence)]
+        # The position argument will be in the format np.array([pos_index])
+        insert_position = position[0]
+        sequence.insert(insert_position, tensor)
+    else:
+        # Default position of insertion is at the end of the sequence.
+        sequence.append(tensor)
+    return np.array(sequence)
```
BowenBao, line 22: this sets output to `ndarray`, which is later recognized as tensor. Should be list for sequence.

onnx/backend/test/case/node/sequenceinsert.py
vinitra: - since list doesn't have a .copy() function and I didn't want to introduce the copy library, I'm using the list() casting functionality to create a new copy.

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -199,17 +199,19 @@ ResolvedTestCase LoadSingleTestCase(const UnsolvedTestCase& t) {
   st.test_case_name_ = t.test_case_name_;
   ONNX_NAMESPACE::ParseProtoFromBytes(
       &st.model_, raw_model.c_str(), raw_model.size());
-
+  ONNX_NAMESPACE::ValueInfoProto input = &st.model_.graph.input;
```
gramalingam, line 202: If I remember right, "graph.input" is a sequence of inputs. We will have to use the i-th element of this sequence when we process the input-file corresponding to the i-th input down below.

.travis.yml
```diff
@@ -1,83 +0,0 @@
-language: generic
```
BowenBao, line 1: I might be missing something.. why are we removing this file?

vinitra: my mistake! that was a missed file merge.

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -199,25 +198,62 @@ ResolvedTestCase LoadSingleTestCase(const UnsolvedTestCase& t) {
...
     for (auto& input_file : test_data.input_filenames_) {
       std::string input_data;
+      ONNX_NAMESPACE::ValueInfoProto input_info;
       LoadSingleFile(input_file, input_data);
-      ONNX_NAMESPACE::TensorProto input_proto;
-      ONNX_NAMESPACE::ParseProtoFromBytes(
-          &input_proto, input_data.c_str(), input_data.size());
-      proto_test_data.inputs_.emplace_back(std::move(input_proto));
+      input_info = st.model_.graph().input(test_data_counter);
+      if(input_info.doc_string().find(tensor_str) != std::string::npos) {
```
BowenBao, line 212: curious why are we able to determine input type through `doc_string`? Is there some part in code that we are setting 'Sequence/Tensor/Map' in `doc_string`?

onnx/backend/test/cpp/driver_test.cc
```diff
@@ -211,10 +211,11 @@ class ONNXCppDriverTest : public ::testing::TestWithParam<
       onnxBackendID backendID) {
     // Check Model
     ONNX_NAMESPACE::checker::check_model(model_);
-    // Check Input&Output Tensors
+    // Check Input & Output Data
     ONNX_NAMESPACE::checker::CheckerContext ctx;
     for (auto proto_test_data : protos_) {
       for (auto input : proto_test_data.inputs_) {
+        // TODO: address case of sequence / map input
```
BowenBao: Complete the checks for sequence / map :)

onnx/backend/test/cpp/driver/test_driver.cc
vinitra: It prints out with the type in it, so I was using that as a metric, but I agree -- it's an imperfect solution since it's not actually being tested. ValueInfo has a type argument which has a TypeProto value. Let me see if comparing the types works better.

vinitra: Updated by comparing classes of TypeProto. Not sure if this is much better, but definitely seems less arbitrary than comparing doc strings.

onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -199,25 +198,62 @@ ResolvedTestCase LoadSingleTestCase(const UnsolvedTestCase& t) {
...
-      std::string output_data = "";
+      std::string output_data;
+      ONNX_NAMESPACE::ValueInfoProto output_info;
+      output_info = st.model_.graph().output(test_data_counter);
       LoadSingleFile(output_file, output_data);
-      ONNX_NAMESPACE::TensorProto output_proto;
-      ONNX_NAMESPACE::ParseProtoFromBytes(
-          &output_proto, output_data.c_str(), output_data.size());
-      proto_test_data.outputs_.emplace_back(std::move(output_proto));
+      if(typeid(output_info.type()) == typeid(ONNX_NAMESPACE::TypeProto::Tensor)) {
```
gramalingam, line 238: I don't think we should use "typeid" here. I doubt if it would work. Protobuf has its own encoding for "oneof" : see https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#oneof-embedded-message-fields ... I think we should probably be using output_info.type().has_tensor_type() ... similarly, has_sequence_type() and has_map_type() for below.

gramalingam: It would be good to validate if the right branch is executing here. I suspect none of these branches may be executing, but maybe I am wrong.

onnx/checker.cc
```diff
@@ -228,6 +228,14 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
 #undef check_field
 }
 
+void check_sequence(const SequenceProto& sequence) {
+  enforce_has_repeated_field(sequence, values);
```
gramalingam, line 232: (a) enforce_has_repeated_fields requires that there be at least 1 repeated value. Do we want that condition for sequence/map here? That wouldn't allow an empty sequence to be an input, for example. May be better to remove that check?
(b) However, I think there are other sanity checks that could be done on the data: one check is to iterate through all elements of the sequence and calling the corresponding check function on each element; a second check is to verify that all elements are of the same type.

onnx/backend/test/cpp/driver/test_driver.cc
vinitra: Updated to use .has_tensor_type() and related functions!

vinitra: After testing with the new methods, it is hitting those if branches. :)

onnx/checker.cc
vinitra: Added the iterative checks, as well as helper functions for key value pair and sequence map element testing. Removed the has_repeated_field checks.

onnx/checker.cc
```diff
@@ -228,12 +228,60 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+}
+
+// Check that the underlying values for SequenceProto and MapProto are
+// valid by calling the respective check functions on the SequenceMapElements.
+void check_sequence_map_element(
+  const SequenceMapElement& seq_map_elem,
+  const CheckerContext& ctx,
+  const std::string& seq_map_name) {
+    if (seq_map_elem.elem_type() == SequenceMapElement::TENSOR) {
+      check_tensor(seq_map_elem.tensor_value(), ctx);
```
gramalingam, line 267: Add "enforce_has_field(seq_map_elem, tensor_value)" before check_tensor. Similarly for the other cases below. Similarly, for "elem_type" above.

gramalingam:
@linkerzhang : do you have any further comments on this PR? Thanks!

onnx/checker.cc
vinitra: added.

onnx/onnx-data.in.proto
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) ONNX Project Contributors.
+// Licensed under the MIT license.
```
linkerzhang, line 2: what's the difference between onnx-data.proto and onnx-data-ml.proto?

onnx/onnx-data.in.proto
```diff
@@ -0,0 +1,98 @@
...
+  optional MapProto map_value = 5;
+}
+
+// Sequences
+//
+// Defines a dense, ordered, collection of elements that are of homogeneous types.
+message SequenceProto {
+
+  optional string name = 1;
+  repeated SequenceMapElement values = 2;
```
linkerzhang, line 42: In this way, there may be a tensor and a map in one sequence, which I don't think is correct.

onnx/onnx-data.in.proto
```diff
@@ -0,0 +1,98 @@
...
+  optional SequenceMapElement value = 8;
+}
+
+// Maps
+//
+// Specifies an associative table, defined by a key type and a value type.
+message MapProto {
+  optional string name = 1;
+
+  repeated KeyValuePair pairs = 2;
```
linkerzhang, line 97: A better way might be define two fields. one for keys, one for values, which should be a Sequence.

onnx/onnx-data.in.proto
gramalingam: What if we do something similar to AttributeProto: define the DataType here, and use "repeated TensorProto", "repeated SparseTensorProto", etc.? Would that be better?

onnx/onnx-data.in.proto
gramalingam: One option is to "inline" the fields in KeyValuePair here ("key_type" and all the "*_data") and make all the _data fields repeated. The value can be a SequenceProto. With the implicit constraint that both sequences have the same length. Would that work?

gramalingam: Or, we can define a "KeySequenceProto" as above.

onnx/onnx-data.in.proto
vinitra: No difference. It was automatically generated when ONNX_ML=1 using gen_proto.py.

onnx/onnx-data.in.proto
vinitra: I think Rama's suggested structure makes a lot of sense. In an offline discussion, Rama explained the strongly typed nature of these data structures in ONNX and how they operate differently than in Python (where you can have {'key':'value', 1: 32} as a valid dictionary), which was the reason behind this original design.

vinitra: As per my earlier interpretation, having a tensor and a map in one sequence was a valid option. I understand now that it should not be allowed. :)

onnx/onnx-data.in.proto
vinitra: Option 1: remove SequenceMapElement and KeyValuePair + their related helper functions. Implement the new proto structure in MapProto itself with a number of repeated key_data fields, a key_type, a value_type, and a repeated value (Sequence) field.
Option 2: Create a KeySequenceProto (instead of KeyValuePair) to not clutter the MapProto and allow us to index the dictionary pairs as separate elements, with a single element SequenceProto as the value and the same structure for the keys.

Not sure which option is better -- it might be nice to index the corresponding key and value in KeySequenceProtos if the structure exists as repeated elements instead of indexing the separate sequences in parallel, so I'm leaning towards option 2. Do you have an opinion before I implement it, @linkerzhang?

onnx/onnx-data.in.proto
linkerzhang: that's bad. if no difference needed, then no need having two files. "// #if ONNX-ML" should not in the file.

onnx/onnx-data.in.proto
linkerzhang: the map should be just one sequence of keys and one sequence of values. no?

onnx/onnx-data.in.proto
```diff
@@ -0,0 +1,76 @@
...
+  optional string name = 1;
+
+  // All MapProto data types must have the same length of keys and values.
+
+  // The data type of the key.
+  // This field MUST have a valid TensorProto.DataType value of
+  // INT8, INT16, INT32, INT64, UINT8, UINT16, UINT32, UINT64, or STRING
+  optional int32 key_type = 2;
+
+  // Every element of keys has to be a tensor of one of the following data types
```
gramalingam, line 66: Every key has to be a scalar value. Why should we make it a tensor here? Just raising the question to understand the pros and cons. Traditionally, ONNX used a tensor of zero dimensions to represent scalars. But that is in a different context. In this context, is it useful represent the key as a tensor of zero dimensions? Or, would it be better to just use a "repeated int64" (to cover all the int cases) and a "repeated bytes" to cover the string case?

vinitra: I do think it makes more sense in terms of space to implement keys as a repeated sequence of ints / bytes. Will implement accordingly.

onnx/checker.cc
```diff
@@ -228,6 +228,92 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
 #undef check_field
 }
 
+void check_sequence(const SequenceProto& sequence, const CheckerContext& ctx) {
+  enforce_has_field(sequence, elem_type);
+  if (sequence.elem_type() == SequenceProto::TENSOR) {
+    enforce_has_field(sequence, tensor_values);
```
gramalingam, line 234: I think we can drop this check above, since tensor_values is a repeated field and since we (presumably) want to allow for the case with zero elements in the sequence.

onnx/checker.cc
```diff
@@ -228,6 +228,92 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+    for (const TensorProto& tensor : sequence.tensor_values()) {
+      check_tensor(tensor, ctx);
+    }
+  }
+  else if (sequence.elem_type() == SequenceProto::SPARSE_TENSOR) {
+    enforce_has_field(sequence, sparse_tensor_values);
```
gramalingam, line 240: Same here and below for other cases: drop check in repeated case where zero-elements is valid.

vinitra: Thanks for the recommendation. Removed the relevant checks.

vinitra:
> Please also show the model example/usage of sequence data serialization in the PR.

> btw, given this change is indeed affecting the IR, please move the operator and its test data into a separate PR.

Since we are not updating the IR version and the purpose was to enable using these data types in testing infrastructure, I think it makes sense to have an example of a Sequence operator in this PR. Referencing the earlier comments above to indicate why it was included.

CMakeLists.txt
```diff
@@ -163,7 +163,7 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP NAME SRCS HDRS ROOT_DIR DEPEND)
     set(ABS_FILE ${ROOT_DIR}/${INFILE})
     get_filename_component(FILE_DIR ${ABS_FILE} DIRECTORY)
     get_filename_component(FILE_WE ${INFILE} NAME_WE)
-    if(ONNX_ML)
+    if(ONNX_ML AND NOT(FILE_WE STREQUAL "onnx-data"))
```
askhade, line 166: Add comment here to explain why you are adding this exception for onnx-data

onnx/numpy_helper.py
```diff
@@ -118,6 +118,131 @@ def from_array(arr, name=None):  # type: (np.ndarray[Any], Optional[Text]) -> Te
...
+    Inputs:
+        lst: a Python list
+        name: (optional) the name of the sequence.
+    Returns:
+        sequence: the converted sequence def.
+    """
+    sequence = SequenceProto()
+    if name:
+        sequence.name = name
+    elem_type = type(lst[0])
```
BowenBao, line 157: Can we extend this to allow creating from empty list? Maybe also add optional dtype parameter such that type can be specified if the list is empty.

onnx/checker.cc
```diff
@@ -228,6 +228,106 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+    }
+
+  int num_keys = 0;
+  if (!map.has_keys()) {
+    num_keys = map.keys_size();
+  } else if (!map.has_string_keys()) {
+    num_keys = map.string_keys_size();
+  }
+
+  // if map.key_type == tensor -> tensor_values_size()
```
BowenBao, line 320: nit: remove comments

onnx/checker.cc
```diff
@@ -228,6 +228,106 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+  // check_sequence(map.values(), ctx);
+
+  // if (num_keys != map.values_size()){
```
BowenBao, line 323: should this be uncommented?

onnx/checker.cc
```diff
@@ -228,6 +228,106 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+      ++used_fields;                        \
+      check_type(type);                     \
+    }
+
+    check_repeated_field(keys, TensorProto::INT64);
+    check_repeated_field(string_keys, TensorProto::STRING);
+
+  #undef check_repeated_field
+
+    // Normally, used_fields is expected to be 1.
```
BowenBao, line 305: nit: indent

onnx/checker.cc
vinitra: resolved with new fixes in latest update!

CMakeLists.txt
vinitra: Addressed

onnx/numpy_helper.py
vinitra: Addressed!

onnx/backend/test/case/node/__init__.py
```diff
@@ -105,7 +105,7 @@ def _extract_value_info(input, name, ele_type=None):  # type: (Union[List[Any],
     if isinstance(input, list):
         return onnx.helper.make_sequence_value_info(
             name=name,
-            elem_type=onnx.mapping.NP_TYPE_TO_TENSOR_TYPE[input[0].dtype],
+            elem_type=onnx.mapping.PYTHON_TYPE_TO_ELEMENT_TYPE[type(input[0])],
```
gramalingam, line 108: The existing "make_sequence_value_info" seems to have a limitation, as it assumes that we only allow sequences of tensors and the elem-type parameter seems to be that of the type of elements in the tensor. So, I think this is not going to be consistent with that. We need to either (a) Restrict ourselves to sequences of tensors, in which case we should pick up the element-type of the numpy array here, or (b) Have a more comprehensive type-constructor (which would have to be recursive to handle sequences/maps contained inside sequences).

onnx/mapping.py
```diff
@@ -59,3 +59,16 @@
...
+
+STORAGE_ELEMENT_TYPE_TO_FIELD = {
+    int(SequenceProto.TENSOR): 'tensor_values',
+    int(SequenceProto.SPARSE_TENSOR): 'sparse_tensor_values',
+    int(SequenceProto.SEQUENCE): 'sequence_values',
+    int(SequenceProto.MAP): 'map_values'
+}
+
+PYTHON_TYPE_TO_ELEMENT_TYPE = {
+    np.ndarray: int(SequenceProto.TENSOR),
```
askhade, line 71: this is not a python type

onnx/numpy_helper.py
```diff
@@ -118,6 +118,139 @@ def from_array(arr, name=None):  # type: (np.ndarray[Any], Optional[Text]) -> Te
...
+        sequence: the converted sequence def.
+    """
+    sequence = SequenceProto()
+    if name:
+        sequence.name = name
+
+    if len(lst) == 0:
+        elem_type = dtype
+    else:
+        elem_type = type(lst[0])
```
askhade, line 163: this does not seem right. The elem_type here will be the type of the input 1 in the list... it is not compatible with sequence.elem_type .... this code will not work...
The type should be one of the following:
enum DataType {
    UNDEFINED = 0;
    TENSOR = 1;
    SPARSE_TENSOR = 2;
    SEQUENCE = 3;
    MAP = 4;
  }

onnx/mapping.py
vinitra: changed the name accordingly

onnx/backend/test/case/node/__init__.py
vinitra: Right now, the implementation is option (a). A future PR will address (b).

vinitra:
@linkerzhang, as Azure Pipelines CIs are passing, this PR looks ready to go. The requested changes to onnx-data.proto have been implemented. Can you take another look?

Minor caveats:
1. CircleCI build is currently broken for all PRs and is being fixed here: https://github.com/onnx/onnx/pull/2941
2. TravisCI is failing because it's finding a diff in onnx/onnx-data.proto from the generated file because the onnx/onnx.proto import statement is changing across both options. It's important to have the differing import statements (import onnx/onnx-ml.proto vs onnx/onnx.proto because only one is generated based on the ONNX_ML value). @askhade and I are brainstorming a way around this for the CI check, but the PR is entirely functional.

onnx/onnx-data.in.proto
```diff
@@ -0,0 +1,85 @@
...
+  repeated int64 keys = 3;
+
+  // If keys are strings, they are represented by the repeated bytes field
+  // string_keys below.
+  repeated bytes string_keys = 4;
+
+  // The data type of the value.
+  // This field MUST have a valid SequenceProto.DataType value of
+  // TENSOR, SPARSE_TENSOR, MAP, SEQUENCE
+  optional int32 value_type = 5;
```
linkerzhang, line 82: this is not needed, right? The type has been specified in SequenceProto already.

onnx/onnx-data_pb.h
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) ONNX Project Contributors.
```
linkerzhang, line 1: is this file needed?

onnx/onnx-data.in.proto
```diff
@@ -0,0 +1,85 @@
...
+// #endif
+
+// This file contains the proto definitions for MapProto and
+// SequenceProto. These protos are used to represent the data structures
+// of maps and sequence for use in test data or ModelProto.
+
+// Sequences
+//
+// Defines a dense, ordered, collection of elements that are of homogeneous types.
+// Sequences can be made out of tensors, maps, or sequences.
```
linkerzhang, line 20: so if it's "TENSOR", the tensor's element type should also be the same, but can be different shapes, right? if yes, I'd suggest to add this clarification in the comments too.

gramalingam: The clarification can point to https://github.com/onnx/onnx/blob/master/docs/IR.md#static-tensor-shapes for a more complete description. Whether the tensors can have different shapes or not depends on the type/shape associated with the corresponding "ValueInfo".

gramalingam: For example, "Sequence<Tensor<float, [M,N]>" means that all tensors have same shape. But "Sequence<Tensor<float, [*,*]>" means they can have different shapes (but all of rank 2), where the "*" means the corresponding dimension has no symbolic/constant value.

gramalingam: Finally, "Sequence<Tensor<float, omitted>>" means that the different tensors can have different ranks, when the "shape" itself is omitted from the tensor-type.

onnx/onnx-data_pb.h
vinitra: Removing this file is not trivial -- perhaps something to investigate in a follow-on PR. onnx-operators.proto has a similar onnx-operators_pb.h as well.

onnx/onnx-data.in.proto
vinitra: Removed! Thanks for the feedback.

onnx/onnx-data.in.proto
vinitra: Addressed and included these points in the recent update.

vinitra:
:shipit:

PR #2583, opened 2020-02-03
add utility function to make reference attribute 

linkerzhang:
When creating function op, make reference attribute whose name is not the same as the attribute it refers.


onnx/defs/attr_proto_util.h
```diff
@@ -8,16 +8,49 @@
...
+    const std::string& attr_name,
+    const std::vector<TensorProto>& values);
+AttributeProto MakeAttribute(
+    const std::string& attr_name,
+    const std::vector<GraphProto>& values);
 
-AttributeProto MakeRefAttribute(const std::string& attr_name, AttributeProto_AttributeType type);
+// Make a "reference" attribute for a node in a function body.
+// <attr_name> specifies the attribute name of both the function node and its
+// function body node. They're using same attribute name.
```
wschin:
```suggestion
// function body node. They're using the same attribute name.
```

onnx/defs/attr_proto_util.h
```diff
@@ -8,16 +8,49 @@
...
+// <type> specifies the attribute type.
+AttributeProto MakeRefAttribute(
+    const std::string& attr_name,
+    AttributeProto_AttributeType type);
+
+// Make a "reference" attribute for a node in a function body.
+// <attr_name> specifies the attribute name of the function body node.
+// <referred_attr_name> specifies the referred attribute name of the function
+// node. <type> specifies the attribute type.
```
wschin:
```suggestion
// node.
// <type> specifies the attribute type.
```

onnx/defs/attr_proto_util.cc
```diff
@@ -51,9 +51,16 @@ ADD_LIST_ATTR_IMPL(GraphProto, AttributeProto_AttributeType_GRAPHS, graphs)
 AttributeProto MakeRefAttribute(
     const std::string& attr_name,
     AttributeProto_AttributeType type) {
+  return MakeRefAttribute(attr_name, attr_name, type);
+}
+
+AttributeProto MakeRefAttribute(
```
wschin: Not sure how but I feel it's better to have at least one test.

linkerzhang:
this is already covered by the PR #2575

PR #2584, opened 2020-02-05
allow attribute type to be float, floats, int, ints, string, strings

linkerzhang:
allow attribute type to be float, floats, int, ints, string, strings when expected attribute type is tensor. This will ask a IR version bump, I think.

### DIFF
--- a/onnx/defs/schema.cc
+++ b/onnx/defs/schema.cc
@@ -97,58 +97,83 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
 void OpSchema::CheckInputOutputType(struct InferenceContext& ctx) const {
   std::unordered_map<std::string, std::string> type_constraints;
   // check all input types
-  for (size_t in_idx = 0; in_idx < ctx.getNumInputs() && in_idx < inputs_.size(); ++in_idx) {
-    const auto& param      = inputs_[in_idx];
-    const auto& type_str   = param.GetTypeStr();
+  for (size_t in_idx = 0;
+       in_idx < ctx.getNumInputs() && in_idx < inputs_.size();
+       ++in_idx) {
+    const auto& param = inputs_[in_idx];
+    const auto& type_str = param.GetTypeStr();
     const auto& param_type = ctx.getInputType(in_idx);
-    const auto& all_types  = param.GetTypes();
-    if (nullptr == param_type || param_type->value_case() == TypeProto::VALUE_NOT_SET) {
+    const auto& all_types = param.GetTypes();
+    if (nullptr == param_type ||
+        param_type->value_case() == TypeProto::VALUE_NOT_SET) {
       continue;
-    } else if (!all_types.empty() && all_types.find(Utils::DataTypeUtils::ToType(*param_type)) == all_types.end()) {
-      fail_check(param.GetName(), " typestr: ", type_str, ", has unsupported type: ", *Utils::DataTypeUtils::ToType(*param_type));
+    } else if (
+        !all_types.empty() &&
+        all_types.find(Utils::DataTypeUtils::ToType(*param_type)) ==
+            all_types.end()) {
+      fail_check(
+          param.GetName(),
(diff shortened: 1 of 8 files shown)


onnx/defs/schema.cc
```diff
@@ -288,10 +313,27 @@ void OpSchema::Verify(const NodeProto& node) const {
           "Unrecognized attribute: ", name, " for operator ", node.op_type());
     }
 
-   // Type would be UNDEFINED if not set
+    // Type would be UNDEFINED if not set
     if (attr_proto.type() != expected_type) {
-      fail_check(
-          "Mismatched attribute type in '", node.name() + " : " + name, "'");
+      // Expected attribute type is "Tensor", in this case, "FLOAT", "FLOATS",
+      // "INT", "INTS", "STRING", "STRINGS" should be also OK.
```
linkerzhang: this is the only change, all others are format changes.

wschin:
Any shape inference tests for attributes?

TMVector:
Should this PR include changes to the spec?

wschin:
Not sure if we want to bump op version.

TMVector:
@gramalingam [commented](https://github.com/onnx/onnx/pull/2575#issuecomment-582581695) about compatibility. I can imagine this could definitely break some things.

I think we are trying to address a couple of issues.

### Allow an attribute to hold one of many types
I think this should include sequences and maps, which I expect will be added to `AttributeProto` in #2581).

We could address this by:
* Adding a new AttributeType `ANY_VALUE` which means any attribute type apart from `GRAPH` or `GRAPHS`. Any type restrictions (e.g. only `int` or `float`) would have to be in documentation only. This type tag would only be used in operator specifications -- node attributes would use the existing type tags.
* Adding lots of attributes (`value_int`, `value_ints`, ...) and specifying in the documentation only one should be used. This would be done ad-hoc -- it has already been done to some extent for some operators, and even `Constant` has `value` and `sparse_value`.

### Lift non-tensor attribute values in to the value level
This is the situation raised in #2575, and this could be addressed by:
* Changing `Constant` to have an attribute which can have any type, as above. Any values which are not valid graph values (`int`, `ints`, `float`, `floats`, `string`, `strings`) would be promoted to 0D or 1D tensors.

linkerzhang:
I think adding an attribute type "any" is tricky (sorry for this wording), it actually is adding a very customized type for this "Constant" requirement right now.

I got the comment from @gramalingam in PR #2575 - adding multiple attributes in "Constant", say, "value_int", "value_ints", "value_float", "value_floats", "value_string", "value_strings" to make "Constant" op be able to accept all these types' attributes. it's not that bad, as it only affects ONE operator (even though it's not that beautiful).

However, I'd still add more on why I'm still suggesting to make this change. AttributeType: INT, INTS, FLOAT, FLOATS, STRING, STRINGS were designed (instead of reusing type TENSOR), is because, they simplify things when the attribute is just a scalar or an array, though they can be represented as a Tensor. This means, these 6 types are indeed "TENSOR" representable.

About version bump: This PR will ask for a IR version bump.
1. there's NO backward compatibility issue - any old models (without this PR) can still be runnable.
2. there's forward compatibility issue - a new model (with using this statement) can't be loaded and run against a runtime if it does not support this PR, so that IR version should be bumped to differentiate the new model from old models.

wschin:
> 1. there's NO backward compatibility issue - any old models (without this PR) can still be runnable.
> 2. there's forward compatibility issue - a new model (with using this statement) can't be loaded and run against a runtime if it does not support this PR, so that IR version should be bumped to differentiate the new model from old models.

Do we prefer to bump operator version? Increasing IR version for adding some attributes to an operator seems too big.

onnx/defs/schema.cc
wschin: I thought we would only touch `Constant`. Why do we add these lines?

linkerzhang:
abandon this PR due to agreement on adding more support on Constant op

Issue #2548, opened 2020-01-14
onnx.onnx_cpp2py_export.checker.ValidationError: Node () has input size 1 not in range [min=2, max=3].


zhonhel:
Hi all~

I'm converting a pytorch model to an onnx model.
The convert process doesn't report any error or warning.
But when I check the onnx model, using this function:
```
onnx.checker.check_model(model)
```
I get the following error:
```
onnx.onnx_cpp2py_export.checker.ValidationError: Node () has input size 1 not in range [min=2, max=3].

==> Context: Bad node spec: input: "626" output: "634" op_type: "Pad" attribute { name: "mode" s: "constant" type: STRING } attribute { name: "pads" ints: 0 ints: 0 ints: 1 ints: 1 ints: 0 ints: 0 ints: 1 ints: 1 type: INTS } attribute { name: "value" f: 0 type: FLOAT }
```
Can anyone help me?  Thank you very much!

linkerzhang:
adding @spandantiwari for thoughts.

Looks like the pytorch model's pad is converted into Pad-2, however, the model's opset information is using op set version 11 (Pad-11 needs 2-3 inputs)

zhonhel:
@linkerzhang 
Thank you very much！
I indeed use parameter
```opset_version=11```
in 
```torch.onnx.export```
function.
Because if I use:
```opset_version=9(default)```
it will occur the following error:
```RuntimeError: ONNX export failed: Couldn't export operator aten::upsample_bilinear2d```

**Comment by zhonhel** [comment]
@linkerzhang 
I find the error:
```
RuntimeError: ONNX export failed: Couldn't export operator aten::upsample_bilinear2d
```
is because I use the follow op in pytorch:
```
F.interpolate(x, size=self.out_size, mode="bilinear", align_corners=True)
```
when i change this op to:
```
F.interpolate(x, size=self.out_size, mode="bilinear", align_corners=False)
```
This error is solved.
But when I use mxnet load the generated onnx model using this function: `mxnet.contrib.onnx.import_model`, I get the following error:
```
  File "(/path/to)/miniconda3/lib/python3.7/site-packages/mxnet/contrib/onnx/onnx2mx/import_onnx.py", line 118, in <listcomp>
    inputs = [self._nodes[i] for i in node.input]
KeyError: 'entry.conv1.0.weight'
```

linkerzhang:
have no idea about mxnet, try onnx runtime? :)

mrlzla:
[...]

PR #2590, opened 2020-02-06
fix the attribute types section in IR.md

linkerzhang:


docs/IR.md
```diff
@@ -366,7 +366,7 @@ _Historical Notes_: The following extensions were considered early on, but were
 
 ### Attribute Types
 
-The type system used for attributes is a superset of that used for of inputs and outputs. In addition to tensors, attribute values may be scalar numerical values, strings, and graphs. Sequences are available for attributes in both ONNX and ONNX-ML. Maps are not available for attributes in either variant.
```
TMVector: I took this to refer to `AttributeProto.tensors`, but I suppose `tensors` allows heterogeneous sequences, where sequence values are homogenous in value type, and will probably be added in #2548.

Maybe there should be something said about allowing multiple tensors and graphs in one attribute?

gramalingam: "a superset of" seems incorrect too. How about replacing it by "related to but slightly different from"?

gramalingam: I also took "sequences" to refer to the "repeated" version of all attributes. I also note that there is a PR in progress (https://github.com/onnx/onnx/pull/2581 ) to define the message format for Sequences and Maps. It does not yet seem to add this to AttributeProto, but I assume that is the intention or end-goal.

gramalingam: I am also not clear what "either variant" refers to ... sounds like a leftover error. Why not simplify this to:
```
The type system used for attributes is related to but slightly different from that used for of inputs and outputs.
Attribute-values may be a dense tensor, or sparse tensor, or a scalar numerical value, or a string, or a graph, or
repeated values of one of the abovementioned types.
```

gramalingam: Currently, some ops encode a map as a pair of attributes (repeated keys, repeated values).

vinitra: I am in favor of the wording @gramalingam mentioned above. Definitely adds clarity.

PR #2592, opened 2020-02-07
Allow Constant operator to promote scalar and list to tensors.

jeremycochoy:
This PR follow comments https://github.com/onnx/onnx/pull/2575#issuecomment-582698230 and https://github.com/onnx/onnx/pull/2575#issuecomment-582581695 in order to allow Function operators like https://github.com/onnx/onnx/pull/2575 to promote scalar values to tensors.

## Content of the PR
This PR define new attributes `value_int`, `value_ints`, `value_float`,  `value_floats`,  `value_string`, `value_strings` that take *scalar types* for the Constant operator.

## Promoting scalar
Right now scalar and tensors have two separated type systems. This do not create any trouble with operator that are implemented in backend as the backend is responsible of mixing scalars and tensors in the way he want.
But for operators defined as functions (i.e. graph) such mixing should happen directly in the defined graph. Unfortunately, we miss an operator that can do this "mixing" of types. One way to solve this issue is simply to allow the `Constant` operator to lift the *scalars* to *tensors*.

---

I am not use to this code base so let me know if something is missing :)

### DIFF
--- a/onnx/defs/generator/defs.cc
+++ b/onnx/defs/generator/defs.cc
@@ -76,7 +76,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           auto* value_string = ctx.getAttribute("value_string");
           auto* value_strings = ctx.getAttribute("value_strings");
 
-          std::vector<int> non_null_attr = {
+          std::vector<bool> non_null_attr = {
             (nullptr != value),
             (nullptr != sparse_value),
             (nullptr != value_int),
--- a/onnx/defs/generator/defs.cc
+++ b/onnx/defs/generator/defs.cc
@@ -29,32 +29,32 @@ ONNX_OPERATOR_SET_SCHEMA(
             false)
         .Attr(
             "value_int",
-            "The value for the elements of the output tensor.",
+            "The value for the sole element for the scalar, int64, output tensor.",
             AttributeProto::INT,
             false)
         .Attr(
             "value_ints",
-            "The value for the elements of the output tensor.",
+            "The values for the elements for the 1D, int64, output tensor.",
             AttributeProto::INTS,
             false)
         .Attr(
             "value_float",
-            "The value for the elements of the output tensor.",
+            "The value for the sole element for the scalar, float32, output tensor.",
             AttributeProto::FLOAT,
             false)
         .Attr(
             "value_floats",
-            "The value for the elements of the output tensor.",
+            "The values for the elements for the 1D, float32, output tensor.",
             AttributeProto::FLOATS,
             false)
         .Attr(
             "value_string",
-            "The value for the elements of the output tensor.",
+            "The value for the sole element for the scalar, UTF-8 string, output tensor.",
             AttributeProto::STRING,
             false)
         .Attr(
             "value_strings",
-            "The value for the elements of the output tensor.",
+            "The values for the elements for the 1D, UTF-8 string, output tensor.",
             AttributeProto::STRINGS,
             false)
         .Output(
@@ -103,7 +103,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             if (!value_int->has_i())
               fail_shape_inference("Attribute 'value_int' expect an integer.")
             updateOutputElemType(ctx, 0, TensorProto::INT64);
-            appendDim(getOutputShape(ctx, 0), 1);
+            updateOutputShape(ctx, 0, TensorShapeProto());
             return;
           }
 
@@ -121,7 +121,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             if (!value_float->has_f())
               fail_shape_inference("Attribute 'value_float' expect a float.")
             updateOutputElemType(ctx, 0, TensorProto::FLOAT);
-            appendDim(getOutputShape(ctx, 0), 1);
+            updateOutputShape(ctx, 0, TensorShapeProto());
             return;
           }
 
@@ -139,7 +139,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             if (!value_string->has_s())
               fail_shape_inference("Attribute 'value_string' expect a string.")
             updateOutputElemType(ctx, 0, TensorProto::STRING);
-            appendDim(getOutputShape(ctx, 0), 1);
+            updateOutputShape(ctx, 0, TensorShapeProto());
             return;
           }
 
@@ -164,8 +164,10 @@ ONNX_OPERATOR_SET_SCHEMA(
               appendDim(output_shape, sparse.dims(i));
             return;
           }
+
           fail_shape_inference(
-              "One of the attributes 'value' or 'sparse_value' must be specified for a Constant node.")
+              "TypeAndShapeInferenceFunction implementation incomplete: "
+              "this line should never be reached.")
         }));
 
 static const char* ConstantOfShape_ver9_doc = R"DOC(
--- a/onnx/defs/generator/defs.cc
+++ b/onnx/defs/generator/defs.cc
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 
 #include <cmath>
-#include <numeric>
 #include <algorithm>
 #include "onnx/defs/function.h"
 #include "onnx/defs/schema.h"
--- a/onnx/defs/generator/defs.cc
+++ b/onnx/defs/generator/defs.cc
@@ -14,7 +14,7 @@ or value_* must be specified.
 
 ONNX_OPERATOR_SET_SCHEMA(
     Constant,
-    11,
+    12,
     OpSchema()
         .SetDoc(Constant_ver11_doc)
         .Attr(
--- a/onnx/defs/generator/defs.cc
+++ b/onnx/defs/generator/defs.cc
@@ -8,7 +8,7 @@
 
 namespace ONNX_NAMESPACE {
 static const char* Constant_ver11_doc = R"DOC(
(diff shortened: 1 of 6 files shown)


onnx/defs/generator/defs.cc
```diff
@@ -38,10 +70,26 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            (nullptr != value_ints),
+            (nullptr != value_float),
+            (nullptr != value_floats),
+            (nullptr != value_string),
+            (nullptr != value_strings)
+          };
+          if (std::count(non_null_attr.begin(), non_null_attr.end(), true) != 1)
             fail_shape_inference(
-                "Only one of the attributes 'value' or 'sparse_value' must be specified for a Constant node.");
+                "One and only one of the attributes 'value', 'value_*' or 'sparse_value' must be specified for a Constant node.");
```
jeremycochoy: The constraint changed: Here at least one attribute should be defined AND no more than one attribute should be defined.

onnx/defs/generator/defs.cc
```diff
@@ -51,6 +99,60 @@ ONNX_OPERATOR_SET_SCHEMA(
             return;
           }
 
+          if (nullptr != value_int) {
+            // OpSchema::Verify check ensures that the attribute value has_i():
+            if (!value_int->has_i())
+              fail_shape_inference("Attribute 'value_int' expect an integer.")
```
jeremycochoy: This is indeed a bit redundant with the type definition of value_int, but maybe redundancy is a good thing? 😅

onnx/defs/generator/defs.cc
```diff
@@ -26,6 +27,36 @@ ONNX_OPERATOR_SET_SCHEMA(
             "The value for the elements of the output tensor in sparse format.",
             AttributeProto::SPARSE_TENSOR,
             false)
+        .Attr(
+            "value_int",
+            "The value for the elements of the output tensor.",
```
TMVector: I would suggest something like:
> The value for the sole element for the scalar, int64, output tensor.

> The values for the elements for the 1D, int64, output tensor.

> The value for the sole element for the scalar, float32, output tensor.

and so on.

This would make it clear what the output tensor will actually look like (shape, element type).

onnx/defs/generator/defs.cc
```diff
@@ -51,6 +98,60 @@ ONNX_OPERATOR_SET_SCHEMA(
             return;
           }
 
+          if (nullptr != value_int) {
+            // OpSchema::Verify check ensures that the attribute value has_i():
+            if (!value_int->has_i())
+              fail_shape_inference("Attribute 'value_int' expect an integer.")
+            updateOutputElemType(ctx, 0, TensorProto::INT64);
+            appendDim(getOutputShape(ctx, 0), 1);
```
TMVector: IMO the scalar attributes should produce a scalar tensor, which should have zero dimensions https://github.com/onnx/onnx/blob/master/docs/IR.md#static-tensor-shapes

onnx/defs/generator/defs.cc
TMVector: There's also a message on line 168 that needs updating, although it should never be reached now that you check the `<= 1` part at the top.

onnx/defs/generator/defs.cc
```diff
@@ -51,6 +98,60 @@ ONNX_OPERATOR_SET_SCHEMA(
             return;
           }
 
+          if (nullptr != value_int) {
+            // OpSchema::Verify check ensures that the attribute value has_i():
+            if (!value_int->has_i())
+              fail_shape_inference("Attribute 'value_int' expect an integer.")
+            updateOutputElemType(ctx, 0, TensorProto::INT64);
+            updateOutputShape(ctx, 0, TensorShapeProto());
```
jeremycochoy: An alternative is simply `getOutputShape(ctx, 0)` since the shape attribute doesn't exists yet (I tried to pick the one that was semantically more sounding).

onnx/defs/generator/defs.cc
```diff
@@ -2,13 +2,14 @@
...
 #include <cmath>
+#include <algorithm>
 #include "onnx/defs/function.h"
 #include "onnx/defs/schema.h"
 
 namespace ONNX_NAMESPACE {
 static const char* Constant_ver11_doc = R"DOC(
-A constant tensor. Exactly one of the two attributes, either value or sparse_value,
-must be specified.
+A constant tensor. Exactly one of the provided attributes, either value, sparse_value,
```
wschin:
```suggestion
This operator produces a constant tensor. Exactly one of the provided attributes, either value, sparse_value,
```

onnx/defs/generator/defs.cc
```diff
@@ -51,6 +98,60 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              fail_shape_inference("Attribute 'value_int' expect an integer.")
+            updateOutputElemType(ctx, 0, TensorProto::INT64);
+            updateOutputShape(ctx, 0, TensorShapeProto());
+            return;
+          }
+
+          if (nullptr != value_ints) {
+            // OpSchema::Verify check ensures that the attribute value has ints.
+            if (value_ints->ints_size() < 1)
+              fail_shape_inference("Attribute 'value_ints' expect a collection of integers.")
```
wschin:
```suggestion
              fail_shape_inference("Attribute 'value_ints' expect a list of integers.")
```

onnx/defs/generator/defs.cc
```diff
@@ -51,6 +98,60 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              fail_shape_inference("Attribute 'value_float' expect a float.")
+            updateOutputElemType(ctx, 0, TensorProto::FLOAT);
+            updateOutputShape(ctx, 0, TensorShapeProto());
+            return;
+          }
+
+          if (nullptr != value_floats) {
+            // OpSchema::Verify check ensures that the attribute value has ints.
+            if (value_floats->floats_size() < 1)
+              fail_shape_inference("Attribute 'value_floats' expect a collection of floats.")
```
wschin:
```suggestion
              fail_shape_inference("Attribute 'value_floats' expect a list of floats.")
```

onnx/defs/generator/defs.cc
```diff
@@ -51,6 +98,60 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              fail_shape_inference("Attribute 'value_string' expect a string.")
+            updateOutputElemType(ctx, 0, TensorProto::STRING);
+            updateOutputShape(ctx, 0, TensorShapeProto());
+            return;
+          }
+
+          if (nullptr != value_strings) {
+            // OpSchema::Verify check ensures that the attribute value has ints.
+            if (value_strings->strings_size() < 1)
+              fail_shape_inference("Attribute 'value_strings' expect a collection of strings.")
```
wschin:
```suggestion
              fail_shape_inference("Attribute 'value_strings' expect a list of strings.")
```

onnx/defs/generator/defs.cc
```diff
@@ -2,13 +2,14 @@
 // Licensed under the MIT license.
 
 #include <cmath>
+#include <algorithm>
 #include "onnx/defs/function.h"
 #include "onnx/defs/schema.h"
 
 namespace ONNX_NAMESPACE {
 static const char* Constant_ver11_doc = R"DOC(
```
wschin: @linkerzhang, do you think we should bump the operator's version?

onnx/defs/generator/defs.cc
gramalingam: Change to "updateOutputShape(ctx, 0, {})". I think "TensorShapeProto()" would represent an unknown shape (of unknown rank). It is more precise to specify output tensor is of rank 0.

onnx/defs/generator/defs.cc
```diff
@@ -51,6 +98,60 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            updateOutputShape(ctx, 0, TensorShapeProto());
+            return;
+          }
+
+          if (nullptr != value_ints) {
+            // OpSchema::Verify check ensures that the attribute value has ints.
+            if (value_ints->ints_size() < 1)
+              fail_shape_inference("Attribute 'value_ints' expect a list of integers.")
+            updateOutputElemType(ctx, 0, TensorProto::INT64);
+            appendDim(getOutputShape(ctx, 0), value_ints->ints_size());
```
gramalingam: May be slightly simpler to write "updateOutputShape(ctx, 0, {value_ints->ints_size()})". Same comment for other statements below.

jeremycochoy: This syntax do not compile. The compiler don't know how to create a `TensorShapeProto` from this vector.

onnx/defs/generator/defs.cc
jeremycochoy: I think some constructors / types specifications are missing to allow this syntax to compile. :/

gramalingam: May be the compiler is unable to infer the type of the parameter from an empty list {}. In any case, I realize that the existing code is fine. I was confusing this with a type without a defined shape ... a TensorShapeProto() should represent a shape with an empty list of dims, so that's fine. Please ignore my previous suggestion.

onnx/defs/generator/defs.cc
gramalingam: Sorry, please ignore my suggestion. It needs to be a bit more complicated, needs a Dim, not integer value.

PR #2596, opened 2020-02-12
Kezhan/function update

linkerzhang:
Add more clarification about function.

### DIFF
--- a/docs/IR.md
+++ b/docs/IR.md
@@ -25,7 +25,7 @@ ONNX is an open specification that consists of the following components:
 
 3)  Definitions of built-in operators.
 
-Of these, #1 and #2 are covered herein; the built-in operators are covered separately in documents listed at the end of this. Specifically, built-in operators are divided into a set of primitive operators and functions. A function refers to a composed operator which can be semantically represented by a sub-graph (called function body) with using primitive ops. Functionality-wise, an ONNX compatible framework or runtime may inline a function body to execute it if it does not have corresponding implementation of the function.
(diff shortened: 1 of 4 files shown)


docs/IR.md
```diff
@@ -25,7 +25,7 @@ ONNX is an open specification that consists of the following components:
 
 3)  Definitions of built-in operators.
 
-Of these, #1 and #2 are covered herein; the built-in operators are covered separately in documents listed at the end of this.
+Of these, #1 and #2 are covered herein; the built-in operators are covered separately in documents listed at the end of this. Specifically, built-in operators are divided into a set of primitive operators and functions. A function refers to a composed operator which can be semantically represented by a sub-graph (called function body) with using primitive ops. Functionality-wise, an ONNX compatible framework or runtime may inline a function body to execute it if it does not have corresponding implementation of the function.
```
gramalingam: How about changing the middle line to "A function is an operator whose semantics is formally expressed via expansion into a sub-graph (called the function body) using other operators (and functions). Functionality-wise ..."

linkerzhang: Your wording is better, adopted. Thanks!

onnx/defs/gen_doc.py
```diff
@@ -202,9 +202,10 @@ def format_value(value):  # type: (Any) -> Text
         s += '</dl>\n'
 
     # Function Body
-    if schema.has_function or schema.has_context_dependent_function:  # type: ignore
-        s += '\n#### Function\n'
-        s += '\nThe Function can be represented as a function.\n'
+    # TODO: this should be refactored to show the function body graph's picture (DAG).
```
TMVector: This isn't always going to be possible with context dependent functions :/

PR #2597, opened 2020-02-12
Allow function body to rely on other operator sets

linkerzhang:
A function body graph can rely on other operator sets by specifying its own opset_imported field, besides the operator set that this function belongs to.

### DIFF
--- a/onnx/defs/schema.cc
+++ b/onnx/defs/schema.cc
@@ -746,12 +746,6 @@ bool OpSchema::BuildContextDependentFunction(
 }
 
 OpSchema& OpSchema::FunctionBody(const std::vector<NodeProto>& func_nodes) {
-  // By default, the function body graph is relying on the OperatorSet this
-  // function belongs to.
-  auto relied_opset = function_body_.mutable_opset_import()->Add();
-  relied_opset->set_domain(this->domain());
-  relied_opset->set_version(this->SinceVersion());
-
   for (const auto node : func_nodes) {
     auto new_node = function_body_.add_node();
     new_node->CopyFrom(node);
@@ -794,6 +788,11 @@ void OpSchema::BuildFunction(FunctionProto& function_body) const {
   for (auto& a : attributes_) {
     function_body.add_attribute(a.first);
   }
+  // By default, the function body graph is relying on the OperatorSet this
+  // function belongs to.
+  auto relied_opset = function_body_.mutable_opset_import()->Add();
+  relied_opset->set_domain(this->domain());
+  relied_opset->set_version(this->SinceVersion());
 }
 
 void OpSchema::Finalize() {
--- a/onnx/defs/schema.cc
+++ b/onnx/defs/schema.cc
@@ -790,7 +790,7 @@ void OpSchema::BuildFunction(FunctionProto& function_body) const {
   }
   // By default, the function body graph is relying on the OperatorSet this
   // function belongs to.
-  auto relied_opset = function_body_.mutable_opset_import()->Add();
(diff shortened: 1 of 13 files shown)


onnx/defs/schema.cc
```diff
@@ -705,26 +730,45 @@ void OpSchema::ParseAndSetTypes(
...
+    const FunctionBodyBuildContext& ctx,
+    FunctionProto& functionProto) const {
   if (functionBuilder_)
     return functionBuilder_(ctx, *this, functionProto);
   else
     return false;
 }
 
 OpSchema& OpSchema::FunctionBody(const std::vector<NodeProto>& func_nodes) {
+  // By default, the function body graph is relying on the OperatorSet this
```
linkerzhang: this is the non-formatting change.

onnx/defs/schema.cc
```diff
@@ -705,26 +730,45 @@ void OpSchema::ParseAndSetTypes(
...
+  relied_opset->set_version(this->SinceVersion());
+
   for (const auto node : func_nodes) {
     auto new_node = function_body_.add_node();
     new_node->CopyFrom(node);
   }
   return *this;
 }
 
+OpSchema& OpSchema::FunctionBody(
```
linkerzhang: this is non-formatting change.

onnx/defs/schema.h
```diff
@@ -663,6 +663,7 @@ class OpSchema final {
   }
 
   OpSchema& FunctionBody(const std::vector<NodeProto>& func_nodes);
+  OpSchema& FunctionBody(const std::vector<NodeProto>& func_nodes, const std::vector<OperatorSetIdProto>& opsets);
```
linkerzhang: API changes for dynamic function will be added later (maybe in a separate PR).

onnx/defs/schema.cc
```diff
@@ -97,58 +97,83 @@ OpSchemaRegistry* OpSchemaRegistry::Instance() {
...
     const auto& param_type = ctx.getInputType(in_idx);
-    const auto& all_types  = param.GetTypes();
-    if (nullptr == param_type || param_type->value_case() == TypeProto::VALUE_NOT_SET) {
+    const auto& all_types = param.GetTypes();
+    if (nullptr == param_type ||
+        param_type->value_case() == TypeProto::VALUE_NOT_SET) {
       continue;
-    } else if (!all_types.empty() && all_types.find(Utils::DataTypeUtils::ToType(*param_type)) == all_types.end()) {
-      fail_check(param.GetName(), " typestr: ", type_str, ", has unsupported type: ", *Utils::DataTypeUtils::ToType(*param_type));
+    } else if (
```
gramalingam: On a different note: can we increase the line-width used for clang-format please? I feel it is too small and breaks up code into too many lines. (Nothing to do with this PR, just struck me.)

onnx/defs/schema.cc
```diff
@@ -705,26 +730,45 @@ void OpSchema::ParseAndSetTypes(
...
     auto new_node = function_body_.add_node();
     new_node->CopyFrom(node);
   }
   return *this;
 }
 
+OpSchema& OpSchema::FunctionBody(
+    const std::vector<NodeProto>& func_nodes,
+    const std::vector<OperatorSetIdProto>& relied_opsets) {
+  for (auto& relied_opset : relied_opsets) {
```
gramalingam: I think it would be a good idea to include the function's domain/since-version in this case also (or, at least if it is not a duplicate).

linkerzhang: it's included, it's call the other overloaded FunctionBody, which adds that.

PR #2601, opened 2020-02-14
FunctionExpandHelper - remove the original function node

postrational:
The `FunctionExpandHelper` currently expands the selected node with a given `FunctionProto`, but the original node remains in the graph. 

This PR adds logic which removes the original function node from the graph.


linkerzhang:
@postrational do you mind to update the branch please?

onnx/defs/function.cc
```diff
@@ -13,6 +13,41 @@ std::string InteralTensorNameGenerator(
...
+    return false;
+  }
+  if (lhs.op_type() != rhs.op_type()){
+    return false;
+  }
+  if(lhs.attribute_size()!=rhs.attribute_size()){
+    return false;
+  }
+  for(int i=0; i<lhs.attribute_size(); i++){
+    if(lhs.attribute(i).name()!=rhs.attribute(i).name()){
```
linkerzhang: attribute value should also be checked?

linkerzhang: and also domain? and doc string?

do you think it's better to check the two nodeproto's serialized string instead?

onnx/defs/function.cc
```diff
@@ -105,6 +140,14 @@ void FunctionExpandHelper(
       }
     }
   }
+
+  // Remove function node from graph
+  for (int i = 0; i < g.node().size(); ++i) {
```
linkerzhang: can this be done when doing in-lining above? this removal may be not effective if the model is very large.

onnx/defs/function.cc
Ewa21: Yes, I agree that `SerializeToString` method from Protobuf will work better than my custom function.

onnx/defs/function.cc
postrational: @linkerzhang This part itself cannot be done in the in-lining above as it's currently written.

AFAICT, current in-lining is based on adding all the nodes of the expanded subgraph to the end of the main graph's node list:

https://github.com/onnx/onnx/blob/b8966dc607b4a4f495c68c2c44996cada4c2c843/onnx/defs/function.cc#L79-L80

The graph this approach generates is **not sorted** topologically.

We could try to expand this PR, to replace the function node with the expanded subgraph (instead of deleting it after the expansion).

Issue #2507, opened 2019-12-17
Reshape operator -- handling zero values is confusing


jignparm:
The current implementation of `Reshape` states:

`A dimension could also be 0, in which case the actual dimension value is unchanged (i.e. taken from the input tensor).`

This is inconsistent with NumPy and TensorFlow, and moreover causes unexpected shape behavior. For example, if the input has shape **[0,255]** and the request is to reshape to **[255, 0]**, then the current `Reshape` op try to reshape the input to **[255, 255]**, which is not intuitive and seemingly incorrect.

This unexpected behavior is causing some TF models (converted to ONNX) to fail at runtime.

Proposal for opset-12 is to remove the statement above in the current spec, and bring `Reshape` behavior in line with NumPy behavior.

If there are specific reasons why this should not be changed, a clarifying example would be great.

linkerzhang:
looks to me a reasonable fix.

jignparm:
Proposed 2 separate PRs to handle this. Only 1 should be merged (preferably #2553), contingent on feedback.

1> PR #2553 : Make it work like NumPy 
2> PR #2552 : Retain current behavior, but make it work like NumPy if needed via new attribute 'allowzero'

neginraoof:
Can we keep the same behavior with -1 dim value?

jignparm:
> Can we keep the same behavior with -1 dim value?

Can you clarify? The current behavior of reshape is:

* 0 = copy from input
* -1 = infer from remaining dimensions

The 2 PRs above change the meaning of '0' values, but not '-1'. Did you mean that there should be a change in the meaning of -1?

PR #2602, opened 2020-02-14
Modify Expand to support -1 dim value [Release 1.7]

neginraoof:
SImilar to Pytorch expand op, support shape input with -1 dim value, in which case the actual dimension value will remain unchanged.


linkerzhang:
@ebarsoum @postrational "-1" has different meanings across ONNX ops. It may be good to have a consistent definition across all of them to avoid confusion. See issue #2507

spandantiwari:
@neginraoof  - would it be OK to use '0' instead of '-1' for keeping the dimension unchanged. I think there are other ops that are use 0 for that semantics. We should try to be consistent. 

@linkerzhang - Issue https://github.com/onnx/onnx/issues/2507 has been open for a while, and I see multiple options open for that. Do you know if any decision has been made on that? 
If yes, then we can follow the same semantics here.

spandantiwari:
cc: @houseroad

gramalingam:
My understanding from issue 2507 is that using 0 is problematic, and that -1 is better.

docs/Operators.md
```diff
@@ -5146,6 +5151,40 @@ expect(node, inputs=[data, new_shape], outputs=[expanded],
...
+<summary>dim_negative</summary>
+
+```python
+node = onnx.helper.make_node(
+    'Expand',
+    inputs=['data', 'new_shape'],
+    outputs=['expanded'],
+)
+shape = [3, 3, 1]
+new_shape = [-1, 3, 6]
```
gramalingam: Why can't we use new_shape = [1, 3, 6] instead?

neginraoof: This test is for the case when we have -1 in expand shape input.

neginraoof:
Just a note:
I recently updated constant of shape to create multidimensional tensors with dim value 0.
https://github.com/onnx/onnx/pull/2567

docs/Operators.md
gramalingam: I was using this to illustrate my previous question: why do we need this extension of the op, if we can use "1" instead? Am I misunderstanding something?

neginraoof: We do not always know the dim value we want to expand to.
Let's say you have a tensor of size (?, ?, 1) and you wan to expand to (?, ?, 5) where ? could be unknown.

gramalingam: Then, you just specify the new-shape to be (1, 1, 5) ... if you apply this to a tensor of size (M,N,1), won't you end up with the shape (M, N, 5)?

neginraoof: No,  you cannot expand (M, N, 1) tensor to (1, 1, K)

neginraoof:
@linkerzhang 
Thanks for pointing the PR. I believe -1 is a good choice, specially if we decide to allow zero.

docs/Operators.md
gramalingam: Bidirectional broadcast between the shapes (1, 1, K) and (M, N, 1) will produce the shape (M, N, K). And I believe Expand (the existing version) does bidirectional-broadcast, as far as I can see. So, what am I missing?

gramalingam: My understanding from the existing documentation is that ONNX Expand differs from numpy's broadcast_to in that ONNX uses bidirectional broadcast, unlike numpy.

neginraoof: This is more of an export issue, since for exporting PT or TF models that allow -1, we need to change the shape input tensor.

gramalingam: I see. So, the semantics of "-1" will be the same as that of "1" (in ONNX)? (Not sure if PT or TF have the exact same semantics for 1 ... it does look odd to have two ways of saying the same thing.)

neginraoof:
Closing this for now. As suggested by @gramalingam, we can have a workaround in export.

PR #2603, opened 2020-02-14
Fix CI build break

snnn:
1. Add conda's Library\bin folder to the PATH, because libprotobuf.dll is there.
2. Fix the protobuf version mismatch problem. Currently the C/C++ part is 3.9.2 but the python part is 3.11.3. This PR changes them to the same version: 3.11.3
3. USE_MSVC_STATIC_RUNTIME should be disabled when the libprotobuf is from conda
4. We should set the gtest_force_shared_crt variable according to the value of USE_MSVC_STATIC_RUNTIME

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,10 +30,12 @@ install:
 - cmd: git submodule update --init --recursive
 
 before_build:
-- cmd: SET PATH=%CONDA_PREFIX%;%CONDA_PREFIX%\Scripts;%PATH%
+#TODO:We should put libprotobuf.dll within onnx's package.
+#Then we don't need to add %CONDA_PREFIX%\Library\bin at here
+- cmd: SET PATH=%CONDA_PREFIX%;%CONDA_PREFIX%\Scripts;%CONDA_PREFIX%\Library\bin;%PATH%
 - cmd: SET _prefix=%CONDA_PREFIX:~0,14%
 - cmd: SET _arch=%CONDA_PREFIX:~15,18%
-- cmd: conda install -y -c conda-forge libprotobuf=3.9.2
+- cmd: conda install -y libprotobuf=3.11.3 protobuf=3.11.3 numpy
 - >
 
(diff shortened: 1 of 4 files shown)


appveyor.yml
```diff
@@ -47,7 +49,14 @@ build_script:
...
 - cmd: cd c:\projects\onnx
 - cmd: set ONNX_BUILD_TESTS=1
-- cmd: set CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+#The latest libprotobuf Windows packages in anaconda was built with          
+#        -Dprotobuf_BUILD_SHARED_LIBS=ON 
+#        -Dprotobuf_MSVC_STATIC_RUNTIME=OFF
+#We must align with them. We must set USE_MSVC_STATIC_RUNTIME and 
+#ONNX_USE_PROTOBUF_SHARED_LIBS based on their config
+- cmd: set USE_MSVC_STATIC_RUNTIME=0
+# Don't put quotation marks in the following line
```
vinitra: Is this comment needed?

vinitra:
Thanks for the PR!

appveyor.yml
snnn: Take a look of the old line. It's wrong. It shouldn't have the mark.

snnn: With the  quotation mark, it won't work as expected.

snnn:
BTW, this is why ONNX_USE_MSVC_STATIC_RUNTIME should be OFF when ONNX_USE_PROTOBUF_SHARED_LIBS is ON.

Each copy of the CRT library has its own heap manager, allocating memory in one CRT library and passing the pointer across a DLL boundary to be freed by a different copy of the CRT library is a potential cause for heap corruption.

So it will be dangerous when you pass a protobuf object between DLL boundaries and if the ownership of the object is transferred. 
https://docs.microsoft.com/en-us/cpp/c-runtime-library/potential-errors-passing-crt-objects-across-dll-boundaries?view=vs-2019

PR #2604, opened 2020-02-14
Remove PyTorch tests from CI

neginraoof:
Removing PyTorch-ONNX tests from CI


neginraoof:
cc @houseroad @postrational for review.

BowenBao:
LGTM, please rebase with master, and remove the [WIP] tag. Thanks

gramalingam:
The broken circle-ci adds some complexity to PR merging process as there are questions of whether it is okay to merge the PR or not. Is anyone looking at the circle-ci failures? If not, it seems like a waste to have it.

gramalingam:
@houseroad , that's useful to know. Can you clarify which text-cases does this test (models, nodes)?

snnn:
Maybe it is out of topic, but honestly speaking, I don't think onnxruntime can do too much things there. 

If it the op has been implemented in onnxruntime, then we can verifiy if onnxruntime's result matches with the reference in ONNX.  But such things only happen after the op finalized and released. Too late. 

If it the op isn't implemented yet, then currently onnxruntime can't run it, even when it has function body. We should have a way to verify if the functon body works but unfortunately onnxruntime can't help much now. I doubt what pytorch can do too.  We need a plan. We need to test it. But the test tool mustn't only partially works, or just do its best effort.

neginraoof:
@gramalingam @houseroad 
We are currently only running Caffe2 backend tests as part of CI (ORT tests are ignored).
Shall we decide which backend tests to keep? Also, do we want to keep test_operators, test_models, and test_onnx_opsets?

askhade:
> If we remove this CI, then we don't have a backend test to verify the ONNX testing data. Then it's very easy to introduce wrong testing data. I am okay to switch to ONNX runtime to replace PyTorch/Caffe2.

@houseroad : Can you please elaborate what exactly is not being tested?

@neginraoof : Can you restart this PR? Let's understand what testing is missing in azurepipelines  and try to add it so that we can remove circle ci

neginraoof:
I'm rebasing this.

linkerzhang:
I agree with Lu that having one backend (either pytorch or onnxruntime) added into ONNX CI as monitoring/verifying the correctness of testing data is a good idea. It can also help backend (pytorch or onnxruntime) to know when there's a break, though they should not be "required" CI pass. 

@askhade maybe instead of removing pytorch, onnxruntime should also be added here, but both of them are not "required" CI passes.

jacky82226:
Provide another viewpoint here:
I think at least the tests for old ops in [test_pytorch_onnx_caffe2](https://github.com/pytorch/pytorch/blob/9a3e16c773496b16e6c02f6e3e020be5bb485ea0/test/onnx/test_pytorch_onnx_caffe2.py#L2497) might be unnecessary.
I am enabling shape inference check in https://github.com/onnx/onnx/pull/2783, but I bumped into some test failures for op7 and op7 tests. The reason is current onnx does not support the deprecated op: ConstantFill which these tests are using.

PR #2606, opened 2020-02-16
Add operators LessOrEqual and GreaterOrEqual (as functions)

jeremycochoy:
# LessOrEqual and GreaterOrEqual

Looking at the list of missing operators https://github.com/onnx/onnx/issues/1646 the LessOrEqual and GreaterOrEqual have not been implemented yet. ( from issue https://github.com/onnx/onnx/issues/870 )

In this PR I define both operators and implement them as functions.

## Description of operators

  * Operator `LessOrEqual` can be implemented as `Not \circ Greater`.
  * Operator `GreaterOrEqual` can be implemented as `Not \circ Less`.

## The equivalent numpy code
```
# Less Or Equal
z = np.less_equal(x, y)
# Greater Or Equal
z = np.greater_equal(x, y)
```


TMVector:
This won't handle NaNs properly, so maybe the `(a > b or a == b)` form is better?
```python
import numpy as np
np.nan >= 1 # False
not (np.nan < 1) # True!
```

jeremycochoy:
> This won't handle NaNs properly, so maybe the `(a > b or a == b)` form is better?
> 
> ```python
> import numpy as np
> np.nan >= 1 # False
> not (np.nan < 1) # True!
> ```

Good point... I wanted to avoid this spelling to prevent two call to (Tensor,Tensor)->Tensor operators, but right now it is not equivalent to its definition (indeed pretty annoying that on floats `> or ==` is different from `not >`).

So if NaNs are considered part of the domain on which operators should be defined, we would have to either change the actual definition, or use the `(a > b or a == b)` implementation. (It also make me think that there is some inconsistencies in pytorch in this case ...)

Which choice is the best?

TMVector:
I think it's fine to have not-amazing performance for functions, since backends can implement kernels if they care. However, yes, you could use the `ContextDependentFunction` functionality to have a separate function body for int types and float types

A third option could be to not provide function bodies at all, to push backends to implement it natively, since it's a fairly primitive operation.

jeremycochoy:
> ContextDependentFunction

Not sure it is worth adding complexity only for integer models (I personally do not need it 😊). But If you remember one implementation using the ContextDependentFunction feature in the code base, I'd like to have a look just for curiosity.

I changed the graph to the `(<)OR(=)`  so that we align to Numpy's behavior.

TMVector:
The `ContextDependentFunction` feature was added in https://github.com/onnx/onnx/pull/2551/files#diff-11e5546ae75100fbca62f077a8b8a904 which also has an example of using it.

linkerzhang:
@jeremycochoy can you resolved the conflicts?

linkerzhang:
@jeremycochoy do you mind to help resolving the conflicts? thank you!

jeremycochoy:
> @jeremycochoy do you mind to help resolving the conflicts again? thank you! New check-ins introduce more conflicts.....

Merged the master and regenerated the documentation. It should be good (just waiting for ci to complete).

linkerzhang:
@jeremycochoy did you sign the CLA? as I'm seeing the licence/cla above is not green :)

PR #2607, opened 2020-02-17
Generate node backend tests for celu operator

TMVector:
Running `tools/update_doc.sh` from master creates these previously untracked files


jeremycochoy:
😱 Something is wrong with the Celu backend tests. I will see if I can fix it. Sorry. 😞

TMVector:
No worries @jeremycochoy, happens to everyone :)

I think the problem might actually be with the code which creates the expanded-functions test not handling ref attrs correctly[0]. I'll give fixing it a quick try now in this PR.

[0] https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/__init__.py#L59-L63

jeremycochoy:
Thanks. :)

I am also wondering if the lines at https://github.com/onnx/onnx/blob/807c62cf7e4c96ce49040bcf073b7e4a054f28a5/onnx/defs/function.cc#L75-L77 may require some modification too ?

jeremycochoy:
One last mistake I found that should result in test failing:
https://github.com/onnx/onnx/blob/c978d102c9fe5b3609c6d8ff87e2ab41149edd42/onnx/backend/test/case/node/celu.py#L17-L21

The attribute `alpha=2` (should be `alpha=2.`) is not forwarded to the node, so there should be a mismatch between the backend computing with `alpha=1.` and numpy computing with `alpha=2.`.

linkerzhang:
please fix Linux ci failure.


+flake8
2388
./onnx/backend/test/case/node/celu.py:22:18: E251 unexpected spaces around keyword / parameter equals
2389
./onnx/backend/test/case/node/celu.py:22:20: E251 unexpected spaces around keyword / parameter equals

onnx/backend/test/case/node/celu.py
```diff
@@ -14,10 +14,12 @@ class Celu(Base):
 
     @staticmethod
     def export():  # type: () -> None
+        alpha = 2
```
jeremycochoy: Shouldn't it be alpha=2. for the tests to pass?

TMVector: Hmm weird -- I must have somehow missed including the actual source when I fixed it in 65fc77f7a84a3dc15d26bfd70175c2ced7289984 :man_facepalming: Good catch!

PR #2608, opened 2020-02-17
Add support for integer tensors to Min and Max

TMVector:
Two more operators for https://github.com/onnx/onnx/issues/2528. Added tests for int8.


docs/TestCoverage.md
```diff
@@ -4663,6 +4663,23 @@ expect(node, inputs=[data_0, data_1], outputs=[result],
        name='test_max_two_inputs')
 ```
 
+</details>
+<details>
+<summary>max_int8</summary>
+
+```python
+data_0 = np.array([ 3, 2,-1]).astype(np.int8)
```
wschin: May we have one test for each added type?

onnx/defs/math/defs.cc
```diff
@@ -623,10 +623,12 @@ All inputs and outputs must have the same data type.
         "T",
         OpSchema::Variadic);
     schema.Output(0, name, "Output tensor.", "T");
-    schema.TypeConstraint(
-        "T",
-        {"tensor(float16)", "tensor(float)", "tensor(double)"},
-        "Constrain input and output types to float tensors.");
+    if (with_type_constraint) {
+      schema.TypeConstraint(
```
wschin: Can we make the information to construct type constraint an input of this function? For example,
```
std::function<void(OpSchema&)> ElementwiseMultiOpDocGenerator(
    const char* name,  const vector<string> allowed_types = {"tensor(float16)", "tensor(float)", "tensor(double)"})
```
Using a flag is less explicit than passing the right information directly.

docs/TestCoverage.md
TMVector: I didn't do that because a [previous comment](https://github.com/onnx/onnx/pull/2532#issuecomment-571787425) on a different PR said there were too many tests, but I would very happily add one for each type.

wschin: I don't feel that comment is valid. Any code path not tested can be broken. :)

onnx/defs/math/defs.cc
TMVector: I removed the type constraint from that function all together, I think that's clearest :)

linkerzhang:
@TMVector please resolve the conflicts.

PR #2611, opened 2020-02-18
Fixed CumSum axis is not optional and test case errors 

pluradj:
The axis input for CumSum is described as optional, however it was not denoted
correctly in the operator schema.

Signed-off-by: Jason Plurad <pluradj@us.ibm.com>


askhade:
@pluradj : Is there anything blocking this PR? Can you resolve conflicts and run CIs again?

askhade:
@pluradj : What is the value addition here for making axis as optional? Per the op schema axis is not optional and since multiple frameworks have different value for axis can we simply fix the documentation here to remove optional from the input description for this op? This way it is clear that axis is not an optional input?

pluradj:
I'm not the original author of the CumSum spec (https://github.com/onnx/onnx/pull/2030). The original author @jeffsaremi described `axis` as optional in the [description](https://github.com/onnx/onnx/blob/2c03cff041798ec23dfe419c6567c78db59d7234/onnx/defs/math/defs.cc#L1516) but didn't declare it as optional in the schema. I don't know what the correct intentions were in this situation, and I don't have a strong preference either way. So if the project committers would rather have this PR update the doc instead, I'd be happy to make that update. Just let me know.

askhade:
@pluradj , @postrational : In the version 11 of this operator even though the intention of the author was to make axis optional the schema was not defined in such a way. In this PR the schema is being changed to make axis optional and therefore I think this warrants a opset version bump.

postrational:
Well, we need to make a decision, either we bump the op version, or we just remove the "optional" wording.
We should remove the "optional" wording from the current version in either case to be consistent with schema.

askhade:
> Well, we need to make a decision, either we bump the op version, or we just remove the "optional" wording.
> We should remove the "optional" wording from the current version in either case to be consistent with schema.

I am leaning for removing optional wording because the schema will remain intact and therefore we don't need any version bump. I think we should avoid unnecessary version bumps this creates a lot of overhead for onnx backends.

pluradj:
Updated doc so `axis` is required, no breaking changes. @postrational @askhade Please review and let me know if you'd like any other changes before merging.

docs/Operators.md
```diff
@@ -3903,7 +3903,7 @@ node = onnx.helper.make_node(
     outputs=['y']
 )
 x = np.array([1., 2., 3., 4., 5.]).astype(np.float64)
-axis = np.array([0]).astype(np.int32)
+axis = np.array(0).astype(np.int32)
```
postrational, line 3906: Why are we removing the array brackets here? 

According to numpy docs this argument should be an array or an object exposing the array interface:
https://numpy.org/doc/stable/reference/generated/numpy.array.html

pluradj: The CumSum specification documentation states that `axis` is a 0-D tensor (scalar). We are removing the array brackets so `axis` isn't a 1-D tensor.

```
>>> axis = np.array([0]).astype(np.int32)
>>> axis.ndim
1
>>> axis = np.array(0).astype(np.int32)
>>> axis.ndim
0
```

pluradj: numpy array scalars https://numpy.org/doc/stable/reference/arrays.scalars.html

pluradj:
Is there anything else need to be updated to get this fix merged? Please let me know. Thanks.

postrational:
@ebarsoum Are we ready to merge this? If so, please do.

pluradj:
@postrational @ebarsoum @askhade Could somebody point me at the ONNX commit process? What does it take to get a PR through review, approval, and merge to master? Where can I find a list of ONNX committers? Perhaps I'm missing more information here to help move things along, but I don't see anything mentioned in the [CONTRIBUTING.md](https://github.com/onnx/onnx/blob/master/docs/CONTRIBUTING.md).

I'm a first time contributor to ONNX, and I'm getting discouraged from contributing more because this rather small unit test case fix is taking almost 6 months to get merged. Thanks.

gramalingam:
Yes, this is ready to be merged in. Unfortunately, we need to update the branch and wait for CI to complete.

pluradj:
Appreciate your help @gramalingam and thank you for getting this merged.

Any input on the questions I listed previously would go a long way in helping other contributors in the future. Thanks.

> Could somebody point me at the ONNX commit process? What does it take to get a PR through review, approval, and merge to master? Where can I find a list of ONNX committers? Perhaps I'm missing more information here to help move things along, but I don't see anything mentioned in the CONTRIBUTING.md.

PR #2616, opened 2020-02-21
Adding UnfoldToDepth op [1.7 Release]

neginraoof:
Adding ImageToCol operator.
This is equivalent of PyTorch nn.Unfold and TensorFlow Tf.extract_image_patches


neginraoof:
cc @postrational @houseroad for review

neginraoof:
cc @wschin @gramalingam @BowenBao @lara-hdr for review.
Thanks a lot.

onnx/defs/nn/defs.cc
```diff
@@ -184,6 +184,98 @@ void convPoolShapeInference(
...
+  std::vector<int64_t> block_shape;
+  if (getRepeatedAttribute(ctx, "block_shape", block_shape)) {
+    if (block_shape.size() != n_input_dims) {
+      fail_shape_inference("Attribute size has incorrect size");
+    }
+  } else {
+    fail_shape_inference("Attribute block_shape must be specified");
+  }
+
+  size_t block_num_element = input_shape.dim(1).dim_value();
```
gramalingam: I think this also needs a has_dim_value check: if it is not known, the corresponding output-dim should be left undefined.

onnx/backend/test/case/node/imagetocol.py
```diff
@@ -0,0 +1,86 @@
...
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+from typing import List
+
+
+def im2col_2d_reference_implementation(x, shape, padding=[0, 0]):  # type: (np.ndarray, List[int], List[int]) -> np.ndarray
```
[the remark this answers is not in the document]
neginraoof: I've updated the implementation with padding and strides.

onnx/defs/nn/defs.cc
```diff
@@ -184,6 +184,98 @@ void convPoolShapeInference(
...
+      ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+
+  *output_shape->add_dim() = input_shape.dim(0);
+  output_shape->add_dim()->set_dim_value(block_num_element);
+
+  size_t last_dim_size = 1;
+  int block_shape_size = static_cast<int>(block_shape.size());
+  for (int i = 0; i < block_shape_size; ++i) {
+    if (!input_shape.dim(2 + i).has_dim_value()) {
+      continue;
```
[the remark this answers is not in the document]
neginraoof: Thanks,
I think we should add the dimension with unknown value in this case. Fixed this.

onnx/backend/test/case/node/imagetocol.py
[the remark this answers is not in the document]
neginraoof: I've added stride and padding for the reference implementation.

docs/Operators.md
```diff
@@ -7135,6 +7136,121 @@ Other versions of this operator: <a href="Changelog.md#If-1">If-1</a>
 </dl>
 
 
+### <a name="ImageToCol"></a><a name="imagetocol">**ImageToCol**</a>
+
+  The ImageToCol operator rearranges blocks from an input tensor into columns, and returns
+  the concatenated columns.
```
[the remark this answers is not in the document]
neginraoof: I added explanations for input/output shapes.

neginraoof:
All tests are passing except for a timeout from:
 Xcode: xcode9.3 PYTHON_VERSION=python2

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated

I'm triggering the CI again.

onnx/defs/nn/defs.cc
```diff
@@ -783,6 +877,78 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        "convolution. The output dimensions are functions "
+        "of the kernel size, stride size, and pad lengths.",
+        "T");
+    schema.TypeConstraint(
+        "T",
+        {"tensor(float16)", "tensor(float)", "tensor(double)"},
+        "Constrain input and output types to float tensors.");
+    schema.Attr(
+        "block_shape",
+        "The size of the extracted blocks [H, W].",
```
BowenBao: nit: update this to [D1, D2, ..., Dn] as well

docs/Operators.md
```diff
@@ -7135,6 +7136,153 @@ Other versions of this operator: <a href="Changelog.md#If-1">If-1</a>
 </dl>
 
 
+### <a name="ImageToCol"></a><a name="imagetocol">**ImageToCol**</a>
```
linkerzhang: This op looks to me is a tensor layout change op, which is similar with SpaceToDepth, so putting it into the same folder as "SpaceToDepth" makes more sense.

docs/Operators.md
```diff
@@ -7135,6 +7136,153 @@ Other versions of this operator: <a href="Changelog.md#If-1">If-1</a>
...
+
+  The ImageToCol operator extracts sliding blocks from an input tensor, and concatenates these blocks
+  in the last dimension.
+  Given an input of shape (N x C x D1 x D2 ... x Dn), output would be a 3-D tensor of shape:<br/>
+  
+  ```(N, C * reduce-mul(kernel_size), reduce-mul(block_size))```
```
linkerzhang: the "kernel_size" is confusing ...

docs/Operators.md
linkerzhang: Pytorch name "unfold" looks to be better than this one, since it does not have to be an image, and it's indeed unfolding a window/block.

linkerzhang: With this new op, can "SpaceToDepth" be replaced/deprecated? looks to me this one is more general.

To replace "SpaceToDepth", a little bit more needs to be updated, as "SpaceToDepth" is not squeezing the last 2 dimensions into one (while this op does). Maybe adding all parameters used to define the block into "SpaceToDepth" is better?

linkerzhang:
Looks to me:
an updated “SpaceToDepth” (with more flexible block definition) + “Squeeze” is a better idea than adding this new op. Thoughts?

docs/Operators.md
neginraoof: Thanks @linkerzhang 
I talked about this with Spandan and Emad as well. So, as you mentioned, this operator is a the more general version of SapceToDepth (except the mode attribute is not supported). I think we can modify this operator's name and functionality further (maybe not flattening the last dimension, support for CRD/DCR re-arrangemen modes, etc.), and change SapceToDepth to a function that uses this operator internally.
However,  SpaceToDepth and DepthToSpace are currently complimentary ops, and I think it's better to keep them as such. We can add an operator equivalent for ColToImage (with a more general naming), and change both SpaceToDepth and DepthToSpace operators as functions.

neginraoof: Yes, I agree. I will update the operator name.

linkerzhang: do you have plan to do the move? (move SpaceToDepth to function).

I am also OK to deprecate SpaceToDepth if this OP can be changed to not merge the last two dimension (split them in to sub dimensions).

ONNX op has versioning, it's definitely OK to deprecate one as long as there's new one which can do the job @ebarsoum / @postrational  may comment more on this.

ebarsoum: Based on our discussion, this work will move to 1.8. And we will update the existing SpaceToDepth and DepthToSpace to be more general.

linkerzhang: great! confirm: so the plan is to update SpaceToDepth/DepthToSpace more general and move this one to function?

postrational: Should we change the Milestone to 1.8?

onnx/defs/nn/defs.cc
```diff
@@ -783,6 +877,78 @@ ONNX_OPERATOR_SET_SCHEMA(
     11,
     OpSchema().FillUsing(ConvOpSchemaGenerator("a filter")));
 
+std::function<void(OpSchema&)> UnfoldToDepthOpSchemaGenerator() {
```
linkerzhang: again, this op should be put in the "tensor" folder instead of "nn", it belongs to the same category as "SpaceToDepth", which is used for changing a tensor's layout.

neginraoof:
@linkerzhang Thanks a lot for reviewing the PR.
I have realized that the SpaceToDepth op cannot be directly replaced by UnfoldToDepth operator, and we need to perform a combination of reshape/transpose ops on the output tensor to match the exact behavior. However, we also need the UnfoldToDepth operator since it supports the more general cases. I think the best idea is to change SpaceToDepth op to a function in the next release.
Please let me know what do you think.

onnx/defs/tensor/defs.cc
```diff
@@ -8,6 +8,174 @@
...
+    std::string doc = R"DOC(
+The UnfoldToDepth operator extracts sliding blocks from an input tensor, and concatenates these blocks
+in the last dimension.
+Given an input of shape (N x C x D1 x D2 ... x Dn), output would be a 3-D tensor of shape:<br/>
+
+```(N, C * reduce-mul(block_size), reduce-mul(num_blocks))```
+<br/>
+Where number of blocks extracted from each spatial dimension d is:
+```
+block_size[d] = floor((input_spatial_shape[d] + 2 * padding[d] − dilation[d] * (kernel_size[d] − 1) − 1) / stride[d]) + 1
```
linkerzhang: typo? block_size -> num_blocks?

onnx/defs/tensor/defs.cc
```diff
@@ -8,6 +8,174 @@
...
+        0,
+        "Y",
+        "Output data tensor that contains the result of the "
+        "unfold. The output has three dimensions, and dimension "
+        "values are funciton of the kernel size, stride size, and "
+        "pad lengths.",
+        "T");
+    schema.TypeConstraint(
+        "T",
+        {"tensor(float16)", "tensor(float)", "tensor(double)"},
```
linkerzhang: for image, I think uint8 should be supported, please check whether you want to have other numeric types support.

onnx/defs/tensor/defs.cc
```diff
@@ -8,6 +8,174 @@
...
+        "Constrain input and output types to float tensors.");
+    schema.Attr(
+        "block_shape",
```
linkerzhang: rename to "block_size" is better? as "block_size is used the other places.

onnx/backend/test/case/node/imagetocol.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import absolute_import
```
linkerzhang: you'd rename this file too.

onnx/defs/tensor/defs.cc
```diff
@@ -8,6 +8,174 @@
...
+        "values are funciton of the kernel size, stride size, and "
+        "pad lengths.",
+        "T");
+    schema.TypeConstraint(
+         "T",
+         OpSchema::all_numeric_types(),
+         "Constrain input and output types to numeric tensors.");
+    schema.Attr(
+        "block_size",
+        "The size of the extracted blocks [D1, D2, ..., Dn].",
```
linkerzhang: please clarify the default value for "block_size", given it's optional.

docs/Operators.md
linkerzhang: @postrational I think 1.7 will take this one

PR #2620, opened 2020-02-21
Change dtype of dd_da in gradient test to float32

shinh:
Before:

```
>>> onnx.load_tensor('onnx/onnx/backend/test/data/simple/test_gradient_of_add_and_mul/test_data_set_0/output_1.pb')
data_type: 11
name: "dd_da"
raw_data: "\000\000\000\000\000\000\020@"'
```

After:

```
>>> onnx.load_tensor('onnx/onnx/backend/test/data/simple/test_gradient_of_add_and_mul/test_data_set_0/output_1.pb')
data_type: 1
name: "dd_da"
raw_data: "\000\000\200@"
```


shinh:
@wschin Let me send a notification, thanks!

wschin:
Some mac CI fails because a missing python package `enum`. I have rerun them. If that package is still missing, we need to fix it..

PR #2623, opened 2020-02-24
Generate tests for MeanSquareDistance and SoftmaxCrossEntropyLoss

TMVector:
I noticed some backend tests on master hadn't been generated. Looks like the function-expanded ones for #2570 and #2573, so should probably go in for 1.7.


TMVector:
Maybe a check could be added to CI, similar to how (I believe) it checks docs are up to date

linkerzhang:
exactly, I like the idea on adding a check on the doc update, same as proto and operator docs. @TMVector  are you interested in it? :)

TMVector:
Unfortunately generating these tests seems to expose issues with the definition of MeanSquareDistance and SoftmaxCrossEntropyLoss, plus with function expansion. I'm looking at fixes now.

onnx/backend/test/case/node/__init__.py
```diff
@@ -22,22 +22,28 @@
 from onnx.onnx_operators_pb import FunctionProto
 
 
+# FIXME(TMVector): Any reason we can't get rid of this and use the C++ helper directly?
```
TMVector: TODO/question

gramalingam: The changes look good to me, thanks! I guess we can import the C++ version into Python (provided there is no subtle difference between them due to bugs, etc., I suppose). I guess that could be a different PR, given that we are close to the cutoff date for the next ONNX release.

PR #2625, opened 2020-02-26
Feature/fft ifft

jeremycochoy:
## ⚠️ Work in progress

The fourrier transform operators have been mentionned in https://github.com/onnx/onnx/issues/1646. They are a must have for many signal processing operations, from audio signal to images.

In this PR, I add the specifications for the FFT/iFFT/RFFT/iRFFT operators.

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -2378,4 +2378,76 @@ ONNX_OPERATOR_SET_SCHEMA(
             }
 
         }));
+
+static float fft_signal_ndim_default = 1.;
+
+static const char* FFT_ver12_doc = R"DOC(
+Fast Fourier Trensform.
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    FFT,
+    12,
+    OpSchema()
+        .SetDoc(FFT_ver12_doc)
+        .Attr(
+            "signal_ndim",
+            "The number of dimension of the input signal."
+            "Values can be 1, 2 or 3.",
+            AttributeProto::FLOAT,
+            fft_signal_ndim_default)
+        .Input(0, "scores", "The predicted outputs.", "T")
+        .Input(
+            1,
+            "input",
+            "A complex signal of dimension signal_ndim."
+            "The last dimension of the tensor should be 2,"
+            "representing the real and imaginary components of complex numbers,"
+            "and should have at least signal_ndim + 2 dimensions."
+            "The first dimension is the batch dimension.",
+            "T")
+        .Output(
+            0,
+            "output",
+            "The fourier transform of the input vector,"
+            "using the same format.",
+            "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+            propagateElemTypeFromInputToOutput(ctx, 0, 0);
+            auto signal_ndim = getAttribute(ctx, "signal_ndim", fft_signal_ndim_default);
+        
+            if (!hasInputShape(ctx, 0)) {
+                fail_shape_inference(
+                    "Input tensor should have a shape.")
+            }
+
+            const TensorShapeProto& input_shape =
+                ctx.getInputType(0)->tensor_type().shape();
+            const int rank = static_cast<int>(input_shape.dim_size());
+
+            if (rank != signal_ndim + 2) {
+                fail_shape_inference(
+                    "Rank of input tensor should be signal_ndim + 2 (input rank:",
+                    rank, "signal_ndim + 2:", signal_ndim + 2, ").");
+            }
+
+            auto last_input_dim = input_shape.dim(rank - 1);
+            if (!last_input_dim.has_dim_value()) {
+                fail_shape_inference(
+                    "The last dimension of the input tensor should be 2.");
+            }
+            if (last_input_dim.dim_value() != 2) {
+                fail_shape_inference(
+                    "The last dimension of the input tensor should be 2 instead of ",
+                    last_input_dim.dim_value(), ".");
+            }
+
+            propagateShapeFromInputToOutput(ctx, 0, 0);
+
+        }));
+
 } // namespace ONNX_NAMESPACE
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -2379,7 +2379,7 @@ ONNX_OPERATOR_SET_SCHEMA(
 
         }));
 
-static float fft_signal_ndim_default = 1.;
+static int64_t fft_signal_ndim_default = 1;
 
 static const char* FFT_ver12_doc = R"DOC(
 Fast Fourier Trensform.
@@ -2394,11 +2394,10 @@ ONNX_OPERATOR_SET_SCHEMA(
             "signal_ndim",
             "The number of dimension of the input signal."
             "Values can be 1, 2 or 3.",
-            AttributeProto::FLOAT,
+            AttributeProto::INT,
             fft_signal_ndim_default)
-        .Input(0, "scores", "The predicted outputs.", "T")
         .Input(
-            1,
+            0,
             "input",
             "A complex signal of dimension signal_ndim."
             "The last dimension of the tensor should be 2,"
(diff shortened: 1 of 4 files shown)


jeremycochoy:
@linkerzhang  Excuse me for the interruption, if you have 1 or 2 minutes free could you have a very quick look just to let me know if I am heading in the right direction before I add the 3 remaining Ops ?

7oud:
@jeremycochoy I am also interested in FFT related ops. Expect your commit, I had a look the code commit in your branch, but I did not find the implementation of FFT in C++, maybe it will use cuFFT?

jeremycochoy:
> @jeremycochoy I am also interested in FFT related ops. Expect your commit, I had a look the code commit in your branch, but I did not find the implementation of FFT in C++, maybe it will use cuFFT?

@7oud This question (implementation detail) will (or already have) be addressed by the ONNX backends (onnruntime, tensorflow, pytorch, etc...). This PR will contain only the description of the operators (possible inputs and expected outputs).

7oud:
@jeremycochoy I am not very clear about ONNX. Based on your reply, I think that the ONNX is a architecture of model, which contains only op name and params, but not the implementation of ops. if I wanna run ONNX model, addtional backend(runtime) is needed, just like onnx-runtime, is it right? 
I have a  practical problem, my trained tensorflow model includes fft op, I wanna convert the model to ONNX, then use TensorRT to build the ONNX model to trt engine, and finally run the model in Tensorrt. However the fft op is unsupported in both ONNX and TensorRT now. If wanna achieve this work, what steps are need? Maybe includes
- onnx repo adds fft descriptions, just like what you did
- onnx-tensorrt (Tensorrt backend for onnx) adds fft implementation detail, maybe cuFFT
- tensorflow-onnx, adds parsers to convert tensorflow fft op to onnx fft op
Could you give some advice about it？




running fft/ifft in tensorrt. which are not supported now. 
If tensorrt officially wanna support fft op, what steps should be done, 
Now, I have to run fft/ifft in nvidia tensorrt, which are unsupported now. A possible solution is adding a plugin op of FFT

jeremycochoy:
> @jeremycochoy I am not very clear about ONNX. Based on your reply, I think that the ONNX is a architecture of model, which contains only op name and params, but not the implementation of ops. if I wanna run ONNX model, addtional backend(runtime) is needed, just like onnx-runtime, is it right?
> I have a practical problem, my trained tensorflow model includes fft op, I wanna convert the model to ONNX, then use TensorRT to build the ONNX model to trt engine, and finally run the model in Tensorrt. However the fft op is unsupported in both ONNX and TensorRT now. If wanna achieve this work, what steps are need? Maybe includes
> 
> * **onnx** adds fft descriptions, just like what you did
> * **onnx-tensorrt** (Tensorrt backend for onnx) adds fft implementation detail, maybe cuFFT
> * **tensorflow-onnx**, adds parsers to convert tensorflow fft op to onnx fft op
> 
> Could you give some advice about it？

In your case I would first make sure TensorRT can handle FFT in their model. Once it is the case, you can probably edit the model without FFT to had them by hand to the TensorRT model. This is a short term solution. The long term solution is :
* A ONNX op for fft/ifft.
* Add to onnx-tensorrt the support for writing models with FFT from ONNX graphs which contain them.
* Add convertion from tensorflow to the ONNX op.

Also, maybe the Issue related to the FFT/iFFT/RFFT/iRFFT ops would be better suited for this conversations. 🙂

postrational:
@jeremycochoy We discussed this PR during the Operators SIG meeting today and we would like to move forward with this proposal.

We're wondering how we could expand the description of the operation and tests to make sure we don't run into edge cases which have different behaviors on implementations such as NumPy, PyTorch and Tensorflow.

diyessi:
FFT and iFFT are implementations of DFT/iDFT. I know TensorFlow and PyTorch called it "FFT" but that's not a good reason for ONNX to continue calling it the wrong name.
There are more DFT variants in Torch; there is also the related DCT (discrete cosine transform).

onnx/defs/math/defs.cc
```diff
@@ -2594,4 +2594,157 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T")
+        .TypeConstraint(
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+            propagateElemTypeFromInputToOutput(ctx, 0, 0);
+            const int signal_ndim = getAttribute(ctx, "signal_ndim", fft_signal_ndim_default);
+        
+            if (!hasInputShape(ctx, 0)) {
```
gramalingam, line 2644: Statically unknown shapes are allowed during inference. Inference should not fail in these cases, but can leave the output shapes unspecified.

onnx/defs/math/defs.cc
```diff
@@ -2594,4 +2594,157 @@ ONNX_OPERATOR_SET_SCHEMA(
...
```
gramalingam, line 2720: See comment above. Inference should not fail under this condition.

onnx/defs/math/defs.cc
```diff
@@ -2594,4 +2594,157 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+```
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    FFT,
+    13,
+    OpSchema()
+        .SetDoc(FFT_ver13_doc)
+        .Attr(
+            "signal_ndim",
```
gramalingam, line 2616: Specifying the number-of-dimensions as an attribute is uncommon for ONNX ops. Will this add any value, or can we omit it? It is okay, however, for the spec to restrict the number of dimensions (to 1, 2, or 3) if that is useful. But

onnx/defs/math/defs.cc
```diff
@@ -2594,4 +2594,157 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "The number of dimension of the input signal."
+            "Values can be 1, 2 or 3.",
+            AttributeProto::INT,
+            fft_signal_ndim_default)
+        .Input(
+            0,
+            "input",
+            "A complex signal of dimension signal_ndim."
```
gramalingam, line 2624: ONNX introduced complex-numbers as a type, though no ops yet exist to support them. Seems worthwhile to make a call on this now. We could use complex as a type-specification here. If not, may be we should omit the complex-type from ONNX, if we settle on using the more basic float types.

onnx/defs/math/defs.cc
```diff
@@ -2594,4 +2594,157 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+static int64_t fft_signal_ndim_default = 1;
+
+static const char* FFT_ver13_doc = R"DOC(
+Fast Fourier Transform.
+
+This implementation suppose the DFT (Discrete Fourier Transform) defined by
+```
+A_k =  \sum_{m=0}^{n-1} a_m \exp\left\{-2\pi i{mk \over n}\right\}
+\qquad k = 0,\ldots,n-1.
+```
```
gramalingam, line 2607: May be worth clarifying that `a_m` is the input tensor (indexed) and that `A_k` is the output tensor

onnx/defs/math/defs.cc
```diff
@@ -2594,4 +2594,157 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
+    FFT,
```
gramalingam, line 2611: The numpy version appears to have some extra attributes: output-length adjustment (relative to input-length), and a normalization mode. They also support real/complex inputs. We should also specify how the output is laid out (which index corresponds to positive frequency and which for negative frequency, etc.)

masahi:
Do we want to support real value input FFT? I don't know which of real/complex valued input is more common in practice, especially for deep learning use cases.

j-paulus:
@masahi I can only talk for my needs, but for many (most?) audio applications, a real-valued input to DFT is the main use scenario. Either as a stand-alone or as a part of STFT. If supporting real-valued inputs in addition to complex-valued ones allows optimizing the computational operations, this would be beneficial for the overall efficiency.

onnx/defs/math/defs.cc
xadupre: Supported complex would make it easier to convert models into ONNX. Otherwise, the converter needs to remember which tensor is complex. But having complex means also extend operator Cast to convert the complex tensor into real. But if FFT operator does not produce complex, removing complex everywhere would be less confusing.

onnx/defs/math/defs.cc
xadupre, line 2611: For performance, adding operator RFFT would be great too. That would reduce the computation cost by 2.

pfeatherstone:
Is there a timeline on when this will be supported?

garymm:
Should be closed since it's a subset of https://github.com/onnx/onnx/pull/3741?

napohou:
Hi @jeremycochoy Does your code of https://github.com/jeremycochoy/onnx contain FFT implementation? If so, which branch should we use, master or feature/fft-ifft?

PR #2627, opened 2020-02-26
Update documentation for SoftmaxCrossEntropyLoss

KsenijaS:
Remove the line "Usualy, it's a one-hot representation of ground-truth class." in the documentation that is related to labels, since labels don't have one hot representation.


lara-hdr:
@wschin @liqunfu @linkerzhang @ebarsoum 
Can we merge this asap and include it to the ONNX release to avoid having a bug in the ONNX doc?
thanks

Issue #1730, opened 2019-01-10
Big Endian support in numpy helper


kavanabhat:
Are there any plans to support big endian format in the onnx numpy helper? This is needed for save/retrieval of onnx models on big endian systems. Here is a sample error output when trying to import the onnx backend for caffe2 on AIX.

$python
import onnx
import caffe2.python.onnx.backend
WARNING:root:This caffe2 python run does not have GPU support. Will run in CPU only mode.
Traceback (most recent call last):
 File "<stdin>", line 1, in <module>
 File "/tmp/pytorch_python2_64/build/caffe2/python/onnx/backend.py", line 35, in <module>
   import onnx.numpy_helper
 File "/home/user1/onnx/build/lib.aix-7.2-2.7/onnx/numpy_helper.py", line 16, in <module>
   'Numpy helper for tensor/ndarray is not available on big endian '
RuntimeError: Numpy helper for tensor/ndarray is not available on big endian systems yet.

houseroad:
I think this is an issue in the upstream project, i.e., Numpy. I would suggest to create issue in Numpy repository.

entossdev:
numpy has big-endian support. The above error is being thrown from the onnx numpy_helper.py located at https://github.com/onnx/onnx/blob/master/onnx/numpy_helper.py which is part of this project. So, please reopen the issue and let us know any plans on adding big endian support.

fjfjfan:
@houseroad I have the same problem, could you please help reopen the issue?  Thanks a lot!

kavanabhat:
Model loading and saving works fine on big endian systems with big endian error check removed in https://github.com/onnx/onnx/blob/master/onnx/numpy_helper.py. So, removal of this check allows model load/save on same type of platforms. To support load/save of onnx models across big/little endian systems may need changes in the onnx formats to record the endianness of the data and the processing tools to do the conversion as needed.

shajithchandran:
For those who are interested in inferencing on Big Endian Machine..
I wrote a tool to convert Little Endian ONNX Model to Big Endian Format and vice versa. Using the tools, i was able to run some of the models from ONNX Model Zoo on Big Endian Machine (AIX).
[https://github.com/shajithchandran/onnx-convertor-le-be](url)

PR #2633, opened 2020-03-01
Convert endianness of test dataset to pass tests of onnx-mlir on big-endian machines

imaihal:
Since the endianness of test dataset is little-endian, conversions are required on big-endian machines.  Related issue is also written in https://github.com/onnx/onnx/issues/1730

This PR fixes small part of the issue, and we need to convert more to pass all onnx tests by `pytest`. However, this enables all of [onnx-mlir](https://github.com/onnx/onnx-mlir) test to pass. The test cases are written in [test code](https://github.com/onnx/onnx-mlir/blob/master/test/backend/test.py#L63-L308)

I would like to hear your comments or suggestions about this approach.


onnx/backend/test/runner/__init__.py
```diff
@@ -298,15 +298,21 @@ def run(test_self, device):  # type: (Any, Text) -> None
...
                 ref_outputs = []
                 ref_outputs_num = len(glob.glob(os.path.join(test_data_dir, 'output_*.pb')))
                 for i in range(ref_outputs_num):
                     output_file = os.path.join(test_data_dir, 'output_{}.pb'.format(i))
                     tensor = onnx.TensorProto()
                     with open(output_file, 'rb') as f:
                         tensor.ParseFromString(f.read())
-                    ref_outputs.append(numpy_helper.to_array(tensor))
+                    if sys.byteorder == 'big':
+                        ref_outputs.append(numpy_helper.to_array(tensor).byteswap())
```
postrational: This changes the test runner, but not the underlying helper method.
With this change tests may pass, but the effect may be misleading.

Should the `byteswap()` call be moved into `numpy_helper.to_array` itself?

onnx/numpy_helper.py
```diff
@@ -12,7 +12,7 @@
 from six import text_type, binary_type
 from typing import Sequence, Any, Optional, Text, List
 
-if platform.system() != 'AIX' and sys.byteorder != 'little':
+if (platform.system() != 'AIX' and platform.processor() != 's390x') and sys.byteorder != 'little':
```
postrational: If we did add the `byteswap` workaround to `numpy_helper.to_array`, would we need this increasingly complex condition here?

This condition is growing complex. It would be a good idea to add a small, well named helper function such as `_get_system_byteorder_supported()` and move the complexity there.

snnn:
But the spec says all the data must be little-endian on the disk, you can convert it to BE on-the fly.

postrational:
> But the spec says all the data must be little-endian on the disk.

Would checking for `system_byteorder` in `numpy_helper.to_array` interfere with that?

onnx/backend/test/runner/__init__.py
imaihal: Thanks for the comments! 
At first I thought it might be better to do it, but I was not sure if it is always ok to byteswap the array in `numpy_helper.to_array()` .

imaihal:
I found comments which says `raw_data` of `TensorProto` should be stored in little-endian order.
https://github.com/onnx/onnx/blob/3368834cf0b1f0ab9838cf6bdf78a27299d08187/onnx/onnx.in.proto#L538-L539

On big-endian machines, converting big to little before storing in `TensorProto`, and little to big in `numpy_helper.to_array()` works well?

postrational:
> On big-endian machines, converting big to little before storing in `TensorProto`, and little to big in `numpy_helper.to_array()` works well?

I would say that it's critical that models persisted to disk always maintain one byte order (little-endian). The tools we use to load them into memory may be adaptable, so using `numpy_helper.to_array()` for this seems logical to me.

@linkerzhang What do you think?

imaihal:
OK. I will modify `numpy_helper` to do it. 
I assume `tensor.raw_data` in `numpy_helper.to_array()`  is always little endian.

imaihal:
> But the spec says all the data must be little-endian on the disk.

> models persisted to disk always maintain one byte order (little-endian)

@postrational I couldn't find the description about byte order in the spec. Could you please tell me where they are described?

imaihal:
I updated `numpy_helper.to_array()` and `numpy_helper.from_array()`. This PR assumes `TensorProto.raw_data` in the inputs of `to_array()` and in the outputs of `from_array()` are always little endian even on big-endian machine.

onnx/numpy_helper.py
imaihal: I removed this condition check because we can say "numpy helper is available on big-endian system" by adding endian conversion in numpy helper.

imaihal:
Could @postrational, @linkerzhang, or someone review my PR?

linkerzhang:
@postrational any more comments?

imaihal:
@postrational Do you have any comments?

PR #2640, opened 2020-03-04
Fix copy paste error of Min op test case

take-cheeze:


linkerzhang:
Hmmm, merged too quickly, @take-cheeze the test data should also re-generated (updated).

Issue #2628, opened 2020-02-27
Op spec documentation should not be part of the binary that consumes ONNX.


pranavsharma:
Several parts of the op sec like the main op description, attributes, input and output descriptions become part of the binary that consumes ONNX e.g. onnxruntime causing an increase in its size due to strings that take no part in the execution of the model or its verification. 

Setting ```__ONNX_NO_DOC_STRINGS``` doesn't really help here since (1) it's not used in the SetDoc(string) overload (see https://github.com/onnx/onnx/blob/master/onnx/defs/schema.cc#L444) and (2) it's not used at all for attributes, inputs and outputs. We should do something similar to https://github.com/onnx/onnx/blob/master/onnx/defs/schema.h#L322 to fix these things.

pranavsharma:
I plan to fix this before the next ONNX release.

pranavsharma:
fyi...I'm working on this currently.

PR #2643, opened 2020-03-05
Reduce binary size of libraries consuming ONNX (part 1/2)

pranavsharma:
Several parts of the op sec like the main op description, attributes, input and output descriptions become part of the binary that consumes ONNX e.g. onnxruntime causing an increase in its size due to strings that take no part in the execution of the model or its verification.

Setting __ONNX_NO_DOC_STRINGS doesn't really help here since (1) it's not used in the SetDoc(string) overload (see https://github.com/onnx/onnx/blob/master/onnx/defs/schema.cc#L444) and (2) it's not used at all for attributes, inputs and outputs. We should do something similar to https://github.com/onnx/onnx/blob/master/onnx/defs/schema.h#L322 to fix these things.

This PR takes care of the SetDoc calls and inputs/outputs. I'll send a separate PR for the attributes since they're a bit more involved and this PR is already big. These and the attribute changes together reduce the binary size of onnxruntime by at least 95k (tested using vs2017) with LTO enabled.

Reference: https://github.com/onnx/onnx/issues/2628


onnx/defs/logical/defs.cc
```diff
@@ -168,11 +172,19 @@ Bitwise shift operator performs element-wise operation. For each input element,
...
+  return "";
+#endif
+}
+
 ONNX_OPERATOR_SET_SCHEMA(
     BitShift,
     11,
     OpSchema()
-        .SetDoc(std::string(BitShift_ver11_doc) + GenerateBroadcastingDocMul())
+        .SetDoc(GetBitShiftDoc())
```
pranavsharma: The addition of the output from GenerateBroadcastingDocMul() inhibits the linker from throwing away BitShift_ver11_doc. Hence the introduction of the function GetBitShiftDoc().

onnx/defs/logical/old.cc
```diff
@@ -16,6 +16,7 @@ inline void logicalOpInference_opset1(InferenceContext& ctx) {
 std::function<void(OpSchema&)> BinaryLogicDocGenerator_opset1(
     const char* name) {
   return [=](OpSchema& schema) {
+#ifndef __ONNX_NO_DOC_STRINGS
```
pranavsharma: This kind of setup of the doc doesn't help in removing these strings from the binary. Hence the check for __ONNX_NO_DOC_STRINGS here.

onnx/defs/schema.h
```diff
@@ -329,7 +351,14 @@ class OpSchema final {
     return *this;
   }
 
-  OpSchema& SetDoc(std::string doc);
+  OpSchema& SetDoc(const std::string& doc) {
```
pranavsharma: Changed to const-ref. No need to pay for the copy of doc/description if we're not going to copy it. Ditto for Input() and Output() functions.

pranavsharma:
cc @linkerzhang

pranavsharma:
Can this make it to 1.7 release?

pranavsharma:
Is there any existing issue going on with CircleCI? I see that ci/circleci: py3.6-clang7-ubuntu16.04 is failing for other PRs too that were a day or so old.

linkerzhang:
circleci is not a required ci for onnx, which may be ignored. @spandantiwari for the awareness of the failure, btw.

onnx/defs/math/defs.cc
```diff
@@ -538,13 +554,21 @@ output data (Tensor<T>) where the function `f(x) = slope * x for x < 0`,
 `f(x) = x for x >= 0`., is applied to the data tensor elementwise.
 )DOC";
 
+static std::string GetPReluVer9Doc() {
+#ifndef __ONNX_NO_DOC_STRINGS
```
linkerzhang: suggestion: create a macro for this pattern, so that all later added ones can reuse the pattern and we don't need to revisit this issue any more.

pranavsharma: Currently there are 3 ways SetDoc is accomplished.
* [This](https://github.com/onnx/onnx/blob/master/onnx/defs/generator/defs.cc#L19) is easy to eliminate since it is taken care of in the SetDoc function itself. 
* [This](https://github.com/onnx/onnx/blob/master/onnx/defs/logical/defs.cc#L175) can be hidden behind a generic macro as follows:
```
.SetDoc(GET_OP_DOC_STR(std::string(BitShift_ver11_doc) + GenerateBroadcastingDocMul())

#ifndef __ONNX_NO_DOC_STRINGS
#define GET_OP_DOC_STR(doc_str) (doc_str)
#else
#define GET_OP_DOC_STR(doc_str) ("")
#endif
```
* Finally the third way is [this](https://github.com/onnx/onnx/blob/master/onnx/defs/logical/defs.cc#L28). This is a bit tricky to do in a generic way due to the way doc is constructed. Having the ```ifndef __NO_DOC_STRINGS``` inside SetDoc is not enough in this case. Surrounding SetDoc with this macro also doesn't work since it's too late. Once the 'doc' variable is used, the linker can't throw it away in the generated binary. I don't have a good solution for this one right now. Some more thought might be required.

I can change the 2nd one behind a generic macro.

onnx/defs/logical/defs.cc
```diff
@@ -7,39 +7,43 @@
...
 std::function<void(OpSchema&)> BinaryLogicDocGenerator(const char* name) {
   return [=](OpSchema& schema) {
+#ifndef __ONNX_NO_DOC_STRINGS
     std::string doc = R"DOC(
 Returns the tensor resulted from performing the `{name}` logical operation
 elementwise on the input tensors `A` and `B` (with Numpy-style broadcasting support).
 
 {broadcast_doc}
 )DOC";
-        ReplaceAll(doc, "{name}", name);
```
linkerzhang: how do you think of adding macro on the declaration/definition of this "replaceall" function? so that in the case of "__ONNX_NO_DOC_STRING" defined, there's no such function for developers to call. if it's still called, there will be compilation error reminding that they should not call it.

linkerzhang: I was thinking how to avoid people keep adding more such descriptions and we'll keep doing this job of removing them :)

linkerzhang: maybe, this mechanism should be done on the "SetDoc" API, which is more obvious and stable.

pranavsharma: Yeah, I'm not super happy with this ReplaceAll setup either. We would rather not visit this issue every time :). I'll think of something. The ReplaceAll setup is also not uniform in the way the replacement is done.

pranavsharma: Updated.

pranavsharma:
Any more comments that need to be addressed here?

gramalingam:
Unfortunately, there appears to be some conflicts to be resolved.

pranavsharma:
Can this be merged once CI is complete?

snnn:
This change broke raspberrypi build.

snnn:
This change also break the build on ubuntu 14.04.

PR #2644, opened 2020-03-06
Regenerate test data of Min-12

take-cheeze:
Related too https://github.com/onnx/onnx/pull/2640


TMVector:
I think only the files related to the operators you've updated should be changed -- it seems to have been the policy to keep old files (the only change for everything else is (should be) opset/IR version numbers).

Thanks for fixing my mistake btw :slightly_smiling_face:

PR #2641, opened 2020-03-04
Fix Greater/LessOrEqual function definition

take-cheeze:

PR #2645, opened 2020-03-06
Fix Greater/LessOrEqual function definition

take-cheeze:
CLA fix for: https://github.com/onnx/onnx/pull/2641
closes #2641


linkerzhang:
@jeremycochoy can you also help to fix the test data of these two functions please? I believe the test data checked in was also wrong because of the bug fixed by this PR.

linkerzhang:
@take-cheeze you may want to update the test data accordingly if @jeremycochoy can't offer help in time.

The function test data's generation is,
when adding a test case for a function op, there will be two sets of test data generated, with same inputs/outputs, but different models, one model is with the function op as it is in the model, the other model is having the function body graph inlineed. This is also ensuring the function body graph is defined correctly.

take-cheeze:
Sorry to forget the test data update. 
https://github.com/onnx/onnx/pull/2645/commits/7b4897daecb7f6cc9b139745078a1ef04293865d should fix it

jeremycochoy:
Seams like I read this a bit too late. 😅

PR #2605, opened 2020-02-15
Fixing ONNX Debug mode on Windows, adding ONNX Debug CI for Windows

vinitra:
To resolve #1880, install onnx from source in CI w/ Debug mode for Windows. Linux / Mac builds are in #2651.

Issue #1880, opened 2019-03-21
At least one CI build and install ONNX in debug mode


houseroad:
Otherwise, some check may fail silently. Such as https://github.com/onnx/onnx/blob/master/onnx/defs/schema.cc#L820

https://github.com/onnx/onnx/pull/1878 is for fixing the checks. But we need CI to detect such breakage.

houseroad:
cc: @linkerzhang @prasanthpul @zrphercule @bddppq

snnn:
Then don't enable 'COVERAGE' . I thought it was by design.

PR #2651, opened 2020-03-10
Adding CI for ONNX Debug mode (Linux, OSX)

vinitra:
Adding a Linux and OSX build for ONNX Debug mode, as per #1880. Including ONNX_ML enabled CI as well. Addressing ONNX Debug CI and errors for windows in #2605.


vinitra:
@linkerzhang, can you take a look?

gramalingam:
LGTM, thanks! I have one question though: how do we decide which combinations are worth adding CI for? There isn't a ONNX_DEBUG version of all combinations. It makes sense, as doubling the number of CIs would probably be too much, but wondering.

vinitra:
> LGTM, thanks! I have one question though: how do we decide which combinations are worth adding CI for? There isn't a ONNX_DEBUG version of all combinations. It makes sense, as doubling the number of CIs would probably be too much, but wondering.

We chose one version with ONNX ML enabled, one version with it disabled. We also chose 2 OS types to vary the format and catch more errors (Linux / Mac). We were thinking of also including a Windows build, but found some errors with the ONNX Debug package on Windows that need to be resolved. These are being addressed in a separate PR: https://github.com/onnx/onnx/pull/2605

Thanks for the approval!

Issue #2414, opened 2019-10-23
Please require op version adapter in future breaking PRs


matteosal:
The version converter is currently missing several adapters: https://github.com/onnx/onnx/blob/master/onnx/version_converter.py#L64

Add this lack of adapters to the fast pace at which new opset versions are released, and to the strict rules that demand bumping an operator's version, and you obtain that external implementations are forced to support several opset versions explicitly in order to have a truly solid ONNX import/export functionality.

The current situation with Conv is particularly interesting: the opset version was switched to 11 in [this](https://github.com/onnx/onnx/pull/2291) PR for a change of documentation, but no adapter was added. This means that frameworks supporting opset 11 only will formally have to reject any old convolution operator, but importing them ignoring the version bump will actually work, because the spec didn't actually change.

Supporting multiple opset versions is additional burden placed on external implementations that can limit the potential of ONNX to allow seamless switchting between frameworks. On the other hand, if the version converter had complete coverage, external frameworks would only need to support the latest opset when it's released, and switching between frameworks will be guaranteed to work.

I am a developer of Mathematica's neural network framework at Wolfram Research, and we're currently working on our ONNX importer. We are very interested in having the adapter expanded, and are considering adding missing adapters for this reason.
Covering the 90+ adapters which are currently missing is a big endeavor which would take a long time, but at least the number could be kept from growing by requiring that every PR which bumps the op version also adds the related adapter.

Is this a reasonable request? Thanks!

prasanthpul:
There have been discussions about requiring a converter to be added when ops are updated. @onnx/sig-operators to comment. @matteosal suggest you join the next operators SIG meeting as well

rkpandya:
I was trying to convert from opset 11 to 9, is that currently possible? I get the error :

matteosal:
@rkpandya the adapter for that conversion is not implemented (although the internal spec is identical in both versions). 
I have added several adapters in [this PR](https://github.com/onnx/onnx/pull/2664/files) including that one, so your conversion should work there (but it might fail due to other missing adapters)

rkpandya:
Thanks! The main reason I had to use opset 11 was the aten operators, specifically bilinear interpolation.

rkpandya:
Anything else I can try here to make this work?

askhade:
@matteosal : 
[...]

PR #2664, opened 2020-03-17
More version adapters

matteosal:
This adds several missing version adapters and sorts registered adapters by version in the code (I have found it easier to figure out what's missing). I have added adapters for the operators that we (Wolfram Research) are going to support for Mathematica 12.1. 

In [this](https://github.com/onnx/onnx/issues/2414) issue I've described the problem of missing adapters for external implementations and asked if it was possible to require adapters to be always added when specs are changed. I'll take this opportunity to also renew this request, hoping that developers here will agree that this is an important point to ensure that the format has good coverage across frameworks.

This is my first PR in ONNX, so there are some details I'm actually not sure about, and I sorry in advance for any wrong practice :). More comments in the diff.


onnx/version_converter/BaseConverter.h
```diff
@@ -54,10 +54,10 @@ class BaseVersionConverter {
           if (adapter_ptr != target_map->second.end()) {
             return *(adapter_ptr->second);
           } else {
-            ONNX_ASSERTM(false, "No Adapter To Version %s for %s", initial.c_str(), op_name.c_str());
+            ONNX_ASSERTM(false, "No Adapter To Version %s for %s", target.c_str(), op_name.c_str());
           }
```
matteosal: Looks like the intended version number was `target` here. 

In the change below, I found it clearer to reword the message in that way.

onnx/version_converter/adapters/broadcast_forward_compatibility.h
```diff
@@ -22,14 +22,23 @@ class BroadcastForwardCompatibility final : public Adapter {
...
+            // Add dimensions to the left
+            for (int i = 0; i < axis; i++) {
+              axes.emplace_back(i);
+              new_sizes.emplace_back(Dimension(1));              
+            }
+            new_sizes.insert(new_sizes.end(), B_sizes.begin(), B_sizes.end());
+            // Add dimensions to the right
+            for (int i = 0; i < (int) (A_sizes.size() - B_sizes.size() - axis); i++) {
+              axes.emplace_back(B_sizes.size() + axis + i);
               new_sizes.emplace_back(Dimension(1));
```
matteosal: I went looking for the old broadcast behaviour [here](https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Add-6), which states that dimensions could also be added to the left of the shape list. But the previous adapter was only adding them to the right.

onnx/version_converter/convert.h
```diff
@@ -105,120 +107,341 @@ class DefaultVersionConverter : public BaseVersionConverter {
...
+      registerAdapter(make_unique<RemoveConsumedInputs>("Sqrt",
+        OpSetID(5), OpSetID(6)));
+      registerAdapter(make_unique<RemoveConsumedInputs>("Sub",
+        OpSetID(5), OpSetID(6)));
+      registerAdapter(make_unique<RemoveConsumedInputs>("Tanh",
+        OpSetID(5), OpSetID(6)));
+
+      /******** 6 -> 5 ********/
+      registerAdapter(make_unique<TypeRestriction>("Add",
+        OpSetID(6), OpSetID(5), broadcast_unallowed_types));
```
matteosal: I'm not sure the converters for `6 -> 5` below this comment are correct. I followed what was done for `"Add"`.

onnx/version_converter/convert.h
```diff
@@ -105,120 +107,341 @@ class DefaultVersionConverter : public BaseVersionConverter {
...
-        OpSetID(8), OpSetID(7)));
       registerAdapter(make_unique<CompatibleAdapter>("Mean",
         OpSetID(7), OpSetID(8)));
+      registerAdapter(make_unique<CompatibleAdapter>("Sum",
+        OpSetID(7), OpSetID(8)));
+      registerAdapter(make_unique<CompatibleAdapter>("MaxPool",
+        OpSetID(7), OpSetID(8)));
+
+      /******** 8 -> 7 ********/  
+      /* Should the adapters for Max, Min, Mean be CompatiblAdapter here? */
```
matteosal: The three adapters for `"Max"`, `"Min"` and `"Mean"` here were pre-existing as `BroadcastBackwardCompatibility`, but it really looks they should be `CompatibleAdapter`. What is the right adapter?

matteosal:
I have tested these changes using the adapter test suite and extra upgrade tests I wrote for myself.

My extra tests consisted in creating model files for all the operators involved, creating them at the lowest possible opset version (attaching the script I used to create them). I then ran the version converter on each of them converting to version 12, to ensure that the full stack of adapters was available for all the operators. I have focused on version upgrades because I felt they are much more critical, but a similar thing can be done with version downgrades. 

Are you interested in adding something like this to the test suite?

[gen_upgrade_test_data.zip](https://github.com/onnx/onnx/files/4344602/gen_upgrade_test_data.zip)

matteosal:
Any feedback on this?

vinitra:
Thanks for the contribution, @matteosal! cc: @linkerzhang, @KsenijaS to take a look

matteosal:
Another ping on this :)

matteosal:
@vinitra @linkerzhang @KsenijaS any news on this?

onnx/version_converter/convert.h
jcwchen: According to https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Mean-8,
(same as Max-8 and Min-8), they started to support broadcasting from opset8. 
In that case, using `BroadcastBackwardCompatibility` here makes more sense to me.

matteosal: Indeed. Removing the comment then

onnx/version_converter/BaseConverter.h
jcwchen: Good catch

onnx/version_converter/adapters/broadcast_forward_compatibility.h
jcwchen: I guess the reason is the old broadcast behaviour (opset 6) only supports limited broadcast (only adding dimensions to the right one). 

If the op is upgraded to opset 7, it will allow Multi-directional Broadcasting. Therefore, the version_converter needs to do Unidirectional Broadcast (only adding dimensions to the right one) for opset 6 first. Otherwise, opset 7 would misunderstand how to broadcast these old ops. 

IMO, adding dimensions to the left here looks unnecessary to me.

onnx/version_converter/convert.h
jcwchen: `Add` and `Mul` started to support `int32`, `int64`, `uint32`, `uint64` since opset 6 because of supporting broadcast.
I think most of other ops like `Sum` and `Ceil`, which do not support broadcast in opset 6, do not need to use `TypeRestriction` here.

onnx/version_converter/adapters/broadcast_forward_compatibility.h
matteosal: By "left" and "right" I don't mean the left and right inputs of the node, but left and right of the dimension list of the right (second) input, i.e. prepending (left) and appending (right) dimensions. What I added is still adding dimensions to B only, but allows adding them on both sides of the dimension list. Previously, this adapter was only adding them to the right (i.e. appending them, via `axes.emplace_back(B_sizes.size() + i)`).

The examples in the changelog I linked are both prepending and appending dimensions:
```
shape(A) = (2, 3, 4, 5), shape(B) = (,), i.e. B is a scalar tensor
shape(A) = (2, 3, 4, 5), shape(B) = (1, 1), i.e. B is an 1-element tensor
shape(A) = (2, 3, 4, 5), shape(B) = (5,)
shape(A) = (2, 3, 4, 5), shape(B) = (4, 5)
shape(A) = (2, 3, 4, 5), shape(B) = (3, 4), with axis=1
shape(A) = (2, 3, 4, 5), shape(B) = (2), with axis=0
```

matteosal: Ah I realize that the reason why dimensions were not prepended might be that opset 7 (and higher versions) allow prepending dimensions themselves, so there's no reason to explicitly prepend them in the adapter. Is that correct? I will revert this change if yes.

onnx/version_converter/convert.h
matteosal: Ok I see the point now. I went through the changelog checking which operator extended their types in opset 6, and modified the adapters accordingly

onnx/version_converter/adapters/broadcast_forward_compatibility.h
jcwchen: Sorry for my previous misunderstanding.
But I think the answer is yes. The key point is for broadcast in opset 7, it does not need an axis to append anymore.
Besides, as you mentioned, it can handle prepending dimensions. 
So the only thing we need to do here for opset 6 to 7 is appending the dimensions by the axis.

jcwchen:
The adapter of `Clip` and `Dropout` looks good to me.
@matteosal Thank you for the hard work!

onnx/version_converter/convert.h
```diff
@@ -102,123 +104,303 @@ class DefaultVersionConverter : public BaseVersionConverter {
...
+        OpSetID(8), OpSetID(7)));
       registerAdapter(make_unique<BroadcastBackwardCompatibility>("Mean",
         OpSetID(8), OpSetID(7)));      
-      registerAdapter(make_unique<BroadcastForwardCompatibility>("Mul",
-        OpSetID(6), OpSetID(7)));
-      registerAdapter(make_unique<TypeRestriction>("Mul",
-        OpSetID(6), OpSetID(5), broadcast_unallowed_types));
-      registerAdapter(make_unique<RemoveConsumedInputs>("Mul",
-        OpSetID(5), OpSetID(6)));
-      registerAdapter(make_unique<CompatibleAdapter>("Gemm",
```
jcwchen, line 136: registerAdapter(make_unique<CompatibleAdapter>("Gemm",
        OpSetID(6), OpSetID(5)));
registerAdapter(make_unique<CompatibleAdapter>("Relu",
        OpSetID(6), OpSetID(5)));
registerAdapter(make_unique<CompatibleAdapter>("Sum",
        OpSetID(6), OpSetID(5)));
registerAdapter(make_unique<CompatibleAdapter>("Dropout",
        OpSetID(6), OpSetID(5)));

onnx/version_converter/convert.h
```diff
@@ -102,123 +104,303 @@ class DefaultVersionConverter : public BaseVersionConverter {
...
+      registerAdapter(make_unique<TypeRestriction>("Sub",
+        OpSetID(6), OpSetID(5), broadcast_unallowed_types));
+      registerAdapter(make_unique<TypeRestriction>("Div",
+        OpSetID(6), OpSetID(5), broadcast_unallowed_types));
+      registerAdapter(make_unique<TypeRestriction>("Abs",
+        OpSetID(6), OpSetID(5), int_unallowed_types));
+      registerAdapter(make_unique<TypeRestriction>("Neg",
+        OpSetID(6), OpSetID(5), neg_unallowed_types));
+      registerAdapter(make_unique<BatchNormalization_6_5>());
+
```
jcwchen, line 211:
```suggestion
registerAdapter(make_unique<CompatibleAdapter>("Gemm",
        OpSetID(6), OpSetID(5)));
registerAdapter(make_unique<CompatibleAdapter>("Relu",
        OpSetID(6), OpSetID(5)));
registerAdapter(make_unique<CompatibleAdapter>("Sum",
        OpSetID(6), OpSetID(5)));
registerAdapter(make_unique<CompatibleAdapter>("Dropout",
        OpSetID(6), OpSetID(5)));
```

onnx/version_converter/adapters/broadcast_forward_compatibility.h
matteosal: Reverted

matteosal:
Is there something missing still? What about the tests I suggested [here](https://github.com/onnx/onnx/pull/2664#issuecomment-600175746)?

matteosal:
@jcwchen @askhade  a ping on this :)
What about the tests I've suggested [here](https://github.com/onnx/onnx/pull/2664#issuecomment-600175746)?

jcwchen:
Hi @matteosal Sorry for the late reply.
I checked your test file. Is it something like encapsulating the helper function?
IMO, to test the version_converter, maybe we can simply follow the original test [here](https://github.com/onnx/onnx/blob/master/onnx/test/version_converter_test.py).
It seems unnecessary to save the created model.

Thanks.

matteosal:
> Hi @matteosal Sorry for the late reply.
> I checked your test file. Is it something like encapsulating the helper function?
> IMO, to test the version_converter, maybe we can simply follow the original test [here](https://github.com/onnx/onnx/blob/master/onnx/test/version_converter_test.py).
> It seems unnecessary to save the created model.
> 
> Thanks.

matteosal:
> Is it something like encapsulating the helper function?

Sorry I don't undarstand this

> IMO, to test the version_converter, maybe we can simply follow the original test [here](https://github.com/onnx/onnx/blob/master/onnx/test/version_converter_test.py).
> It seems unnecessary to save the created model.
> 
> Thanks.

Ok fine, saving the model is not necessary, but I think the testing philosophy can be improved by automatically picking the highest possible opset version for upgrades (and the lowest possible for downgrades). 

The current tests are designed to test conversions between a fixed paier of opset versions (e.g, `Sum Adapter: 5 -> 8` and `Sum Adapter: 8 -> 5`). This forces developers to continously add tests as operators are changed. On the other hand, if all tests used the current opset version as the target version the number additional tests would decrease (at least for upgrading). E.g. `Sum 5 -> 8`, is only testing a single adapter, but if the current opset version is picked instead of 8 (so currently v13), ALL the adapters for `Sum` will be tested automatically without having to add new. 

This mechanism is not perfect because it doesn't guarantee that all the upgared codepaths will be tested, so one might still have to add new tests once in a while, but I think it's still an improvement

jcwchen:
> Sorry I don't understand this

According to the file you provided, I thought the file is only for simplifying the process of graph creation.

> This mechanism is not perfect because it doesn't guarantee that all the upgraded code paths will be tested, so one might still have to add new tests once in a while, but I think it's still an improvement

Thank you for the elaboration. I think I finally get this.
The idea looks good to me. Do you have any other existing test code which can demo your idea?
It would help us to understand your proposal. Thank you!

onnx/version_converter/adapters/clip_10_11.h
```diff
@@ -0,0 +1,64 @@
...
+      if(has_min) {
+        addMinInit(graph, node);
+      }
+      if(has_max) {
+        if(!has_min) {addMinDefault(graph, node);}
+        addMaxInit(graph, node);
+      }
+    }
+
+    void addMinInit(std::shared_ptr<Graph> graph, Node *node) const {
```
askhade, line 27: Do you need 3 methods? You can simply create 1 generic method which takes a few extra params for the values of min, max, default no?

onnx/version_converter/adapters/clip_10_11.h
```diff
@@ -0,0 +1,64 @@
+// Adapter for Clip in default domain from version 10 to 11
+
+#pragma once
+#include <limits>
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+class Clip_10_11 final : public Adapter {
```
askhade, line 8: Why is this adapter needs? After opset 6 clip was only changed in opset 11 so just having a 6 -> 11 adapter should be sufficient right? Is this not the case today?

onnx/version_converter/convert.h
```diff
@@ -102,123 +104,368 @@ class DefaultVersionConverter : public BaseVersionConverter {
...
+        OpSetID(8), OpSetID(9)));
+      registerAdapter(make_unique<BatchNormalization_8_9>());
+      registerAdapter(make_unique<Scan_8_9>());
+      registerAdapter(make_unique<Upsample_8_9>());
+
+      /******** 9 -> 8 ********/
+      registerAdapter(make_unique<CompatibleAdapter>("BatchNormalization",
+        OpSetID(9), OpSetID(8)));  
+      registerAdapter(make_unique<ExtendSupportedTypes>("Flatten",
+        OpSetID(9), OpSetID(8)));
```
askhade, line 305: why are these adapters needed? There is no opset 8 version for Flatten

onnx/version_converter/adapters/clip_10_11.h
matteosal: Adapters are always converting between contiguous opset versions. The version converter upgrades operators making jumps of one version point at a time, and if the operatos has changed for a particular jump it goes look for an adapter. So the way Clip is upgraded from 6 to 11 is 6 -> 7 -> 8 -> 9 -> 10 -> 11, where only the last step requires an adapter.

This also answers your comment below.

onnx/version_converter/convert.h
matteosal: There is no Flatten-8, but Flatten exists in opset version 8 anyway, as Flatten-1. The adapter 9 -> 8 is needed to downgrade Flatten from opset version 9 to any version before that.

askhade:
> > Is it something like encapsulating the helper function?
> 
> Sorry I don't undarstand this
> 
> > IMO, to test the version_converter, maybe we can simply follow the original test [here](https://github.com/onnx/onnx/blob/master/onnx/test/version_converter_test.py).
> > It seems unnecessary to save the created model.
> > Thanks.
> 
> Ok fine, saving the model is not necessary, but I think the testing philosophy can be improved by automatically picking the highest possible opset version for upgrades (and the lowest possible for downgrades).
> 
> The current tests are designed to test conversions between a fixed paier of opset versions (e.g, `Sum Adapter: 5 -> 8` and `Sum Adapter: 8 -> 5`). This forces developers to continously add tests as operators are changed. On the other hand, if all tests used the current opset version as the target version the number additional tests would decrease (at least for upgrading). E.g. `Sum 5 -> 8`, is only testing a single adapter, but if the current opset version is picked instead of 8 (so currently v13), ALL the adapters for `Sum` will be tested automatically without having to add new.
> 
> This mechanism is not perfect because it doesn't guarantee that all the upgared codepaths will be tested, so one might still have to add new tests once in a while, but I think it's still an improvement


Agree with the overall idea... Can you create a separate PR for this. Thanks!

matteosal:
> > > Is it something like encapsulating the helper function?
> > 
> > 
> > Sorry I don't undarstand this
> > > IMO, to test the version_converter, maybe we can simply follow the original test [here](https://github.com/onnx/onnx/blob/master/onnx/test/version_converter_test.py).
> > > It seems unnecessary to save the created model.
> > > Thanks.
> > 
> > 
> > Ok fine, saving the model is not necessary, but I think the testing philosophy can be improved by automatically picking the highest possible opset version for upgrades (and the lowest possible for downgrades).
> > The current tests are designed to test conversions between a fixed paier of opset versions (e.g, `Sum Adapter: 5 -> 8` and `Sum Adapter: 8 -> 5`). This forces developers to continously add tests as operators are changed. On the other hand, if all tests used the current opset version as the target version the number additional tests would decrease (at least for upgrading). E.g. `Sum 5 -> 8`, is only testing a single adapter, but if the current opset version is picked instead of 8 (so currently v13), ALL the adapters for `Sum` will be tested automatically without having to add new.
> > This mechanism is not perfect because it doesn't guarantee that all the upgared codepaths will be tested, so one might still have to add new tests once in a while, but I think it's still an improvement
> 
> Agree with the overall idea... Can you create a separate PR for this. Thanks!

Ok I will, although not in the near future

onnx/version_converter/adapters/clip_10_11.h
matteosal: Of course! I made this when I was starting getting familiar with the system and wanted to keep things very simple. Changed it to use the generic method.

askhade:
> > > > Is it something like encapsulating the helper function?
> > > 
> > > 
> > > Sorry I don't undarstand this
> > > > IMO, to test the version_converter, maybe we can simply follow the original test [here](https://github.com/onnx/onnx/blob/master/onnx/test/version_converter_test.py).
> > > > It seems unnecessary to save the created model.
> > > > Thanks.
> > > 
> > > 
> > > Ok fine, saving the model is not necessary, but I think the testing philosophy can be improved by automatically picking the highest possible opset version for upgrades (and the lowest possible for downgrades).
> > > The current tests are designed to test conversions between a fixed paier of opset versions (e.g, `Sum Adapter: 5 -> 8` and `Sum Adapter: 8 -> 5`). This forces developers to continously add tests as operators are changed. On the other hand, if all tests used the current opset version as the target version the number additional tests would decrease (at least for upgrading). E.g. `Sum 5 -> 8`, is only testing a single adapter, but if the current opset version is picked instead of 8 (so currently v13), ALL the adapters for `Sum` will be tested automatically without having to add new.
> > > This mechanism is not perfect because it doesn't guarantee that all the upgared codepaths will be tested, so one might still have to add new tests once in a while, but I think it's still an improvement
> > 
> > 
> > Agree with the overall idea... Can you create a separate PR for this. Thanks!
> 
> Ok I will, although not in the near future

Can you create an issue with these details to track this work. This way in future someone else who has bandwidth can also pick this workitem. Thanks!

onnx/version_converter/adapters/argmax_argmin_12_11.h
```diff
@@ -0,0 +1,26 @@
...
+class ArgMaxArgMin_12_11 final : public Adapter {
+  public:
+    explicit ArgMaxArgMin_12_11(const std::string& op_name)
+      : Adapter(op_name, OpSetID(12), OpSetID(11)) {}
+
+    void adapt_argmax_argmin_12_11(std::shared_ptr<Graph> graph, Node* node) const {
+    	Symbol select_last_index = Symbol("select_last_index");
+	    if (node->hasAttribute(select_last_index)) {
+	      ONNX_ASSERTM(node->i(select_last_index) == 0,
+	          "select_last_index != 0 is not supported in version 11");
```
askhade, line 16: nit: Change error message to something like : Opset version 11 only supports select_last_index == 0?

askhade:
LGTM. Once the test failures are fixed I will approve the PR.
Thanks!

matteosal:
> This mechanism is not perfect because it doesn't guarantee that all the upgared codepaths will be tested, so one might still have to add new tests once in a while, but I think it's still an improvement

Done https://github.com/onnx/onnx/issues/2931

matteosal:
@askhade some CI test is failing reporting errors about some ipynb file:
```
==================================== ERRORS ====================================
________________________ ERROR collecting test session _________________________
../../../.conda/envs/py3.6/lib/python3.6/site-packages/pluggy/hooks.py:286: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
../../../.conda/envs/py3.6/lib/python3.6/site-packages/pluggy/manager.py:93: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
../../../.conda/envs/py3.6/lib/python3.6/site-packages/pluggy/manager.py:87: in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
../../../.conda/envs/py3.6/lib/python3.6/site-packages/nbval/plugin.py:115: in pytest_collect_file
    return IPyNbFile(path, parent)
../../../.conda/envs/py3.6/lib/python3.6/site-packages/_pytest/nodes.py:95: in __call__
    warnings.warn(NODE_USE_FROM_PARENT.format(name=self.__name__), stacklevel=2)
E   pytest.PytestDeprecationWarning: Direct construction of IPyNbFile has been deprecated, please use IPyNbFile.from_parent.
E   See https://docs.pytest.org/en/stable/deprecations.html#node-construction-changed-to-node-from-parent for more details.
=========================== short test summary info ============================
ERROR  - pytest.PytestDeprecationWarning: Direct construction of IPyNbFile ha...
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 0.16s ===============================
```

I am not familiar with CI internals, but this really seems unrelated to my changes. I don't know what action should be taken.

jcwchen:
@matteosal
It only fails with python3 and you did not change anything on the python side. 
I think the ipynb failure should be not related to your PR. Sorry for the confusion.

onnx/version_converter/adapters/argmax_argmin_12_11.h
```diff
@@ -0,0 +1,26 @@
...
+#pragma once
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+class ArgMaxArgMin_12_11 final : public Adapter {
+  public:
+    explicit ArgMaxArgMin_12_11(const std::string& op_name)
+      : Adapter(op_name, OpSetID(12), OpSetID(11)) {}
+
+    void adapt_argmax_argmin_12_11(std::shared_ptr<Graph> graph, Node* node) const {
```
matteosal, line 12: @askhade one job is still failing because `graph` is unused here, which triggers a warning treated as error: https://travis-ci.org/github/onnx/onnx/jobs/713539738#L2049
The problem is, if I remove the `graph` argument here I get the same exact warning for `adapt(std::shared_ptr<Graph> graph, Node* node)` a few lines below, and that can't be changed because that's how adapters are supposed to be called by the converter: https://github.com/onnx/onnx/blob/master/onnx/version_converter/convert.cc#L98

So I don't know how to solve this.

onnx/version_converter/adapters/argmax_argmin_12_11.h
matteosal: Done

onnx/version_converter/adapters/argmax_argmin_12_11.h
askhade: you can simply change this to : 
void adapt_argmax_argmin_12_11(std::shared_ptr<Graph> , Node* node) const {

matteosal: Ah of course, unnamed arguments! Done

matteosal:
@askhade can this be merged now?

PR #2666, opened 2020-03-17
Update Pow input types in Opset 12

lara-hdr:
Update Pow in opset 12 to allow operands of different types as inputs, and allow int types inputs.


onnx/defs/math/defs.cc
```diff
@@ -505,24 +505,48 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .SetDoc(std::string(Pow_ver12_doc) + GenerateBroadcastingDocMul())
         .Input(0, "X", "First operand, base of the exponent.", "T")
-        .Input(1, "Y", "Second operand, power of the exponent.", "T")
+        .Input(1, "Y", "Second operand, power of the exponent.", "T1")
         .Output(0, "Z", "Output tensor (same size as X)", "T")
         .TypeConstraint(
             "T",
-            {"tensor(float16)", "tensor(float)", "tensor(double)"},
-            "Constrain input and output types to float tensors.")
+            {"tensor(uint8)",
```
linkerzhang: I don't think it's a good idea to have 8-bit integer support added as this, the output will overflow if it has same type as input.

What's the motivation of adding integer support right now? can adding int32/int64 support be good enough? if yes, don't add 8/16 bit integers for now.

lara-hdr: that's a good point, thanks.
this is needed for the function ONNX::MeanSquaredDistance that executes a Pow operation with X as a float and Y as an int.
I added ints for X as well for completeness since it is a valid operation that we would need eventually. I removed 8-16 bits types for T (X and output).

snnn:
Seems not ok.
I checked your branch and run a local test,  I got even more failures. 

test_pow_types_float_cpu says: 
"onnx.onnx_cpp2py_export.checker.ValidationError: Y typestr: T, has unsupported type: tensor(int32)"

lara-hdr:
@snnn I haven't tested with ORT yet. I am having some build issues when building with a new ONNX version. Is that the complete error you are seeing?

snnn:
So, this bug fix is for test_mean_square_distance_mean_3d_expanded and some others, not for test_softmax_cross_entropy_mean_3d, right?

lara-hdr:
@snnn yes, this is the link for softmax cross entropy: https://github.com/microsoft/onnxruntime/pull/3237.
@codemzs mentioned still seeing a "Could not find an implementation for the node Max(12)" error after the fix.

snnn:
> @snnn yes, this is the link for softmax cross entropy: microsoft/onnxruntime#3237.
> @codemzs mentioned still seeing a "Could not find an implementation for the node Max(12)" error after the fix.

That's normal. Because onnxruntime doesn't have an implementation for the OP yet, that's expected.

onnx/defs/math/defs.cc
```diff
@@ -505,24 +505,42 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .SetDoc(std::string(Pow_ver12_doc) + GenerateBroadcastingDocMul())
         .Input(0, "X", "First operand, base of the exponent.", "T")
-        .Input(1, "Y", "Second operand, power of the exponent.", "T")
+        .Input(1, "Y", "Second operand, power of the exponent.", "T1")
         .Output(0, "Z", "Output tensor (same size as X)", "T")
         .TypeConstraint(
             "T",
-            {"tensor(float16)", "tensor(float)", "tensor(double)"},
-            "Constrain input and output types to float tensors.")
+            {"tensor(int32)",
```
spandantiwari: The set of allowed types T is more limited than T1 for exponent. Is there a strong reason not to support unsigned integers?

lara-hdr: I restricted the types of T following this comment https://github.com/onnx/onnx/pull/2666#discussion_r394071964 to avoid overflow.
Another option would be to accept any type for X as well. 
But then the output type would be impossible to infer..

lara-hdr:
@ebarsoum @wschin for review

docs/Operators.md
```diff
@@ -11996,13 +11998,13 @@ node = onnx.helper.make_node(
 
 x = np.array([1, 2, 3]).astype(np.float32)
 y = np.array([4, 5, 6]).astype(np.float32)
-z = np.power(x, y)  # expected output [1., 32., 729.]
+z = pow(x, y)  # expected output [1., 32., 729.]
```
wschin: Please add tests for newly supported types.

lara-hdr: there are 2 tests in export_types() involving int64.
Would adding 2 others for int32 be sufficient?

wschin: Your new tests are good enough. Thanks.

PR #2668, opened 2020-03-18
Run stricter model checker in onnx backend test

snnn:
For example, it can detect the shape inference errors that was just found in SoftmaxCrossEntropyLoss and mean_square_distance_mean


onnx/backend/base.py
```diff
@@ -71,7 +71,7 @@ def prepare(cls,
                 **kwargs  # type: Any
                 ):  # type: (...) -> Optional[BackendRep]
         # TODO Remove Optional from return type
-        onnx.checker.check_model(model)
```
snnn: This is the major change.

snnn:
@bddppq Could you please help take a look at the test_scalar_type failures in Caffe2Backend
 tests? I think the test case itself is wrong.

bddppq:
cc @spandantiwari @houseroad 
I believe the test case is intentionally testing comparison between float and int (as PyTorch supports it). The problem is probably that the onnx exporter does not do type promotion during conversion.

```python
class ComparisonModel(torch.nn.Module):
    def forward(self, x, y):
        return x.ge(0.5) & y.le(2)
    
x = torch.ones(2, 3, dtype=torch.int32)
y = torch.ones(2, 3, dtype=torch.float32)
self.run_model_test(ComparisonModel(), input=(x, y), train=False, batch_size=BATCH_SIZE)
...
>       inferred_model_str = C.infer_shapes(model_str, check_type)
E       onnx.onnx_cpp2py_export.checker.ValidationError: A typestr: T, has unsupported type: tensor(int32)```

PR #2669, opened 2020-03-19
Fix testcase of batchnorm.py

shinh:
The reduction axes for saved_mean and saved_var were wrong.


shinh:
CC: @lara-hdr

shinh:
Closing since https://github.com/onnx/onnx/pull/2674 fixes one more issue

PR #2675, opened 2020-03-20
Update sequence test case so input is not scalar and splits are specified

skottmckay:


onnx/backend/test/case/model/sequence.py
```diff
@@ -330,14 +330,15 @@ def make_graph(
...
         seq_len_node = onnx.helper.make_node('SequenceLength', ['seq_1'], ['len'])
 
-        tensor_shape = []  # type: ignore
-        len_shape = []  # type: ignore
+        tensor_shape = ['n']  # type: ignore
+        splits_shape = [3]  # type: ignore
 
         x = np.array([]).astype(np.float32)
-        out_len = np.int64(0)
+        splits = np.array([0, 0, 0]).astype(np.int64)
```
wschin: Is this a valid `splits`? From the [spec](https://github.com/onnx/onnx/blob/master/docs/Operators.md#splittosequence), I see
```
the sum of entries in 'split' must be equal to the dimension size of input tensor on 'axis'.
```

PR #2680, opened 2020-03-25
Add 'ignore_index' input in the spec for SoftmaxCrossEntropyLoss and NLLLoss.

codemzs:
This change fixes the spec for SoftmaxCrossEntropyLoss and NLLLoss to make it compatible with PyTorch.


docs/Changelog.md
```diff
@@ -14814,6 +14814,8 @@ This version of the operator has been available since version 12 of the default
 <dd>Target tensor of shape (N) or (N, d1, d2, ..., dk). Target element value shall be in range of [0, C).</dd>
 <dt><tt>weight</tt> (optional) : T</dt>
 <dd>Optional rescaling weight tensor. If given, it has to be a tensor of size C. Otherwise, it is treated as if having all ones.</dd>
+<dt><tt>ignore_index</tt> (optional) : Tind</dt>
```
wschin: If we can set the `weight` of a single class to 0, why do we need `ignore_index`?

codemzs: Hacking around is not good. You will increase the computation cost with this workaround. We are keeping this spec strictly for PyTorch. Next version will have a provision for TensorFlow.

lara-hdr:
@codemzs, thanks for the PR :)
I have some comments but I cannot add them inline since the lines are not modified;

NllLoss and SoftmaxCrossEntropy are functions not operators; so they have a function body in ONNX that we will need to update, even if there is a kernel.

The functions are defined in onnx/defs/math/defs.cc, line 1831 and 2320;
We will need to update the 'if' clauses 'if (!ctx.hasInput(2))' to deal with cases with 2, 3, 4 inputs.In case ignore_iindex is specified (4 inputs), we can use ScatterElements to put a 0 in the weight on the ignored index in the weights.

onnx/defs/math/defs.cc
```diff
@@ -2367,38 +2373,44 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "K-dimensional loss. Otherwise, it is a scalar.",
             "T")
-	.Output(
-	    1,
-	    "log_prob",
-	    "Log probability tensor. If the output of softmax is prob, its value is log(prob).",
-	    "T",
-	    OpSchema::Optional)
-        .TypeConstraint(
+        .Output(
```
lara-hdr: why is this is required now?
PyTorch only has 1 output for NllLoss/CrossEntropy

onnx/backend/test/case/node/softmaxcrossentropy.py
```diff
@@ -10,7 +10,7 @@
 from . import expect
 
 
-def softmaxcrossentropy(x, target, weight=None, reduction='mean'):  # type: ignore
+def softmaxcrossentropy(x, target, weight=None, reduction='mean', ignore_index=-100):  # type: ignore
```
lara-hdr: maybe set ignore_index to None?
(Unlike PyTorch, the spec does not say that the default is -100.)

onnx/backend/test/case/node/softmaxcrossentropy.py
```diff
@@ -178,3 +180,27 @@ def export_softmaxcrossentropy_mean_weights():  # type: () -> None
 
         # Check results
         expect(node, inputs=[x, labels, weights], outputs=[sce], name='test_softmax_cross_entropy_mean_weight')
+
+    @staticmethod
+    def export_softmaxcrossentropy_mean_weights_ignore_index():  # type: () -> None
```
lara-hdr: can we add a case with no weights and ignore_index specified?

onnx/backend/test/case/node/softmaxcrossentropy.py
```diff
@@ -10,7 +10,7 @@
 from . import expect
```
lara-hdr: Should the nllloss tests be updated?

snnn:
Is there any tests to verify if the function body works as expected?

onnx/defs/math/defs.cc
KsenijaS: it's optional output for softmaxcrossentropy, it was a part of the original PR

docs/Changelog.md
gramalingam: I don't understand this. Using weights seems the better "non-hacking" solution to me. We already have weights. Is it aligned to PyTorch or TF or does it handle both? If the weights doesn't correctly handle both PyTorch and TF, we already have a problem that needs to be fixed.

onnx/defs/math/defs.cc
lara-hdr: @KsenijaS initially it was optional, but this PR makes it required.
I am curious of the reason since PyTorch does not have it as an output

onnx/defs/math/defs.cc
```diff
@@ -1897,6 +1897,12 @@ ONNX_OPERATOR_SET_SCHEMA(
             "If given, it has to be a tensor of size C. Otherwise, it is treated as if having all ones.",
             "T",
             OpSchema::Optional)
+        .Input(
```
lara-hdr: why not make it an attribute?

lara-hdr: Talked with @codemzs offline, he is ok to make this an attribute for NllLoss and SoftmaxCrossEntropy.
cc : @SherlockNoMad

gramalingam:
> Is there any tests to verify if the function body works as expected?

Every "node" test-case is also expanded into a test-case for the corresponding function-body: for example, like here: https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/node/test_softmax_cross_entropy_mean_3d_expanded

onnx/defs/math/defs.cc
lara-hdr: my bad, it is still optional; the "OpSchema::Optional" is after the below red lines.

docs/Changelog.md
```diff
@@ -14801,6 +14801,8 @@ This version of the operator has been available since version 12 of the default
 #### Attributes
 
 <dl>
+<dt><tt>ignore_index</tt> : int (default is -100)</dt>
```
wschin: Where does this default come from? If we don't have a good default value, maybe we can make this attribute non-optional?

docs/Operators.md
```diff
@@ -18394,6 +18398,37 @@ expect(node, inputs=[x, labels, weights], outputs=[sce], name='test_softmax_cros
...
+node = onnx.helper.make_node('SoftmaxCrossEntropyLoss',
+                             inputs=['x', 'y', 'w'],
+                             outputs=['z'],
+                             reduction=reduction,
+                             ignore_index=ignore_index)
+
+# Define operator inputs.
+np.random.seed(0)
+x = np.random.rand(3, 5).astype(np.float32)
+labels = np.random.randint(0, high=5, size=(3, ))
```
wschin: Maybe you want to make sure this `labels` contain `0`, which will be ignored.

onnx/backend/test/case/node/softmaxcrossentropy.py
```diff
@@ -20,13 +20,15 @@ def softmaxcrossentropy(x, target, weight=None, reduction='mean'):  # type: igno
         N, C = input_shape
         neg_gather_element_input = np.zeros((N, ), dtype=np.float32)
         for i in range(N):
-            neg_gather_element_input[i] = -inp[i][target[i]]
+            if target[i] != ignore_index:
```
wschin: Just want to double-check. Do you have some python code to compare pytorch and this reference implementation?

docs/Changelog.md
codemzs: Resolved offline with @wschin. In his words "I understand composing Pytorch SoftmaxCrossEntropy is hard (but still doable with ops including LabelEncoder). I am fine with this change because using LabelEncoder doesn’t sound a good idea. I feel we can still keep in mind that ONNX is an IR for both of Tensorflow and Pytorch. If we only consider Pytorch, the value of ONNX may decrease."

Yes, this spec also supports TensorFlow’s tf.nn.sparse_softmax_cross_entropy_with_logits but we will need a minor change in the TensorFlow converter to permute and reshape the logits from [N, D1, D2…Dk, C] to [N, C, D1, D2…Dk]

docs/Changelog.md
codemzs: It is coming from PyTorch but I can make it non-optional.

codemzs:
@wschin I have removed the default value for `ignore_index` but still kept it as optional because it is easier to implement the function body. In the converter we will pass an `ignore_index` only if its value provided is within the range [0, C). I did this at the suggestion of @lara-hdr.

onnx/backend/test/case/node/softmaxcrossentropy.py
codemzs: We spoke offline, now that ignore_index is not an input so this is not needed.

onnx/backend/test/case/node/softmaxcrossentropy.py
lara-hdr: Yes, here it is.
[sce_test.zip](https://github.com/onnx/onnx/files/4400389/sce_test.zip)

codemzs: thanks @lara-hdr for running these tests and validating them with PyTorch.

onnx/defs/math/defs.cc
```diff
@@ -1939,7 +1939,8 @@ ONNX_OPERATOR_SET_SCHEMA(
             std::string("mean"))
         .Attr(
             "ignore_index",
-            "Specifies a target value that is ignored and does not contribute to the input gradient.",
+            "Specifies a target value that is ignored and does not contribute to the input gradient. "
+            "It is an optional value and valid values are [0, C).",
```
snnn: Usually the optional attribute should have a default value defined in the schema, unless the default value is too complex to be provided there.

codemzs: Please read @wschin 's comment. ORT also has a way to get default value when one is not defined, i.e `GetAttrOrDefault`

snnn: Thank you for clarifying this.

onnx/defs/math/defs.cc
```diff
@@ -1837,33 +1843,66 @@ bool BuildContextDependentFunctionBody(const FunctionBodyBuildContext& ctx, cons
...
-        body.push_back({{"loss"}, "ReduceSum", {"loss_Ndd"}, {MakeAttribute("keepdims", (int64_t)0)}});  
-      }
-    } 
-  } else {
-    body.push_back({{"weight_gather"}, "Gather", {"weight", "target"}});
-    body.push_back({{"loss_unweighted"}, "Squeeze", {"loss_N1dd"}, {MakeAttribute("axes", std::vector<int64_t>({1}))}});
-    if (ctx.getAttribute("reduction")->s() == "none") {
-      body.push_back({{"loss"}, "Mul", {"loss_unweighted", "weight_gather"}});
+    if(ctx.getAttribute("ignore_index") == nullptr)
+    {
```
snnn: Nit: please have a consistent coding style. (The position of "{" )

codemzs:
@wschin / @ebarsoum / @gramalingam Please merge this change. All of the **SoftmaxCrossEntropyLoss** & **NegativeLogLikelihoodLoss** ONNX tests on ONNX Runtime are now passing with the updated spec. @lara-hdr  has also verified the outputs match with that of PyTorch. 

image

Thank you, @chinhuang007 for giving me time till end of weekend to wrap this change up, much appreciated! This change is ready to be shipped with ONNX 1.7 release.

CC: @SherlockNoMad

chinhuang007:
@codemzs Thank you for the contribution!

PR #2687, opened 2020-03-30
Rel 1.7.103 verify

chinhuang007:
The PR is to bring rel-1.7.0 up-to-date and set up a temporary version number for release verification.


chinhuang007:
@onnx/sig-archinfra @onnx/sig-operators Please help review so we can start the release process. Thanks.

PR #2690, opened 2020-03-31
Update reference implementation and add tests for SoftmaxCrossEntropy and NegativeLogLikelihood

lara-hdr:
Update the reference implementation to include un-handled case and implementations for 3d and 4d inputs and add tests.
This PR does not introduce any spec changes.


gramalingam:
Is this needed to be included in the 1.7 release?

lara-hdr:
@gramalingam yes that would be great.

cc: @codemzs

lara-hdr:
@faxu @prasanthpul @wschin 

This PR is good to have in the release. 
It only updates the reference implementation and adds some tests, so it does not introduce a spec change.

faxu:
CC @chinhuang007

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -21,16 +21,31 @@ def compute_negative_log_likelihood_loss(input, target, weight=None, reduction='
         for i in range(N):
             if target[i] != ignore_index:
                 neg_gather_element_input[i] = -input[i][target[i]]
-    else:
+    elif len(input_shape) == 3:
```
gramalingam, line 24: This may be a naive question: if we transpose the first two axes of input (so that C becomes the first axis) and flatten into a 2D (making all axes except first one into one axis), would it help simplify things? Of course, "target" also has to be similarly flattened. In other words, is there any difference in the treatment of axes other than C? If all those axes are treated the same way, it would simplify the logic to flatten them.

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -21,16 +21,31 @@ def compute_negative_log_likelihood_loss(input, target, weight=None, reduction='
...
                         neg_gather_element_input[i][d1][d2] = -input[i][target[i][d1][d2]][d1][d2]
+    else:
+        raise NotImplementedError
 
     loss = neg_gather_element_input
+
+    if weight is None and ignore_index is not None:
+        c = input_shape[1]
+        weight = np.ones(c, dtype=np.float32)
+        weight[ignore_index] = np.float32(0)  # type: ignore
```
gramalingam, line 47: I am a bit confused: it looks like "ignore_index" is being handled in two places (both in the conditions above which update neg_gather_element_input only for values other than ignore_index, and here using weight). Is that redundant?

chinhuang007:
@faxu I have no problem to add this to the release 1.7 branch. However it will take another two days after this PR is merged to go through the build, publish, and verification cycle. Can we consider putting it as a post 1.7 fix? Since the release is owned by the ArchInfra SIG, so @linkerzhang please advise what we should do, in case additional PRs "should" be part of 1.7. Thanks!

gramalingam:
On the topic of 1.7 release, please note that https://github.com/onnx/onnx/pull/2691 is a bug fix to shape-inference. That PR has been merged into master. I assume that if we include this PR, we will include the other one too?

neginraoof:
@linkerzhang It would be great if we could add this other PR that Rama mentioned above (this a bug fix) to 1.7: https://github.com/onnx/onnx/pull/2691 
Thanks

chinhuang007:
@gramalingam Yes, if we decide to include this PR in 1.7 release, I assume #2691 will be included. I don't see the reason to cherry pick particular PR at this point unless told by the Archinfra SIG to do so.

onnx/backend/test/case/node/negativeloglikelihoodloss.py
gramalingam: At the least, "input = input.reshape((N,C,-1))" and "target = target.reshape((N,-1))" should let us handle different D cases uniformly.

onnx/backend/test/case/node/negativeloglikelihoodloss.py
lara-hdr: it's not really redundant. 
For example, in case the reduction is mean and ignore_index is specified, we don't want to include all the target values in the mean operation but only the non-ignored ones. This is why this if is added.

onnx/backend/test/case/node/negativeloglikelihoodloss.py
lara-hdr: good point. I will try that.

onnx/backend/test/case/node/negativeloglikelihoodloss.py
gramalingam: Yes, but I think this makes the earlier "if-condition" in line 22/29/37 redundant, I think. IIUC, if ignore_index is specified, then we end up introducing a weight and ensure that the weight is zero for ignore_index, and end up executing line 72 "loss = gather_weight * loss". So, why do we need the earlier if-condition in line 22 etc.? (I might be wrong. Please verify before changing it.)

wschin:
Is this PR tested with pytorch? How did you verify the correctness of the new cases?

onnx/backend/test/case/node/negativeloglikelihoodloss.py
wschin, line 47: weight[ignore_index] sometime may not be valid. `ignore_idex` can be 100 while the weight is just a 10-element vector.

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -40,12 +55,19 @@ def compute_negative_log_likelihood_loss(input, target, weight=None, reduction='
...
                         gather_weight[i] = 0
-
-            if len(input_shape) == 3:
+            elif len(input_shape) == 3:
                 for i in range(input_shape[0]):
-                    for j in range(input_shape[1]):
+                    for j in range(input_shape[2]):
                         if target[i][j] == ignore_index:
                             gather_weight[i][j] = 0
+            elif len(input_shape) == 4:
```
wschin, line 63: humm... if you have case for `len(input_shape) == 3`, why do you need another case? We can flatten `[N, C, H, W]` to `[N, C, H*W]` and apply the 3-D branch.

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -11,74 +11,60 @@
...
     if len(input_shape) == 2:
-        N, C = input_shape
         neg_gather_element_input = np.zeros((N, ), dtype=np.float32)
         for i in range(N):
             if target[i] != ignore_index:
                 neg_gather_element_input[i] = -input[i][target[i]]
-    elif len(input_shape) == 3:
-        N, C, dim1 = input_shape
-        neg_gather_element_input = np.zeros((N, dim1), dtype=np.float32)
+    else:
```
wschin, line 41: `len(input_shape) == 1` will enter this block? If yes, can we throw?

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -11,74 +11,60 @@
...
+        D = input.shape[2]
+        neg_gather_element_input = np.zeros((N, D), dtype=np.float32)
         for i in range(N):
-            for d1 in range(dim1):
+            for d1 in range(D):
```
wschin, line 45:
```suggestion
            for d in range(D):
```
maybe?

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -11,74 +11,60 @@
...
-        N, C = input_shape
         neg_gather_element_input = np.zeros((N, ), dtype=np.float32)
         for i in range(N):
             if target[i] != ignore_index:
                 neg_gather_element_input[i] = -input[i][target[i]]
-    elif len(input_shape) == 3:
-        N, C, dim1 = input_shape
-        neg_gather_element_input = np.zeros((N, dim1), dtype=np.float32)
+    else:
+        D = input.shape[2]
```
wschin, line 42:
```suggestion
        // Get a dimension from the reshaped input. If the original input shape is [N, C, H, W], the D here should be H * W because we reshape [N, C, H, W] to [N, C, H * W].
        D = input.shape[2]
```

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -11,74 +11,60 @@
...
+        # when the target valeu is > C, it doesn't matter which value we are
+        # taking in gather_weight, since it will be set to 0 in the following lines
+        gather_weight = np.take(weight, target, mode='clip')
+        if ignore_index is not None:
+            gather_weight = np.where(target == ignore_index, 0, gather_weight).astype(dtype=np.float32)
+    elif ignore_index is not None:
+        gather_weight = np.where(target == ignore_index, 0, 1).astype(dtype=np.float32)
+
+    # if input is 4-d and above, make it 3-d
+    if len(input_shape) > 3:
```
wschin, line 32: We can reshape input shape `[N, C]` to `[N, C, 1]` and then all the following code become applicable to all cases without if-else. Note that we still need to throw if the input rank is less than 2.

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -11,52 +11,60 @@
...
     input_shape = input.shape
+    target_shape = target.shape
+    N = input_shape[0]
+    C = input_shape[1]
+
+    # initialize the positional weights when required
+    gather_weight = None
+    if weight is not None:
+        # setting mode='clip' to deal with ignore_index > C cases.
+        # when the target valeu is > C, it doesn't matter which value we are
```
wschin, line 23:
```suggestion
        # when the target valeu is > C or < 0, it doesn't matter which value we are
```

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -11,52 +11,60 @@
...
+        # taking in gather_weight, since it will be set to 0 in the following lines
```
wschin, line 24:
```suggestion
        # taking in gather_weight, since it will be set to 0 in the following fi-block.
```

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -11,52 +11,60 @@
...
+        gather_weight = np.take(weight, target, mode='clip')
+        if ignore_index is not None:
+            gather_weight = np.where(target == ignore_index, 0, gather_weight).astype(dtype=np.float32)
+    elif ignore_index is not None:
+        gather_weight = np.where(target == ignore_index, 0, 1).astype(dtype=np.float32)
```
wschin, line 29:
```suggestion
        // Set `ignore_index`'s loss weight to 0. The loss tensor will be multiplied by this weight tensor, so `ingore_index`'s loss value will be eliminated.
        gather_weight = np.where(target == ignore_index, 0, 1).astype(dtype=np.float32)
```

lara-hdr:
@chinhuang007 this is ready for merge

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -220,10 +231,264 @@ def export_input_shape_is_NCd1d2_with_weight_reduction_sum_ignore_index():  # ty
...
+        np.random.seed(0)
+        input = np.random.rand(N, C, d1).astype(np.float32)
+        target = np.random.randint(0, high=C, size=(N, d1))
+        target[0][0] = np.int64(1)
+
+        negative_log_likelihood_loss = compute_negative_log_likelihood_loss(input, target, weight=None, reduction=reduction, ignore_index=ignore_index)
+
+        expect(node, inputs=[input, target], outputs=[negative_log_likelihood_loss],
+            name='test_negative_log_likelihood_loss_input_shape_is_NCd1_ignore_index')
+    """
```
codemzs, line 333: This test assumes ignore_index is within range and if this is failing then we must investigate why function body definition is not working because it should handle cases where ignore_index is within range but if it is not handling that case then we must investigate and fix instead of commenting the test. Please note, the "update of function body" work item is for handling ignore_index that is **out of** range.

PR #2691, opened 2020-03-31
Fix for einsum shape ineference

neginraoof:
There's a bug in einsum shape inference for inferring the output rank with ellipsis in right hand side of the equation.

### DIFF
--- a/onnx/test/shape_inference_test.py
+++ b/onnx/test/shape_inference_test.py
@@ -2980,6 +2980,14 @@ class TestShapeInference(unittest.TestCase):
             [],)
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (None, None))])  # type: ignore
 
+    def test_einsum_outer_prod(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, (3, 5)),
+             ('y', TensorProto.FLOAT, (7, 9))],
+            [make_node('Einsum', ['x', 'y'], ['z'], equation='ij,ab->ijab')],
+            [],)
+        self._assert_inferred(graph, [make_tensor_value_info('z', TensorProto.FLOAT, (None, None, None, None))])  # type: ignore
+
     def test_einsum_sum_along_dim(self):  # type: () -> None
         graph = self._make_graph(
             [('x', TensorProto.FLOAT, (3, 4))],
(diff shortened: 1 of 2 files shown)


neginraoof:
Would appreciate it if you could review this PR @gramalingam

gramalingam:
Thanks for the fix. Regarding the line
```cpp
num_ellipsis_indices = rank - term.size() + 3;
```
can we add a check whether rank >= (term.size()-3)? What happens if the input rank is less than the minimum expected?

onnx/test/shape_inference_test.py
```diff
@@ -2980,6 +2980,14 @@ def test_einsum_transpose(self):  # type: () -> None
             [],)
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (None, None))])  # type: ignore
 
+    def test_einsum_outer_prod(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, (3, 5)),
+             ('y', TensorProto.FLOAT, (7, 9))],
+            [make_node('Einsum', ['x', 'y'], ['z'], equation='ij,ab->ijab')],
+            [],)
+        self._assert_inferred(graph, [make_tensor_value_info('z', TensorProto.FLOAT, (None, None, None, None))])  # type: ignore
```
gramalingam: So, the inference doesn't currently infer the output-shape of (3,5,7,9), even though it has the necessary information to infer it? We don't have to do this in the current PR/release, but I think we should make the shape-inference more complete to infer the dimensions where possible.

neginraoof:
> Thanks for the fix. Regarding the line
> 
> ```c++
> num_ellipsis_indices = rank - term.size() + 3;
> ```
> 
> can we add a check whether rank >= (term.size()-3)? What happens if the input rank is less than the minimum expected?

Thanks, I just added this check.

onnx/test/shape_inference_test.py
```diff
@@ -2989,11 +2997,38 @@ def test_einsum_sum_along_dim(self):  # type: () -> None
 
     def test_einsum_ellipsis(self):  # type: () -> None
         graph = self._make_graph(
-            [('x', TensorProto.FLOAT, (3, 4))],
+            [('x', TensorProto.FLOAT, (3, 4, 4))],
             [make_node('Einsum', ['x'], ['y'], equation='... ii ->... i')],
             [],)
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (None, None))])  # type: ignore
 
+    def test_einsum_ellipsis_2(self):  # type: () -> None
```
hariharans29: Can we add an "implicit" version of this test too ? Something like "...ij,...jk" alone ?

hariharans29:
Thanks for the fix. 

There are still some cases to check:

1) Make sure the letters we see (inputs and outputs) are in the allowed range ['a', 'z']
2) Make sure we don't have a letter in the output that we haven't seen in the input
3) Make sure we don't have a letter that repeats in the output
4) Make sure we don't have an ellipsis in the output when we don't have one in any of the inputs

Can come in a later PR though.

onnx/test/shape_inference_test.py
neginraoof: Sure, I added this test.

onnx/test/shape_inference_test.py
neginraoof: So actually I did not look into shape inference since Einsum could results in many different operations. I'm not sure if it's okay to have shape inference supported partially, based on the equation. If you think otherwise, we can look into supporting some cases.

neginraoof:
> Thanks for the fix.
> 
> There are still some cases to check:
> 
> 1. Make sure the letters we see (inputs and outputs) are in the allowed range ['a', 'z']
> 2. Make sure we don't have a letter in the output that we haven't seen in the input
> 3. Make sure we don't have a letter that repeats in the output
> 4. Make sure we don't have an ellipsis in the output when we don't have one in any of the inputs
> 
> Can come in a later PR though.

Actually I did not add any checks for the equation string itself. Though these tests can be considered for rank inference too. I can send a follow up PR.

onnx/test/shape_inference_test.py
gramalingam: As I mentioned above, let us not do it for this PR or this release. I can help do this after the release.

neginraoof: Thanks @gramalingam

neginraoof:
cc @linkerzhang Woulc it be possible to include this fix in 1.7 branch please?

snnn:
Do you need to regenerate the test data as well?

neginraoof:
@snnn this is merged. But should shape inference test data be re-generated?

neginraoof:
@snnn Could you please let me know if this will be part of 1.7 release?

PR #2696, opened 2020-04-03
Update `ignore_index` documentation for SoftmaxCrossEntropyLoss and NegativeLogLikelihoodLoss

codemzs:
`ignore_index` can be arbitrary integer, so it should not be constrained by the number of classes in the classification problem. This PR only changes the spec, so reference implementation and function body need to be updated too. Note that all current tests are ok because their `ignore_index` doesn't exceed the maximum class index.


codemzs:
@chinhuang007 This is a very small **one line documentation change** that is important and I will really appreciate if you can cherry pick this into the release branch. It is extremely low risk as it is just a documentation change but it clarifies the behavior of the spec.

docs/Changelog.md
```diff
@@ -14802,7 +14802,7 @@ This version of the operator has been available since version 12 of the default
 
 <dl>
 <dt><tt>ignore_index</tt> : int</dt>
-<dd>Specifies a target value that is ignored and does not contribute to the input gradient. It is an optional value and valid values are [0, C).</dd>
+<dd>Specifies a target value that is ignored and does not contribute to the input gradient. Its an optional value.</dd>
```
spandantiwari, line 14805: nit: "It's"

spandantiwari:
Just as note for posterity - today the function subgraph for this function still assumes that the `ignore_indices` are restricted to [0, C). The subgraph will be updated in a separate PR later.

codemzs:
@spandantiwari Thanks. I just want to add to it that updating the function body implementation is needed but it is not a ship blocker since we have the implementation of the op in ORT and ORT is the primary user of this spec. I will try to get the function body also implemented in a different PR but will leave it to folks to decide if we should take it for 1.7

linkerzhang:
@codemzs not exactly, please don't take an assumption that ORT is the "only" or primary user of this spec. that's not true. ONNX does have many partners. Per my personal investigation now, there're many mobile/edge related inference engines which are depending on ONNX. That said, making the spec correct (including function body correctness) is super important. @ebarsoum @spandantiwari @prasanthpul

codemzs:
@linkerzhang thanks for the merge, I will try my best to get correct the function body as well for 1.7 release. A little new to this ONNX world so made the naive assumption but thanks for educating me.

PR #2703, opened 2020-04-06
Make NegativeLogLikelihoodLoss function body handle out of range ignore_index and re-enable all tests

codemzs:
This change completes the function body for NegativeLogLikelihoodLoss and ensures it can handle `ignore_index` that is out of [0, C) range and within [0. C) range and re-enables all 6 of the disabled tests.


onnx/defs/math/defs.cc
```diff
@@ -1937,42 +1944,61 @@ bool BuildContextDependentFunctionBody(
...
+                      {"target_shape"},
                       {MakeAttribute("value", ToDimensionOneFloatTensor(1))}});
-      body.push_back(
-          {{"weights_default"},
-           "ScatterElements",
-           {"const_weights_ones", "const_ignore_index", "const_zero_float"}});
-      body.push_back(
-          {{"weight_gather"}, "Gather", {"weights_default", "target"}});
+
+      body.push_back({{"weight_gather"}, "Where", {"squeeze_mask", "const_zero_float", "const_weights_ones_float"}});
```
gramalingam, line 1988: Minor nit: Can't we use "const_one_float" instead of "const_weights_ones_float"? Doen's "squeeze_mask" already have the desired shape? Just wondering.

codemzs: Good point, let me try.

---
In reply to: [403821108](https://github.com/onnx/onnx/pull/2703#discussion_r403821108) [](ancestors = 403821108)

codemzs: This works. Let me push the commit.

---
In reply to: [403821282](https://github.com/onnx/onnx/pull/2703#discussion_r403821282) [](ancestors = 403821282,403821108)

codemzs:
@chinhuang007 This is the **last** commit to go with ONNX 1.7 release, it **does not** add any new features but it fixes all the gaps we found with the function body for **NegativeLogLikelihoodLoss** and re-enables all of the disabled tests. Without this change we will shipping a function body that is incomplete and as @linkerzhang pointed [here](https://github.com/onnx/onnx/pull/2696#issuecomment-608954671) function body correctness is extremely important even though ONNX Runtime has an operator implementation. We have several partners that *may* not use ONNX Runtime and instead may depend on function body definition.

Thank you @gramalingam and @skottmckay for all your help and doing the review today!!

CC: @ebarsoum @wschin @prasanthpul @spandantiwari @faxu @SherlockNoMad @linkerzhang

PR #2713, opened 2020-04-08
Updated shape inference to fix the bug for scalars

neginraoof:
There's a bug in shape inference for einsum for scalar tensors.


linkerzhang:
Thanks for fixing bugs!

Interesting... fixing shape inference logic should not change test models/data, right?

onnx/defs/math/defs.cc
```diff
@@ -2178,17 +2178,18 @@ void einsumRankInference(
 
   // Parse the left-hand side
   std::stringstream str(left_equation);
-  while (std::getline(str, term, ',')) {
+  while (!str.eof()) {
```
gramalingam, line 2181: I am trying to understand why the earlier version exited the while-loop at all. Anyone know? The getline spec says that it returns a reference to a stream. (Just want to double-check what the difference in behavior here.)

onnx/defs/math/defs.cc
```diff
@@ -2178,17 +2178,18 @@ void einsumRankInference(
...
+    std::getline(str, term, ',');
     auto ellipsis_index = term.find("...");
+    if (numInputs <= num_operands) {
+      fail_shape_inference(
+          "Number of input tensors does not match the operands in the equation.");
+    }
+    size_t rank =
+    ctx.getInputType(num_operands)->tensor_type().shape().dim_size();
```
gramalingam, line 2189: Is the formatting produced by clang? Looks odd ... I expected some indentation of the second line.

gramalingam:
There is a CI failure due to a flake8 complaint, needs some space character after comma or something like that.

neginraoof:
cc @wschin @gramalingam Would you please review? Thanks.

onnx/defs/math/defs.cc
neginraoof: No, I just updated it

neginraoof:
@gramalingam Thanks for review.

neginraoof:
@linkerzhang I'm just updating the ONNX IR version for models

neginraoof:
Failures here are from Caffe2 backend tests:
FAILED test/onnx/test_pytorch_onnx_caffe2.py::TestCaffe2Backend_opset7::test_scalar_type
FAILED test/onnx/test_pytorch_onnx_caffe2.py::TestCaffe2BackendEmbed_opset7::test_scalar_type
FAILED test/onnx/test_pytorch_onnx_caffe2.py::TestCaffe2Backend_opset8::test_scalar_type
FAILED test/onnx/test_pytorch_onnx_caffe2.py::TestCaffe2BackendEmbed_opset8::test_scalar_type

onnx/defs/math/defs.cc
gramalingam: Ok, I understand ... it is implicitly converting the stream object into a bool, after the "getline()" is done ... and so it wasn't processing the "empty-string" getline returns in the last iteration.

neginraoof: Thanks for clarifying this

gramalingam:
Hi @linkerzhang : the merge of PR 2268 seems to be causing CircleCI failure (the problem is not with the PR, but the model-checker is exposing failures in some pre-existing model-export). Should that PR be reverted? Or, should we go ahead and merge this PR in without CircleCI? Thanks.

PR #2714, opened 2020-04-08
refactoring: early return 'if else' -> 'if'

marload:
Hello.
I think 'if' syntax is better than 'if else' syntax in the early return pattern.


gramalingam:
Curious: is this just as a matter of coding-style, or does it have some effect (on performance)?

marload:
@gramalingam 
Matter of code style.

PR #2721, opened 2020-04-14
Clarify operator set reference and definition

natke:


docs/Versioning.md
```diff
@@ -65,7 +65,7 @@ Because onnx.proto is expected to be consumed by multiple independent developers
 
 ONNX is defined such that the IR can evolve independently from the set of operators. In ONNX, operators represent both the signature and semantics of a given operation. Operators are abstract interfaces in that they do not imply a specific implementation; rather, they are simply the contract between a model author and the implementations that model may execute on.
 
-A given operator is identified by a three-tuple: `(domain, op_type, and op_version)`. This is written as `domain.op_type:op_version` in prose (e.g., `com.acme.FastConv:3`). Nodes in graphs always refer to operators by their three-part identifier. Breaking opset changes include:
+A given operator is identified by a three-tuple: `(domain, op_type, and op_version)`. This is written as `domain.op_type:op_version` in prose (e.g., `com.acme.FastConv:3`). Nodes in graphs always refer to operators by their three-part identifier. Breaking operator changes include:
```
linkerzhang, line 68: good catch!

docs/Versioning.md
```diff
@@ -94,6 +94,8 @@ operator id MUST be greater than any extant `op_version` for the
 > 4. Register the new operator in the corresponding `operator_sets`
 >    header file.
 
+### Operator Sets
```
linkerzhang, line 97: I'm not sure it's good to add this line so that a separate "operator sets" section is there. It may mislead readers on getting the points that "Versioning is against an operator set, instead of an operator".

natke: That's fair. I do think a clear definition of an `Operator set` is necessary though. Especially given that users who are converting to ONNX from another format often need to specify the opset, and will need to know precisely what this means

linkerzhang: Agree that a separate section of "OperatorSet and OperatorSet Versioning" should be there. do you want to do another review and refactor of this section "Operator Versioning" and this new "OperatorSet" section please? 

basically, things should be highlighted, 1) Versioning is against OperatorSet, instead of Operator. 2) Operator is identified by tuple <name, domain, since_version> (currently, it's saying "op_version", which is not accurate and misleading). 3) Every ONNX release will at most have 1 OperatorSet version bump (if and only if there's any breaking changes) 4) list the details of "breaking changes".

natke: Yep sure, will do!

natke: @linkerzhang I created a new Operator set sub-section

PR #2722, opened 2020-04-14
Fix shape inference for unfold to depth

neginraoof:
Fix shape inference for unfold to depth.
Dimension should be = C * prod({kernel_size})

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -114,7 +114,7 @@ Given an input of shape (N x C x D1 x D2 ... x Dn), output would be a 3-D tensor
 <br/>
 Where number of blocks extracted from each spatial dimension d is:
 ```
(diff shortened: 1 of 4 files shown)


neginraoof:
cc @wschin @gramalingam @hariharans29  for review. Thanks a lot.

onnx/defs/tensor/defs.cc
```diff
@@ -56,7 +56,7 @@ void unfoldToDepthShapeInference(
   for (int i = 0; i < static_cast<int>(block_size.size()); i++) {
     // accounting for dilation, how big is the block in this dimension
     effective_block_size[i] = (effective_block_size[i] - 1) * dilations[i] + 1;
-    block_num_element *= effective_block_size[i];
+    block_num_element *= block_size[i];
```
linkerzhang, line 59: good catch!

PR #2724, opened 2020-04-15
[WIP] Add safety check to DataTypeUtils::FromString

snnn:
Whatever the input is, the code shouldn't crash.


onnx/defs/data_type_utils.cc
```diff
@@ -186,13 +191,16 @@ void DataTypeUtils::FromString(
   } else if (s.LStrip("map")) {
     s.ParensWhitespaceStrip();
     size_t key_size = s.Find(',');
+    if (key_size == std::string::npos)
```
snnn, line 194: This edge case was discovered by [klee](https://klee.github.io/)

snnn:
> you may want to have "fail early" here, as type string recognition is in a very early stage when loading the lib (it's also the same in ORT). Changing to return "nullptr" may not be a good idea as it hides the root cause of a failure, although you have a "status" returned, but it's not recorded or thrown out anywhere.

Currently, onnxruntime, almost does nothing when onnxruntime.dll is loaded. All the initialization are postponed to session creation time .

linkerzhang:
don't quite understand the point. I was saying that "early failure" should be advocated for this piece of codes. With your change, although there's a failed "status" returned, but it was swallowed and a nullptr without any error message is returned. I think this is not good? (but of course, the original code of crashing is not good neither).

onnx/defs/data_type_utils.cc
```diff
@@ -96,15 +97,19 @@ DataType DataTypeUtils::ToType(const TypeProto& type_proto) {
   std::lock_guard<std::mutex> lock(GetTypeStrLock());
   if (GetTypeStrToProtoMap().find(typeStr) == GetTypeStrToProtoMap().end()) {
     TypeProto type;
-    FromString(typeStr, type);
+    Common::Status st = FromString(typeStr, type);
+    if (!st.IsOK())
+      return nullptr;
```
wschin, line 102: Can we print something in this case? Maybe st.ErrorMessage(..) or something like that?

onnx/defs/data_type_utils.cc
```diff
@@ -186,13 +191,16 @@ void DataTypeUtils::FromString(
   } else if (s.LStrip("map")) {
     s.ParensWhitespaceStrip();
     size_t key_size = s.Find(',');
+    if (key_size == std::string::npos)
+      return Common::Status(
+          Common::NONE, Common::INVALID_ARGUMENT, "invalid type string");
```
wschin, line 196:
```suggestion
          Common::NONE, Common::INVALID_ARGUMENT, "invalid type string: ", type_str);
```

onnx/defs/data_type_utils.cc
```diff
@@ -243,13 +255,20 @@ bool DataTypeUtils::IsValidDataTypeString(const std::string& type_str) {
...
+  if (!IsValidDataTypeString(type_str))
+    return Common::Status(
+        Common::NONE, Common::INVALID_ARGUMENT, "invalid type string");
 
   TypesWrapper& t = TypesWrapper::GetTypesWrapper();
-  tensor_data_type = t.TypeStrToTensorDataType()[type_str];
+  auto iter = t.TypeStrToTensorDataType().find(type_str);
+  if (iter == t.TypeStrToTensorDataType().end())
+    return Common::Status(
+        Common::NONE, Common::INVALID_ARGUMENT, "invalid type string");
```
wschin, line 269:
```suggestion
        Common::NONE, Common::INVALID_ARGUMENT, "invalid type string: ", type_str);
```
At least user can search for the returned `type_str` in their code to find the error.

onnx/defs/data_type_utils.cc
```diff
@@ -96,15 +97,19 @@ DataType DataTypeUtils::ToType(const TypeProto& type_proto) {
...
     GetTypeStrToProtoMap()[typeStr] = type;
   }
   return &(GetTypeStrToProtoMap().find(typeStr)->first);
 }
 
 DataType DataTypeUtils::ToType(const std::string& type_str) {
   TypeProto type;
-  FromString(type_str, type);
+  Common::Status st = FromString(type_str, type);
+  if (!st.IsOK())
```
wschin, line 111: Can we throw and print out the error message from `st` here?

onnx/defs/data_type_utils.cc
snnn: No. Unless onnx has a Logger interface, otherwise it's better to not print anything in onnx code. Because it's often applications want to take a totally control on logging, they don't want us use stdout/stderr directly.

onnx/defs/data_type_utils.cc
snnn: I prefer to change it to return Common::Status instead, to make it consist with the other function.  It look bad when mixing the use of Common::Status and exception.  
What do you think?

wschin: I always prefer to propagate error to the surface. Let's return `Common::Status`. :)

onnx/defs/data_type_utils.cc
wschin: Makes sense. We can still return `Common::Status` as you suggested below.

PR #2725, opened 2020-04-15
Dropout, SoftmaxCrossEntropyLoss, NegativeLogLikelihoodLoss and Celu

codemzs:
- **Dropout**
  - Add `training_mode` input to Dropout spec to support training and inference mode by keeping default ratio as 0.5.
  - Fixes Dropout reference implementation as it was not scaling and did not have a provision to return mask.
  - Dropout only had two tests but this change adds several more tests to cover all possible cases.

- **SoftmaxCrossEntropyLoss**
  - Fixes reference implementation because it did not have a provision to return log_probability which is an optional parameter in the spec.
  - Adds ~60 more tests for SoftmaxCrossEntropyLoss to cover optional log_probability scenario.
  - Fixes function definition because it did not have an implementation to return optional parameter log_probability in addition it was computing softmax incorrectly because Max operator was being used to find max class score per example instead of ReduceMax(axis = 1). Max is used for finding element wise max across multiple tensors, not sure why this was considered for a single tensor and to find max for every class in each example. To sum up scores per class ReduceSum was being used without specifying axis = 1, this was summing across all axis instead of just along the classes.

- **NegativeLogLikelihoodLoss** 
  - Fixes function definition to handle target values of all possible supported types, previously it was only handling int32 and not int64.
  - Fix ShapeInference because it had incorrect assert on the dimensions of weights input.

- **Celu**
  - Fix  function body because attribute type was being passed as float_value attribute instead of tensor typed attribute.


onnx/backend/test/case/node/dropout.py
```diff
@@ -11,19 +11,14 @@
 from . import expect
 from onnx import helper
 
+def dropout(X, drop_probability=0.5, seed=0):  # type: ignore
```
gramalingam, line 14: This PR seems to also change the op specification about the default value of ratio/drop_probability. The current spec says that default value is 0. Two questions: (a) Why change the default-value? I understand that it makes it more compatible with PyTorch/TF, but I wonder why it was made 0 in the first-place and what things would break if we change it thus, (b) If we must change the default-value, the spec also needs to be changed/updated.

codemzs: @gramalingam The rationale is to be compatible with PyTorch/TF and even ORT has default as 0.5. Happy to update the spec but lets first understand why the default of zero was chosen, maybe @wschin might know?

linkerzhang: @gramalingam good catch!

I don't suggest to change this default value. I believe the "drop elimination" optimizer was assuming that the default value is 0 so that in inference case it's just a copy and can be removed.

codemzs: Thanks, @linkerzhang then in that case we need to change the "default_dropout" test to pass the probability of 0.5 because on ORT side default probability is set to 0.5 (same as PyTorch/TF). In anycase the reference implementation is not doing the scaling as it is done in PyTorch/TF and ORT.

gramalingam: @codemzs : it sounds like ORT implementation does not match the spec then. Either the spec needs to be changed or the ORT implementation needs to change (I don't think it is just a matter of changing the test-case).

onnx/defs/math/defs.cc
```diff
@@ -424,6 +424,37 @@ max(0,x) + min(0,alpha*(exp(x/alpha)-1))
...
+  t.add_dims(1);
+  return t;
+}
+
+bool BuildContextDependentFunctionBodyCelu(
+    const FunctionBodyBuildContext& ctx,
+    const OpSchema& schema,
+    FunctionProto& functionProto) {
+  std::vector<FunctionBodyHelper::NodeDef> body;
+float alpha = ctx.getAttribute("alpha")->f();
```
gramalingam, line 438: Use
```
   ctx.getAttribute("alpha") ? ctx.getAttribute("alpha")->f() : 1.0f
```

codemzs: Thanks for the reminder I was planning to checking for missing value but doesn't; this attribute already have a default value set in the schema? Also I think we should use the default const below. 

ctx.getAttribute("alpha") ? ctx.getAttribute("alpha")->f() : celu_default_alpha


---
In reply to: [410777048](https://github.com/onnx/onnx/pull/2725#discussion_r410777048) [](ancestors = 410777048)

onnx/defs/math/defs.cc
```diff
@@ -424,6 +424,37 @@ max(0,x) + min(0,alpha*(exp(x/alpha)-1))
...
+  body.push_back({{"alpha"},
+                  "Constant",
+                  {},
+                  {MakeAttribute("value", ToDimensionOneFloatTensor(alpha))}});
+
+  body.push_back({{"X_alpha"}, "Div", {"X", "alpha"}});
+  body.push_back({{"Elu_Result"}, "Elu", {"X_alpha"}, {{"alpha", 1.f}}});
```
gramalingam, line 445: Is that supposed to be 1.f or alpha from above?

codemzs: I think 1.f is the default value for the attribute but we should use celu_default_alpha instead. What do you think?

---
In reply to: [410777439](https://github.com/onnx/onnx/pull/2725#discussion_r410777439) [](ancestors = 410777439)

onnx/defs/math/defs.cc
gramalingam: I think it is just safer. I am not sure if the check is needed (this can be called from ORT, but it can also be called from other ONNX client-code ... I recall that ORT fills in default values, but not sure if that also happens in the ONNX context).

codemzs: Safer for sure.

---
In reply to: [410779075](https://github.com/onnx/onnx/pull/2725#discussion_r410779075) [](ancestors = 410779075)

onnx/defs/math/defs.cc
gramalingam: I don't know this op. But looking at the spec, it says the value is
```
   max(0,x) + min(0,alpha*(exp(x/alpha)-1))
```
So, I assume both alphas in this equation must be the attribute-value retrieved above?

gramalingam: But I am confused about the "max(0,x)" part too. How does the above equation get rewritten in the form shown in this function-body?

codemzs: Both alphas are attribute-value retrieved above that we put in a constant tensor by the name "alpha" 

---
In reply to: [410779367](https://github.com/onnx/onnx/pull/2725#discussion_r410779367) [](ancestors = 410779367)

codemzs: You make a good point, the test passed so I didn't look too closely but let me think.

---
In reply to: [410779760](https://github.com/onnx/onnx/pull/2725#discussion_r410779760) [](ancestors = 410779760)

codemzs: [https://github.com/onnx/onnx/pull/2575](https://github.com/onnx/onnx/pull/2575) that explains more on the implementation.

---
In reply to: [410779924](https://github.com/onnx/onnx/pull/2725#discussion_r410779924) [](ancestors = 410779924,410779760)

codemzs: ``Tensor celu(const Tensor & self, Scalar alpha) {
  double inv_alpha = 1. / alpha.to<double>();
  return at::elu(self, alpha, Scalar(1.0), Scalar(inv_alpha));
}``

---
In reply to: [410780998](https://github.com/onnx/onnx/pull/2725#discussion_r410780998) [](ancestors = 410780998,410779924,410779760)

gramalingam: May be this is okay, since there is the multiplication by alpha down below.

codemzs: yep

---
In reply to: [410781381](https://github.com/onnx/onnx/pull/2725#discussion_r410781381) [](ancestors = 410781381)

onnx/backend/test/case/node/dropout.py
linkerzhang: I suggest not updating the spec, which costs more than the benefits it brings.

codemzs:
@linkerzhang Thank you for reviewing. Please also review other changes in this PR. We have a meeting on Monday to close down on the default value for Dropout. Lastly, I will appreciate if you can please not merge PRs in the master branch unless you have very high confidence they are low risk. The reason being we are trying to release 1.7 with quality and only want PRs to go in that we have confidence in or are bug fixes, this is the reason we made the call to revert several ops that were merged last minute.

onnx/backend/test/case/node/dropout.py
codemzs: @linkerzhang I have made a note of your suggestion but the final decision will be taken by the team on Monday morning, this includes @ebarsoum , @gramalingam , @wschin , @yuanbyu , @SherlockNoMad , @prasanthpul , etc.

codemzs: Thanks @linkerzhang ! lets see what other reviewers have to say? @ebarsoum @yuanbyu @wschin @gramalingam

codemzs: @linkerzhang Please take a look again, we added a training_mode input to Dropout so that existing inference behavior does not break while the operator can support training that can be dynamically switched with default probability of 0.5

onnx/backend/test/case/node/dropout.py
```diff
@@ -12,21 +12,25 @@
...
+            return X, np.ones(X.shape, dtype=bool)
+        else:
+            return X
 
-    for index in indices_to_dropout:
-        flattened_x[index] = 0.
-    y = flattened_x.reshape(x.shape)
-    return y
+    np.random.seed(seed)
+    mask = np.random.uniform(0, 1.0, X.shape) >= drop_probability
```
SherlockNoMad, line 23: We cannot match the output value in the onnx test, right? as the mask is randomly generated, and different backend will generate different mask.

onnx/backend/test/case/node/dropout.py
```diff
@@ -37,8 +41,133 @@ def export_default():  # type: () -> None
...
+            'Dropout',
+            inputs=['x', 'r', 't'],
+            outputs=['y', 'z'],
+        )
+
+        x = np.array([-1, 0, 1]).astype(np.float32)
+        r = np.float32(0.75)
+        t = np.bool_(True)
+        y, z = dropout(x, r, training_mode=t, return_mask=True)
+        expect(node, inputs=[x, r, t], outputs=[y, z], name='test_training_dropout_mask')
```
SherlockNoMad, line 140: Wouldn't this test fail in ORT's onnx_test, as mask is random.

onnx/backend/test/case/node/dropout.py
codemzs: It is currently failing but as discussed offline I want to see it I can some how approximate the comparison but if that is not feasible I’ll skip _expect_ as you suggested.__

codemzs: After setting the seed attribute of dropout to zero and also in the reference implementation, tests on ORT pass.

onnx/backend/test/case/node/dropout.py
codemzs: After setting the seed attribute of dropout to zero and also in the reference implementation, tests on ORT pass.

onnx/backend/test/case/node/dropout.py
codemzs: I have updated ORT test infra to count for number of zeros and make sure they are around the same count as baseline.

onnx/backend/test/case/node/dropout.py
codemzs: I have updated ORT test infra to count for number of zeros and make sure they are around the same count as baseline.

onnx/defs/math/defs.cc
```diff
@@ -2527,40 +2545,42 @@ bool BuildContextDependentFunctionBodySCE(
...
+                  "ReduceMax",
+                  {"scores"},
+                  {MakeAttribute("axes", std::vector<int64_t>({1}))}});
+
   body.push_back({{"X_Sub"}, "Sub", {"scores", "X_Max"}});
   body.push_back({{"X_Exp"}, "Exp", {"X_Sub"}});
-  body.push_back({{"X_RS"}, "ReduceSum", {"X_Exp"}});
+  body.push_back({{"X_RS"}, "ReduceSum", {"X_Exp"}, {MakeAttribute("axes", std::vector<int64_t>({1}))}});
   body.push_back({{"X_Div"}, "Div", {"X_Exp", "X_RS"}});
-  body.push_back({{"log_prob"}, "Log", {"X_Div"}});
```
SherlockNoMad, line 2535: These few lines looks like LogSoftmax to me... Don't we have a LogSoftmax onnx op for this ?

onnx/defs/math/defs.cc
```diff
@@ -2527,40 +2545,42 @@ bool BuildContextDependentFunctionBodySCE(
...
-                       MakeRefAttribute("ignore_index", AttributeProto::INT)}});
-    }
+  body.push_back({{"X_Log"}, "Log", {"X_Div"}});
+
+  // Review(mzs): Ideally we want to reuse the output from Log for sub-graph output as well but
+  // looking at the graph resolve code it does not include graph outputs as intermediate outputs, hence
+  // if intermediate X_log is renamed as log_prob then it will be treated as graph output and will not 
+  // be available to NegativeLogLikelihoodLoss. May be my understanding is incorrect or there is a bug in
+  // function population code in ORTbut I will dig further to be 100%.
+  // In the meantime we just replicate the log by recomputing it or may be copy if there is an op for that?
```
SherlockNoMad, line 2565: use an Identity op should sovle this 
log_prob = Identity(X_Log) ?

onnx/defs/math/defs.cc
codemzs: It also seemed that way to me but LogSoftmax does not seem to yield the same results.

onnx/defs/math/defs.cc
codemzs: Perfect, thanks!

codemzs:
@linkerzhang Can you please review one more time and see if your hold can be removed? I believe we have addressed your concerns. We want to ship ONNX 1.7 next and this PR is critical for that. I will really appreciate if you can review this PR at the earliest.

onnx/defs/nn/defs.cc
```diff
@@ -1811,9 +1811,17 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "If it's non-zero, output will be a random dropout of the scaled input, which is typically "
-            "the case during training.",
+            "the case during training. It is an optional value, if not specified it will default to 0.5.",
             "T1",
             OpSchema::Optional)
+        .Input(
+            2,
+            "training_mode",
+            "If set to true then it indicates dropout is being used for training. It is an optional value hence unless  "
+            "specified explicitly it is false and ratio will be zero in which case it will mimic inference mode where "
```
gramalingam, line 1821: Change to "... unless specified explicitly, it is false. If it is false, ratio is ignored and the operation mimics inference mode ..."?

onnx/defs/math/defs.cc
```diff
@@ -441,18 +472,8 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-                                         "Constant",
-                                         {},
-                                         {MakeRefAttribute(
-                                             "value_float",
-                                             "alpha",
-                                             AttributeProto::FLOAT)}},
-             {{"X_alpha"}, "Div", {"X", "alpha"}},
-             {{"Elu_Result"}, "Elu", {"X_alpha"}, {{"alpha", 1.f}}},
-             {{"Y"}, "Mul", {"alpha", "Elu_Result"}}})));
+        .SetContextDependentFunctionBodyBuilder(BuildContextDependentFunctionBodyCelu)
```
gramalingam, line 475: I thought the plan was to restrict the type to only float, is that right? Should line 442/473 be changed to allow T to be only tensor(float)?

onnx/defs/math/defs.cc
```diff
@@ -2527,40 +2545,42 @@ bool BuildContextDependentFunctionBodySCE(
...
-                      {MakeRefAttribute("reduction", AttributeProto::STRING),
-                       MakeRefAttribute("ignore_index", AttributeProto::INT)}});
-    }
+  body.push_back({{"X_Log"}, "Log", {"X_Div"}});
+
+  // Review(mzs): Ideally we want to reuse the output from Log for sub-graph output as well but
+  // looking at the graph resolve code it does not include graph outputs as intermediate outputs, hence
+  // if intermediate X_log is renamed as log_prob then it will be treated as graph output and will not 
+  // be available to NegativeLogLikelihoodLoss. May be my understanding is incorrect or there is a bug in
+  // function population code in ORTbut I will dig further to be 100%.
```
gramalingam, line 2564: I suspect it might be a bug in ORT function expansion code. E.g., something similar to what is done in ONNX might be required: see: https://github.com/onnx/onnx/blob/2dd6a1ba5fcdb1d47cd485d5033e6cf804100d49/onnx/defs/function.cc#L46

codemzs: Agreed, I will track this separately, spoke offline and @gramalingam is fine with the workaround for now.

gramalingam:
Any other feedback from anyone else, before this is merged? Thanks!

PR #2730, opened 2020-04-17
Azure Pipelines - Linux / MacOS CIs

vinitra:
In line with #2729, we are creating Azure pipelines for Linux, MacOS CI builds.

Checking in the YML files will enable us to run Azure pipelines (free for open source tools) in conjunction with current CIs (appveyor, travis, circleCI) without any breaking changes. Once we confirm everything is working well, we can have one CI format for all pipelines and deprecate alternative solutions.

Builds of Azure pipelines: https://dev.azure.com/onnx-pipelines/onnx/_build

Plan: Deprecate Travis builds and have Linux, MacOS, and Windows pipelines in the same CI service. This does not have as much of a time constraint as we are currently on Travis free tier.


vinitra:
> wondering where will the CI show on the page?

These CIs will become part of the PR checks / be included as badges in the READMEs. This cannot be implemented before the pipelines are enabled, which requires these YAMLs to be checked in first.

Issue #2614, opened 2020-02-20
ONNX release 1.7


linkerzhang:
Per ONNX community agreement, ONNX 1.7 release is being prepared. Key dates as below.
1.	PRs (for 1.7) merging due date – 2/25/2020.
2.	Release branch (for 1.7) creation date – 2/26/2020.
3.	1.7 release date – 3/2/2020.

Please active PR owners, mark your PR with “1.7 release” tag (in PR subject) and actively pushing the merge before the due date.

linkerzhang:
Hi ONNX partners, 
Due to some key PRs not merged yet, per the community agreement, ONNX 1.7 release will be pushed out for one week. Key dates updated as below,
a.	PRs (for 1.7) merging due date – 3/3/2020.
b.	Release branch (for 1.7) creation date – 3/4/2020.
c.	1.7 release date – 3/9/2020.

linkerzhang:
@chinhuang007 how do you think of tracking the issues left for 1.7 release here please? (it's a little bit hard to track issues in the gitter message room). Thank you!

codemzs:
### <div align="center">Issues that were fixed to ensure ONNX 1.7 ships with quality.</div>

| **Operator**  | **Offending PR**  | **Issue** | **Fix Description** | **Fix PR** |
|:-:|:-:|:-:|:-:|:-:|
| SoftmaxCrossEntropyLoss | https://github.com/onnx/onnx/pull/2573  | Incorrect label tensor type as int32/int64 in the spec instead of float. | Update the spec. | https://github.com/onnx/onnx/pull/2667 |
| SoftmaxCrossEntropyLoss | https://github.com/onnx/onnx/pull/2573  | Spec is missing `ignore_index` attribute, effectively rendering it useless for use with PyTorch models. Real world use case with HuugingFace model uses `ignore_index` equal to -1. | Update the spec, reference implementation, function body, tests.  | https://github.com/onnx/onnx/pull/2680 , https://github.com/onnx/onnx/pull/2703 , https://github.com/onnx/onnx/pull/2700, https://github.com/onnx/onnx/pull/2696 |
| SoftmaxCrossEntropyLoss | https://github.com/onnx/onnx/pull/2573  | Incorrect reference implementation.  | Fix bugs in reference implementation and handle 3-D, 4-D, and other cases.  | https://github.com/onnx/onnx/pull/2690 |
[...]

PR #2741, opened 2020-04-24
Move all training operators to a preview domain

wschin:
This PR creates a new ONNX domain, `ai.onnx.preview.training`, for storing training operators while ONNX Training is in preview. Because ONNX Training Spec is in preview and subject to change, it's better to put related operators to a domain which does not interact with other operators. They main purpose is to allow development around these complicated operators and avoid polluting other namespace.


wschin:
> Hmmm, the "experimental" op concept was removed because it's hard to tell onnx partners to not depend on it till it becomes production quality. That brings lots of troubles on the versioning maintenance.
> 
> I may suggest to remove these ops from the repo if we think they're not in spec quality, and we can add them back when ready.
> 
> @prasanthpul @ebarsoum @postrational

It's a chicken-and-egg problem. The community already agrees to put training spec in **preview**, this PR offers a clearer way for doing it.

gramalingam:
A nit: it would be better to use a symbolic constant like AI_ONNX_EXPERIMENT_DOMAIN instead of "ai.onnx.experiment" everywhere. (But it occurs in lots of places, so probably lot of work to fix it. So, I am also okay with leaving it as is.) If this is changed, I would recommend "experimental" instead of "experiment".

onnx/defs/operator_sets-experiment.h
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) ONNX Project Contributors.
```
gramalingam, line 1: Change file name to use underscores consistently (it is a mix of underscore and hyphen right now)?

wschin:
> A nit: it would be better to use a symbolic constant like AI_ONNX_EXPERIMENT_DOMAIN instead of "ai.onnx.experiment" everywhere. (But it occurs in lots of places, so probably lot of work to fix it. So, I am also okay with leaving it as is.) If this is changed, I would recommend "experimental" instead of "experiment".

It's not quick but it's done. The new domain name is `ai.onnx.experimental` now.

onnx/defs/operator_sets-experiment.h
wschin: Done.

gramalingam:
@linkerzhang : the purpose of the experimental domain is to enable an open discussion of an op's spec and to experiment with it before finalizing the spec. I agree there was an issue with it previously and it was removed. The previous problem was that some partners took a dependence on the experimental op. It has to be made clear that users/partners should not depend on experimental ops, as they are subject to changes and even deletion. ONNX implementations are under no obligation to implement or support experimental ops. This is partly motivated by a number of issues that showed up in ops that were introduced in the last cycle as we move towards the 1.7 release. To avoid such problems, it is necessary to iterate through ONNX spec development, its use (in producers, exporters) and implementation (in backends) and this experimental domain is intended to help with that. Does that make sense? There is an Operator sig meeting next week to discuss this proposal.

codemzs:
@gramalingam Fair point but my only concern with this approach is that experimental domain _could_ become a dumping ground for specs. Who will take the responsibility to move an op from experimental stage to production stage and if we can time bound this upgrade?  I tend to take a more conservative approach with these things and agree with@linkerzhang.

linkerzhang:
I agree with @codemzs and don't think it's a good idea to have another "experimental" concept in ONNX in any way. I am seeing two risks: 
1. How to ensure that no one is going to depend on these "experimental" stuff?
2. How to ensure that these "experimental" concepts will NOT be used as an area of putting not agreed stuff?

linkerzhang:
Adding more: are we going to allow onnx folks to add other "experimental domains" for adding their own ops?

linkerzhang:
> > Hmmm, the "experimental" op concept was removed because it's hard to tell onnx partners to not depend on it till it becomes production quality. That brings lots of troubles on the versioning maintenance.
> > I may suggest to remove these ops from the repo if we think they're not in spec quality, and we can add them back when ready.
> > @prasanthpul @ebarsoum @postrational
> 
> It's a chicken-and-egg problem. The community already agrees to put training spec in **preview**, this PR offers a clearer way for doing it.

@wschin I don't agree that it's a "chicken-egg" problem here. The "egg" (preview doc) does NOT have to be coded and pushed into this repo, instead, it could be a markdown doc only there for discussion (before it turns to be chicken :)). Let's keep the ONNX repo with the "standard spec" quality.

linkerzhang:
> @linkerzhang : the purpose of the experimental domain is to enable an open discussion of an op's spec and to experiment with it before finalizing the spec. I agree there was an issue with it previously and it was removed. The previous problem was that some partners took a dependence on the experimental op. It has to be made clear that users/partners should not depend on experimental ops, as they are subject to changes and even deletion. ONNX implementations are under no obligation to implement or support experimental ops. This is partly motivated by a number of issues that showed up in ops that were introduced in the last cycle as we move towards the 1.7 release. To avoid such problems, it is necessary to iterate through ONNX spec development, its use (in producers, exporters) and implementation (in backends) and this experimental domain is intended to help with that. Does that make sense? There is an Operator sig meeting next week to discuss this proposal.

No. ONNX development should not and does not have to be in this repo. Taking ONNX runtime as an example, it's very straightforward to have all ops registered and implemented in ONNX runtime firstly and verified with releases, and then proposing them to ONNX community. Taking ONNX repo as an experimental place is NOT a good idea, which was tried and proved before.

codemzs:
If someone wants to implement an _experimental_ spec (saved as a MD file as @linkerzhang suggested and I agree) then they can't they use the exact same strategy we have in place for ONNX Runtime Training where we define our own schema definition file locally as CONTRIB, see [gradient_schema_defs.cc ](https://github.com/microsoft/onnxruntime/blob/master/orttraining/orttraining/core/graph/gradient_schema_defs.cc)? We have so many gradient operators and not even one of those operators has a spec checked into ONNX, no?

As we make this important decision on the new policy to check-in specs into ONNX, I urge all of us to please also take a look at number of [issues ](https://github.com/onnx/onnx/issues/2614#issuecomment-619508755) we had to investigate and fix so that ONNX 1.7 can be released with quality. These issues that were found last minute delayed the release and were a direct result of checking-in specs that did not have a corresponding runtime implementation, inaccurate reference implementation, inadequate tests and were not validated to appropriate quality levels.

wschin:
> > @linkerzhang : the purpose of the experimental domain is to enable an open discussion of an op's spec and to experiment with it before finalizing the spec. I agree there was an issue with it previously and it was removed. The previous problem was that some partners took a dependence on the experimental op. It has to be made clear that users/partners should not depend on experimental ops, as they are subject to changes and even deletion. ONNX implementations are under no obligation to implement or support experimental ops. This is partly motivated by a number of issues that showed up in ops that were introduced in the last cycle as we move towards the 1.7 release. To avoid such problems, it is necessary to iterate through ONNX spec development, its use (in producers, exporters) and implementation (in backends) and this experimental domain is intended to help with that. Does that make sense? There is an Operator sig meeting next week to discuss this proposal.
> 
> No. ONNX development should not and does not have to be in this repo. Taking ONNX runtime as an example, it's very straightforward to have all ops registered and implemented in ONNX runtime firstly and verified with releases, and then proposing them to ONNX community. Taking ONNX repo as an experimental place is NOT a good idea, which was tried and proved before.

It's true for a single operator. However, training is a large framework level change, which requires much more resources/time/discussion/development-iteration. It involves exporter, runtime, spec. Are we able to do all the three things in the same time? In addition, even Pytorch doesn't have full support of auto-grad; they can do only a very limited part of it. Do you consider Pytorch should not release auto-grad APIs?

wschin:
> If someone wants to implement an _experimental_ spec (saved as a MD file as @linkerzhang suggested and I agree) then can't they use the exact same strategy we have in place for ONNX Runtime Training where we define our own schema definition file locally as CONTRIB, see [gradient_schema_defs.cc ](https://github.com/microsoft/onnxruntime/blob/master/orttraining/orttraining/core/graph/gradient_schema_defs.cc)? We have so many gradient operators and not even one of those operators has a spec checked into ONNX, no?
> 
> As we make this important decision on the new policy to check-in specs into ONNX, I urge all of us to please also take a look at number of [issues ](https://github.com/onnx/onnx/issues/2614#issuecomment-619508755) we had to investigate and fix so that ONNX 1.7 can be released with quality. These issues were found last minute, delayed the release and were a direct result of checking-in specs that did not have a corresponding runtime implementation, inaccurate reference implementation, inadequate tests and were not validated to appropriate quality levels.

They CANNOT implement the spec if we don't write it in C++. How do you make sure everyone translate English into the same C++ code? I am aware of the pain of ONNX 1.7 release and I am also very aware of the year-long pain of creating a training spec where everyone had their opinion. If we don't freeze the agreement formally here. I don't know who will be willing to do this again.

codemzs:
What do you mean? How are all the gradient operator spec implemented in CONTRIB? Where is the formal C++ spec for them? these don’t even have a MD file for _experimental_ discussion.

People can implement the spec privately, experiment by implementing the op in some
ONNX RT and once they have sufficient confidence then they submit to ONNX.

wschin:
> What do you mean? How are all the gradient operator spec implemented in CONTRIB? Where is the formal C++ spec for them? these don’t even have a MD file for _experimental_ discussion.
> 
> People can implement the spec privately, experiment by implementing the op in some
> ONNX RT and once they have sufficient confidence then they submit to ONNX.

@ebarsoum already said in our meeting he will have a document for that experimental domain.

codemzs:
Which meeting?

ebarsoum:
Guys, we need a staging domain to iterates on operators and gather feedbacks. This is different than the old experimental, which had FX specific operators that we needed to support.

@linkerzhang we will document that this is only to gather feedbacks and iterate on the OPs. Once it is ready and implemented in a runtime, then we can move it to core.

codemzs:
@ebarsoum @wschin Can this _staging_ spec not be in a MD file, essentially all the same syntax but in a MD file?

What advantage do we get by having it as a code in CC file and then we are creating this additional experimental domain?

If someone wants to implement they can copy the content and implement locally just like we do in [gradient_schema_defs.cc ](https://github.com/microsoft/onnxruntime/blob/master/orttraining/orttraining/core/graph/gradient_schema_defs.cc)?

chinhuang007:
If we are introducing this new 'staging domain' in release 1.7, I think we need to clearly articulate the motivation and what to expect in terms of quality, readiness, testing, docs, etc for the ops to be included in this domain so that the broader community knows exactly what to expect and what to do with the staging domain ops. Hope this will serve ONNX community better than the "experimental  ops" experience.

linkerzhang:
@wschin I may mislead you. It's very critical to have training support in ONNX. The work that have been done are very appreciated. 

This discussion now is NOT saying to revert all "training" related stuff from ONNX, but only the "OPs" which folks think are not ready (and need more iterations). All the IR related stuff are good enough with good quality.

Even for "OPs", not all operators have to be reverted, but those need more iterations/changes.

@ebarsoum @gramalingam The big reason that another kind of "experimental" area for "OPs" should not be created is that,
1. what's the bar of the quality? how do you want to control it? it's not only targeting for this specific "training" scenario, right?
2. what's the expectation of "versioning"? 
3. what's the timeline of either promoting them or removing them from ONNX?

This 3 questions were met previously, and obviously it's hard to give answers to them.

Meanwhile, after all IR related stuff introduced into ONNX (for training), "OPs" should NOT be a blocking issue for any ONNX partner (say, ORT) to try more on ONNX training. As @codemzs mentioned, there're already many "contrib" ops in ORT, and they're being verified by ORT in real biz scenarios.

ebarsoum:
@linkerzhang  and @chinhuang007 can you attend next operator meeting (This Wednesday)? 

1. The quality if experimental domain is the same as other ONNX operator currently, the bar isn't lower with . With the exception, if an OP isn't yet implemented by any framwork. This isn't only for training.
2. Experimental OPs aren't part of the release, version will be there for tracking. We will be explicit about it, no backward compatibility, it is just a preview to gather feedbacks and iterate on the OPs. If we have FooV5 in experimental and decided to promote it to core, then it will move to core as FooV1 and we will delete FooV5.
3. That is a great questions, those operators will be reviewed by the operators SIG, and operator SIG will ping people if there is no traction on some OPs. Operator SIG will decide when to remove the operators.

We will add a big banner in the md file that those OPs are for feedback only from framework developers, and they aren't part of the spec.

linkerzhang:
> @linkerzhang and @chinhuang007 can you attend next operator meeting (This Wednesday)?
> 
> 1. The quality if experimental domain is the same as other ONNX operator currently, the bar isn't lower with . With the exception, if an OP isn't yet implemented by any framwork. This isn't only for training.
> 2. Experimental OPs aren't part of the release, version will be there for tracking. We will be explicit about it, no backward compatibility, it is just a preview to gather feedbacks and iterate on the OPs. If we have FooV5 in experimental and decided to promote it to core, then it will move to core as FooV1 and we will delete FooV5.
> 3. That is a great questions, those operators will be reviewed by the operators SIG, and operator SIG will ping people if there is no traction on some OPs. Operator SIG will decide when to remove the operators.
> 
> We will add a big banner in the md file that those OPs are for feedback only from framework developers, and they aren't part of the spec.

Haha, what time will be meeting hosted? I'll skip it if it's a mid-night time slot (after 4PM saves my life if it's doable :))

@ebarsoum, Sir, I still don't see a strong reason that "OPs" needs to be previewed, especially needs to be previewed with "c++" changes in ONNX repo.

Even stepping back a little bit, if there're a group of "OPs" which need to be previewed for discussion, then a separate .md file created and merged in ONNX repo should be good enough. Say, operators_for_preview.md.

If "Preview" ops will be promoted to "production" ops with v1 again, what's the point of tracking their versions in "Preview"? are we expecting partners will play with "Preview" ops with versioning? if yes, that will be very bad, as when "preview" ops promoted with V1 later, troubles will be introduced to partners (depending on Preview versions), if not, then no need having "versioning" for Preview ops (this is again why I think a .md document is good enough for preview instead of "c++" changes).

chinhuang007:
@ebarsoum Sure, I will attend the Operators SIG meeting this Wed to understand this new "staging domain" concept.

wschin:
We cannot stage spec in MD. It creates spec even without trying to compile it. If company A creates a operator in MD while company B tries to implement it, do you suggest company B to translate English or whatever in MD into their C++ code? Would it be possible that the C++ code in A's mind is always identical to the C++ code in B's mind? Most machine learning software has preview APIs. I don't see any problem with a similar concept in ONNX. I understand you worry a lot about `previous experimental` ops. Please let me share an observation: Microsoft encouraged the uses of experimental ops, created a lot of models with experimental ops, and finally felt painful when dropping them. The problem comes from the **promotion** of experimental ops inside Microsoft, not creating some experimental ops.

linkerzhang:
> We cannot stage spec in MD. It creates spec even without trying to compile it. If company A creates a operator in MD while company B tries to implement it, do you suggest company B to translate English or whatever in MD into their C++ code? Would it be possible that the C++ code in A's mind is always identical to the C++ code in B's mind? Most machine learning software has preview APIs, where `preview` means that this is the best setting given our limited knowledge and please join us to make it better. I don't see any problem with a similar concept in ONNX. I understand you worry a lot about `previous experimental` ops. Please let me share an observation: Microsoft encouraged the uses of experimental ops in production, created a lot of models with experimental ops, and finally felt painful when dropping them. The problem comes from the **promotion** of experimental ops inside Microsoft, not creating some experimental ops.
> 
> As for recent bugs in ONNX-1.7, I think they occur because some authors didn't fully follow the stages in proposing new operators while bad reviewers like me were not perfectly doing their job. If everyone follow the rules from now on, we will have a better world. But, again, the problem does not come from a domain of `ai.onnx.experimental`.

Sorry that I am still not convinced. Take ORT contrib ops as an example, there're bunch of ops being created/registered/verified in ORT without touching ONNX spec or experimenting/staging in ONNX spec. 

The point of having this "Preview" is not about "compilation", is about showing it to public for discussion, right? That's why I was saying a markdown file is good enough for this purpose. That means, internally, for a partner, say, ORT again, the ops' codes (registration codes) may still follow the same way as other "contrib" ops.

wschin:
> > We cannot stage spec in MD. It creates spec even without trying to compile it. If company A creates a operator in MD while company B tries to implement it, do you suggest company B to translate English or whatever in MD into their C++ code? Would it be possible that the C++ code in A's mind is always identical to the C++ code in B's mind? Most machine learning software has preview APIs, where `preview` means that this is the best setting given our limited knowledge and please join us to make it better. I don't see any problem with a similar concept in ONNX. I understand you worry a lot about `previous experimental` ops. Please let me share an observation: Microsoft encouraged the uses of experimental ops in production, created a lot of models with experimental ops, and finally felt painful when dropping them. The problem comes from the **promotion** of experimental ops inside Microsoft, not creating some experimental ops.
> > As for recent bugs in ONNX-1.7, I think they occur because some authors didn't fully follow the stages in proposing new operators while bad reviewers like me were not perfectly doing their job. If everyone follow the rules from now on, we will have a better world. But, again, the problem does not come from a domain of `ai.onnx.experimental`.
> 
> Sorry that I am still not convinced. Take ORT contrib ops as an example, there're bunch of ops being created/registered/verified in ORT without touching ONNX spec or experimenting/staging in ONNX spec.
> 
> The point of having this "Preview" is not about "compilation", is about showing it to public for discussion, right? That's why I was saying a markdown file is good enough for this purpose. That means, internally, for a partner, say, ORT again, the ops' codes (registration codes) may still follow the same way as other "contrib" ops.

They are not contrib ops. They are decisions made by Training WG. A full training spec combines all training ops and ONNX format changes --- no one should be missed. Do you suggest Training WG use ORT to store their progress? There are many cases we need to write C++ spec to do something. Can MD solve my concern on cross-company/team development and ambiguity of text? For your concerns, it's really up to how we use this experimental domain. They may or may not depending on our investment to ONNX.

codemzs:
@wschin Thanks, I think your concerns are reasonable but they don’t warrant a new domain. We can start with MD file and progress towards a domain if needed. Directly jumping into a new domain few days before a release is risky and seems reckless especially since the release was already delayed due to quality issues.

wschin:
> @wschin Thanks, I think your concerns are reasonable but they don’t warrant a new domain. We can start with MD file and progress towards a domain if needed. Directly jumping into a new domain few days before a release is risky and seems reckless especially since the release was already delayed due to quality issues.

What you said is not very different than Ke suggestion. I have repeated many times. No, MD is not enough given we have finished their C++ spec (if Training WG already goes beyond MD stage, why do they need to move back?). I don't think new domain is a problem. We originally have a `ai.onnx.training` domain for those training ops. It works just fine and it's not a change we can't affort.

codemzs:
@wschin Your C++ argument doesn’t seem strong. Much of world’s planning happens in plain text format. Please refer to C++ IR discussions in the community, they don’t introduce “experimental” syntax, it’s all in pdfs and emails. Please don’t try to reinvent the wheel.

wschin:
> @wschin Your C++ argument doesn’t seem strong. Much of world’s planning happens in plain text format. Please refer to C++ standard discussions in the community, they don’t introduce “experimental” syntax, it’s all in pdfs and emails. Please don’t try to reinvent the wheel.
> 
> You also missed my point on risking the release. I’m seriously concerned.

It's not about planning. It's a decision went through year-long discussion and must be stored in a most solid way and their next working items will have to use C++. It's not necessary for us to follow C++ community. Pytorch and Tensorflow have experimental APIs. Why can't we follow them? Taking conservative step is fine for traditional software. For deep learning, we need to move fast and iterate on it.

codemzs:
@wschin Deep learning is slowly turning into a programming languages problem hence C++ was a more relevant example. This idea of a experimental domain only popped up recently, it needs more review.

wschin:
> @wschin Deep learning is slowly turning into a programming languages problem hence C++ was a more relevant example. This idea of a experimental domain only popped up recently, it needs more review.

Deep learning becoming programming language doesn't stop TF or Pytorch from using experimental APIs.

We can restrict the uses of experimental domain. It's under Operator SIG's control.

gramalingam:
A quick question: would it help to distinguish "master releases" from other "experimental branches"? In my opinion, the important question is not whether we use text or code, but whether it is part of the standard or not. I think there is clear agreement (from both sides) that the "experimental" part is not part of the "standard". The question then becomes what is the best, unambiguous way, to clarify that something is not part of the standard. Is the "experimental domain" sufficient to clarify it is not part of the standard? If not, should we exclude it from the "official master release" and have it only in other branches?

gramalingam:
(Just to clarify the previous comment, the above decision also has implications for including "training as preview" in the master-release.)

codemzs:
@gramalingam The master branch should only contain standard. May be you can keep experimental stuff in an experimental branch. Fair?

gramalingam:
@Emad's suggestion "We will add a big banner in the md file that those OPs are for feedback only from framework developers, and they aren't part of the spec." is also one way to emphasize the distinction. We could take it one step further and make experimental.md a different file to further emphasize it is not part of the standard. It seems the question is about how heavyweight or lightweight a mechanism we want to keep the distinction between the standard and experimental clear.

codemzs:
@gramalingam I’m aware of the banner comment. Let’s start with an experimental branch and move up if needed.

gramalingam:
Speaking of C++ standards: this is what https://isocpp.org/std/status says:
```
Starting in 2012, the committee has transitioned to a “decoupled” model where major pieces of
work can progress independently from the Standard itself and be delivered in “feature branch” TSes.
Vendors can choose to implement these, and the community can gain experience with the
std::experimental version of each feature. This lets us learn and adjust each feature’s design based
on experience before it is cast in stone when merged into the “trunk” C++ Standard itself. In the
meantime, the Standard can be delivered on a more regular cadence with smaller and more
predictable batches of features. This approach also helps C++ compilers to track the Standard more
closely and add both the experimental and the draft-final C++ features in a more consistent order.
```

wschin:
> Speaking of C++ standards: this is what https://isocpp.org/std/status says:
> 
> ```
> Starting in 2012, the committee has transitioned to a “decoupled” model where major pieces of
> work can progress independently from the Standard itself and be delivered in “feature branch”
> TSes. Vendors can choose to implement these, and the community can gain experience with the
> std::experimental version of each feature. This lets us learn and adjust each feature’s design
> based on experience before it is cast in stone when merged into the “trunk” C++ Standard
> itself. In the meantime, the Standard can be delivered on a more regular cadence with smaller
> and more predictable batches of features. This approach also helps C++ compilers to track the
> Standard more closely and add both the experimental and the draft-final C++ features in a
> more consistent order.
> ```

I guess this reflects one thing: even C++ has a stage to make progress on features in preview. Such a preview stage is natural to community-based developments. `std::experimental` sounds a minor of `ai.onnx.experimental`.

codemzs:
@wschin First this is recent, two, it says preview features are in its own feature branch which is what I suggested previously. No one is arguing it is bad to have preview features, just not in the same place as production feature.

PR #2743, opened 2020-04-25
Add versions to operators in table of contents

natke:
Added operator and function versions to the Operator table of contents.


codemzs:
I feel like this might be a good change to take for ONNX 1.7 release. @natke can you please fix build errors soon if possible? We plans cut the release early next week preferably on Monday.

natke:
@codemzs wrote:
> I feel like this might be a good change to take for ONNX 1.7 release. @natke can you please fix build errors soon if possible? We plans cut the release early next week preferably on Monday.

Sure, will do! On first glance, as someone who is more familiar with the build system, do you have any pointers on what is causing the break?

gramalingam:
One possibility for the build-break is that operators-ml.md also needs to be regenerated and checked in. I see only operators.md updated. (This might require setting environment variable ONNX_ML before document generation, I don't remember exact details.)

wschin:
To regenerate all documents under Windows + Anaconda, you can run
```
F:\repos\onnx> .\tools\update_doc.bat
```

codemzs:
@natke Sorry for the delay in my reply, I almost missed your message amid ONNX 1.7/ORT 1.3 release but below are steps I follow:

From the root do:

1. Run tools\update_doc.bat 
    - After you run this command you will see bunch files being modified, that is because the command regenerates all MD, ONNX and PB files. ONNX files are generated from test files (in python) and placed, these files contain the model, PB files are also generated from test files but these contain input to the test and expected output from the test that comes from the reference implementation from the python test files. Both ONNX and PB files are placed [here](https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/node) and reference implementation and test files can be found [here](https://github.com/onnx/onnx/tree/master/onnx/backend/test/case/node).
    - Our test runners on ORT side, open these test files, execute the model (ONNX file) with inputs (PB file) and compare the output with the PB (output) files and to make sure results from ORT kernel matches reference implementation, which is why it is important to have the reference implementation correct.
2. git add *.md
3. git add */*softmax_cross_entropy_* [Optional, do only if you have modified any python file that may affect test, the example depicts softmax_cross_entropy related changes, you will need to repeat this step for all sorts of files you want to keep]
4. git checkout */*.onnx */*.json */*.pb
5. git checkout *.proto*
6. git status
7. git add any files that you may modified like *.cc files that contain spec or *.py that may contain test/reference implementation changes and make sure the updated files are also added.

I have also forked your branch and updated the MD files and pushed them at [this ](https://github.com/codemzs/onnx/tree/add-version-to-ops)branch in my remote, I wanted to push them in your branch but got 403 access denied.

Would it be possible to update this branch soon so that we can take for ONNX 1.7 release? Thanks!

codemzs:
> One possibility for the build-break is that operators-ml.md also needs to be regenerated and checked in. I see only operators.md updated. (This might require setting environment variable ONNX_ML before document generation, I don't remember exact details.)

I don't think you need to since _update.bat/sh_ does that for you, see below output from running it:
Using e:\continuum\anaconda2\envs\onnx\lib\site-packages
Finished processing dependencies for onnx==1.7.0

(onnx) E:\github\onnx>python onnx\backend\test\cmd_tools.py generate-data

(onnx) E:\github\onnx>python onnx\backend\test\stat_coverage.py

(onnx) E:\github\onnx>python onnx\defs\gen_doc.py

(onnx) E:\github\onnx>set ONNX_ML=0

(onnx) E:\github\onnx>python onnx\defs\gen_doc.py

(onnx) E:\github\onnx>set ONNX_ML=1

natke:
@codemzs, thank you for the instructions! I've pulled the latest change into this branch and rebuilt, but it looks like the build is stalled. Any ideas on this one?

natke:
@codemzs I've updated to the latest and re-ran the build. A couple of the tests are timing out - any ideas for that?

PR #2744, opened 2020-04-25
Catch >2GB models in checker

jacky82226:
Try to enable full_check


snnn:
See #2668

1. If you enable the check, then one of the CI build will fail. But there is nothing wrong with ONNX. PyTorch should be fixed. So you should get familiar with PyTorch codebase and fix their bug first. 

2. Then you can go back enable the check. However,  it's not just as easy as turn on a flag. The shape inference context implementation itself ignore exceptions. So even when there are inference errors,  they'll be silently ignored. see: https://github.com/onnx/onnx/blob/master/onnx/shape_inference/implementation.cc#L221

So, the flag you are changing is indeed very deceptive.

jacky82226:
> See #2668
> 
> 1. If you enable the check, then one of the CI build will fail. But there is nothing wrong with ONNX. PyTorch should be fixed. So you should get familiar with PyTorch codebase and fix their bug first.
> 2. Then you can go back enable the check. However,  it's not just as easy as turn on a flag. The shape inference context implementation itself ignore exceptions. So even when there are inference errors,  they'll be silently ignored. see: https://github.com/onnx/onnx/blob/master/onnx/shape_inference/implementation.cc#L221
> 
> So, the flag you are changing is indeed very deceptive.

Hi @snnn, I am trying to make the onnx checker stricter so I set the flag first to see whether it could pass the CI. I will improve it according to your ideas. Thank you for the helpful instructions!

onnx/defs/schema.cc
```diff
@@ -82,6 +85,12 @@ void OpSchema::CheckInputOutputType(struct InferenceContext& ctx) const {
         !all_types.empty() &&
         all_types.find(Utils::DataTypeUtils::ToType(*param_type)) ==
             all_types.end()) {
+      cout << "catch!" << endl;
```
gramalingam, line 88: Is this meant to be for debugging?

jacky82226: Yes. Sorry that it will be removed. The current version is still in progress.

jacky82226:
After enabling full_check, it encounters the pytorch test error: Type consistency error
Might need this [PR](https://github.com/pytorch/pytorch/pull/37787) to fix the test failure in Pytorch.

jacky82226:
### Limitations
1. Protobuf file has a hard limit of 2GB
2. After loading models, onnx cannot know the original path of the loaded model
### Facts
1. Passing protobuf>2GB (model) with loaded raw_data will lose some infomration to onnx-checker
2. Passing protobuf>2GB (model) without loaded raw_data will not pass onnx-checker 
3. Passing protobuf<2GB (model) with loaded raw_data works well to C++ checker
4. Passing model path works well to C++ checker 
### Proposals
1. Add one more data_field in ModelProto: model_path (Prevent to modify the interface of onnx.load_model)
2. After loading those external tensors to raw_data, correct the data_location from external_data to DEFAULT.
3. For those protobuf>2GB, onnx should provide model_path to onnx-checker instead of onnx model itself. (However, I think detecting whether the model is > 2GB might be a little messy. How about we provide model_path to onnx-checker for **those models with external tensors** ?)
### Drawbacks
1. Hard to maintain tests on various platforms like the path issue. (\/ vs \\) Current solution is making the path even to various platform for matching testing results
2. Will adding one more field compromise some tests in ModelProto?
### TO-DOs:
1. Decide when to use model_path instead of model for onnx-checker
2. Add tests for checking large model (>2GB with external tensors)
3. Organize the commit history
### References
ONNX https://github.com/onnx/onnx/pull/678
ORT https://github.com/microsoft/onnxruntime/pull/520

jacky82226:
I made some modification as POC. @gramalingam Please help me to review the proposal. Thank you.

gramalingam:
It seems reasonable to add model_path to the python object representing a model. But adding it to the proto seems unnecessary. I don't think we should pollute the proto definition with features/fields needed only in a Python or C++ utility function. If the protoc-generated python-class/object doesn't allow addition of new fields/attributes, we should be able to workaround that, I think.

jacky82226:
> It seems reasonable to add model_path to the python object representing a model. But adding it to the proto seems unnecessary. I don't think we should pollute the proto definition with features/fields needed only in a Python or C++ utility function. If the protoc-generated python-class/object doesn't allow addition of new fields/attributes, we should be able to workaround that, I think.

Agree that we should not pollute the proto definition so I did tried it before. However, like you mentioned, protoc-generated python-class/object doesn't allow addition of new fields/attributes. A possible solution would be introducing a new model object in Python including the original model proto and filepath, but it will change the return type of load_model and influence the usage.

jacky82226:
Since the modification for enabling two functionality might conflict, it would be better to decouple different tasks from this PR. 
This PR focus on supporting large models for onnx-checker. Catching inference error has been moved to anther PR https://github.com/onnx/onnx/pull/2783.

onnx/checker.py
```diff
@@ -24,8 +24,13 @@
 from typing import TypeVar, Callable, Any, Type, cast, Union, Text
 from six import string_types
 import onnx.shape_inference
+import sys
+import warnings
 
 
+# Limitation of single protobuf file is 2GB
+MAXIMUM_PROTOBUF = 1900000000  # loose boundary
```
gramalingam, line 32: Why do we need a "loose" boundary? Can we define it exactly?

onnx/checker.py
```diff
@@ -90,6 +95,10 @@ def check_model(model, full_check=False):  # type: (Union[ModelProto, Text], boo
         C.check_model_path(model)
         m = onnx.load(model)
     else:
+        # If the protobuf is larger than 2GB,
+        # remind user should use the model path to check
+        if sys.getsizeof(model.SerializeToString()) > MAXIMUM_PROTOBUF:
```
gramalingam, line 100: Better to avoid serializing the model twice.

jacky82226: Good catch.

onnx/checker.py
jacky82226: I was afraid that the function is approximate. Just did some experiment show the calculation is precise enough. Should set 2GB for the readability. Thank you for catching this.

onnx/checker.py
```diff
@@ -90,7 +95,12 @@ def check_model(model, full_check=False):  # type: (Union[ModelProto, Text], boo
         C.check_model_path(model)
         m = onnx.load(model)
     else:
-        C.check_model(model.SerializeToString())
+        # If the protobuf is larger than 2GB,
+        # remind user should use the model path to check
+        protobuf_string = model.SerializeToString()
+        if sys.getsizeof(protobuf_string) > MAXIMUM_PROTOBUF:
+            warnings.warn(str('This protobuf of onnx model is too large (>2GB). Call check_model with model path instead.'))
```
gramalingam, line 102: Does warn throw an exception? Wouldn't it be better for the check_model to be in the else branch?

jacky82226: Warn just throws a message without interrupting the code. 
Sounds good. Models larger than 2gb won't pass the checker anyway so it is no need to run the checker.

jacky82226:
### Description

- Throw an error if the model is larger than 2GB in onnx.checker
- Fix a load_external_tensor bug (change into DEFAULT after loading)
- Add tests for 2GB models and refactor original test cases

### Motivation and Context
Current onnx.checker doest not support models larger than 2GB because they exceed the maximum of protobuf. By contrast, C++ checker works well with models with external tensors.
Therefore, when the model is larger than 2GB, users should use onnx.checker with model path instead of the model proto.

jacky82226:
I will add some documents for using load_external_data in this PR

jacky82226:
- Add a ExternalData.md to tell users how to deal with external data
- Add example code in PythonAPIOverview
- Add a Link to ExternalData.md in IR.md

PR #2674, opened 2020-03-20
Update Batchnorm test

lara-hdr:
Updating batchnorm tests to succed with ORT backend.

PR #2750, opened 2020-04-26
Revert BatchNormalization(12)

codemzs:
This change reverts BatchNormalization(12) spec (below commits) because it is not fully verified to work as it should.

Surgically removed batch norm specific changes from #2568
#2674

Unfortunately, this spec was one among the few specs that was checked in a rush between 02/2020 and 03/2020. Its presence has jeopardized ONNX 1.7 release but removing it will give us appropriate confidence level in the quality for ONNX 1.7 release.


codemzs:
This PR depends on https://github.com/pytorch/pytorch/pull/37309 that is causing CI failure with Caffe2 tests not being able to find BatchNorm(12)

PR #2752, opened 2020-04-27
Update ORT version

neginraoof:


neginraoof:
@houseroad Since ORT tests are not running on CI, do we still need to update ORT version?

PR #2755, opened 2020-04-30
Update ONNX build instructions for readability, ARM 64 instructions

vinitra:
ONNX Runtime is planning to publish an ARM 64 package. As it takes a dependency on ONNX, the developer experience might be broken if the ARM version of ONNX is not easy to install. In lieu of a ARM 64 binary for ONNX for the 1.7 release, I'm including some build instructions in the README. 

Additionally, adding some more detailed instructions on common errors / build procedures and fixing some minor grammatical inconsistencies for readability.


snnn:
BTW, currently, building from onnx on Ubuntu is ok. But it doesn't work on CentOS/RHEL/manylinux/... . 
On these systems, you need to open:
https://github.com/onnx/onnx/blob/master/CMakeLists.txt#L124
and search replace all the "/lib" to "/lib64".

snnn:
Similar, if you need debug version of ONNX, search things like "NAMES protobuf-lite", append a letter 'd' at the end.

vinitra:
> BTW, currently, building from onnx on Ubuntu is ok. But it doesn't work on CentOS/RHEL/manylinux/... .
> On these systems, you need to open:
> https://github.com/onnx/onnx/blob/master/CMakeLists.txt#L124
> and search replace all the "/lib" to "/lib64".

I've included these instructions in the README. For now, I'm not including the Debug instructions because it's not used as frequently. Thanks!

wschin:
> > BTW, currently, building from onnx on Ubuntu is ok. But it doesn't work on CentOS/RHEL/manylinux/... .
> > On these systems, you need to open:
> > https://github.com/onnx/onnx/blob/master/CMakeLists.txt#L124
> > and search replace all the "/lib" to "/lib64".
> 
> I've included these instructions in the README. For now, I'm not including the Debug instructions because it's not used as frequently. Thanks!

Including Debug instructions helps new devs working on this topic. It'd be great to have it in this PR. Thank you.

gramalingam:
BTW, I am curious about the instructions for building protobuf from source for a Windows installation. I use protobuf installed from conda, and it works for me. However, I had to install a specific version of protobuf (=3.5.1) from conda. Does anyone have more concrete information? May be it will help to suggest that people installing from conda should use version 3.5.1 (even though we may officially recommend building protobuf from source)?

snnn:
> BTW, I am curious about the instructions for building protobuf from source for a Windows installation. I use protobuf installed from conda, and it works for me. However, I had to install a specific version of protobuf (=3.5.1) from conda. Does anyone have more concrete information? May be it will help to suggest that people installing from conda should use version 3.5.1 (even though we may officially recommend building protobuf from source)?

There are 3 special variables:
1. Env variable: USE_MSVC_STATIC_RUNTIME
2. Cmake variables: ONNX_USE_PROTOBUF_SHARED_LIBS -DProtobuf_USE_STATIC_LIBS
You must set all of these 3 correctly. 

For example, one possible combination is:
```
$Env:USE_MSVC_STATIC_RUNTIME=1
$Env:CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=OFF -DProtobuf_USE_STATIC_LIBS=ON
```
If you protobuf was built from source with static link to CRT and doesn't have an DLL.

If you got the libprotobuf from conda, you need to check how the conda package was built.  The package you referred changed its build command around that time(3.5.1).  

Most people don't know these things, they just use the default setting. They didn't think it's nesessary to set the 3 variables manually. 

And, all of the above are based on an assumption that you're using the latest cmake. Otherwise, depending on which cmake version you have, the instruction may differ.

vinitra:
Based on @snnn's instructions above and @jacky82226's experience with installing on Windows, I've aligned the "Build with Anaconda on Windows" instructions with the new Azure Pipelines Windows CI.

wschin:
@vinitra, can you add @snnn's explanation to `common error` section? I mean this:
```
Cmake variables: ONNX_USE_PROTOBUF_SHARED_LIBS Protobuf_USE_STATIC_LIBS
You must set all of these 3 correctly. If ONNX_USE_PROTOBUF_SHARED_LIBS is ON, then Protobuf_USE_STATIC_LIBS must be off and USE_MSVC_STATIC_RUNTIME must be 0. If ONNX_USE_PROTOBUF_SHARED_LIBS is OFF then Protobuf_USE_STATIC_LIBS must be ON and USE_MSVC_STATIC_RUNTIME can be 1 or 0.
```

vinitra:
> @vinitra, can you add @snnn's explanation to `common error` section? I mean this:
> 
> ```
> Cmake variables: ONNX_USE_PROTOBUF_SHARED_LIBS Protobuf_USE_STATIC_LIBS
> You must set all of these 3 correctly. If ONNX_USE_PROTOBUF_SHARED_LIBS is ON, then Protobuf_USE_STATIC_LIBS must be off and USE_MSVC_STATIC_RUNTIME must be 0. If ONNX_USE_PROTOBUF_SHARED_LIBS is OFF then Protobuf_USE_STATIC_LIBS must be ON and USE_MSVC_STATIC_RUNTIME can be 1 or 0.
> ```

Included!

vinitra:
> ## Binaries
> A binary build of ONNX is available from [Conda](https://conda.io), in [conda-forge](https://conda-forge.org/):
> ```
> conda install -c conda-forge onnx
> ```

While trying these instructions, I noticed the onnx version was only 1.1.1 in a Python 3.6 environment. Should we be updating the binary release with new releases? Does it make sense to remove this if it's not up to date?

wschin:
> > ## Binaries
> > A binary build of ONNX is available from [Conda](https://conda.io), in [conda-forge](https://conda-forge.org/):
> > ```
> > conda install -c conda-forge onnx
> > ```
> 
> While trying these instructions, I noticed the onnx version was only 1.1.1 in a Python 3.6 environment. Should we be updating the binary release with new releases? Does it make sense to remove this if it's not up to date?

ONNX on Anaconda is not up-to-date. Someone needs to do the work. You can create an issue for tracking it. I guess it can be prioritized based on demand.

PR #2756, opened 2020-05-01
Control build paralllelism via `--parallel` option

malfet:
When building with python3 respect setuptools `--parallel` option, i.e.:
`python3 setup.py build --parallel=4` should not spawn more than 4 compiler jobs at once
Also, if neither of the options is specified, check `MAX_JOBS` environment variable

Test Plan:  `MAX_JOBS=4 pip3 install  git+https://github.com/malfet/onnx@follow-parallel-build-option` and see how many jobs are spawned

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -139,10 +139,14 @@ class cmake_build(setuptools.Command):
     built = False
 
     def initialize_options(self):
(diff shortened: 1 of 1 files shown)


orionr:
@houseroad are you able to help here? Thanks.

PR #2761, opened 2020-05-05
Avoid experimental word also internally

wschin:
Rename internal `experimental` names to `preview`. This PR does not affect spec. It only changes internal things.


onnx/defs/operator_sets_experiment.h
```diff
@@ -1,38 +0,0 @@
-// Copyright (c) ONNX Project Contributors.
```
wschin, line 1: This file is renamed to operator_sets_preview.h.

PR #2764, opened 2020-05-06
Release 1.7 verification 105

chinhuang007:
This PR takes the latest change from master to rel-1.7.0 branch along with updated version number and preview domain name. The version number will be changed back to 1.7.0 after release candidate passes verification.


codemzs:
@chinhuang007 It seems to only have last commit from master? I don’t see SoftmaxCrossEntopyLoss for example.

chinhuang007:
@codemzs Please be specific on what is missing. My branch is 1 commit ahead of master. So it is a superset of master.

codemzs:
@chinhuang007 That clarifies. Thanks.

VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.7.0
+1.7.105
```
wschin, line 1: Is this change intended? Does 105 have a special meaning?

chinhuang007: Yes, this is for the version number in test pypi. We have to use a different number each time we have a release candidate. Once the final verification is complete, we will change it back to 1.7.0 in the release branch.

chinhuang007: There is no special meaning other than I started from 101 and increment by 1 for each release candidate.

PR #2770, opened 2020-05-12
add bfloat16 to all numeric types

souptc:
Add bfloat16 to all_numeric_type list, so we can use it for most operators.


onnx/defs/schema.h
```diff
@@ -566,7 +567,8 @@ class OpSchema final {
         "tensor(int64)",
         "tensor(float16)",
         "tensor(float)",
-        "tensor(double)"};
+        "tensor(double)",
+        "tensor(bfloat16)"};
```
linkerzhang, line 571: should we also add it in the "all_tensor_types" too?

souptc: yes i added that too.

wschin:
We need some basic shape inference tests for this new type. I am not sure how hard to add some numerical tests (like creating input and output from Tensorflow), but if it's not very hard, we should add some.

onnx/defs/math/defs.cc
```diff
@@ -887,7 +910,8 @@ ONNX_OPERATOR_SET_SCHEMA(
              "tensor(uint32)",
              "tensor(uint64)",
              "tensor(int32)",
-             "tensor(int64)"},
+             "tensor(int64)",
+             "tensor(bfloat16)"},
```
wschin, line 914: Not sure if bfloat16 should be added in the beginning of this list. @linkerzhang, any comment?

onnx/defs/tensor/defs.cc
```diff
@@ -2088,7 +2093,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             static_cast<int64_t>(1))
         .TypeConstraint(
             "T1",
-            {"tensor(float)", "tensor(double)"},
+            {"tensor(float)", "tensor(double)", "tensor(bfloat16)"},
```
wschin, line 2096: I don't think we need bfloat16 for this operator. At least, it doesn't support float16 now.

gramalingam:
@wschin : did you mean node tests? I think shape-inference should not depend on the primitive type (type-inference does, but if it correctly works for multiple types already, I don't think it is important to add new ones). But I agree that test-cases for the runtime would be helpful.

wschin:
> @wschin : did you mean node tests? I think shape-inference should not depend on the primitive type (type-inference does, but if it correctly works for multiple types already, I don't think it is important to add new ones). But I agree that test-cases for the runtime would be helpful.

Yes, I mean node tests + a shape inference test. For shape inference test, I think we just need one for this type.

onnx/defs/tensor/defs.cc
souptc: removed

onnx/defs/logical/old.cc
```diff
@@ -6,6 +6,99 @@
 using namespace ONNX_NAMESPACE;
 
 namespace ONNX_NAMESPACE {
+
+inline void unaryLogicalOpInference_opset12(InferenceContext& ctx) {
```
gramalingam, line 10: Is this function used, or is it a copy-paste error?

gramalingam:
@souptc : thanks for the PR! As discussed above, is there any way to add some test-data for bfloat16 for at least one or two ops, to verify the representation? May be casting from float32 to bfloat16 and vice-versa?

onnx/defs/logical/old.cc
souptc: removed.

gramalingam:
@wschin : do you have any further comments or concerns? Thanks.

onnx/backend/test/case/node/cast.py
```diff
@@ -29,10 +30,40 @@ def export():  # type: () -> None
...
-            if 'STRING' != from_type:
+            input_type = None
+            output_type = None
+            if 'BFLOAT16' == from_type or 'BFLOAT16' == to_type:
+                np_fp32 = np.array([u'0.47892547', u'0.48033667', u'0.49968487', u'0.81910545',
+                    u'0.47031248', u'0.816468', u'0.21087195', u'0.7229038',
+                    u'NaN', u'INF', u'+INF', u'-INF'], dtype=np.float32)
+                little_endisan = sys.byteorder == 'little'
+                np_uint16_view = np_fp32.view(dtype=np.uint16)
+                np_bfp16 = np_uint16_view[1::2] if little_endisan else np_uint16_view[0::2]
```
gramalingam, line 46: So, I guess this trick works correctly for NaN, INF, -INF (since you have those values in the test-case)?

souptc: yes i have the test data there, and it is done by the standard conversion between float and bfloat16. I didn't find the definition of NaN in bfloat16 as bfloat16 is not in IEEE, but as TF apply the same conversion for all the float32 numbers, i think it should be correct.

gramalingam:
@linkerzhang @postrational @wschin : is this okay to merge in? Thanks!

PR #2772, opened 2020-05-14
Add support for per-axis scaling for QDQ linear nodes

kevinch-nv:
Add support for the scales and zero points to be 1-D tensors to support per-axis quantizing and dequantizing of tensors.

Schema changes:

- Added `axis` attribute to denote which axis the scales/zero points will operate on.
- Updated schema to opset 13.


kevinch-nv:
@onnx/sig-operators-approvers please help review. I targeted this change against opset 13 since ONNX 1.7 was recently released, let me know if there are any more files I should change to reflect the opset update.

kevinch-nv:
@askhade can you help review? 

Also @onnx/sig-operators-approvers sorry for the spam, but I'm not exactly sure why CI is failing on this change. Can someone help root cause the failure? Thanks!

docs/TestCoverage.md
```diff
@@ -2694,7 +2694,46 @@ expect(node, inputs=[x], outputs=[y],
...
+               [[  5,  24],
+                [ 24,  87],
+                [ 32,  13]],
+
+               [[245,  99],
+                [  4, 142],
+                [121, 102]],],], dtype=np.uint8)
+x_scale = np.array([2, 4, 5], dtype=np.float32)
+x_zero_point = np.array([84, 24, 196], dtype=np.uint8)
+y = np.array([[[[-162,   10],
```
askhade, line 2720: Instead of adding direct output please add a method to do the output computation.

docs/TestCoverage.md
```diff
@@ -7521,7 +7560,45 @@ expect(node, inputs=[a, a_scale, a_zero_point, b, b_scale, b_zero_point, y_scale
...
+               [[ -76,    0],
+                [   0,  252],
+                [  32,  -44]],
+
+               [[ 245, -485],
+                [-960, -270],
+                [-375, -470]],],], dtype=np.float32)
+y_scale = np.array([2, 4, 5], dtype=np.float32)
+y_zero_point = np.array([84, 24, 196], dtype=np.uint8)
+y = np.array([[[[  3,  89],
```
askhade, line 7585: Same as above please add a method to compute this output. Adding a method will benefit anybody who is implementing this operator in their runtime.

onnx/defs/quantization/defs.cc
```diff
@@ -83,11 +93,16 @@ ONNX_OPERATOR_SET_SCHEMA(
             "y",
             "N-D full precision output tensor. It has same shape as input 'x'.",
             "tensor(float)")
+        .Attr(
+            "axis",
```
askhade, line 97: Please update the documentation to include axis. How is axis incorporated in this formula "y = (x - x_zero_point) * x_scale" ?

Other quantization ops like QLinearMatmul which allow 1-D tensors for scale and zero point do not have axis attribute, I am trying to understand whether we need to update those ops too?

askhade:
FYI @tracysh @yufenglee

askhade:
> @askhade can you help review?
> 
> Also @onnx/sig-operators-approvers sorry for the spam, but I'm not exactly sure why CI is failing on this change. Can someone help root cause the failure? Thanks!

How did you update docs like Operators.md, Changelog.md etc? I just checked the Linux CI and looks like it is failing due to formatting differences between your version and the expected version. You can check "https://github.com/onnx/onnx/blob/master/tools/update_doc.sh" to see how to auto generate these files.

onnx/defs/quantization/defs.cc
```diff
@@ -34,6 +37,11 @@ ONNX_OPERATOR_SET_SCHEMA(
             "y",
             "N-D quantized output tensor. It has same shape as input 'x'.",
             "T2")
+        .Attr(
+            "axis",
+            "(Optional) The axis of the channel dimension of the input tensor.",
+            AttributeProto::INT,
+            static_cast<int64_t>(1))
```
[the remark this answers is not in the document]
kevinch-nv: Good point. I've updated the MR description and schema changes to a general per-axis QDQ support, and included the fact that axis can be negative for negative indexing.

docs/TestCoverage.md
kevinch-nv: Done!

docs/TestCoverage.md
kevinch-nv: Done!

onnx/defs/quantization/defs.cc
kevinch-nv: Which documentation do I need to update specifically? I've added it as an attribute for these ops, and generally I've seen in the ONNX spec that attributes don't need to be explicitly explained.

I'm targeting only the QDQ Linear ops in this MR since these are the most general quantization operators.

askhade: I am trying to understand why do we need axis at all? We do per channel quantization and data is in format NCHW so we know where channel is?

onnx/defs/quantization/defs.cc
```diff
@@ -54,27 +62,30 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            "Scale for input 'x'. It's a scalar, which means a per-tensor/layer quantization.",
+            "Scale for input 'x'. It can be a scalar, which means a per-tensor/layer dequantization, "
+            "or a 1-D tensor for per-axis dequantization.",
             "tensor(float)")
         .Input(
             2,
             "x_zero_point",
-            "Zero point for input 'x'. It's a scalar, which means a per-tensor/layer quantization. "
+            "Zero point for input 'x'. It can be a scalar, which means a per-tensor/layer dequantization, "
+            "or a 1-D tensor of size C for per-channel dequantization. "
```
askhade, line 88: description not consistent. line 82 mentions per-axis dequantization and line 88 says per-channel dequantization

kevinch-nv: Good catch, fixed.

onnx/defs/quantization/defs.cc
kevinch-nv: For QuantizeLinear and DequantizeLinear ops, they currently only support per-tensor quantization. This MR is for expanding the definition to also include per-axis quantization.

onnx/defs/quantization/defs.cc
```diff
@@ -6,26 +6,29 @@
...
-            "Scale for doing quantization to get 'y'. It's a scalar, which means a per-tensor/layer quantization.",
+            "Scale for doing quantization to get 'y'. It can be a scalar, which means per-tensor/layer quantization, "
+            "or a 1-D Tensor for per-axis quantization.",
             "tensor(float)")
         .Input(
             2,
             "y_zero_point",
-            "Zero point for doing quantization to get 'y'. It's a scalar, which means a per-tensor/layer quantization. "
+            "Zero point for doing quantization to get 'y'. It can be a scalar, which means a per-tensor/layer quantization, "
+            "or a 1-D tensor of size C for per-channel quantization."
```
askhade: same here - description incosistent... line 25 vs line 31. Also add what C is for example tensor of size C where C = number of channels

kevinch-nv: Fixed in latest commit

kevinch-nv:
@askhade Can I get a final review / approval? Thanks!

kevinch-nv:
@askhade ping ^^

linkerzhang:
@askhade any more comments?

kevinch-nv:
@linkerzhang can someone else help review this MR if @askhade does not respond? Thanks!

linkerzhang:
the PR looks good to me. @ebarsoum @postrational to check it

askhade:
The only open issue from my side for this PR is that why do we need axis attribute for QuantizeLinear and DequantizeLinear ops... Is the axis for per axis quantization ever going to be other than the channel axis? If no then why do we need an axis attribute? ONNX only supports NCHW data so channel axis is always known...

skyw:
> The only open issue from my side for this PR is that why do we need axis attribute for QuantizeLinear and DequantizeLinear ops... Is the axis for per axis quantization ever going to be other than the channel axis? If no then why do we need an axis attribute? ONNX only supports NCHW data so channel axis is always known...

I don't think QuantizeLinear and DequantizeLinear need to be limited by what convolution layout ONNX supports. They look independent as QuantizeLinear and DequantizeLinear can take any tensor. Even for convolution, conv and convtranspose use different layout for weight, **M x C/group x kH x kW** vs **C x M/group x kH x kW**, axis needs to be 1 and 0 respectively.

onnx/defs/quantization/defs.cc
```diff
@@ -42,7 +50,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             "T2",
             {"tensor(int8)", "tensor(uint8)"},
             "Constrain 'y_zero_point' and 'y' to 8-bit integer tensor.")
-        .SetDoc(QuantizeLinear_ver10_doc)
+        .SetDoc(QuantizeLinear_ver13_doc)
         .TypeAndShapeInferenceFunction(
             [](ONNX_NAMESPACE::InferenceContext& ctx) {
               propagateElemTypeFromInputToOutput(ctx, 2, 0);
```
yufenglee, line 56: There is a bug here. y_zero_point is optional. if it doesn't exist, we should set the output type uint8.

yufenglee:
Thanks @kevinch-nv for adding the attribute! It is one feature that is in my backlog. Overall, it looks good to me. Only the bug in the QuantizeLinear inference. Though it is not introduced in this PR, it would be great if you can help to fix it.

linkerzhang:
@kevinch-nv an open comment is about whether "axis" should be specified in these two operators, or only per channel (axis = 1) needs to be supported. @askhade has this question here https://gitter.im/onnx/operators

linkerzhang:
> Thanks @kevinch-nv for adding the attribute! It is one feature that is in my backlog. Overall, it looks good to me. Only the bug in the QuantizeLinear inference. Though it is not introduced in this PR, it would be great if you can help to fix it.

Let's merge this PR and have a separate PR to fix the bug you mentioned, either from you or Kevin :). Sounds good?

onnx/defs/quantization/defs.cc
kevinch-nv: I've opened https://github.com/onnx/onnx/issues/2839 to track this issue, we can resolve this in another MR.

kevinch-nv:
I've opened an issue (https://github.com/onnx/onnx/issues/2839) to track the type error @yufenglee mentioned, and resolved the opened conversations. @linkerzhang can you help merge? Thanks!

postrational:
LGTM

PR #2773, opened 2020-05-14
Deprecate Travis CI

vinitra:
Remove Travis CI (as these options are covered in Linux CI and Mac OS CI). 

Add C++ API tests to Linux CI. Add debug option to Mac OS CI. Revision: As this has already been done in https://github.com/onnx/onnx/pull/2908, I will solely focus on the travis ci deprecation.


jacky82226:
Hi @vinitra,
I was wondering that why we don't have
```
backend-test-tools generate-data
```
in [Windows-CI.yml](https://github.com/onnx/onnx/blob/master/.azure-pipelines/Windows-CI.yml)?
In fact, I cannot run backend-test-tools generate-data on my Windows...

.azure-pipelines/MacOS-CI.yml
```diff
@@ -30,9 +37,10 @@ jobs:
       conda install -c conda-forge pybind11 protobuf
       brew install protobuf
       set ONNX_ML=${onnx_ml}
+      set DEBUG=${onnx_debug}
       set ONNX_BUILD_TESTS=1
       set USE_MSVC_STATIC_RUNTIME=0
-      set CMAKE_ARGS=-DONNX_USE_PROTOBUF_SHARED_LIBS=ON -DProtobuf_USE_STATIC_LIBS=OFF -DONNX_USE_LITE_PROTO=ON
+      set CMAKE_ARGS=-DONNX_WERROR=ON -DONNX_USE_PROTOBUF_SHARED_LIBS=ON -DProtobuf_USE_STATIC_LIBS=OFF -DONNX_USE_LITE_PROTO=ON
```
jacky82226, line 43: For Mac OS and Linux OS, set XXX=YYY is a useless command.
Should use export XXX=YYY instead.
Actually I think for Mac OS and Linux OS we don't need to modify these flags for building?
At least I can build onnx successfully without modifying these flags on my Mac.

vinitra: > Hi @vinitra,
> I was wondering that why we don't have
> 
> ```
> python onnx/backend/test/stat_coverage.py
> backend-test-tools generate-data
> ```
> 
> in [Windows-CI.yml](https://github.com/onnx/onnx/blob/master/.azure-pipelines/Windows-CI.yml)?
> In fact, I cannot run backend-test-tools generate-data on my Windows...

Not sure. I believe it wasn't there in the original Windows build CI. I think one CI that checks these should be plenty.

vinitra: I've updated the comment above accordingly :) I suppose since it was working, I never questioned it. Thanks for the catch.

vinitra:
@askhade -- ready to merge?

askhade (approved):

PR #2774, opened 2020-05-14
Update setup info for ONNX pip account

codemzs:
Updates the below Author, Author-email and License in ONNX pip account.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -325,13 +325,14 @@ setuptools.setup(
     ext_modules=ext_modules,
     cmdclass=cmdclass,
     packages=packages,
+    license='MIT',
     include_package_data=True,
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -325,13 +325,14 @@ def run(self):
     ext_modules=ext_modules,
     cmdclass=cmdclass,
     packages=packages,
+    license='MIT',
```
vinitra, line 328: We're moving to Apache 2.0 licensing with the shift to LFAI (perhaps in the future, if not now?). @prasanthpul might have more thoughts on this.

prasanthpul: we'll need to update once we actually switch (maybe later this month). at this moment it MIT

PR #2783, opened 2020-05-20
Catch inference errors in checker

jcwchen:
**Description**
- Catch these shape inference errors then throw in the end
- Correct some tests after enabling shape inference
- Improve previous code: put `checkShapesAndTypes` in `mergeShapesAndTypes` (check before merge)
- While encountering experimental or unsupported operators, throw a warning instead of an error (Some tests in pytorch or caffe2 use them) 

**Motivation and Context**
Decouple a task from https://github.com/onnx/onnx/pull/2744
Originally, the shape inference context implementation itself ignore exceptions.
Trying to enable this and some tests will fail.


onnx/defs/tensor/defs.cc
```diff
@@ -108,8 +108,8 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::all_tensor_types(),
             "Constrain input and output types to all tensor types.")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
-          // Type inference
-          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          // Type inference, pass true to tell it is reshape 
+          propagateElemTypeFromInputToOutput(ctx, 0, 0, true);
```
gramalingam, line 112: Why do we need this? The general assumption in type-inference is that input-types are known. What is the use-case / scenario where this does not hold?

jacky82226: For [tests ](https://github.com/onnx/onnx/blob/cfab05a5f1cf33065a7a2ccc3df8020707e8998f/onnx/test/shape_inference_test.py#L42), these two inputs are set to TensorProto.UNDEFINED. I thought it should be right here because I have no idea what else tensor types we can set for them. Therefore, I chose to modify the checker part not to check the tensors type for reshape ops. 

BTW, actually there are many conflicts in tests after applying inference errors catching because the inference check is kind of general (every tensors need type definition) and there might be some exceptions. For example, ["Loop"](https://github.com/onnx/onnx/blob/cfab05a5f1cf33065a7a2ccc3df8020707e8998f/onnx/defs/controlflow/old.cc#L750). There are some unknown tensors like "M" or "cond" will trigger the inference error [here](https://github.com/onnx/onnx/blob/cfab05a5f1cf33065a7a2ccc3df8020707e8998f/onnx/defs/shape_inference.h#L165). 

My current solution is to make checker not check tensor types for these particular ops (reshape and Loop). In that case, I don't need to modify the inference graph and break something. Or maybe we could just let inference errors become warnings? Do you have any idea about how to modify it correctly?

gramalingam: For the first part: I think it would be better to change TestShapeInference if possible, instead of changing the shape-inference code.

gramalingam: For the loop case: where is the test that is failing? Again, even with a Loop, the expectation is that when we analyze a complete model, we should be able to infer the types inside a Loop from the outer context.

onnx/defs/controlflow/old.cc
```diff
@@ -817,59 +778,46 @@ scan_output values of the scan construct (similar to the concatenated intermedia
...
 scanned) for both inputs and outputs, specify a scan_input axis and scan_output axis
 value of 1.
-
 Note that because of the ONNX restriction that only the last parameter of an operator can
 be variadic, the initial-states and scan-inputs are listed together as one input parameter.
 Similarly, the final-states and scan-outputs are listed together as one output parameter.
 The attribute num_scan_inputs indicates the number M of scan-inputs.
-
 The behavior of
-
```
vinitra, line 854: @jacky82226, can you remove the whitespace changes in your PR?

jacky82226: Good catch. Thank you.

onnx/defs/tensor/defs.cc
jacky82226: > For the first part: I think it would be better to change TestShapeInference if possible, instead of changing the shape-inference code.

If so, I am thinking to use TensorProto.String instead of TensorProto.UNDEFINED.

jacky82226: > For the loop case: where is the test that is failing? Again, even with a Loop, the expectation is that when we analyze a complete model, we should be able to infer the types inside a Loop from the outer context.

Test is [Here](https://github.com/pytorch/pytorch/blob/master/test/onnx/test_pytorch_onnx_caffe2.py#L2215). Tests involving forloop or while will all fail. I think the tests should be OK so I focus on modifying the checker part.  

Originally the checker will check input including "M" and "cond", but I don't know where to remove them from the graph. Do you know where to modify the input and output of graph (g_) [here](https://github.com/onnx/onnx/blob/cfab05a5f1cf33065a7a2ccc3df8020707e8998f/onnx/shape_inference/implementation.cc#L434)?  (In other words, which function calls the Loop in old.cc?) Not sure whether it is changeable. If yes, we might remove them in shape_inference because they look useless for shape inference.

jacky82226: > > For the first part: I think it would be better to change TestShapeInference if possible, instead of changing the shape-inference code.
> 
> If so, I am thinking to use TensorProto.String instead of TensorProto.UNDEFINED in shape_inference_test.py.

NVM. Just found a better way to do this. Should use the type from the seed_value.

onnx/test/shape_inference_test.py
```diff
@@ -39,8 +41,8 @@ def _make_graph(self,
                 input_value_infos.append(seed_value_info)
             else:
                 value_info.append(seed_value_info)
-                input_value_infos.append(make_tensor_value_info('SEED_' + seed_name, TensorProto.UNDEFINED, ()))
-                input_value_infos.append(make_tensor_value_info('UNKNOWN_SHAPE_' + seed_name, TensorProto.UNDEFINED, ()))
+                input_value_infos.append(make_tensor_value_info('SEED_' + seed_name, proto_type, ()))
+                input_value_infos.append(make_tensor_value_info('UNKNOWN_SHAPE_' + seed_name, proto_type, ()))
```
gramalingam, line 43: shape should have a a fixed type (INT64) independent of seed-value's type: see: https://github.com/onnx/onnx/blob/master/docs/Operators.md#Reshape

onnx/test/shape_inference_test.py
```diff
@@ -24,12 +24,14 @@ def _make_graph(self,
             initializer = []
         names_in_initializer = set(x.name for x in initializer)
         input_value_infos = []
+        proto_type = TensorProto.FLOAT  # Should be set; checker does not allow UNDEFINED
```
gramalingam: Better to move this into the else branch of the first if-else inside the for-loop below (line 35/37)

onnx/test/shape_inference_test.py
jacky82226: Great catch. Set the shape correctly and then some tests don't need to be changed.

onnx/defs/tensor/defs.cc
jacky82226: It fails [here](https://github.com/onnx/onnx/blob/cfab05a5f1cf33065a7a2ccc3df8020707e8998f/onnx/defs/shape_inference.h#L165). I suspect the reason is the inference graph contains "M" and "cond" and both of them are TensorProto.UNDEFINED. And the output of Loop node is TensorProto.UNDEFINED, too. 

Yes, I can successfully create such model from PyTorch and it looks normal on Netron. 

My understanding is shape inference works normal for the inference part. Except for checking the tensors: it needs each tensor has defined tensor type. In that case, there are some inputs and outputs (like a condition) for "Loop" do not meet this requirement. 

To avoid checking this kind of tensors, I need to modify the input and output of the graph [here](https://github.com/onnx/onnx/blob/cfab05a5f1cf33065a7a2ccc3df8020707e8998f/onnx/shape_inference/implementation.cc#L434) as mentioned. However, I am still looking for where is the code for calling the Loop Node (construct the inference graph) or how can I modify the inference graph without breacking the logic.

gramalingam: The first question we should answer is: is there a problem in the model (this implies that the problem is with the exporter that created the model) or not (in this case the bug is in the type inference code).

jacky82226: Agree. I just checked the code on the Pytorch test  and found that the loop model can pass the onnx.checker.check_model(model) first but fail on the onnx.shape_inference.infer_shapes(model). I will review the Pytorch exporter part as well. Thank you for the reminder.

jacky82226: @gramalingam Thank you for the previous instructions for solving the input issue. 

Another problem is about checking the output. The Loop node checker fails [here](https://github.com/onnx/onnx/blob/e094e101da8cd2a10d6bf5f5ab692b50943a1470/onnx/defs/controlflow/old.cc#L588). There are two output for Loop node and both of them are `TypeProto::ValueCase::VALUE_NOT_SET` [here](https://github.com/onnx/onnx/blob/e094e101da8cd2a10d6bf5f5ab692b50943a1470/onnx/shape_inference/implementation.cc#L470). It seems that there is a bug for the graph output not updated as well, but the output of Loop model looks normal (has valid type) by Netron. 

Since we cannot solve this problem in the same way as the input (not providing inferred type), what would you suggest to solve this? (Maybe skip the output checker?)

gramalingam: I think the problem is here: https://github.com/onnx/onnx/blob/e094e101da8cd2a10d6bf5f5ab692b50943a1470/onnx/shape_inference/implementation.cc#L267 ... when we infer the type of X, we need to check if X is a graph output. If it is a graph output, we should update the type of the output, instead of adding a new valueinfo. So, we need another "else if" branch to handle this.

onnx/shape_inference/implementation.cc
```diff
@@ -303,10 +302,12 @@ static void InferShapesImpl(
...
   }
     for (auto& vi : *g->mutable_output()) {
-    auto iter = outputTypesByName.find(vi.name());
-    if (iter != outputTypesByName.end() && iter->second != NULL) {
-      vi.mutable_type()->mutable_tensor_type()->set_elem_type(iter->second);
-    }
+      if (!vi.name().empty()) {
+        auto iter = outputTypesByName.find(vi.name());
+        if (iter != outputTypesByName.end() && iter->second != NULL) {
+          vi.mutable_type()->mutable_tensor_type()->set_elem_type(iter->second);
```
gramalingam, line 308: The basic idea here is good. However, the existing code has another limitation, which shows up here too: it assumes that the inputs/outputs of the graph are all tensor-types. It would be nice to update this to allow it to be of other types (like sequence, etc.) also.

gramalingam: But may be we should make it a TODO comment and address it separately, rather than let this PR become too big. I don't think it is affecting any existing models currently.

onnx/shape_inference/implementation.cc
```diff
@@ -265,9 +272,15 @@ static void InferShapesImpl(
           existingType = iter->second;
           checkShapesAndTypes(*inferredType, *existingType);
         } else {
-          auto vi = g->add_value_info();
-          vi->set_name(n.output(i));
-          existingType = vi->mutable_type();
+            auto output_iter = outputTypesByName.find(n.output(i));
+            if (output_iter != outputTypesByName.end() && inferredType->tensor_type().elem_type() != TensorProto::UNDEFINED) {
```
gramalingam, line 276: I am a bit confused by this and the lines below at line number 304: both seem to serve the same purpose. Is there a reason we need both? I think the loop at line 304 should be sufficient, and we can leave this part as before. What do you think?

onnx/shape_inference/implementation.cc
```diff
@@ -179,9 +179,13 @@ static void InferShapesImpl(
       valueTypesByName[vi.name()] = vi.mutable_type();
   }
   for (auto& vi : *g->mutable_output()) {
+    
+    // Save names of output with undefined types
     if (vi.mutable_type() != NULL && !vi.mutable_type()->mutable_tensor_type()->has_elem_type()) {
```
gramalingam, line 184: I think protobuf implementation will create a type field if it does not exist when we call "mutable_type". Please see: https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#embeddedmessage . I think it is safer to first check "vi.has_type()" before calling mutable_type.

onnx/shape_inference/implementation.cc
```diff
@@ -146,6 +142,8 @@ void mergeShapesAndTypes(
 void mergeShapesAndTypes(
     const TypeProto& inferredType,
     TypeProto* existingType) {
+  // Check before merge
+  checkShapesAndTypes(inferredType, *existingType);
   if (inferredType.has_tensor_type()) {
```
gramalingam, line 147: We need to add a case for "if (inferredType.value_case() == VALUE_NOT_SET)".

onnx/shape_inference/implementation.cc
```diff
@@ -84,6 +84,10 @@ void checkShapesAndTypes(
     const TypeProto& existingType) {
   const auto inferredTypeCase = inferredType.value_case();
   const auto existingTypeCase = existingType.value_case();
+  if (existingTypeCase == TypeProto::ValueCase::VALUE_NOT_SET) {
```
gramalingam, line 87: Please add "|| inferredTypeCase == TypeProto::ValueCase::VALUE_NOT_SET" to the if-condition. It will make it more robust, I think.

onnx/shape_inference/implementation.cc
```diff
@@ -211,23 +209,24 @@ static void InferShapesImpl(
...
         continue;
       }
     } else if (schema->HasFunction()) {
       try {
         InferShapeForFunctionNode(
           schema->GetFunction(), schema_registry, ctx);
       } catch (const ONNX_NAMESPACE::InferenceError& function_ex) {
-        (void)function_ex;
+        // Continue with inference for remaining nodes
+        inference_errors.push_back(std::string(function_ex.what()));
```
gramalingam, line 229: I think we should replace the checks in line 243 and 248 by "if (inferredType.value_case() == VALUE_NOT_SET)".

onnx/shape_inference/implementation.cc
gramalingam: Sorry, it was a  copy-paste error. I meant the case existingType.value_case == VALUE_NOT_SET needs to be handled. We need to copy the type from inferredType to existingType.

gramalingam: Ok, please ignore my comment. I think the following code works fine. "existingType->mutable_tensor_type()" will create the tensor_type if not already present.

onnx/shape_inference/implementation.cc
```diff
@@ -241,11 +240,9 @@ static void InferShapesImpl(
       }
       for (int i = 0; i < n.output_size(); ++i) {
         const auto* inferredType = ctx.getOutputType(i);
-        if (!inferredType->has_tensor_type() &&
-            !inferredType->has_sequence_type()) {
+        if (inferredType->value_case() == TypeProto::ValueCase::VALUE_NOT_SET) {
```
gramalingam, line 243: Great, thanks! Can we also eliminate the lines 246 to 254? It will help simplify the code and make it easier to maintain and extend in the future.

jacky82226:
I think the test failure of opset7 and opset8 is because it used `ConstantFill` instead of `ConstantOfShape` for `torch.randn` before `torch.onnx._export`. The current onnx shape inference cannot inference `ConstantFill` (old experiment op) so it fails (empty input for the next node). @spandantiwari @BowenBao Do you have any comment of this? Thanks.

jcwchen:
According to https://github.com/onnx/onnx/pull/1434, while encountering deprecated operators, shape inference or checker will simply stop checking and give a warning to remind users.

onnx/checker.cc
```diff
@@ -814,6 +799,30 @@ void check_model(const ModelProto& model) {
...
+                                          "GRUUnit",
+                                          "GivenTensorFill",
+                                          "ImageScaler",
+                                          "ParametricSoftplus",
+                                          "Scale",
+                                          "ScaledTanh"};
+
+bool checkIsExperimental(std::string node_op_type) {
+  if (experimental_ops.count(node_op_type)) {
+    std::cerr << "Warning: " << node_op_type << " was a removed "
```
askhade, line 817: This method as the name suggests should only check whether the given op is experimental or not... .i.e return true if it is or false if not... the calling code should decide what to do with the result. I would say move the error out of this method

Also in case of onnx checker the warning message should be something like - checker does not support models with experimental ops

onnx/checker.h
```diff
@@ -140,5 +140,7 @@ void check_function(
 void check_model(const ModelProto& model);
 void check_model(const std::string& model_path);
 
+bool checkIsExperimental(std::string node_op_type);
```
askhade, line 143: nit: change name to check_is_experimental_op

jcwchen:
If onnx.checker stops while encountering experimental operators, caffe2-onnx would bump into an error because it uses onnx.shape_inference.infer_shapes for the conversion [here](https://github.com/pytorch/pytorch/blob/b7b99ab0c8f82100177729b9751481852d83e77e/caffe2/python/onnx/backend.py#L708). And there are some models use experimental operators which can be successfully converted previously... Still thinking how to solve it.

jcwchen:
**Another issue**
Shape inference cannot work with unsupported operators like operators from `org.pytorch._caffe2`.

**My proposal**
To solve this, while shape inference encounters unsupported operators, it will throw a warning instead of throw an error.

Now all the tests can be passed on CircleCI now. This PR is ready for review.

onnx/shape_inference/implementation.cc
```diff
@@ -283,6 +295,14 @@ static void InferShapesImpl(
       throw;
```
askhade, line 295: It is unclear which exceptions are thrown vs which ones are caught and reported at the end. 
Seems like you keep the throw statement here because if we reach this catch block then it means Type/Shape inference failed for this node and we do not want to continue right?

In the earlier cases the exceptions are caught and stored in inference_errors because there was not enough info available to infer shapes and types so instead of returning we decide to continue ?

onnx/shape_inference/implementation.cc
```diff
@@ -227,36 +239,53 @@ static void InferShapesImpl(
...
       continue;
     }
     auto domain_version = dit->second;
-
     const auto schema =
         schema_registry->GetSchema(n.op_type(), domain_version, n.domain());
     InferenceContextImpl ctx(
         n, valueTypesByName, inputDataByName, &graphInferenceContext);    
     if (!schema) {
+      std::cerr << "Warning: Shape inference does not support this operator: "
```
gramalingam, line 258: May be change the message to "Warning: Unsupported operator " << n.op_type() << ". No schema registered for this operator."

onnx/shape_inference/implementation.cc
```diff
@@ -505,10 +524,17 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
 
   return graphOutputTypes;
 }
+
+std::string getErrorWithNodeInfo(NodeProto n, std::runtime_error err) {
+  std::string op_name = n.has_name() ? n.name() : "no name";
+  return "(op_type:" + n.op_type() + ", name:" + op_name + "): " + err.what();
```
gramalingam, line 530: Minor suggestion:
```cpp
  std::string op_name = n.has_name() ? (", node name: " + n.name()) : "";
  return "(op_type:" + n.op_type() + op_name + "): " + err.what();
```

onnx/shape_inference/implementation.cc
gramalingam: Sorry I missed this earlier. My opinion is that this should be very much like error-messages produced by a compiler. Ideally, we should report as many distinct errors as possible. So, even for checkShapesAndTypes, if it is possible for the checker to continue and check the remaining model, it would be great to do so. However, whatever we do, it is important to avoid the possibility of issues like null-dereferencing. (E.g., like some of the issues that Ashwini recently fixed.) This is one of the main reasons to terminate early (in case we are not sure whether the subsequent code will robustly handle missing types etc.).

onnx/shape_inference/implementation.cc
jcwchen: Both modified. Thanks for the suggestion.

onnx/shape_inference/implementation.cc
gramalingam: So, I would say that the first priority should be to make sure the code-base is robust (in terms of avoiding assumptions that an input-type is known when doing inference for a node). When we have that robustness, we can relax more of these early-termination errors and continue with checking the rest of the model (thus, this is a secondary priority).

PR #2794, opened 2020-05-26
Add document to explain how to define operator's differentiability

wschin:
Per ONNX Training WG's discussion, we agree to add a guideline for defining operator's differentiability. PRs of differentiability definition should follow it. Please let us know if any comments.

In addition, one thing which is not quite clear to us is --- should new operators always come with differentiability definition? Maybe this is a question to Operator SIG.


linkerzhang:
"should new operators always come with differentiability definition?" I'd suggest not :). That's also why when adding such flag, an "unknown" option is there. Thoughts?

wschin:
> "should new operators always come with differentiability definition?" I'd suggest not :). That's also why when adding such flag, an "unknown" option is there. Thoughts?

The major motivation to force people to define differentiability is to make sure a operator is as matured as possible in its first version. But I am ok with your suggestion. I am not sure how to balance  "inference-friendly" and "completeness", and we chose "completeness".

gramalingam:
This looks nice. My gut feeling is that executable-code that can be used by ONNX frameworks would be much more valuable. In other words, an version of method 3 using a standardized interface. Doing method 1 is a good way to verify that outputY of opZ is differentiable wrt inputX, but what is the payoff? Is it worth investing a lot of effort in this, as opposed to building a version of method 3 (even if it takes longer to finish)?

docs/DefineDifferentiability.md
```diff
@@ -0,0 +1,58 @@
...
+
+The differentiability definition of an operator consists of several aspects.
+
+- Differentiable inputs, which can be referenced in Gradient's `xs` attribute.
+- Differentiable outputs, which can be referenced in Gradient's `y` attribute.
+- The math equation to compute the Jacobian matrix (or tensor). If a variable (input or output) is differentiable or not is judged by math. If the Jacobian matrix (or tensor) exists, then the considered operator has some differentiable inputs and outputs.
+
+There are several strategies to implement auto-differentiation such as forward accumulation, backward accumulation, and dual variable.
+Because most deep learning frameworks are backward-based, the reviewers should ensure the PR authors provide enough details on that.
+Subsequently, we discuss several methods to define the differentiability for ONNX operator.
```
gramalingam, line 11: change "define" to "verify".

gramalingam:
Overall: this seems useful advice for developers on how to figure out what's differentiable. But I am not clear what should be made mandatory in a PR. In uncomplicated cases, option (2) would seem to be the simplest, but it is not clear what sort of examples is expected. What do you need as examples to show that addition or multiplication is differentiable?

docs/DefineDifferentiability.md
wschin: Ok.

wschin:
> This looks nice. My gut feeling is that executable-code that can be used by ONNX frameworks would be much more valuable. In other words, an version of method 3 using a standardized interface. Doing method 1 is a good way to verify that outputY of opZ is differentiable wrt inputX, but what is the payoff? Is it worth investing a lot of effort in this, as opposed to building a version of method 3 (even if it takes longer to finish)?

If the author knows how Pytorch/Tensorflow operators may be converted to  ONNX, Method 1 is an easy way to make sure a differentiability tag is correct without digging into math. The goal here is to have shortest path to observe or prove that inputs or outputs are differentiable or not. Method 3 is similar to Method 1, but Method 3 requires the author to build a correct forward operator first before applying auto-diff. Forward pass could be difficult sometime.

For reuse Method 3, as you said, we need a standard interface (I guess it would be a small runtime like ONNX). Do we have such a interface today? It seems this is related to Alibaba's test framework.

docs/DefineDifferentiability.md
```diff
@@ -0,0 +1,100 @@
...
+```
+A = [a1, a2]
+B = [b1, b2]
+C = [c1, c2]
+```
+Let `X = [a1, a2, b1, b2]` and `Y = [c1, c2]` and consider Add as a function which maps `X` to `Y`.
+Then, this function's Jacobian matrix is a 4-by-2 matrix,
+```
+J = [[dc1/da1, dc2/da1],
+     [dc1/da2, dc2/da2],
```
wschin, line 72: Switch these two rows.

docs/DefineDifferentiability.md
```diff
@@ -0,0 +1,99 @@
+# A Short Guide on Defining Differentiability for ONNX operators
```
gramalingam, line 1: "Defining Differentiability" => "the Differentiability tag"

docs/DefineDifferentiability.md
```diff
@@ -0,0 +1,99 @@
...
+
+The differentiability definition of an operator consists of several aspects.
```
gramalingam, line 3: Add "The ONNX operator schema for each operator includes a Differentiability tag for each input and output. In this document, we explain the meaning of this tag and how to ensure that a correct tag-value is provided in a PR that proposes a new operator or updates an operator. Briefly, the tag identifies the set of differentiable inputs and differentiable outputs of an operator. The meaning of the tag is that the partial derivative of each differentiable output is defined with respect to each differentiable output."

docs/DefineDifferentiability.md
```diff
@@ -0,0 +1,99 @@
...
+
+- Differentiable inputs, which can be referenced in Gradient's `xs` attribute.
+- Differentiable outputs, which can be referenced in Gradient's `y` attribute.
+- The math equation to compute the Jacobian matrix (or tensor). If a variable (input or output) is differentiable or not is judged by math. If the Jacobian matrix (or tensor) exists, then the considered operator has some differentiable inputs and outputs.
+
+There are several strategies to implement auto-differentiation such as forward accumulation, backward accumulation, and dual variable.
+Because most deep learning frameworks are backward-based, the reviewers should ensure the PR authors provide enough details on that.
```
gramalingam, line 10: I didn't quite understand. I guess this means that the PR should contain details on how the backward-prop can be implemented? The subsequent discussion seems to imply that this is not necessary? Is this optional?

docs/DefineDifferentiability.md
```diff
@@ -0,0 +1,99 @@
...
+Subsequently, we discuss several methods to verify the differentiability for ONNX operator.
```
gramalingam, line 11: We present a couple of methods below to verify the differentiability tag of ONNX operators.

docs/DefineDifferentiability.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Method 1: Reuse Existing Deep Learning Frameworks
+The first way is to show that the considered operator's backward operation exists in an existing framework such as Pytorch or Tensorflow. In this case, the author should provide a runnable python script which computes the backward pass of the considered operator. The author should also points out how to map the Pytorch or Tensor code to ONNX format (for example, the author can call `torch.onnx.export` to save an ONNX model). The following script shows the differentiability of ONNX Reshape using  Pytorch.
```
gramalingam, line 14: "points" => "point"

gramalingam:
Thanks @wschin for the document. I think it will help people understand the process, especially backward-propagation. (However, I am not sure if it is realistic to expect a PR author to include this level of detail in every PR. I suspect we may have to make a decision on a case-by-case basis. Let us see how it goes.)

docs/DefineDifferentiability.md
wschin: No. The author here means "author of differentiability tag", not the author of operator.

docs/DefineDifferentiability.md
wschin: Ok.

PR #2797, opened 2020-05-28
Rename tests for NLLLoss and SoftmaxCrossEntropy

KsenijaS:
Test names are too long and are causing following failure when running on Docker image on PyTorch CI:
```
panic: runtime error: comparing uncomparable type tar.headerError
```


KsenijaS:
@houseroad  please review. Thanks!

PR #2802, opened 2020-05-28
Revert "Rename tests for NLLLoss and SoftmaxCrossEntropy (#2797)"

BowenBao:
This reverts commit cfab05a5f1cf33065a7a2ccc3df8020707e8998f.

Another PR will be submitted shortly after to include the necessary file updates for resolving the naming issue.


gramalingam:
So, I assume there is some issue with the previous file-rename? Does something break?

KsenijaS:
@gramalingam the first PR added all files generated by tools/update_doc.sh instead only files related to NLLLoss and SCE.

gramalingam:
Do we know why other files were modified by the update script? I assume they appeared to be different from the repo version. I think someone else faced a similar issue, so I wonder if something needs to be fixed.

PR #2828, opened 2020-06-15
Adding support to update dimensions of limited layers

### LINKED ISSUE #2799: Bug in Input and output dimension update
I think there is a **bug**, update input and output dimension flow is iterating over every input and output key in the model, instead it is should iterate only over desired number of keys fed via parameter input.
```python
 update_model_dims.update_inputs_outputs_dims(model, {
                    "input_1": ['b', 3, 'w', 'h'],  "input_2": ['b', 4],
                },{
                    "output": ['b', -1, 5]
                 }   )
```

https://github.com/onnx/onnx/blob/599f3da985cf8a790a619a505abce167d87c2c30/onnx/tools/update_model_dims.py#L70-L81

akansal1:
Fixes #2799

### DIFF
--- a/onnx/tools/update_model_dims.py
+++ b/onnx/tools/update_model_dims.py
@@ -10,7 +10,7 @@ from onnx import ModelProto, ValueInfoProto
 import onnx.checker
 
 
-def update_inputs_outputs_dims(model, input_dims, output_dims):  # type: (ModelProto, Dict[Text, List[Any]], Dict[Text, List[Any]]) -> ModelProto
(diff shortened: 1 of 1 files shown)


onnx/tools/update_model_dims.py
```diff
@@ -69,15 +69,17 @@ def update_dim(tensor, dim, j, name):  # type: (ValueInfoProto, Any, int, Text)
 
     for input in model.graph.input:
         input_name = input.name
-        input_dim_arr = input_dims[input_name]
-        for j, dim in enumerate(input_dim_arr):
-            update_dim(input, dim, j, input_name)
+        input_dim_arr = input_dims.get(input_name)
```
kanthagirish-rit, line 72: `input_dims.get(input_name, [])` will help you avoid the `if` statement below

Issue #2821, opened 2020-06-11
Installing package on Windows fails on too long filenames


Korijn:
Installing onnx fails on my Windows 10 machine due to filepaths being too long ("De bestandsnaam of -extensie is te lang" is dutch for "The filename or -extension is too long"):

`ERROR: Could not install packages due to an EnvironmentError: [WinError 206] De bestandsnaam of -extensie is te lang: 'c:\\users\\korij_000\\appdata\\local\\pypoetry\\cache\\virtualenvs\\zb-hps-algo-bkp0o5gs-py3.6\\Lib\\site-packages\\onnx\\backend\\test\\data\\node\\test_negative_log_likelihood_loss_input_shape_is_NCd1d2d3_none_no_weight_negative_ignore_index_expanded\\test_data_set_0'`

I am able to work around this issue by installing to a folder higher up in my filesystem, but this filename looks a bit excessive.

askhade:
@jacky82226 : Let's fix this

jacky82226:
Hi @Korijn,
Thank you for reporting this bug! 
I was wondering that did you encounter other files also encounter this error?
I suspect the problem is [here](https://github.com/onnx/onnx/blob/cc2230603422bae893d5bc900d2d773ab34400a4/onnx/backend/test/case/node/negativeloglikelihoodloss.py#L428) and it has been update in a recent [PR](https://github.com/onnx/onnx/commit/f11f50fdb5f069eaf74af56f54edb0213559b833).
It surprisingly passed Windows-CI somehow, maybe it's because the maximum length of filename varies on different Windows machines.
I will propose a PR to fix this. Thanks.

Korijn:
If I may make a suggestion: don't include your tests and especially test data in your wheels. It's pointless and makes the package much bigger.

jacky82226:
> If I may make a suggestion: don't include your tests and especially your test data in your wheels. It's not needed at runtime and makes the package much bigger.

Good point. Did you use pip install onnx? It should be unnecessary to include them while installing the library...

Korijn:
> > If I may make a suggestion: don't include your tests and especially your test data in your wheels. It's not needed at runtime and makes the package much bigger.
> 
> Good point. Did you use pip install onnx? It should be unnecessary to include them while installing the library...

Yes I did.

snnn:
To get away from such issue:
[...]

PR #2850, opened 2020-06-19
Remove backend/test for building package

jacky82226:

Therefore, I don't see there is any benefit to put these testing files into the package. If someone has concern about it, please let me know. THANKS!


jacky82226:
Since there are some partners using the testing data under onnx, it is not a good time to remove those test data right now.

PR #2857, opened 2020-06-24
Shorten test names to prevent long filename issue

jcwchen:

However, there are some users depend on those test data in onnx:
e.g., https://github.com/NervanaSystems/ngraph-onnx/blob/master/tests/test_backend.py#L54

To compromise, the setup.py should provide an option (ONNX_EXCLUDE_TESTS) for some users to exclude those test data while installing onnx if they don't need them. The default value is False so it won't influence current usage.

If this modification has been approved, I will add a description in README.md for users to use.


jacky82226:
cc @ebarsoum @postrational for review. Thank you!

linkerzhang:
Quick question, what's your suggestion when releasing ONNX, should the test data included or not?

jacky82226:
IMO, a released library should not include test data because general users won't use them and it costs more effort while installing the package. 
However, it has been added to onnx for a while so there are some existing users depends on them... Ideally I hope them can additionally add a flag to install the test data. In that case, we can set excluding test data as default.

linkerzhang:
> IMO, a released library should not include test data because general users won't use them and it costs more effort while installing the package.
> However, it has been added to onnx for a while so there are some existing users depend on them... Ideally I hope they can additionally add a flag to install the test data. In that case, we can set excluding test data as default.

I agree with you!

I was thinking that "ONNX_EXCLUDE_TESTS" should be 1 by default in that way and have folks explicitly set it to 0 when needed.

askhade:
> > IMO, a released library should not include test data because general users won't use them and it costs more effort while installing the package.
> > However, it has been added to onnx for a while so there are some existing users depend on them... Ideally I hope they can additionally add a flag to install the test data. In that case, we can set excluding test data as default.
> 
> I agree with you!
> 
> I was thinking that "ONNX_EXCLUDE_TESTS" should be 1 by default in that way and have folks explicitly set it to 0 when needed.

I agree with this. However right now changing this to exclude tests by default might affect dependencies and this will probably need more time to get approval from community. To ease installation experience issues I would say let's check this in with default set to include tests (the way it is right now) and then open an issue for this to get approval from the community.

askhade:
@postrational ping :)

askhade:
> I don't think that removing test data from the package is the right solution for the long filename problem.
> It would break systems which were built to rely on this test data.
> 
> Instead of deleting the data, we should change the long filenames to shorter ones.

We are not proposing to delete the test data. This PR is just adding an option to exclude test data. With this PR the current behavior does not change. As I mentioned in the previous comment, removing test data will need longer discussion and approval from community members. 

@jcwchen : As part of this PR please also include the change to rename long names to short names

postrational:
> This PR is not proposing to delete the test data. This PR is just adding an option to exclude test data. 

Excluding test data will effectively delete them from user system, because after `pip install onnx` they will not have the test data installed. We could add an option like `pip install onnx[tests]` as an alternative to the users. Making users build their own packages from sources is not viable.

askhade:
> > This PR is not proposing to delete the test data. This PR is just adding an option to exclude test data.
> 
> Excluding test data will effectively delete them from user system, because after `pip install onnx` they will not have the test data installed. We could add an option like `pip install onnx[tests]` as an alternative to the users. Making users build their own packages from sources is not viable.

We should definitely rename the long names to short ones. However this PR additionally adds an option for someone to exclude test data. The onnx package will by default have the test data

jcwchen:
Thank you everyone for the feedback. I just shortened some long filenames.
There is a similar PR for renaming (shorten) tests filenames. https://github.com/onnx/onnx/pull/2803

A few file names are quite long... The longest one is `test_sce_NCd1d2d3_none_no_weight_negative_ignore_index_log_prob_expanded` which consists 72 characters.
The test names with "ignore_index" are usually long name. (5 files consist 65+ characters)
To make sure all names follow the same format, I replaced all the filenames which contain "ignore_index" into "ii". 

In addition, I modified Operator.md and TestCoverage.md. (where involving these tests)
After modification, the longest filename is now 63 characters.

Since I am not one of the authors for this part, I am not sure whether this modification would break something (like tests in Pytorch). It's better to cc them to review this and prevent writing longer test name in the future.

setup.py
```diff
@@ -290,8 +291,13 @@ def run(self):
 # Packages
 ################################################################################
 
+# Exclude test data while building onnsx
```
postrational, line 294:
```suggestion
# Exclude test data while building ONNX
```

jcwchen:
Since adding external flag to exclude test data is still debatable, this PR focuses on shortening those long test names to prevent the long filename issue on Windows. I just removed the modification in `setup.py`. 

@postrational please review. Thank you so much!

Issue #2859, opened 2020-06-25
Mac CI builds failing


prasanthpul:
MacOS build pipelines are timing out waiting for user input to install openssl package

vinitra:
This has been addressed by Jacky's PR! #2861.

PR #2861, opened 2020-06-25
Fix timeout in MacOS-CI

jcwchen:
According to https://github.com/onnx/onnx/issues/2859, the MacOS-CI has a timeout problem.

### DIFF
--- a/.azure-pipelines/MacOS-CI.yml
+++ b/.azure-pipelines/MacOS-CI.yml
@@ -27,7 +27,7 @@ jobs:
       git submodule update --init --recursive
       python -m pip install --upgrade setuptools
       python -m pip install numpy
-      conda install -c conda-forge pybind11 protobuf
+      conda install -y -c conda-forge pybind11 protobuf
       brew install protobuf
       set ONNX_ML=${onnx_ml}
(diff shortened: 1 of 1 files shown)


askhade:
This pipeline has passed in the past so what changed now? If without -y install halts for user input then how did this pipeline pass before?

jacky82226:
> This pipeline has passed in the past so what changed now? If without -y install halts for user input then how did this pipeline pass before?

Since the setting here is fixed on our end, I think maybe there are some updates on conda package (need extra confirm for installing some packages). Setting yes should fix this issue.

askhade:
> > This pipeline has passed in the past so what changed now? If without -y install halts for user input then how did this pipeline pass before?
> 
> Since the setting here is fixed on our end, I think maybe there are some updates on conda package (need extra confirm for installing some packages). Setting yes should fix this issue.

Do other CIs need a similar fix?

jcwchen:
I did check other CIs:
For Linux-CI, the conda environment has been created initially before the scripts so it should be fine 
For Windows-CI, --yes has been added.
Thank you for pointing out.

Issue #2495, opened 2019-12-08
how to uninstall onnx?


ultracxx1997:
I just do

**Instructions to build ONNX on windows in anaconda environment**

```
# Use conda-forge protobuf
conda install -c conda-forge protobuf=3.9.2 numpy

# Get ONNX
git clone https://github.com/onnx/onnx.git
cd onnx
git submodule update --init --recursive

# Set environment variable for ONNX to use protobuf shared lib
set CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"

# Build ONNX
# Optional : Set environment variable `ONNX_ML=1` for onnx-ml

python setup.py install
```
but when i done the steps,some errors happened.
my python script running on pycharm didn't work.
`import onnx ` 

can run normally, but when the program run on onnx.xxx()
a error would be caught.  
so now i just want undo the changes of build onnx made to my windows10.
uninstall it.  i try to delete the files **ONNX**,but that doesn't seem to work.
how should i do?

jcwchen:
Hi @ultracxx1997,
Sorry for the late reply.
You can git clean those untracked files and reinstall onnx.
```
git clean -xdf
python setup.py install
```
BTW, to remove those installed files. You need other command to track those installed files:
```
python setup.py install --record installed_files.txt
```
Then you can remove all files in `installed_files.txt`.

ultracxx1997:
thank you
&nbsp;

nnn


&nbsp;

PR #2862, opened 2020-06-25
Clean command in setup.py

jcwchen:
**Description**: 
Add a clean command in setup.py.
Remove files in .gitignore (which means those files are probably produced by build)
Set a list for exception (keep them when cleaning) in .gitignore 

**Motivation and Context**

- A user is looking for this solution: https://github.com/onnx/onnx/issues/2495.
- Actually I have the same requirement for using clean command to clean all build files from setup.py. 
`python setup.py clean` is useful if developers want to tune some environment settings and rebuild the onnx package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,9 +53,6 @@ onnx/*_pb.pyi
 # PyCharm files
 .idea
 
-# Visual Studio Code files
-.vscode
-
 # OSX dir files
 .DS_Store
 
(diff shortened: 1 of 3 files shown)


askhade:
How is this different than git clean?

jcwchen:
> How is this different than git clean?

Basically they are the same. The only difference would be with this PR we can keep some user-defined files like `.vs` or `.vscode`(no need to remove for reinstall), which can be defined by `.gitignore`. Therefore, this requirement is minor...maybe we can close this PR.

PR #2863, opened 2020-06-26
Fix environment settings for CI

jcwchen:
**Description**: 
Remove unnecessary environment settings for Linux and Mac.
Use `export` instead of `set` on Linux and Mac to validate the setting for `ONNX_ML` and `ONNX_BUILD_TESTS`.
In addition, add backend test on Windows (not sure whether it is necessary)

**Motivation and Context**
Some environment settings might be unnecessary for Linux and Mac.
Should use `export` instead of `set` on Unix-like operating systems.


jcwchen:
@vinitra could you take a look on this?
Thank you!

jcwchen:
Has been updated in https://github.com/onnx/onnx/pull/2773/files.

Issue #2869, opened 2020-06-30
ort-nightly version not found in CircleCI script


jcwchen:
# Bug Report

### Describe the bug
For recent PRs, while running CircleCI, they all encountered the same error by this command:
```
pip install -i https://test.pypi.org/simple/ ort-nightly==1.2.0.dev202004201
```
Error:
```
+ pip install -i https://test.pypi.org/simple/ ort-nightly==1.2.0.dev202004201
Looking in indexes: https://test.pypi.org/simple/
Collecting ort-nightly==1.2.0.dev202004201
  Could not find a version that satisfies the requirement ort-nightly==1.2.0.dev202004201 (from versions: 1.3.0.dev202005079, 1.3.0.dev202005111, 1.3.0.dev202005112, 1.3.0.dev202005121, 1.3.0.dev202005123, 1.3.0.dev202005171, 1.3.0.dev202006022, 1.3.0.dev202006031, 1.3.0.dev202006061, 1.3.0.dev202006071, 1.3.0.dev202006081, 1.3.0.dev202006091, 1.3.0.dev202006101, 1.3.1.dev202006111, 1.3.1.dev202006115, 1.3.1.dev202006161, 1.3.1.dev202006171, 1.3.1.dev202006172, 1.3.1.dev202006191, 1.3.1.dev202006192, 1.3.1.dev202006221, 1.3.1.dev202006254)
No matching distribution found for ort-nightly==1.2.0.dev202004201
You are using pip version 19.0, however version 20.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```
I am looking deeply into this issue.

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): ubuntu16.04
- ONNX version (*e.g. 1.7*):  1.7
- Python version: 3.6

### Reproduction instructions
- Describe the code to reproduce the behavior.
```
pip install -i https://test.pypi.org/simple/ ort-nightly==1.2.0.dev202004201
```

### Expected behavior
This version should be found and pass the build in CircleCI.

askhade:
We should update the version to 1.3.1 as it is the latest ort version. The older versions were deleted recently.

PR #2870, opened 2020-06-30
Update ort version in CircleCI

jcwchen:
According to https://github.com/onnx/onnx/issues/2869, update the ort version to 1.3.1 in CircleCI.


jcwchen:
Besides ort, pip install torchvision from their master branch is broken as well. It seems they are fixing this right now.

jcwchen:
It seems to me that the pip install torchvision problem has been solved.
With this PR, CircleCI should work normally now.

Issue #2877, opened 2020-07-03
collect_snippets writes to console


lutzroeder:
# Bug Report

### Describe the bug
`collect_snippets` writes warning messages to console.

### System information
- OS Platform and Distribution:  macOS Catalina
- Python version: 3.7

### Reproduction instructions
```
git clone --quiet https://github.com/onnx/onnx.git
pushd onnx
git submodule update --quiet --init --recursive
python -m venv ../env
source ../env/bin/activate
python -m pip install --quiet --upgrade pip
python -m pip install --quiet protobuf
export ONNX_ML=1
export ONNX_NAMESPACE=onnx
python -m pip install --quiet .
popd
python -c "from onnx.backend.test.case import collect_snippets; snippets = collect_snippets()"
deactivate
```

### Expected behavior
No text is printed to console.

### Actual behavior
```
~/env/lib/python3.7/site-packages/onnx/backend/test/case/node/reduce_log_sum_exp.py:32: RuntimeWarning: overflow encountered in log
  np.exp(data), axis=tuple(axes), keepdims=keepdims == 1))
~/env/lib/python3.7/site-packages/onnx/backend/test/case/node/reduce_log_sum_exp.py:67: RuntimeWarning: overflow encountered in log
  keepdims=keepdims == 1))
~/env/lib/python3.7/site-packages/onnx/backend/test/case/node/reduce_log_sum_exp.py:103: RuntimeWarning: overflow encountered in log
  keepdims=keepdims == 1))
~/env/lib/python3.7/site-packages/onnx/backend/test/case/node/reduce_log_sum_exp.py:136: RuntimeWarning: overflow encountered in log
  keepdims=keepdims == 1))
```

@neginraoof

jcwchen:
Hi @lutzroeder,
I encountered the same error only on MacOS (Windows works fine) when using some commands.
Using longer float type should fix this warning. I will propose a PR https://github.com/onnx/onnx/pull/2878 and cc @neginraoof for review.

Thank you for reporting this bug!

PR #2878, opened 2020-07-04
Use longer float to prevent overflow warning 

jcwchen:
**Description**
To prevent overflow warning, using longer float type `np.float64` can fix this warning.

**Motivation and Context**
https://github.com/onnx/onnx/issues/2877
I encountered the same error only on MacOS (Windows works fine) when using `python onnx/defs/gen_doc.py`, `python onnx/backend/test/stat_coverage.py` or `backend-test-tools generate-data`.


jcwchen:
cc @neginraoof for review. Thank you!

jcwchen:
According to https://github.com/onnx/onnx/blob/master/docs/Changelog.md#reducelogsum-13,
it's better to use `double` instead of float64.

jcwchen:
@ebarsoum @gramalingam could you help to review and merge this PR? Thank you!

BTW, currently please ignore the CI failures from circleCI and travisCI. They are not related to this PR and we are solving them in another PR https://github.com/onnx/onnx/pull/2890.

PR #2879, opened 2020-07-06
Update softmax family ops behavior to align with other frameworks (fix #2289)

### LINKED ISSUE #2289: Non-standard Softmax behavior
What is the reason for the non-standard behaviour of the ONNX `Softmax` operator (also `LogSoftmax` and `HardMax`)? More specifically, why do we limit (or coerce) the input to be a 2D vector? Many other deep learning frameworks ([TensorFlow](https://www.tensorflow.org/api_docs/python/tf/nn/softmax), [PyTorch](https://pytorch.org/docs/stable/nn.html#softmax), [MXNet](http://mxnet.incubator.apache.org/versions/1.5.0/api/python/ndarray/ndarray.html?highlight=softmax#mxnet.ndarray.softmax)) allow the input of the softmax to be N-dimensional and use the `axis` argument to specify which dimension of the input should be softmaxed (such that the sum of the elements in that dimension is 1). 

It seems to me like the current behaviour is less flexible than the N-dimensional behaviour, and ends up requiring tricks like first transposing the input before applying softmax (as seen in #1543). I also find it a bit strange that an activation function changes the shape of its input.

I propose that we standardise the ONNX Softmax op to follow the behaviour of the deep learning frameworks mentioned above. If the existing behaviour is needed for some applications, it can easily be recovered by reshaping tensors to be 2D. As far as I know, this is a much less common use-case (it isn't implemented in the most popular frameworks) so this shouldn't affect too many users. 

I am more than happy to make a PR for these changes, but I wanted to find out if I was missing some key insight into why have the current behaviour before making any changes.

daquexian:
Fix https://github.com/onnx/onnx/issues/2289

This PR updates softmax and logsoftmax so that they are functions and align with other frameworks (tf, pytorch, etc), and hardmax is also updated as a primitive operator


onnx/defs/math/defs.cc
```diff
@@ -49,29 +49,24 @@ std::function<void(OpSchema&)> SoftmaxFamilyDocGenerator(
...
     schema.SetDoc(doc);
     schema.Attr(
-        "axis",
-        "Describes the axis of the inputs when coerced "
-        "to 2D; defaults to one because the 0th axis most likely describes "
-        "the batch_size. Negative value means counting dimensions "
-        "from the back. Accepted range is [-r, r-1] where r = rank(input).",
-        AttributeProto::INT,
-        static_cast<int64_t>(1));
+        "axis", axis_attr, AttributeProto::INT, static_cast<int64_t>(-1));
```
linkerzhang, line 69: The version should also be bumped, given the default value of "axis" changed.

linkerzhang: In my opinion, if it's only default value change, I'd suggest to not change it. The benefit is not that big with fair change (version bump).

onnx/defs/math/defs.cc
```diff
@@ -821,14 +816,120 @@ ONNX_OPERATOR_SET_SCHEMA(
 ONNX_OPERATOR_SET_SCHEMA(
     Softmax,
     13,
-    OpSchema().FillUsing(
-        SoftmaxFamilyDocGenerator("softmax", "normalized exponential")));
+    OpSchema()
+        .FillUsing(
+            SoftmaxFamilyDocGenerator("softmax", "normalized exponential"))
+        .SetContextDependentFunctionBodyBuilder(
+            [](const FunctionBodyBuildContext& ctx,
```
linkerzhang, line 823: did you verify the correctness of the subgraph?  (the "expanded" and non "expanded" model generated in this PR, feeding them same inputs will get same outputs)

onnx/defs/math/defs.cc
daquexian: I think the current softmax op version has already been bumped since last release (current version is already 13). Is it necessary to bump it once more?

linkerzhang: no in that way. Only one version bump in one release.

docs/Changelog.md
```diff
@@ -16045,16 +16045,9 @@ This version of the operator has been available since version 13 of the default
...
-  coerced into one. For an arbitrary n-dimensional tensor
-  input \in [a_0, a_1, ..., a_{k-1}, a_k, ..., a_{n-1}] and k is
-  the axis provided, then input will be coerced into a 2-dimensional tensor with
-  dimensions [a_0 * ... * a_{k-1}, a_k * ... * a_{n-1}]. For the default
-  case where axis=1, this means the input tensor will be coerced into a 2D tensor
-  of dimensions [a_0, a_1 * ... * a_{n-1}], where a_0 is often the batch size.
-  In this situation, we must have a_0 = N and a_1 * ... * a_{n-1} = D.
-  Each of these dimensions must be matched correctly, or else the operator
-  will throw errors. The output tensor has the same shape
+  The input does not need to explicitly be a 2D vector. The "axis" attribute
```
wschin, line 16048: Could  you add the math for this operator (like it had before)?

onnx/defs/math/defs.cc
fdwr: Yeah, changing any of these {attribute names, attribute default values, tensors meanings} is a version breaking change.

onnx/defs/math/defs.cc
fdwr: Shouldn't this functional decomposition include a Flatten call? (the part that coerces the shape to 2D)

Here's the decomposition I wrote down before:
https://fdwr.github.io/LostOnnxDocs/OperatorFormulas.html

```
function SoftMax2D(Input; Output; axis):
  MaxInput = ReduceMax(Input, axes=[1], keepdims=1)
  ExpInput = Exp(Input - MaxInput)
  ReducedExpInput = ReduceSum(ExpInput, axis=[1], keepdims=1)
  Output = ExpInput / ReducedExpInput
endfunction

function F(Input; Output; axis):
  FlattenedInput = Flatten(Input, axis) // Flatten to 2D
  NormalizedInput = SoftMax2D(FlattenedInput)
  Output = Reshape(NormalizedInput, Shape(X))
endfunction
```

docs/Changelog.md
daquexian: What it had before is not the math of the operator. It just describes how the n-d vector is coerced into a 2-d vector. However, I will still add the math in the spec.

onnx/defs/math/defs.cc
daquexian: @linkerzhang @fdwr I have verified the correctness of the subgraph by inferencing the subgraph in onnxruntime and comparing the result with the reference numpy implementation

daquexian:
@linkerzhang @wschin @fdwr I have updated the PR according to the comments. Please review again. Thanks!

docs/Changelog.md
```diff
@@ -17855,20 +17851,14 @@ This version of the operator has been available since version 13 of the default
...
-  input \in [a_0, a_1, ..., a_{k-1}, a_k, ..., a_{n-1}] and k is
-  the axis provided, then input will be coerced into a 2-dimensional tensor with
-  dimensions [a_0 * ... * a_{k-1}, a_k * ... * a_{n-1}]. For the default
-  case where axis=1, this means the input tensor will be coerced into a 2D tensor
-  of dimensions [a_0, a_1 * ... * a_{n-1}], where a_0 is often the batch size.
-  In this situation, we must have a_0 = N and a_1 * ... * a_{n-1} = D.
-  Each of these dimensions must be matched correctly, or else the operator
-  will throw errors. The output tensor has the same shape
-  and contains the softmax values of the corresponding input.
+   Softmax(input, axis) = Exp(input) / ReduceSum(Exp(input), axis=axis, keepdims=1)
```
wschin, line 17856: Great!

docs/Changelog.md
```diff
@@ -17855,20 +17851,14 @@ This version of the operator has been available since version 13 of the default
...
-  case where axis=1, this means the input tensor will be coerced into a 2D tensor
-  of dimensions [a_0, a_1 * ... * a_{n-1}], where a_0 is often the batch size.
-  In this situation, we must have a_0 = N and a_1 * ... * a_{n-1} = D.
-  Each of these dimensions must be matched correctly, or else the operator
-  will throw errors. The output tensor has the same shape
-  and contains the softmax values of the corresponding input.
+   Softmax(input, axis) = Exp(input) / ReduceSum(Exp(input), axis=axis, keepdims=1) 
+  
+  The input does not need to explicitly be a 2D vector. The "axis" attribute
+  indicates the dimension along which {name} will be performed.
```
wschin, line 17859: name??

docs/Operators.md
```diff
@@ -18416,20 +18415,14 @@ expect(node, inputs=[x, starts, ends, axes, steps], outputs=[y],
...
-  will throw errors. The output tensor has the same shape
-  and contains the softmax values of the corresponding input.
+  The operator computes the normalized exponential values for the given input:
+  
+   Softmax(input, axis) = Exp(input) / ReduceSum(Exp(input), axis=axis, keepdims=1) 
+  
+  The input does not need to explicitly be a 2D vector. The "axis" attribute
+  indicates the dimension along which {name} will be performed.
+  The output tensor has the same shape
+  and contains the {name} values of the corresponding input.
```
wschin, line 18425:
```suggestion
  and contains the Softmax values of the corresponding input.
```

docs/Changelog.md
daquexian: @wschin Thanks! I missed the `ReplaceAll` for `{name}`. It is fixed now.

daquexian:
@wschin I have updated this PR. Is it ok to merge it? Thanks!

wschin:
> @wschin @linkerzhang I have updated this PR. Is it ok to merge it? Thanks!

Could you please sync this branch with master again? Sorry for being late.

daquexian:
> Could you please sync this branch with master again? Sorry for being late and we want this PR in for this release.

Done. Do I need to sign off according to the DCO check? I try signing off the commits according to its instructions but I found it makes git history confusing.

wschin:
> > Could you please sync this branch with master again? Sorry for being late and we want this PR in for this release.
> 
> Done. Do I need to sign off according to the DCO check? I tried signing off the commits according to its instructions but I found it makes git history confusing.

@prasanthpul, Is DCO required now?

PR #2880, opened 2020-07-06
[WIP] Add a installation flag to exclude test data

jcwchen:
**Description**
To compromise, the setup.py should provide an option (ONNX_EXCLUDE_TESTS) for some users to exclude those test data while installing onnx if they don't need them. The default value is False so it won't influence current usage.
Related PR: https://github.com/onnx/onnx/pull/2857, move the original idea to this WIP PR.

**Motivation and Context**
However, there are some users depend on those test data in onnx:
e.g., https://github.com/NervanaSystems/ngraph-onnx/blob/master/tests/test_backend.py#L54

The solution (adding a flag for installation or moving those test/data to another package on PyPI) is still debatable at this moment.


jcwchen:
Close it due to inactivity.

PR #2888, opened 2020-07-08
Return empty string from ToDataTypeString() when tensor_data_type not found

prabhat00155:
Github issue: https://github.com/microsoft/onnxruntime/issues/4349.
This would fix segmentation fault caused by models(invalid) having nodes(that need input) with no inputs.


onnx/defs/data_type_utils.cc
```diff
@@ -170,7 +170,10 @@ std::string DataTypeUtils::ToDataTypeString(int32_t tensor_data_type) {
   TypesWrapper& t = TypesWrapper::GetTypesWrapper();
   auto iter = t.TensorDataTypeToTypeStr().find(tensor_data_type);
   assert(t.TensorDataTypeToTypeStr().end() != iter);
-  return iter->second;
+  if (t.TensorDataTypeToTypeStr().end() != iter) {
+    return iter->second;
+  }
+  return std::string();
```
linkerzhang, line 176: this is swallowing "error". An assertion should be added, rather than returning empty string.

prabhat00155: An assertion is already there(https://github.com/onnx/onnx/blob/456ba4cace5b21cbd64b579ebd48e83aa4081b76/onnx/defs/data_type_utils.cc#L172). Assertions are for debugging(works in debug mode: Assertion failed: (t.TensorDataTypeToTypeStr().end() != iter), function ToDataTypeString, file onnx_projects/onnxruntime/cmake/external/onnx/onnx/defs/data_type_utils.cc, line 183.), we should not access `second` if the data type is not present in the hashmap. This causes seg fault(as you can see in the issue linked above).

pranavsharma: Yes, we shouldn't swallow the error. In this case the undefined data type is coming from the model (external input). Typically assert()'s are used to check programming mistakes, not to validate user inputs. I suggest we return an error status code that can be checked by callers. I realize this can be cascading and affect many callers, but provides better semantics by setting clear expectations that this method can return an error.

linkerzhang: If changing the API is not good (affecting many users), throwing exception/error?

prabhat00155: Updated code to throw runtime error.

onnx/defs/data_type_utils.cc
```diff
@@ -112,6 +112,9 @@ const TypeProto& DataTypeUtils::ToTypeProto(const DataType& data_type) {
   std::lock_guard<std::mutex> lock(GetTypeStrLock());
   auto it = GetTypeStrToProtoMap().find(*data_type);
   assert(it != GetTypeStrToProtoMap().end());
```
postrational, line 114: Should we remove the `assert`? It seems redundant now.

onnx/defs/data_type_utils.cc
pranavsharma: If we're throwing an exception, throwing std::invalid_argument would probably more sense. Also, throwing an exception doesn't free us from changing the caller. If the caller doesn't catch it, std::terminate will get called.

onnx/defs/data_type_utils.cc
prabhat00155: Done.

prabhat00155:
I don't have write access to merge this PR.

askhade:
> I don't have write access to merge this PR.

Done

PR #2890, opened 2020-07-09
Fix travis-CI on Mac with Python3 and circleCI with stable vision

jcwchen:
**Description**
Modify travis-CI: use Python 3.6.5 instead of the latest Python 3.8.

**Motivation and Context**
It seems that travis-CI is failing (Mac with Python3) in recent PRs.
Originally `brew upgrade python` used python 3.7. It changes to use 3.8 recently and `mypy` type check fails in Python 3.8. 
Error log is here:
```
 clang -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -Iast27/Include -I/usr/local/Cellar/python@3.8/3.8.3_2/Frameworks/Python.framework/Versions/3.8/include/python3.8 -c ast27/Parser/acceler.c -o build/temp.macosx-10.13-x86_64-3.8/ast27/Parser/acceler.o
  ast27/Parser/acceler.c:13:10: fatal error: 'pgenheaders.h' file not found
  #include "pgenheaders.h"
           ^~~~~~~~~~~~~~~
  1 error generated.
  error: command 'clang' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for typed-ast
  Running setup.py clean for typed-ast
Failed to build typed-ast
Installing collected packages: typed-ast, mypy, onnx
    Running setup.py install for typed-ast ... error
    ERROR: Command errored out with exit status 1:
```


prasanthpul:
I thought we are deprecating Travis CI?

jcwchen:
> I thought we are deprecating Travis CI?

Oh you are right. Thank you for the reminder.

prasanthpul:
but if Travis deprecation is not happening soon, then we should do this for now. @vinitra any ETA?

jcwchen:
Agree. Besides, it seems that circleCI keeps failing because the master branch of https://github.com/pytorch/vision fails... circleCI uses `pytorch/vision` library from the master branch.

vinitra:
Thanks for the PR! Deprecation was happening in https://github.com/onnx/onnx/pull/2773 but work was temporarily paused. It will hopefully go in by the end of the week

askhade:
Travis failure is happening because of the following issue : https://github.com/python/mypy/issues/6545
Copying an excerpt from the above issue here : 

mypy currently supports targeting Python versions other than the version used to run mypy. That's true, and we will still do that for older Python versions. (E.g. you can use Python 3.8 to run mypy targeting Python 2.7 through 3.7.) But if you want to target the latest Python version, you will have to use that latest version to run mypy. I think it's a relatively small price to pay -- after all, if your package claims support for Python 3.8, you probably want to use Python 3.8 to run your unit tests anyway.



Right now your change works because we don't build onnx with 3.8... but once we add this to the CI matrix we will need to make another change to make sure mypy can then work with this new target too... Please add some brief comments to document the same.

.circleci/test.sh
```diff
@@ -6,7 +6,7 @@ source /tmp/venv/bin/activate
 
 # install torchvision from master
 # the one on pypi requires cuda
-pip install -q git+https://github.com/pytorch/vision.git
+pip install -q git+https://github.com/pytorch/vision.git@86b6c3e22e9d7d8b0fa25d08704e6a31a364973b
```
askhade, line 9: add a comment here explaining why this is being done

jcwchen:
> Travis failure is happening because of the following issue : [python/mypy#6545](https://github.com/python/mypy/issues/6545)
> Copying an excerpt from the above issue here :
> 
> mypy currently supports targeting Python versions other than the version used to run mypy. That's true, and we will still do that for older Python versions. (E.g. you can use Python 3.8 to run mypy targeting Python 2.7 through 3.7.) But if you want to target the latest Python version, you will have to use that latest version to run mypy. I think it's a relatively small price to pay -- after all, if your package claims support for Python 3.8, you probably want to use Python 3.8 to run your unit tests anyway.
> 
> Right now your change works because we don't build onnx with 3.8... but once we add this to the CI matrix we will need to make another change to make sure mypy can then work with this new target too... Please add some brief comments to document the same.

Thank you for the investigation. Another thing is, I tried to upgrade python to 3.8 and set it back to 3.6 or 3.7, but it still failed because `mypy` will run by the latest python on Mac OS somehow. Therefore, I chose to brew install from certain old python version (actually it's not a good way to install python on Mac) instead of brew upgrade python first then brew switch.

I agree with you that this solution is just temporary... We definitely need to set it back to the python 3.8 in the future.

jcwchen:
It seems that it encountered an accidental problem. Close and open to trigger the CI again.

jcwchen:
:shipit:

PR #2893, opened 2020-07-13
Define differentiability of Pool/Conv Ops

ma-hei:
This commit adds the differentiability tags to the
MaxPool, AveragePool, RoiPool and Conv operators.


ma-hei:
Differentiability of the Pooling operator is shown by Method 1, "Reuse Existing Deep Learning Frameworks", described here https://github.com/onnx/onnx/pull/2794.

Showing differentiability of MaxPool input/output and export of model from MXnet to ONNX:

```python
from mxnet import gluon, nd, autograd, context
from mxnet.contrib import onnx as onnx_mxnet
import numpy as np

net = gluon.nn.HybridSequential()
net.add(gluon.nn.MaxPool1D(pool_size=2))
net.hybridize()

x = nd.random.uniform(-1,1,(1,3,4))
labels = nd.random.uniform(-1,1,(1,3,2))
L2loss = gluon.loss.L2Loss()

# showing differentiability of input
x.attach_grad()
with autograd.record():
    output = net(x)
    loss = L2loss(output, labels)

loss.backward()
print(x.grad)

# showing differentiability of output
with autograd.record():
    output = net(x)
    output.attach_grad()
    loss = L2loss(output, labels)

loss.backward()
print(output.grad)

net.export("maxpool1d")

sym = './maxpool1d-symbol.json'
params = './maxpool1d-0000.params'
onnx_file = './maxpool1d.onnx'

converted_model_path = onnx_mxnet.export_model(sym, params, [x.shape], np.float32, onnx_file)
``` 

We can see the derivatives of the input and output from the print statements:
```
[[[ 0.          0.05175247  0.          0.1822437 ]
  [ 0.         -0.01127579  0.          0.2635129 ]
  [ 0.         -0.11336635  0.12441261  0.        ]]]
<NDArray 1x3x4 @cpu(0)>

[[[ 0.05175247  0.1822437 ]
  [-0.01127579  0.2635129 ]
  [-0.11336635  0.12441261]]]
<NDArray 1x3x2 @cpu(0)>
```

ma-hei:
Differentiability of the Conv Operator is shown similarly:
```python

net = gluon.nn.HybridSequential()
net.add(gluon.nn.Conv1D(1, 2, weight_initializer = 'ones'))
net.hybridize()
net.initialize(ctx=context.cpu(0))

net_params = net.collect_params('0.weight|0.bias')

x = nd.random.uniform(-1,1,(1, 1, 3))
x.attach_grad()
label = nd.random.uniform(-1,1,(1,1,2))

with autograd.record():
    output = net(x)
    loss = L2loss(output, label)

loss.backward()
print(x.grad)
print(net_params['0.weight'].grad())
print(net_params['0.bias'].grad())

#export of model to onnx
net.export("conv1d")
sym = './conv1d-symbol.json'
params = './conv1d-0000.params'
onnx_file = './conv1d.onnx'

converted_model_path = onnx_mxnet.export_model(sym, params, [x.shape], np.float32, onnx_file)

```
We see the derivatives of the input as well as kernel and bias parameters from the print statements:
```
[[[-0.45106208 -0.21056682  0.24049526]]]
<NDArray 1x1x3 @cpu(0)>

[[[0.09440736 0.1604658 ]]]
<NDArray 1x1x2 @cpu(0)>

[-0.21056682]
<NDArray 1 @cpu(0)>
```

ma-hei:
@wschin could you give some advice on how to reproduce those build failures locally? I see that when you added the differentiability tag to operators in onnx/defs/tensor/defs.cc  (https://github.com/onnx/onnx/pull/2723/files#), you added the parameter "OpSchema::Single" to operators such as Transpose, Scatter, etc. Why was that necesary?

wschin:
>             OpSchema::Optional)

NonDifferentiable. Because `PoolOpSchemaGenerator` is changed, we need to tag all places where `PoolOpSchemaGenerator` is used.

---
Refers to: onnx/defs/nn/defs.cc:368 in c0e1989. [](commit_id = c0e198907ce184cedba99e87e44b25e2d3fca1f2, deletion_comment = False)

onnx/defs/nn/defs.cc
```diff
@@ -672,19 +680,31 @@ std::function<void(OpSchema&)> RoiPoolOpSchemaGenerator(const char* name) {
...
     schema.Output(
         0,
         "Y",
         "RoI pooled output 4-D tensor of shape (num_rois, channels, pooled_shape[0], pooled_shape[1]).",
-        "T");
+        "T",
+        OpSchema::Single,
+        true,
+        1,
+        OpSchema::NonDifferentiable);
```
wschin, line 707: I guess it's differentiable? It's similar to one-to-one mapping.

ma-hei: Yes, definitely. That was a mistake. Regarding the comment above (line 368): This is the indices output of MaxPool (not the actual pooling output). It's just the indices of the max value in the input window (an integer value if I understand correctly). Is it really differentiable?

ma-hei:
@wschin I revised the commit according to your comments. Thanks for catching those two issues.

onnx/defs/nn/defs.cc
wschin: Yes, my understanding is that indexes are not differentiable but the output pooled from input is differentiable. :)

PR #2899, opened 2020-07-15
Extend ControlFlow to allow Sequence type for inputs and outputs

BowenBao:
The update is to support more flexible list operations within loops. With Scan output, much can be supported by doing loop optimization. However, more complex cases requires more flexible loop carried dependencies

Example
```python
x_s = split(input)
outs = []
for x in x_s:
    if some_condition:
        outs.append(x)
return outs
```

After merging https://github.com/onnx/onnx/pull/2581, test cases can be added.


onnx/defs/controlflow/defs.cc
```diff
@@ -590,7 +590,8 @@ ONNX_OPERATOR_SET_SCHEMA(
         .Output(
             0,
             "v_final_and_scan_outputs",
-            "Final N loop carried dependency values then K scan_outputs",
+            "Final N loop carried dependency values then K scan_outputs. "
+            "Scan outputs must be Tensors.",
```
gramalingam, line 594: I think the inference function above does not check that scan-outputs are tensors. Can we make the check stricter for scan outputs?

onnx/defs/controlflow/defs.cc
```diff
@@ -337,9 +337,9 @@ void LoopInferenceFunction(InferenceContext& ctx) {
 
       const bool is_loop_state_var = i < num_loop_state_vars;
 
-      if (!subgraph_output_type->has_tensor_type()) {
+      if (!subgraph_output_type->has_tensor_type() && !subgraph_output_type->has_sequence_type()) {
```
gramalingam, line 340: Is there any reason (specific to the Loop op) to restrict the outputs to be of any type? E.g., why not maps? Or, sequences of sequences? (This is probably allowed by above check, but not by the type-constraint on the op below, which restricts it to sequences-of-tensors.) 

I think allowing the output-type to be unconstrained would require some extensions in the Schema implementation, which, I think, assumes that the set of possible types can be finitely enumerated and stored. It would be worth taking a look to see if this can be generalized. This would avoid having to generalize the Loop op again the next time if we want to allow other types.

onnx/defs/controlflow/defs.cc
```diff
@@ -604,7 +605,15 @@ ONNX_OPERATOR_SET_SCHEMA(
             " if the dimensions or data type of these scan_outputs change across loop"
             " iterations.",
             AttributeProto::GRAPH)
-        .TypeConstraint("V", OpSchema::all_tensor_types(), "All Tensor types")
+        .TypeConstraint(
+            "V",
+            [](){
+              auto t = OpSchema::all_tensor_types();
```
gramalingam, line 611: See comment above. If we could generalize the Schema implementation so that a type-variable "V" can be unconstrained (that is, without any TypeConstraint specified for it), we could omit this type-constraint. I think it involves some non-trivial work to extend this, but it would be really useful and great.

onnx/defs/controlflow/defs.cc
BowenBao: updated

onnx/defs/controlflow/defs.cc
BowenBao: Today there does not exist operators that use/produce such flexible types. I agree to support types like sequence of sequence, we need to update Schema to relax the assumption of types being a finite set. However, I'm not sure if we should start this change with Loop, instead of a real use case. Should such op be added in the future, the infrastructure will have to be updated, and I'm relatively ok to generalize Loop (and If) accordingly in that opset update.

gramalingam: The problem is that when that happens, we will have to bump the Loop operator's version again even though a Loop implementation itself does not really depend on the types of these values. This leads to unnecessary version-updates, which impacts exporters, convertors, backends. A loop is polymorphic over this type V and the constraints on the type V are artificial. (If this had been done correctly for the earlier version, even this extension would not have been needed).

askhade:
Once https://github.com/onnx/onnx/pull/2925 is checked in, updating an op with an existing data type will not require opset version update. @gramalingam since sequence is an existing type this PR will not require version bump right?

gramalingam:
@askhade my understanding is that doesn't apply here, since the inference-code is impacted here.

BowenBao:
@gramalingam @askhade I updated the PR with test cases and extended the scope to include if operator.

onnx/backend/test/case/node/if.py
```diff
@@ -0,0 +1,129 @@
...
+from __future__ import unicode_literals
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+class Loop(Base):
```
gramalingam, line 13: Change name from Loop to If?

onnx/defs/controlflow/defs.cc
```diff
@@ -263,6 +263,23 @@ void IfInferenceFunction(InferenceContext& ctx) {
 
       UnionShapeInfo(
           else_output->tensor_type().shape(), *if_output->mutable_tensor_type());
+    } else if (then_output->has_sequence_type() && then_output->sequence_type().elem_type().has_tensor_type()) {
```
gramalingam, line 266: Could we move this logic (the complete if-then-else) as a function, say "UnionTypeInfo(const TypeProto& type1, TypeProto& type2)" into the shared utility file https://github.com/onnx/onnx/blob/master/onnx/defs/shape_inference.h ? It will help if the same logic is required elsewhere and if we generalize this in the future to handle map-types, etc. as well.

gramalingam: The if-then-else in lines 237 to 245 should also be merged into the same function. The code here needs to just say
```cpp
   *if_output = *then_output;
   UnionTypeInfo(*else_output, *if_output);
```

onnx/defs/shape_inference.h
```diff
@@ -805,4 +913,62 @@ inline void UnionShapeInfo(
...
+          "Mismatched tensor element type:",
+          " source=",
+          source_elem_type,
+          " target=",
+          target_elem_type);
+    }
+
+    UnionShapeInfo(source_type.tensor_type().shape(), *target_type.mutable_tensor_type());
+  } else if (target_type.has_sequence_type()) {
+    if (source_type.sequence_type().elem_type().has_tensor_type() &&
```
gramalingam, line 953: Better to just call "UnionTypeInfo(source_type.sequence_type().elem_type(), target_type.sequence_type().elem_type())" ... that will handle the case correctly in case one is a sequence of tensors and the other is a sequence of non-tensor-type.

onnx/defs/controlflow/defs.cc
BowenBao: Thanks, updated

onnx/defs/shape_inference.h
BowenBao: thanks! makes sense

gramalingam: Thanks for addressing it! I think the nested condition "source_type ... has_tensor_type() && target_type ... has_tensor_type()" should also be removed.

gramalingam: Was that meant to check "has_elem_type()"? I assumed a sequence-type will always have an "elem_type". We could check for that to be safe I suppose ... but we should probably treat it as an error if there is no elem_type.

Issue #2874, opened 2020-07-01
Shape inference doesn't work unless initializer tensors are also listed as inputs


matteosal:
# Bug Report

### Describe the bug
As title says

### System information
- OS Platform and Distribution: Ubuntu 18.04  
- ONNX version: 1.7 release commit
- Python version: 3.6.9
- GCC/Compiler version: 7.5.0
- CMake version: 3.17.3
- Protobuf version: 3.12

### Reproduction instructions
```
import onnx
from onnx import helper, shape_inference
from onnx import TensorProto

print("************************* Without weights as inputs ******************************")
model_path = "without_weights_as_inputs.onnx"
original_model = original_model = onnx.load(model_path)
onnx.checker.check_model(original_model)

print('Model inputs\n{}'.format(original_model.graph.input))
print('\nValue info before shape inference:\n{}'.format(original_model.graph.value_info))

# Apply shape inference on the model
inferred_model = shape_inference.infer_shapes(original_model)

# Check the model and print Y's shape information
onnx.checker.check_model(inferred_model)
print('\nValue info after shape inference:\n{}'.format(inferred_model.graph.value_info))

print("************************* With weights as inputs ******************************")
model_path = "with_weights_as_inputs.onnx"
original_model = original_model = onnx.load(model_path)
onnx.checker.check_model(original_model)

print('Model inputs\n{}'.format(original_model.graph.input))
print('\nValue info before shape inference:\n{}'.format(original_model.graph.value_info))

# Apply shape inference on the model
inferred_model = shape_inference.infer_shapes(original_model)

# Check the model and print Y's shape information
onnx.checker.check_model(inferred_model)
print('\nValue info after shape inference:\n{}'.format(inferred_model.graph.value_info))
```
Model `"without_weights_as_inputs.onnx"` has tensors in the initializer list that are not listed as inputs. Model `"with_weights_as_inputs.onnx"` has those tensors listed as inputs as well. The above script shows that only model `"with_weights_as_inputs.onnx"` has a successful shape inference, while shapes are not inferred for the first one. The [IR documentation](https://github.com/onnx/onnx/blob/master/docs/IR.md) allows initializer tensor not being listed as inputs, stating that
> When an initializer has the same name as a graph input, it specifies a default value for that input. When an initializer has a name different from all graph inputs, it specifies a constant value.

So it seems that shape inference should work on `"without_weights_as_inputs.onnx"`

- ONNX models
[...]

Issue #2873, opened 2020-07-01
Version converter 11 -> 12 fails with "IndexError: Input s is undefined!"


matteosal:
# Bug Report

### Describe the bug
The version converter fails on the attached model (and not becase the adapter is missing)

### System information
- OS Platform and Distribution: Ubuntu 18.04  
- ONNX version: 1.7 release commit
- Python version: 3.6.9
- GCC/Compiler version: 7.5.0
- CMake version: 3.17.3
- Protobuf version: 3.12

### Reproduction instructions
```
import onnx
from onnx import version_converter, helper

# Preprocessing: load the model to be converted.
model_path = "path/to/shared-bn-arrays-v11.onnx"
original_model = onnx.load(model_path)

converted_model = version_converter.convert_version(original_model, 12)
print("done!")
```
The script doesn't get to the final print, but fails with 
```
Traceback (most recent call last):
  File "script.py", line 8, in <module>
    converted_model = version_converter.convert_version(original_model, 12)
  File "/usr/local/lib/python3.6/dist-packages/onnx-1.7.0-py3.6-linux-x86_64.egg/onnx/version_converter.py", line 166, in convert_version
    converted_model_str = C.convert_version(model_str, target_version)
IndexError: Input s is undefined!
```

- ONNX model
[shared-bn-arrays-v11.zip](https://github.com/onnx/onnx/files/4858440/shared-bn-arrays-v11.zip)

### Expected behavior
The script should not fail and properly upgrade the model (or at least fail saying there are no adapters)

askhade:
There is an underlying issue in version converter. It relies on the C++ IR which I believe has not been updated after IR v3. Because of this I think it expects initializers also be added as graph inputs. 
If you try to change the version of your model to IRv3 or create a model with initializers also as inputs then I think this will work.

matteosal:
Yes, by including all the initializers in the graph inputs it works, thanks. So this seems related to [this other issue](https://github.com/onnx/onnx/issues/2874) I created today.

jcwchen:
Fortunately, it can be fixed by correcting `shape_inference` about the IR issue because `version_converter` uses `shape_inference` before the conversion. This PR https://github.com/onnx/onnx/pull/2901 should help.

jcwchen:
It should be resolved by https://github.com/onnx/onnx/pull/2901. Please reopen it if you still encounter this issue. Thanks.

jcwchen:
[...]

Issue #2660, opened 2020-03-15
A tensor which is in graph initializer,but not in graph input list, have no  TypeProto info.  


PenghuiCheng:
I'm runing a DLRM model to inference shape, but I found that Gemm op can't inference output shape, the reason is that Gemm op input 1(which is input B) is in graph initializer list, and not in graph input list. This model is converted from pytorch. I want to know, why input TypeProto list have no initializer information, is it reasonable?

linkerzhang:
initializers do not have to be included in graph inputs. 
1. if initializers are also in graph inputs then they're just default values and can be overridden by user inputs.
2. if initializers are not in graph inputs, they're constants.

The point here should be, the shape inference implementation should be able to use initializer data (shape).

This issue should be the same as #2655 

@TMVector @PenghuiCheng any of you would like to offer a fix for this gap please?

TMVector:
I haven't had time to work on ONNX lately (sorry for the late reply), but I have been using this method as a stopgap at work.
```python
def add_value_info_for_constants(model : onnx.ModelProto):
    """
    Currently onnx.shape_inference doesn't use the shape of initializers, so add
    that info explicitly as ValueInfoProtos.
    Mutates the model.
    Args:
        model: The ModelProto to update.
    """
    # All (top-level) constants will have ValueInfos before IRv4 as they are all inputs
    if model.ir_version < 4:
        return

    def add_const_value_infos_to_graph(graph : onnx.GraphProto):
        inputs = {i.name for i in graph.input}
        existing_info = {vi.name: vi for vi in graph.value_info}
        for init in graph.initializer:
            # Check it really is a constant, not an input
            if init.name in inputs:
                continue

            # The details we want to add
            elem_type = init.data_type
            shape = init.dims

            # Get existing or create new value info for this constant
            vi = existing_info.get(init.name)
            if vi is None:
                vi = graph.value_info.add()
                vi.name = init.name

            # Even though it would be weird, we will not overwrite info even if it doesn't match
            tt = vi.type.tensor_type
            if tt.elem_type == onnx.TensorProto.UNDEFINED:
                tt.elem_type = elem_type
            if not tt.HasField("shape"):
                # Ensure we set an empty list if the const is scalar (zero dims)
                tt.shape.dim.extend([])
                for dim in shape:
                    tt.shape.dim.add().dim_value = dim

        # Handle subgraphs
        for node in graph.node:
            for attr in node.attribute:
[...]

Issue #2655, opened 2020-03-11
Shape inference doesn't use initializer shapes


TMVector:
### Issue description
It looks like shape inference doesn't make use of initializer shapes, so even super simple graphs can fail to gain any useful info from shape inference.

I suppose one caveat should be that shape inference shouldn't rely on initializers which are also inputs, since their shape could definitely change at runtime.

### MRE
Run shape inference with the exact same model, but with and without adding explicit value info for the initializer.

<details>

```python
model = onnx.helper.make_model(onnx.helper.make_graph(
    nodes=[
        onnx.helper.make_node("Gemm", ["x", "m"], ["y"]),
        onnx.helper.make_node("Identity", ["y"], ["z"]),
    ],
    name="graph1",
    inputs=[
        onnx.helper.make_tensor_value_info("x", onnx.TensorProto.FLOAT, ["dim1",7]),
    ],
    outputs=[
        onnx.helper.make_tensor_value_info("z", onnx.TensorProto.FLOAT, ["dim2",13]),
    ],
    initializer=[
        onnx.numpy_helper.from_array(np.zeros((7,13), dtype=np.float32), name="m"),
    ],
    value_info=[
    ],
))
```
**Without explicit value info for initializer `m`**
```
>>> onnx.shape_inference.infer_shapes(model).graph.value_info
[name: "y"
type {
  tensor_type {
    elem_type: 1
  }
}
]
```
**_With_ explicit value info for initializer `m`**
```
>>> model.graph.value_info.extend([onnx.helper.make_tensor_value_info("m", onnx.TensorProto.FLOAT, [7,13])])
>>> onnx.shape_inference.infer_shapes(model).graph.value_info
[name: "m"
type {
  tensor_type {
    elem_type: 1
    shape {
      dim {
        dim_value: 7
      }
      dim {
        dim_value: 13
      }
    }
  }
}
, name: "y"
type {
  tensor_type {
    elem_type: 1
    shape {
      dim {
        dim_param: "dim1"
      }
      dim {
        dim_value: 13
      }
    }
  }
}
]
```

</details>

linkerzhang:
there might be a gap of not using initializer when doing shape inference, though, in terms of interface design, we do have getInputData in InferenceContext to support shape inference with known input data (before running model).

jcwchen:
It should be resolved by https://github.com/onnx/onnx/pull/2901. Please reopen it if you still encounter this issue. Thanks.

PR #2901, opened 2020-07-15
Fix IR gap for shape inference: update input from initializer

jcwchen:
**Description**
Fix the IR gap for shape inference by updating the `graph.input` `ValueProto` from `graph.initializer` if missing.

**Motivation and Context**
https://github.com/onnx/onnx/issues/2655, https://github.com/onnx/onnx/issues/2660, https://github.com/onnx/onnx/issues/2873, https://github.com/onnx/onnx/issues/2874
There is a rule in the latest IR: 
`graph.initializer` does not have to be included in `graph.input`.

However, old IR only considers the `graph.input` for `shape_inference` and `version_converter`.
This PR is fixing this IR gap.


onnx/shape_inference/implementation.cc
```diff
@@ -185,6 +185,30 @@ static void InferShapesImpl(
...
     inputDataByName[tp.name()] = &tp;
+    
+    // This is for testing new IR: some tensors can only exist in initializer and not in input
+    // So shape_inference should make use of initializer shapes
+    // Store initializer shape info in value_info as well
+    TypeProto_Tensor* newTensorFromInitializer;
+    auto iter = valueTypesByName.find(tp.name());
+    // If it already exists in input, simply use the original input
+    if (iter != valueTypesByName.end()) {
+      newTensorFromInitializer = valueTypesByName[tp.name()]->mutable_tensor_type();
```
jcwchen, line 196: If such tenor has been found in input, maybe we don't need to do anything here from the initializer?

onnx/shape_inference/implementation.cc
```diff
@@ -183,8 +183,34 @@ static void InferShapesImpl(
   }
 
   std::unordered_map<std::string, const TensorProto*> inputDataByName;
+  GraphProto inputGraph; // to not modify original graph; use temp graph
```
gramalingam, line 186: I am confused. Why do we need a temp graph? Where is this eventually used?

gramalingam: I guess you are using it for automatically allocating and freeing new ValueInfoProto below? I guess we just need a TypeProto, actually. At the least, it is worth documenting.

gramalingam: But it might be simpler to just use a vector of TypeProto, or something like that, to do the memory management.

onnx/shape_inference/implementation.cc
```diff
@@ -183,8 +183,34 @@ static void InferShapesImpl(
...
     inputDataByName[tp.name()] = &tp;
+    
+    // Support new IR: some tensors can only exist in initializer and not in input
+    // So shape_inference should make use of initializer shapes
+    // Store initializer shape info in value_info as well
+    TypeProto_Tensor* newTensorFromInitializer;
+    auto iter = valueTypesByName.find(tp.name());
+    // If it already exists in input, simply use the original input
+    if (iter != valueTypesByName.end()) {
+      continue; // skip if shape info exists in input
```
gramalingam, line 197: I think it would be useful, in this case, to check that the declared type/shape (from input lis) is compatible with the actual value in the initializer. I suggest creating a new "inferred" type (just like newTensorFromInitializer below) and calling checkShapesAndTypes on the original type and this new inferred type.

onnx/shape_inference/implementation.cc
```diff
@@ -183,8 +183,34 @@ static void InferShapesImpl(
...
+    if (iter != valueTypesByName.end()) {
+      continue; // skip if shape info exists in input 
+      // newTensorFromInitializer = valueTypesByName[tp.name()]->mutable_tensor_type();
+    }
+    // If not, add a new input into temp graph
+    else {
+      ValueInfoProto *newValueForInitializer = inputGraph.add_input();
+      newValueForInitializer->set_name(tp.name());
+      valueTypesByName[tp.name()]= newValueForInitializer->mutable_type();
+      newTensorFromInitializer = valueTypesByName[tp.name()]->mutable_tensor_type();
```
gramalingam, line 205: I think "initializerTensorType" would be a better name for "newTensorFromInitializer" ... it is a tensor-type, not a tensor.

onnx/shape_inference/implementation.cc
```diff
@@ -183,8 +183,34 @@ static void InferShapesImpl(
...
+    // So shape_inference should make use of initializer shapes
+    // Store initializer shape info in value_info as well
+    TypeProto_Tensor* newTensorFromInitializer;
+    auto iter = valueTypesByName.find(tp.name());
+    // If it already exists in input, simply use the original input
+    if (iter != valueTypesByName.end()) {
+      continue; // skip if shape info exists in input 
+      // newTensorFromInitializer = valueTypesByName[tp.name()]->mutable_tensor_type();
+    }
+    // If not, add a new input into temp graph
```
askhade, line 200: this is not the right approach
for ir_version 4 and above allow initializers to not be in graph inputs (this means they are constant) we should simply use the shape from intializer

onnx/shape_inference/implementation.cc
askhade: the behavior varies from ir_version < 4 vs ir_version >=4
In case of ir_version < 4 the check which Rama suggested makes sense
In case of ir_version >=4 the shape from input should be prefer as the value in initializer is merely a default value

gramalingam: A couple of comments: (a) Whatever the IR version, the type of the input-variable should be consistent with the value in the initializer. It is okay if the type of the input-variable has less precise-shape than the initializer, but it should be consistent. If the type of the input-variable is a float-tensor and the initializer is an int-tensor, it is incorrect; similarly, if the shape of the input-variable is [B, 10] while the initializer-value has shape [1, 5] it is incorrect. (b) If the input is present, we should use the shape given in the input-type, and not the more-precise shape from the initializer (even though we can check that they are consistent).

onnx/shape_inference/implementation.cc
jcwchen: > I guess you are using it for automatically allocating and freeing new ValueInfoProto below? I guess we just need a TypeProto, actually. At the least, it is worth documenting.

Yes. I also felt weird about creating a useless graph... Sounds like I can simply use `valueTypesByName[tp.name()]= new TypeProto()` for creating a new TypeProto. Thank you for the good catch.

gramalingam: Yes, that would be better ... but then you need to remember to "free" the allocated TypeProto at the end. (That is why I was referring to memory management.) As long as that happens correctly, this would be good.

onnx/shape_inference/implementation.cc
```diff
@@ -185,6 +186,30 @@ static void InferShapesImpl(
...
+    // set the shape according to the initializer shape info
+    if (tp.dims_size() != 0) {
+      initializerTensorType->set_elem_type(tp.data_type()); // TensorProto_DataType_FLOAT
+      TensorShapeProto* shape = initializerTensorType->mutable_shape();
+      for (int i = 0 ; i < tp.dims_size(); ++i) {
+        shape->add_dim()->set_dim_value(tp.dims(i)); 
+      }
+    }
+    auto iter = valueTypesByName.find(tp.name());
+    // If it already exists in input, check input and initializer is sync
```
skottmckay, line 202: My understanding was that an input may override the shape of the initializer providing a default value, so when you're doing the shape inferencing you can only use shape info from a constant initializer (an initializer is constant if IR < 4, or if it has no matching input if IR >= 4). @gramalingam do you agree?

onnx/shape_inference/implementation.cc
```diff
@@ -183,33 +184,31 @@ static void InferShapesImpl(
...
+      }
+    }
     auto iter = valueTypesByName.find(tp.name());
-    // If it already exists in input, simply use the original input
+    // If it already exists in input, check input and initializer is sync
     if (iter != valueTypesByName.end()) {
-      continue; // skip if shape info exists in input 
-      // newTensorFromInitializer = valueTypesByName[tp.name()]->mutable_tensor_type();
+        checkShapesAndTypes(*initializerTensorType, *valueTypesByName[tp.name()]->mutable_tensor_type());
+        mergeShapesAndTypes(*initializerTensorType, valueTypesByName[tp.name()]->mutable_tensor_type());
```
gramalingam, line 205: We should only do the "check". We should NOT call "merge".

onnx/shape_inference/implementation.cc
gramalingam: Roughly, yes. More precisely: we should NOT do the merge below.

Case 1: If X is both an input and an initializer (regardless of IR version), we should use ONLY the shape information provided in the input type; however, it should be the case that the initializer-value should be consistent with the input-type, so it is okay to CHECK for this consistency. For example, it is okay for the input-type to have a shape [M,N] and the initializer to have shape [100, 256]. It is not okay for the input-type to have shape [M,N] and the initializer to have shape [10, 10, 256]. 

Case 2: X is only an initializer and not an input. (Here IR should be >= 4.) We can use the actual shape of the initializer in the inferencing.

Does that sound right?

gramalingam: @skottmckay : the one point I am not clear about is your comment that an initializer is a constant if IR < 4. I need to check back to see if this was the conclusion we reached for IR < 4.

onnx/shape_inference/implementation.cc
jcwchen: Just created a condition to handle the `ir_version` stuff. Thank you for catching this.

onnx/shape_inference/implementation.cc
skottmckay: ONNX Runtime treats initializers as constant if IR version is < 4 as the initializers had to have matching graph inputs, and to not do so would prevent constant folding running. The ONNX spec doesn't require this though, so I agree with your description of Case 1 and Case 2 handling and that an IR check shouldn't be necessary here.

jcwchen:
@gramalingam @askhade @skottmckay Thank you so much for the detailed review!
To conclude the discussion above:
1. There is no condition for merging input and initializer.
2. If they both exist, the only thing we need to do is to check whether these two are consistent (like rank, dimension mapping).
3. If input exists, we should always take the shape from input.
4. If input doesn't exist and ir_version>=4, we should use the shape from initializer for the input shape.

I suppose the current implementation should meet these rules.
Next, I will write more graph-level tests to test this functionality.

jcwchen:
Add two more changes:
1. If the model uses old ops <= 8 (corresponded IR version is <=3), the shape inference won't update the input shape from the initializer shape.
2. If the model has any opset from other doman (e.g., pytorch-caffe2), the shape inference won't update the input shape from the initializer shape either.

I suppose these two kinds of models should not follow the new IR so they don't need to update input from initializer.

@gramalingam please comment if you have other thought. Thank you!

onnx/shape_inference/implementation.cc
```diff
@@ -185,6 +202,35 @@ static void InferShapesImpl(
...
+    auto iter = valueTypesByName.find(tp.name());
+    // If it already exists in input, check input and initializer is sync
+    // use shape info from input (input has priority over initializer)
+    if (iter != valueTypesByName.end()) {
+        checkShapesAndTypes(*initializerTensorType, *valueTypesByName[tp.name()]->mutable_tensor_type());
+    }
+    // Support IR>=4: some tensors can only exist in initializer and not in input
+    // So shape_inference should make use of initializer shapes
+    // Store initializer shape info in value_info as well    
+    else if (ir_version >= 4 && opset_version >= 9){
```
gramalingam, line 229: I don't understand why we need to look at the opset_version here. Technically, it is okay to use IR version > 4 with opset_version < 9. In this scenario, if we have an initializer that is not an input, we can treat it as a constant and it seems okay to execute the following.

jcwchen: The previous modification without considering the opset in this PR will break some tests in pytorch (graphs converted from `torch.onnx.export`). Those graphs are using old opsets (< 9), but it seems that the ir_version is always the latest one in `torch.onnx.export`. In that case, do we still need to update the input from the initializer? I thought old opset corresponds to old ir_version.

The failure is as follows:
https://circleci.com/gh/onnx/onnx/5368?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
They are using opsets from pytorch.caffe2 domain.

gramalingam: But what causes the checks to fail? Do those models have initializers that are not in the input-list? (Otherwise, the behavior should not be affected by this PR ... does that sound right?) If those models have initializers that are not in the input-list, how should we handle them? They should be handled as constants, right? If something breaks when we treat them as constants, then there must be some *other* issue ... either in the pytorch model exporter or within the onnx checker. Does that make sense?

gramalingam: Looking at "test_while_cond" for example: it appears that the exporter is exporting ops (relational comparison ops) with int types, though ONNX does not have them. So, it is an invalid model, and it is correct for the checker to produce an error message. I think that we need a single general solution to the problem of incorrect ONNX models exported from these backends (for example, using some explicit list of known incorrect models from the checker in CI, or something similar).

jcwchen: Agreed. In fact, these models keep failing in different conditions (https://github.com/onnx/onnx/pull/2783) because they have some opsets which do not come from ONNX. Then the solution will be same as what we discussed before as you mentioned above.

Therefore, I removed the opset version stuff and only consider the opset domain: if the model is using some opsets other than ONNX domain, the shape inference won't update the input from initializer. Besides, I will propose a PR about deprecating the CircleCI and consider to remove those tests (invalid models to ONNX). 

Thank you for the explanation @gramalingam

gramalingam: I think even the opset check is not needed. The initializer semantics has nothing to do with opsets. It is related *only* to IR version. The opset check can lead to unintuitive behavior; if I simply add a new domain "my.special.domain" for one particular node-op, it will suddenly impact the inference/checking for other initializers/nodes/ops which have no connection to "my.special.domain". Thanks.

jcwchen: Understood. Just removed other domain stuff.

onnx/test/shape_inference_test.py
```diff
@@ -3232,6 +3232,80 @@ def test_celu_function_output_shape(self):  # type: () -> None
...
+        initializer_shape = (8, 7)
+        original_model = self.prepare_input_initializer_tensors(initializer_shape, None)
+        original_model.ir_version = 3  # test ir_version < 4
+
+        self.not_infer_empty_input_from_initializer(original_model)
+
+    def test_infer_initializer_input_mismatch(self):  # type: () -> None
+        # Catch error if initializer and input mismatch
+        initializer_shape = (8, 7)
+        input_shape = (-4, 3)
```
gramalingam, line 3283: Dimensions are expected to have non-negative values. "-4" is not a meaningful dimension. Changing it to "4" would be better.

onnx/test/shape_inference_test.py
```diff
@@ -3232,6 +3232,80 @@ def test_celu_function_output_shape(self):  # type: () -> None
...
+
+        inferred_model = onnx.shape_inference.infer_shapes(original_model)
+        # Cannot infer from initializer; nothing in value_info; pop fail
+        self.assertRaises(IndexError, inferred_model.graph.value_info.pop)
+
+    def test_infer_with_initializer_without_input_below_ir4(self):  # type: () -> None
+        # This is for testing IR<4: tensors must exist both in initializer and input
+        # So shape_inference should not make use of initializer shapes
+        initializer_shape = (8, 7)
+        original_model = self.prepare_input_initializer_tensors(initializer_shape, None)
```
gramalingam, line 3275: In IR version 3, we expect every initializer to also be in input. Does this not cause an error? I think it should be better to use (None, None) as input_shape.

onnx/test/shape_inference_test.py
jcwchen: Changed. Thanks

onnx/test/shape_inference_test.py
jcwchen: Using ir_version=3 does not cause an error with current checker.
Using `None` here is to not pass input (empty list) to `make_tensor`.

gramalingam: Hmmm. But it will create a model with x and y as initializer, but not as inputs, in IR version 3. I think it is working because you are calling only shape-inference here ... but the model will not pass this check in checker.cc https://github.com/onnx/onnx/blob/434cf85abde5b4eb3aa537bcbe892b39eed3e9c6/onnx/checker.cc#L679 , I think. So, isn't it an invalid model?

jcwchen:
Added free memory for created `initializertype`.

onnx/shape_inference/implementation.cc
```diff
@@ -473,6 +506,10 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
 
   return graphOutputTypes;
 }
-
+void deleteCreatedTensors(std::vector<TypeProto*> initializerTypeList) {
```
gramalingam, line 509: Thanks for adding this! Minor nit: "deleteCreatedTypes" would be a better name than "deleteCreatedTensors".

jcwchen: Good idea.

onnx/test/shape_inference_test.py
jcwchen: You are right... It will be caught by `ValidationError: x in initializer but not in graph input`.
I will use (None, None) as input to produce a valid model. Just changed this test.
Thank you for the good catch!

Issue #2613, opened 2020-02-19
BN fusion optimizer can break graph.


xkszltl:
Anything calling `addInitializerAndInput` relies on the `next_unique_` to create new node (e.g. fused weight/bias of Conv+BN).
However, if the model is previously written to file, this counter may not be correct.

I'm not sure how the counter is recovered in loading (probably by recreating `Value`), but I have a model file traced from pytorch and the optimizer picked some number already used as name, and break the graph.

Suggest to fix by taking the `max(stoull(name), next_unique_)` in any name-setting functions when the name is a number.

xkszltl:
The current workaround is add a prefix to the followings to free up all the namespace
```
model.graph.initializer.name
model.graph.input[].name
model.graph.output[].name
model.graph.node[].input[]
model.graph.node[].output[]
```

dashesy:
Almost anything with the optimizer is broken. Same issue [here[(https://github.com/onnx/onnx/issues/2401)

xkszltl:
@dashesy This is what I end up with:
```
def tag_namespace(self, model, ns):
    existing  = [i.name for i in model.graph.initializer]
    existing += [i.name for i in model.graph.input      ]
    existing += [i.name for i in model.graph.output     ]
    for node in model.graph.node:
        existing += node.input
        existing += node.output
    for tensor in model.graph.initializer:
        if tensor.name.isdigit():
            tensor.name = "{}_{}".format(ns, tensor.name)
            if tensor.name in existing:
                print("[Warning] Name conflict:", tensor.name)
    for i, value_info in enumerate(model.graph.input):
        if value_info.name.isdigit():
            model.graph.input[i].name = "{}_{}".format(ns, value_info.name)
            if model.graph.input[i].name in existing:
                print("[Warning] Name conflict:", model.graph.input[i].name)
    for i, value_info in enumerate(model.graph.output):
        if value_info.name.isdigit():
            model.graph.output[i].name = "{}_{}".format(ns, value_info.name)
            if model.graph.output[i].name in existing:
                print("[Warning] Name conflict:", model.graph.output[i].name)
    for node in model.graph.node:
        for i, name in enumerate(node.input):
            if name.isdigit():
                node.input[i] = "{}_{}".format(ns, name)
                if node.input[i] in existing:
                    print("[Warning] Name conflict:", node.input[i])
        for i, name in enumerate(node.output):
            if name.isdigit():
                node.output[i] = "{}_{}".format(ns, name)
[...]

PR #2906, opened 2020-07-16
Fix next_unique

okdshin:
When `initializer_list` has numerical name, `next_unique` conflicts with it. The conflict causes breaking graph.
This PR fix the part depending on `next_unique`. See also #2613.


onnx/common/ir.h
```diff
@@ -866,6 +866,29 @@ friend struct Value;
 
   std::vector <OpSetID> opset_versions_;
 
+  bool isNameInAnyNode(const std::string& name) const {
+      auto f = [&name](const Value* v){ return v->uniqueName() == name; };
+      for(const Node* node : all_nodes) {
+          auto found_in = std::find_if(node->inputs().begin(), node->inputs().end(), f);
+          auto found_out = std::find_if(node->outputs().begin(), node->outputs().end(), f);
+          if(found_in != node->inputs().end() || found_out != node->outputs().end()) {
```
askhade, line 874: nit: If you do the check for found_in before checking the node outputs then you can exit early when the name is found in node inputs.

askhade:
Overall looks good to me. There is a small comment I added apart from that once you sign the DCO we can check this in. Thanks!

onnx/common/ir.h
```diff
@@ -866,6 +866,29 @@ friend struct Value;
 
   std::vector <OpSetID> opset_versions_;
 
+  bool isNameInAnyNode(const std::string& name) const {
```
gramalingam, line 869: What happens when we have nested-scopes produced via control-flow operators? Don't you want to ensure uniqueness handles that too?

daquexian:
@askhade @gramalingam Since it is important to onnx optimizer, I updated it myself in https://github.com/onnx/onnx/pull/3071

daquexian:
The content of this PR has been merged as #3071.

PR #2908, opened 2020-07-17
fixes for linux and mac ci pipelines

askhade:
In Azurepipelines all pipeline failures are not caught, with Travis CI no more being required for PR merge this is a big issue as the PRs are not accurately validated. This PR fixes error reporting for mac and linux pipelines and adds missing checks from travis pipeline to azurepipelines

PR #2910, opened 2020-07-17
Enable LITE_PROTO, ml, debug, warning as error, align with Travis in Azure CIs

vinitra:
Updating matrix of possibilities for Linux and Mac CIs to align with and eventually deprecate Travis.
Follow on from #2908, should be merged before #2773.


.azure-pipelines/MacOS-CI.yml
```diff
@@ -33,7 +54,7 @@ jobs:
       export ONNX_BUILD_TESTS=1
       export DEBUG=${onnx_debug}
       export ONNX_ML=${onnx_ml}
-      export CMAKE_ARGS="-DONNX_USE_LITE_PROTO=ON -DONNXIFI_DUMMY_BACKEND=ON"
+      export CMAKE_ARGS=-DONNX_USE_LITE_PROTO=${onnx_lite} -DONNXIFI_DUMMY_BACKEND=ON
```
askhade, line 57: quotation marks are required otherwise it does not recognize all the args. This is from MacOS CI for this PR
/Users/runner/work/_temp/d95fff7f-6a8a-44fd-a5d0-8d57263db9dc.sh: line 10: export: `-DONNXIFI_DUMMY_BACKEND=ON': not a valid identifier

.azure-pipelines/MacOS-CI.yml
```diff
@@ -10,13 +10,34 @@ jobs:
...
         python.version: '3.6'
         onnx_ml: 1
-    maxParallel: 3
+        onnx_debug: 0
+        onnx_lite: OFF
+      py36-ml-debug:
+        python.version: '3.6'
+        onnx_ml: 1
+        onnx_debug: 1
+        onnx_lite: OFF
```
askhade, line 39: ${onnx_lite} is empty. I think you need to put OFF or ON inside quotation marks

This is from MACOS CI for this PR
ONNX version          : 1.7.0
--   ONNX NAMESPACE        : ONNX_NAMESPACE_FOO_BAR_FOR_CI
--   ONNX_BUILD_TESTS      : ON
--   ONNX_BUILD_BENCHMARKS : OFF
--   ONNX_USE_LITE_PROTO   : 
--   ONNXIFI_DUMMY_BACKEND : OFF
--   ONNXIFI_ENABLE_EXT    : OFF

.azure-pipelines/Linux-CI.yml
```diff
@@ -47,12 +57,11 @@ jobs:
       export ONNX_ML=${onnx_ml}
       export DEBUG=${onnx_debug}
       export ONNX_BUILD_TESTS=1
-      export CMAKE_ARGS="-DONNXIFI_DUMMY_BACKEND=ON"
+      export CMAKE_ARGS=-DONNXIFI_DUMMY_BACKEND=ON -DONNX_USE_LITE_PROTO=${onnx_lite}
```
askhade, line 60: quotation marks are needed here. without it the second arg is not recognized

.azure-pipelines/MacOS-CI.yml
vinitra: Thanks! I've fixed this in the latest update.

.azure-pipelines/Linux-CI.yml
```diff
@@ -11,23 +11,33 @@ jobs:
...
       py37:
         python.version: '3.7'
         onnx_ml: 0
         onnx_debug: 0
+        onnx_lite: "OFF"
       py37-ml:
         python.version: '3.7'
         onnx_ml: 1
         onnx_debug: 0
+        onnx_lite: "OFF"
```
askhade, line 34: this still did not take effect.... I think you need single quotation 'OFF'

vinitra: Thanks for letting me know! Will try that.

.azure-pipelines/MacOS-CI.yml
```diff
@@ -10,13 +10,34 @@ jobs:
...
         python.version: '2.7'
         onnx_ml: 0
+        onnx_debug: 0
+        onnx_lite: 'OFF'
+      py27-ml:
+        python.version: '2.7'
+        onnx_ml: 1
+        onnx_debug: 0
+        onnx_lite: 'OFF'
+      py27-ml-lite:
```
askhade, line 20: python 2.7 is going to be deprecated soon.. can you simply add the LITE option to Py36-ml?

vinitra: added!

.azure-pipelines/MacOS-CI.yml
```diff
@@ -33,7 +54,7 @@ jobs:
       export ONNX_BUILD_TESTS=1
       export DEBUG=${onnx_debug}
       export ONNX_ML=${onnx_ml}
-      export CMAKE_ARGS="-DONNX_USE_LITE_PROTO=ON -DONNXIFI_DUMMY_BACKEND=ON"
+      export CMAKE_ARGS='-DONNXIFI_DUMMY_BACKEND=ON -DONNX_USE_LITE_PROTO='${onnx_lite}
```
askhade, line 52: I dont think it is still working : 

ONNX version          : 1.7.0
--   ONNX NAMESPACE        : ONNX_NAMESPACE_FOO_BAR_FOR_CI
--   ONNX_BUILD_TESTS      : ON
--   ONNX_BUILD_BENCHMARKS : OFF
--   ONNX_USE_LITE_PROTO   : 
--   ONNXIFI_DUMMY_BACKEND : ON

askhade: may be try something like this : 

if [ "${onnx_lite}" == "ON" ]; then
  export CMAKE_ARGS="${CMAKE_ARGS} -DONNX_USE_LITE_PROTO=ON"
fi

snnn:
BTW, if ONNX is dynamically link to protobuf, and if ONNX is a share lib(e.g. an python extension), then it should use the protobuf lite, otherwise if something else in the same progress want to use protobuf and onnx too, it will cause conflict and won't work. The dynamic version of libprotobuf full(not lite) requires all the *.proto files be compiled in a single DLL(or .so) file.

.azure-pipelines/Linux-CI.yml
```diff
@@ -37,22 +47,30 @@ jobs:
...
       createCustomEnvironment: true
       environmentName: 'py$(python.version)'
-      packageSpecs: 'python=$(python.version) protobuf'
+      packageSpecs: 'python=$(python.version)'
+      installOptions: '-c conda-forge protobuf'
 
   - script: |
       python -m pip install --upgrade pip
       sudo apt-get install -qq -o=Dpkg::Use-Pty=0 -y --no-install-recommends dos2unix
+      sudo apt-get install -y --no-install-recommends protobuf-compiler libprotoc-dev
```
askhade, line 56: Is this needed? given you are adding it in conda on line 51

vinitra: It was needed to resolve CI errors, even with the conda-forge install. The CI was failing because it could not identify the "protobuf compiler" without that line.

vinitra: Referenced run: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=3188&view=logs&jobId=2105a913-e5a3-5f74-1c30-4f8b7a56f2db&j=2105a913-e5a3-5f74-1c30-4f8b7a56f2db&t=c2afa4d8-3a4c-5e2d-1542-4328675067bd

snnn: Then you could run into protobuf mismatch problem. You'll use the protoc from the operating system(protobuf-compiler package), but libprotobuf from conda(based on your description). Otherwise you don't need two. 

If the could not identify the "protobuf compiler", please check if your PATH env variable has setup correctly. Please reference conda's document to see which env vars need be updated.

askhade: I looked at your reference run. The protobuf package is not being installed in the conda environment that is why it cant find it.

snnn: Ideally, protoc, libprotobuf and protobuf python package, all these three components should have the same version. But it seems ok if the python one has higher version than the other two.

vinitra: I've removed the two dependencies (libprotoc-dev and protobuf-compiler) and used a conda install -c conda-forge protobuf line in the script directly (like in the CI for onnxmltools -- https://github.com/onnx/onnxmltools/pull/207).

vinitra: It seems like the install error is a possible bug with azure pipelines' conda setup task. Thanks for the advice!

snnn:
A tiny suggestion: this is getting more complex, and I feel it could be avoided. 

First,  ONNX must have a pipeline that based on manylinux docker images, because the official packages must be from there. And that one doesn't have conda.  If the CI pipeline largely differ with the packaging pipeline, it will create trouble for you. You may often need to pay extra time on repairing the packaging pipeline. 

(Please feel free to ignore this)

askhade:
> A tiny suggestion: this is getting more complex, and I feel it could be avoided.
> 
> First, ONNX must have a pipeline that based on manylinux docker images, because the official packages must be from there. And that one doesn't have conda. If the CI pipeline largely differ with the packaging pipeline, it will create trouble for you. You may often need to pay extra time on repairing the packaging pipeline.
> 
> (Please feel free to ignore this)

Agree with Changming the CI pipelines should be close to the packaging pipelines. Let's use a different PR to do this work. Goal of this PR is to bring the azurepipelines on par with the Travis pipeline so that Travis can be deprecated.

askhade:
Based on our offline discussion following needs to be fixed/verified in this PR
1. Make sure onnx_ml=0 setting is working. Right now onnx_ml=0 is not set properly so this never gets set in the CI. More importantly do we need this at all? Can it always be just set to 1?
2. Debug setting is actually working ( it think you already fixed this)
3. Linux and Mac CIs have the "treat warning as error flag turned on" just like Travis CI
4. Right now git diff --exit-code sets the exit code but azure pipeline does not treat it s error. Add exit code check for the git diff tests

.azure-pipelines/Windows-CI.yml
```diff
@@ -67,5 +67,5 @@ jobs:
       pip install --quiet -e .[mypy]
       python setup.py typecheck
 
-      git diff --exit-code
+      git diff --exit-code  -- . ':(exclude)onnx/onnx-data.proto' ':(exclude)onnx/onnx-data.proto3'
```
askhade, line 70: on windows platform you dont nee quotation marks
git diff --exit-code  -- . :(exclude)onnx/onnx-data.proto :(exclude)onnx/onnx-data.proto3
should work

.azure-pipelines/Linux-CI.yml
vinitra: Hi @askhade, @snnn   -- I can't seem to get the CIs to pass without these two dependencies. It's not finding them in the source package, even when I move the conda install -c conda-forge protobuf line in this block. Is it ok to install these dependencies as well?

askhade: As snnn explained above this might create other issues because you will end up using protoc and libprotobuf from different sources. Can you check whether protobuf compiler was indeed installed? If yes then fixing the PATH should resolve this issue otherwise can you simply get rid of conda? can you use virtualenvv for env management?

.azure-pipelines/Windows-CI.yml
```diff
@@ -67,5 +67,5 @@ jobs:
       pip install --quiet -e .[mypy]
       python setup.py typecheck
 
-      git diff --exit-code
+      git diff --exit-code  -- . :(exclude)onnx/onnx-data.proto :(exclude)onnx/onnx-data.proto3
```
jcwchen, line 70:
```suggestion
      git diff --exit-code  -- . :(exclude)onnx/onnx-data.proto :(exclude)onnx/onnx-data.proto3
      IF NOT %ERRORLEVEL% EQU 0 (
        @echo "git diff returned failures"
        EXIT 1
      )      
```

jcwchen: Catch the error for Windows

jcwchen (changes requested):

.azure-pipelines/Windows-CI.yml
vinitra: thanks for the suggestion!

PR #2911, opened 2020-07-17
Adding differentiability tags to Math operators, needed for ONNX Training

sveta-levitan:
Added the tags to about half of all Math operators.
Most math operations are clearly differentiable, at least in most points and when values are float or double.
Some can be discussed and argued either way.
I originally marked "floor" and "ceiling" as non-differentiable because their derivatives are 0 in most points and not defined for integer arguments, but PyTorch reports their derivatives as 0, and Wei-Sheng recommends marking them as differentiable.
Also, I assumed that the divisor in "mod" is fixed and hence not differentiable, let me know if that is incorrect.
@wschin

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -365,7 +365,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .TypeConstraint(
             "T",
             {"tensor(float16)",
@@ -2306,7 +2306,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .Input(
             1, 
             "a_scale", 
@@ -2315,7 +2315,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .Input(
             2, 
             "a_zero_point", 
@@ -2324,7 +2324,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .Input(
             3, 
             "b", 
@@ -2333,7 +2333,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .Input(
             4, 
             "b_scale", 
@@ -2342,7 +2342,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .Input(
             5, 
             "b_zero_point", 
@@ -2351,7 +2351,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .Input(
             6, 
             "y_scale", 
@@ -2360,7 +2360,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .Input(
             7, 
             "y_zero_point", 
@@ -2369,7 +2369,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .Output(
             0, 
             "y", 
@@ -2378,7 +2378,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             OpSchema::Single,
             true,
             1,
-            OpSchema::Differentiable)
+            OpSchema::NonDifferentiable)
         .TypeConstraint(
             "T1",
             {"tensor(int8)", "tensor(uint8)"},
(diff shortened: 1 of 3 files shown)


wschin:
Yes, your description on `Mod` is correct. I also did a small experiment for you using Pytorch.

```python
import torch
import torch.nn as nn

# A single-operator model. It's literally a Pytorch Reshape.
# Note that Pytorch Reshape can be directly mapped to ONNX Reshape.
class MyModel(nn.Module):
  def __init__(self):
    super(MyModel, self).__init__()

  def forward(self, x, t):
    y = torch.fmod(x, 2)
    y.retain_grad()
    return y

model = MyModel()

x = torch.tensor([[1., -1], [1., 1.]], requires_grad=True)
t = torch.tensor([[1., 2], [3., 4.]], requires_grad=True)
y = model(x, t)
dy = torch.tensor([[1., 2.], [3., 4.]])

torch.autograd.backward([y],
  grad_tensors=[dy],
  retain_graph=True,
  create_graph=True,
  grad_variables=None)

# This example shows the input and the output in Pytorch are differentiable.
# From the exported ONNX model below, we also see that "x" is the first input
# of ONNX Reshape and "y" the output of ONNX Reshape. Therefore, we can say
# the first input and the output of ONNX Reshape are differentiable.
print(x.grad)
print(t.grad)
print(y.grad)

with open('model.onnx', 'wb') as f:
  torch.onnx.export(model, (x, t), f, opset_version=11)
```

sveta-levitan:
Here is Pytorch code for checking differentiablity of Floor:
```
import torch
import torch.nn as nn

# A one-operator model for checking differentiability of Floor in Pytorch.
class MyModel(nn.Module):
  def __init__(self):
    super(MyModel, self).__init__()

  def forward(self, x):
    y = torch.floor(x)
    y.retain_grad()
 
    return y

model = MyModel()

x = torch.tensor([[1.1, -1.1], [1.2, 1.9]], requires_grad=True)
y = model(x)
y.retain_grad()
dy = torch.tensor([[1., 2.], [3., 4.]])

torch.autograd.backward([y],
  grad_tensors=[dy],
  retain_graph=True,
  create_graph=True,
  grad_variables=None)

# This example shows the input and the output in Pytorch are differentiable.
# From the exported ONNX model below, we also see that "x" is the input
# "y" the output of ONNX Reshape. Therefore, we can say
# the input and the output of ONNX Floor are differentiable.
print("x.grad:", x.grad)
print("y.grad:", y.grad)

with open('modelFloor.onnx', 'wb') as f:
  torch.onnx.export(model, (x), f, opset_version=11)
```

sveta-levitan:
For the operation Clip the min and max arguments must be scalars in PyTorch clamp, hence they can't have gradients, are non-differentiable. The first input and the output are differentiable.

onnx/defs/math/defs.cc
```diff
@@ -280,8 +386,22 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-        .Input(0, "X", "Input tensor", "T")
-        .Output(0, "Y", "Output tensor", "T")
+        .Input(0, 
+            "X", 
+            "Input tensor", 
+            "T",
+            OpSchema::Single,
+            true,
+            1,
+            OpSchema::NonDifferentiable)
```
wschin, line 396: It's partial differentiable. I am ok to make it non-differentiable for a more conservative setting. If needed, changing from conservative (non-differentiable) to aggressive (differentiable) won't cost us too much.

onnx/defs/math/defs.cc
```diff
@@ -258,8 +350,22 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Output(0, 
+            "Y", 
+            "Output tensor", 
+            "T",
+            OpSchema::Single,
+            true,
+            1,
+            OpSchema::Differentiable)
```
wschin, line 368:
```suggestion
            OpSchema::NonDifferentiable)
```
It should be consistent to `Ceil`'s tags.

onnx/defs/math/defs.cc
```diff
@@ -1753,25 +2432,55 @@ ONNX_OPERATOR_SET_SCHEMA(
     10,
     OpSchema()
         .SetDoc(MatMulInteger_ver10_doc)
```
wschin, line 2434: Let's make this op entirely non-differentiable because all its inputs and outputs are integer tensors.

onnx/defs/math/defs.cc
```diff
@@ -2271,22 +3028,41 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "target",
             "Target tensor of shape (N) or (N, d1, d2, ..., dk). Target element value shall be in range of [0, C). "
             "If ignore_index is specified, it may have a value outside [0, C) and the target values should either be "
             "in the range [0, C) or have the value ignore_index.",
-            "Tind")
+            "Tind",
+            OpSchema::Single,
+            true,
+            1,
+            OpSchema::Differentiable)
```
wschin, line 3046:
```suggestion
            OpSchema::NonDifferentiable)
```
This field contains labels in classification problems.

onnx/defs/math/defs.cc
```diff
@@ -2271,22 +3028,41 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "weight",
             "Optional rescaling weight tensor. "
             "If given, it has to be a tensor of size C. Otherwise, it is treated as if having all ones.",
             "T",
-            OpSchema::Optional)
-        .Output(0, "loss", "The negative log likelihood loss", "T")
+            OpSchema::Optional,
+            true,
+            1,
+            OpSchema::Differentiable)
```
wschin, line 3056: I also want to make this non-differentiable because I never seen anyone trained this tensor.

onnx/defs/math/defs.cc
```diff
@@ -1691,15 +2298,87 @@ ONNX_OPERATOR_SET_SCHEMA(
     10,
     OpSchema()
         .SetDoc(QLinearMatMul_ver10_doc)
```
wschin, line 2300: Let's also make all its inputs and outputs non-differentiable because they are integer tensors. The differentiable of integer doesn't exist in math so I prefer to go with a conservative setting here.

onnx/defs/math/defs.cc
sveta-levitan: They are integer, but that is just a compact representation, I think they should still be differentiable. Even double precision numbers do not provide completely continuous range of values. If we make those non-differentiable, then quantized training won't be possible.

wschin (approved):

Issue #2912, opened 2020-07-20
Backend tests per operator version?


alrevuelta:
# Ask a Question

### Question
We are implementing a [pure C99 runtime](https://github.com/alrevuelta/cONNXr) and we are using the onnx backend tests in `onnx/onnx/backend/test/case/node/`, but we have some questions:
* Which operator version is being tested in the `node` test cases? I can't see that a specific one is provided, so I assume that it takes the latest one available? Wouldn't be better to have tests per operator version per operator?
* Assuming that the latest operator version is taken, is this right? What if I generate the `.onnx` files with `cmd_tools.py` using a older onnx version? In this case I would be generating models for an older version that might not match the test vectors provided.
* Do you foresee any major changes in the testing `test/case/node` area in the short future?

### Further information
-backend, test.

- Is this issue related to a specific model?  
NA

### Notes
Any additional information, code snippets.

snnn:
You can fetch all the history version of ONNX and use cmd_tools.py to re-generate the test data, it will be fine except the tool doesn't handle onnx model zoo models very well.

alrevuelta:
I guess it would do the trick, but would be nice to have the tests organised per version right in place. Is there any official mapping between the release and the operator set version? I mean, [here](https://github.com/onnx/onnx/releases/tag/v1.7.0) I can see that for example that the operator set 12 was introduced in v1.7.0 release. How can I do this mapping?

Would you accept a PR that walks through all onnx versions, gets the test cases per operator version and places them in `onnx/onnx/backend/test/case/node/`. So in the future we can keep adding tests per operator version?

Thanks!

snnn:
Sounds great. @askhade , what do you think?

askhade:
Answering your questions - 

- Which operator version is being tested in the node test cases? I can't see that a specific one is provided, so I assume that it takes the latest one available? Wouldn't be better to have tests per operator version per operator?

The test data is generated based on the latest opset for the node. Everytime an op is updated the test is also updated.

- Assuming that the latest operator version is taken, is this right? What if I generate the .onnx files with cmd_tools.py using a older onnx version? In this case I would be generating models for an older version that might not match the test vectors provided.

Yes the latest operator version is taken and this is the intended behavior. I don't think cmd_tools.py has the option to generate data for older version. If we add this option then we will need to verify schema or run onnx checker to validate the generated model

- Do you foresee any major changes in the testing test/case/node area in the short future?
[...]

PR #2918, opened 2020-07-22
Add programmatic access to version-table and extend make-model

gramalingam:
The existing version of make-model uses the latest IR-VERSION as the IR version if the caller does not explicitly supply an IR version. This PR makes the version-table available for programmatic access (in Python) and provides an extended version of make-model that generates the minimum required version from the opsets specified (on a best-effort basis).


onnx/test/helper_test.py
```diff
@@ -271,6 +271,17 @@ def test_model_metadata_props(self):  # type: () -> None
...
+    def test_model_irversion(self):  # type: () -> None
+        def test(opset_version, ir_version):  # type: (int, int) -> None
+            graph = helper.make_graph([], "my graph", [], [])
+            model = helper.make_model_gen_version(graph, opset_imports=[helper.make_opsetid("", opset_version)])
+            self.assertEqual(model.ir_version, ir_version)
+        # opset version 9 requires minimum ir_version 4, etc.
+        test(9, 4)
+        test(10, 5)
+        test(11, 6)
+        test (12, 7)
```
jcwchen, line 283:
```suggestion
        test(12, 7)
```

gramalingam:
> can you add a cross reference between this script and the versioning doc at https://github.com/onnx/onnx/blob/master/docs/Versioning.md#released-versions
> 
> Whenever one is updated, the other needs to be updated too.

Done, thanks.

onnx/test/helper_test.py
```diff
@@ -271,6 +271,17 @@ def test_model_metadata_props(self):  # type: () -> None
         dupe.value = 'Other'
         self.assertRaises(checker.ValidationError, checker.check_model, model_def)
 
+    def test_model_irversion(self):  # type: () -> None
+        def test(opset_version, ir_version):  # type: (int, int) -> None
+            graph = helper.make_graph([], "my graph", [], [])
+            model = helper.make_model_gen_version(graph, opset_imports=[helper.make_opsetid("", opset_version)])
+            self.assertEqual(model.ir_version, ir_version)
+        # opset version 9 requires minimum ir_version 4, etc.
+        test(9, 4)
```
askhade, line 280: can you add some tests with multiple imports... 
another questions : Is this acceptable ai.onnx 11, ai.onnx.ml 2. ai.onnx.training 1 ? Can you add such a test case with expected result

alrevuelta:
This is very useful for #2912 thanks!. Just one concern. Regarding `VERSION_TABLE` that comes from [here](https://github.com/onnx/onnx/blob/master/docs/Versioning.md#released), is the `v1.2` entry right?

I can't find `v1.2` neither in [onnx PyPi](https://pypi.org/project/onnx/#history) nor in the [GitHub Releases](https://github.com/onnx/onnx/releases). The alternatives are `v1.2.1` or `v1.2.2`.

Since @prasanthpul mentioned this:

> can you add a cross reference between this script and the versioning doc at https://github.com/onnx/onnx/blob/master/docs/Versioning.md#released-versions
Whenever one is updated, the other needs to be updated too

I thought it would be appropriate to bring this up here.

onnx/test/helper_test.py
gramalingam: I added more tests. About the second question: this will simply choose the "maximum" among the "minimum IR version" required for each opset in the list. So, it is acceptable. The rationale for this is that we should keep the IR version backward compatible as much as possible. Any model that is valid in IR version N should be a valid model in IR version n+1.

gramalingam:
@alrevuelta : that's a good question. I merely copied the text document over. May be this is the history, and perhaps not all versions were published. @prasanthpul , do you know? Should we do something different for those versions?

onnx/helper.py
```diff
@@ -16,6 +16,57 @@
 from typing import Text, Sequence, Any, Optional, Dict, Union, TypeVar, Callable, Tuple, List, cast
 import numpy as np  # type: ignore
 
+VersionRowType = Union[Tuple[Text, int, int, int], Tuple[Text, int, int, int, int]]
+VersionTableType = List[VersionRowType]
+
+# This is a copy of the documented version in https://github.com/onnx/onnx/blob/master/docs/Versioning.md#released-versions
+# Both must be updated whenever a new version of ONNX is released.
```
askhade, line 23: This table needs to be updated before every release right? Can you change this to something like : This must be updated before creating release candidate packages... 
Exactly when the doc is updated does not matter as much as when this table is updated.

onnx/defs/schema.h
```diff
@@ -865,22 +866,39 @@ class OpSchemaRegistry final : public ISchemaRegistry {
...
+
     // Add customized domain to min/max version.
     // Onnx partners are able to use onnx operator schema api to
     // register customized op in their own domain.
     void AddDomainToVersion(
         const std::string& domain,
         int min_version,
-        int max_version) {
+        int max_version,
+        int last_release_version = -1) {
```
askhade, line 893: since this method is used to add custom domains to this map, not sure last_release_version  makes sense here...

onnx/defs/schema.h
```diff
@@ -889,6 +907,10 @@ class OpSchemaRegistry final : public ISchemaRegistry {
     // Key: domain. Value: <lowest version, highest version> pair.
     std::unordered_map<std::string, std::pair<int, int>> map_;
 
+    // Key: domain. Value: most recent release opset version. Note that
+    // the highest opset version may be ahead of the most recent release's opset version.
+    std::unordered_map<std::string, int> last_release_version_map_;
+
```
askhade, line 913: can you also add an api to get the last release version number

onnx/defs/schema.h
gramalingam: Isn't it possible that the same scenario arises for custom-domains? Namely, the "last official release" vs "current master branch"? Why expose a different API for onnx-domains and custom-domains? If we can keep them as similar/close to each other as possible, it seems better ... it will avoid unexpected issues later on, I think.

onnx/defs/schema.h
gramalingam: Interesting that this does not already exist. It seems more broadly usable, so let me create a separate "version.h" in onnx/onnx/common.

onnx/defs/schema.h
askhade: We can keep this here... I was just wondering whether this will ever be used... last release version makes sense for onnx domains because when backends pick the domain and version map thay want to be able to distinguish between the released vs yet to be released versions... I dont see how this is applicable for custom domains.

Having said that I agree with your point that keeping the apis as similar as possible is better. Can you add a comment to this method with some description for this last_release_version param.

askhade:
Circle CI is fixed now. Merging will master should resolve the failure

onnx/defs/schema.h
gramalingam: Added comment. My thinking is that if we use "my.custom.domain", it could have its own release cycle independent of ONNX. But it would still make sense to talk about the "last released opset version" of "my.custom.domain".

onnx/helper.py
gramalingam: I am not sure I understand the distinction you want to make. It seems important that both are updated. It seems better to update both at the same time (ideally). As to the "when" question: whenever we start a new opset version (for ai.onnx or ai.onnx.ml etc.), it seems to me we should update all of these ... does that sound right?

askhade: I got confused by the wording:
Both must be updated whenever a new version of ONNX is **released** 
This made me think that the updates need to happen before a release and that is why I made the comment that it should be clear that the update needs to happen before the release package is created (with documentation 1 can always update them after the release)

What you said now in the comment makes more sense to me - 
whenever we start a new opset version (for ai.onnx or ai.onnx.ml etc.), it seems to me we should update all of these

docs/Versioning.md
```diff
@@ -169,3 +169,7 @@ ONNX version|File format version|Opset version ai.onnx|Opset version ai.onnx.ml|
 1.5.0|5|10|1|-
 1.6.0|6|11|2|-
 1.7.0|7|12|2|1
+
+A programmatically accessible version of the above table is available [here](../onnx/helper.py). Limited version number
```
ebarsoum, line 173: Why we need a text here? You should update "https://github.com/onnx/onnx/blob/master/docs/OnnxReleases.md".

onnx/common/version.h
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) ONNX Project Contributors.
+// Licensed under the MIT license.
+
+#pragma once
+
+namespace ONNX_NAMESPACE {
+
+// Represents the most recent release version. Updated with every release.
+constexpr const char* LAST_RELEASE_VERSION = "1.7.0";
```
ebarsoum, line 9: Can we read it from: https://github.com/onnx/onnx/blob/master/VERSION_NUMBER?

docs/Versioning.md
gramalingam: Thanks for the suggestion. I updated the OnnxReleases.md file also.

onnx/common/version.h
gramalingam: We can't read that file at run-time, right? The one thing we could do is to auto-generate one file from the other. But that seems non-trivial and not worth it (at this point).

gramalingam: (Summarizing an offline discussion.) Can we make auto-generation of these files a separate task and do it in a separate PR? It will be better to decide on a solution for all the version-numbers in multiple files, instead of just for this one file.

askhade:
@ebarsoum @gramalingam what is the status on this? The changes look good to me and I am ok with adding auto-generation as part of a separate PR.

gramalingam:
I am also fine with completing this PR.

alrevuelta:
I mentioned this some weeks ago but it didn't got much attention. Can someone have a look? Is it right?

> This is very useful for #2912 thanks!. Just one concern. Regarding VERSION_TABLE that comes from here, is the v1.2 entry right?
I can't find v1.2 neither in onnx PyPi nor in the GitHub Releases. The alternatives are v1.2.1 or v1.2.2.

PR #2922, opened 2020-07-24
Allow recurrent operations to be batchwise (reprise)

matteosal:
This is a continuation of this old PR: https://github.com/onnx/onnx/pull/2284

The goal is to add a representation of recurrent operators in which the batch dimension comes before the sequence dimension in the input and output shapes. This representation is enabled by setting the new attribure `time_major` to 0 (default 1 for backward compatibility).

The previous PR also introduced a breaking change, changing the default order of the dimensions for all inputs and outputs so that the `num_directions dim` comes last (if present). It seems to me that such change is not acceptable, because it doesn't affect just the operator but also the other ops attached to its inputs and outputs. Unless the version converter is also able to add `Transpose` operators in the right places, this change will render old models invalid after the recurrent op is upgraded. [This comment](https://github.com/onnx/onnx/pull/2284#issuecomment-545061703) to the old PR suggested to introduce another attribute to regulate this behaviour, which sounds reasonable to me. So what should I do with this one?


postrational:
There seems to be a merge conflict in this PR, which removes the `(differentiable)` annotations from inputs and outputs.

Also, is there a reason `third_party/pybind11` was bumped?

Merge conflicts in other files need to be resolved and we should check why CI isn't passing.

onnx/defs/rnn/defs.cc
```diff
@@ -58,6 +66,15 @@ std::function<void(OpSchema&)> RNNDocGenerator(const char* /*name*/) {
         "Must be one of forward (default), reverse, or bidirectional.",
         AttributeProto::STRING,
         std::string("forward"));
+    schema.Attr(
+        "batch_major",
```
wschin, line 70: I am not sure if this makes sense to everyone --- maybe we can use more general flags `batch_axis` (type: int64) and `sequence_axis` (type: int64)? For current setting, `batch_axis=1` and `sequence_axis=0`.

postrational:
Let's make sure we have a shape inference test for the reverse case.

onnx/test/shape_inference_test.py
```diff
@@ -976,6 +976,21 @@ def _rnn_bidirectional(self, seqlen, batchsize, inpsize, hiddensize):  # type: (
             make_tensor_value_info('all', TensorProto.FLOAT, (seqlen, 2, batchsize, hiddensize)),
             make_tensor_value_info('last', TensorProto.FLOAT, (2, batchsize, hiddensize))])
 
+    def test_rnn_batchwise(self):  # type: () -> None
+        self._rnn_batchwise(64, 32, 10, 4)
+
+    def _rnn_batchwise(self, seqlen, batchsize, inpsize, hiddensize):  # type: (int, int, int, int) -> None
```
spandantiwari, line 982: Maybe I missed it, but is there a shape inference test added for the bidirectional case with the new flag? If not, it would be nice to add one.

onnx/defs/rnn/defs.cc
```diff
@@ -19,22 +19,30 @@ void RNNShapeInference(InferenceContext& ctx) {
   if (hidden_size_value > 0)
     hidden_size.set_dim_value(hidden_size_value);
 
+  auto batch_major_value = getAttribute(ctx, "batch_major", 0);
```
spandantiwari, line 22: nit: If the final name of the `batch_major` or `batch_axis` flag is changed, it may be helpful to update the name of this variable.

onnx/defs/rnn/defs.cc
gramalingam: Re. the more general flag: are there uses-cases (either models or systems) where batch-axis other than 0 or 1 is used? If not, it may be reasonable to leave it in the proposed form (covering the two scenarios that are used in practice, I assume). Allowing arbitrary batch-axis and sequence-axis is conceptually simple, but there may be implementation costs in specific scenarios.

matteosal:
> There seems to be a merge conflict in this PR, which removes the `(differentiable)` annotations from inputs and outputs.
> 
> Also, is there a reason `third_party/pybind11` was bumped?
> 
> Merge conflicts in other files need to be resolved and we should check why CI isn't passing.

matteosal:
> There seems to be a merge conflict in this PR, which removes the `(differentiable)` annotations from inputs and outputs.
> 
> Also, is there a reason `third_party/pybind11` was bumped?
> 
> Merge conflicts in other files need to be resolved and we should check why CI isn't passing.

The `(differentiable)` flags look good to me now. If there's still something wrong, can you point out? I have also fixed the `pybind11` version and the conflicts

onnx/defs/rnn/defs.cc
matteosal: In the last operator SIG meeting, it was decided to adopt the more general flag `batch_axis`, but restricting its values to 0 and 1, leaving extensions for later if needed.

However, putting myself back to work on this, I've realized that `batch_axis = 1|0` is not descriptive of what `batch_major = 0|1` does, because the shapes look like this:
* `batch_major = 0` (old case):
`input_shape = `[seq_length, batch_size, input_size]`
`output_shape = `[seq_length, num_directions, batch_size, hidden_size]`
* `batch_major = 1` (new case):
`input_shape = `[batch_size, seq_length, input_size]`
`output_shape = `[batch_size, seq_length, num_directions, hidden_size]`

Hence `batch_major = 1` does correspond to `batch_axis = 0`, but `batch_major = 0` corresponds to `batch_axis = 1` for the input and `batch_axis = 2` for the output. So maybe it's better to go back to `batch_major = 0|1`? The current spec for the batch axis is not even consistent between input and output, so generalizing it would require two separate flags, which might be too much for something that doesn't even have a current use case...

What should I do?

onnx/test/shape_inference_test.py
matteosal: I've added it

onnx/defs/rnn/defs.cc
wschin: Having two int flags (`input_batch_axis` and `output_batch_axis`) sound ok to me. It doesn't hurt and help maintaining this op's extensibility. What do you think?

postrational: We need to consider if we're not making the op too flexible, which may put unnecessary burden on backend implementors.

matteosal: So what should I do with this? How to take a decision?

matteosal: Can this be moved forward? Is it ok to just go for `batch_major = 0|1`?

gramalingam: My personal opinion: it seems ok to use batch_major = 0|1.

matteosal:
@prasanthpul @askhade sorry for leaving this unfinished for so long. I got back to it trying to finalize it. I'm sticking for a single `batch_major` attribute for all 3 ops, moved all the changes to opset 14 and added all the adapters.

The only thing left to do is to regenerate the test files, but `backend-test-tools generate-data` is failing with a couple of errors:
```
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 792, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.VersionConflict: (onnx 1.8.0 (/usr/local/lib/python3.8/dist-packages/onnx-1.8.0-py3.8-linux-x86_64.egg), Requirement.parse('onnx==1.7.0'))
```
```
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 787, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'onnx==1.7.0' distribution was not found and is required by the application
```

Any help with this one?

matteosal:
Commit signoff here is very impractical due to many many past conflicts that should be fixed again in the rebase. I've reopened this PR here: https://github.com/onnx/onnx/pull/3217

PR #2923, opened 2020-07-24
More recurrent tests (reprise)

matteosal:
This is a continuation of this old PR: https://github.com/onnx/onnx/pull/2279


kraiskil:
Hello all.
What is the status of this PR? What would be the best way to nudge it forward?
Would be nice to have (at least some of) these backend node unit tests in onnx.

matteosal:
> Hello all.
> What is the status of this PR? What would be the best way to nudge it forward?
> Would be nice to have (at least some of) these backend node unit tests in onnx.

I agree that additional rnn tests are needed, but this PR is so old that probably needs to start over again, especially to incorporate new tests to the new `layout = 1` setting (https://github.com/onnx/onnx/pull/3217). Unfortunately I don't have much time now and I intend to dedicate all the bandwidth I have for ONNX to this one, which is far more important: https://github.com/onnx/onnx/pull/3344

So I intend to get back to this but not anytime soon.

kraiskil:
Thanks for the update @matteosal. 
These unit tests are really useful as complements to documentation when implementing a backend, but I can see how there would be more urgent tasks :) Thanks in advance for when you can put time into this.

PR #2925, opened 2020-07-24
Remove Add new type restriction on operator versioning.

ebarsoum:
As discussed in our SIG operator meeting (https://github.com/onnx/sigs/blob/master/operators/meetings/016-20200724.md), adding new type to operator input/output doesn't warrant operator version change.


docs/Versioning.md
```diff
@@ -80,6 +82,8 @@ The following are not breaking:
 * Clarifications of specification ambiguities to match prevailing
   implementation practice.
 
+* Adding new type to inputs or outputs.
```
gramalingam, line 85: How about adding the following? _Earlier such changes were considered breaking changes requiring an increase in the version-number. However, in practice, the cost of increasing the version-number in this situation outweighs the benefits. This is particularly so since backends, in practice, rarely guarantee complete coverage of all types that exist in the standard._

prasanthpul:
@ebarsoum Please edit the the initial comment of this PR to include links to the detailed notes of the discussion and decision

docs/Versioning.md
```diff
@@ -71,7 +71,9 @@ A given operator is identified by a three-tuple: `(domain, op_type, and op_versi
 
 * Adding/removing/reordering inputs or outputs.
 
-* Adding/removing types supported by inputs and outputs, and changing types used by attributes.
+* Removing types supported by inputs or outputs.
```
askhade, line 74: why bump up version for removing types? do we need to treat add and remove different?

ebarsoum: Because removing type can break existing implementation.

ebarsoum:
@postrational, @linkerzhang and @chinhuang007 any feedback?

docs/Versioning.md
```diff
@@ -80,6 +82,8 @@ The following are not breaking:
 * Clarifications of specification ambiguities to match prevailing
   implementation practice.
 
+* Adding new type to inputs or outputs. Earlier such changes were considered breaking changes requiring an increase in the version-number. However, in practice, the cost of increasing the version-number in this situation outweighs the benefits. This is particularly so since backends, in practice, rarely guarantee complete coverage of all types that exist in the standard.
```
chinhuang007, line 85: I agree adding a new type to inputs or outputs does not need to be considered as breaking changes. However we should find a way to communicate this kind of changes with broader community such as developers for converters. Additionally, if data type specific units tests are added, some backends will potentially fail standard backend tests although the opset version stays the same. Some clarification would help.

ebarsoum: @chinhuang007 the way that I think about it, it will be based on ONNX release version. For example adding new types to OpSet 12 wont affect ONNX 1.7 release, but it will be part of 1.8 release.

linkerzhang: "adding new types to OpSet 12 wont affect ONNX 1.7 release"

An model with the added (new) type can't be run against framework/runtime which supports opset12, right?

ebarsoum: @linkerzhang that is already the case, I am not aware of any framework that support all the types.

ebarsoum:
@chinhuang007 any other feedback?

gramalingam:
UPDATE: After further discussions, the decision was to NOT pursue this relaxation. The main motivation here was the introduction of the bfloat16 type which impacted a number of ops. But it did not seem safe enough to justify this relaxation. Closing this PR at this point.

PR #2929, opened 2020-07-28
Fixes for Windows CI

askhade:
Adds logic to catch CI failures and properly report them and adds C++ API tests to the CI.

Note: test_external_data is failing in windows CI when checking a model >2GB. Windows CI swallows this error. This is however covered in Linux and Mac OS CIs. Fix for this test failure will be added as part of a separate PR

### DIFF
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -8,6 +8,10 @@ jobs:
     vmImage: 'vs2017-win2016'
   strategy:
     matrix:
+      py37_onnx_ml:
+        python.version: '3.7'
+        onnx_ml: 1
+        onnx_verify_proto: 0
       py37:
         python.version: '3.7'
         onnx_ml: 0
@@ -16,13 +20,9 @@ jobs:
         python.version: '3.6'
         onnx_ml: 0
         onnx_verify_proto: 0
-      py37_onnx_ml:
-        python.version: '3.7'
-        onnx_ml: 1
-        onnx_verify_proto: 0
       py36_verify_proto:
         python.version: '3.6'
-        onnx_ml: 0
+        onnx_ml: 1
         onnx_verify_proto: 1
     maxParallel: 4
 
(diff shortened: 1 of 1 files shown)


.azure-pipelines/Windows-CI.yml
```diff
@@ -49,16 +49,39 @@ jobs:
...
+        # TODO fix test-external_data test for models > 2GB
+        # There is a known issue in test_external_data in windows CI.
+        # Swallowing the failures temporarily. Fix will be added in a separate PR
+        # Linux and Mac CI are configured to throw errors for pytest failures.
+        @echo "pytest failed"
+      )
+
+      # onnx c++ API tests
+      .setuptools-cmake-build/Release/onnx_gtests.exe
+      IF NOT %ERRORLEVEL% EQU 0 (
```
jcwchen, line 66: Just for curios: when do we need to catch the error from last command?
I don't understand why AzurePipeline sometimes catch the failure normally but sometimes does not.

jcwchen:
I wrote that test... If you have not started yet, I can propose a PR and correct that error. Thank you!

jcwchen:
BTW, @gramalingam and I saw there are some errors if we run `python setup.py typecheck` on our local Windows. We are not sure that whether these errors are caused by our local machine or onnx itself. 
Since this PR will catch typecheck errors properly in Windows-CI, we need to be aware of whether it happens in the CI machine.

.azure-pipelines/Windows-CI.yml
```diff
@@ -49,16 +49,39 @@ jobs:
...
+        @echo "onnx_gtests failed"
+        EXIT 1
+      )
+
+      # check auto-gen files up-to-date
       python onnx/defs/gen_doc.py
       python onnx/gen_proto.py -l
       python onnx/gen_proto.py -l --ml
+      python onnx/backend/test/stat_coverage.py
```
jcwchen, line 75: Do we need to add the error catch here too?

.azure-pipelines/Windows-CI.yml
```diff
@@ -49,16 +49,39 @@ jobs:
...
-
       python setup.py --quiet install
+
+      # onnx python api tests
       pytest
+      IF NOT %ERRORLEVEL% EQU 0 (
+        # TODO fix test-external_data test for models > 2GB
+        # There is a known issue in test_external_data in windows CI.
+        # Swallowing the failures temporarily. Fix will be added in a separate PR
+        # Linux and Mac CI are configured to throw errors for pytest failures.
```
jcwchen, line 60: Please note that the test has been handled (skipped) so the error catch here is ready to go. Thank you!

PR #2935, opened 2020-07-30
Catch unexpected MemoryError for large model test on Windows-CI

jcwchen:
**Description**
I checked the Azure Pipeline doc it seems that Microsoft-hosted agents cannot increase the memory. If we cannot enhance our CI Windows machine, the workaround might be to catch `MemoryError` as well. In that case, the test is useless on Windows, but it has been covered on Linux and Mac. (This test worked normally on them) 

**Motivation and Context**
https://github.com/onnx/onnx/pull/2929
`TestLarge2GBExternalData` tests 2 GB models with `onnx.checker`. However, some machines will encounter `MemoryError` because of the hardware limitation. (e.g., Azure Pipeline Windows Machine)

### DIFF
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -53,7 +53,7 @@ jobs:
       set CMAKE_ARGS=-DONNX_USE_PROTOBUF_SHARED_LIBS=ON -DProtobuf_USE_STATIC_LIBS=OFF -DONNX_USE_LITE_PROTO=ON
 
       python setup.py --quiet install
-      pytest
+      pytest -s
       IF NOT %ERRORLEVEL% EQU 0 (
         @echo "pytest failed"
         EXIT 1
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -53,7 +53,7 @@ jobs:
       set CMAKE_ARGS=-DONNX_USE_PROTOBUF_SHARED_LIBS=ON -DProtobuf_USE_STATIC_LIBS=OFF -DONNX_USE_LITE_PROTO=ON
 
       python setup.py --quiet install
-      pytest -s
+      pytest
       IF NOT %ERRORLEVEL% EQU 0 (
         @echo "pytest failed"
         EXIT 1
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -56,6 +56,7 @@ jobs:
       pytest
       IF NOT %ERRORLEVEL% EQU 0 (
         @echo "pytest failed"
+        EXIT 1
       )
 
       python onnx/defs/gen_doc.py
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -54,6 +54,10 @@ jobs:
 
       python setup.py --quiet install
       pytest
+      IF NOT %ERRORLEVEL% EQU 0 (
+        @echo "pytest failed"
+      )
+
       python onnx/defs/gen_doc.py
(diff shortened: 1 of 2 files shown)


onnx/test/test_external_data.py
```diff
@@ -285,7 +285,7 @@ def create_test_model_both(self):  # type: () -> Tuple[Text, ModelProto]
     # 2GB models would fail onnx.checker with ModelProto but pass with model path
     def test_check_model_by_model(self):  # type: () -> None
         model = onnx.load_model(self.model_filename, load_external_data=False)
-        with pytest.raises(ValueError):
+        with pytest.raises((ValueError, MemoryError)):
```
askhade, line 288: For memory error add a log statement stating --> because of mem limitation this test was not executed.

jcwchen: Good idea. Just added.

.azure-pipelines/Windows-CI.yml
```diff
@@ -54,6 +54,10 @@ jobs:
 
       python setup.py --quiet install
       pytest
+      IF NOT %ERRORLEVEL% EQU 0 (
+        @echo "pytest failed"
+      )
```
askhade, line 59: need to add EXIT 1 otherwise it will continue and mark pipeline as green

onnx/test/test_external_data.py
```diff
@@ -285,9 +285,12 @@ def create_test_model_both(self):  # type: () -> Tuple[Text, ModelProto]
     # 2GB models would fail onnx.checker with ModelProto but pass with model path
     def test_check_model_by_model(self):  # type: () -> None
         model = onnx.load_model(self.model_filename, load_external_data=False)
-        with pytest.raises(ValueError):
+        try:
             load_external_data_for_model(model, self.temp_dir)  # Exceeds maximum protobuf
-            checker.check_model(model)  # checker catches 2GB models as well
+        except MemoryError:
+            print("Warning: Because of the hardware limitation (memory), this test was not executed.")
+            return
```
askhade, line 292: take a look at windows CI - it is printing this warning but still marks the pytest as failed

jcwchen: Thank you for catching this. Just modified.

askhade: In the latest windows CI run there is no warning message saying this test was skipped. Can you fix the CI.
It is OK to skip this test on Windows because of hardware limitation but we need to explicitly call it out.

jcwchen: By default `pytest` prints stdout only if the test fails. Even though adding `-s` to `pytest` can capture the stdout from tests, it will apply to all tests. Is it acceptable?

askhade: instead of catching mem error can you simply skip this test on windows something like this : https://docs.pytest.org/en/stable/skipping.html#skipping-test-functions and see if pytest can add skip tests in summary

jcwchen: @askhade do you mean something like this?
```
@pytest.mark.skipif(os.name == 'nt', reason="Because of the hardware limitation (memory) in Windows-CI, this test was not executed.")
```
Done, but the linux failed. It seems like the old problem is happening again. https://github.com/onnx/onnx/pull/2933

askhade: did you try this :  @pytest.mark.skipif(sys.platform.startswith("win"), reason="Because of the hardware limitation (memory) in Windows-CI, this test was not executed.")

askhade: Linux failure is because of a type check error : 
onnx/test/test_external_data.py:288: error: Untyped decorator makes function "test_check_model_by_model" untyped

jcwchen: Thanks for the catch. Just fixed. That test has been successfully skipped on Windows.

Issue #2932, opened 2020-07-29
The checker fails for missing shape, but IR allows it


matteosal:
[IR.md](https://github.com/onnx/onnx/blob/master/docs/IR.md) states that the "shape" property can be missing from input and output specs:

> When the shape property is absent for an input, a tensor value of any shape may be passed from the caller. When the shape property is absent for an output, the caller should expect that the output value may be of any shape.

But running the checker on the attached file fails saying that "shape" is required:
```
>>> model = onnx.load("file.onnx")
>>> onnx.checker.check_model(model)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/dist-packages/onnx-1.7.0-py3.6-linux-x86_64.egg/onnx/checker.py", line 102, in check_model
    C.check_model(protobuf_string)
onnx.onnx_cpp2py_export.checker.ValidationError: Field 'shape' of type is required but missing.
```

Am I missing something here? Is the model invalid for some other reason?

[file.zip](https://github.com/onnx/onnx/files/4996240/file.zip)

jcwchen:
Hi @matteosal,

Similar to https://github.com/onnx/onnx/issues/2492?
I agree this behavior of checker does not meet the description in IR...
@linkerzhang what do you think about this difference between the new IR and the current checker?

skottmckay:
I believe graph inputs require a shape, node inputs do not. 

The shape may have symbolic (dim_param is set) or unknown (neither dim_param or dim_value is set) dimensions in it.

linkerzhang:
yes, Main graph's inputs and outputs must have type and shape information. Do you @skottmckay or @jcwchen want to create a PR to update the IR.md please?

linkerzhang:
It's not a BUG in checker but an issue of the documentation, btw. @matteosal

jcwchen:
> yes, Main graph's inputs and outputs must have type and shape information. Do you @skottmckay or @jcwchen want to create a PR to update the IR.md please?

Happy to help. Thank you for the explanation @skottmckay  @linkerzhang

matteosal:
> I believe graph inputs and outputs require a shape as well. Node inputs do not.

Well, for that matter, node inputs are not even required to exist at all. Then I think that this section I mentioned should also be made clearer:

> When the shape property is absent for an input, a tensor value of any shape may be passed from the caller. When the shape property is absent for an output, the caller should expect that the output value may be of any shape.

[...]

Issue #2492, opened 2019-12-06
"shape" field should not be enforced on TypeProto::kTensorType in onnx checker


samratdebroy:
When running the onnx checker on a model with a tensor that has a missing "shape" parameter, the checker throws the following error:

```
C.check_model(model.SerializeToString())
onnx.onnx_cpp2py_export.checker.ValidationError: Field 'shape' of type is required but missing.
```

looking at [checker.cc](https://github.com/onnx/onnx/blob/master/onnx/checker.cc) I see that this is a result of the following code:
```
 case TypeProto::kTensorType: {
      const auto& type = value_info.type().tensor_type();
      enforce_has_field(type, elem_type);
      enforce_has_field(type, shape);
    } break;
``` 
but the [ONNX standard](https://github.com/onnx/onnx/blob/master/docs/IR.md#input--output-data-types) clearly states that a missing "shape" parameter on a tensor is how we indicate that the tensor takes in zero or more dimensions of data:

>  A zero-dimension tensor is distinct from a tensor of unknown dimensionality, which is indicated by an absent 'shape' property in the Tensor record.
as well as:

> The use of the string "*" (as a dimension variable) to denote a sequence of zero or more dimensions of unknown cardinality. This is not supported. In the current implementation, the number of dimensions in a shape MUST represent the rank of the tensor. A tensor of unknown rank is represented using a TypeProto::Tensor object with no shape, which is legal.

I encountered this problem while trying to visualize the output of one of the intermediary nodes of the model by marking it as an output of the model itself:
```
model = onnx.load(model_path)
intermediate_layer_value_info = onnx.helper.make_tensor_value_info('output_name', onnx.TensorProto.FLOAT, None)
model.graph.output.extend([intermediate_layer_value_info])
onnx.checker.check_model(model)
```

linkerzhang:
The checker is good in this case.

The standard is saying, every graph's inputs/outputs MUST have detail type information specified in a model. Type information contains "shape".

For your case, as you're actually trying to add an intermediate node's output, whose type information is actually not a MUST (as it can be inferred), as a graph output, so the failure is shown.

To resolve the problem, you may infer the graph/model to understand the intermediate node's output's type information and make the type information complete when adding it into graph output list.

samratdebroy:
[...]

PR #2937, opened 2020-07-31
Declaration about shape info for graph I/O and node I/O in IR.md

jcwchen:
**Description**
To resolve the confusion, add more declaration about shape for graph I/O (input and output) and node I/O in IR.md.

**Motivation and Context**
https://github.com/onnx/onnx/issues/2492, https://github.com/onnx/onnx/issues/2932
According to the original IR.md, users are confused about whether the shape of I/O must exist.

cc @linkerzhang @skottmckay @matteosal


docs/IR.md
```diff
@@ -370,7 +370,9 @@ Which is referenced by the Tensor type message:
   }
 ```
 
-The empty list of dimension sizes, [], is a valid tensor shape, denoting a zero-dimension (scalar) value. A zero-dimension tensor is distinct from a tensor of unknown dimensionality, which is indicated by an absent 'shape' property in the Tensor record. When the shape property is absent for an input, a tensor value of any shape may be passed from the caller. When the shape property is absent for an output, the caller should expect that the output value may be of any shape.
+The empty list of dimension sizes, [], is a valid tensor shape, denoting a zero-dimension (scalar) value. A zero-dimension tensor is distinct from a tensor of unknown dimensionality, which is indicated by an absent 'shape' property in the Tensor record. When the shape property is absent for a node input, a tensor value of any shape may be passed from the caller. When the shape property is absent for a node output, the caller should expect that the output value may be of any shape.
```
jcwchen, line 373: Specify:
input -> node input
output -> node output

gramalingam: I suggest changing the "When the shape property is absent ..." part as below:
```
When the shape property is absent in the type of a value, it indicates that the corresponding runtime value
may have any shape. However, types with a missing shape are not permitted in all contexts. In particular,
the inputs and outputs of a model are required to *have* a shape, indicating the rank of inputs and outputs,
even though the exact dimensions need not be specified.
```

docs/IR.md
```diff
@@ -370,7 +370,9 @@ Which is referenced by the Tensor type message:
   }
 ```
 
-The empty list of dimension sizes, [], is a valid tensor shape, denoting a zero-dimension (scalar) value. A zero-dimension tensor is distinct from a tensor of unknown dimensionality, which is indicated by an absent 'shape' property in the Tensor record. When the shape property is absent for an input, a tensor value of any shape may be passed from the caller. When the shape property is absent for an output, the caller should expect that the output value may be of any shape.
+The empty list of dimension sizes, [], is a valid tensor shape, denoting a zero-dimension (scalar) value. A zero-dimension tensor is distinct from a tensor of unknown dimensionality, which is indicated by an absent 'shape' property in the Tensor record. When the shape property is absent for a node input, a tensor value of any shape may be passed from the caller. When the shape property is absent for a node output, the caller should expect that the output value may be of any shape. 
+
+Note that node input and output can be completely absent from the graph (and can be filled by shape inference), but graph input and output must always specify a shape.
```
gramalingam, line 375: I suggested a reworded version which merges this statement into the previous paragraph. If you want the first part of the above, I suggest using "Note that type and shape information for intermediate values in the graph, used as inputs and outputs of nodes, are optional but not compulsory as they can be inferred, but type and rank information is compulsory for inputs and outputs of the model.

gramalingam:
Another clarification that may be useful is that the treatment is slightly different for "model inputs/outputs" (which are inputs/outputs of top-level graphs) and "nested graph inputs/outputs" (which are inputs/outputs of graphs that specify the loop-body or body of an if-then-else).

docs/IR.md
```diff
@@ -190,7 +190,7 @@ Each main (top-level) graph MUST define the names and types of its inputs and ou
 Name|Type|Description
 |---|---|---|
 name|string|The name of the value/parameter.
-type|Type|The type of the value.
+type|Type|The type of the value **including shape information**.
```
skottmckay, line 193: This table seems to be the general description of ValueInfo, which currently is directly after this statement
> Each main (top-level) graph MUST define the names and types of its inputs and outputs, which are specified as ‘value info’ structures, having the following properties:

ValueInfo is used for multiple things and not just graph inputs/outputs so it may be better to put a general description of ValueInfo first, and in that say 'Type' may include shape information.

Following that we could have the 'MUST' statement to describe the additional requirements on the main graph inputs and outputs. I'd maybe move Rama's suggestion ("In particular,
the inputs and outputs of a model are required to *have* a shape, indicating the rank of inputs and outputs,
even though the exact dimensions need not be specified.") into that. 

Something like
> Each main (top-level) graph MUST define the names, types and shapes of its inputs and outputs, which are specified as ‘value info’ structures. The main graph inputs and outputs are required to *have* a shape, indicating the rank, even though the exact dimensions need not be specified.

It may be helpful to be more specific about what 'not specified' could equate do. Unfortunately we don't talk about dim_value vs dim_param vs unknown until much later in the document under 'Static tensor shapes'.

jcwchen:
Updated. @gramalingam @skottmckay Thank you for the revision!

docs/IR.md
skottmckay: Being a little pedantic here as it's a spec. Can we group things together slightly better?

Specifically line 188 has the 'MUST' for a graph, then we have the ValueInfo structure (although nothing explicitly says it's the ValueInfo structure), then we have some details about nested subgraphs and other graph requirements. 

Would something like the following be slightly better organized?

'Graphs have the following properties:'
\<table with GraphProto fields>

'ValueInfo has the following properties:'
\<table with ValueInfo fields>

\<Various other requirements>
Each main (top-level) graph MUST define...
Nested subgraphs (specified as attribute values) MUST define...
Each graph MUST specify ...
etc. etc.

jcwchen: Good idea. This flow looks clearer to me. Just moved the table.

docs/IR.md
```diff
@@ -370,7 +372,10 @@ Which is referenced by the Tensor type message:
...
+The empty list of dimension sizes, [], is a valid tensor shape, denoting a zero-dimension (scalar) value. A zero-dimension tensor is distinct from a tensor of unknown dimensionality, which is indicated by an absent 'shape' property in the Tensor record. When the shape property is absent for a node input, a tensor value of any shape may be passed from the caller. When the shape property is absent in the type of a value, it indicates that the corresponding runtime value
+may have any shape. However, types with a missing shape are not permitted in all contexts. In particular,
+the inputs and outputs of a model (top-level graphs) are required to *have* a shape, indicating the rank of inputs and outputs,
+even though the exact dimensions need not be specified.
```
skottmckay, line 378: We probably don't need the info on top level graph inputs/outputs requiring a shape here as it's specified earlier. 

Are there any other known contexts where a type with a missing shape is permitted? If not should we drop the the 'However, types with a missing shape are not permitted in all contexts.' as well? @gramalingam are you aware of other scenarios where a shape is required?

askhade: these 2 sentences seem redundant:
When the shape property is absent for a node input, a tensor value of any shape may be passed from the caller. When the shape property is absent in the type of a value, it indicates that the corresponding runtime value
may have any shape.

Can these 2 be combined : 
When the shape property is absent in the type of a value (including node input), it indicates that the corresponding runtime value may have any shape.

docs/IR.md
```diff
@@ -185,14 +185,16 @@ input|ValueInfo[]|The input “parameters” of the graph, possibly initialized
...
 
 Name|Type|Description
 |---|---|---|
 name|string|The name of the value/parameter.
-type|Type|The type of the value.
+type|Type|The type of the value **including shape information**.
 doc_string|string|A human-readable documentation for this value. Markdown is allowed.
 
+Each main (top-level) graph MUST define the names, types and shapes of its inputs and outputs, which are specified as ‘value info’ structures. The main graph inputs and outputs are required to have a shape, indicating the rank, even though the exact dimensions need not be specified.
+
```
askhade, line 197: @skottmckay , @gramalingam : Regarding the type info - it is clear from this statement that the graph inputs and outputs must have types but what about the node inputs? Are they required to have type info? 
This will determine how we handle the case of absent type info more specifically when contxt->getInputType(<input index>) is nullptr. If all inputs (graph and intermediate node inputs) require type info then the node level type inference can simply fail when context->getInputType(<input index>) is a nullptr otherwise the node level type inference should stop and return when type info is not available.

gramalingam: For node inputs (that are not graph inputs), the type information is optional. The model does not *need* to specify it. But in a "correct model", type-inference will be able to infer all these types automatically. (That is to answer your question. In terms of the document: I believe this is mentioned in the document, at least it was at some point in this PR; I haven't checked the latest version.)

docs/IR.md
gramalingam: I sort of agree with the idea underlying Scott's point above. To expand on this: this section is meant to explain how to interpret a shape (or a missing-shape) in a type. This is an explanation of the "class TypeProto". It is not meant to be a statement/descriptions of (further) constraints that may be imposed on the use of a TypeProto in particular contexts ... those belong to a description of those contexts.

On the other hand, if you look at the two issues Jacky mentions as motivation, I feel some redundancy may be okay and it may be okay to repeat some information here to avoid some misunderstanding/confusion. 

What if we rephrase part of the above as below
```
This sub-section describes how to interpret a missing-shape or a shape with missing dimensions etc.
However, specific usage contexts may impose further constraints on a type and shape.
For example, the inputs and outputs of a model (top-level graph) are required to *have* a shape, ... ,
as described in section ...
```

docs/IR.md
jcwchen: I think it is just mentioned in the next sentence?
> Nested subgraphs (specified as attribute values) MUST define the names of its inputs and outputs and MAY define the types of its inputs and outputs.

Do we need to further indicate the input/output of Nested subgraphs as the node input/output here?

Another section which is talking about this:
>  When the shape property is absent for a node input, a tensor value of any shape may be passed from the caller. When the shape property is absent in the type of a value, it indicates that the corresponding runtime value may have any shape. However, types with a missing shape are not permitted in all contexts. In particular, the inputs and outputs of a model (top-level graphs) are required to have a shape, indicating the rank of inputs and outputs, even though the exact dimensions need not be specified.

docs/IR.md
jcwchen: Just modified. @gramalingam can you check again whether I did the right way?
Thanks!

gramalingam: Thanks @jcwchen. Can we add Ashwini's suggestion to the end (those two lines seem to have gotten deleted, without a replacement): namely,
```
When the shape property is absent in the type of a value (including node input),
it indicates that the corresponding runtime value may have any shape.
```

jcwchen: Just added. Thanks for the reminder.

PR #2941, opened 2020-08-03
[CircleCI] Use stable libturbojpeg for torch/vision

jcwchen:
**Description**
To prevent the link (download dependency for torch/vision) from being invalid again, it should use a stable version.

**Motivation and Context**
Related to https://github.com/onnx/onnx/pull/2909, the latest version of libturbojpeg becomes invalid just now.


jcwchen:
:shipit:

PR #2946, opened 2020-08-05
[ONNX] Resolve segfault on Input without tensor data in ConstantofShape

shubhambhokare1:
Added check to avoid segfault on empty input tensor in ConstantofShape op


onnx/defs/generator/defs.cc
```diff
@@ -231,7 +231,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             // from the input 'shape' tensor, then we add the same number
             // of dimensions (without any dim_value information) to the
             // output.
-            if (ctx.getInputType(0)->tensor_type().has_shape()) {
+            if (ctx.getInputType(0) != nullptr && ctx.getInputType(0)->tensor_type().has_shape()) {
```
gramalingam, line 234: (a) I think we can just use
```cpp
   if (hasInputShape(ctx, 0))
```
which will be more consistent with the style elsewhere.

(b) Is there some case where this is failing? I am a bit concerned that the real issue might be somewhere else. This type is expected to exist (have a non-null value) at this point. If the model does not have an input here, I would have expected the earlier checker to have caught that the node is missing an input. I suppose it might be possible to have some cascading error, if the type was not inferred due to some earlier error.

shubhambhokare1: I have changed the code such that it is similar to one the you mentioned in (a)
For (b), The case in which this is failing is if the input to the ConstantofShape node in the onnx model looks like:
  input {
      name: "9"
  }
In this particular case, the previous check tries to access type or even tensor_type which don't have any values assigned to them

PR #2948, opened 2020-08-06
Doc update for Resize to correct some wrong information

HectorSVC:
The scales is set as Single input, it's not possible to set an empty string name for it because we have schema check for Single input(https://github.com/onnx/onnx/blob/72b701f7a55cafa4b8ab66a21dc22da0905b2f4c/onnx/defs/schema.cc#L227-L234). The doc is not correct. The only way is to set a nonempty string as the name but set it to empty data (empty shape).
Another option is to set scales as Optional input, but it requires opset version change.

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -15277,7 +15277,7 @@ x_original = length_resized > 1 ? start_x * (length_original - 1) + x_resized * 
 <dt><tt>roi</tt> : T2</dt>
 <dd>1-D tensor given as [start1, ..., startN, end1, ..., endN], where N is the rank of X. The RoIs' coordinates are normalized in the coordinate system of the input image. It only takes effect when coordinate_transformation_mode is "tf_crop_and_resize"</dd>
 <dt><tt>scales</tt> : tensor(float)</dt>
-<dd>The scale array along each dimension. It takes value greater than 0. If it's less than 1, it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should be the same as the rank of input 'X'. Only one of 'scales' and 'sizes' can be specified. If 'size' is needed, the user can use an empty string as the name of 'scales' in this operator's input list.</dd>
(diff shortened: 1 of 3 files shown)


askhade:
both changelog.md and operator.md needs to be updated. Please use python onnx/defs/gen_doc.py script to generate these files. They should not be manually updated.

docs/Operators.md
```diff
@@ -15277,7 +15277,7 @@ x_original = length_resized > 1 ? start_x * (length_original - 1) + x_resized *
 <dt><tt>roi</tt> : T2</dt>
 <dd>1-D tensor given as [start1, ..., startN, end1, ..., endN], where N is the rank of X. The RoIs' coordinates are normalized in the coordinate system of the input image. It only takes effect when coordinate_transformation_mode is "tf_crop_and_resize"</dd>
 <dt><tt>scales</tt> : tensor(float)</dt>
-<dd>The scale array along each dimension. It takes value greater than 0. If it's less than 1, it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should be the same as the rank of input 'X'. Only one of 'scales' and 'sizes' can be specified. If 'size' is needed, the user can use an empty string as the name of 'scales' in this operator's input list.</dd>
+<dd>The scale array along each dimension. It takes value greater than 0. If it's less than 1, it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should be the same as the rank of input 'X'. Only one of 'scales' and 'sizes' can be specified. If 'size' is needed, the user can use a nonempty string as the name of 'scales' but set it to empty data (zero shape) in this operator's input list.</dd>
```
askhade, line 15280: Can this -> "If 'size' is needed, the user can use a nonempty string as the name of 'scales' but set it to empty data (zero shape) in this operator's input list."
be changed to ->
"If 'size' is specified, then set scales to empty data (zero shape) in this operator's input list."

Basically scales is not an optional input, adding empty data will imply that input name will be non-empty.

daquexian:
@HectorSVC I'm the main author of the resize op. I'm planning to submit a PR to update the resize op version, as I found the "tf_half_pixel_for_nn" coordinate_transformation_mode is indeed unnecessary and misleading. I think it is better to also make scales as an optional input in my PR. What's your opinion? Thanks!

HectorSVC:
@daquexian Make scales as an optional input is definitely the best solution. I didn't do it because I don't want to bump up the opset version. Since you need to bump up the version any way for your coming PR, I think you can do the changes for scales together. Thanks!

Issue #2647, opened 2020-03-10
ONNX InferenceContext implementation triggers segfault failures


linkerzhang:
https://github.com/onnx/onnx/blob/master/onnx/shape_inference/implementation.h#L91

ONNX type and shape InferenceContext implementation sets all types as nullptr by default if not manually set, and, current shape inference functions implemented do not check "nullptr" accordingly, with these two, using ONNX package to run inference (type and shape) will sometimes have segfault failures (nullptr exception).

This needs to be fixed by either updating the implementation to make it never have nullptr, or fix all shape inference functions to check that, for example, https://github.com/onnx/onnx/blob/master/onnx/defs/generator/defs.cc#L234 needs checking the input type to ensure it's not nullptr.

faxu:
@linkerzhang are you working on or planning a PR for this?

shubhambhokare1:
https://github.com/onnx/onnx/pull/2946
gets rid of a possible segfault issue in 
https://github.com/onnx/onnx/blob/master/onnx/defs/generator/defs.cc#L234

PR #2950, opened 2020-08-07
Add nullptr check to type inference mtds to avoid segfaults

askhade:
Addresses #2647 

ONNX type and shape InferenceContext implementation sets all types as nullptr by default (https://github.com/onnx/onnx/blob/master/onnx/shape_inference/implementation.h#L91). All node level type and shape inference functions do not check for nullptr before dereferencing the pointer and this results in segfaults. 

This PR fixes type and shape inference functions which do not have this check to validate TypeProto* is not a nullptr before proceeding.


onnx/defs/tensor/defs.cc
```diff
@@ -2896,14 +2888,17 @@ ONNX_OPERATOR_SET_SCHEMA(
           } else {
             // 'axis' is provided.
             int axis = static_cast<int>(axisAttr->i());
-            const TensorShapeProto& input_shape =
+            if(!xTensorProto->tensor_type().has_shape()) {
```
BowenBao, line 2891: should this part be updated to check with `(nullptr == xTensorProto) || !xTensorProto->tensor_type().has_shape()`?

askhade: it's not required in this case. This will be checked as part of the type inference.
see line 2849 ->propagateElemTypeFromInputToOutput(ctx, 0, 0);

BowenBao:
LG, please sync with master branch.

onnx/defs/sequence/defs.cc
```diff
@@ -88,6 +88,10 @@ ONNX_OPERATOR_SET_SCHEMA(
 
           std::vector<int> input_elem_types;
           for (size_t i = 0; i < numInputs; ++i) {
+            auto input_type = ctx.getInputType(i);
+            if(nullptr == input_type){
+                fail_type_inference("Input type for input at index ", i, " is null. Type info is expected.");
+            }
             input_elem_types.emplace_back(ctx.getInputType(i)->tensor_type().elem_type());
```
gramalingam, line 95: Replace "ctx.getInputType(i)" by input_type?

PR #2954, opened 2020-08-11
Add Some Differentiable Tags for The Rest Ops in tensor/defs.cc

wschin:
This PR tags `Resize` and `Compress`. This PR also changes the differentiable tags for `Round` and `Sign`. I think it's better to mark them non-differentiable until we agree a way to compute gradient at their discontinuous points. 

`Resize`:
We use Pytorch `Upsample` to understand the differentiability of ONNX `Resize`. As shown in my Python code below, there is only one differentiable input and one differentiable output. The rest parameters such as `scale` are constants and not differentiable.
```python
import torch
import torch.nn as nn

# A single-operator model. It's literally a Pytorch Reshape.
# Note that Pytorch Reshape can be directly mapped to ONNX Reshape.
class MyModel(nn.Module):
  def __init__(self):
    super(MyModel, self).__init__()
    self.layer = nn.Upsample(scale_factor=2, mode='bilinear')

  def forward(self, x):
    y = self.layer(x)
    y.retain_grad()
    return y

model = MyModel()

x = torch.arange(1, 5, dtype=torch.float32).view(1, 1, 2, 2)
x.requires_grad = True
y = model(x)
dy = torch.tensor(torch.ones_like(y))

torch.autograd.backward([y],
  grad_tensors=[dy],
  retain_graph=True,
  create_graph=True,
  grad_variables=None)

# This example shows the input and the output in Pytorch are differentiable.
# From the exported ONNX model below, we also see that "x" is the first input
# of ONNX Reshape and "y" the output of ONNX Reshape. Therefore, we can say
# the first input and the output of ONNX Reshape are differentiable.
print(x.grad)
print(y.grad)

with open('model9.onnx', 'wb') as f:
  torch.onnx.export(model, x, f, opset_version=9)
with open('model11.onnx', 'wb') as f:
  torch.onnx.export(model, x, f, opset_version=11)
```

`Compress`:
`Compress` is similar to `Gather`. It extracts some slices of tensors from input. Its backward operator should just reverse the forward mapping.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -1564,7 +1564,10 @@ ONNX_OPERATOR_SET_SCHEMA(
             "data",
             "Tensors with at least max(dims) dimensions.",
             "T",
-            OpSchema::Single)
+            OpSchema::Single,
+            true,
+            1,
+            OpSchema::Differentiable)
         .Input(
             1,
             "axes",
@@ -1580,7 +1583,10 @@ ONNX_OPERATOR_SET_SCHEMA(
             "squeezed",
             "Reshaped tensor with same data as input.",
             "T",
-            OpSchema::Single)
+            OpSchema::Single,
+            true,
+            1,
+            OpSchema::Differentiable)
         .TypeConstraint(
             "T",
             OpSchema::all_tensor_types_with_bfloat(),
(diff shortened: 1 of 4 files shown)


askhade (approved):

onnx/defs/tensor/defs.cc
```diff
@@ -2028,20 +2034,29 @@ ONNX_OPERATOR_SET_SCHEMA(
             "X",
             "N-D tensor",
             "T",
-            OpSchema::Single)
```
wschin, line 2031: Revert changes for 'Upsample`. It is already deprecated.

askhade (changes requested):
revert changes for Upsample as it is a deprecated op

wschin:
> revert changes for Upsample as it is a deprecated op

Done.

askhade:
Since the ops which are being updated in this PR has changed please update the PR description to reflect the same

wschin:
> Since the ops which are being updated in this PR has changed please update the PR description to reflect the same

Sure. Done.

onnx/defs/tensor/defs.cc
```diff
@@ -1564,7 +1564,10 @@ ONNX_OPERATOR_SET_SCHEMA(
             "data",
             "Tensors with at least max(dims) dimensions.",
             "T",
-            OpSchema::Single)
+            OpSchema::Single,
```
askhade, line 1567: I dont see the Squeeze changes reflected in the Operators and Changelog files

wschin: Please see line 20508 in Operators.md.

askhade (approved):

sveta-levitan (approved):
Looks great.

gramalingam (approved):

PR #2955, opened 2020-08-12
Schema change to support dynamic shapes in ORT

ashbhandare:
This PR changes the schema of the following ops: 
ReduceSum -> 'axes' attribute to input
Squeeze-> 'axes' attribute to input
Unsqueeze-> 'axes' attribute to input
Split -> 'split' attribute to input

- It also modifies the function body of ops SoftmaxCrossEntropyLoss and NegativeLogLikelihoodLoss where they use the above ops with a modified schema.

- The changes also include adding adapters for opset version conversion of these ops from 12->13 and 13->12.

- This includes required changes to existing and new tests.


askhade:
1. You are converting attributes for operators to inputs right? Please update the PR title to reflect the same. Current title is ambiguous.
2. Add a description to the PR which includes all the ops which you are changing

onnx/defs/reduction/defs.cc
```diff
@@ -59,19 +80,38 @@ False instead of True.)DOC";
...
         keep_dims = attr_proto->i();
       }
+      auto noop_attr_proto = ctx.getAttribute("noop_with_empty_axes");
+      if (noop_attr_proto) {
+        noop_with_empty_axes = noop_attr_proto->i();
+      }
+      std::vector<int64_t> axes;
+      size_t num_inputs = ctx.getNumInputs();
+      if (num_inputs == 2) { // axes is input
+        const TensorProto* axesInitializer = ctx.getInputData(1);
```
gramalingam, line 95: Check if the axes attribute is specified and produce an error-message in that case?

onnx/defs/reduction/defs.cc
```diff
@@ -59,19 +80,38 @@ False instead of True.)DOC";
...
       if (attr_proto) {
         keep_dims = attr_proto->i();
       }
+      auto noop_attr_proto = ctx.getAttribute("noop_with_empty_axes");
+      if (noop_attr_proto) {
+        noop_with_empty_axes = noop_attr_proto->i();
+      }
+      std::vector<int64_t> axes;
+      size_t num_inputs = ctx.getNumInputs();
+      if (num_inputs == 2) { // axes is input
```
gramalingam, line 94: I wonder if it would be safer to use the condition
```cpp
if ( (num_inputs == 2) && ctx.getInputType(1))
```
to handle the case where the second input is specified using an empty string.

onnx/defs/tensor/defs.cc
```diff
@@ -524,7 +533,14 @@ ONNX_OPERATOR_SET_SCHEMA(
           int split_dim_value = static_cast<int>(split_dim.dim_value());
 
           std::vector<int64_t> split;
-          if (getRepeatedAttribute(ctx, "split", split)) {
+          size_t num_inputs = ctx.getNumInputs();
+          if (num_inputs == 2) { //'split' is input
```
gramalingam, line 537: Same comment as before, about adding the condition "ctx.getInputType(1)".

onnx/defs/reduction/defs.cc
gramalingam: (In the general case, a node may have a missing optional input followed by another input that is present; so, it is insufficient to check the number of inputs to figure out if an optional input is present or not. For the special case of the last input, as in this case, the above may work, but we would need to test it in the case where the second input is specified using an empty string. So, I think it is safer to add the above extra condition.)

ashbhandare: changed

gramalingam:
It would be helpful if someone familiar with the optimizer and convertor could also review the PR, as it has updates to those too.

gramalingam:
Can you fix the CI errors? Thanks!

askhade:
Looks good to me. Please address the CI failures and make sure to regenerate the test data before checking in.

onnx/defs/tensor/defs.cc
```diff
@@ -465,6 +465,16 @@ ONNX_OPERATOR_SET_SCHEMA(
             true,
             1,
             OpSchema::Differentiable)
+        .Input(
```
linkerzhang, line 468: Version bump is needed.

askhade: @linkerzhang : The version for this op is already bumped to 13 as part of PR https://github.com/onnx/onnx/pull/2770

daquexian:
Should we also change other ops (ReduceMax/Min/...) in the ReduceX family (they all have an 'axes' attribute) so that they can be consistent with ReduceSum?

Issue #2658, opened 2020-03-13
Error when building onnx in Ubuntu


DuZzzs:
environment：
```
Ubuntu 16.04.5 LTS
python 3.7
GPU：GTX 2060
```
command:
```
conda install -c conda-forge protobuf numpy
conda install pybind11 -c conda-forge

git clone https://github.com/onnx/onnx.git
cd onnx
git submodule update --init --recursive
python setup.py install
```
commits : c2fefcbfa052dd3b602ec825456184d275e91876
Error:
```
[  8%] Built target onnxifi_loader
[  8%] Built target onnxifi_dummy
[  8%] Running C++ protocol buffer compiler on /home/dzws/tensorrt/mxnet/3rdparty/onnx-tensorrt/third_party/onnx/.setuptools-cmake-build/onnx/onnx-ml.proto
/usr/bin/env: "python\r": No such file or directory
--mypy_out: protoc-gen-mypy: Plugin failed with status code 127.
CMakeFiles/gen_onnx_proto.dir/build.make:61: recipe for target 'onnx/onnx-ml.pb.cc' failed
make[2]: *** [onnx/onnx-ml.pb.cc] Error 1
CMakeFiles/Makefile2:178: recipe for target 'CMakeFiles/gen_onnx_proto.dir/all' failed
make[1]: *** [CMakeFiles/gen_onnx_proto.dir/all] Error 2
make[1]: *** Waiting for unfinished tasks....
[ 11%] Built target onnxifi_wrapper
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
Traceback (most recent call last):
  File "setup.py", line 336, in <module>
    'backend-test-tools = onnx.backend.test.cmd_tools:main',
  File "/home/miniconda3/envs/mxnet_tensorrt/lib/python3.7/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/home/miniconda3/envs/mxnet_tensorrt/lib/python3.7/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/home/miniconda3/envs/mxnet_tensorrt/lib/python3.7/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/home/miniconda3/envs/mxnet_tensorrt/lib/python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
[...]

Issue #1314, opened 2018-08-22
onnx build script always depends on python2


snnn:
OS: fedora 27
build steps:
```bash
dnf install -y curl eigen3-devel gcc gcc-c++ zlib-devel cmake make ninja-build git python3-devel python3-numpy libunwind icu aria2 rsync python3-setuptools python3-wheel
export ONNX_ML=1
aria2c --download-result=hide -d /tmp/src  https://github.com/onnx/onnx/archive/v1.2.2.tar.gz
tar -xf /tmp/src/onnx-1.2.2.tar.gz -C /tmp/src
cd /tmp/src/onnx-1.2.2
python3 setup.py  bdist_wheel
```
output:
```
[  5%] Running C++ protocol buffer compiler on /tmp/src/onnx-1.2.2/.setuptools-cmake-build/onnx/onnx-ml.proto
/usr/bin/env: ‘python’: No such file or directory
--mypy_out: protoc-gen-mypy: Plugin failed with status code 127.
gmake[2]: *** [CMakeFiles/gen_onnx_proto.dir/build.make:62: onnx/onnx-ml.pb.cc] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:100: CMakeFiles/gen_onnx_proto.dir/all] Error 2
gmake: *** [Makefile:130: all] Error 2
Traceback (most recent call last):
  File "setup.py", line 327, in <module>
    'backend-test-tools = onnx.backend.test.cmd_tools:main',
  File "/usr/lib/python3.6/site-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib64/python3.6/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib64/python3.6/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/lib64/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/usr/lib/python3.6/site-packages/wheel/bdist_wheel.py", line 199, in run
    self.run_command('build')
  File "/usr/lib64/python3.6/distutils/cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "/usr/lib64/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/usr/lib64/python3.6/distutils/command/build.py", line 135, in run
    self.run_command(cmd_name)
[...]

PR #2961, opened 2020-08-18
Make onnx build not always depend on python2

jcwchen:
**Description**
Remove !~/usr/bin/env python to avoid some systems use it for python.

**Motivation and Context**
https://github.com/onnx/onnx/issues/1314, https://github.com/onnx/onnx/issues/2658
onnx build script always depends on python2 on some Linux machines like Ubuntu 16.04 or CentOS.
The reason is `/usr/bin/python` always points to python2 in these systems.


CMakeLists.txt
```diff
@@ -239,11 +239,8 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP NAME SRCS HDRS ROOT_DIR DEPEND)
       if(ONNX_GEN_PB_TYPE_STUBS)
         # Haven't figured out how to generate mypy stubs on Windows yet
         if(NOT WIN32)
-          # If onnx was packaged to pypi from Windows, protoc-gen-mypy.py is
-          # missing the +x flag. Add it.
-          execute_process(COMMAND chmod +x ${ROOT_DIR}/tools/protoc-gen-mypy.py)
-          set(PROTOC_MYPY_PLUGIN_FILE ${ROOT_DIR}/tools/protoc-gen-mypy.py)
-        else(NOT WIN32)
```
jcwchen, line 246: I don't know why it set `NOT WIN32` here. It seems that the mypy check has never been used on Windows before.

jcwchen:
Reopen to Rerun CI

jcwchen:
Reopen again. The model urls are sometimes invalid.

CMakeLists.txt
askhade: Not sure I understand this. In windows CI this check is run...

jcwchen:
The CIs in ONNX are not comprehensive enough and I afraid it is still broken in some platforms...
@snnn Do you think this PR can solve this issue https://github.com/onnx/onnx/issues/1314? Thank you

CMakeLists.txt
```diff
@@ -77,6 +77,12 @@ if(NOT ONNX_NAMESPACE)
   set(ONNX_NAMESPACE "onnx")
 endif()
 
+# find_package Python has replaced PythonInterp and PythonLibs since cmake 3.12
+# Use the following command in the future; now this is only compatible with the latest pybind11
+# find_package(Python ${PY_VERSION} COMPONENTS Interpreter DevelopmentREQUIRED)
+find_package(PythonInterp ${PY_VERSION} REQUIRED)
+find_package(PythonLibs ${PY_VERSION} REQUIRED)
```
jcwchen, line 84: Originally ONNX only uses `find_package(PythonInterp)` and `find_package(PythonLibs)` to find python library on Windows... Move here can help finding the right Python on Linux, but I am not sure whether this will break something in some platforms.

snnn (approved):

jcwchen:
:shipit:

tools/protoc-gen-mypy.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+python -u  "$PWD"/../tools/protoc-gen-mypy.py
```
snnn, line 2: But it is still python not python3?
You should always use the python executable which was used for running setup.py to run this script. 
In the cases of setup.py was not used, the path of the  python executable should be got from the find_package calls.

tools/gen_coverage_report.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
```
snnn, line 1: I think you don't need to remove these lines. They're harmless.

tools/protoc-gen-mypy.sh
jcwchen: Thank you @snnn for bringing this up.
Yes, I thought of this, but originally the interface of using `protoc-gen-mypy` here is fixed. The only way is to call a script for `protoc-gen-mypy` so I don't know how to use `PYTHON_EXECUTABLE` from CMake for this (Could we call a CMake environment variable from an outside script?)

Otherwise, since onnx won't support python2 since ONNX 1.8, could we just use python3 here?

tools/gen_coverage_report.py
jcwchen: If we are using `PYTHON_EXECUTABLE` to call them, these shebangs should be fine. If not, they still will be troubles, right? Why don't we just remove them to prevent future mistakes?

snnn:
I think a quick fix would be generating tools/protoc-gen-mypy.py on the fly and putting the full path of PYTHON_EXECUTABLE after the shabang.

jcwchen:
> I think a quick fix would be generating tools/protoc-gen-mypy.py on the fly and putting the full path of PYTHON_EXECUTABLE after the shabang.

Thank you @snnn for the helpful suggestion. I have got PYTHON_EXECUTABLE in the script and recovered all shabangs.
Please help me to review this. Thanks!

tools/protoc-gen-mypy.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# use PYTHON_EXECUTABLE from CMake to get the right python to execute
+PYTHON_EXECUTABLE=$(grep PYTHON_EXECUTABLE:FILEPATH CMakeCache.txt | cut -d "=" -f2)
```
snnn, line 3: See https://cmake.org/cmake/help/v3.19/command/configure_file.html

jcwchen: Thank you for this. Updated. It is more elegant now.

snnn (approved):

snnn (approved):

CMakeLists.txt
snnn: Should they be required only when BUILD_ONNX_PYTHON  is ON?

jcwchen: Make sense. Added a condition for this. Thanks.

snnn (approved):

askhade (approved):

Issue #2965, opened 2020-08-19
Shape inference on the Torchvision’s Mask R-CNN causes a segmentation fault


przybyszewskiw:
# Bug Report

### Is the issue related to model conversion?
Probably not.

### Describe the bug
When I try to run shape inference on the Torchvision's Mask R-CNN it causes a segfault. However, `check_model` doesn't return any warning/exception.

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): Linux Ubuntu 18.04.4
- ONNX version (*e.g. 1.7*):  1.6.0
- Python version: 3.6.10
- PyTorch version: 1.6.0a0+9907a3e
- Torchvision version: 0.7.0a0
- GCC/Compiler version (if compiling from source): N/A
- CMake version: 3.14.0
- Protobuf version: 3.12.2
- Visual Studio version (if applicable): N/A

### Reproduction instructions
```
import torch
import torchvision
import onnx

model = torchvision.models.detection.maskrcnn_resnet50_fpn(pretrained=True)
model.eval()
x = [torch.rand(3, 800, 800)]
torch.onnx.export(model, x, "mask_rcnn.onnx", opset_version = 11)
onnx_model = onnx.load("mask_rcnn.onnx")
onnx.checker.check_model(onnx_model)
onnx.shape_inference.infer_shapes(onnx_model)
```

### Expected behavior
Shape inference doesn't cause any errors.

jcwchen:
Hi @przybyszewskiw,
Sorry that current shape inference does not catch the exception properly. 
This segfault bug will be fixed in this PR: https://github.com/onnx/onnx/pull/2783.

After applying this PR, I tried to check this model with `onnx.checker.check_model('mask_rcnn.onnx', full_check=True)` and got the error as follows:
```
Invalid tensor data type.
```

For detailed error messages, I "shape_inference" the model and got the error as follows:
```
Shape inference error(s): (op_type:Gather, name:Gather_2359): [TypeInferenceError] Input 0 expected to have tensor type
(op_type:Gather, name:Gather_2360): [TypeInferenceError] Input 0 expected to have tensor type
(op_type:Slice, name:Slice_2389): [TypeInferenceError] Input 0 expected to have tensor type
(op_type:Slice, name:Slice_2398): [TypeInferenceError] Input 0 expected to have tensor type

```

przybyszewskiw:
Hi @jcwchen,
[...]

PR #2983, opened 2020-08-28
fix type inference issue (scalar initializers and Resize)

gramalingam:
Fix two issues in type inference

(a) Scalar initializers type (tensor element type) was not being set.

(b) Resize type inference had the check for shape-inference before the type-inference (failing to propagate type).

These changes fix the issue mentioned in https://github.com/onnx/onnx/issues/2965


onnx/shape_inference/implementation.cc
```diff
@@ -191,15 +191,13 @@ static void InferShapesImpl(
...
+    initializerTensorType->set_elem_type(tp.data_type());
     // set the shape according to the initializer shape info
-    // if the shape info of initializer is not empty
-    if (tp.dims_size() != 0) {
-      initializerTensorType->set_elem_type(tp.data_type());
-      TensorShapeProto* shape = initializerTensorType->mutable_shape();
-      for (int i = 0 ; i < tp.dims_size(); ++i) {
-        shape->add_dim()->set_dim_value(tp.dims(i));
-      }
+    TensorShapeProto* shape = initializerTensorType->mutable_shape();
```
askhade, line 196: I think this check -> tp.dims_size() !=0 is needed before output tensor shape is initialized?
This can cause issues because has_shape for this tensor will return true from here onwards and depending what the calling code (whether it does necessary checks or not ) it can cause segfaults

gramalingam: The dims_size check here is not related to whether the shape is known or not known. Here, the shape is known exactly. If dims_size is 0, it means it is rank 0 (a scalar). This check is different from a "has_shape" kind of check.

jcwchen: I added `tp.dims_size() != 0` because 3 tests in `shape_inference_test` will fail without this. My thought was same as @askhade that this can check whether the shape is initialized.
If `dims_size=0` means a scalar, there might be other potential bugs in `splitToSequence` because the ranks of initializer and input are different: `RuntimeError: Inferred shape and existing shape differ in rank: (0) vs (1)
`.

gramalingam: It is possible that there are bugs elsewhere (in requiring rank 0 or rank 1 for supposedly-scalar inputs). Or, if the spec calls for rank 1, and the model uses rank 0, it is an error in the model. Adding the check here will only mask problems elsewhere, I think.

gramalingam: The split_to_sequence testcases had an issue. I fixed them. Hope this goes through the CI.

jcwchen:
The fix of resize, initializer and tests looks good to me. Thank you for proposing this @gramalingam

PR #2987, opened 2020-09-01
Add size check to make_tensor

gramalingam:
Add a check in make_tensor to ensure that the specified dims and the number of elements in the tensor are consistent. Fix a couple of bugs in shape-inference-tests that were violating this condition.


linkerzhang (approved):
good enhancement! Thank you!

PR #2994, opened 2020-09-03
util: function to extract part of an ONNX model

jackwish:
Sometimes, people would like to _extract_ part of an model for development, validation or other purpose.

With `onnx.util.extract`, this is doable by specifying the input and output tensor names.


jackwish:
@linkerzhang @houseroad Would you mind to take a look at this PR if you have time? I am not sure where ONNX repo may accept such utils.

gramalingam:
Looks useful to me. A quick question/comment: I think this does not handle control-flow ops (like loop/if-then-else) correctly, since the nested scopes (sub-graphs) in those ops may contain references to other tensors from the outer scope (outer graph). If that is not fixed, it may be worth documenting it in the description as a limitation.

jackwish:
@gramalingam Thanks for your comment! Yes, I didn't try on control-flow ops, but maybe it's insignificant:
* If the _boundary_ (defined by input and output tensors) is not cutting the subgraph, control-flow ops are just like other ops.
* If the _boundary_ is cutting the subgraph, it would be a pretty tricky scenario.  But yes, this is not supported.

I will document it clear later. Thanks for your advice!

jackwish:
The doc issue mentioned in https://github.com/onnx/onnx/pull/2994#issuecomment-697845963 has been resolved.

Would you mind to take a look at this if you have time? @gramalingam @linkerzhang @houseroad

onnx/utils.py
```diff
@@ -21,3 +24,138 @@ def polish_model(model):  # type: (ModelProto) -> ModelProto
     model = onnx.optimizer.optimize(model)
     onnx.checker.check_model(model)
     return model
+
+
+class Extractor:
+    def __init__(self, model):  # type: ignore
```
askhade, line 30: it would be best if you add type annotations.

onnx/utils.py
```diff
@@ -21,3 +24,138 @@ def polish_model(model):  # type: (ModelProto) -> ModelProto
...
+        tensors = []
+        for name in names_keep:
+            tensors.append(tmap[name])
+        for name in names_add:
+            # activation become input or output
+            tensors.append(self.vimap[name])
+
+        # sort the tensors
+        new_tmap = self._buildNameDict(tensors)
+        sorted_tensors = [new_tmap[name] for name in new_names]
```
askhade, line 56: why is sorting required here?

onnx/utils.py
```diff
@@ -21,3 +24,138 @@ def polish_model(model):  # type: (ModelProto) -> ModelProto
...
+
+class Extractor:
+    def __init__(self, model):  # type: ignore
+        self.model = onnx.shape_inference.infer_shapes(model)
+        self.graph = self.model.graph
+        self.wmap = self._buildNameDict(self.graph.initializer)
+        self.vimap = self._buildNameDict(self.graph.value_info)
+
+    @staticmethod
+    def _buildNameDict(objs):  # type: ignore
```
askhade, line 37: adapt similar naming convention as other python files, basically this should be _build_name_dict

onnx/utils.py
```diff
@@ -21,3 +24,138 @@ def polish_model(model):  # type: (ModelProto) -> ModelProto
...
+        meta = {
+            'ir_version': self.model.ir_version,
+            'opset_imports': self.model.opset_import,
+            'producer_name': 'onnx.utils.extract',
+        }
+        return onnx.helper.make_model(graph, **meta)
+
+    def extract(self, input_names=None, output_names=None):  # type: ignore
+        graph = self.model.graph
+        if input_names is None:
```
askhade, line 113: if input or output names are not provided then you extract the entire graph. Why not display an error instead?

onnx/utils.py
```diff
@@ -21,3 +24,138 @@ def polish_model(model):  # type: (ModelProto) -> ModelProto
...
+        self.model = onnx.shape_inference.infer_shapes(model)
+        self.graph = self.model.graph
+        self.wmap = self._buildNameDict(self.graph.initializer)
+        self.vimap = self._buildNameDict(self.graph.value_info)
+
+    @staticmethod
+    def _buildNameDict(objs):  # type: ignore
+        return {obj.name: obj for obj in objs}
+
+    def _filterTensors(self, originals, new_names):  # type: ignore
```
askhade, line 40: nit: rename originals to existing_tensors and new_names to something like to_be_filtered or to_be_extracted?

docs/PythonAPIOverview.md
```diff
@@ -242,13 +242,32 @@ Function `polish_model` runs model checker, optimizer, shape inference engine on
...
+
+```python
+import onnx
+
+input_path = 'path/to/the/original/model.onnx'
+output_path = 'path/to/save/the/extracted/model.onnx'
+input_names = ['input_0', 'input_1', 'input_2']
+output_names = ['output_0', 'output_1']
+
+onnx.utils.extract(input_path, output_path, input_names, output_names)
```
askhade, line 264: rename to extract_model?

onnx/utils.py
gramalingam: I think it is to ensure that the generated model has inputs/outputs in the order specified by the user (via the parameters).

onnx/utils.py
gramalingam: Interestingly, in this case it ends up being a "model optimization" that eliminates nodes that are not used (in computing any of the output values). So, the current behavior seems to have a useful usecase. But it is worth documenting.

onnx/utils.py
```diff
@@ -21,3 +24,138 @@ def polish_model(model):  # type: (ModelProto) -> ModelProto
...
+        sorted_tensors = [new_tmap[name] for name in new_names]
+        return sorted_tensors
+
+    def _filterInputs(self, names):  # type: ignore
+        return self._filterTensors(self.graph.input, names)
+
+    def _filterOutputs(self, names):  # type: ignore
+        return self._filterTensors(self.graph.output, names)
+
+    def _dfsSearchReachableNodes(self, node_output_name, graph_input_names, reachable_nodes):  # type: ignore
```
gramalingam, line 65: It seems like if a graph-input X is required to compute a specified-output Y, but if X is not in the specified-inputs, we could end up producing an incorrect model. May be it would help to check for this condition?

onnx/utils.py
jackwish: Yes, this is to make the order of inputs/outputs of the generated model to be the same as user input. This is not necessary but having this won't make user supervised.

onnx/utils.py
jackwish: Yes, the _missing input-X_ won't be taken as _input of the resulted model_. But I am not planning to detect such a corner case as it seems not to be easy work. Hope that the user may avoid this, and it can be caught by `onnx.check_model()` if we really encounter some...

jackwish:
@askhade @gramalingam Thank you for your careful review, I have updated accordingly, please help to have a look. :)

onnx/utils.py
```diff
@@ -21,3 +24,154 @@ def polish_model(model):  # type: (ModelProto) -> ModelProto
...
+    The sub-model is defined by the names of the input and output tensors *exactly*.
+
+    Note: For control-flow operators, e.g. If and Loop, the _boundary of sub-model_,
+    which is defined by the input and output tensors, should not _cut through_ the
+    subgraph that is connected to the _main graph_ as attributes of these operators.
+
+    Arguments:
+        input_path (string): The path to original ONNX model.
+        output_path (string): The path to save the extracted ONNX model.
+        input_names (list of string): The names of the input tensors that to be extracted.
```
askhade, line 160: Looks like when input_names are not provided the the extracted sub graph uses the original inputs. This should be documented.

onnx/utils.py
```diff
@@ -21,3 +24,154 @@ def polish_model(model):  # type: (ModelProto) -> ModelProto
...
+            output_names,  # type: List[Text]
+    ):  # type: (...) -> List[NodeProto]
+        reachable_nodes = list()  # type: ignore
+        for name in output_names:
+            self._dfs_search_reachable_nodes(name, input_names, reachable_nodes)
+        # needs to be topology sorted.
+        nodes = [n for n in self.graph.node if n in reachable_nodes]
+        return nodes
+
+    def _collect_reachable_tensors(
```
askhade, line 93: To optimize I think this can be done along with _collect_reachable_nodes. This is nice to have.

onnx/utils.py
jackwish: All member functions have type annotations except two which are sort complex...

onnx/utils.py
jackwish: Remove the _fallback_ like logic - input/output names are taken as they are.

jackwish: For the _model optimization_ like behavior, as we are extracting sub-models - which is removing _unused_ or _unneeded_ nodes and tensors - it seems to be something as expected to me :)

onnx/utils.py
jackwish: Original input/output names are used in the _first commit_, and the current ([change](https://github.com/onnx/onnx/pull/2994/commits/dfe238f82e10f2a1c294841838535747651f851a?file-filters%5B%5D=.py#diff-498e239a2d86d6bcf902378224869c79L114)) implementation simply takes them as they are - empty is empty.

onnx/utils.py
jackwish: Yes, we can do that. IMO, _two passes_ is sort clearer to clarify the logic. And, as `_collect_reachable_tensors` doesn't walk the graph but iterate nodes, I guess it won't be significant :) Anyway, I can rewrite it if we really want that.

jackwish:
@askhade @gramalingam I accidentally clicked _Resolve conversation_ button before *Comment* button, and lost some comments in the previous update. I have added the important ones back. If any code or comment confuses you, please raise.

onnx/utils.py
askhade: input names being empty is not valid right? Then shouldn't there be a check for this on line 167?

onnx/utils.py
askhade: As I mentioned this is not a must have but a nice to have.

askhade:
Apart from the 1 open comment I have regarding input names check rest of the PR looks good to me. Once you clarify fix this I will approve the PR.

Thanks for your contribution!

onnx/utils.py
jackwish: It could be surprising but an empty input name is valid actually, though not very meaningful in practice... Empty input names are cases that the extracted model is using _constant inputs_ i.e. initializers. @askhade 

The example below is a `Transpose` node with constant input, and we are extracting the model itself - as it has only one node.
```py
import onnx
from onnx import helper, TensorProto

# generate original ONNX model
M, N, K = 2, 3, 4

wdata = [i for i in range(N*K)]
weight = helper.make_tensor('weight', TensorProto.FLOAT, [N, K], wdata)

output = helper.make_tensor_value_info('output', TensorProto.FLOAT, [K, N])

transpose = helper.make_node('Transpose', ['weight'], ['output'], name='transpose', perm=(1, 0))

graph = helper.make_graph([transpose, ], 'demo', [], [output], [weight])
opset = helper.make_operatorsetid(onnx.defs.ONNX_DOMAIN, 9)
model = helper.make_model(graph, producer_name='toy-demo', ir_version=6, opset_imports=[opset])

onnx.checker.check_model(model)
onnx.save_model(model, 'example.onnx')

# extract model
onnx.utils.extract_model('example.onnx', 'extracted.onnx', list(), ['output'])
```

onnx/utils.py
jackwish: Let's see if we can improve this in the future :)

jackwish:
> Apart from the 1 open comment I have regarding input names check rest of the PR looks good to me. Once you clarify fix this I will approve the PR.
> 
> Thanks for your contribution!

Thank you @askhade, I have updated [the comment](https://github.com/onnx/onnx/pull/2994#discussion_r501600166) regarding your concern.

askhade (approved):

askhade:
@jackwish : Please merge with master again. It is best if you give permission to code maintainers to merge with master to make this more efficient. Thanks!

jackwish:
done @askhade

PR #2997, opened 2020-09-03
Replace data_type_utils.cc throw calls

vinitra:
Follow on to #2888. Removing exception handling in ONNX Runtime is an important step for reducing the size for an ORT minimal build for mobile. This is a key deliverable of the ORT 1.5 release.

Short term fix approach: return an empty TypeProto and empty string short term instead of throwing errors.

For the TypeProto the caller can check:
`returned_type_proto.value_case() == TypeProto::VALUE_NOT_SET`

### DIFF
--- a/onnx/defs/data_type_utils.cc
+++ b/onnx/defs/data_type_utils.cc
@@ -109,8 +109,6 @@ DataType DataTypeUtils::ToType(const std::string& type_str) {
 }
 
 const TypeProto& DataTypeUtils::ToTypeProto(const DataType& data_type) {
-  // If the DataType input is invalid, this function will return an empty TypeProto
-  // and return_value.value_case() will return TypeProto::VALUE_NOT_SET.
   std::lock_guard<std::mutex> lock(GetTypeStrLock());
   auto it = GetTypeStrToProtoMap().find(*data_type);
   if (GetTypeStrToProtoMap().end() == it) {
@@ -172,7 +170,6 @@ std::string DataTypeUtils::ToString(
 }
 
 std::string DataTypeUtils::ToDataTypeString(int32_t tensor_data_type) {
-  // If int32_t input is invalid, this function will return an empty string.
   TypesWrapper& t = TypesWrapper::GetTypesWrapper();
   auto iter = t.TensorDataTypeToTypeStr().find(tensor_data_type);
   if (t.TensorDataTypeToTypeStr().end() == iter) {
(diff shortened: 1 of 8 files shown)


onnx/defs/data_type_utils.cc
```diff
@@ -172,7 +173,7 @@ std::string DataTypeUtils::ToDataTypeString(int32_t tensor_data_type) {
   TypesWrapper& t = TypesWrapper::GetTypesWrapper();
   auto iter = t.TensorDataTypeToTypeStr().find(tensor_data_type);
   if (t.TensorDataTypeToTypeStr().end() == iter) {
-    throw std::invalid_argument("Invalid tensor data type.");
+    return "Invalid tensor data type.";
```
gwang-msft, line 176: The 1st iter of the previous PR #2888 was trying to return an empty string, does the "Invalid tensor data type." actually mean anything to the caller?

onnx/defs/data_type_utils.cc
```diff
@@ -112,7 +112,8 @@ const TypeProto& DataTypeUtils::ToTypeProto(const DataType& data_type) {
   std::lock_guard<std::mutex> lock(GetTypeStrLock());
   auto it = GetTypeStrToProtoMap().find(*data_type);
   if (GetTypeStrToProtoMap().end() == it) {
-    throw std::invalid_argument("Invalid data type.");
+    static const TypeProto empty_type_proto;
```
skottmckay, line 115: Maybe add a comment to the header saying that if the input was invalid they will get an empty TypeProto back, and return_value.value_case() will return TypeProto::VALUE_NOT_SET.

onnx/defs/data_type_utils.cc
gramalingam: Agree. I see pros/cons, but for code that handles this, comparing against an empty string is simpler and easier.

gramalingam: Or, a named string constant.

onnx/defs/data_type_utils.cc
vinitra: Thanks for the good suggestion! Addressed.

onnx/defs/data_type_utils.cc
vinitra: Resolved with an empty string.

onnx/defs/data_type_utils.cc
```diff
@@ -109,10 +109,13 @@ DataType DataTypeUtils::ToType(const std::string& type_str) {
 }
 
 const TypeProto& DataTypeUtils::ToTypeProto(const DataType& data_type) {
+  // If the DataType input is invalid, this function will return an empty TypeProto
```
askhade, line 112: Could you add this in the class header data_type_utils.h where this function is declared.

vinitra: Updated!

askhade (approved):

gramalingam (approved):

Issue #1609, opened 2018-11-13
ConvTranspose auto_pad spec contradiction?


BowenBao:
In operators documentation, ConvTranpose states that 

> output_shape can also be explicitly specified in which case pads values are auto generated using these equations:
> 
> total_padding[i] = stride[i] * (input_size[i] - 1) + output_padding[i] + kernel_shape[i] - output_shape[i]
> **If (auto_pads != SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)**
> **Else: pads[start_i] = total_padding[i] - (total_padding[i]/2); pads[end_i] = (total_padding[i]/2).**

which is to pad extra at upper when auto_pads is _NOT_ SAME_UPPER, however in

> auto_pad : string (default is NOTSET)
> auto_pad must be either NOTSET, SAME_UPPER, SAME_LOWER or VALID. Where default value is NOTSET, which means explicit padding is used. SAME_UPPER or SAME_LOWER mean pad the input so that the output size match the input.In case of odd number **add the extra padding at the end for SAME_UPPER** and at the beginning for SAME_LOWER. VALID mean no padding. DEPRECATION NOTE: auto_pad is only intended to support legacy uses, and for framework authors, one is explicitly encouraged to use explicit padding specified in the pads attribute.

the spec says to pad extra at upper when auto_pads is SAME_UPPER.

linkerzhang:
Thank you! Good catch! The formula should be fixed. Do you want to send a PR for it please?

jcwchen:
It should be resolved by https://github.com/onnx/onnx/pull/3440. Thanks!

Issue #2787, opened 2020-05-24
ConvTranspose auto_pad and explicit pads combination should be invalid


vuzelac-cadence:
```
X = helper.make_tensor_value_info('X', TensorProto.FLOAT, [1, 1, 2, 2])
W = helper.make_tensor_value_info('W', TensorProto.FLOAT, [1, 1, 3, 3])
B = helper.make_tensor_value_info('B', TensorProto.FLOAT, [1])
Y = helper.make_tensor_value_info('Y', TensorProto.FLOAT, [1, 1, 3, 3])
node_def = helper.make_node("ConvTranspose", ["X", "W", "B"], ["Y"], auto_pad="SAME_UPPER", strides=[1, 1], pads=[0, 1, 1, 0])
```

The above excerpt doesn't generate error although the spec for ConvTranspose says it should:
```
pads : list of ints
_...This attribute cannot be used simultaneously with auto_pad attribute...
```

If they can be used together, auto_pad description hints pads are valid only if auto_pad=NONSET. However, I see pads attribute applies even when auto_pad=SAME_UPPER/SAME_LOWER
`

TMVector:
Hi @vuzelac-cadence, thanks for the report. The `onnx.helper.make_node` function has no operator-specific checks. It is the role of the ONNX checker to catch problems like this, but it is not exhaustive by any means. The specification has the final say on whether a node is a valid instantiation of an operator, and ONNX backends/compilers/tools should work with such nodes.
If you would like to add this check to the ONNX checker (if it isn't already there), then please submit a PR :)

vuzelac-cadence:
@TMVector , I failed to write the whole code but the checker is invoked and it passes, and that was the reason to submit this issue, not the helper code. Sorry about that. So, the check is not there.
Hope you can find time to describe better what  this means " **it is not exhaustive by any means**".

IMO, it should run against everything in the spec, otherwise it's incomplete (or can simply say that it has bugs). Is it really that you intended it to be non-exhaustive ?

TMVector:
@vuzelac-cadence thanks for clarifying. Yes, the ONNX project is dependent on people and companies dedicating time to work on it, and unfortunately the checker is not complete, but even with lots of people working full time on it I'm sure it wouldn't be perfect :)
I believe it tries to cover the majority of issues by checking machine-readable properties of operators, such as accepted input types and attributes.

faxu:
CC @jcwchen

jcwchen:
[...]

PR #2097, opened 2019-06-11
Fix ConvTranspose auto_pad spec contradiction (#1609) 

zhaozhixu:
Fixed a contradiction in ConvTranspose auto_pad specification, as @BowenBao said in issue #1609 

> In operators documentation, ConvTranpose states that

    output_shape can also be explicitly specified in which case pads values are auto generated using these equations:

    total_padding[i] = stride[i] * (input_size[i] - 1) + output_padding[i] + kernel_shape[i] - output_shape[i]
    If (auto_pads != SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)
    Else: pads[start_i] = total_padding[i] - (total_padding[i]/2); pads[end_i] = (total_padding[i]/2).

> which is to pad extra at upper when auto_pads is NOT SAME_UPPER, however in

    auto_pad : string (default is NOTSET)
    auto_pad must be either NOTSET, SAME_UPPER, SAME_LOWER or VALID. Where default value is NOTSET, which means explicit padding is used. SAME_UPPER or SAME_LOWER mean pad the input so that the output size match the input.In case of odd number add the extra padding at the end for SAME_UPPER and at the beginning for SAME_LOWER. VALID mean no padding. DEPRECATION NOTE: auto_pad is only intended to support legacy uses, and for framework authors, one is explicitly encouraged to use explicit padding specified in the pads attribute.

> the spec says to pad extra at upper when auto_pads is SAME_UPPER.

PR #3000, opened 2020-09-05
Fix ConvTranspose: enhance attribute check

jcwchen:
**Description**

- Enhance the attribute check for ConvTranspose: `pads` and `auto_pads` can not be used at the same time.
- Add a unit test for it
- (move it to another PR) Update the document for ConvTranspos: make the example be consistent as the doc description

**Motivation and Context**
https://github.com/onnx/onnx/issues/2787


docs/Changelog.md
```diff
@@ -10743,7 +10743,7 @@ This version of the operator has been available since version 11 of the default
   output_shape can also be explicitly specified in which case pads values are auto generated using these equations:
   
     total_padding[i] = stride[i] * (input_size[i] - 1) + output_padding[i] + ((kernel_shape[i] - 1) * dilations[i] + 1) - output_shape[i]
-    If (auto_pads != SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)
+    If (auto_pads == SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)
```
askhade, line 10746: Is this intentional? Why change this ?

onnx/defs/nn/defs.cc
```diff
@@ -1300,7 +1304,7 @@ If the pads parameter is provided the shape of the output is calculated via the
 output_shape can also be explicitly specified in which case pads values are auto generated using these equations:
 
   total_padding[i] = stride[i] * (input_size[i] - 1) + output_padding[i] + ((kernel_shape[i] - 1) * dilations[i] + 1) - output_shape[i]
-  If (auto_pads != SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)
+  If (auto_pads == SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)
```
askhade, line 1307: same comment as above - why is this change needed?

docs/Changelog.md
jcwchen: It is another old issue related to ConvTranspose which hasn't been resolved: https://github.com/onnx/onnx/issues/1609 There is also a PR for it: https://github.com/onnx/onnx/pull/2097
The original reason is the description is not consistent with the example. I thought it's a good chance to push this forward so I added this fix in this PR as well.

jcwchen:
Updating the document for ConvTranspos will move to another PR because the modification needs further review.

askhade:
@gramalingam @ebarsoum @postrational : This PR enforces a condition which is already part of the spec. I do not think we need a version bump here. Thoughts?

askhade (approved):

onnx/defs/nn/defs.cc
```diff
@@ -1191,6 +1191,10 @@ void convTransposeShapeInference(InferenceContext& ctx) {
     if (pads.size() != n_input_dims * 2) {
       fail_shape_inference("Attribute pads has incorrect size");
     }
+    const auto* auto_pad_attr = ctx.getAttribute("auto_pad");
+    if (nullptr != auto_pad_attr) {
+      fail_shape_inference("The pads attribute cannot be used simultaneously with auto_pad attribute");
```
gramalingam, line 1196: Just a minor nit: this is neither a shape_inference error nor a type_inference error. Unfortunately, we have only "fail_type_inference" and "fail_shape_inference". This is okay, but it might be good to have a consistent categorization of errors (may be adding another category of errors) ... not necessarily in this PR, could be later also.

gramalingam (approved):

gramalingam:
> This PR enforces a condition which is already part of the spec. I do not think we need a version bump here. Thoughts?
Agree, this seems fine.

gramalingam:
> @gramalingam @ebarsoum @postrational : This PR enforces a condition which is already part of the spec. I do not think we need a version bump here. Thoughts?

Agree

PR #3002, opened 2020-09-06
Fix float16 data convert issue in numpy_helper.to_array

jcwchen:
**Description**
Since the float16 data is stored as int32, simply using `numpy.astype` for the type conversion would cause a bug. Should further use `numpy.view` to get the right original value.

**Motivation and Context**
https://github.com/onnx/onnx/commit/41ae7f8551c688d56a31e9b38b41221d072a412e
There is bug for converting float16 data from int323 data in `numpy_helper.to_array`


onnx/numpy_helper.py
```diff
@@ -55,13 +55,23 @@ def to_array(tensor):  # type: (TensorProto) -> np.ndarray[Any]
...
                 or tensor_dtype == TensorProto.COMPLEX128):
             data = combine_pairs_to_complex(data)
-        return (
-            np.asarray(
-                data,
-                dtype=storage_np_dtype)
-            .astype(np_dtype)
-            .reshape(dims)
-        )
+        # F16 was stored as int32; Need view to get the original value
```
askhade, line 58: nit: is stored

askhade (approved):
LGTM

jcwchen:
:shipit:

Issue #2995, opened 2020-09-03
Version converter 9-->10 fails with 'Input conv_first.weight is undefined!'


Lisicong9203:
# Bug Report

### Describe the bug
The version converter fails on the attached model 

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*):  Ubuntu 18.04.4 LTS
- ONNX version (*e.g. 1.7*):  1.7.0
- Python version: Python 3.6.9
- GCC/Compiler version (if compiling from source): 
- CMake version:
- Protobuf version:
- Visual Studio version (if applicable):

### Reproduction instructions
- Describe the code to reproduce the behavior.
```
import onnx
from onnx import version_converter, helper
model_path = 'deart_model_fp16.onnx'
original_model = onnx.load(model_path)

converted_model = version_converter.convert_version(original_model, 10)

print('The model after conversion:\n{}'.format(converted_model))

```

error:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-8-4409d2eda4ac> in <module>
      6 # https://github.com/onnx/onnx/blob/master/onnx/version_converter.py#L21
      7 # Apply the version conversion on the original model
----> 8 converted_model = version_converter.convert_version(original_model, 10)
      9
     10 print('The model after conversion:\n{}'.format(converted_model))

~/tf.t130/lib/python3.6/site-packages/onnx/version_converter.py in convert_version(model, target_version)
    164         raise ValueError('VersionConverter only accepts int as target_version, incorrect type: {}'.format(type(target_version)))
    165     model_str = model.SerializeToString()
--> 166     converted_model_str = C.convert_version(model_str, target_version)
    167     return onnx.load_from_string(converted_model_str)

IndexError: Input conv_first.weight is undefined!
```

Onnx model:
 [deart_model_fp16.onnx](https://drive.google.com/open?id=1tO3P9ZxxqCZqi-4RoWQxxGYv2c2i4yrI)

jcwchen:
Hi @Lisicong9203,
It seems that you provide a wrong file (`deart_model.pt`). Could you upload the onnx model again? Thanks.

Lisicong9203:
@jcwchen  I'm so sorry about that , i have update the onnx file.

Lisicong9203:
[...]

PR #3007, opened 2020-09-10
Fix IR gap for version_converter: update from initializer

jcwchen:
**Description**
Use the value from initializer for `version_converter`.

**Motivation and Context**
https://github.com/onnx/onnx/issues/2873, https://github.com/onnx/onnx/issues/2995
For ir_version  >= 4, `graph.initializer` does not have to be included in `graph.input`.
However, old IR only considers the `graph.input` for `version_converter`.
This PR is fixing this IR gap.


onnx/common/ir.h
```diff
@@ -985,6 +989,11 @@ friend struct Value;
   Value * addInput() {
     return input_->addOutput();
   }
+
+  void popInput() {
+    input_->popOutput();
```
linkerzhang, line 994: confusing function call...  popoutput from input :).

jcwchen: I also felt confused for the original `addInput`...but the behavior is correct according to the original design.
I just added some comments for these two functions and hope it can resolve the confusion.
Thank you for bringing this up!

onnx/common/ir_pb_converter.cc
```diff
@@ -266,7 +267,27 @@ std::unique_ptr<Graph> graphProtoToGraph(
...
+      if (value_by_name_of.count(vip.name()) == 0) {
+        auto v = g->addInput();
+        v->setElemType(vip.elem_type());
+        std::vector<Dimension> sizes;
+        for (auto size: vip.sizes()) {
+          sizes.push_back(size);
+        }
+        v->setSizes(sizes);
+        v->setUniqueName(vip.name());
+        value_by_name_of[vip.name()] = v;
```
fumihwh, line 287: Could this part be replaced by using `addInitializerAndInput` in `ir.h`?

jcwchen: Good suggestion. Updated. Thank you

jcwchen:
Still, adding initializer to input and then remove it is kind of a hacky method to work around this. 
@daquexian @fumihwh since you are working on optimizer with `common/ir`, do you have other solution to fix this IR gap issue in your case? Thanks.

daquexian:
> @daquexian @fumihwh since you are working on optimizer with common/ir, do you have other solution to fix this IR gap issue in your case? Thanks.

I have no clear idea either. IMO we can use mlir infra (maybe based on https://github.com/onnx/onnx-mlir) instead of implementing the ir ourselves, but obviously, it cannot be done in the short term.

linkerzhang (approved):

linkerzhang:
shall we merge this? @daquexian @jcwchen

jcwchen:
> shall we merge this? @daquexian @jcwchen

We are still looking for a better way to do this... This PR is more like a hacky workaround (Add initializer to input then remove it).

jcwchen:
Create a new PR to solve DCO issue: https://github.com/onnx/onnx/pull/3676 Close it now and the work will be continued there.

PR #3008, opened 2020-09-10
Create a tool for creating TrainingInfoProto and a test

wschin:
That test can be seen as an example for using that tool.

### DIFF
--- a/onnx/test/training_tool_test.py
+++ b/onnx/test/training_tool_test.py
@@ -95,11 +95,12 @@ class TestTrainingTool(unittest.TestCase):
             list(model.graph.input) + list(model.training_info[0].algorithm.input),
             list(model.graph.output) + list(model.training_info[0].algorithm.output),
             list(model.graph.initializer) + list(model.training_info[0].algorithm.initializer))
-        
+
         # Wrap full training graph as a ModelProto so that we can run checker.
         full_training_model = helper.make_model(full_training_graph)
         full_training_model_with_shapes = shape_inference.infer_shapes(full_training_model)
         onnx.checker.check_model(full_training_model_with_shapes)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/onnx/test/training_tool_test.py
+++ b/onnx/test/training_tool_test.py
@@ -7,7 +7,7 @@ import unittest
 import onnx
 import numpy as np
 from onnx.tools import update_model_dims
-from onnx import helper, numpy_helper, TensorProto
+from onnx import helper, numpy_helper, shape_inference, TensorProto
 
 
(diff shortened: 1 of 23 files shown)


onnx/test/training_tool_test.py
```diff
@@ -0,0 +1,104 @@
...
+        training_info = helper.make_training_info(
+            training_graph, [(B_name, Y_name)], None, None)
+
+        # Create a model with both inference and training information.
+        model = helper.make_model(inference_graph)
+        # Check if the inference-only part is correct.
+        onnx.checker.check_model(model)
+        # Insert training information.
+        new_training_info = model.training_info.add()
+        new_training_info.CopyFrom(training_info)
```
wschin, line 86: After this line, `print (model)` prints this model in text format. It should show inference graph + training graph.

onnx/test/training_tool_test.py
```diff
@@ -0,0 +1,104 @@
...
+        # corresponding fields.
+        full_training_graph = helper.make_graph(
+            list(model.graph.node) + list(model.training_info[0].algorithm.node),
+            'full_training_graph',
+            list(model.graph.input) + list(model.training_info[0].algorithm.input),
+            list(model.graph.output) + list(model.training_info[0].algorithm.output),
+            list(model.graph.initializer) + list(model.training_info[0].algorithm.initializer))
+        
+        # Wrap full training graph as a ModelProto so that we can run checker.
+        full_training_model = helper.make_model(full_training_graph)
```
wschin, line 100: You can also `print (full_training_model)` to see a full training algorithm graph.

onnx/test/training_tool_test.py
```diff
@@ -0,0 +1,104 @@
...
+        A_shape = [2, 2]
+        A_name = 'A'
+        A = np.array(A_shape).astype(np.float)
+        A_initializer = numpy_helper.from_array(A, name=A_name)
+        A_value_info = helper.make_tensor_value_info(
+            A_name, TensorProto.FLOAT, A_shape)
+
+        B_shape = [2, 2]
+        B_name = 'B'
+        B = np.array(B_shape).astype(np.float)
```
askhade, line 25: This will create an np array [2, 2] with shape [2] whereas on line 27 when creating B_value_info the shape is set as [2,2]. What am I missing here?

wschin: My bad. I should create a random function here. 

My mistake here is a good example to support the need of having a runtime test (either ORT or a small runtime) in ONNX repo.

sveta-levitan (approved):
Looks great to me.

askhade (approved):

PR #3012, opened 2020-09-11
Make shape_inference accept model path (support >2GB models)

jcwchen:
**Description**
Create a new API in Python `onnx.shape_inference.infer_shapes_path` for shape_inference
Make shape_inference accept model path in C++
Output the inferred model into the specified path via C++ (Default is the original model path)
Update Python API document

**Motivation and Context**
`onnx.checker` does support string path to `check_model`. It is useful for models > 2GB. `onnx.shape_inference` should support it as well.


jcwchen:
cc @yufenglee @YUNQIUGUO

onnx/shape_inference/implementation.h
```diff
@@ -195,6 +195,12 @@ void InferShapes(
     const ISchemaRegistry* schema_registry = OpSchemaRegistry::Instance()
     );
 
+void InferShapes(
+    const std::string& model_path,
+    const bool check_type = false,
```
gramalingam, line 200: It would be nice to add a comment explaining this flag "check_type". I don't recall and the name itself doesn't clarify it.

onnx/shape_inference/implementation.cc
```diff
@@ -371,6 +372,38 @@ void InferShapes(
...
+    fail_check(
+        "Unable to parse model from file:",
+        model_path,
+        ". Please check if it is a valid protobuf file of model.");
+  }
+  InferShapes(model, check_type, schema_registry);
+  // Save the inferred model to the original model path
+  // TODO catch fail status
+  std::fstream output(model_path, std::ios::out | std::ios::trunc | std::ios::binary);
+  if (!model.SerializeToOstream(&output)) {
```
gramalingam, line 400: Would it be better to support a separate input_model_path and a separate output_model_path?

onnx/shape_inference/implementation.h
jcwchen: I was also confused about this... I think it is related to the full_check for checker. I will add some comments for it. Thank you for the suggestion!

onnx/shape_inference/implementation.cc
jcwchen: Sounds good. We can add an optional arguments, `output_model_path`, for Python API. And I think the `output_model_path` needs to be in the same directory as the original model_path because this output only works with the main onnx model file... We should state this limitation for the `output_model_path`.

jcwchen, line 400: This PR will fail in Linux-CI because LITE_PROTO on Linux does not support `SerializeToOstream`(Windows and Mac do support)... No sure whether LITE_PROTO is a must for building onnx.

jcwchen: Solved by directly using output string to file stream.

jcwchen:
Updated. This PR is ready for review. Thanks

jcwchen:
Hi @gramalingam,
Do you have other concern about this PR? I hope to merge it for ONNX 1.8 Release. Thank you!

gramalingam (approved):

gramalingam:
Hi @jcwchen : do you need to signoff on this PR?

jcwchen:
> Hi @jcwchen : do you need to signoff on this PR?

Thanks for the approval. I am still looking for a good way to sign-off old commits excluding merge commits... Simply use `git rebase HEAD~24 --signoff` will mess the commit log

jcwchen:
Signed. After all CIs pass, let's :shipit: . Thank you @gramalingam

Issue #3013, opened 2020-09-11
Shape inference issue with Loop output


askhade:
# Bug Report

### Describe the bug
Reported by @BowenBao . Adding the issue here to be able to track it and reference in a PR.
[loop_repro.zip](https://github.com/onnx/onnx/files/5211869/loop_repro.zip)
Attached are 2 models loop_pass and loop_fail
Download and run the script
import onnxruntime
import torch
x = torch.randn(3,4,5)
y = torch.randn(4,5)

sess = onnxruntime.InferenceSession('model.onnx')
ort_out = sess.run(None, {sess.get_inputs()[0].name: x.numpy(), sess.get_inputs()[1].name: y.numpy()})

Observe the following error message 
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Node (Concat_33) Op (Concat) [ShapeInferenceError] axis must be in [-rank, rank-1].

However, inspecting the model it is obvious the input has rank 3 which makes the range valid for axis=2.

askhade:
The root cause is that in loop shape inference it starts output shape inference without checking whether shape info is available. This causes propagation of incomplete shape which messes up the following node's shape inference.

PR #3014, opened 2020-09-12
fix shape inference for loop

askhade:
Fixes issue:  https://github.com/onnx/onnx/issues/3013


BowenBao (approved):
LGTM, thanks!

gramalingam (approved):

neginraoof:
@askhade 
Looks like you have updated the Loop shape inference for opset8, but not 11.
I think we need a similar logic to check whether shape ino is available in line:
https://github.com/askhade/onnx/blob/2cf8ea9567d7cf8e5ac19ae88ba7de5046a98ba7/onnx/defs/controlflow/old.cc#L886

PR #3019, opened 2020-09-13
Update spec for Convtranspose to make it sync

jcwchen:
**Description**
Update the spec to make it synchronous. (Description part vs if-else part)
Description part should be corrected. In addition, onnxruntime should also be corrected.: https://github.com/microsoft/onnxruntime/blob/a5530358c9236f20e2ad3d6ce968219b21426874/onnxruntime/core/providers/cpu/nn/conv_transpose_attributes.h#L184)
The behavior of SAME_UPPER and SAME_LOWER should be same as the same attribute (auto_pad) used by other pooling or conv related operators.

**Motivation and Context**
https://github.com/onnx/onnx/issues/1609,  https://github.com/onnx/onnx/pull/2097
The spec for ConvTranspose is not synchronous.


jcwchen:
cc @gramalingam @ebarsoum @postrational. Thank you!

gramalingam:
@jcwchen : is this still relevant (given the PR https://github.com/onnx/onnx/pull/3188 )? Thanks!

jcwchen:
> @jcwchen : is this still relevant (given the PR #3188 )? Thanks!

Yes. I believe so. This issue is only related to ConvTranspose that its SAME_UPPER/SAME_LOWER has contradicted definition in the document. There is a related fix in ONNXRuntime as well. I will go back to deal with this when I have more bandwidth. Thank you!

Issue #3004, opened 2020-09-08
Using GCC(<5.0) compile ONNX may cause some error.


462630221:
# Ask a Question
Using GCC(<5.0) compile ONNX may cause some shallow copy error.

### Question
For example:
'''
//...
Tensor B=A;                            // A is Tensor, it's is_raw_data_ is true.
B.scale_by_first_dim(t);         //  t is also a Tensor
//...
'''

- This case, we only want change Tensor B value. But if we using GCC(<5.0) compile ONNX. It will change the value of Tensor A as well. Tensor B=A  happen shallow copy. But we want deep copy.

- The reason for this problem is that gcc(<5.0) use Copy-On-Write(COW) for std::string. If Tensor B is raw_data and Tensor::scalar_by_first_dim which use pointer to change values will also change Tensor A values.  I think  using pointer and system don't detect it, so did not happened COW.

- Adding copy constructor can avoid this problem. I'm not sure this is a bug to ONNX. But I don't see ONNX requiring a version of GCC either.

Thanks!

'''
//onnx/common/tensor.h
...
struct Tensor final {
private:
    //...
    std::string raw_data_;
    //...
public:
    Tensor()
  	: is_segment_(false)
  	, segment_begin_(0)
  	, segment_end_(0)
  	, has_name_(false)
  	, elem_type_(ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED)
  	, is_raw_data_(false)
  	{}
  //...
}  
'''

daquexian:
Wow! Is it the root cause of https://github.com/onnx/onnx/issues/2677? @askhade

daquexian:
I confirmed that the following patch fixes #2677

```diff
diff --git a/onnx/common/tensor.h b/onnx/common/tensor.h
index 1db2e0f8..3cdf8850 100644
--- a/onnx/common/tensor.h
+++ b/onnx/common/tensor.h
@@ -12,6 +12,49 @@
 namespace ONNX_NAMESPACE {

 struct Tensor final {
+    Tensor(const Tensor& other):
+        is_segment_(other.is_segment_) ,
+        segment_begin_(other.segment_begin_) ,
+        segment_end_(other.segment_end_) ,
+        has_name_(other.has_name_) ,
+        name_(other.name_) ,
+        elem_type_(other.elem_type_) ,
+        sizes_(other.sizes_),
+        float_data_(other.float_data_),
+        double_data_(other.double_data_),
[...]

PR #3021, opened 2020-09-15
Using GCC(<5.0) compile ONNX may cause some error.( #3004)

462630221:
Modify tensor.h. GCC(<5.0) use Copy-On-Write(COW) for std::string. It may cause shallow copy. So I add copy constructor that force deep copy in struct tensor, to avoid shallow copy.


onnx/common/tensor.h
```diff
@@ -50,6 +50,51 @@ struct Tensor final {
...
+    // Deep copy. Avoid copy on write when using gcc<5.0
+    string_data_.resize(other.string_data_.size());
+    for(unsigned int i=0; i<other.string_data_.size(); ++i) {
+      string_data_[i] = std::string( other.string_data_[i].data(), other.string_data_[i].size() );
+    }
+    name_ = std::string(other.name_.data(), other.name_.size());
+    raw_data_ = std::string(other.raw_data_.data(), other.raw_data_.size());
+  }
+
+  Tensor& operator = (const Tensor &ohther){
```
daquexian, line 74: typo, ohther->other

onnx/common/tensor.h
```diff
@@ -50,6 +50,51 @@ struct Tensor final {
...
+    if(this != &other){
+        is_segment_ = other.is_segment_;
```
daquexian, line 76: inconsistent indent

onnx/common/tensor.h
```diff
@@ -50,6 +50,51 @@ struct Tensor final {
...
+        has_name_ = other.has_name_;
+        elem_type_ = other.elem_type_;
+        sizes_ = other.sizes_;
+        float_data_ = other.float_data_;
+        double_data_ = other.double_data_;
+        int32_data_ = other.int32_data_;
+        int64_data_ = other.int64_data_;
+        is_raw_data_ = other.is_raw_data_;
+        // Deep copy. Avoid copy on write when using gcc<5.0
+        string_data_.resize(other.string_data_.size());
```
daquexian, line 88: This copy assignment implementation doesn't guarantee strong exception safety. Maybe [copy-and-swap](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Copy-and-swap) like [here](https://github.com/onnx/onnx/issues/3004#issuecomment-692487726) is a better implementation (It's just an advice).

daquexian: > 1 for Inconsistent definition of copy constructor and assignment ('Rule of Two')

LGTM analysis also complains about this

onnx/common/tensor.h
```diff
@@ -50,6 +50,51 @@ struct Tensor final {
...
+  , segment_begin_(other.segment_begin_)
+  , segment_end_(other.segment_end_)
+  , has_name_(other.has_name_)
+  , elem_type_(other.elem_type_)
+  , sizes_(other.sizes_)
+  , float_data_(other.float_data_)
+  , double_data_(other.double_data_)
+  , int32_data_(other.int32_data_)
+  , int64_data_(other.int64_data_)
+  , is_raw_data_(other.is_raw_data_) {
```
daquexian, line 64: uint64_data_ is missing

daquexian (approved):
I have verified that this patch fixes #3004

daquexian:
@462630221 Could you please sign the CLA and merge the latest master branch?

linkerzhang (approved):
lgtm.

PR #3022, opened 2020-09-15
fix cmake build when there is no unversioned python interpreter

helmutg:
When building with cmake without setting PYTHON_EXECUTABLE, the build
fails unless the system has an unversioned python. The unversioned
variant is phased out by various Linux distributions including Debian.
Use find_package to locate python as it knows how to find versioned
variants.


jcwchen:
Hi @helmutg,
Thank you for proposing this.
I think another PR https://github.com/onnx/onnx/pull/2961 is trying to solve the same problem as yours.
Could you try that PR to solve your problem? (Remove those files with `#!/usr/bin/env python`)

Other than that, can this PR fix your issue? I thought if those shebangs exist, some OS systems will still use `/usr/bin/python` as python for those files while building onnx and cause the error.

helmutg:
These PRs are certainly related on a conceptual level. However, this one is not about using Python 2.x vs using Python 3.x. I can already build onnx with only Python 3.X installed and no Python 2.x provided that I pass `-DPYTHON_EXECUTABLE=python3` to cmake. So from my pov, https://github.com/onnx/onnx/pull/2961 seems practically solved and all that is left is cleaning up unused(!) shebang lines and such, which is what the PR actually does.

This PR is about actually detecting a Python 3.x when it is not linked as `/usr/bin/python`, but only linked as `/usr/bin/python3.X` and `/usr/bin/python3`. The build system does not use the shebang lines anyway it explicitly calls Python, which is why this separate PR is necessary.

I hope this makes the need more clear, please follow up if it doesn't.

CMakeLists.txt
```diff
@@ -195,11 +195,7 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP NAME SRCS HDRS ROOT_DIR DEPEND)
       file(MAKE_DIRECTORY "${OUTPUT_PROTO_DIR}")
     endif()
 
-    if("${PYTHON_EXECUTABLE}" STREQUAL "")
-      set(_python_exe "python")
-    else()
-      set(_python_exe "${PYTHON_EXECUTABLE}")
-    endif()
+    find_package(PythonInterp REQUIRED)
```
jcwchen, line 198:
```suggestion
    if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")
        # find_package Python has replaced PythonInterp and PythonLibs since cmake 3.12
        find_package(Python ${PY_VERSION} REQUIRED)
    else()
        find_package(PythonInterp ${PY_VERSION} REQUIRED)
        find_package(PythonLibs ${PY_VERSION} REQUIRED)
    endif()
```

jcwchen: I suggest using `find_package(Python)` for `cmake >=3.12.0`.

`find_package(PythonInterp)` and `find_package(PythonLibs)` has been deprecated since cmake 3.12.0
https://cmake.org/cmake/help/v3.12/module/FindPythonInterp.html
Even though they can work now, it would be better to use the latest API to prevent failures in the future.

BTW, in line#446, there is a same function doing the same thing. We should try to merge these things. Thanks!

jcwchen (changes requested):

CMakeLists.txt
helmutg: I think it should be `find_package(Python ${PY_VERSION} REQUIRED COMPONENTS Interpreter)` for recent cmake.

I actually considered using that when writing the patch, but opted for not doing so for two reasons:
* The minimum cmake version that is declared in the `CMakeLists.txt` does not allow doing so.
* The deprecated variant is already elsewhere.

Therefore it is consistent to also use it here. Using the current syntax is a good thing, but only if it allows removing the deprecated code. Keeping both seems to be the worst of both ways. Given cmake's approach to backwards compatibility, it does not seem likely for it to remove the deprecated variant anytime soon.

For these reasons, I'd leave it as is.

Merging it with the later invocation conceptually makes sense. However, these were separate prior to this PR (i.e. not a regression) and we're trying to fix a build failure here. Cleaning this up risks deferring this bug fix. In case you insist, where would you place the shared `find_package` call? Around line 80?

jcwchen: Thank you for the explanation. My viewpoint is we should use the latest API whenever applicable. Actually the original deprecated variant (line#446) in CMakeLists.txt might cause an error on Windows https://github.com/onnx/onnx/pull/1631.

Anyway it's a suggestion and I understand your concern. If you don't have bandwidth to handle this, I will try it in my PR. Thank you for catching this issue.

(In case you are interested, around line 80 is good.)

helmutg:
What do you think is missing to merge this?
* You suggested using the non-deprecated `find_package(Python COMPONENTS INTERPRETER)` which I think is orthogonal to the bug at hand and should be done when the minimum CMake version is raised. Can we agree to not do that here?
* You asked to fuse the `find_package(PythonInterp ...)` calls. I implemented this.
* circleci seems to be universally broken.

Is there anything left? Should I click the "Resolve conversation" button above?

jcwchen:
`circleci` will be deprecated so you can just ignore it. Yes you can resolve the conversation. @onnx/sig-archinfra-approvers please review this PR. Thanks.

helmutg:
What is missing to merge this?

helmutg:
Could someone explain why this was closed? What exactly has changed?

helmutg:
I tried merging the branch and noticed that no changes are left. The issue apparently was fixed via f8a613acd77f63ff558f451ef85bf81454070010.

PR #3026, opened 2020-09-20
mark scales as optional and remove tf_half_pixel_for_nn in resize op

daquexian:
1. mark 'scales' of the resize op as an optional input (https://github.com/onnx/onnx/pull/2948#issuecomment-671950755). 

2. remove coordinate_transformation_mode 'tf_half_pixel_for_nn'. This mode was introduced by me for resize_nearest half_pixel of tensorflow, but it is indeed unnecessary. 
   The equation of nearest + half_pixel is `x' = round((x + 0.5) / scale - 0.5)`, while in tf it is optimized as `x' = floor((x + 0.5) / scale)`. As a result, I introduced a transformation mode 'tf_half_pixel_for_nn' for the transformation `(x + 0.5) / scale` in PR https://github.com/onnx/onnx/pull/2057. 
    However, as shown above, it should be replaced with a plain 'half_pixel' + 'round_prefer_ceil'


docs/Operators.md
```diff
@@ -15669,15 +15666,15 @@ x_original = length_resized > 1 ? start_x * (length_original - 1) + x_resized *
...
-<dt><tt>scales</tt> : tensor(float)</dt>
-<dd>The scale array along each dimension. It takes value greater than 0. If it's less than 1, it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should be the same as the rank of input 'X'. Only one of 'scales' and 'sizes' can be specified. If 'size' is specified, then set scales to empty data (zero shape) in this operator's input list.</dd>
+<dt><tt>scales</tt> (optional) : tensor(float)</dt>
+<dd>The scale array along each dimension. It takes value greater than 0. If it's less than 1, it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should be the same as the rank of input 'X'. Only one of 'scales' and 'sizes' can be specified. If 'size' is needed, the user can use an empty string as the name of 'scales' in this operator's input list.</dd>
 <dt><tt>sizes</tt> (optional) : tensor(int64)</dt>
 <dd>The size of the output tensor. The number of elements of 'sizes' should be the same as the rank of input 'X'. Only one of 'scales' and 'sizes' can be specified.</dd>
```
hariharans29, line 15679: Since both scales and sizes are now "optional" per schema - should we go ahead and include a line that says - one of the two optional inputs MUST be specified. We already have a line that says "only one can be specified...", maybe we can say - "one of the two MUST be specified and it is an error if both are specified."

Also does the onnx checker catch cases where the Resize node has only two inputs - which means it is missing one of the two ? Also, `Inputs (2-4) ` - is that correct ?

docs/Operators.md
```diff
@@ -15669,15 +15666,15 @@ x_original = length_resized > 1 ? start_x * (length_original - 1) + x_resized *
 <dd>Four modes: round_prefer_floor (default, as known as round half down), round_prefer_ceil (as known as round half up), floor, ceil. Only used by nearest interpolation. It indicates how to get "nearest" pixel in input tensor from x_original, so this attribute is valid only if "mode" is "nearest".</dd>
 </dl>
 
-#### Inputs (3 - 4)
+#### Inputs (2 - 4)
 
 <dl>
 <dt><tt>X</tt> : T1</dt>
 <dd>N-D tensor</dd>
 <dt><tt>roi</tt> : T2</dt>
```
askhade, line 15674: why not make roi optional as well? since it only takes effect when coordinate_transformation_mode is "tf_crop_and_resize"

docs/Operators.md
```diff
@@ -15669,15 +15666,15 @@ x_original = length_resized > 1 ? start_x * (length_original - 1) + x_resized *
 <dd>Four modes: round_prefer_floor (default, as known as round half down), round_prefer_ceil (as known as round half up), floor, ceil. Only used by nearest interpolation. It indicates how to get "nearest" pixel in input tensor from x_original, so this attribute is valid only if "mode" is "nearest".</dd>
 </dl>
 
-#### Inputs (3 - 4)
+#### Inputs (2 - 4)
```
askhade, line 15669: Should this be 2 to 3? Because Sizes and Scales should not be specified together right.

askhade:
@daquexian : Marking this PR for 1.8 release. Let's try and get this change checked in before 1.8 to avoid another version bump for resize.

docs/Operators.md
daquexian: @askhade @hariharans29 I agree that here should be 2-3, but this line is generated automatically. Is there any way to overwrite it?

docs/Operators.md
daquexian: > Also does the onnx checker catch cases where the Resize node has only two inputs - which means it is missing one of the two ? Also, Inputs (2-4) - is that correct ?

Agreed. But it seems like the infra (checker and doc generator) cannot support "one of these optional inputs must be specified".

daquexian:
@askhade @hariharans29 I have updated the spec to "one of the two MUST be specified and it is an error if both are specified" and made the input "roi" also optional. When it comes to "input (2-4)" and the checker problem, maybe we need to update the infra to support the case "one of these optional inputs must be specified" first?

hariharans29:
> @askhade @hariharans29 I have updated the spec to "one of the two MUST be specified and it is an error if both are specified" and made the input "roi" also optional. When it comes to "input (2-4)" and the checker problem, maybe we need to update the infra to support the case "one of these optional inputs must be specified" first?

Thanks for adding the line to the spec. It leaves no place for ambiguity atleast with the spec part. It does a look a little weird to see (2-4) when actually the correct line there must read (2-3), but looks like ONNX doesn't have the infrastructure to deal with "inter-dependence" of optional inputs/outputs (i.e.) we are not able to say - 'scales' and 'sizes' are individually optional both they are NOT both optional. So, I guess we can leave it for now and enhance the infrastructure for later.

For the second aspect of the PR - deprecating `tf_half_pixel_for_nn`- does this cause any issue for a model that is using this attribute and trying to upgrade to opset-13 ? It seems like (from your comment in the PR) that all the converter (or opset upgrader) would have to do is replace `tf_half_pixel_for_nn` with `half_pixel' + 'round_prefer_ceil` and the upgrade will be seamless. Is this correct or do you foresee any other upgrade issues faced by models already using this attribute ?

docs/TestCoverage.md
```diff
@@ -9760,7 +9760,6 @@ data = np.array([[[
 ]]], dtype=np.float32)
 
 roi = np.array([], dtype=np.float32)
-scales = np.array([], dtype=np.float32)
```
hariharans29, line 9763: should we add a test where scales exists and sizes does not ?

onnx/defs/tensor/defs.cc
```diff
@@ -2131,15 +2128,15 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "T2",
-            OpSchema::Single)
+            OpSchema::Optional)
         .Input(
             2,
             "scales",
             "The scale array along each dimension. It takes value greater than 0. If it's less than 1,"
             " it's sampling down, otherwise, it's upsampling. The number of elements of 'scales' should"
-            " be the same as the rank of input 'X'. Only one of 'scales' and 'sizes' can be specified. If 'size' is specified, then set scales to empty data (zero shape) in this operator's input list.",
+            " be the same as the rank of input 'X'. One of 'scales' and 'sizes' MUST be specified and it is an error if both are specified. If 'size' is needed, the user can use an empty string as the name of 'scales' in this operator's input list.",
```
hariharans29, line 2137: nit: If 'sizes' is needed (not if 'size' is needed) for consistency

hariharans29:
Overall, I am good with this PR - left a couple of minor comments

onnx/defs/tensor/defs.cc
daquexian: Thanks. Fixed

docs/TestCoverage.md
daquexian: Thanks. There are already such tests before this PR submitted (e.g., https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/resize.py#L341) .

daquexian:
> For the second aspect of the PR - deprecating tf_half_pixel_for_nn- does this cause any issue for a model that is using this attribute and trying to upgrade to opset-13 ? It seems like (from your comment in the PR) that all the converter (or opset upgrader) would have to do is replace tf_half_pixel_for_nn with half_pixel' + 'round_prefer_ceil and the upgrade will be seamless. Is this correct or do you foresee any other upgrade issues faced by models already using this attribute ?

@hariharans29 Yes, it is correct. All the converter only have to replace replace 'tf_half_pixel_for_nn' with 'half_pixel' + 'round_prefer_ceil'.

hariharans29 (approved):
LGTM

askhade:
@daquexian : It seems the tests were not updated after roi was made optional. Can you update the test script and regenerate the tests again. Thanks!

askhade (changes requested):
@daquexian : It seems the tests were not updated after roi was made optional. Can you update the test script and regenerate the tests again. Thanks!

daquexian:
> @daquexian : It seems the tests were not updated after roi was made optional. Can you update the test script and regenerate the tests again. Thanks!

Thanks. Updated.

askhade:
@daquexian : The tests are checked in with the test script. So you need to update the generated tests as well. I think you only updated the test script. 
FYI: We are cutting the release branch tomorrow morning and as much as possible I want to check this change in otherwise we will need another version bump for resize op.

gramalingam:
It is true that the documentation-generator does not have infrastructure to fix the number of inputs expected. However, the type-and-shape-inference checker _can_ be augmented to verify that exactly one of two inputs is specified. (Well, sort of ... we can do the check using both getNumInputs and getInputType; we will need to use getInputType as a proxy for whether a particular input is specified; it is not perfect, but it is reasonable. The only minor issue would be a misleading error-message in the case where an input is specified, but type-inference could not infer its type.

gramalingam:
Specifically: a utility function like below should help:
```cpp
   inline bool hasInput(InferenceContext& ctx, size_t n) {
       return ctx.getNumInputs() > static_cast<size_t>(n) && ctx.getInputType(n);
}
```
We can then add a check one of the two inputs is specified.

askhade:
Shape inference code for resize needs fixes. https://github.com/onnx/onnx/blob/53057782d19a59548705802fca756d87dee032e2/onnx/defs/tensor/utils.cc#L73
Current shape inference code will pick sizes input only if number of inputs is 4 but with roi and scales being optional inputs shape inference will fail.

daquexian:
> Current shape inference code will pick sizes input only if number of inputs is 4 but with roi and scales being optional inputs shape inference will fail.

I don't think so. 
https://github.com/onnx/onnx/blob/88e5d3e39c11c79a38f43af2a23147895f6f226b/onnx/shape_inference/implementation.h#L59-L65
https://github.com/onnx/onnx/blob/88e5d3e39c11c79a38f43af2a23147895f6f226b/onnx/shape_inference/implementation.h#L87-L89

The `allInputTypes_` will be `[input_type, nullptr, nullptr, sizes_type]` and the number of inputs will still be 4.

daquexian:
> It is true that the documentation-generator does not have infrastructure to fix the number of inputs expected. However, the type-and-shape-inference checker _can_ be augmented to verify that exactly one of two inputs is specified. (Well, sort of ... we can do the check using both getNumInputs and getInputType; we will need to use getInputType as a proxy for whether a particular input is specified; it is not perfect, but it is reasonable. The only minor issue would be a misleading error-message in the case where an input is specified, but type-inference could not infer its type.

Good advice! Let's merge this PR for 1.8 release and I can submit another pr for this in the future.

gramalingam (approved):

askhade (approved):

PR #3035, opened 2020-09-30
Define the differentiable variables for NN operator

rajeevnalawadi:
Pytorch has 3 specific modes for quantization
For training case the “Quantization aware training” QAT is applicable as described in the below linked page
https://pytorch.org/docs/stable/quantization.html#quantization-workflows

Snippet from the page under Quantization workflows & QAT (#3) “…Computations will take place in FP32 but with values clamped and rounded to simulate the effects of INT8 quantization…”

Code for checking QAT
https://pytorch.org/tutorials/advanced/static_quantization_tutorial.html

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -3395,18 +3395,18 @@ Other versions of this operator: <a href="Changelog.md#ConvTranspose-1">1</a>
 #### Inputs (2 - 3)
 
 <dl>
-<dt><tt>X</tt> : T</dt>
+<dt><tt>X</tt> (differentiable) : T</dt>
 <dd>Input data tensor from previous layer; has size (N x C x H x W), where N is the batch size, C is the number of channels, and H and W are the height and width. Note that this is for the 2D image. Otherwise the size is (N x C x D1 x D2 ... x Dn)</dd>
-<dt><tt>W</tt> : T</dt>
+<dt><tt>W</tt> (differentiable) : T</dt>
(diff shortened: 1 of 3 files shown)


onnx/defs/nn/defs.cc
```diff
@@ -1191,10 +1271,6 @@ void convTransposeShapeInference(InferenceContext& ctx) {
     if (pads.size() != n_input_dims * 2) {
       fail_shape_inference("Attribute pads has incorrect size");
     }
-    const auto* auto_pad_attr = ctx.getAttribute("auto_pad");
```
wschin, line 1194: Why is this removed?

rajeevnalawadi, line 1194: Manual add

docs/Operators.md
```diff
@@ -3283,20 +3283,20 @@ This version of the operator has been available since version 10 of the default
 #### Inputs (2 - 4)
 
 <dl>
-<dt><tt>x</tt> : T1</dt>
+<dt><tt>x</tt> (non-differentiable) : T1</dt>
```
wschin, line 3286: Why is it non-differentiable? To speed up the merge, we probably should split this PR into two PRs (one for normal ops and the other one for quantized ops)?

docs/Operators.md
```diff
@@ -12744,30 +12744,30 @@ This version of the operator has been available since version 10 of the default
 #### Inputs (8 - 9)
 
 <dl>
-<dt><tt>x</tt> : T1</dt>
+<dt><tt>x</tt> (differentiable) : T1</dt>
```
wschin, line 12747: This is differentiable but in `ConvInteger` the input `X` isn't.

docs/Operators.md
rajeevnalawadi: Removed the differentiable variables for QlinearConv and ConvInteger operators

docs/Operators.md
rajeevnalawadi: Removed the differentiable variables for QlinearConv and ConvInteger operators

wschin (approved):
Thanks for the hard work!

PR #3036, opened 2020-09-30
Added reserve, move, const, size->empty

Yukigaru:
I've found that the authors of the code base use `reserve` extensively what is great, but the code lacked reserves in some spots, so I've decided to add. There are also a few other minor things to add such as std::move and others.

We can discuss a particular change if you don't agree and I may explain each.

All tests have passed.
All diffs have been reviewed before sending.


Yukigaru:
Okay, I read about DCO and will do it soon

Yukigaru:
DCO done.

onnx/defs/schema.cc
```diff
@@ -96,9 +96,10 @@ void OpSchema::CheckInputOutputType(struct InferenceContext& ctx) const {
     }
     if (param.GetIsHomogeneous()) {
       const auto& type_proto = Utils::DataTypeUtils::ToType(*param_type);
-      if (type_constraints.find(type_str) == type_constraints.end()) {
+      auto it = type_constraints.find(type_str);
```
yuslepukhin, line 99: Assuming we will not find many inconsistent duplicate registrations you can also combine insert/find and eliminate double lookup if you want to be real efficient. Something like:
`auto p = type_constraints.emplace(type_str, *type_proto);
if(!p.second) {
// failed to insert due to a duplicate, check consistency
if(p.first.second != *type_proto) fail()
}
`

onnx/defs/schema.cc
```diff
@@ -400,14 +401,14 @@ OpSchema& OpSchema::NumInputs(std::set<int> allowed_input_nums) {
 OpSchema& OpSchema::NumOutputs(std::set<int> allowed_output_nums) {
   num_outputs_allowed_ =
       [MOVE_CAPTURE_IF_CPP14(allowed_output_nums)](int n) -> bool {
-    return allowed_output_nums.count(n);
+    return allowed_output_nums.find(n) != allowed_output_nums.end();
```
yuslepukhin, line 404: count() usage widespread, probably more readable. Make it count(n) > 0;

onnx/defs/schema.h
```diff
@@ -1060,6 +1061,7 @@ class OpSchemaRegistry final : public ISchemaRegistry {
 
   static const std::vector<OpSchema> get_all_schemas() {
     std::vector<OpSchema> r;
+    r.reserve(4);
```
yuslepukhin, line 1064: We do not want hardcoded reserve(). 
In general, while reserve() usage may be beneficial, it will not make most of the code vastly more efficient. There are two reasons for that 1) the types we often use are primitive types, cheap to copy 2) in addition, the amount of elements is very small, it would fit into a single allocation block if such allocation is necessary at all, because std::vector as well std::string now has some inline space, which eliminates a need for any allocations/re-allocations for many cases.

onnx/defs/shape_inference.h
```diff
@@ -594,6 +594,7 @@ inline void bidirectionalBroadcastShapeInference(
     const TensorShapeProto& shapeR,
     TensorShapeProto& resultShape) {
   std::vector<const TensorShapeProto*> shapes;
+  shapes.reserve(2);
```
yuslepukhin, line 597: Ditto, hardcoded reserve()

onnx/defs/tensor/defs.cc
```diff
@@ -2586,6 +2588,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           propagateElemTypeFromInputToOutput(ctx, 1, 0);
           if (hasNInputShapes(ctx, 3)) {
             std::vector<const TensorShapeProto*> shapes;
+            shapes.reserve(3);
```
yuslepukhin, line 2591: Same thing

onnx/version_converter/adapters/broadcast_forward_compatibility.h
```diff
@@ -28,7 +28,10 @@ class BroadcastForwardCompatibility final : public Adapter {
             n->addInput(inputs[1]);
             std::vector<int64_t> axes;
             std::vector<Dimension> new_sizes = B_sizes;
-            for (int i = 0; i < (int) (A_sizes.size() - B_sizes.size()); i++) {
+            auto size = (int) (A_sizes.size() - B_sizes.size());
```
yuslepukhin, line 31: No C-casts. Also, please, keep sizes in size_t type.
set size to zero if B_sizes.size() is > A_sizes.size().

yuslepukhin (changes requested):

onnx/defs/schema.h
Yukigaru: As I know std::vector implementations does not have SVO/SBO. It's explained here: https://stackoverflow.com/questions/48669413/is-sso-used-in-any-other-standard-library-containers-other-than-stdstring
But I will remove that for the same reason as before: it's hard-coded number.

onnx/common/ir.h
```diff
@@ -1131,6 +1131,7 @@ inline const Graph * Value::owningGraph() const {
 
 inline void Value::replaceAllUsesWith(Value * newValue) {
   ONNX_ASSERT(owningGraph() == newValue->owningGraph());
+  newValue->uses_.reserve(uses().size());
   for(auto u : uses()) {
```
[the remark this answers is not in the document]
Yukigaru: Checked, uses() returns vector of Use objects, Use is a struct of size 16 bytes. I think it can be easily copied.

onnx/defs/schema.cc
Yukigaru: nice trick!

Yukigaru:
@yuslepukhin, thank you for the review! Will update the PR soon.

Yukigaru:
Done, updated.

onnx/defs/schema.h
```diff
@@ -1059,6 +1059,7 @@ class OpSchemaRegistry final : public ISchemaRegistry {
  public:
   static const std::vector<OpSchema> get_all_schemas_with_history() {
     std::vector<OpSchema> r;
+    r.reserve(4);
```
linkerzhang, line 1062: why "4"?

Yamatao: I will remove that reserve. The logic was that 4 is better than 0 (because that incurs 3 reallocs to get to capacity=4), but not high enough to waste memory.

Yamatao:
Done, removed hard-coded .reserve(4). Tests - passed.

onnx/defs/schema.h
Yukigaru: The logic was that 4 is better than 0 (because that incurs 3 reallocs to get to capacity=4), but not high enough to waste memory. But I will remove that diff.

Yukigaru:
Done, removed hard-coded .reserve(4). Tests - passed.

yuslepukhin (approved):

yuslepukhin:
@Yukigaru You will need to to rebase/merge off master.

askhade (approved):

wschin (approved):

Issue #2765, opened 2020-05-08
What's the output shape of gather


mzmssg:
I expected gather works like `numpy.take` because it‘s used to verify gather result.
But in [Gather doc](https://github.com/onnx/onnx/blob/master/docs/Operators.md#gather)
```
  data = [
      [1.0, 1.2, 1.9],
      [2.3, 3.4, 3.9],
      [4.5, 5.7, 5.9],
  ]
  indices = [
      [0, 2],
  ]
  axis = 1,
  output = [
      [
          [1.0, 1.9],
          [2.3, 3.9],
          [4.5, 5.9],
      ],
  ] 
# output_shape (1, 3, 2)
```
Numpy result is
```
import numpy as np

data = [
    [1.0, 1.2, 1.9],
    [2.3, 3.4, 3.9],
    [4.5, 5.7, 5.9],
]
indices = [
    [0, 2],
]
out = np.take(data, indices, axis=1)
print(out.shape) 
# (3, 1, 2)
```
Literally, it seems to be (1, 2, 3) (i_0, i_1 from indices and j_0 from data)
> Let k = indices[i_{0}, ..., i_{q-1}] Then output[i_{0}, ..., i_{q-1}, j_{0}, ..., j_{r-2}] = input[j_{0}, k, j_{1}, ..., j_{r-2}]

So which is the expect output shape?

TMVector:
Hi @mzmssg, thanks for the report. The [code examples](https://github.com/onnx/onnx/blob/master/docs/Operators.md#examples-44) in the specification should be checked against ONNX backends, so they should be correct. The code you quote from the operator description could well be wrong (it looks like it is) -- feel free to raise a PR fixing the description. You would need to edit the description string at [onnx/defs/tensor/defs.cc#L1068](https://github.com/onnx/onnx/blob/master/onnx/defs/tensor/defs.cc#L1068) string and regenerate the docs with [tools/update_doc.sh](https://github.com/onnx/onnx/blob/master/tools/update_doc.sh)

mzmssg:
@TMVector Get you, will fix the doc later

CyanHillFox:
@TMVector @mzmssg The document is still not correct.

daquexian:
@mzmssg @TMVector @CyanHillFox Fixed by #3040

Issue #2346, opened 2019-09-23
Gather example with axis=1 not correct


pulzar:
In the operators spec, the example for Gather op with axis=1 produces an incorrect shape:

[https://github.com/onnx/onnx/blob/master/docs/Operators.md#Gather](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Gather)

```
  data = [
      [1.0, 1.2, 1.9],
      [2.3, 3.4, 3.9],
      [4.5, 5.7, 5.9],
  ]
  indices = [
      [0, 2],
  ]
  axis = 1,
  output = [
      [
          [1.0, 1.9],
          [2.3, 3.9],
          [4.5, 5.9],
      ],
  ]
```

The output shape in this example is (1, 3, 2). However, I believe the correct output is (3, 1, 2):

```
>>> data = [
...       [1.0, 1.2, 1.9],
...       [2.3, 3.4, 3.9],
...       [4.5, 5.7, 5.9],
...   ]
>>> indices = [
...       [0, 2],
...   ]
>>> np.take(data, indices, axis=1)
array([[[1. , 1.9]],

       [[2.3, 3.9]],

       [[4.5, 5.9]]])
>>> np.take(data, indices, axis=1).shape
(3, 1, 2)
```

skottmckay:
Would it be better to simplify a little more as well and just have data with shape {3, 3}, indices with shape {2}, and output with shape {3, 2}?

  data = [
      [1.0, 1.2, 1.9],
      [2.3, 3.4, 3.9],
      [4.5, 5.7, 5.9]
  ]
  indices = [0, 2]
  axis = 1,
  output = [
      [1.0, 1.9],
      [2.3, 3.9],
      [4.5, 5.9]
  ]

daquexian:
Fixed by #3040

Issue #1708, opened 2018-12-21
example in gather op spec is wrong


HectorSVC:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Gather
Example 2: data = [ [1.0, 1.2, 1.9], [2.3, 3.4, 3.9], [4.5, 5.7, 5.9], ] indices = [ [0, 2], ] axis = 1, output = [ [ [1.0, 1.9], [2.3, 3.9], [4.5, 5.9], ], ]
for this example, the output shape should be [3, 1, 2] instead of [1, 3, 2], correct output = [ [ [1.0, 1.9]], [[2.3, 3.9]], [[4.5, 5.9], ], ]

daquexian:
Fixed by #3040

PR #3040, opened 2020-10-03
fix the incorrect gather op example and add test

daquexian:
Related issue: https://github.com/onnx/onnx/issues/2346, https://github.com/onnx/onnx/issues/2765, https://github.com/onnx/onnx/issues/1708

This PR fixes the incorrect example in gather op docs and add a test for 2-d indices


askhade (approved):
LGTM

wschin (approved):

PR #3041, opened 2020-10-05
Improve sparse tensor support when used as sparse initializers

yuslepukhin:
Clearly specify in proto files that values name is considered a SparseTensor name.
Enforce values name is present and not empty in checker.
Enforce that the sparse tensor initializers name is unique across initializers and sparse initializers.
Add python tests.

### DIFF
--- a/onnx/checker.cc
+++ b/onnx/checker.cc
@@ -234,18 +234,16 @@ void check_sequence(const SequenceProto& sequence, const CheckerContext& ctx) {
     for (const TensorProto& tensor : sequence.tensor_values()) {
       check_tensor(tensor, ctx);
     }
-  }
-  else if (sequence.elem_type() == SequenceProto::SPARSE_TENSOR) {
-    for (const SparseTensorProto& sparse_tensor : sequence.sparse_tensor_values()) {
+  } else if (sequence.elem_type() == SequenceProto::SPARSE_TENSOR) {
+    for (const SparseTensorProto& sparse_tensor :
+         sequence.sparse_tensor_values()) {
       check_sparse_tensor(sparse_tensor, ctx);
     }
-  }
-  else if (sequence.elem_type() == SequenceProto::SEQUENCE) {
+  } else if (sequence.elem_type() == SequenceProto::SEQUENCE) {
     for (const SequenceProto& seq : sequence.sequence_values()) {
       check_sequence(seq, ctx);
     }
-  }
-  else if (sequence.elem_type() == SequenceProto::MAP) {
+  } else if (sequence.elem_type() == SequenceProto::MAP) {
     for (const MapProto& map : sequence.map_values()) {
       check_map(map, ctx);
     }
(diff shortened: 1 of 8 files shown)


onnx/checker.cc
```diff
@@ -424,6 +419,12 @@ void check_sparse_tensor(
 
   const TensorProto& values = sparse_tensor_proto.values();
   check_tensor(values, ctx);
+  // SparseTensors make use of values name
+  // and this is important for use as inputs or sparse initializers
+  enforce_has_field(values, name);
+  if (values.name().empty()) {
```
gramalingam, line 425: One question: TensorProtos and SparseTensorProtos can appear in two contexts. (a) In initializers, and (b) As an attribute-value in a node. Is the name compulsorily needed in both contexts or only when they are used in initializers? This is more a question about how to interpret the existing standard/spec. My interpretation is that the name is required in case (a), not necessarily in case (b). If so, I think we can have this check when checking initializers, instead of at this place. In either case, I think we should have the same requirement/check for dense initializers as well as sparse initializers.

gramalingam: I agree that the spec is not very explicit about this. I am inferring it indirectly, since onnx.proto specifies "field MUST be present" if it is compulsory for all instances. Further, AttributeProto has its own name.

yuslepukhin: If there are no other opinions, I will make the name optional for sparse tensor but enforce it for sparse initializers.

askhade: I think that is best. I also inferred the spec the similar way. It seems like the name is required for initializers and not in the case of attribute value

onnx/onnx-ml.proto
```diff
@@ -423,8 +423,8 @@ message GraphProto {
   optional string name = 2;   // namespace Graph
 
   // A list of named tensor values, used to specify constant inputs of the graph.
-  // Each TensorProto entry must have a distinct name (within the list) that
-  // MAY also appear in the input list.
+  // Each TensorProto entry must have a distinct name (between initializer and sparse_initializer)
```
gramalingam, line 426: change to "Each TensorProto and SparseTensorProto entry"?

gramalingam (approved):

onnx/checker.cc
```diff
@@ -663,27 +658,48 @@ void check_graph(
...
     }
-    check_tensor(init, ctx);
   }
 
-  // TODO: Need to check that sparse-initializers names are distinct from
-  // initializer names. It looks like the existing checker does not check for
-  // certain duplication of names: e.g., two entries in the initializer list
-  // with same name. Will add a new integrated check.
   for (const auto& sparse_init : graph.sparse_initializer()) {
+    const auto& name = sparse_init.values().name();
```
askhade, line 692: no need for enforce_has_field(sparse_init, name); here?

onnx/test/checker_test.py
```diff
@@ -89,6 +89,24 @@ def test_check_graph(self):  # type: () -> None
         graph.initializer[0].name = 'X'
         checker.check_graph(graph)
 
+    def test_check_graph_duplicate_init_names(self):  # type: () -> None
```
askhade, line 92: can you add one more tests to test the enforce_has_field behavior for attribute and initializer. This will make the intent explicit.

askhade:
Linux CI is failing due to type check failure
onnx/test/checker_test.py:306: error: Incompatible default for argument "name" (default has type "None", argument has type "str")
type check failed

askhade (approved):

PR #3042, opened 2020-10-06
fix unused param warnings

askhade:
Fixes unused param warnings caused when "__ONNX_NO_DOC_STRINGS" is defined.


gramalingam:
On an unrelated point: can we increase the line-width used in clang-format? I think we would have more readable code then!

gramalingam (approved):

askhade:
> On an unrelated point: can we increase the line-width used in clang-format? I think we would have more readable code then!

I updated it to 120. (ORT also used 120)

PR #3059, opened 2020-10-15
Advance NLL loss to opset 13

ashbhandare:


ebarsoum (approved):

gramalingam:
It would be great if you could add a couple of lines explaining why the change is required. I understand an op used in the body has changed: which op? Does it affect the semantics of this function?

gramalingam (approved):

askhade (approved):

PR #3060, opened 2020-10-16
[WIP] Bump to Opset 13: DynamicQuantizeLinear

jcwchen:
**Description**
Bump DynamicQuantizeLinear to opset 13.
In addition, update related documents and tests.

**Motivation**
Since QuantizeLinear has changed in opset 13 and DynamicQuantizeLinear does use it, it's better to bump DynamicQuantizeLinear to opset 13 as well


jcwchen:
cc @yufenglee

onnx/defs/quantization/defs.cc
```diff
@@ -145,9 +145,9 @@ y = saturate (round (x / y_scale) + y_zero_point)
...
 ONNX_OPERATOR_SET_SCHEMA(
     DynamicQuantizeLinear,
-    11,
+    13,
     OpSchema()
-        .SetDoc(DynamicQuantizeLinear_ver11_doc)
+        .SetDoc(DynamicQuantizeLinear_ver13_doc)
         .Input(0, "x", "Input tensor", "T1")
         .Output(0, "y", "Quantized output tensor", "T2")
         .Output(1, "y_scale", "Output scale. It's a scalar, which means a per-tensor/layer quantization.", "tensor(float)")
```
yufenglee, line 153: We'd better to make DynamicQuantizeLinear to support per-channel too.

askhade:
please remove changes to all the tests... only include tests related to the op\function you are changing

onnx/defs/quantization/defs.cc
yufenglee: It looks hard to implement with FunctionBody because there is no way to generate axes attribute for ReduceMin correctly.

jcwchen:
> please remove changes to all the tests... only include tests related to the op\function you are changing

Just manually removed. I was thinking whether these is a good way to detect those mattered models automatically but failed... I suppose `cmd_tools.py` should be corrected to only update the mattered models.

jcwchen:
After discussion, we decide to extend reducemax and reducemin by attribute "complement_axis" to add axis attribute for dynamicQuantLinear:
1. Update ReduceMax and ReduceMin with an optional boolean attribute, complement_axis (default is False).
2. Go back to DynamicQuantizeLinear and add an optional axis attribute just like QuantizeLinear
3. Update FunctionBody of DynamicQuantizeLinear to use ReduceMax and ReduceMin with complement_axis=True and target axis (added attribute from 2.)

onnx/defs/reduction/defs.cc
```diff
@@ -221,7 +221,8 @@ If keepdims equal 0, then the resulting tensor have the reduced dimension pruned
 If select_last_index is True (default False), the index of the last occurrence of the {name} 
 is selected if the {name} appears more than once in the input. Otherwise the index of the 
 first occurrence is selected.
-The type of the output tensor is integer.)DOC";
+The type of the output tensor is integer.
```
askhade, line 224: This the wrong method. You are adding this to ArgReduceDocGenerator. It should be added to ReduceDocGenerator.
Also all reduction ops use ReduceDocGenerator so you need to add a conditional whether complementary axis should be added or not

onnx/defs/reduction/defs.cc
```diff
@@ -221,7 +221,8 @@ If keepdims equal 0, then the resulting tensor have the reduced dimension pruned
...
+complement_axis needs be used simultaneously with axis attribute. If complement_axis is True (default False), it will reduce other axes except the target axis (complementarily).)DOC";
```
gramalingam: Suggested rewording: "The attribute complement_axis is used in conjunction with the axis attribute, and has a default value of False. If complement_axis is True, all axes except the specified axis will be reduced."

onnx/defs/reduction/defs.cc
```diff
@@ -239,6 +240,11 @@ The type of the output tensor is integer.)DOC";
         "Whether to select the last index or the first index if the {name} appears in multiple indices, default is False (first index).",
         AttributeProto::INT,
         static_cast<int64_t>(0));
+    schema.Attr(
+        "complement_axis",
+        "Whether to reduce axes complementarily, default is False (reduce target axis). This attribute needs be used simultaneously with axis attribute.",
```
gramalingam, line 273: Nit: Suggest changing "needs to be used simultaneously" to "is used in conjunction with". This is just to avoid creating the impression that the attribute must be specified if axis is specified.

onnx/defs/quantization/defs.cc
```diff
@@ -164,9 +171,9 @@ ONNX_OPERATOR_SET_SCHEMA(
           {// nodes: {outputs, op, inputs, attributes}
            FunctionBodyHelper::Const<float>("Q_Min", 0.f),
            FunctionBodyHelper::Const<float>("Q_Max", 255.f),
-           {{"X_Min"}, "ReduceMin", {"x"}, {MakeAttribute("keepdims", int64_t(0))}},
+           {{"X_Min"}, "ReduceMin", {"x"}, {MakeAttribute("keepdims", int64_t(0)), MakeRefAttribute("axes", "axis", AttributeProto::INTS), MakeAttribute("complement_axis", int64_t(1))}},
```
jcwchen, line 174: The attribute which DynamicQuantizeLinear is using was AttributeProto::INT, but the attribute which Reduce is using is AttributeProto::INTS... In that case, it seems that we cannot use MakeRefAttribute directly here and the conversion from INT to INTS is needed. Does current FunctionBodyHelper support it?

jcwchen: If not, I am thinking that maybe we can introduce other attribute (single integer) to represent the specified single quant dimension in reduce function instead of "complement_axis" (boolean).

onnx/defs/reduction/defs.cc
jcwchen: Updated. Thanks

onnx/defs/reduction/defs.cc
jcwchen: All updated. Thanks.

onnx/defs/quantization/defs.cc
gramalingam: If we use a ContextDependentFunctionBody, we can do it. See https://github.com/onnx/onnx/blob/a7a0fec7f25cae567429af62b7eaaee1c3f0e247/onnx/defs/math/defs.cc#L2830 for an example. The context allows us to get the attribute-value, and use it to construct new attribute-values.

jcwchen: @gramalingam I have just sent an email and haven't seen you replied here... Thank you so much and I will give this a try.

askhade: Add attribute to QuantizeLinear node as well :)

onnx/defs/quantization/defs.cc
```diff
@@ -143,15 +143,22 @@ y = saturate (round (x / y_scale) + y_zero_point)
...
+    13,
     OpSchema()
-        .SetDoc(DynamicQuantizeLinear_ver11_doc)
+        .SetDoc(DynamicQuantizeLinear_ver13_doc)
         .Input(0, "x", "Input tensor", "T1")
         .Output(0, "y", "Quantized output tensor", "T2")
         .Output(1, "y_scale", "Output scale. It's a scalar, which means a per-tensor/layer quantization.", "tensor(float)")
         .Output(2, "y_zero_point", "Output zero point. It's a scalar, which means a per-tensor/layer quantization.", "T2")
+        .Attr(
+            "axes",
```
askhade, line 158: this should be axis and not axes. In the context of this function axis should be singular.

onnx/defs/quantization/defs.cc
```diff
@@ -143,15 +143,22 @@ y = saturate (round (x / y_scale) + y_zero_point)
...
+            "(Optional) The axis of the quantization dimension of the input tensor. Negative value means counting dimensions from the back. Accepted range is [-r, r-1] where r = rank(input)",
+            AttributeProto::INTS,
```
askhade, line 160: same comment as above. axis here should be same as axis in QuantizeLinear

onnx/defs/reduction/defs.cc
```diff
@@ -33,7 +34,10 @@ tensor has the same rank as the input if keepdims equal 1. If keepdims equal 0,
 the resulted tensor have the reduced dimension pruned.
 
 The above behavior is similar to numpy, with the exception that numpy default keepdims to
-False instead of True.)DOC";
+False instead of True.
+
+The attribute complement_axis is used in conjunction with the axis attribute, and has a default value of False.
+If complement_axis is True, all axes except the specified axis will be reduced.)DOC";
```
askhade, line 40: Doc for all reduction ops will be updated. complement_axes related documentaiton should only be added for ReduceMax and ReduceMin

onnx/defs/reduction/defs.cc
```diff
@@ -33,7 +34,10 @@ tensor has the same rank as the input if keepdims equal 1. If keepdims equal 0,
 the resulted tensor have the reduced dimension pruned.
 
 The above behavior is similar to numpy, with the exception that numpy default keepdims to
-False instead of True.)DOC";
+False instead of True.
+
+The attribute complement_axis is used in conjunction with the axis attribute, and has a default value of False.
```
askhade, line 39: I suggest rewording this a little bit : 

The above behavior is similar to numpy, with the following exceptions
1. numpy defaults keepdims to False instead of True
2. This op uses complement _axes attribute which defaults to false but when set to true indicates all axes except the specified axes will be reduced.
The attribute complement_axis is used in conjunction with the axis attribute, and has a default value of False.

onnx/defs/reduction/defs.cc
```diff
@@ -78,6 +82,15 @@ False instead of True.)DOC";
           AttributeProto::INTS,
           OPTIONAL_VALUE);
     }
+    if (complement_axis) {
+      schema.Attr(
+          "complement_axis",
+          "Whether to reduce axes complementarily, "
```
askhade, line 88: Whether to use complement of axes for reduction

onnx/defs/reduction/defs.cc
```diff
@@ -78,6 +82,15 @@ False instead of True.)DOC";
           AttributeProto::INTS,
           OPTIONAL_VALUE);
     }
+    if (complement_axis) {
+      schema.Attr(
+          "complement_axis",
```
askhade, line 87: rename as complement_axes?

onnx/defs/reduction/defs.cc
```diff
@@ -78,6 +82,15 @@ False instead of True.)DOC";
...
+          "Whether to reduce axes complementarily, "
+          "default is False (reduce target axis). "
+          "This attribute is used in conjunction with axis attribute.",
```
askhade, line 90: in conjunction with "axes" attribute. In context of reduction ops axes is plural so use the axes instead of axis

onnx/defs/quantization/defs.cc
jcwchen: Using axes here was just for experiment. Updated it as axis

onnx/defs/quantization/defs.cc
jcwchen: What do you mean by adding attribute to QuantizeLinear node?

onnx/defs/quantization/defs.cc
```diff
@@ -145,38 +145,57 @@ y = saturate (round (x / y_scale) + y_zero_point)
...
+                {{"X_Range"}, "Sub", {"X_Max_Adjusted", "X_Min_Adjusted"}},
+                {{"Scale"}, "Div", {"X_Range", "Q_Max"}},
+                {{"Min_Scaled"}, "Div", {"X_Min_Adjusted", "Scale"}},
+                {{"Initial_ZeroPoint_FP"}, "Sub", {"Q_Min", "Min_Scaled"}},
+                {{"Clipped_ZeroPoint_FP"}, "Clip", {"Initial_ZeroPoint_FP", "Q_Min", "Q_Max"}},
+                {{"Rounded_ZeroPoint_FP"}, "Round", {"Clipped_ZeroPoint_FP"}},
+                {{"Zeropoint"}, "Cast", {"Rounded_ZeroPoint_FP"}, {MakeAttribute("to", int64_t(2))}},
+                {{"y_scale"}, "Identity", {"Scale"}},
+                {{"y_zero_point"}, "Identity", {"Zeropoint"}},
+                {{"y"}, "QuantizeLinear", {"x", "Scale", "Zeropoint"}}});
```
askhade, line 191: axis should also be passed to QuantizeLinear

onnx/defs/quantization/defs.cc
```diff
@@ -145,38 +145,57 @@ y = saturate (round (x / y_scale) + y_zero_point)
...
-           {{"Rounded_ZeroPoint_FP"}, "Round", {"Clipped_ZeroPoint_FP"}},
-           {{"Zeropoint"}, "Cast", {"Rounded_ZeroPoint_FP"}, {MakeAttribute("to", int64_t(2))}},
-           {{"y_scale"}, "Identity", {"Scale"}},
-           {{"y_zero_point"}, "Identity", {"Zeropoint"}},
-           {{"y"}, "QuantizeLinear", {"x", "Scale", "Zeropoint"}}})));
+        .SetContextDependentFunctionBodyBuilder(
+            [](const FunctionBodyBuildContext& ctx,
+                const OpSchema& schema,
+                FunctionProto& functionProto) -> bool {
+                const auto axis = ctx.getAttribute("axis") != nullptr ? ctx.getAttribute("axis")->i() : OPTIONAL_VALUE;
```
askhade, line 172: nit: DEFAULT_VALUE

onnx/defs/reduction/defs.cc
```diff
@@ -21,20 +21,34 @@ std::vector<std::string> GetSupportedDataTypesForReductionOps(
...
+inline std::string GenerateGeneralRedunctionDoc(std::string name) {
+  return "Computes the " + name + " of the input tensor's element along the provided axes. The resulted"
+"tensor has the same rank as the input if keepdims equal 1. If keepdims equal 0, then"
+"the resulted tensor have the reduced dimension pruned.\n"
+"The above behavior is similar to numpy, with the exception that numpy default keepdims to"
+"False instead of True.";
+}
+
+inline std::string GenerateComplementAxesDoc() {
+  return "\nThe above behavior is similar to numpy, with the following exceptions:"
```
askhade, line 33: This line and the next is present in GenerateGeneralRedunctionDoc as well. I suggest keep it simple just keep 2 separate strings no need for concatenation

jcwchen:
Thank @askhade for all the helpful comments. I have just updated. Right now I am working on fixing the original test, `test_dynamicquantizelinear_min_adjusted_expanded_cpu`, which is failing because `RuntimeError: Inferred shape and existing shape differ in rank: (1) vs (0)`.

onnx/defs/reduction/defs.cc
```diff
@@ -21,20 +21,34 @@ std::vector<std::string> GetSupportedDataTypesForReductionOps(
   return OpSchema::numeric_types_for_math_reduction_with_bfloat();
 }
 
+inline std::string GenerateGeneralRedunctionDoc(std::string name) {
```
askhade, line 24: this does not have to be a function. can be const char* reduction_doc

onnx/defs/quantization/defs.cc
jcwchen: OPTIONAL_VALUE is a global variable (`false`) which is the default value of "axes" attribute in Reduction functions. Actually I don't know what the proper default value here is if the "axis" attribute does not specify in `DynamicQuantizeLinear`. I believe this is why `test_dynamicquantizelinear_min_adjusted_expanded_cpu` is failing.

onnx/defs/reduction/defs.cc
```diff
@@ -21,20 +21,39 @@ std::vector<std::string> GetSupportedDataTypesForReductionOps(
...
-False instead of True.)DOC";
-                        ReplaceAll(doc, "{name}", name););
+    std::string whole_reduction_doc = GenerateGeneralRedunctionDoc(name);
+    if (complement_axes) {
+      whole_reduction_doc += GenerateComplementAxesDoc();
+    } else {
+      whole_reduction_doc += GenerateOnlyKeepdimsDoc();
+    }
+    POPULATE_OP_DOC_STR(doc = R"DOC({whole_reduction_doc})DOC";
+                        ReplaceAll(doc, "{whole_reduction_doc}", whole_reduction_doc.c_str()););
```
askhade, line 56: why do you need replace all? initially it was used to replace all occurrences of name with the op name...

docs/Changelog.md
```diff
@@ -17248,12 +17305,9 @@ This version of the operator has been available since version 13 of the default
 
 ### <a name="ReduceLogSum-13"></a>**ReduceLogSum-13**</a>
 
-  Computes the log sum of the input tensor's element along the provided axes. The resulted
-  tensor has the same rank as the input if keepdims equal 1. If keepdims equal 0, then
-  the resulted tensor have the reduced dimension pruned.
+  Computes the log sum of the input tensor's element along the provided axes. The resulted tensor has the same rank as the input if keepdims equal 1. If keepdims equal 0, then the resulted tensor have the reduced dimension pruned.
```
askhade, line 17308: regenerate docs

onnx/defs/quantization/defs.cc
gramalingam: The quantize op spec says that the default value is 1, doesn't it?

gramalingam: We should be using the default (1) for DynamicQuantizeLinear (not from Reduction ops).

jcwchen: If that is the case, it seems that it will break 3 existing test cases... I want to make sure one thing:
1. (Original) ReduceMax and ReduceMin (keepdims=0, axes is not specified; which means reduce all dimensions)
2. (Updated) ReduceMax and ReduceMin (axes=1, keepdims=0, complement_axes=1)

Will these default cases, 1 and 2, produce the same shape?

yufenglee:
> <dd>Output scale. It's a scalar, which means a per-tensor/layer quantization.</dd>

It can be either a scalar for per-tensor or a 1D dimension tensor with same size as the axis specified for per-channel, 

---
Refers to: docs/Operators.md:4818 in 223343b. [](commit_id = 223343ba0cae5086c39974ce6d8d7c0318121bdd, deletion_comment = False)

yufenglee:
> <dd>Output zero point. It's a scalar, which means a per-tensor/layer quantization.</dd>

Similar as y_scale

---
Refers to: docs/Operators.md:4820 in 223343b. [](commit_id = 223343ba0cae5086c39974ce6d8d7c0318121bdd, deletion_comment = False)

onnx/defs/quantization/defs.cc
```diff
@@ -145,38 +145,57 @@ y = saturate (round (x / y_scale) + y_zero_point)
...
     DynamicQuantizeLinear,
-    11,
+    13,
     OpSchema()
-        .SetDoc(DynamicQuantizeLinear_ver11_doc)
+        .SetDoc(DynamicQuantizeLinear_ver13_doc)
         .Input(0, "x", "Input tensor", "T1")
         .Output(0, "y", "Quantized output tensor", "T2")
         .Output(1, "y_scale", "Output scale. It's a scalar, which means a per-tensor/layer quantization.", "tensor(float)")
         .Output(2, "y_zero_point", "Output zero point. It's a scalar, which means a per-tensor/layer quantization.", "T2")
```
yufenglee, line 154: Please update to reflect that they can be a tensor for per-channel quantization.

onnx/defs/quantization/defs.cc
```diff
@@ -145,38 +145,57 @@ y = saturate (round (x / y_scale) + y_zero_point)
...
+        .Attr(
+            "axis",
+            "(Optional) The axis of the quantization dimension of the input tensor. Negative value means counting dimensions from the back. Accepted range is [-r, r-1] where r = rank(input)",
+            AttributeProto::INT,
+            static_cast<int64_t>(1))
```
yufenglee, line 159: There is an issue here. In QuantizeLinear, we use the size of scale to determine if it is per-tensor or per-channel, i.e., if size of scale is 1, then it is per-tensor; if size of scale is same the x.shape[axis], then it is a per-channel. Here we can not just use the axis to differentiate it.

jcwchen:
According to @yufenglee 's feedback, this PR needs more discussion about the per-channel thing (will introduce more attribute to indicate the usage). We won't forward this before ONNX 1.8. Close this PR now.

PR #3064, opened 2020-10-20
Add type context to FunctionBodyBuildContext

gramalingam:
The current context-dependent FunctionBodyBuilder can make use of the context-information (such as attribute values and the availability of optional inputs/outputs) in constructing the function-body. However, the context does not currently contain information about the types of the inputs, which is somewhat restrictive. For example, the function-body construction for the NegativeLogLikelihoodLoss has examples illustrating this. This line https://github.com/onnx/onnx/blob/ff047ef41d5d91ac1aaa8ebe85ed449c8f0f3604/onnx/defs/math/defs.cc#L2912 for example, tries to construct a zero-value of the same type as some input in a roundabout fashion.

This PR adds the needed type-context information.


onnx/defs/schema.h
```diff
@@ -29,12 +29,17 @@ struct FunctionBodyBuildContext {
   virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
   virtual bool hasInput(int i) const = 0;
   virtual bool hasOutput(int i) const = 0;
+  // getInputType(i) should return null for missing optional inputs, or if
+  // type-inference could not infer the input-type (erroneous model).
+  virtual const TypeProto* getInputType(int i) const = 0;
```
askhade, line 34: rename i to say input_index for better readability

onnx/backend/test/case/node/__init__.py
```diff
@@ -156,7 +156,16 @@ def expect(node,  # type: onnx.NodeProto
         atol=1e-7,
     ))
 
-    expanded_function_nodes = function_testcase_helper(node, name)
+    # Create list of types for node.input, filling a default TypeProto for missing inputs:
+    def merge(node_inputs, present_value_info):  # type: (List[Text], List[onnx.ValueInfoProto]) -> List[TypeProto]
```
askhade, line 160: I dont understand what this is exactly trying to do. Can you please elaborate... 

For example when node_inputs does not exist why return [] and not [TypeProto()] 
Also if (node_inputs[0] != ''): what does this check really accomplish?

askhade:
Can you add a test case for this too.

onnx/backend/test/case/node/__init__.py
gramalingam: I added comments to the beginning of the function to clarify what the inputs/outputs look like. This is needed because of the existing style of creating test-cases which omit values for optional-inputs when they are missing. So, we need to adjust for missing-optional-parameters when creating the list of input-types.

onnx/defs/schema.h
gramalingam: I felt it would be simplest to maintain compatibility with the existing methods like hasInput/hasOutput. If we change it, we should do it for all simultaneously. (A related note: if we do change it, it would be more useful to change the type to unsigned as well; but then it starts becoming a more significant change ... something to think about.) Any thoughts on what exactly we should do?

askhade: I did not realize other methods used the same convention. For this PR I suggest updating this and other methods to use a readable name instead of i ... and we can update the type to uint as part of another PR.

gramalingam: Done

askhade (approved):

onnx/defs/schema.h
```diff
@@ -27,15 +27,20 @@ namespace ONNX_NAMESPACE {
 
 struct FunctionBodyBuildContext {
   virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
-  virtual bool hasInput(int i) const = 0;
-  virtual bool hasOutput(int i) const = 0;
+  virtual bool hasInput(int inputIndex) const = 0;
+  virtual bool hasOutput(int inputIndex) const = 0;
+  // getInputType(i) should return null for missing optional inputs, or if
+  // type-inference could not infer the input-type (erroneous model).
+  virtual const TypeProto* getInputType(int inputIndex) const = 0;
```
wschin, line 34: Maybe we should throw if a type is missing? We can write another helper function `virtual bool hasInputType(int inputIndex)` for checking if input type exists before calling `getInputType(int inputIndex)`.

onnx/cpp2py_export.cc
```diff
@@ -64,19 +64,27 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
-        if (op->HasContextDependentFunction()) {
-          FunctionBodyBuildContextImpl ctx(proto);
-          FunctionProto func_proto;
-          op->BuildContextDependentFunction(ctx, func_proto);
-          func_proto.SerializeToString(&func_bytes);
-        }
-        return py::bytes(func_bytes);
-      });
+      .def(
+          "get_context_dependent_function",
```
wschin, line 68: I feel it's better to have a test for this function.

wschin (approved):

onnx/defs/schema.h
gramalingam: I think this is simpler than throwing an exception. We already have optional inputs to handle anyway. It is good to make the caller be aware and handle the possibility explicitly. (It is easier to ignore/forget to handle exceptions.)

onnx/cpp2py_export.cc
gramalingam: Agree with test-case. But I don't think we can do it from python since we need to define a new dummy function for this purpose, which needs to be done in C++. Let me look into this separately because some test-infrastructure is required I think.

PR #3068, opened 2020-10-22
Both update value_info and output for shape_inference undefined output

jcwchen:
**Description**
- Both update value_info and output for output with undefined type 
- Throw `ONNX_NAMESPACE::InferenceError` instead of `std::runtime_error`, which is more proper for debugging

**Motivation**
shape_inference used to update “undefined” output to value_info, but after this PR (https://github.com/onnx/onnx/pull/2783), if the type of output is undefined, shape_inference will directly merge inferred type into the output instead of creating a new value_info with the output’s name. It should be fine for valid models, but it will influence the behavior of invalid models (like missing output type initially).

However, part of `torch.onnx.export` replies on `onnx.shape_inference` and it only considers value_info. As a workaround, shape_inference now both updates value_info and output for output without defined type.
Please note that in the future, shape_inference will only update output in this case.

### DIFF
--- a/onnx/shape_inference/implementation.cc
+++ b/onnx/shape_inference/implementation.cc
@@ -321,9 +321,16 @@ static void InferShapesImpl(
         if (iter != valueTypesByName.end()) {
           existingType = iter->second;
         } else {
+          // Create a new value_info if defined type does not exist
           auto vi = g->add_value_info();
           vi->set_name(n.output(i));
           existingType = vi->mutable_type();
+          // For undefined output type, update both value_info and output for now
+          // Update existing output with undefined type: assign inferred type to it
+          iter = undefinedValueTypesByName.find(n.output(i));
+          if (iter != undefinedValueTypesByName.end()) {
+            *iter->second = *inferredType;
+          }
         }
 
         // Now we can merge pre-existing and inferred info
@@ -331,13 +338,6 @@ static void InferShapesImpl(
 
         // Make merged info available to further inference.
         valueTypesByName[n.output(i)] = existingType;
-
-        // For udefined output type, update both value_info and output for now
-        // If the output type is undefined, merge inferred type to it
-        iter = undefinedValueTypesByName.find(n.output(i));
-        if (iter != undefinedValueTypesByName.end()) {
-          mergeShapesAndTypes(*inferredType, iter->second);
-        }
       }
     } catch (const std::runtime_error& err) {
       std::cerr << getErrorWithNodeInfo(n, err) << std::endl;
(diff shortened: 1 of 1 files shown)


onnx/shape_inference/implementation.cc
```diff
@@ -322,6 +331,13 @@ static void InferShapesImpl(
 
         // Make merged info available to further inference.
         valueTypesByName[n.output(i)] = existingType;
+
+        // For udefined output type, update both value_info and output for now
+        // If the output type is undefined, merge inferred type to it
+        iter = undefinedValueTypesByName.find(n.output(i));
+        if (iter != undefinedValueTypesByName.end()) {
+          mergeShapesAndTypes(*inferredType, iter->second);
```
gramalingam, line 339: A few independent comments:
(a) Can we just do an assignment instead of calling mergeShapesAndTypes? "* iter->second = *inferredType"?
(b) Can we move this into the else-branch in line 323?

askhade: agree with Rama on both the points...

jcwchen: Both (a) and (b) are doable.
For (b), I think `inferredType` is actually determined on line 311 and this update (update original undefined output) only occurs in the else-branch on line 323. So it looks good to move there.

gramalingam (approved):

onnx/shape_inference/implementation.cc
```diff
@@ -166,23 +166,32 @@ static void InferShapesImpl(
...
+    if (vi.has_type()) {
       valueTypesByName[vi.name()] = vi.mutable_type();
+    }
   }
   for (auto& vi : *g->mutable_output()) {
-    // Some output type might be undefined
-    // To assgin inferred type to them,
-    // Also save names of output with undefined types
-    valueTypesByName[vi.name()] = vi.mutable_type();
+    // TODO: Remove this condition and undefinedValueTypesByName in the future
```
askhade, line 186: either remove this comment or explain why this condition and undefinedValueTypesByName  needs to be removed. My suggestion remove this comment from here and create an issue to track this item. The first step is to add the clarification in the spec and then update the code here.

onnx/shape_inference/implementation.cc
```diff
@@ -166,23 +166,32 @@ static void InferShapesImpl(
...
+    if (vi.has_type()) {
+      valueTypesByName[vi.name()] = vi.mutable_type();
+    } else {
+      // Some output type might be undefined in subgraph. e.g., Loop Op
+      // To assgin inferred type to them,
```
askhade, line 191: nit: line 191 and 192 can be reworded as Saving names of outputs with undefined types to allow assigning inferred types to them.

onnx/shape_inference/implementation.cc
jcwchen: Sounds good. I have removed it and tracked here: https://github.com/onnx/onnx/issues/3069

askhade (approved):

BowenBao (approved):
Looks great, thank you for updating this!

askhade (approved):

PR #3071, opened 2020-10-25
Fix next unique

daquexian:
This PR is based on @okdshin's work https://github.com/onnx/onnx/pull/2906. It looks like the original author @okdshin has no enough time to maintain his PR, but his PR will help onnx optimizer a lot. So I updated it myself, including the check of subgraph and some code style refinement.


daquexian:
@okdshin I signed off your commits to make this PR mergeable. Please contact me if you want to sign it off yourself and this PR is not merged.

okdshin:
No problem. Thank you.

onnx/common/ir.h
```diff
@@ -866,6 +866,50 @@ friend struct Value;
 
   std::vector <OpSetID> opset_versions_;
 
+  bool isExistingName(const std::string& name) const {
+    if (std::find(initializer_names_.begin(), initializer_names_.end(), name) !=
```
askhade, line 870: where applicable use const iterators cbegin() and cend()

onnx/common/ir.h
```diff
@@ -866,6 +866,50 @@ friend struct Value;
 
   std::vector <OpSetID> opset_versions_;
 
+  bool isExistingName(const std::string& name) const {
```
askhade, line 869: nit: isExistingName -> doesNameExist

even better would be isNameUnique

askhade (changes requested):
Added a few comments. Overall looks good to me

daquexian:
> Added a few comments. Overall looks good to me

Thanks! Updated

onnx/common/ir.h
```diff
@@ -866,6 +866,50 @@ friend struct Value;
...
+        } else if (node->kindOf(attr) == AttributeKind::gs) {
+          for (const auto& subgraph : node->gs(attr)) {
+            if (subgraph->isNameUnique(name)) {
+              return true;
+            }
+          }
+        }
+      }
+      const auto found_in =
+          std::find_if(node->inputs().begin(), node->inputs().end(), f);
```
askhade, line 891: const iterators can be used here too and on line line 896

onnx/common/ir.h
```diff
@@ -866,6 +866,50 @@ friend struct Value;
 
   std::vector <OpSetID> opset_versions_;
 
+  bool isNameUnique(const std::string& name) const {
```
askhade, line 869: isNameUnique == true logically suggests that the function is going to return true if name is unique and false otherwise but the return value of this function is opposite of this and is confusing.

onnx/common/ir.h
```diff
@@ -866,6 +866,50 @@ friend struct Value;
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+
+  size_t getNextUnique() {
+      std::string next_unique_name = ONNX_NAMESPACE::to_string(++next_unique_);
+      while(isNameUnique(next_unique_name)) {
```
askhade, line 907: isNameUnique should return false when name is not unique and this check should be changed to while(!isNameUnique(next_unique_name))

onnx/common/ir.h
daquexian: node->inputs/outputs() returns an ArrayRef object which doesn't have a cbegin() or cend() method

onnx/common/ir.h
daquexian: my bad 😂

askhade (approved):

PR #3077, opened 2020-10-28
Fix cpp headers and libraries installation when using setup.py

AnthonyBarbier:
- Some of the generated protobuf headers were missing
- None of the shared libraries were packaged
- The CMake targets were missing.

The aim of this patch is to fix the Conda package so that it can be used to build user applications.
Unfortunately I don't know much about Conda so I haven't been able to generate a new package.


setup.py
```diff
@@ -210,12 +211,21 @@ def run(self):
...
+
 
 class build_py(setuptools.command.build_py.build_py):
     def run(self):
         self.run_command('create_version')
         self.run_command('cmake_build')
 
+        self.copy_tree(os.path.join(CMAKE_BUILD_DIR,"install","include"),
+                os.path.join(TOP_DIR, self.build_lib, "onnx", "include"))
+        self.copy_tree(os.path.join(CMAKE_BUILD_DIR,"install","lib"),
```
jcwchen, line 227:
```suggestion
        self.copy_tree(os.path.join(CMAKE_BUILD_DIR, "install", "lib"),
```
Add a space after , to meet flake8. L225 needs the same fix.

jcwchen (changes requested):

jcwchen:
Thank you for proposing this PR. 
Questions: Are you saying it's a conda-specific bug? Is PyPi related to this?

AnthonyBarbier:
It's a bug related to all the packages generated using `setuptools.py` (Only the pure CMake build seems to be installing all the files as expected).
Which means the `pip install` / `setuptools.py` or `conda` packages all have the problem

AnthonyBarbier:
Is the circleci supposed to pass ? Pybind11 seems to be missing ?

jcwchen:
> Is the circleci supposed to pass ? Pybind11 seems to be missing ?

circleci will be deprecated so you can just ignore it.

BTW, currently I am releasing ONNX 1.8.0, typically we will remove those produced .pb during release process: https://github.com/onnx/onnx/blob/master/docs/OnnxReleases.md
I guess the reason is these .pb files will be protobuf sensitive. For instance, if the wheel built with newer protobuf, these .pb files would possibly not work with older protobuf, and vice versa. Therefore, as library's prospective, onnx won't release anything has strict protobuf dependency since the document does not mention any specific protobuf requirement. If users want to use these .pb files, they will need to build on their own based on their protobuf version.

Does it make sense to you? Thanks.

AnthonyBarbier:
> > Is the circleci supposed to pass ? Pybind11 seems to be missing ?
> 
> circleci will be deprecated so you can just ignore it.
> 
> BTW, currently I am releasing ONNX 1.8.0, typically we won't upload those produced .pb during release process.
> I guess the reason is these .pb files will be protobuf sensitive. For instance, if the wheel built with newer protobuf, these .pb files would possibly not work with older protobuf, and vice versa. Therefore, as library's prospective, onnx won't release anything has strict protobuf dependency since the document does not mention any specific protobuf requirement. If users want to use these .pb files, they will need to build on their own based on their protobuf version.
> 
> Does it make sense to you? Thanks.

You should make sure that none of the public headers include any of these pb files then: as it stands the library you distribute is pretty useless as the public headers keep having missing includes which means the user actually needs to rebuild them, unless I'm missing something ?

AnthonyBarbier:
The.pb.cc is compiled in the onnx library, so whatever you distribute already depends on a specific version of protobuf, no?

jcwchen:
To clarify: I am releasing ONNX 1.8.0 "PyPI" Package. And yes, it will depend on certain protobuf version, but we will do our best effort to make sure the built Python API supports backward and forward compatibility of protobuf. By contrast, these header files are more like intermediate files and won't be included in the Python package because we cannot guarantee whether these produced C++ headers from the release pipeline support backward and forward compatibility for protobuf.

If any user is interested in using ONNX through C/C++, they can generate these headers from source (under .setuptools-cmake-build/onnx) based on their local protoc, which is more compatible for them to use in their environments.

AnthonyBarbier:
Just for context: we only use the schema.h to let the users define their own custom op.
I just want a way for the user to have access to it without having to faff with building onnx or protobuf. 
Ideally just using Conda. It doesn't have to be python, I'm only interested in them being able to access the cpp interface

AnthonyBarbier:
Here is an example: https://github.com/graphcore/popart/blob/sdk-release-1.3/examples/custom_ops/custom_op.cpp#L43
Basically the op is built as a standalone shared library that python will then `dlopen`.

AnthonyBarbier:
FYI in the end I've created a separate Conda recipe to package only the C++ library https://github.com/conda-forge/staged-recipes/pull/13108

Cc @ezyang @marcelotrevisani

jcwchen:
Sorry for the late reply because I was busy with ONNX 1.8 Release...
I believe conda package is a better scenario for using ONNX through C/C++ because the protobuf version in conda would be fixed. @AnthonyBarbier Thank you so much for the contribution in conda-forge!

PR #3083, opened 2020-11-02
build protobuf in pipeline

askhade:

### DIFF
--- a/.github/workflows/release_win.yml
+++ b/.github/workflows/release_win.yml
@@ -1,9 +1,10 @@
 name: WindowsRelease
 
 on:
-  push:
-    branches:
-      - rel-*
+   push:
+    branches: [ master, rel-* ]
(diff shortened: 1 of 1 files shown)


snnn (approved):
LGTM.

But setup.py still need be changed. It should say ONNX  requires protobuf python package >=3.11.3

jcwchen:
LGTM. Thanks

jcwchen (approved):

askhade:
> LGTM.
> 
> But setup.py still need be changed. It should say ONNX requires protobuf python package >=3.11.3

Because we are pinning protobuf to 3.11.3 right now the package works with older versions too... for this release I prefer not to change this... after this release we will move protobuf to 3.13 and then add "ONNX requires protobuf python package >=3.13.0
" in setup.py

PR #3090, opened 2020-11-03
Use runtime_error instead of ONNX_NAMESPACE::InferenceError for shape inference

jcwchen:
**Description**
Use runtime_error instead of ONNX_NAMESPACE::InferenceError for onnx.shape_inference.

**Motivation**
On some platforms, it cannot recognize the error from ONNX_NAMESPACE::InferenceError without registering the custom errors. (For instance, tensorflow-onnx.) Besides, onnx.shape_inference should use the same error to handle the exception caused by onnx.shape_inference, at least for ONNX 1.8 Release.


askhade (approved):

gramalingam:
Hi @jcwchen : thanks. But for documentation purpose, can you clarify what exactly is the requirement about exceptions? We do internally use various classes of exceptions, and it is unclear to me what the caller is assuming in this case, or what the contract is.

jcwchen:
> Hi @jcwchen : thanks. But for documentation purpose, can you clarify what exactly is the requirement about exceptions? We do internally use various classes of exceptions, and it is unclear to me what the caller is assuming in this case, or what the contract is.

Currently throwing shape_inference error here causes some converters crashing on Windows with certain python version... Even though the exception here might be not related to the crash, it is safer to use the native exception (std::runtime_error) here instead of custom exception.

And yes, I agree with you we should discuss what shape_inference should throw in different scenarios so this modification has not been settled down yet, but that would be something we need to do after ONNX 1.8 Release. Thank you.

gramalingam:
Ok. But it sounds like there may be some other underlying issue relating to python-versions that needs to be fixed too, even though it may not be a priority right now.

gramalingam (approved):

onnx/shape_inference/implementation.cc
```diff
@@ -346,11 +346,11 @@ static void InferShapesImpl(
   deleteCreatedTypes(initializerTypeList);
   // Throw shape inference error if any
   if (!inference_errors.empty()) {
-    std::string full_errors = "Shape inference error(s): ";
+    std::cerr << "Shape inference error(s): " << std::endl;
```
gramalingam, line 349: Sorry, saw this just now. Is this temporary debugging?

jcwchen:
Another PR has fixed this.

PR #3095, opened 2020-11-05
Update pybind11 to 2.6 to solve seg fault on Windows

jcwchen:
**Description**
- Update pybind11 2.6.0
- Use runtime_error instead of ONNX_NAMESPACE::InferenceError for onnx.shape_inference.

**Motivation**
On some platforms, it cannot recognize the error from ONNX_NAMESPACE::InferenceError without registering the custom errors. (For instance, tensorflow-onnx.) Besides, onnx.shape_inference should use the same error to handle the exception caused by onnx.shape_inference, at least for ONNX 1.8 Release.


snnn:
What's the bug?

jcwchen:
> What's the bug? 

If onnx throws any kind of error from c++ to python (only if onnx builds with vs2019 and pybind11 2.4.3), converter using this onnx will occur segmentation fault. Therefore either building release onnx with vs2017- or pybind11     2.6.0 can resolve this issue... Will upgrading pybind11 bring any side effect? Otherwise using lower vs version is also  a solution. Thanks.

jcwchen:
We decide to update pybind11 (2.6.0) to fix this segmentation bug.

snnn:
"converter using this onnx will occur segmentation fault"

Then what if onnx was used alone without the converter, will it fail?

Is it a problem of ONNX or a compatibility issue of these two python packages?

jcwchen:
> "converter using this onnx will occur segmentation fault"
> 
> Then what if onnx was used alone without the converter, will it fail?
> 
> Is it a problem of ONNX or a compatibility issue of these two python packages?

ONNX itself works normally with the failed model from converter without involving the converter. 

- (vs2017 + pybind 2.4.3) worked
- (vs2019 + pybind 2.6.0) worked
- (vs2019 + pybind 2.4.3) failed

Therefore, we suppose it's a compatibility issue between VS Compiler and pybind11.

jcwchen:
We have tested between the workable and unworkable environment, and VS Compiler and pybind11 are the only variance... PyPI library versions and environment variables did not matter in our experiments.

gramalingam (approved):

askhade (approved):

snnn:
I highly suggest we should find the root cause, instead of trying all the different combinations to get a working one. And it's very likely the other people may entercounter the same issue again, no everyone is using our prebuilt packages, many of the users build onnx from source.  Even ORT is using the source form of ONNX, and vs2019 which seems won't work.

snnn:
> Will upgrading pybind11 bring any side effect? 

I think it should be ok. But if you don't know what went wrong, how will you be sure upgrading pybind will solve the issue? Would it happen again in a different build setting?

PR #3101, opened 2020-11-09
Function type-context unit-test

gramalingam:
Add unit test for function type-context extension.

### DIFF
--- a/onnx/test/cpp/function_context_test.cc
+++ b/onnx/test/cpp/function_context_test.cc
@@ -92,12 +92,15 @@ TEST(FunctionAPITest, TypeContextTest) {
   EXPECT_TRUE(schema);
   EXPECT_FALSE(schema->HasFunction());
   EXPECT_TRUE(schema->HasContextDependentFunction());
+
   NodeProto nodeProto;
   TypeProto floatTypeProto;
-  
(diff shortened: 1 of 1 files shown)


onnx/test/cpp/function_context_test.cc
```diff
@@ -0,0 +1,120 @@
...
+#include "onnx/defs/schema.h"
+
+using namespace ONNX_NAMESPACE::checker;
+
+namespace ONNX_NAMESPACE {
+namespace Test {
+
+// Utilities. TODO: Turn them into reusable ONNX utilities for use by
+
+TensorProto ToTensor(double value, TensorProto_DataType elem_type) {
```
askhade, line 15: Can you use https://github.com/onnx/onnx/blob/master/onnx/defs/tensor_proto_util.cc#L10 ?

gramalingam: No, this is a generalization of that. The difference is that this determines what to construct dynamically, based on the dynamic value of elem_type, while the other one has to be statically determined based on a static type.

gramalingam: This is worth pushing into the other utility (as commented in the above code), but I don't think it is ready yet. It works for the test-case. But there are issues like how to convert a float into a float16 value which need to be resolved for a fully-general solution.

askhade (approved):

PR #3105, opened 2020-11-10
fix: cleanup from #3084

### LINKED ISSUE #3084: Cannot build with pybind11 2.6.0
# Bug Report

### Describe the bug

We cannot build with pybind11 2.6 released recently, 2.5 works fine.
Basically a `-I` for `Python.h` is missing.

A bug is filed to them: https://github.com/pybind/pybind11/issues/2632
They mentioned it could be caused by the wrong use in onnx's cmake.
Could you take a look?

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): CentOS 7 / Ubuntu 18.04
- ONNX version (*e.g. 1.7*): Built from latest release tag (1.7.0)
- Python version: distro stock python3
- GCC/Compiler version (if compiling from source): gcc-9 / gcc-8
- CMake version: Built from latest release tag
- Protobuf version: Built from latest release tag

henryiii:
* Make `PYTHON_INCLUDE_DIRS` public (and use the official output "S" form)
* Quote directories to protect against directories with spaces

Fixes #3084


jcwchen:
> * Make PYTHON_INCLUDE_DIRS public (and use the official output "S" form)
> * Quote directories to protect against directories with spaces

It seems that it only can find `Python.h` under `PYTHON_INCLUDE_DIR`... I encountered the same error as CIs in my local.
Even I used the latest commit from pybind11 (>2.6.0) but the problem still persists. @henryiii  Do you have any idea what happened? Thank you.

jcwchen:
Maybe because ONNX does not build pybind11 while installing? (It only uses pybind11 as headers) Therefore, the produced Python.h cannot be found from uninstalled pybind11 anyway?

CMakeLists.txt
```diff
@@ -405,23 +405,26 @@ if(BUILD_ONNX_PYTHON)
   target_include_directories(onnx_cpp2py_export PRIVATE
                              $<BUILD_INTERFACE:${ONNX_ROOT}>
                              $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
-                             $<INSTALL_INTERFACE:include>
-                             ${PYTHON_INCLUDE_DIR})
+                             $<INSTALL_INTERFACE:include>)
```
henryiii, line 408: Python has not been found yet, so this was doing nothing.

CMakeLists.txt
```diff
@@ -405,23 +405,26 @@ if(BUILD_ONNX_PYTHON)
...
-      target_include_directories(onnx_cpp2py_export PUBLIC
-        ${ONNX_ROOT}/third_party/pybind11/include)
+  find_package(pybind11 2.2 CONFIG)
+  if(NOT pybind11_FOUND)
+    if(EXISTS "${ONNX_ROOT}/third_party/pybind11/include/pybind11/pybind11.h")
+      add_subdirectory("${ONNX_ROOT}/third_party/pybind11")
+      # Workaround for <2.6 in subdirectory mode, remove after update
+      if(NOT DEFINED pybind11_INCLUDE_DIRS)
+        set(pybind11_INCLUDE_DIRS "${PYBIND11_INCLUDE_DIR}")
+      endif()
```
henryiii: Before 2.6 we did not provide a unified interface in both modes. Once the submodule is updated (next commit), this is simpler.

CMakeLists.txt
```diff
@@ -405,23 +405,26 @@ if(BUILD_ONNX_PYTHON)
...
 
   # pybind11 is a header only lib
-  find_package(pybind11 2.2)
-  if(pybind11_FOUND)
-    target_include_directories(onnx_cpp2py_export PUBLIC
-      ${pybind11_INCLUDE_DIRS})
-  else()
-    if(EXISTS ${ONNX_ROOT}/third_party/pybind11/include/pybind11/pybind11.h)
-      target_include_directories(onnx_cpp2py_export PUBLIC
-        ${ONNX_ROOT}/third_party/pybind11/include)
```
henryiii, line 419: This path never found Python, so not sure how it was working...

CMakeLists.txt
```diff
@@ -447,8 +450,6 @@ if(BUILD_ONNX_PYTHON)
   target_link_libraries(onnx_cpp2py_export PRIVATE onnx)
 
   if(MSVC)
-    find_package(PythonInterp ${PY_VERSION} REQUIRED)
-    find_package(PythonLibs ${PY_VERSION} REQUIRED)
```
henryiii, line 451: These are already found as part of pybind11's Config or subdirectory modes.

CMakeLists.txt
```diff
@@ -405,23 +405,26 @@ if(BUILD_ONNX_PYTHON)
...
   # pybind11 is a header only lib
-  find_package(pybind11 2.2)
-  if(pybind11_FOUND)
-    target_include_directories(onnx_cpp2py_export PUBLIC
-      ${pybind11_INCLUDE_DIRS})
-  else()
-    if(EXISTS ${ONNX_ROOT}/third_party/pybind11/include/pybind11/pybind11.h)
-      target_include_directories(onnx_cpp2py_export PUBLIC
-        ${ONNX_ROOT}/third_party/pybind11/include)
+  find_package(pybind11 2.2 CONFIG)
```
henryiii, line 411: Including the word CONFIG here gives a better message. You could also add QUIET, since it's okay if it's not found.

henryiii:
CMake should _really_ be using targets (see [Modern CMake](https://cliutils.gitlab.io/modern-cmake/)), but this fixes the current issues (commit 1), then updates to 2.6.1 for the built in pybind11 (commit 2).

henryiii:
PS: This is fine except for one failing check that seems to be unrelated.

jcwchen:
circleCI can be ignored because it will be deprecated soon. Originally the usage of pybind11 in CMakeLists.txt was quite outdated... @henryiii Thank you so much for the modification and explanation. One question: so the commit ID for pybind11 in this PR will be pybind11 official v.2.6.1 release commit, right?

henryiii:
Yes, commit one only fixes the usage of variables. Commit 2 updates to 2.6.1. For either commit, they will still prioritize an "installed" pybind11 over the local one, and they will still allow 2.2+ to be found (I can obviously change that if you want).

henryiii:
Using 2.9.1, same changes as before.

jcwchen (commented):
Thank you for the quick update! The CIs look good. We will further verify this change in some related repos to prevent potential issues.

garymm (approved):
I tested applying this change in the pytorch/pytorch repo and it doesn't seem to cause any problems, so LGTM.

jcwchen (approved):
Thank you for the fix. I have tested it in ONNXRuntime and overall looks good.

gramalingam (approved):

PR #3106, opened 2020-11-11
fix collections deprecation error

annajung:
Running `pytest` outputs the following deprecation warning for `collections`.
Since it will stop working in 3.9, I made a small change to import ABC correctly using the supported format
I've kept it python 2 compatible as I don't think `onnx` is marked as python 3 only. I can get rid of try/except if that's not the case

```
onnx/test/checker_test.py::TestChecker::test_check_graph_topologically_sorted
  /Users/xxxxx/workspace/onnx/onnx/helper.py:339: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    is_iterable = isinstance(value, collections.Iterable)
```

### DIFF
--- a/.azure-pipelines/Linux-CI.yml
+++ b/.azure-pipelines/Linux-CI.yml
@@ -127,8 +127,7 @@ jobs:
 
       # Mypy only works with our generated _pb.py files when we install in develop mode, so let's do that
       pip uninstall -y onnx
-      pip install --no-use-pep517 -e .[mypy]
-      python setup.py --quiet typecheck
+      pip install mypy
       if [ $? -ne 0 ]; then
         echo "type check failed"
         exit 1
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -3,12 +3,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-try:
-    # python3
-    from collections import abc as collections_abc
-except ImportError:
-    # python2
-    import collections as collections_abc
+from collections.abc import Iterable
 
 import numbers
 from six import text_type, integer_types, binary_type
@@ -342,7 +337,7 @@ def make_attribute(
     if doc_string:
         attr.doc_string = doc_string
 
-    is_iterable = isinstance(value, collections_abc.Iterable)
+    is_iterable = isinstance(value, Iterable)
     bytes_or_false = _to_bytes_or_false(value)
     # First, singular cases
     # float
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -3,7 +3,13 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import collections
+try:
+    # python3
+    from collections import abc as collections_abc
+except ImportError:
+    # python2
+    import collections as collections_abc
+
 import numbers
 from six import text_type, integer_types, binary_type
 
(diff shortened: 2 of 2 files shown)


annajung:
looks like it doesn't work on Linux, closing

askhade:
@annajung : Linux CI is failing because of type check error:
onnx/helper.py:11: error: Name 'collections_abc' already defined
type check failed

One more thing... since we have deprecated support for python 2.7 for onnx packages we can go ahead and deprecate this in  CIs as well

annajung:
I think the problem is that my solution supports both python 2 and 3 which then is causing the error that `collections_abc` is already defined. 

and thank you for the information! I can remove the line for the second import for python 2 and hopefully, that will resolve it. Also, I can look into removing 2.7 tests from the CI!

annajung:
Build errors are the following
```
onnx/helper.py:6: error: No library stub file for standard library module 'collections.abc'
onnx/helper.py:6: note: (Stub files are from https://github.com/python/typeshed)
```

According to [mypy docs](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-type-hints-for-standard-library-module),
```
If you are getting a “No library stub file for standard library module” error, this means that you are attempting to import something from the standard library which has not yet been annotated with type hints.
```

However, [typeshed]() does support Iterable as seen [here](https://github.com/python/typeshed/blob/2949a9289ebdf2d3af427f9b5f5a4d4dbd0005a0/stdlib/3/collections/abc.pyi#L14)

I will close out the PR and create an issue to get more feedback on how to approach this problem.

askhade:
have you checked this: https://github.com/python/mypy/issues/3905

annajung:
Hi @askhade, I did see the option to hide the error using `–ignore-missing-imports`, but I didn't think that was the right approach. Is this applicable here? I assumed that any error produced by this line `python setup.py --quiet typecheck` should result in a fail build.

annajung:
looks like using the latest `mypy` worked with changing the import for `collections`. Maybe I can hold off on this until `mypy` version can be changed. I can look into the effort required to change the version if there are not objections :)

.azure-pipelines/Linux-CI.yml
```diff
@@ -127,8 +127,7 @@ jobs:
 
       # Mypy only works with our generated _pb.py files when we install in develop mode, so let's do that
       pip uninstall -y onnx
-      pip install --no-use-pep517 -e .[mypy]
-      python setup.py --quiet typecheck
+      pip install mypy
```
askhade, line 130: what is the intention behind this change?
pip install --no-use-pep517 -e .[mypy] --> will install onnx from current directory along with dependencies listed in the requirement group [] .i.e mypy in this case

whereas pip install mypy will install mypy

annajung: Hi @askhade thanks for following up. sorry, I thought I closed out this pr.

The change here is a wip, just to test to see if the latest `mypy` would work with Linux builds. 
It's not the change I want to commit/merge. I will close out this PR and either clean this up or open up a new PR based on work https://github.com/onnx/onnx/issues/3110

Thanks!

annajung:
Closing, details here https://github.com/onnx/onnx/pull/3106#discussion_r525429910

PR #3107, opened 2020-11-11
Remove python 2.7 CI build tests

annajung:
Follow up on https://github.com/onnx/onnx/pull/3106#issuecomment-725583770, with onnx deprecating support for python 2.7, modified builds to remove jobs that test python 2.7 compatability

### DIFF
--- a/.azure-pipelines/Linux-CI.yml
+++ b/.azure-pipelines/Linux-CI.yml
@@ -7,11 +7,6 @@ jobs:
     vmImage: 'Ubuntu-16.04'
   strategy:
     matrix:
-      py27:
-        python.version: '2.7'
-        onnx_ml: 0
-        onnx_debug: 0
-        onnx_lite: 1
       py36:
         python.version: '3.6'
         onnx_ml: 0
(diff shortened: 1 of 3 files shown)


.azure-pipelines/Linux-CI.yml
```diff
@@ -134,18 +125,4 @@ jobs:
         exit 1
       fi
 
-      # Mypy only works with Python 3
-      if [ "$(python.version)" != "2.7" ]; then
```
askhade, line 138: why remove this? The check is for !2.7

annajung: ah thank you for pointing that out. misread the code, will update!

annajung: well, actually since python version will never equal to 2.7, is the if statement still needed here?

annajung: I will put back the logic, but no longer wrap it in if statement :)

annajung: Updated! PTAL when you can, thanks again for the review!

askhade (approved):

PR #3113, opened 2020-11-13
Update Reshape op: Add 'allowzero' attribute - Cont

BowenBao:
Most of the changes were directly from #2552. The original PR is outdated after several ONNX releases. This PR brings it up to date.

======== From #2552 ========

This PR retains the behavior of old Reshape op, but introduces 'allowzero' attribute, which (if set) allows tensors to be reshaped with zeros in target shape, similar to NumPy.

Previous behavior was :

    input
    data = tensor of shape [1, 0]
    shape = [0, 1]
    output
    result = tensor of shape [1, 1]

The previous behavior is unchanged, unless 'allowzero' is set. If set, the behavior is as below:

    input
    data = tensor of shape [1, 0]
    shape = [0, 1]
    allowzero = 1
    output
    result = tensor of shape [0, 1]


BowenBao:
cc @gramalingam, @askhade, @spandantiwari, @postrational

onnx/defs/schema.h
```diff
@@ -893,7 +893,7 @@ class OpSchemaRegistry final : public ISchemaRegistry {
       // Version corresponding last release of ONNX. Update this to match with
       // the max version above in a *release* version of ONNX. But in other
       // versions, the max version may be ahead of the last-release-version.
-      last_release_version_map_[ONNX_DOMAIN] = 13;
+      last_release_version_map_[ONNX_DOMAIN] = 14;
```
askhade, line 896: this should be 13... this refers to the released opset version.

BowenBao: thanks, do i need to change line 886 above back to 13?

onnx/test/shape_inference_test.py
```diff
@@ -232,6 +232,24 @@ def test_reshape_static_shape_inferred(self):  # type: () -> None
             initializer=[make_tensor('shape', TensorProto.INT64, (3,), (0, 3, -1))])
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.UINT8, (2, 3, 4))])
 
+    def test_reshape_static_shape_zero(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.UINT8, (1, 0, 0)),
```
gramalingam, line 237: Can we make the shape of 'x' (1,1,1) ? Otherwise, this example is not a valid example, since the input and output don't have same size. Better to avoid invalid examples in our test-suite.

gramalingam: In the future, we could technically extend the shape-inference to reject such a test-case.

gramalingam:
LGTM thanks ... I had just one minor comment about one example.

onnx/test/shape_inference_test.py
BowenBao: Thx for catching, updated.

gramalingam (approved):

gramalingam:
Any idea why the LGTM:C/C++ analysis CI is failing?

gramalingam:
@postrational @askhade : any further comments on this extension before we merge this in?

BowenBao:
> 
> 
> Any idea why the LGTM:C/C++ analysis CI is failing?

I checked that other PR which include C++ modifications also fail this test https://github.com/onnx/onnx/pull/3141/checks?check_run_id=1533692794 with the same error Failed with bad exit code during 'Checkout'. Not sure what's the root cause.

gramalingam:
@BowenBao : sorry, it looks like some conflicts need to be resolved now before we can merge this

BowenBao:
> 
> 
> @BowenBao : sorry, it looks like some conflicts need to be resolved now before we can merge this

Will take a look

BowenBao:
@gramalingam rebased and all tests passed. Could you help merge the PR?

PR #3119, opened 2020-11-17
Fix test generation of SequenceInsert to front

take-cheeze:
`bool(np.array([0]))` is `False`


hariharans29 (approved):

jcwchen (approved):
LGTM. Thank you for proposing this. Could @onnx/sig-operators-approvers review this? Thanks.

gramalingam (approved):

PR #3120, opened 2020-11-18
Fix update_doc.sh behaving differently on different platforms

jcwchen:
**Description**
Some test scripts behave differently on different platforms. For instance, numpy will use `numpy.int32` by default on Windows and use `numpy.int64` by default on Mac. Specifying the variable type should resolve this issue.

- Specify `numpy.int64` for numpy functions
- Compare the old proto and the produced proto and only update if they are actually different since some outputs are produced differently with the same numpy code (numpy behaves slightly differently on different platforms)

**Motivation**
Currently if a user would like to modify certain operator and update related model/input/output, update_doc will update a lot of unrelated files. So the update needs to be done manually. This PR is try to make update_doc.sh only update those mattered files to prevent wrong update.


onnx/backend/test/cmd_tools.py
```diff
@@ -21,9 +23,44 @@
 def generate_data(args):  # type: (argparse.Namespace) -> None
 
     def prepare_dir(path):  # type: (Text) -> None
-        if os.path.exists(path):
-            shutil.rmtree(path)
-        os.makedirs(path)
```
jcwchen, line 26: Don't remove existing protos for the following comparison.

askhade:
Per offline discussion
this can be simplified to just provide an option to user to specify ops for which they want the data to be regenerated. Just this fix is not enough to resolve the current issue because the test generator creates the model for latest opset... therefore even though these diffs are fixed, the script will still modify a lot of models because the opset would have changed to the latest opset

jcwchen:
For now, just simply use --op_type to specify the target operator while generating data.

PR #3127, opened 2020-11-24
Deprecate outdated CircleCi: remove PyTorch tests from CI

jcwchen:
**Description**
To deprecate CircleCI, remote files under .circleci/

**Motivation and Context**
The test script in CircleCI is obsolete (some caffe2 tests cannot be passed by the latest ONNX shape inference) so currently CircleCI is failing... We don't see the value to keep it.


.circleci/build.sh
```diff
@@ -1,51 +0,0 @@
-#!/bin/bash
-
```
jcwchen, line 2: Besides removing these files, circleCI setting needs to be changed.

askhade (approved):

neginraoof (approved):
LGTM. Thanks for removing tests.

PR #3131, opened 2020-11-27
update ir for subgraph and optimizer, add some const methods

daquexian:
1. Add `forSelfAndEachSubGraph` and `forEachNode` to iterate over every (sub)graph and node. The latter method will be used in onnx optimizer.
2. Update `uses()` to include the uses in subgraph. It fixes some optimizer bugs about subgraph. It may not be the optimal way, as I'm not so familiar with the design of the existing ir. Any advice is welcomed.
3. Update `replaceAllUsesWith` to propagate `sizes`, `data_type` and `unique_name` if needed.
4. Fix the bug that opset version is not set in subgraph.
5. Add some missing const methods


onnx/common/ir.h
```diff
@@ -1250,5 +1300,23 @@ inline const_graph_node_list_iterator Node::reverseIterator() const {
...
 }
 
+inline const use_list Value::uses() const {
+  use_list all_uses = uses_in_current_graph_;
+  owningGraph()->forEachNode([this, &all_uses](const Node* node) {
+    if (node->owningGraph() == this->owningGraph()) {
+      // skip non-subgraph
+      return;
+    }
+    if (node->kind() == kCaptured) {
```
daquexian, line 1310: The names from outer scope are imported as a `kCaptured` node, https://github.com/onnx/onnx/blob/master/onnx/common/ir_pb_converter.cc#L277

onnx/common/ir.h
```diff
@@ -1250,5 +1300,23 @@ inline const_graph_node_list_iterator Node::reverseIterator() const {
   return iterator().reverse();
 }
 
+inline const use_list Value::uses() const {
```
jcwchen, line 1303: Maybe add some comments to explain what this function (`uses`) is?

daquexian: Thanks! Updated.

daquexian:
In this PR I also updated `replaceAllUsesWith` to propagate `sizes`, `data_type` and `unique_name` if needed.

onnx/common/ir.h
```diff
@@ -992,6 +982,14 @@ friend struct Value;
...
   }
 
+  size_t getNextUnique() {
+      std::string next_unique_name = ONNX_NAMESPACE::to_string(++next_unique_);
+      while(!isNameUnique(next_unique_name)) {
+          next_unique_name = ONNX_NAMESPACE::to_string(++next_unique_);
+      }
+      return next_unique_;
+  }
+
```
jcwchen, line 992: Is `getNextUnique` a const function? Did you move it to public because optimizer wants to use it?

daquexian: > Is getNextUnique a const function?

It is not const (but idempotent)

> Did you move it to public because optimizer wants to use it?

Yes. Since the methods like `addInitializer(Tensor initializer, std::string name)` is public, `getNextUnique` should also be public to get a unique name as the argument.

linkerzhang (approved):

jcwchen (approved):

daquexian:
@linkerzhang @jcwchen could you please merge this?

onnx/common/ir.h
```diff
@@ -1138,6 +1136,41 @@ friend struct Value;
...
+        } else if (node->kindOf(attr) == AttributeKind::gs) {
+          for (const auto& subgraph : node->gs(attr)) {
+            subgraph->forSelfAndEachSubGraph(fn);
+          }
+        }
+      }
+    }
+  }
+
+  void forSelfAndEachSubGraph(std::function<void(const Graph*)> fn) const {
```
askhade, line 1156: I am confused why do you need a const and a non const version of these methods? cant the invoker const cast and directly call the non const version?

onnx/common/ir.h
```diff
@@ -1177,14 +1210,68 @@ inline const Graph * Value::owningGraph() const {
   return node()->owningGraph();
 }
 
+// `captured` nodes in subgraph determines which value it captures
+// by storing the value's unique name, so old unique names in `captured` nodes
+// should also be updated.
+inline Value* Value::setUniqueName(const std::string &name, bool rename_subgraph_captured_nodes) {
+  if (has_unique_name() && rename_subgraph_captured_nodes) {
+    Graph *graph = owningGraph();
```
askhade, line 1218: use auto where applicable

onnx/common/ir.h
```diff
@@ -1177,14 +1210,68 @@ inline const Graph * Value::owningGraph() const {
...
-  newValue->uses_.reserve(uses().size());
-  for(auto u : uses()) {
+  // propagate sizes and elem type
+  if (this->has_sizes()) {
+    newValue->setSizes(this->sizes());
+  }
+  if (this->elemType() != ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED) {
+    newValue->setElemType(this->elemType());
+  }
+  Graph* graph = owningGraph();
```
askhade, line 1246: this can be moved above line 1238

onnx/common/ir.h
daquexian: I don't think it is a good habit to do const_cast and call the non const version.
For the const `forSelfAndEachSubGraph` method, the argument of `fn` is `const Graph`, that means `fn` cannot (and should not) modify the graph. For the non-const version there is not such a constraint

PR #3135, opened 2020-12-01
Add default value for type-context in FunctionBodyBuildContextImpl constructor 

gramalingam:
Add a default-value for type-context in FunctionBodyBuildContextImpl constructor for backward-compatibility.

This will allow users to upgrade to this version of ONNX without breaking existing usage, allowing them to add the type-context when needed.

Add a test-case to test the usage as well.


askhade (approved):

wschin (approved):
LGTM.

Issue #3086, opened 2020-11-03
Backend test runner doesn't honor SequenceProto inputs/outputs


hariharans29:
# Bug Report

Noticed that the onnx backend test runner doesn't have logic [in place](https://github.com/onnx/onnx/blob/5cc790e03b2af691c55926a11c116eb9edaa689b/onnx/backend/test/runner/__init__.py#L298) to handle `SequenceProto`s in the model's test inputs/outputs. Currently, it assumes that everything is a TensorProto. There are exisitng node tests that use sequences as inputs/outputs (for example: if_seq, loop13_seq, sequence_insert_at_front, sequence_insert_at_back) that is unable to be handled by the runner. 

This change added support for `SequenceProto`s : https://github.com/onnx/onnx/pull/2581

PR #3136, opened 2020-12-01
Enable input/output as SequenceProto for test runner

jcwchen:
**Description**

- Check whether input/output is a SequenceProto
- Check whether it's a empty sequence
- Refactor original code

ORT can test ONNX backend runner for `test_if_seq_cpu` and `test_loop13_seq_cpu` with this PR. Once this PR is merged, I will send a PR in ORT to remove the skips for those tests. 

**Motivation**
Solve issue: https://github.com/onnx/onnx/issues/3086
After https://github.com/onnx/onnx/pull/2581 merged, test runner does not further handle Sequence input and output. This PR is make test runner detect SequenceProto and convert it to numpy correctly


jcwchen:
cc @hariharans29 for review. Thank you!

onnx/backend/test/runner/__init__.py
```diff
@@ -295,21 +295,32 @@ def run(test_self, device):  # type: (Any, Text) -> None
                 inputs_num = len(glob.glob(os.path.join(test_data_dir, 'input_*.pb')))
                 for i in range(inputs_num):
                     input_file = os.path.join(test_data_dir, 'input_{}.pb'.format(i))
-                    tensor = onnx.TensorProto()
-                    with open(input_file, 'rb') as f:
-                        tensor.ParseFromString(f.read())
-                    inputs.append(numpy_helper.to_array(tensor))
+                    self._load_proto(input_file, inputs)
```
gramalingam, line 298: I think that the best approach is to get the types of inputs from the model (read above) and use the types to determine how to load the data. See, for example, here: https://github.com/onnx/onnx/blob/b353e8ba02539ace81dcb762dc78ff4df27cd9b8/onnx/backend/test/cpp/driver/test_driver.cc#L208

onnx/backend/test/runner/__init__.py
```diff
@@ -295,21 +295,30 @@ def run(test_self, device):  # type: (Any, Text) -> None
...
                                             atol=model_test.atol)
 
         self._add_test(kind + 'Model', model_test.name, run, model_marker)
+
+    def _load_proto(self, proto_filename, target_list, model_type_proto):  # type: (Text, List[Union[np.ndarray[Any], List[Any]]], TypeProto) -> None
+        with open(proto_filename, 'rb') as f:
+            protobuf_content = f.read()
+            type_proto_string = str(model_type_proto)
+            # this proto is a SequenceProto
+            if 'sequence_type' in str(type_proto_string):
```
jcwchen, line 316: I cannot find a good way to detect the type of TypeProto in Python so I use a sort of hacky way here...

onnx/backend/test/runner/__init__.py
jcwchen: Makes sense. However, it's hard to match the model.input/output to protobuf files (before knowing the right type and loading)... So I assume the order of model.input/output is same as the order of input/output proto files.

gramalingam: I don't understand your concern. The code looks fine to me, since for the i-th input/output we generate the corresponding file name, and get the type from graph.input and graph.output.

onnx/backend/test/runner/__init__.py
gramalingam: I think you can use ```model_type_proto.HasField("sequence_type")``` and, similarly, for the other cases. See https://developers.google.com/protocol-buffers/docs/reference/python-generated#oneof for more information.

onnx/backend/test/runner/__init__.py
```diff
@@ -288,28 +288,33 @@ def run(test_self, device):  # type: (Any, Text) -> None
...
         self._add_test(kind + 'Model', model_test.name, run, model_marker)
+
+    def _load_proto(self, proto_filename, target_list, model_type_proto):  # type: (Text, List[Union[np.ndarray[Any], List[Any]]], TypeProto) -> None
+        with open(proto_filename, 'rb') as f:
+            protobuf_content = f.read()
+            if model_type_proto.HasField('sequence_type'):
+                sequence = onnx.SequenceProto()
+                sequence.ParseFromString(protobuf_content)
+                target_list.append(numpy_helper.to_list(sequence))
+            elif model_type_proto.HasField('tensor_type'):
```
hariharans29, line 317: Do we want to leave a message in the else branch saying : loading proto of that specific type (Map/Sparse Tensor) is currently not supported ?

onnx/backend/test/runner/__init__.py
```diff
@@ -288,28 +288,33 @@ def run(test_self, device):  # type: (Any, Text) -> None
...
+                    self._load_proto(output_file, ref_outputs, model.graph.output[i].type)
                 outputs = list(prepared_model.run(inputs))
                 self.assert_similar_outputs(ref_outputs, outputs,
                                             rtol=model_test.rtol,
                                             atol=model_test.atol)
 
         self._add_test(kind + 'Model', model_test.name, run, model_marker)
+
+    def _load_proto(self, proto_filename, target_list, model_type_proto):  # type: (Text, List[Union[np.ndarray[Any], List[Any]]], TypeProto) -> None
+        with open(proto_filename, 'rb') as f:
```
hariharans29, line 311: Need to check-in tests to cover this function ?

onnx/backend/test/runner/__init__.py
```diff
@@ -288,28 +288,33 @@ def run(test_self, device):  # type: (Any, Text) -> None
...
                     os.path.join(model_dir, "test_data_set*")):
                 inputs = []
                 inputs_num = len(glob.glob(os.path.join(test_data_dir, 'input_*.pb')))
                 for i in range(inputs_num):
                     input_file = os.path.join(test_data_dir, 'input_{}.pb'.format(i))
-                    tensor = onnx.TensorProto()
-                    with open(input_file, 'rb') as f:
-                        tensor.ParseFromString(f.read())
-                    inputs.append(numpy_helper.to_array(tensor))
+                    self._load_proto(input_file, inputs, model.graph.input[i].type)
```
hariharans29, line 297: Just thinking out aloud - does this logic of iterating through input/output proto files handle optional inputs/outputs ? Let us say we have input_0.pb and input_2.pb (with input 1 optional), does it break then ?

jcwchen: https://github.com/onnx/onnx/pull/3136#discussion_r534397071
Just like what @gramalingam said above, the input/output protobuf files are generated by the model.graph in the same order so I think it should be fine? If the input or output is optional, then it should not show in the model.graph either.

onnx/backend/test/runner/__init__.py
jcwchen: Sounds good. Added. Thanks

onnx/backend/test/runner/__init__.py
jcwchen: Good idea. Updated. Thank you

onnx/backend/test/runner/__init__.py
jcwchen: Since it's an inner test utility function and it won't be used by others, it should be OK not to test it? Besides, any issue of it will be caught by the original tests anyway.

hariharans29 (approved):

onnx/backend/test/runner/__init__.py
hariharans29: A comment is nice, but I was suggesting printing something for the user

gramalingam (approved):

PR #3137, opened 2020-12-02
fix function opset imports

askhade:

This PR fixes the handling of opsets imported by Function Proto. 
When function proto imports an opset then it is merged with the model level opset and in cases when function proto does not import any opsets then default is to use the model level opset imports. This fixes cases when function body ops are not from the same domain as the function op itself. It also gives flexibility to function authors to use primitive ops from different domain in function body without tying it to a particular opset.


onnx/shape_inference/implementation.cc
```diff
@@ -454,6 +475,12 @@ void InferShapeForFunctionNode(
   }
 
   for (auto& n : func->node()) {
+    // Resolve domain for node
+    auto it = func_opset_imports.find(n.domain());
+    if (it == func_opset_imports.end()) {
+      return;
```
gramalingam, line 481: Should there be an error/exception here?

onnx/defs/schema.h
```diff
@@ -821,7 +821,8 @@ class OpSchema final {
   void Finalize();
 
   // Build function with information stored in opschema
-  void BuildFunction(FunctionProto& function_body) const;
+  void OpSchema::BuildFunction(
```
gramalingam, line 824: Nit: The qualifier "OpSchema::" is unnecessary, I assume.

gramalingam:
(As discussed offline), it would be great to have a spec/design level discussion and documentation of the underlying issues and proposed resolution, since there seem to be quite a few concerns/issues relating to versioning here.

askhade:
> (As discussed offline), it would be great to have a spec/design level discussion and documentation of the underlying issues and proposed resolution, since there seem to be quite a few concerns/issues relating to versioning here.

Agree. I created an issue to track this: https://github.com/onnx/onnx/issues/3139

gramalingam (approved):

PR #3169, opened 2020-12-15
register python exceptions and update exceptions handling

askhade:


onnx/shape_inference/implementation.cc
```diff
@@ -292,11 +292,10 @@ static void InferShapesImpl(
               function_opset_imports[opset_import.domain()] = static_cast<int>(opset_import.version());
             } else {
               if (it->second != opset_import.version()) {
-                throw ONNX_NAMESPACE::InferenceError(
-                    "ONNX models do not support multiple opset version imports for a domain. Function " +
-                            schema->Name() + " imports opset version " + std::to_string(opset_import.version()) +
-                            " for domain " + (opset_import.domain().empty() ? "ai.onnx" : opset_import.domain()) +
-                            " where as the model imports opset version " + std::to_string(it->second));
+                fail_shape_inference("ONNX models don't support multiple opset version imports for a domain. Function ",
```
jcwchen, line 295: Is it a "shape" inference? I thought using ONNX_NAMESPACE::InferenceError is more proper

jcwchen (commented):
Others LGTM. Thank you so much for organizing exceptions.

onnx/cpp2py_export.cc
```diff
@@ -324,6 +327,8 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
   // Submodule `shape_inference`
   auto shape_inference = onnx_cpp2py_export.def_submodule("shape_inference");
   shape_inference.doc() = "Shape Inference submodule";
+  py::register_exception<InferenceError>(shape_inference, "InferenceError");
```
gramalingam, line 330: Is this intended to cover both shape-inference errors as well as type-errors? Wondering why module is called "shape inference", while the exception is plain "InferenceError".

onnx/shape_inference/implementation.cc
askhade: fail_shape_inference throws a InferenceError... not sure what you mean

onnx/cpp2py_export.cc
askhade: yes both module and exception class are intended to handle both type and shape inference.... I am simply registering the exception here...

onnx/shape_inference/implementation.cc
jcwchen: Yes eventually they will throw the same error. Just thought this one is actually not a "shape" inference error so directly using `fail_shape_inference` looks a little weird to me.

jcwchen (approved):

onnx/cpp2py_export.cc
```diff
@@ -334,19 +334,19 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
   py::register_exception<InferenceError>(shape_inference, "InferenceError");
 
 
-  shape_inference.def("infer_shapes", [](const py::bytes& bytes, bool check_type) {
+  shape_inference.def("infer_shapes", [](const py::bytes& bytes, bool check_type, bool strict_mode) {
```
gramalingam, line 337: I wonder if making strict_mode into an integer (called mode, restricted to values 0 and 1 currently) will give us more flexibility in the future. For example, may be we could treat it as a compatibility_mode indicator indicating that we desire the behavior in version 1.7 or 1.8, etc. In the future, we could potentially end up with more than 2 modes from a compatibility-mode perspective.

askhade: I am leaning towards still keeping the strict_mode as true\false in python and change the one in c++ interface to int... when we add more granularity we can still keep the python api for backward compatibility and add a new api in python for the new approach

gramalingam (approved):

Issue #2324, opened 2019-09-17
Proposal to extend type-constraint language in operator schema


gramalingam:
This note is about the type-constraint language used in the operator schema definitions. Currently, we allow the type (string) of each input and output to be either a type-variable (e.g., "T") or a type-constant (e.g., "tensor(float)"). (E.g., see 
https://github.com/onnx/onnx/blob/e2e12d97af735b8d25ecdbd8cf013c18d5b8d44f/onnx/defs/schema.cc#L635
as well as 
https://github.com/onnx/onnx/blob/e2e12d97af735b8d25ecdbd8cf013c18d5b8d44f/onnx/defs/data_type_utils.h#L19 )

Each type-variable can be bound to a set of type-constants. (See: 
https://github.com/onnx/onnx/blob/e2e12d97af735b8d25ecdbd8cf013c18d5b8d44f/onnx/defs/schema.cc#L602 )

This restricted language means that each input and output has one of finitely many fixed types.

This is a proposal to extend the type-constraint language. As motivation, consider the recently introduced operations involving the new sequence type, such as SequenceAt, SequenceInsert, SequenceErase, SequenceLength. Currently, these operations are all restricted to "Sequence of tensors". Conceptually, the operations can be used for sequences of any type (e.g, a sequence of sequence of tensors). 

The proposal is to
(a) Extend the input/output type string to a type-expression, allowing expressions such as "sequence(T)" where T is a type-variable
(b) Permit type-variables without any type-constraint (meaning that it can be any type whatsoever).

PR #3170, opened 2020-12-15
Add sequence type for identity op

BowenBao:


onnx/defs/tensor/defs.cc
```diff
@@ -2202,15 +2202,20 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            OpSchema::all_tensor_types_with_bfloat(),
-            "Constrain input and output types to all tensor types.")
+            "V",
+            [](){
+              auto t = OpSchema::all_tensor_types_with_bfloat();
+              auto s = OpSchema::all_tensor_sequence_types();
+              t.insert(t.end(), s.begin(), s.end());
+              return t;
+            }(),
+            "Constrain input and output types to all tensor and sequence types.")
```
gramalingam, line 2218: This is okay for this PR. But just wanted to point to the issue discussed here: https://github.com/onnx/onnx/issues/2324 ... in the long run, this is going to be limiting because this does not allow for sequences of sequences of tensors or maps, etc. It would be useful to extend this as discussed in the issue above. Any thoughts about this?

gramalingam:
Minor nit: for the changes in defs.cc, it would be helpful to remove the clang-format changes of existing code (I think it is helpful to have pure formatting changes as a separate PR).

onnx/defs/shape_inference.h
```diff
@@ -393,25 +393,35 @@ inline void appendSingleDimCopiedFromInputTypeToOutputType(
...
 inline void propagateShapeFromInputToOutput(
     InferenceContext& ctx,
     size_t inputIndex,
     size_t outputIndex) {
   auto output_type = ctx.getOutputType(outputIndex);
   auto input_type = ctx.getInputType(inputIndex);
 
-  if (TypeProto::kTensorType != input_type->value_case() ||
-      TypeProto::kTensorType != output_type->value_case()) {
+  if (TypeProto::kTensorType == input_type->value_case() &&
```
gramalingam, line 415: It would be more general to move this if-then-else logic into "propagateShape". For example, then it would work for sequence(sequence(tensor(...))) as well.

gramalingam: @BowenBao : does the above make sense? Thanks!

BowenBao: @gramalingam sorry for the late reply, will take a look and make updates accordingly.

BowenBao: updated

gramalingam (approved):

PR #3171, opened 2020-12-16
fix bugs in external data helpers and add add size thresholds for ser…

askhade:
This PR addresses issues mentioned in https://github.com/onnx/onnx/pull/2332 and a few more.

1. Current code marks all tensors to be serialized as external data. Fixing this to only mark the ones which have raw data.
2. Add a threshold for tensor data size. Only tensor's with size >= threshold will be serialized to external data. This feature was designed to work with model >2GB and with current approach it tries to serialize every single tensor to external file. Therefore adding a threshold so that callers can choose which tensors to serialize
3. Add checks for filename correctness. Current code uses tensor names as filenames. This fails when the names contains illegal characters.

### DIFF
--- a/onnx/external_data_helper.py
+++ b/onnx/external_data_helper.py
@@ -98,7 +98,7 @@ def set_external_data(tensor,  # type: TensorProto
 
 
 def convert_model_to_external_data(model, all_tensors_to_one_file=True, location=None, size_threshold=1024):
-    # type: (ModelProto, bool, Optional[Text], Optional[int]) -> None
+    # type: (ModelProto, bool, Optional[Text], int) -> None
     """
     Call to set all tensors with raw data as external data. This call should preceed 'save_model'.
     'save_model' saves all the tensors data as external data after calling this function.
--- a/onnx/external_data_helper.py
+++ b/onnx/external_data_helper.py
@@ -98,7 +98,7 @@ def set_external_data(tensor,  # type: TensorProto
 
 
 def convert_model_to_external_data(model, all_tensors_to_one_file=True, location=None, size_threshold=1024):
-    # type: (ModelProto, bool, Optional[Text]) -> None
+    # type: (ModelProto, bool, Optional[Text], Optional[int]) -> None
     """
     Call to set all tensors with raw data as external data. This call should preceed 'save_model'.
     'save_model' saves all the tensors data as external data after calling this function.
@@ -220,7 +220,7 @@ def uses_external_data(tensor):  # type: (TensorProto) -> bool
     return tensor.HasField("data_location") and tensor.data_location == TensorProto.EXTERNAL
 
 
-def _remove_external_data_field(tensor, field_key):  # type: (TensorProto, Text) -> None
+def remove_external_data_field(tensor, field_key):  # type: (TensorProto, Text) -> None
     """
     Remove a field from a Tensor's external_data key-value store.
 
--- a/onnx/external_data_helper.py
+++ b/onnx/external_data_helper.py
@@ -208,7 +208,7 @@ def _sanitize_path(path):  # type: (Text) -> Text
 def _is_valid_filename(filename):  # type: (Text) -> bool
     """Utility to check whether the provided filename is valid."""
     exp = re.compile("^[^<>:;,?\"*|/]+$")
-    match = exp.fullmatch(filename)
+    match = exp.match(filename)
     if match:
         return True
     else:
--- a/onnx/external_data_helper.py
+++ b/onnx/external_data_helper.py
@@ -2,9 +2,10 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
-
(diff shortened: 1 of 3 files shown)


yufenglee:
Are there any test cases that need to be updated?

askhade:
> Are there any test cases that need to be updated?

Done

yufenglee (approved):

onnx/external_data_helper.py
```diff
@@ -193,12 +205,22 @@ def _sanitize_path(path):  # type: (Text) -> Text
...
+        return False
+
+
 def uses_external_data(tensor):  # type: (TensorProto) -> bool
     """Return true if the tensor stores data in an external location."""
     return tensor.HasField("data_location") and tensor.data_location == TensorProto.EXTERNAL
 
 
-def remove_external_data_field(tensor, field_key):  # type: (TensorProto, Text) -> None
+def _remove_external_data_field(tensor, field_key):  # type: (TensorProto, Text) -> None
```
jcwchen, line 223: Actually I don't see it has been used. Not sure why it is here.

jcwchen (approved):

wschin (approved):

onnx/external_data_helper.py
```diff
@@ -78,6 +79,9 @@ def set_external_data(tensor,  # type: TensorProto
                       checksum=None,  # type: Optional[Text]
                       basepath=None  # type: Optional[Text]
                       ):  # type: (...) -> None
+    if not tensor.HasField("raw_data"):
+        raise ValueError("Tensor " + tensor.name + "does not have raw_data field. Cannot set external data for this tensor.")
+
     del tensor.external_data[:]
```
wschin, line 85: Maybe we should clean `tensor.raw_data` as well because it's being set to external

wschin: > Maybe we should clean `tensor.raw_data` as well because it's being set to external

onnx/external_data_helper.py
```diff
@@ -93,31 +97,39 @@ def set_external_data(tensor,  # type: TensorProto
...
+    it will be converted to external data. To convert every tensor with raw data to external data set size_threshold=0.
     """
     if all_tensors_to_one_file:
         file_name = Text(uuid.uuid1())
         if location:
             file_name = location
         for tensor in _get_all_tensors(model):
-            set_external_data(tensor, file_name)
+            if tensor.HasField("raw_data") and sys.getsizeof(tensor.raw_data) >= size_threshold:
+                set_external_data(tensor, file_name)
```
gramalingam, line 120: Shouldn't multiple tensors be stored at different offsets within the same file? I don't see any offsets here. How does this work? (Alternatively, if that feature is not being used, may be we can remove it for now until we can add a proper version.)

onnx/external_data_helper.py
```diff
@@ -93,31 +97,39 @@ def set_external_data(tensor,  # type: TensorProto
             entry.value = str(v)
 
 
-def convert_model_to_external_data(model, all_tensors_to_one_file=True, location=None):
+def convert_model_to_external_data(model, all_tensors_to_one_file=True, location=None, size_threshold=1024):
     # type: (ModelProto, bool, Optional[Text]) -> None
     """
-    call to set all tensors as external data. save_model saves all the tensors data as external data after calling this function.
+    Call to set all tensors with raw data as external data. This call should preceed 'save_model'.
+    'save_model' saves all the tensors data as external data after calling this function.
```
gramalingam, line 104: Could we just call save_model as the last step of this function and rename it "save_model_with_external_data" or something like that? What is the value in splitting it into two different calls?

askhade: I don't know why this was split in 2 calls... I am planning to add another method which will do all this in a single call. For now let's keep this method because there are partners who have a dependency on this and I don't want to change this before prior notification.

onnx/external_data_helper.py
askhade: This method simply marks the tensors to use external data. The offsets are properly added when the data is actually written to the file. See save_external_data line 175

onnx/external_data_helper.py
askhade: right now converting a model to external data is done in 2 steps. In the first step tensors are simply marked to use external data and in the second step the data is actually serialized. This method is called in both the steps. So we cant simply clean the raw_data here...

onnx/external_data_helper.py
gramalingam: This is one reason to expose a single method that does both things. Otherwise, we have a fragile dependence between the two parts, with the model being in a weird non-standard representation in between the two steps ... if the user ends up doing other things in between the two steps, the potential for bugs increases.

gramalingam (approved):

onnx/external_data_helper.py
askhade: yes I agree... we should have a single method to do this... I will cover it as part of a new PR

PR #3175, opened 2020-12-16
Add default case to relied_opsets in BuildFunction

jcwchen:
**Description**
https://github.com/onnx/onnx/pull/3137
Build function should consider the case of `relied_opsets` is empty.

**Motivation**
ORT will fail with the latest ONNX because of the following errors:
```
1: [  FAILED  ] GraphTransformationTests.DynamicQuantizeMatMulTest 
1: [  FAILED  ] GraphTransformationTests.DynamicQuantizeMatMulTest_With_Bias
1: [  FAILED  ] GraphTransformationTests.DynamicQuantizeMatMulTest_With_ND_bias
1: [  FAILED  ] GraphTransformationTests.DynamicQuantizeMatMulTest_With_Bias_No_B_ZP
1: [  FAILED  ] GraphTransformationTests.MatMulIntegerToFloatTest
```
The optimizer fusion (MatMulIntegerToFloat) does not happen normally. Something in build function breaks.

### DIFF
--- a/onnx/defs/schema.cc
+++ b/onnx/defs/schema.cc
@@ -786,8 +786,16 @@ void OpSchema::BuildFunction(FunctionProto& function_body, const std::vector<Ope
     function_body.add_attribute(a.first);
   }
 
-  for (auto& relied_opset : relied_opsets) {
-    *(function_body.mutable_opset_import()->Add()) = relied_opset;
+  if (!relied_opsets.empty()) {
+    for (auto& relied_opset : relied_opsets) {
+      *(function_body.mutable_opset_import()->Add()) = relied_opset;
+    }
+  } else {
+    // By default, the function body graph is relying on the OperatorSet this
+    // function belongs to.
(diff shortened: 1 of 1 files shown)


onnx/defs/schema.cc
```diff
@@ -786,8 +786,16 @@ void OpSchema::BuildFunction(FunctionProto& function_body, const std::vector<Ope
     function_body.add_attribute(a.first);
   }
 
-  for (auto& relied_opset : relied_opsets) {
-    *(function_body.mutable_opset_import()->Add()) = relied_opset;
+  if (!relied_opsets.empty()) {
+    for (auto& relied_opset : relied_opsets) {
```
askhade, line 790: why do we need a default here? I am not convinced we need this

jcwchen: After discussion, the original behaviour in ONNX is correct. Instead, ORT should specify what the relied_opsets are for the function op. Close this PR now.

PR #3181, opened 2020-12-23
Add Linux release pipeline

jcwchen:
**Description**
- Move Linux release CI from https://github.com/onnx/wheel-builder to GitHub Action here 
- Add our own manylinux docker image for manylinux wheel build

**Motivation**
Same as Windows Release, ONNX will use GitHub Action to release the wheels on all platforms. This PR is try to enable Linux release with manylinux2010.


workflow_scripts/install_linux_proto.sh
```diff
@@ -0,0 +1,14 @@
+yum -y install wget
+
+export NUMCORES=`grep -c ^processor /proc/cpuinfo`
```
snnn, line 3: nit: this is the number of logical processors, not cores.

workflow_scripts/install_linux_proto.sh
```diff
@@ -0,0 +1,14 @@
...
+
+pb_dir="./cache/pb"
+PB_VERSION=2.6.1
```
snnn, line 6: Is there a centralized place to manage the protobuf versions we are using? In other words, are we using 2.6.1 in all the places?

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,69 @@
...
+  # test this PR for now, remove it before the merge
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
```
snnn, line 16: I suggest you may consider to add 3.9 there.

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,69 @@
...
+        python-version: ${{ matrix.python-version }}
+        architecture: ${{ matrix.architecture }}
+    - name: Install python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pytest==5.4.3 nbval numpy
+        python -m pip install virtualenv
+        python -m virtualenv py$(python.version)
+        source py$(python.version)/bin/activate
+        python -m pip install numpy protobuf==3.11.3
```
snnn, line 38: Here it is 3.11.3.

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,69 @@
...
+    - name: Build manylinux Python wheels
+      env:
+        ONNX_BUILD_TESTS: 1
+        USE_MSVC_STATIC_RUNTIME: 1
+        ONNX_ML: 1
+        CMAKE_ARGS: "-DONNX_USE_LITE_PROTO=ON"
+      uses: RalfG/python-wheels-manylinux-build@v0.3.3-manylinux2010_x86_64
+      with:
+        python-versions: 'cp36-cp36m cp37-cp37m cp38-cp38m'
+        build-requirements: 'numpy protobuf'
```
snnn, line 50: You may consider to install protobuf python package from source. Otherwise you should be careful about the versioning. 
Here you will get the latest protobuf python package, in most of the cases, it's fine. Unless you got a newer version of protoc and libprotobuf C/C++ library.

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,69 @@
...
+        python -m pip install numpy protobuf==3.11.3
+        sudo apt-get install protobuf-compiler libprotoc-dev
+
+    - name: Build manylinux Python wheels
+      env:
+        ONNX_BUILD_TESTS: 1
+        USE_MSVC_STATIC_RUNTIME: 1
+        ONNX_ML: 1
+        CMAKE_ARGS: "-DONNX_USE_LITE_PROTO=ON"
+      uses: RalfG/python-wheels-manylinux-build@v0.3.3-manylinux2010_x86_64
```
snnn, line 47: So, it will use a docker image built from https://github.com/ralfg/python-wheels-manylinux-build ? And we have no control on that, we don't know how often the image is updated? It is ok, but I'd rather not taking such a dependency unless it helps a lot.

jcwchen: Yes you are right. I am doing some experiment with that docker first. Actually many dependencies in that docker need to be updated so using our own docker image would be a better solution as you suggested. Now I am trying to make it work first then will decide what to do here. Thank you!

.github/workflows/release_linux.yml
jcwchen: Yes... Some of these libraries are not the latest ones and they need to be updated manually...

workflow_scripts/install_linux_proto.sh
```diff
@@ -1,15 +1,24 @@
 yum -y install wget
-yum -y install ccache
 
-export NUMCORES=`grep -c ^processor /proc/cpuinfo`
+wget http://ftp.mirrorservice.org/sites/sourceware.org/pub/gcc/releases/gcc-5.2.0/gcc-5.2.0.tar.gz
```
snnn: No, you shouldn't change gcc. If the default one in the manylinux docker image doesn't work, it must because you did something wrong.

jcwchen: Actually the purpose of installing gcc here is to get the latest `libstdc++.so.6`.
While installing 3.11.3 protobuf, it encounters the following error:
```
2020-12-31T04:58:13.8924662Z  You could try using --skip-broken to work around the problem
2020-12-31T04:58:13.8925459Z            Requires: libstdc++.so.6(GLIBCXX_3.4.20)(64bit)
2020-12-31T04:58:13.8926555Z Error: Package: protobuf-3.11.2-2.el8.x86_64 (/protobuf-3.11.2-2.el8.x86_64)
2020-12-31T04:58:13.8927400Z            Requires: libstdc++.so.6(CXXABI_1.3.8)(64bit)
2020-12-31T04:58:13.8928777Z Error: Package: protobuf-compiler-3.11.2-2.el8.x86_64 (/protobuf-compiler-3.11.2-2.el8.x86_64)
2020-12-31T04:58:13.8929927Z            Requires: libstdc++.so.6(GLIBCXX_3.4.21)(64bit)
2020-12-31T04:58:13.8931307Z Error: Package: protobuf-compiler-3.11.2-2.el8.x86_64 (/protobuf-compiler-3.11.2-2.el8.x86_64)
2020-12-31T04:58:13.8932580Z            Requires: libstdc++.so.6(CXXABI_1.3.9)(64bit)
```
The original old libstdc++.so.6 does not have such things so I thought I need to get a newer libstdc++.so.6. To do that in CentOS, I only know I can get it from building gcc... `yum install` certain version of libstdC++ will encounter some issues like recursive dependencies failure. Do you have any suggestion for it? Thank you.

snnn: The link you have is for CentOS 8, but the docker image is based on CentOS 6, so it definitely won't work.
see:
cbs.centos.org/kojifiles/packages/protobuf/3.11.2/2.el8/x86_64/protobuf-3.11.2-2.**el8**.x86_64.rpm

Why don't you just compile protobuf from source? It would be much easier than build gcc.

workflow_scripts/install_linux_proto.sh
```diff
@@ -0,0 +1,15 @@
+export NUMCORES=`grep -c ^processor /proc/cpuinfo`
+
+git clone https://github.com/protocolbuffers/protobuf.git
+cd protobuf
+git checkout 3.11.x
+git submodule update --init --recursive
+./autogen.sh
```
snnn, line 7: https://github.com/protocolbuffers/protobuf/blob/master/src/README.md#binary-compatibility-warning

 --disable-shared and enable PIC.

workflow_scripts/install_linux_proto.sh
```diff
@@ -0,0 +1,15 @@
...
+
+./configure
```
snnn, line 9: CFLAGS="-fPIC -g -O2" CXXFLAGS="-fPIC -g -O2" ./configure --disable-shared
56

workflow_scripts/install_linux_proto.sh
jcwchen: Building protobuf from source is a good idea. Thank you @snnn. Happy New Year!!

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,70 @@
...
+      with:
+        python-versions: 'cp36-cp36m cp37-cp37m cp38-cp38m'
+        build-requirements: 'numpy protobuf'
+        system-packages: 'cmake3 glibc-devel libstdc++-devel python-devel'
```
snnn, line 52: No, don't install python-devel. All you need should be provided by manylinux. Don't get such changes from the system packages.

CMakeLists.txt
```diff
@@ -77,6 +77,12 @@ if(NOT ONNX_NAMESPACE)
   set(ONNX_NAMESPACE "onnx")
 endif()
 
+# find_package Python has replaced PythonInterp and PythonLibs since cmake 3.12
+# Use the following command in the future; now this is only compatible with the latest pybind11
+# find_package(Python ${PY_VERSION} COMPONENTS Interpreter DevelopmentREQUIRED)
+find_package(PythonInterp ${PY_VERSION} REQUIRED)
```
snnn, line 83: It depends on which cmake version you want to use.
For older cmake versions, please see how onnxruntime is handling how.  https://github.com/microsoft/onnxruntime/blob/master/tools/ci_build/github/azure-pipelines/templates/set-python-manylinux-variables-step.yml

For newer ones, please read https://gitlab.kitware.com/cmake/cmake/-/issues/20425

.github/workflows/release_linux.yml
jcwchen: https://github.com/pybind/pybind11/issues/1728
pybind11 needs such library (Python.h) which centOS doesn't have initially... I have tried to fix it from pybind11 side but failed. Do you have any idea about how to fix this without installing python-devel? Thank you!

CMakeLists.txt
```diff
@@ -77,6 +77,12 @@ if(NOT ONNX_NAMESPACE)
...
+find_package(PythonLibs ${PY_VERSION} REQUIRED)
```
jcwchen: https://github.com/onnx/onnx/pull/2961
Apply this PR: Make onnx build not always depend on python2
Originally centOS will use python2 even the python_version is 3.

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,92 @@
...
+    - uses: actions/upload-artifact@v1
+      with:
+        name: wheels
+        path: dist
+
+    - name: Test python3.6 wheel
+      run: |
+        python -m virtualenv py36
+        source py36/bin/activate
+        pip install dist/*-cp36-cp36m-manylinux_x86_64.whl
```
snnn, line 63: it ends with manylinux2010_x86_64.whl

.github/workflows/release_linux.yml
snnn: Replied below.

jcwchen: Do you mean here: `It depends on which cmake version you want to use...` ?
But after some experiments, either using `find_package(PythonInterp` or `find_package(Python` cannot resolve the error as follow:
```
/root/onnx/third_party/pybind11/include/pybind11/detail/common.h:112:20: fatal error: Python.h: No such file or directory
 #include <Python.h>
          ^~~~~~~~~~         
```
I also can repro this in my local CentOS. It seems to me that centOS does not have such library and it needs to be installed from `python-devel`...

.github/workflows/release_linux.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       uses: RalfG/python-wheels-manylinux-build@v0.3.3-manylinux2010_x86_64
       with:
         python-versions: 'cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39'
-        build-requirements: 'numpy'
+        build-requirements: 'numpy protobuf==3.11.3'
```
snnn, line 50: You need to set numpy version as well.

If a package was built with numpy version A, then at runtime it requires numpy version >=A.

jcwchen: Do you have any recommended version? I don't see any numpy requirement in ONNX so I don't know which version is appropriate...

.github/workflows/release_linux.yml
snnn: Yes. Please check how onnxruntime do this.

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Build manylinux Python wheels
+      env:
+        ONNX_BUILD_TESTS: 1
+        USE_MSVC_STATIC_RUNTIME: 1
+        ONNX_ML: 1
+        CMAKE_ARGS: "-DONNX_USE_LITE_PROTO=ON"
+      uses: RalfG/python-wheels-manylinux-build@v0.3.3-manylinux2010_x86_64
+      with:
+        python-versions: 'cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39'
+        build-requirements: 'numpy==1.16.6 protobuf==3.11.3'
```
jcwchen, line 47: Use `numpy==1.16.6` here for building wheel, which is same as what ORT is using

.github/workflows/release_linux.yml
snnn: I don't have. But you may ask the people in sigs.

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,57 @@
...
+  # test this PR for now, remove it before the merge
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
```
snnn, line 15: Is there a plan to add 3.9?

snnn:
Please don't forget to update https://github.com/onnx/onnx/blob/master/setup.py#L300 .
Change it to "numpy>=1.16.6".  (It can be in a separated PR).

snnn (approved):

.github/workflows/release_linux.yml
jcwchen: Since it's a patch for ONNX 1.8, we will stick the same supported version as ONNX 1.8. I think Python 3.9 would probably be added in ONNX 1.9. Thanks.

.github/workflows/manylinux/entrypoint.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Compile wheels
+# Need to be updated if there is a new Python Version
+declare -A python_map=( ["3.5"]="cp35-cp35m" ["3.6"]="cp36-cp36m" ["3.7"]="cp37-cp37m" ["3.8"]="cp38-cp38" ["3.9"]="cp39-cp39")
+declare -A python_include=( ["3.5"]="3.5m" ["3.6"]="3.6m" ["3.7"]="3.7m" ["3.8"]="3.8" ["3.9"]="3.9")
+PY_VER=${python_map[$PY_VERSION]}
+
+# set ONNX build environments
+export ONNX_BUILD_TESTS=1
+export USE_MSVC_STATIC_RUNTIME=1
```
askhade, line 43: nit: not needed

.github/workflows/manylinux/entrypoint.sh
```diff
@@ -0,0 +1,74 @@
...
+export ONNX_ML=1
+export CMAKE_ARGS="-DONNX_USE_LITE_PROTO=ON -DPYTHON_INCLUDE_DIR=/opt/python/${PY_VER}/include/python${python_include[$PY_VERSION]} -DPYTHON_LIBRARY=/usr/lib64/librt.so"
```
askhade, line 45: for 1.8 did we build linux packages with lite proto ON?

askhade (approved):

.github/workflows/manylinux/entrypoint.sh
jcwchen: We didn't. Thanks for the reminder.

PR #3185, opened 2020-12-28
Conv-14 and ConvTranspose-14

jiafatom:
Bump onnx version to 14.
Conv-14 and ConvTranspose-14.


jiafatom:
Move to [PR](https://github.com/onnx/onnx/pull/3188), because this one has DCO issue.

PR #3188, opened 2020-12-29
Update Conv/ConvTranspose/Pool spec on auto_pad

jiafatom:
(1) Make it clear in Conv/ConvTranspose spec on `auto_pad`, also affecting AveragePool, MaxPool, LpPool, ConvInteger, QLinearConv. For AvergePool, MaxPool, LpPool, they use same shape inference logic as Conv: convPoolShapeInference, so we also change it here.
```python
Conv auto_pad='SAME_LOWER'/'SAME_UPPER',
output_shape[i] = ceil(input_shape[i] / strides[i])
ConvTranspose auto_pad='SAME_LOWER'/'SAME_UPPER',
output_shape[i] = input_shape[i] * strides[i]
```
(2) Three shape inference tests added for validation: Conv, ConvTranspose, AvergePool.

Signed-off-by: David <jiafa@microsoft.com>

The original [PR](https://github.com/onnx/onnx/pull/3185) has DCO rebasing failure. Move that PR here.


onnx/defs/nn/defs.cc
```diff
@@ -851,7 +865,7 @@ computes the output.)DOC";
 
 ONNX_OPERATOR_SET_SCHEMA(
     Conv,
-    11,
+    14,
```
jcwchen, line 868: For this opset bump, do you only modify the document? Or you will make other updates? Thanks.

jiafatom: (1) I modify the document (2) I added two shape inference test for Conv-14 and ConvTranspose-14. That is all.
For these two shape inference test, it already works well without any change on shape inference.
This means that the shape inference already has this Conv-14 and ConvTranspose-14 behavior.
Previously we don't have shape inference test for this scenario, so there was no validation previously.

askhade:
please add relevant backend tests for this change

onnx/defs/nn/defs.cc
```diff
@@ -833,7 +847,7 @@ computes the output.)DOC";
         OPTIONAL_VALUE);
     schema.Attr(
         "auto_pad",
-        auto_pad_doc,
+        conv_auto_pad_14_doc,
```
askhade, line 850: If behavior of Conv is changing then the behavior for corresponding quantization ops should change too

onnx/defs/nn/defs.cc
```diff
@@ -21,6 +21,20 @@ const char* auto_pad_doc =
     "SAME_UPPER or SAME_LOWER mean pad the input so that the output spatial size match the input."
     "In case of odd number add the extra padding at the end for SAME_UPPER and at the "
     "beginning for SAME_LOWER. VALID mean no padding.";
+const char* conv_auto_pad_14_doc =
```
askhade, line 24: change name to conv_auto_pad_doc ... 14 is not required here... this file contains schema for the latest version of ops...

yufenglee:
Why do you need to bump up the version? It looks to me that this PR makes the spec more clear and doesn't have a substantial change.

jiafatom:
> Why do you need to bump up the version? It looks to me that this PR makes the spec more clear and doesn't have a substantial change.

There is an internal discussion for this issue, and the current agreement is to bump up the version. You can find it in the email thread.
In short, the current doc is `SAME_UPPER or SAME_LOWER mean pad the input so that the output spatial size match the input.`
The word `match` is vague and usually means `equal`. And current onnxruntime implements the way `equal`. My change need an adjustment by multiplying/dividing the strides - so there will be an onnxruntime change accordingly.

yufenglee:
> > Why do you need to bump up the version? It looks to me that this PR makes the spec more clear and doesn't have a substantial change.
> 
> There is an internal discussion for this issue, and the current agreement is to bump up the version. You can find it in the email thread.
> In short, the current doc is `SAME_UPPER or SAME_LOWER mean pad the input so that the output spatial size match the input.`
> The word `match` is vague and usually means `equal`. And current onnxruntime implements the way `equal`. My change need an adjustment by multiplying/dividing the strides - so there will be an onnxruntime change accordingly.

ORT implements it as ceil(input/stride):
https://github.com/microsoft/onnxruntime-openenclave/blob/bdc51bfe62822b2acd190c3bc98bfc92babf24e5/onnxruntime/core/providers/common.h#L96

yufenglee:
> > > Why do you need to bump up the version? It looks to me that this PR makes the spec more clear and doesn't have a substantial change.
> > 
> > 
> > There is an internal discussion for this issue, and the current agreement is to bump up the version. You can find it in the email thread.
> > In short, the current doc is `SAME_UPPER or SAME_LOWER mean pad the input so that the output spatial size match the input.`
> > The word `match` is vague and usually means `equal`. And current onnxruntime implements the way `equal`. My change need an adjustment by multiplying/dividing the strides - so there will be an onnxruntime change accordingly.
> 
> ORT implements it as ceil(input/stride):
> https://github.com/microsoft/onnxruntime-openenclave/blob/bdc51bfe62822b2acd190c3bc98bfc92babf24e5/onnxruntime/core/providers/common.h#L96

Agree that match is not very clear here, but it doesn't mean 'equal' or it will use 'equal'. I don't think any runtime implementation is so careless to implement it as output[i] = input[i]. If so, they will hit mismatch for model with this setting.

jiafatom:
> Agree that match is not very clear here, but it doesn't mean 'equal' or it will use 'equal'. I don't think any runtime implementation is so careless to implement it as output[i] = input[i]. If so, they will hit mismatch for model with this setting.

Some models are get affected. I have posted an issue for this in onnxruntime, see [here](https://github.com/microsoft/onnxruntime/issues/6067), thanks.

gramalingam:
I agree that the change is best seen as a clarification of the existing spec, so we can avoid bumping up the op version.

yufenglee (approved):

askhade:
The PR looks good to me... @jiafatom please add some backend test for validation

jiafatom:
> The PR looks good to me... @jiafatom please add some backend test for validation

@askhade Thanks for your comments. It looks that the current added test (shape inference test) already covers this PR. 
(1) This PR is mostly a clarification, it does not change any current behavior of these operators.
(2) This PR is about auto_pad 'SAME' on how much padding we need. So the shape inference test is the way to test whether the output_shape is correct. Then it is exactly the PR about. It does not change the way how Conv computes. So the added shape inference test is already enough.
Please let me know if you have further questions.

onnx/defs/nn/defs.cc
```diff
@@ -15,10 +15,18 @@ const char* pads_doc =
...
     "the end of axis `i`. This attribute cannot be used simultaneously with "
     "auto_pad attribute. If not present, the padding defaults to 0 along start and end of each spatial axis.";
-const char* auto_pad_doc =
+const char* conv_auto_pad_doc =
     "auto_pad must be either NOTSET, SAME_UPPER, SAME_LOWER or VALID. Where "
     "default value is NOTSET, which means explicit padding is used. "
-    "SAME_UPPER or SAME_LOWER mean pad the input so that the output spatial size match the input."
+    "SAME_UPPER or SAME_LOWER mean pad the input so that "
+    "`output_shape[i] = ceil(input_shape[i] / strides[i])` for each axis `i`. "
+    "In case of odd number add the extra padding at the end for SAME_UPPER and at the "
```
gramalingam, line 23: Suggest changing the ```In case of odd number''' sentence as below for readability:
```
The padding is split between the two sides equally or almost equally (depending on whether it is even or odd).
In case the padding is an odd number, the extra padding is added at the end for SAME_UPPER
and at the beginning for SAME_LOWER.
```

gramalingam: Same change below also

gramalingam:
@jiafatom : the shape-inference tests and backend tests serve different purposes. Backend tests are meant to test other backends (like ORT) with respect to the ONNX spec (and the shape-inference tests are not used for that purpose). The existing backend tests do not have a test-case to cover the auto-pad cases, so adding them would be useful and valuable. In fact, such a test-case would have caught this exact bug if we had already had such a test-case.

jiafatom:
> @jiafatom : the shape-inference tests and backend tests serve different purposes. Backend tests are meant to test other backends (like ORT) with respect to the ONNX spec (and the shape-inference tests are not used for that purpose). The existing backend tests do not have a test-case to cover the auto-pad cases, so adding them would be useful and valuable. In fact, such a test-case would have caught this exact bug if we had already had such a test-case.

@gramalingam As you see, auto_pad=SAME_X case only happens in dynamic input shape, but I only see static input shape in the backend tests for conv, conv_transpose. Adding a test for static input shape would not cover any real model cases here (when auto_pad=SAME_X). Seems that adding this test does not give benefits.

gramalingam:
The spec does not constrain the use of attributes to dynamic shape cases. Whether the shape is static or dynamic should not be relevant to the spec. My guess is that you are considering the use-cases that arise from the convertor, but the spec is broader and test-cases should cover cases permitted. We are adding this clarification because of a bug in ORT's convtranspose implementation. Why is there no test-case in the ONNX backend tests that would have caught this bug?

jiafatom:
> The spec does not constrain the use of attributes to dynamic shape cases. Whether the shape is static or dynamic should not be relevant to the spec. My guess is that you are considering the use-cases that arise from the convertor, but the spec is broader and test-cases should cover cases permitted. We are adding this clarification because of a bug in ORT's convtranspose implementation. Why is there no test-case in the ONNX backend tests that would have caught this bug?

@gramalingam I totally understand your point from spec perspective. Adding backend test is a burden for developers because of the ORT ConvTranspose bug: Need change ORT code to verify this change. Like I mentioned, adding test in ORT side is more convenient and also cover this. Just take a look at the two recent PRs that get committed (one PR bump onnx version, the other changes shape inference logic):
https://github.com/onnx/onnx/pull/3195
https://github.com/onnx/onnx/pull/2549
There are no backend tests added there. May I know what is the rule for adding backend test in general? Thank you!

gramalingam:
@jiafatom : Thanks, it looks like you were able to add the test-case anyway? As for the other two PRs you mention: one was fixing a bug in shape-inference (a previous bug fix which was unintentionally not duplicated for the older version of the op), but not changing/clarifying the spec in anyway; the second was adding new types, but not changing/clarifying the spec. The point was just that a test-case is useful for this op for this particular scenario. I understand your point about the difficulty of generating test-cases. If test-cases are difficult to generate, we can split the PR and do the test-case separately as needed.

gramalingam (approved):

jiafatom:
@askhade @gramalingam
Thank you very much for your comments, I added the related backend tests as you suggested.
When I add a backend test for ConvTranspose, I find a shape inference bug in ConvTranspose, it is in a very special case when `stride>1 && residue > 0`, so the previous tests do not catch it. The last change on this shape inference logic was written in 6/2019 [here](https://github.com/onnx/onnx/pull/2068), where the core logic was written even earlier. It is very difficult to resolve this bug, I spent several hours to fix this, and include the fix in this PR, and all tests passed.
Thanks again for the comments.

askhade:
LGTM

Regarding your comments on backend test - the idea is when anyone makes an update to an existing op or proposes a new OP, the update or the new OP should be done following some popular framework and should be model driven. If these 2 conditions are meant then it should not be very difficult to generate this test data.

If these test cases existed already then it would have saved us a lot of time because the bug in ORT would have been caught by these tests.

askhade (approved):

PR #3194, opened 2021-01-04
ONNX Parser

gramalingam:
Introduce
(a) A text-based syntax for ONNX models, 
(b) A parser,
(c) A pretty-printer, and
(d) Test-cases

The parser will help with a couple of use-cases: it provides a simpler syntax for defining functions; it also makes it easier to specify test-cases that generate models for testing various aspects.

For example, a toy model can be specified as below:
```
<
  ir_version: 7,
  opset_import: [ "" : 10 ]
>
agraph (float[N, 128] X, float[128,10] W, float[10] B) => (float[N] C)
{
    T = MatMul(X, W)
    S = Add(T, B)
    C = Softmax(S)
}
```


linkerzhang:
May I know the motivation of introducing the text based syntax and associated parser please? Is this for model visualization? Thank you very much!

gramalingam:
@linkerzhang : as mentioned in the above description, there are at least two motivations. One is that this serves as an extension of https://github.com/onnx/onnx/blob/8e99dcff98f32ab30fa31ec1fdbdc5eb0df11f04/onnx/defs/function.h#L29 to help simplify the definition of function bodies. Currently, it is too verbose. Second, it will help write test-cases the generate models fairly easily. For example, the value of this can be seen in how test-cases are written in MLIR/LLVM, where input and output are described in text.

gramalingam:
@linkerzhang : do you have any further comments or feedback? Does this make sense? Thanks!

onnx/defs/parser.h
```diff
@@ -0,0 +1,502 @@
...
+ private:
+  std::string expanded_message_;
+};
+
+#define parse_error(...)                                                                                          \
+  do {                                                                                                            \
+    throw ParseError(ONNX_NAMESPACE::MakeString("[ParseError at position ", (next_ - start_), "]", __VA_ARGS__)); \
+  } while (0)
+
+class PrimitiveTypeNameMap {
```
askhade, line 47: Can you use data_type_utils here?

onnx/defs/printer.cc
```diff
@@ -0,0 +1,154 @@
...
+  const char* sep = "[";
+  const char* comma = ", ";
+
+  os << attr.name() << " = ";
+  switch (attr.type()) {
+    case AttributeProto_AttributeType_INT:
+      os << attr.i();
+      break;
+    case AttributeProto_AttributeType_INTS:
+      for (auto v : attr.ints()) {
```
askhade, line 76: const auto& ?

onnx/defs/printer.cc
```diff
@@ -0,0 +1,154 @@
...
+      os << "\"" << attr.s() << "\"";
+      break;
+    case AttributeProto_AttributeType_STRINGS:
+      for (auto v : attr.strings()) {
+        os << sep << "\"" << v << "\"";
+        sep = comma;
+      }
+      os << "]";
+      break;
+    default:
```
askhade, line 102: Case where attrproto contains a graphproto or tensorproto

onnx/test/cpp/parser_test.cc
```diff
@@ -0,0 +1,219 @@
...
+
+TEST(ParserTest, AttributeTest) {
+  AttributeProto attr;
+
+  OnnxParser::Parse(attr, "x = 2");
+  EXPECT_EQ(attr.name(), "x");
+  EXPECT_EQ(attr.type(), AttributeProto_AttributeType::AttributeProto_AttributeType_INT);
+  EXPECT_EQ(attr.i(), 2);
+
+  OnnxParser::Parse(attr, "x = 0.625");
```
askhade, line 59: curious - will x=0.625f work?

onnx/test/cpp/parser_test.cc
```diff
@@ -0,0 +1,219 @@
...
+}
+)ONNX";
+
+  GraphProto graph;
+  OnnxParser::Parse(graph, code);
+
+  EXPECT_EQ(graph.name(), "agraph");
+  EXPECT_EQ(graph.input_size(), 2);
+  EXPECT_EQ(graph.output_size(), 1);
+  EXPECT_EQ(graph.node_size(), 2);
```
askhade, line 166: can you also add checks to verify the type and shape of inputs and outputs

askhade:
overall looks good to me... can you add a document describing the text syntax with some examples for writing test cases, functions etc...

onnx/defs/printer.cc
gramalingam: I think this should be fine for a primitive type like int/float. There is no efficiency gain using a reference to int/float vs using the actual value.

gramalingam: For other cases, like strings (or not-yet-handled tensors/graphs), I agree a const-reference should be used.

onnx/test/cpp/parser_test.cc
gramalingam: Good question. At this point, attributes support only int or float. So, we can use decimal points to distinguish between the two, and I think that is sufficient. (If we ever have multiple precisions for floats in attributes, we could extend this, but I don't see any demand for that now.)

gramalingam: But for tensors, it could make sense. Depends on the syntax for tensors (whether the type should be specified separately or not). May be worth thinking about.

onnx/defs/parser.h
```diff
@@ -0,0 +1,381 @@
...
+
+  Status Parse(NodeProto& node);
+
+  Status Parse(NodeList& nodelist);
+
+  Status Parse(GraphProto& graph);
+
+  Status Parse(std::string name, GraphProto& graph);
+
+  Status Parse(ModelProto& model);
```
postrational, line 372: Should these methods be part of the `public` interface?

gramalingam: I made some of them private methods, and left some of them as public. In general, I think code constructing parts of a model are common. Not all users are going to construct a complete-model in one go. So, I made what seems a reasonable classification for now. Thanks!

onnx/defs/parser.h
```diff
@@ -0,0 +1,382 @@
...
+
+using AttrList = google::protobuf::RepeatedPtrField<AttributeProto>;
+
+using ValueInfoList = google::protobuf::RepeatedPtrField<ValueInfoProto>;
+
+#define PARSE_ERROR(...)                                                                                        \
+  do {                                                                                                          \
+    return Status(                                                                                              \
+        NONE, FAIL, ONNX_NAMESPACE::MakeString("[ParseError at position ", GetCurrentPos(), "]", __VA_ARGS__)); \
+  } while (0)
```
tomdol, line 37: This macro looks kina hacky. You can achieve the same result by using a variadic template or some other construct that will be type safe.

onnx/defs/parser.h
```diff
@@ -0,0 +1,382 @@
...
+
+#define PARSER_CHECK(cond, msg) \
+  if (!(cond))                  \
+  PARSE_ERROR(msg)
+
+#define CHECK_PARSER_STATUS(status) \
```
tomdol, line 43: I understand that usage of this macro lets you avoid the if-statements at the call site but this is not a very modern-cpp approach. You could even switch to exceptions instead.

onnx/defs/parser.h
```diff
@@ -0,0 +1,382 @@
...
+
+  void RestorePos() {
+    next_ = saved_pos_;
+  }
+
+  std::string GetCurrentPos() {
+    uint32_t line = 1, col = 1;
+    for (const char* p = start_; p < next_; p++) {
+      if (*p == '\n') {
+        line++;
```
tomdol, line 160: In situations like this it is preferred to use `++line` rather than `line++`. The same applies to the for-loop statements increasing a counter/pointer.

tomdol (commented):
2 general comments:
- all macros used in this PR could be replaced with a type-safe features from more modern C++
- a lot of code from `parser.h` could be moved to `parser.cpp` (the implementation of ParserBase in particular)

Reviewed per request of @postrational

onnx/defs/parser.cc
```diff
@@ -0,0 +1,377 @@
...
+          PARSE_TOKEN(intval);
+          // TODO: check values are in the correct range.
+          tensorProto.add_int32_data(intval);
+          break;
+        case TensorProto::DataType::TensorProto_DataType_INT64:
+          PARSE_TOKEN(intval);
+          tensorProto.add_int64_data(intval);
+          break;
+        case TensorProto::DataType::TensorProto_DataType_UINT32:
+        case TensorProto::DataType::TensorProto_DataType_UINT64:
```
mbencer, line 148: Saving uinit64_t as int64_t can cause potential numeric overflow

onnx/defs/parser.h
gramalingam: Done. But it is a pity if a compiler cannot optimize this in this day and age :)

onnx/defs/parser.h
gramalingam: Yes, but ONNX also supports a build-option to build without exceptions. All things considered, it seems better to avoid exceptions for parser-errors.

onnx/defs/parser.cc
gramalingam: Replaced.

onnx/defs/parser.h
gramalingam: Are you referring to MakeString? I am reusing what's being used elsewhere in ONNX. The outer macro was just to hide whether exceptions or status is being used to handle errors. The original implementation did use exceptions, but was switched to use Status due to the overall ONNX goals.

tomdol: No, I meant the whole outer macro. Inside it has this construct with do..while(0), which I've seen many years ago and it was used to adress some specific problem but in a really hacky way. Can this just become a variadic template that builds an error message?

gramalingam: Done.

onnx/defs/parser.h
```diff
@@ -0,0 +1,377 @@
...
+
+  Status Parse(AttrList& attrlist);
+
+  Status Parse(NodeProto& node);
+
+  Status Parse(NodeList& nodelist);
+
+  Status Parse(GraphProto& graph);
+
+  Status Parse(ModelProto& model);
```
postrational, line 357: I'm still not sure why we don't hide all of these `Parse(...Proto&)` methods as `private`

gramalingam: (a) My experience has been that developers construct the different kinds of objects (such as a TensorShapeProto, or TypeProto) frequently and a convenient way to construct them is useful.
(b) Some of the function-definitions require constructing a graph in which some parts are dependent on context-information (like an input type T); the flexibility exposed by the above interface will make this easier.
(c) A more minor point is that it is convenient to have unit-test-cases that test parsing of the different components. (We could possibly address these using friend-classes, but given the other reasons, the above seems simpler.)
In summary, I think the advantages of the above outweigh the disadvantages (if any). (Of course, we can revisit this later, but unless we expose this functionality, there is no way for users to test it or try it out and give feedback.)

postrational (approved):

postrational:
We'll need a review from someone in `onnx/sig-archinfra-approvers`

gramalingam:
@askhade : do you have any further feedback on this? Thanks!

askhade (approved):

PR #3195, opened 2021-01-05
CumSum-14:  add data types float16 and bfloat16

tianleiwu:
Bump onnx opset version to 14.
Add data types of float16 and bfloat16 for inputs/outputs in CumSum operator

Signed-off-by: Tianlei Wu <tlwu@microsoft.com>

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -2512,7 +2512,7 @@ ONNX_OPERATOR_SET_SCHEMA(
 
           matmulShapeInference(ctx, 0, 1);
         }));
-static const char* CumSum_ver11_doc = R"DOC(
+static const char* CumSum_ver14_doc = R"DOC(
 Performs cumulative sum of the input elements along the given axis.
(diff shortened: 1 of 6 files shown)


onnx/defs/math/defs.cc
```diff
@@ -2586,7 +2586,9 @@ ONNX_OPERATOR_SET_SCHEMA(
              "tensor(int32)",
              "tensor(int64)",
              "tensor(float)",
-             "tensor(double)"},
+             "tensor(double)",
+             "tensor(float16)",
+             "tensor(bfloat16)"},
```
gramalingam, line 2591: This looks good. Will some predefined-list like ```numeric_types_for_math_reduction_with_bfloat()``` work? That will help make sure we don't miss this op if a new type is added in the future.

gramalingam: Also, updating description string below will be good (since not all tensor types are allowed).

tianleiwu: Good suggestion. I've updated it.

gramalingam (approved):

Issue #3123, opened 2020-11-21
'NonZero' has_shape() is false


buddhapuneeth:
# Bug Report

### Is the issue related to model conversion?
No. Model is converted successfully from tf2onnx.

### Describe the bug
While reading the onnx model using C++ APIs, I am getting has_shape as false for 'NonZero' operator.

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): Mac OS Catalina
- ONNX version (*e.g. 1.7*):  1.7
- Python version: 3.6
- GCC/Compiler version (if compiling from source): pip based
- CMake version:
- Protobuf version:
- Visual Studio version (if applicable):

### Reproduction instructions
- Describe the code to reproduce the behavior.
```
onnx::ModelProto model;
ParseOnnx(input_stream, model)
const onnx::GraphProto& onnx_graph =  model.graph();
const OnnxValueInfos& value_infos = onnx_graph.value_info();
for(const auto& value_info : value_infos)
 {
     const onnx::TypeProto_Tensor& type_proto = value_info.type().tensor_type();
    if(!type_proto.has_shape())
                {
                  // error: shape is zero
                }
 }
```
- Attach the ONNX model to the issue (where applicable)

### Expected behavior
NonZero will always produces output tensor of RANK 2. It should be empty space.

### Notes

Any additional information

buddhapuneeth:
Adding sample model:
For ONNX Text file:
[where_test.txt](https://github.com/onnx/onnx/files/5592275/where_test.txt)
For Source ONNX and PB files: https://1drv.ms/u/s!Auhnl7nI-7bxgo0xWWM6NCudzN2edA?e=1Xs7XJ

How to reproduce:
```
import onnx
from onnx import helper, shape_inference
from onnx import TensorProto
model1 = onnx.load("where_test.onnx")
# Apply shape inference on the model
inferred_model1 = shape_inference.infer_shapes(model1)

# Check the model and print Y's shape information
onnx.checker.check_model(inferred_model1)
print('After shape inference, the shape info of Y is:\n{}'.format(inferred_model1.graph.value_info))
```
output
```
After shape inference, the shape info of Y is:
[name: "output_raw_output___3:0"
type {
  tensor_type {
    elem_type: 7
  }
}
]
```
Ideally it should contain shape info of `output_raw_output___3:0`

jcwchen:
Hi @buddhapuneeth,
[...]

PR #3207, opened 2021-01-11
Enable Shape Inference for NonZero operator

buddhapuneeth:
Currently Shape Inference is not there NonZero and based on the discussion [here](https://github.com/onnx/onnx/issues/3123).

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -2663,6 +2663,14 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain to all tensor types.")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           updateOutputElemType(ctx, 0, TensorProto::INT64);
+          auto* output_shape = ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+          auto& input_shape = getInputShape(ctx, 0);
+          int64_t rank = input_shape.dim_size();
+          // NonZero output is of rank 2.
+          // First dimension is same as input rank
+          output_shape->add_dim()->set_dim_value(rank);
+          // Second dimension is determined based on actual non-zero's in the input data
+          output_shape->add_dim();
         }));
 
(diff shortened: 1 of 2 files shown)


buddhapuneeth:
@jcwchen can you look into this PR? This is fix for the issue mentioned in the ticket https://github.com/onnx/onnx/issues/3123.

jcwchen:
Hi @buddhapuneeth,
Thank you for proposing this PR. Could you add a simple test in `onnx/test/shape_inference_test.py`?
@onnx/sig-operators-approvers  please review this PR. Thanks!

postrational (approved):

postrational:
Please add a test to `onnx/test/shape_inference_test.py` as @jcwchen mentioned. Otherwise this looks good to me.

onnx/defs/tensor/defs.cc
```diff
@@ -2663,6 +2663,14 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain to all tensor types.")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           updateOutputElemType(ctx, 0, TensorProto::INT64);
+          auto* output_shape = ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+          auto& input_shape = getInputShape(ctx, 0);
+          int64_t rank = input_shape.dim_size();
+          // NonZero output is of rank 2.
+          // First dimension is same as input rank
+          output_shape->add_dim()->set_dim_value(rank);
```
gramalingam, line 2671: I think that we should handle the case where the input rank is not known. Check ```if (hasInputShape(ctx,0))``` before calling ```getInputShape(ctx,0)```

gramalingam: I made the suggested change. If someone could take a look, it would be helpful, thanks!

buddhapuneeth (commented):
@gramalingam thanks for adding the check. The change looks good to me.

gramalingam:
Closing this as this is covered by PR #3364

Issue #3110, opened 2020-11-12
Remove usage of collections import due to deprecation / removal


annajung:
# Bug Report

### Describe the bug
Running `pytest` outputs the following deprecation warning for importing `collections`

```
onnx/test/checker_test.py::TestChecker::test_check_graph_topologically_sorted
  /Users/xxxxx/workspace/onnx/onnx/helper.py:339: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    is_iterable = isinstance(value, collections.Iterable)
```

To support python 3.9, onnx code that uses `import collections` must be replaced with `import collections.abc`

However, changing the import causes Linux builds to fail due to 
```
onnx/helper.py:6: error: No library stub file for standard library module 'collections.abc'
onnx/helper.py:6: note: (Stub files are from https://github.com/python/typeshed)
```

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): All for import deprecation, Linux Ubuntu-16.04 for build failure
- Python version: >= 3.3, specifically will need to make sure onnx works with >= 3.9 since `import collection` will no longer work after 3.9

### Reproduction instructions
1. To see the import deprecation warning run `pytest`
1. To see the Linux build error, change the `onnx/test/checker_test.py` as follows then, run typecheck test `python setup.py --quiet typecheck`
```diff
+ from collections.abc import Iterable
- import collections

+     is_iterable = isinstance(value, Iterable)
-     is_iterable = isinstance(value, collections.Iterable)
```

### Expected behavior
A successful build for all env with deprecation warning removed

### Notes
- PR https://github.com/onnx/onnx/pull/3106 was created to address the problem for `onnx/test/checker_test.py` file. However, that's when I realize that Linux builds fails due to `No library stub file for standard library module 'collections.abc'` error. Researching into the problem, the solution seems to be to upgrade the `mypy` version used. A quick test with the latest version of `mypy` achieved a successful build for Linux.

annajung:
I wanted to create an issue to track the following actions and gather feedback anyone has about upgrading `mypy` version. It looks like it's set to `0.600` and I haven't been able to figure out any reason as to why that is. 

todos
- Look into upgrading `mypy` version for the codebase
- Clean up all files using `import collections`

cc @askhade

annajung:
Oh and I'm happy to take on this task :)

askhade:
[...]

PR #3211, opened 2021-01-12
Update mypy version and fix deprecation warning

annajung:

This is a follow up on https://github.com/onnx/onnx/pull/3106 after a break

Running `pytest` outputs a warning for wrong use of import for `collections`. Since the current way of import will not be supported in 3.9, this PR addresses the import issue with the collections abstract base class by changing the way collections are imported. 

Based on the previous PR created, the Linux build will fail without `mypy` version change due to not being able to support the new way of import https://github.com/onnx/onnx/pull/3106#issuecomment-725691861.

This PR upgrades mypy version to 0.790 and fixes import for collections

fixes https://github.com/onnx/onnx/issues/3110

### DIFF
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -5,7 +5,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import collections
+from collections.abc import Iterable
 import numbers
 from six import text_type, integer_types, binary_type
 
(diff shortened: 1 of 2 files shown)


annajung:
Still failing on Linux builds. Not quite sure how to fix this, will leave it for someone else to pick up.

PR #3213, opened 2021-01-14
[LICENSE] add ALv2 to missed files and remove leftover MIT licenses

szha:
Signed-off-by: Sheng Zha <zhasheng@amazon.com>

While adding licenses to other repos, I discovered that in my last PR my base branch was outdated, so some files were missed. This PR addresses the missed files.

@prasanthpul

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -90,7 +90,7 @@ python setup.py typecheck
 
 # License
 
-[MIT License](https://github.com/onnx/onnx/blob/master/LICENSE)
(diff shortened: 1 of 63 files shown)


szha:
@askhade I found that there are some generated `.proto` files that need update. how do I update them?

gramalingam:
@szha : I think you need to run ```python onnx\gen_proto.py```. Now, it has some options that control the various combinations of files generated. I am not sure about all the combinations that are required, but it looks like the CI settings use ```python onnx/gen_proto.py -l``` and ```python onnx/gen_proto.py -l --ml```

szha:
All done!

gramalingam (approved):

PR #3214, opened 2021-01-14
Improve set_model_dir: introduce path utility to path join

jcwchen:
**Description**
To handle separator in different platforms, introduce path utility to join pathes. With this PR, `set_model_dir` can be more flexible. (No matter whether user gives separator in the end of the model path)

**Motivation**
Previously `set_model_dir` needs to include the separator to join the model_dir and external data file.
https://github.com/microsoft/onnxruntime/pull/6119/


onnx/common/path.cc
```diff
@@ -0,0 +1,22 @@
...
+
+// Copyright (c) ONNX Project Contributors.
+// Licensed under the MIT license.
+
+#include "onnx/common/path.h"
+#include "onnx/string_utils.h"
+
+namespace ONNX_NAMESPACE {
+
+const std::string path_join(const std::string& origin, const std::string& append) {
```
askhade, line 13: why return const string? this is a utility method to join to strings

onnx/common/path.cc
```diff
@@ -0,0 +1,22 @@
...
+    std::string new_path = origin;
+    if (new_path.length() >= k_preferred_path_separator.length() && 
+        new_path.substr(new_path.length () - k_preferred_path_separator.length(), k_preferred_path_separator.length()) != k_preferred_path_separator) {
+        new_path += k_preferred_path_separator;
```
askhade, line 17: this check can be simplified something like
if(origin.find_last_of(k_prefferred_path_separator) == origin.length() - k_preferred_path_separator.length()) {
return origin + k_preferred_path_separator + append ;
}
return origin + append

this way you can avoid creating multiple intermediate strings

onnx/common/path.cc
askhade: can this mtd be inlined?

onnx/checker.cc
```diff
@@ -130,12 +137,15 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
     for (const StringStringEntryProto& entry : tensor.external_data()) {
       if (entry.has_key() && entry.has_value() && entry.key() == "location") {
         has_location = true;
-        if (!std::ifstream(ctx.get_model_dir() + entry.value())) {
+        std::string data_path = path_join(ctx.get_model_dir(), entry.value());
```
askhade, line 140: auto?

onnx/checker.cc
```diff
@@ -130,12 +137,15 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+        // use stat to check whether the file exists
+        struct stat buffer;
+        if (stat((data_path).c_str(), &buffer) < 0) {
```
askhade, line 143: != 0 is a better check
also does this work properly on windows?

onnx/common/path.cc
jcwchen: Remove const function makes more sense. 
It cannot be inlined.

onnx/common/path.cc
jcwchen: Good idea. Updated.

onnx/checker.cc
jcwchen: Yes it works good with Windows.

onnx/common/path.cc
```diff
@@ -0,0 +1,19 @@
...
+
+// Copyright (c) ONNX Project Contributors.
+// Licensed under the MIT license.
+
+#include "onnx/common/path.h"
+
+namespace ONNX_NAMESPACE {
+
+std::string path_join(std::string origin, std::string append) {
+    if (origin.find_last_of(k_preferred_path_separator) == origin.length() - k_preferred_path_separator.length()) {
```
gramalingam, line 13: Is that supposed to be == or != ?

jcwchen: Good catch. It should be !=. Thank you.

onnx/common/path.cc
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+// Copyright (c) ONNX Project Contributors.
+// Licensed under the MIT license.
+
+#include "onnx/common/path.h"
+
+namespace ONNX_NAMESPACE {
+
+std::string path_join(std::string origin, std::string append) {
```
askhade, line 12: why are you making a copy of origin string and the one to append. This should be const ref

onnx/common/path.cc
```diff
@@ -0,0 +1,19 @@
...
+    if (origin.find_last_of(k_preferred_path_separator) == origin.length() - k_preferred_path_separator.length()) {
+        origin += k_preferred_path_separator;
```
askhade, line 14: once you make origin const ref this wont work ... you can simply return origin + path_separator + append here

onnx/common/path.cc
jcwchen: Makes sense. Updated.

askhade (approved):

gramalingam (approved):

PR #3217, opened 2021-01-15
Allow recurrent operations to be batchwise (reprise 2, signoff)

matteosal:
This has the same changes of https://github.com/onnx/onnx/pull/2922

Reopening here to avoid having to fix an endless streak of conflicts in the signoff rebase


matteosal:
@prasanthpul @askhade @wschin @postrational this took way more than it should have, but now it should be ready. Can someone have a look please?

onnx/defs/rnn/defs.cc
```diff
@@ -60,6 +68,15 @@ std::function<void(OpSchema&)> RNNDocGenerator(const char* /*name*/) {
...
         std::string("forward"));
+    schema.Attr(
+        "batch_major",
+        "The shape format of the input X and output Y. "
+        "If 0, the shapes are [seq_length, batch_size, input_size] and "
+        "[seq_length, num_directions, batch_size, hidden_size] respectively."
+        "If not 0, the shapes are [batch_size, seq_length, input_size] and "
+        "[batch_size, num_directions, seq_length, hidden_size] respectively.",
+        AttributeProto::INT,
+        static_cast<int64_t>(1));
```
gramalingam, line 79: I think that the default-value should have same behavior as previous version (giving us better compatibility). Shouldn't that be the value 0?

onnx/defs/rnn/defs.cc
```diff
@@ -60,6 +68,15 @@ std::function<void(OpSchema&)> RNNDocGenerator(const char* /*name*/) {
         "Must be one of forward (default), reverse, or bidirectional.",
         AttributeProto::STRING,
         std::string("forward"));
+    schema.Attr(
+        "batch_major",
+        "The shape format of the input X and output Y. "
+        "If 0, the shapes are [seq_length, batch_size, input_size] and "
+        "[seq_length, num_directions, batch_size, hidden_size] respectively."
+        "If not 0, the shapes are [batch_size, seq_length, input_size] and "
+        "[batch_size, num_directions, seq_length, hidden_size] respectively.",
```
gramalingam, line 77: A question about the shape of the "last" output: should batch-major affect that shape too? Should the "last" be [batch_size, num_directions, hidden_size] or [num_directions, batch_size, hidden_size]? It seems natural to expect "last" to have the same shape as the output, with seq_length omitted. So, just want to make sure that this is as expected (or as in other frameworks).

onnx/defs/rnn/defs.cc
matteosal: Right, this was a mistake. Initially the attribute was called `time_major`, so its default value did replicate the old behaviour. Then I've changed it to `batch_major` but didn't update the default here (but only in the shape inference). Fixed.

onnx/defs/rnn/defs.cc
matteosal: Right, good point. I agree to let `batch_major` control the shape of the initial/final states as well. I have implemented this change

gramalingam: Just as a sanity check: is this consistent with other frameworks that use batch_major? Or, is it irrelevant because the other frameworks encode "num_directions" differently? Just hoping anyone who has concerns with the proposed spec will speak up!

gramalingam (approved):

matteosal:
It seems there is something wrong with the CI on Linux and Mac, failing with:
```
FAILED onnx/test/test_backend_test.py::OnnxBackendNodeModelTest::test_lstm_defaults_cpu
FAILED onnx/test/test_backend_test.py::OnnxBackendNodeModelTest::test_lstm_with_initial_bias_cpu
FAILED onnx/test/test_backend_test.py::OnnxBackendNodeModelTest::test_lstm_with_peepholes_cpu
```
They all fail with a shape inference error, however on my local (Linux) machine the tests are passing, and I have verified that shape inference completes explicitly on those test models.

onnx/defs/rnn/defs.cc
gramalingam: To clarify: I find the presence of num_directions in-between seq-length and batch-size a bit odd. Of course, this was a feature of the original ONNX spec. I am trying to understand whether there was a reason for this, in terms of either making the implementation easier or compatibility with other frameworks. Is there a batch-major implementation motivating this, and if so, what is the layout most convenient for that implementation (for num-directions, specifically). I went back to look at the two preceding PRs that motivate this PR, and the original one actually proposed to put num-directions as the last dimension.

matteosal: I agree that the position of `num_directions` is odd. However, my goal here is only to introduce a setting to swap batch and sequence dimensions. Moving the position of  `num_directions` like in the original PR without a flag controlling it was a bigger change and not crucial to our needs at Wolfram Research, so I felt like leaving it out to focus on `batch_major`.

gramalingam: Just to clarify: we have two cases. When batch_major=0, we have the original layout, where it makes sense to retain the original specification for backward compatibility (and I am not proposing to change this at all). But when batch_major=1, this is entirely new, so I think it makes sense to make the layout as natural for this one as possible. So, just looking for any feedback from any potential user/implementor of this new layout as to what is preferred. Another possibility would be to simply rename the attribute "batch_major" to be "layout" (an enumeration instead of a boolean). This would naturally allow other layouts to be added in the future if desired, by allowing values beyond 0 and 1. Would that make sense? Thanks very much for your thoughts/feedback.

matteosal: I went looking for how tensorflow and pytorch do it.

Tensorflow has a bidirectional wrapper in the Keras module (https://www.tensorflow.org/api_docs/python/tf/keras/layers/Bidirectional) which is strictly batch-major and produces a default shape `[batch, len, num_dir * hidden]`. They have a `merge_mode` parameter (default concat) that can combine the two directions directly by e.g. summing them thus, producing `[batch, len, hidden]`. This is also what we do at Wolfram. They also have `merge_mode = None` which produces a list of two tensors of shape `[batch, len, hidden]`, one for the forward and one for the backward. Thus Tensorflow doesn't seem to have a `num_directions` at all.

Pytorch (https://pytorch.org/docs/stable/generated/torch.nn.LSTM.html) is time-major by default but can switch to batch-major. The two directions are concatenated by default, thus producing `[len, batch, num_dir * hidden]` (or `[batch, len, num_dir * hidden]`). They also have an "unpacked" mode which produces `[len, batch, hidden, num_dir]` (or `[batch, len, hidden, num_dir]`).

So they are all treat `num_dir` pretty differently from ONNX.

gramalingam: I think that suggests [batch, len, num_dir, hidden] would be a better choice. The difference between [batch, len, num_dir, hidden] and [batch, len, num_dir * hidden] is minimal since it does not change the order of data. We can use an inexpensive reshape to switch between the two (without copying data). (I guess for a backend implementation that supports strides in the tensor implementation, most are equivalent; but there are some cases where consecutive data, is helpful.)

gramalingam: [batch, len, num_dir, hidden], however, explicitly indicates the ordering of data in the last two dimensions, and seems preferable to [batch, len, num_dir * hidden].

gramalingam: My suggestion is to (a) use an attribute-name "layout" (to indicate that the layout is changed), and (b) Preferably use [batch, len, num_dir, hidden], assuming it works better for your export. It does impact what happens (or is assumed) by the subsequent nodes, right? So, it must impact the conversion-to-onnx logic.

matteosal: Yes it does impact the conversion but only insofar as adding the right transposition/reshaping operators before and after the recurent op.

Anyway, I found your `layout` suggestion to be good and I have implemented it. The `batch_major` setup can be recovered by reverting the last commit, and I'm fine with both solutions. Now I'd like a final decision to be taken so that we can finally merge this PR.

gramalingam: This looks good to me! I prefer the current version. There is an ONNX operator SIG meeting on thursday (day after tomorrow), in case anyone else has any other opinion to express. I hope we can merge it by thursday!

PR #3218, opened 2021-01-16
Fix find_package PythonLibs not found on some platforms: make it not required

jcwchen:
**Description**
Make find_package PythonLibs not **required.**

**Motivation**
This PR will bring some build failures on some platforms: https://github.com/onnx/onnx/pull/2961
On some platforms, like Android or iOS, CMake's find_package cannot find PythonLibs. It will cause build error. However, CMake does not need that and can still use Python normally in these environments. 
https://github.com/microsoft/onnxruntime/pull/6362


daquexian:
Thanks! When will this pr be merged? :) The same bug affects onnx optimizer

daquexian:
About #2961 there is another bug, I submitted a pr at https://github.com/onnx/onnx/pull/3221

jcwchen:
> Thanks! When will this pr be merged? :) The same bug affects onnx optimizer

Hopefully it will be merged next week. Thanks!

askhade (approved):

PR #3220, opened 2021-01-17
(re-submit) update ir for subgraph and optimizer, add some const methods

daquexian:
Fix some comments in https://github.com/onnx/onnx/pull/3131 and re-submit

@askhade @jcwchen 

---------------------

1. Add forSelfAndEachSubGraph and forEachNode to iterate over every (sub)graph and node. The latter method will be used in onnx optimizer.
2. Update uses() to include the uses in subgraph. It fixes some optimizer bugs about subgraph. It may not be the optimal way, as I'm not so familiar with the design of the existing ir. Any advice is welcomed.
3. Update replaceAllUsesWith to propagate sizes, data_type and unique_name if needed.
4. Fix the bug that opset version is not set in subgraph.
5. Add some missing const methods
6. (newly added) Update ir_pb_converter for https://github.com/microsoft/onnxruntime/issues/6370


daquexian:
I also update ir_pb_converter for https://github.com/microsoft/onnxruntime/issues/6370

onnx/test/version_converter_test.py
```diff
@@ -28,6 +28,9 @@ def _converted(
         # print(type(orig_model))
         converted_model = onnx.version_converter.convert_version(orig_model,
                 target_version)
+        print(str(orig_model))
+        print("---------")
+        print(str(converted_model))
```
jcwchen, line 33: nit: Do you keep these print functions intentionally for debugging? Thank you for the resubmitting.

daquexian: @jcwchen Thanks! I forgot to remove them.

jcwchen (approved):

askhade (approved):

PR #3221, opened 2021-01-17
Fix protoc gen mypy wrapper

daquexian:
It is incorrect to use `$PWD` and `%0` in these wrapper scripts. `$PWD` and `%0` are the directory where the user executes cmake, so they will be wrong in some cases (e.g. onnx is used as a submodule of another project)


jcwchen (approved):
LGTM. Thank you for catching this!

askhade (approved):

PR #3222, opened 2021-01-18
Add argument to specify only update related models for backend generate data

jcwchen:
**Description**
To only generate related backend data in ONNX, add an argument `node` to specify the node for updating. In that case, generating data will only update the models which are using this node.

**Motivation**
https://github.com/onnx/onnx/pull/3120 Previous PR tried to solve the problem that the original generating data will bring a lot of unrelated changes because it behaves differently on different platforms. However, to solve that, it will update a lot of existing models and input/output, which might break something potentially. Alternatively, simply using an argument to generate data can solve this problem as well. So I propose this PR for it.


onnx/backend/test/case/node/__init__.py
```diff
@@ -202,3 +202,15 @@ def collect_testcases():  # type: () -> List[TestCase]
     '''
     import_recursive(sys.modules[__name__])
     return _NodeTestCases
+
+
+def collect_specific_testcases(operator_name):  # type: (Text) -> List[TestCase]
```
askhade, line 207: nit: change name to collect_testcases_by_operator ?

onnx/backend/test/cmd_tools.py
```diff
@@ -80,6 +86,8 @@ def parse_args():  # type: () -> argparse.Namespace
     subparser = subparsers.add_parser('generate-data', help='convert testcases to test data')
     subparser.add_argument('-o', '--output', default=DATA_DIR,
                            help='output directory (default: %(default)s)')
+    subparser.add_argument('-n', '--node', default=None,
+                           help='specify the node to exclude other testcases (default: %(default)s)')
```
askhade, line 90: nit: we want the user to enter operator name which is referred to as op_type in rest of the code and more importantly here (https://github.com/onnx/onnx/blob/master/onnx/onnx-operators.proto#L101) 
so change from node to op_type to match the naming convention used everywhere else...

askhade: specify the node to exclude other testcases -> generates test data only for the specified op_type

onnx/backend/test/case/node/__init__.py
```diff
@@ -202,3 +202,15 @@ def collect_testcases():  # type: () -> List[TestCase]
...
+    '''Collect node test cases which include specific operator
+    '''
+    import_recursive(sys.modules[__name__])
+    # only keep those tests related to this operator
+    specific_testcases = []
+    for testcase in _NodeTestCases:
+        if testcase.model.graph.node[0].op_type == operator_name:
+            specific_testcases.append(testcase)
```
askhade, line 215: instead of creating all the test cases and then picking up the ones for a specific operator can you instead create the test cases only for the requested operator... basically in expect method if the node.op_type does not match the request op_type don't create the test data

onnx/backend/test/cmd_tools.py
gramalingam: Agree with the suggestion to rename the option to indicate it specifies an op name (not a node). Anything like "op_type" or "op" or "op_name" would be better than "node".

jcwchen: I changed to use` --op_type`, but I don't know what the appropriate argument abbreviation is since `-o` is occupied... I chose '-t' eventually. Please let me know if you have better idea. Thanks.

onnx/backend/test/case/utils.py
```diff
@@ -31,3 +31,19 @@ def import_recursive(package):  # type: (ModuleType) -> None
...
+    Takes a package and imports specified modules underneath it
+    """
+    pkg_dir = None  # type: Optional[List[str]]
+    pkg_dir = package.__path__  # type: ignore
+    module_location = package.__name__
+    for (_module_loader, name, ispkg) in pkgutil.iter_modules(pkg_dir):
+        module_name = "{}.{}".format(module_location, name)  # Module/package
+        # only import if op_type is specified and in python test filename
+        if op_type in name:
+            module = importlib.import_module(module_name)
```
jcwchen, line 47: Here we only have the module name (onnx.backend.test.case.node.XXX.py) for filter. And the test python file name might not include op_type name...

onnx/backend/test/case/node/__init__.py
jcwchen: Are you saying extending the arguments for expect to filter op_type? If so, do we need to change a lot of expect functions to add an extra argument?

Actually these testcases are created by `import_module`. So another approach might be changing `import_recursive`, but the test python script name might not include the op_type name. Please check my comments below. Thanks.

jcwchen: Maybe we can introduce a global variable in expect function to filter. I will try it.

askhade: why cant we filter in expect?
The problem with filter while importing the module is we have to make an assumption that python file name will contain op_type... this is very fragile. 

There is no need to change the signature of expect mtd. expect takes in node as the first argument you can get op_type from node as node.op_type and if it does not match the provided op_type then you can filter it.

jcwchen: I think without changing the function `expect`'s interface, we cannot get the given op_type in the `expect` function. op_type needs to be given while calling `expect`.

jcwchen: Solved. Thanks for the proposal!

onnx/backend/test/case/node/__init__.py
```diff
@@ -19,6 +19,7 @@
 
 
 _NodeTestCases = []
+TARGET_OP_TYPE = ""
```
askhade, line 22: rename: _TargetOpType?

onnx/backend/test/cmd_tools.py
```diff
@@ -80,6 +86,8 @@ def parse_args():  # type: () -> argparse.Namespace
     subparser = subparsers.add_parser('generate-data', help='convert testcases to test data')
     subparser.add_argument('-o', '--output', default=DATA_DIR,
                            help='output directory (default: %(default)s)')
+    subparser.add_argument('-t', '--op_type', default=None,
+                           help='specify the node (lowercase) to exclude other testcases -> generates test data only for the specified op_type')
```
askhade, line 90: reword to something like -> op_type for test case generation. (generates test data for the specified op_type only.)

askhade: why lowercase? I dont think operator names are lower case Add is valid op_type... for simplicity you can convert both the provided op_type and node.op_type to lower case before comparison... or maybe python has a string comparison api which ignores case

jcwchen: Good catch. I use lowercase for these two strings to achieve case insensitivity.

askhade (approved):

onnx/backend/test/cmd_tools.py
gramalingam: Why do we want case-insensitivity? ONNX does not specify operator names are case-insensitive, does it? In which case, it seems better to not change the case at all, and expect the user to specify same case as ONNX.

jcwchen: I was afraid that users will use wrong case to update test data and fail, but they should be aware of this while uploading related test data. I agree with you that we should keep it align with case-sensitive ONNX. Just removed lower() to make it case sensitive.

gramalingam (approved):

PR #3227, opened 2021-01-19
ONNX 1.8.1 Release: exception registration and helpers for external_data

jcwchen:
**Description**
cherry-pick those 2 PRs and resolve merge conflict.
Increase the version number to 1.8.1
TODO: ~increase the version number~, release the packages

**Motivation**
Fix unexpected crash in tensorflow-onnx about exception registration: https://github.com/onnx/onnx/pull/3169
Fix bugs for helper function of external data: https://github.com/onnx/onnx/pull/3171


VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.8.0
+1.8.1
```
askhade, line 1: right now change this to 1.8.150 or something for testing. Once our test packages are tested we can change this back to 1.8.1

askhade: other places it is ok to keep it 1.8.1... since the package id is determined based on the version number in this file only

jcwchen: Thanks for the reminder. I will use 1.8.0.1 because 1.8.150 is higher than 1.8.1

askhade (approved):

gramalingam (approved):

PR #3250, opened 2021-01-31
Fix Linux typecheck error caused by ONNX's dependencies

jcwchen:
**Description**
The CI in ONNX should not take care of its dependencies' typecheck... Use `--no-site-packages` to silent those unrelated typecheck errors.

**Motivation**
Linux-CI is failing because the latest numpy (1.20.0) will fail mypy's typecheck.

### DIFF
--- a/.azure-pipelines/Linux-CI.yml
+++ b/.azure-pipelines/Linux-CI.yml
@@ -41,11 +41,12 @@ jobs:
       addToPath: true
 
   - script: |
-      python -m pip install --upgrade pip
       python -m pip install virtualenv
       python -m virtualenv py$(python.version)
       source py$(python.version)/bin/activate
+      python -m pip install --upgrade pip
       python -m pip install numpy protobuf
+      python -m pip numpy
       sudo apt-get install protobuf-compiler libprotoc-dev
       sudo apt-get install -qq -o=Dpkg::Use-Pty=0 -y --no-install-recommends dos2unix
       git submodule update --init --recursive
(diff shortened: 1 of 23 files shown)


jcwchen:
It is caused by the latest numpy version: https://github.com/numpy/numpy/releases/tag/v1.20.0 and python 3.6 was dropped so it only fails with python3.7.

.azure-pipelines/Linux-CI.yml
```diff
@@ -42,10 +42,8 @@ jobs:
 
   - script: |
       python -m pip install --upgrade pip
-      python -m pip install virtualenv
-      python -m virtualenv py$(python.version)
-      source py$(python.version)/bin/activate
       python -m pip install numpy protobuf
+      python -m pip numpy
```
askhade, line 46: I don't understand the solution

jcwchen: I think the goal would be make mypy only check onnx's directory. ONNX should not take care of its dependencies' typecheck... I removed virtualenv because if using it, the installed numpy will be produced in the current dircetory and mypy will check that as well. 

However, even if we remove it, mypy will still catch the type error from numpy in other path... Still investigation.

tools/mypy-onnx.py
```diff
@@ -11,9 +11,9 @@ def main():  # type: () -> None
     try:
         root_folder = os.path.realpath(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
         os.chdir(root_folder)
-
-        subprocess.check_call(["mypy", "."])
-        subprocess.check_call(["mypy", "--py2", "."])
+        # Use silent to prevent mypy catching other typecheck errors which are not related to ONNX itself
```
jcwchen, line 14: Changing this comment now

askhade (approved):

PR #3257, opened 2021-02-03
After ONNX 1.8.1: Update release doc due to new GitHub Action Release pipelines

jcwchen:
**Description**
Remove description about Travis-CI. Add description about the new GitHub Action release pipelines for Linux/Mac.

**Motivation**
We have moved Travis-CI release pipelines to GitHub Action since ONNX 1.8.1. Update release document for the future release.


VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.8.0
+1.8.1
```
jcwchen, line 1: Not sure whether it is needed since 1.8.1 is a patch.

askhade:
overall looks good.
1 comment - we discussed picking the right tag for test pypi wheels so that all the test package tags are < new release tag. Can you mention that somewhere near line 13.

VERSION_NUMBER
askhade: it is needed

askhade (approved):

PR #3258, opened 2021-02-04
fix for naming mismatch between fusion node's outputs and subgraph's

tilaba:
when we fuses a subgraph, we notice that  output names of  new fusion nodes  are randomly generated in a fixed range of number. Therefore a naming conflicts may occur if origin graph contains a large number of nodes , and  nodes wiring would be disturbed. The following pictures shows the result of processing a graph containing 2000 nodes with a fuse_bn_into_conv passes. 
To avoid this, we rewrite the output names of  new fusion nodes with these of fused subgraph. 

image
image

### DIFF
--- a/onnx/common/ir.h
+++ b/onnx/common/ir.h
@@ -503,6 +503,8 @@ public:
     ONNX_ASSERT(outputs().size() == n->outputs().size());
     size_t nOutputs = outputs().size();
     for(size_t i = 0; i < nOutputs; i++) {
+      std::string unique_name = outputs()[i]->uniqueName();
+      n->outputs()[i]->setUniqueName(unique_name);
       outputs()[i]->replaceAllUsesWith(n->outputs()[i]);
     }
   }


askhade:
@daquexian since this is related to optimized can you review this PR

daquexian:
@tilaba Thanks for your contribution! Could you please elaborate a bit about what the problem is? IMO `replaceAllUsesWith` will not affect the unique name. Are you using `onnx.optimizer` (in onnx/onnx repo) or `onnxoptimizer` (in onnx/optimizer) repo?

tilaba:
> @tilaba Thanks for your contribution! Could you please elaborate a bit about what the problem is? IMO `replaceAllUsesWith` does not affect the unique name. Are you using `onnx.optimizer` (in onnx/onnx repo) or `onnxoptimizer` (in onnx/optimizer repo)?

As onnx do not support gelu op, we add a new pass for onnx to fuse certain graph into a gelu op. 
when we create a new op into the graph shown in figure 1 and figure2, we notice that the op's outputs' name are randomly generated as the figure3 shown.  we find that the wiring of new gelu op is related to the  inputs' and output's name as the figure4 shown. To make sure that new gelu op can replace subgraph, it is necessary to keep output's name of mul op and subgraph.
image
**figure 1**

image
**figure 2**

image
**figure 3**

image
**figure 4**

.

daquexian:
@tilaba that's strange. It is right that the name is generated, but it is not mismatch, because `n->replaceAllUsesWith(gelu)` will make all consumers of gelu consume n now. It is not about name. I think there is another reason of your problem. Could you please check your code again? Thanks!

PR #3264, opened 2021-02-08
Added add nodes to onnx model example

Sarthak-Agrawal:
Signed-off-by: Sarthak Agrawal <saagrawal@microsoft.com>


askhade:
Thanks for adding this example! 
While this example does show how to add nodes to existing model this particular way is not generic. You can make this more generic by creating a new list of nodes (by parsing existing list and inserting the new nodes where needed) and then replacing this new list of nodes with existing in the list.

askhade:
@Sarthak-Agrawal : Do you plan to make the requested change to this PR?

PR #3265, opened 2021-02-09
enable no exception build and update exception handling for DataTypeUtils

askhade:
Adding an option for creating a no exception build for ONNX. This gives runtimes an ability to build onnx without exceptions.
This is restricted to c++ only. Exceptions are required with python bindings. When exceptions are disabled we abort at the site of first exception. 

This PR also updates error handling for DataTypeUtils. We throw an exception when any error is encountered. As stated above when exceptions are disabled we abort instead.

Also adding windows CI for testing no exception builds.


onnx/defs/schema.h
```diff
@@ -983,7 +988,7 @@ class OpSchemaRegistry final : public ISchemaRegistry {
           err << "Trying to register schema with name " << op_name
               << " (domain: " << op_domain << " version: " << ver
               << ") from file " << op_schema.file() << " line "
-              << op_schema.line() << ", but it its domain is not"
+              << op_schema.line() << ", but it's domain is not"
```
skottmckay, line 991: 'but its domain is not'.  same on line 1003

onnx/test/cpp/function_context_test.cc
```diff
@@ -122,37 +155,8 @@ TEST(FunctionAPITest, ContextDependentFunctionTest) {
...
-      .Input(0, "X", "Input tensor", "T", OpSchema::Single)
-      .Output(0, "Y", "Output tensor", "T", OpSchema::Single)
-      .TypeConstraint("T", {"tensor(float)", "tensor(double)"}, "Type of the input and output values")
-      .SetContextDependentFunctionBodyBuilder(BuildTwiceFunctionBody);
-  ONNX_NAMESPACE::OpSchemaRegistry::OpSchemaRegisterOnce unused(schema);
-  (void)unused;
-}
-
 TEST(FunctionAPITest, TypeContextTest) {
+  // Does not test registering schema twice when no exceptions are enabled.
```
skottmckay, line 159: Would maybe be simpler to just put all the 'register twice' tests and infrastructure inside an #ifndef ONNX_NO_EXCEPTIONS.

onnx/test/cpp/function_context_test.cc
```diff
@@ -91,7 +107,24 @@ void RegisterTwiceFloatSchema() {
...
+      .SetDoc("This operator returns an output tensor that is twice the input tensor.")
+      .Input(0, "X", "Input tensor", "T", OpSchema::Single)
+      .Output(0, "Y", "Output tensor", "T", OpSchema::Single)
+      .TypeConstraint("T", {"tensor(float)", "tensor(double)"}, "Type of the input and output values")
+      .SetContextDependentFunctionBodyBuilder(BuildTwiceFunctionBody);
+#ifndef ONNX_NO_EXCEPTIONS
+  ONNX_NAMESPACE::OpSchemaRegistry::OpSchemaRegisterOnce unused(schema);
+  (void)unused;
+#endif
+}
```
skottmckay, line 124: Why doesn't RegisterTwiceFloatSchema need a similar update?

askhade: This test was not written properly - RegisterTwiceFloatSchema would register the CustomFunc for the first time and then it was registered for the second time here. Anyways I updated both the test to remove a lot of common code and make the intention of the tests clear.

onnx/checker.h
```diff
@@ -34,9 +34,12 @@ class ValidationError final : public std::runtime_error {
   std::string expanded_message_;
 };
 
-#define fail_check(...)                           \
-  throw ONNX_NAMESPACE::checker::ValidationError( \
-      ONNX_NAMESPACE::MakeString(__VA_ARGS__));
+#ifdef ONNX_NO_EXCEPTIONS
+#define fail_check(...) std::cerr << ONNX_NAMESPACE::MakeString(__VA_ARGS__);\
+abort();
```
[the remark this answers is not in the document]
askhade: I updated these macros first and then decided to add ONNX_THROW and ONNX_THROW_EX . I should have gone back to these macros again and updated them ... updated them now.

onnx/common/common.h
```diff
@@ -2,7 +2,30 @@
...
  */
 
-
 #pragma once
 
 #define ONNX_UNUSED_PARAMETER(x) (void)(x)
+
+#ifdef ONNX_NO_EXCEPTIONS
+#define ONNX_THROW(...)                                 \
+  std::cerr << ONNX_NAMESPACE::MakeString(__VA_ARGS__); \
```
gramalingam, line 11: Add ```do { ... } while (false)``` ?

onnx/common/common.h
```diff
@@ -2,7 +2,30 @@
...
+    std::cerr << ex.what() << std::endl; \
+    abort();                             \
+  } while (false)
+
+#define ONNX_TRY if (true)
+#define ONNX_CATCH(x) else if (false)
+#define ONNX_HANDLE_EXCEPTION(func)
+
+#else
+#define ONNX_THROW(...) throw std::runtime_error(ONNX_NAMESPACE::MakeString(__VA_ARGS__));
```
gramalingam, line 25: Drop semicolon?

onnx/defs/generator/old.cc
```diff
@@ -101,57 +102,62 @@ ONNX_OPERATOR_SET_SCHEMA(
 
           if (nullptr != value_int) {
             // OpSchema::Verify check ensures that the attribute value has_i():
-            if (!value_int->has_i())
+            if (!value_int->has_i()) {
               fail_shape_inference("Attribute 'value_int' expect an integer.")
-                  updateOutputElemType(ctx, 0, TensorProto::INT64);
```
gramalingam, line 106: I realize this is a pre-existing problem, but minor nit: better to add semicolon after fail_shape_inference(...) here and down below

onnx/test/cpp/function_context_test.cc
```diff
@@ -181,5 +126,16 @@ TEST(FunctionAPITest, TypeContextTest) {
   check_function(fnProto, checkerCtx, lexicalScope);
 }
 
+TEST(FunctionAPITest, ContextDependentFunctionTest) {
+  ContextDependentFuntionTest();
+}
+
+// Does not test registering schema twice when no exceptions are enabled.
```
gramalingam, line 133: The two tests here are actually testing different aspects. The first tests a simple function where the parameter is of a fixed type (float), while the second tests a function where the parameter may be one of several types (float or double). The name "RegisterTwiceSchema" may be confusing, but it is meant to register a function called "CustomFunTwice". Does running these two tests cause a problem (such as same function name being registered twice)? If so, I suggest just renaming the functions as "CustomFunTwice1" and "CustomFunTwice2" or use a static boolean to ensure it is registered only once. Thanks!

CMakeLists.txt
```diff
@@ -77,6 +78,32 @@ if(NOT ONNX_NAMESPACE)
...
+    add_definitions(-D_HAS_EXCEPTIONS=0)
+  else(MSVC)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-exceptions -fno-unwind-tables -fno-asynchronous-unwind-tables")
+  endif()
+endif()
+
+if(ONNX_WERROR)
+  if(MSVC)
+    string(APPEND CMAKE_CXX_FLAGS " /WX")
+  elseif(MSVC)
```
skottmckay, line 102: Doesn't elseif need a different condition?

Not sure about the 'else(MSVC)' on line 94 as the cmake doco doesn't mention anything about repeating the condition from the 'if' in the 'else' or 'endif'. 

[https://cmake.org/cmake/help/latest/command/if.html](https://cmake.org/cmake/help/latest/command/if.html)

askhade: it was a typo I mean to add else and not elseif
else does allow condition. 
"Per legacy, the else() and endif() commands admit an optional <condition> argument. If used, it must be a verbatim repeat of the argument of the opening if command."

However to keep it consistent with the rest of the document I removed the condition

skottmckay (approved):

onnx/common/common.h
gramalingam: I am unclear why this is marked resolved. Wouldn't this macro fail in a usage like ```if (cond) ONNX_THROW(...)```?

onnx/common/common.h
gramalingam: Nit: why do we need the semicolon in the macro? Seems preferable to use same style everywhere.

gramalingam (approved):

PR #3266, opened 2021-02-09
Selective load ONNX schema by specific opset_version

jcwchen:
**Description**
- Extend current RegisterOnnxOperatorSetSchema to make it consider given opset_version
- If giving opset_version, only keep the latest opset before that version
- Introduce a status variable to record which version has been loaded. (for Runtime use)
- Introduce option ONNX_DISABLE_STATIC_REGISTRATION to disable static registration for testing selective ONNX schema loading. (Add it in no-exception CI)

**Motivation**
To reduce memory usage in Runtime, make ONNX schema load specific opset_version.


onnx/defs/operator_sets.h
```diff
@@ -965,4 +970,56 @@ inline void RegisterOnnxOperatorSetSchema() {
   RegisterOpSetSchema<OpSet_Onnx_ver14>();
 }
 
+inline void RegisterOnnxOperatorSetSchema(int target_version) {
+  // if the given version has been set, simply return
+  if (OpSchemaRegistry::GetSchemaVersion() == target_version) {
+    return;
+  }
+  OpSchemaRegistry::SetSchemaVersion(target_version);
```
jcwchen, line 978: Check here to prevent repeatedly schema registration for same opset_version. If another opset_version is given, clear the previous one and load again.

jcwchen:
This is the PR for only verifying this in ORT: https://github.com/microsoft/onnxruntime/pull/6959.

onnx/defs/operator_sets.h
```diff
@@ -951,6 +951,8 @@ class OpSet_Onnx_ver14 {
 };
 
 inline void RegisterOnnxOperatorSetSchema() {
+  // 0 means all versions of ONNX schema have been loaded
+  OpSchemaRegistry::Instance()->SetLoadedSchemaVersion(0);
```
askhade, line 955: why not set the default valur to 0 so that you dont need to do anything here

onnx/defs/schema.h
```diff
@@ -1078,6 +1092,7 @@ class OpSchemaRegistry final : public ISchemaRegistry {
    */
   static OpName_Domain_Version_Schema_Map& GetMapWithoutEnsuringRegistration();
   static OpName_Domain_Version_Schema_Map& map();
+  static int loaded_schema_version;
```
askhade, line 1095: i wonder whether this is really needed... it seems like you are only using it to determine whether the debug check should be done or not right? can you reuse the max_version\target version or some other way?

onnx/defs/schema.cc
```diff
@@ -861,13 +866,15 @@ OpName_Domain_Version_Schema_Map& OpSchemaRegistry::map() {
...
       size_t dbg_registered_schema_count = GetRegisteredSchemaCount() - dbg_initial_schema_count;
-
-      ONNX_ASSERTM(
-          dbg_registered_schema_count == ONNX_DBG_GET_COUNT_IN_OPSETS(),
-          "%u schema were exposed from operator sets and automatically placed into the static registry.  "
-          "%u were expected based on calls to registration macros. Operator set functions may need to be updated.",
-          dbg_registered_schema_count,
-          ONNX_DBG_GET_COUNT_IN_OPSETS());
+      // Don't check if the schema was partially loaded
+      if (OpSchemaRegistry::Instance()->GetLoadedSchemaVersion() > 0) {
```
askhade, line 870: did you mean if loaded_schema_version == 0? in operator_sets.h you set this to 0 when you are loading all schemas and that is when you want to run this check right

onnx/defs/operator_sets.h
jcwchen: Actually I set the default value as -1 to represent that the schema has not been loaded yet. If so, setting 0 here is required.

Regarding to a scenario in any Runtime: if Runtime wants to load specified schema during inference, it needs to know whether opset schema has been loaded or not to prevent conflict. This can be done in either ONNX or Runtime. I choose to implement here because I think this thing should be under ONNX-level and it can be reused by every Runtime frameworks.

onnx/defs/schema.h
```diff
@@ -954,7 +954,7 @@ class OpSchemaRegistry final : public ISchemaRegistry {
 
   class OpSchemaRegisterOnce final {
    public:
-    OpSchemaRegisterOnce(OpSchema& op_schema) {
+    OpSchemaRegisterOnce(OpSchema& op_schema, int max_version=0) {
```
askhade, line 957: should max_version be renames to opset_version_to_load? max_version to me seems unintuitive... max_version gives a false impression thart all schemas till a max_version should be loaded but this is not true right... you only want to load schemas valid for a particular opset version

onnx/defs/schema.h
jcwchen: For ONNX, yes it is just for DEBUG mode now, but I think it would be useful in Runtime. You can check more details in my comment above.

onnx/defs/schema.cc
jcwchen: Good catch! Should be ```if loaded_schema_version == 0``` here. Will update. Thanks

onnx/defs/schema.h
jcwchen: Agreed. I was quite confused about `max_version` and `opset_version_to_load` looks better to me. Also, I will add some comments for it to make it clear.

postrational (approved):

onnx/defs/operator_sets.h
askhade: I dont understand... what is the purpose of setting the schema version to 0 ?

onnx/defs/operator_sets.h
```diff
@@ -979,4 +981,25 @@ inline void RegisterOnnxOperatorSetSchema() {
   RegisterOpSetSchema<OpSet_Onnx_ver14>();
 }
 
+inline void RegisterOnnxOperatorSetSchema(int target_version) {
+  // Sets to record the loaded version/ prevent the full operator check in Debug mode
+  OpSchemaRegistry::Instance()->SetLoadedSchemaVersion(target_version);
+  // Keep ordered by descending; return early if target_version has been reached
```
askhade, line 987: can you mention in the comment that the calls to schema registration are required to be in descending order for this to work correctly

onnx/defs/schema.cc
```diff
@@ -18,9 +18,15 @@
 #include "onnx/common/stl_backports.h"
 
 namespace ONNX_NAMESPACE {
+// -1 means ONNX schema hasn't been loaded yet
+// 0 means all versions of ONNX schema have been loaded
+// Other positive integer means the latest ONNX schemas before that version have been loaded
```
askhade, line 23: change comment to : Other positive integer means the ONNX schemas for the specified version have been loaded

onnx/defs/schema.cc
```diff
@@ -861,13 +867,15 @@ OpName_Domain_Version_Schema_Map& OpSchemaRegistry::map() {
...
 #ifndef NDEBUG
       size_t dbg_registered_schema_count = GetRegisteredSchemaCount() - dbg_initial_schema_count;
-
-      ONNX_ASSERTM(
-          dbg_registered_schema_count == ONNX_DBG_GET_COUNT_IN_OPSETS(),
-          "%u schema were exposed from operator sets and automatically placed into the static registry.  "
-          "%u were expected based on calls to registration macros. Operator set functions may need to be updated.",
-          dbg_registered_schema_count,
-          ONNX_DBG_GET_COUNT_IN_OPSETS());
+      // Only checks if the schemas were entirely loaded
```
askhade, line 870: nit change comment to : Only check if schemas for all opset versions are loaded

onnx/defs/schema.h
```diff
@@ -974,6 +978,11 @@ class OpSchemaRegistry final : public ISchemaRegistry {
               << schema.file() << " line " << schema.line() << std::endl;
           fail_schema(err.str());
         }
+        // If opset_version_to_load != 0, only keep the latest one before specified version
```
askhade, line 981: change comment to: Return early if schema for the targeted opset version has already been loaded.

onnx/defs/schema.h
```diff
@@ -974,6 +978,11 @@ class OpSchemaRegistry final : public ISchemaRegistry {
...
+        if (opset_version_to_load != 0 && !m[op_name][op_domain].empty()) {
+          // Stops early because there is another newer OpSchema has been registered
```
askhade, line 983: you can drop this comment

askhade:
overall looks good... can you add some tests for this

gramalingam:
Just trying to understand the requirements here: if the goal is to reduce memory usage, wouldn't it be better to specify a range (min_version, max_version) to be loaded? It seems more likely that people would want to load most recent versions?

onnx/defs/schema.cc
gramalingam: Yes, I agree, I was confused about the intention.

onnx/defs/schema.cc
gramalingam: How about "Check enabled only if schemas for all opset versions are loaded"?

gramalingam (approved):

onnx/defs/operator_sets.h
jcwchen: Default is -1. Setting 0 here to let OpSchemaRegistry know that ONNX opset schema has been **fully** loaded. It would be useful for Runtime to detect whether OpSchemaRegistry has loaded ONNX schema or not to prevent loading conflict.

CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ option(ONNX_BUILD_TESTS "Build ONNX C++ APIs Tests" OFF)
 option(ONNX_USE_LITE_PROTO "Use lite protobuf instead of full." OFF)
 option(ONNXIFI_ENABLE_EXT "Enable onnxifi extensions." OFF)
 option(ONNX_DISABLE_EXCEPTIONS "Disable exception handling." OFF)
+option(ONNX_DISABLE_STATIC_REGISTRATION "Do not create static registration." OFF)
```
askhade, line 29: nit: change comment to : Disable static registration for onnx operator schemas

cmake/Utils.cmake
```diff
@@ -19,6 +19,10 @@ function(add_onnx_global_defines target)
   if(ONNX_USE_LITE_PROTO)
     target_compile_definitions(${target} PUBLIC "ONNX_USE_LITE_PROTO=1")
   endif()
+
+  if(ONNX_DISABLE_STATIC_REGISTRATION)
+    target_compile_definitions(${target} PUBLIC "__ONNX_DISABLE_STATIC_REGISTRATION=1")
```
askhade, line 24: do you need =1?

jcwchen: I followed the same logic as ONNX_ML and ONNX_USE_LITE_PROTO, but yes I think pure `__ONNX_DISABLE_STATIC_REGISTRATION` should be also workable. Let's try it.

onnx/test/cpp/function_context_test.cc
```diff
@@ -78,6 +79,9 @@ BuildFloatFunctionBody(const FunctionBodyBuildContext& ctx, const OpSchema& sche
 
 void RegisterCustomFuncFloatSchema() {
   ONNX_NAMESPACE::OpSchema schema;
+#ifdef __ONNX_DISABLE_STATIC_REGISTRATION
```
askhade, line 82: why do you need this?

onnx/defs/schema.h
```diff
@@ -1077,9 +1090,11 @@ class OpSchemaRegistry final : public ISchemaRegistry {
    * fiasco.
    */
   static OpName_Domain_Version_Schema_Map& GetMapWithoutEnsuringRegistration();
-  static OpName_Domain_Version_Schema_Map& map();
+
+  static int loaded_schema_version;
 
  public:
+  static OpName_Domain_Version_Schema_Map& map();
```
askhade, line 1097: why are you changing the accessibility from private to public?

onnx/test/cpp/function_context_test.cc
jcwchen: I was also considering whether to revert it... Just in case if someone wants to test C++ ONNX with ONNX_DISABLE_STATIC_REGISTRATION and bumps into issue here. However, even if that is the case, the whole test will still fail with schema_test.cc anyway... So I will remove it. Thanks.

onnx/test/cpp/schema_test.cc
```diff
@@ -0,0 +1,41 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <iostream>
+#include "gtest/gtest.h"
```
askhade, line 6: nit: rename file schema_registration_test.cc

onnx/test/cpp/function_context_test.cc
askhade: what would be the issue in that case?

askhade: oh you mean this test and schema_registration_test cannot go together?

onnx/test/cpp/schema_test.cc
```diff
@@ -0,0 +1,41 @@
...
+#include "onnx/defs/operator_sets.h"
+#include "onnx/defs/schema.h"
+
+using namespace ONNX_NAMESPACE;
+
+namespace ONNX_NAMESPACE {
+namespace Test {
+
+TEST(SchemaTest, RegisterCertainOpsetSchema) {
+#ifdef __ONNX_DISABLE_STATIC_REGISTRATION
```
askhade, line 16: add some comments explaining why this test is run only when static registration is disabled

jcwchen:
All updated. Thank you @askhade for all the reviews

onnx/test/cpp/function_context_test.cc
jcwchen: Yes, then it will still fail so the if condition here is useless

onnx/defs/schema.cc
```diff
@@ -844,7 +861,7 @@ OpName_Domain_Version_Schema_Map& OpSchemaRegistry::map() {
       // In debug builds, the number of schema registered in this constructor
       // is compared against the number of calls to schema registration macros.
 #ifndef NDEBUG
-      size_t dbg_initial_schema_count = GetRegisteredSchemaCount();
+      size_t dbg_initial_schema_count = OpSchemaRegistry::GetRegisteredSchemaCount();
```
askhade, line 864: this needs to move back under private right?

jcwchen: Updated. Thanks

askhade (approved):

q-ycong-p:
Hi @jcwchen , thank you very much for introducing this change.

Per my understanding, this allows a Runtime to register ONLY the specific opset schema. However, turning on `ONNX_DISABLE_STATIC_REGISTRATION` does NOT imply a "dynamic" schema registration, because the schema is not automatically "loaded" when required and "offloaded" when not in use. Registered schema holds on allocated memory until program ends. Is this correct?

If so, I want to get your thought on introducing an ONNX API to "offload" schema at runtime. The goal is to reduce MaxRSS in Runtime framework - granted that a Runtime may not need ONNX schema after model loading stage, but still needs to reserve aside RSS for it.

Do you think there's obvious reason against adding such a feature? If not, i'm happy to contribute and would appreciate any pointer. So far the "registration" looks like [it's inserting pairs](https://github.com/jcwchen/onnx/blob/739b6cecc65315bec5018e10f5451df5e6dfb7f9/onnx/defs/schema.h#L1014-L1015) into the static `OpName_Domain_Version_Schema_Map`. What complexities to be considered for removing elements from it at runtime? Thank you very much!

jcwchen:
Hi @q-ycong-p,
Good question. I did consider similar feature you mentioned (offload schema) when I was implementing this PR. I think it is good to have in ONNX-level and don't see any obvious concern. However, if you want to further apply this feature in downstream runtime, it might be more complex. For instance, by default onnxruntime registers all schemas in the beginning and it will need much code change to enable dynamic schema registration (add/offload dynamically). If your final goal is enabling this feature in onnxruntime, I would suggest you confirm this feature request is good to them (like raise a feature request in their repo) before you start. Still, I think such a feature is fairly good-to-have in ONNX. Thank you for your interest!

q-ycong-p:
Thank you @jcwchen for the detailed reply. We have a runtime inference framework that depends on ONNX as one of the supported model formats. However unlike ONNXRuntime, we do not rely on ONNX graph representation for inference, hence this will be a beneficial ONNX-level feature for us, and for other similar downstream runtimes I think.

I will experiment on my side and update further. If easy I'll open a PR, if tricky for me, I may seek community help as a feature request.

q-ycong-p:
Hi @jcwchen, I'd like to ask for your thoughts on a few items:
- Current issue: A runtime may want to register opset schema version VA when loading model A, then register opset schema version VB when loading model B, without terminating the program. From a runtime's perspective, I think it'd expect the registered ONNX schemas to be (1) bumped or downgraded, so that it's equivalent to only VB is ever registered, or (2) all schema versions in U(A, B) are registered. However, the current behavior is to throw [errors due to duplicated schema registration](https://github.com/onnx/onnx/blob/main/onnx/defs/schema.h#L1154). I think this is problematic because a runtime is given the handle to specify schema version to register, but cannot {up,down}grade that version, nor reset it to empty.

Moving forward, to expose the appropriate handles to runtime, I'm thinking (naively), in order of importance:
1. Add func `ONNX_NAMESPACE::DeregisterOnnxOperatorSetSchema` in `operator_sets.h` which empties the static `OpName_Domain_Version_Schema_Map` map, and set `opset_version_to_load=0`.
2. (good to have but I might be missing important things here) Add funcs to allow runtime to bump/downgrade opset schema verison to achieve case (1) in the above example. 

Will this seem to make sense? Any advise against pitfall is appreciated here!

q-ycong-p:
Update: #1 above is added in [OpSchema](https://github.com/onnx/onnx/pull/5221). From a runtime perspective, to achieve #2, it can deregister all ONNX opset schemas first, then register new version as needed, regardless of what was previously registered prior to the clear.

jcwchen:
> 1) bumped or downgraded, so that it's equivalent to only VB is ever registered, or (2) all schema versions in U(A, B) are registered. 

Personally I vote for 2) in ONNX-level. And we can have additional API to remove registered schema (all or specified). In principle, I prefer keeping ONNX simple and providing runtime sufficient onload/offload API for complex scenario. That is to say, we can still use multiple functions from ONNX to achieve 1) in runtime as well.

> However, the current behavior is to throw [errors due to duplicated schema registration](https://github.com/onnx/onnx/blob/main/onnx/defs/schema.h#L1154).

Yes, I think we might need another standalone registration function without that error (or enabling an option to ignore that error in the same function if feasible).

I will take a closer look at your PR and we can continue the discussion there.

q-ycong-p:
I agree that the principle is to keep ONNX simple and provide runtime sufficient onload/offload APIs. Concretely I think this means ONNX to provide APIs for runtime to:
- Register schema (all or specified): API already exists. But need to fix the duplicate schema registration error as you said, so that when a runtime register(VA) and then register(VB), ONNX should have U(VA, VB) loaded without error.
- De-register schema (all or specific): pull#5221 introduces API to naively de-register all. We can add logic to handle specific version, by implementing the opposite of the register functions.

pull#5221 simply de-registers all, for a quick turn-around to measure runtime-side memory delta on my side. I will try iterate in the PR there to expand. Also would really appreciate your input there! Ty!

PR #3267, opened 2021-02-10
Add opset converter Upsample_9 -> Resize_10

Adeykin:


askhade:
@matteosal : Can you help review this PR. Thanks!

matteosal:
> @matteosal : Can you help review this PR. Thanks!

This is a subset of my recent version converter work and will generate conflicts with my PR if merged: https://github.com/onnx/onnx/pull/3344/files#diff-7a1cc6538fe691f51f8b892e4739f6ec7e264f22e471f202adfffb7ba34c6dd0

Some technical comments:
* A `mode` attribute is requested to be present but it's actually optional: https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Upsample-9
* The number of inputs is checked for validity, but these kind of checks are usually performed by the checker and the version converter just assumes that the model is legal
* This PR contains a modification of the version converter loop to allow the replacement of the current node (because Upsample-9 must be destroyed and replaced with Resize-10). The modification is simpler than what I did in my PR, but it will produce a crash when the current node is accessed again [here](https://github.com/onnx/onnx/blob/master/onnx/version_converter/convert.cc#L94). I worked around this problem by returning the pointer to the current op from all adapters: https://github.com/onnx/onnx/pull/3344/files#diff-925153a3b28441cc23a5eb8ba88d3f112634b900a7c200d4de4490b014c1d957R92

PR #3271, opened 2021-02-12
Specify int64 to prevent different numpy behaviour on different platforms 

jcwchen:
**Description**
To prevent different numpy behaviour on different platforms, specify np.int64 for those missing.  For instance, numpy will use numpy.int32 by default on Windows and use numpy.int64 by default on Mac. Specifying the variable type should resolve this issue.

**Motivation**
Continue from https://github.com/onnx/onnx/pull/3120. Currently some backend tests will produce different data because numpy behaves differently on different platforms.


gramalingam (approved):

onnx/backend/test/case/node/mod.py
```diff
@@ -194,9 +194,9 @@ def export_mod_broadcast():  # type: () -> None
             outputs=['z'],
         )
 
-        x = np.arange(0, 30).reshape([3, 2, 5])
-        y = np.array([7])
-        z = np.mod(x, y)
+        x = np.arange(0, 30).reshape([3, 2, 5]).astype(np.int64)
```
askhade, line 197: checked in model has int32 data type... it is best if you use int32 here... this way the model itself wont change... this way 
1. the opset version for the model will remain 10 (right now it will change to 13)
2. there will be less file updates in this PR

jcwchen: Thank you for the reminder. Updated them with int32.

Also, just figure out that we don't need to update other testdata because they are int64 already. This PR only updates the test scripts to prevent issues while generating testdata on Windows.

askhade (approved):

PR #3274, opened 2021-02-12
Check uploaded/generated backend test data in release CIs

jcwchen:
**Description**
- Add a test script to `onnx.checker.check_model` all generated data
- Apply it in all Release CIs (Windows, Linux, Mac)
- ~Add ORT backend test for these backend test data~ Do it in ORT with ONNX nightly package
- Fix some test scripts to prevent numpy producing int32 on 32-bit platforms

**Motivation**
Currently test backend data are not checked therefore the uploaded models might be polluted. Also, to prevent problems like https://github.com/onnx/onnx/pull/3271, it's better to check generated backend data on all platforms.

Random thoughts:
How about other test data in the backend? Do they need to be included?


.azure-pipelines/Linux-CI.yml
```diff
@@ -94,6 +93,22 @@ jobs:
...
+      pip install -i https://test.pypi.org/simple/ ort-nightly
+      python workflow_scripts/test_generated_backend.py
+      if [ $? -ne 0 ]; then
+        echo "check uploaded backend data failed"
+        exit 1
+      fi
+      # onnx.checker all generated backend data
+      python onnx/backend/test/cmd_tools.py generate-data
+      # Not running ORT test because the latest opset_version may not be supported yet
+      python workflow_scripts/test_generated_backend.py --target onnx
```
jcwchen, line 106: Still, running ORT test for generated backend testdata can prevent future test failures in ORT... May need to figure out a way to make ORT run unsupported opset_version. Or we can run this only before every ONNX release.

.azure-pipelines/Linux-CI.yml
```diff
@@ -85,7 +85,6 @@ jobs:
       python onnx/gen_proto.py -l
       python onnx/gen_proto.py -l --ml
       python onnx/backend/test/stat_coverage.py
-      backend-test-tools generate-data
```
jcwchen, line 88: This did not generate anything... remove it because it is not worked. Instead, we only check uploaded/generated in release CIs

.github/workflows/release_linux_x86_64.yml
```diff
@@ -47,6 +47,14 @@ jobs:
...
         pytest
 
+    - name: Test backend test data
+      run: |
+        # onnx.checker all existing backend data
+        python workflow_scripts/test_generated_backend.py
+        # onnx.checker all generated backend data
+        python onnx/backend/test/cmd_tools.py generate-data
+        python workflow_scripts/test_generated_backend.py
+
```
jcwchen, line 57: TODO: This duplicate code among all release CIs will be merged into a same yaml and be reused in the future.
https://github.com/actions/starter-workflows/issues/245 Currently GitHub Action has not supported reused yaml yet.

workflow_scripts/test_generated_backend.py
```diff
@@ -0,0 +1,38 @@
...
+    count = failed_count = 0
+    for root, _, files in os.walk(directory):
+        for file in files:
+            if file.endswith('.onnx'):
+                test_dir_name = os.path.basename(os.path.normpath(root))
+                onnx_model_path = os.path.join(root, file)
+                try:
+                    model = onnx.load(onnx_model_path)
+                    # check model by ONNX checker
+                    onnx.checker.check_model(model)
```
askhade, line 21: call check_model only once after shape inference?

onnx/backend/test/case/node/topk.py
```diff
@@ -20,7 +20,7 @@ def topk_sorted_implementation(X, k, axis, largest):  # type: ignore
         sorted_values = np.flip(sorted_values, axis=axis)
     topk_sorted_indices = np.take(sorted_indices, np.arange(k), axis=axis)
     topk_sorted_values = np.take(sorted_values, np.arange(k), axis=axis)
-    return topk_sorted_values, topk_sorted_indices
+    return topk_sorted_values, np.array(topk_sorted_indices, dtype=np.int64)
```
askhade, line 23: I thought the test case changes were made in a separate PR...

askhade:
overall looks good ... just a few nit suggestions

onnx/backend/test/case/node/topk.py
jcwchen: Because this bug only happened in 32-bit environment and I found it in this PR while running 32-bit release CI.

jcwchen:
> overall looks good ... just a few nit suggestions

All updated. Thanks for the review

askhade (approved):
Please remove the CI trigger for pull requests on master before merging

onnx/backend/test/case/node/negativeloglikelihoodloss.py
```diff
@@ -27,7 +27,8 @@ def compute_negative_log_likelihood_loss(input, target, weight=None, reduction='
         # setting mode='clip' to deal with ignore_index > C or < 0 cases.
         # when the target value is > C or < 0, it doesn't matter which value we are
         # taking in gather_weight, since it will be set to 0 in the following if-block
-        gather_weight = np.take(weight, target, mode='clip')
+        # use np.int_ to make it compatible with x86 machines
+        gather_weight = np.take(weight, np.array(target, dtype=np.int_), mode='clip')
```
gramalingam, line 31: Sorry, I didn't understand this. What precision is this? Why don't we use explicit 32 or 64 bits like elsewhere in this PR?

jcwchen: 32-bit  machines somehow cannot handle numpy.int64 here so I used int_ (which means int32 for 32 machines and int64 for 64 machines). As you mentioned, using a fixed type here is better. I have changed it from int_ to int32. Thank you for the suggestion!

gramalingam (approved):

PR #3280, opened 2021-02-17
Modify existing function in external data helper

### LINKED ISSUE #3228: Updates to external data helpers
Add a new API for converting a model to external data. Today the conversion happens in 2 steps
`external_data_helper.convert_model_to_external_data(<model>,
                                                        <all_tensors_to_one_file>, <size_threshold>)
 save_model(model, output_path)
`
We want to add another api which combines the 2 steps
`
save_model_to_external_data(<model>, <output_path>, <all_tensors_to_one_file>, <size_threshold>, <convert_attributes>)
`

annajung:

Update external data helpers to modify an existing function `convert_model_to_external_data` to take a `model_file_path` and `convert_attribute` to convert attributes to external data only when specified to `true` and to save the model when `model_file_path` is specified

fixes #3228 

assign @askhade for review

### DIFF
--- a/docs/PythonAPIOverview.md
+++ b/docs/PythonAPIOverview.md
@@ -59,10 +59,9 @@ Runnable IPython notebooks:
 ## Converting and Saving an ONNX Model to External Data
 ```python
 import onnx
-from onnx.external_data_helper import covert_and_save_model_to_external_data
 
 onnx_model = ... # Your model in memory as ModelProto
-covert_and_save_model_to_external_data(onnx_model, 'path/to/save/the/model.onnx', all_tensors_to_one_file=True, location='filename', size_threshold=1024, convert_attribute=False)
+onnx.save_model(onnx_model, 'path/to/save/the/model.onnx', save_as_external_data=True, all_tensors_to_one_file=True, location='filename', size_threshold=1024, convert_attribute=False)
 # Then the onnx_model has converted raw data as external data and saved to specific directory
 ```
 
--- a/docs/PythonAPIOverview.md
+++ b/docs/PythonAPIOverview.md
@@ -33,6 +33,16 @@ load_external_data_for_model(onnx_model, 'data/directory/path/')
 # Then the onnx_model has loaded the external data from the specific directory
 ```
 
+## Converting an ONNX Model to External Data
+```python
+from onnx.external_data_helper import convert_model_to_external_data
+
+onnx_model = ... # Your model in memory as ModelProto
+convert_model_to_external_data(onnx_model, all_tensors_to_one_file=True, location='filename', size_threshold=1024, convert_attribute=False)
+# Then the onnx_model has converted raw data as external data
+# Must be followed by save
+```
+
 ## Saving an ONNX Model
 ```python
 import onnx
@@ -44,6 +54,18 @@ onnx.save(onnx_model, 'path/to/the/model.onnx')
 ```
 Runnable IPython notebooks:
 - [save_model.ipynb](https://github.com/onnx/onnx/tree/master/onnx/examples/save_model.ipynb)
+
+
+## Converting and Saving an ONNX Model to External Data
+```python
+import onnx
+from onnx.external_data_helper import covert_and_save_model_to_external_data
+
+onnx_model = ... # Your model in memory as ModelProto
+covert_and_save_model_to_external_data(onnx_model, 'path/to/save/the/model.onnx', all_tensors_to_one_file=True, location='filename', size_threshold=1024, convert_attribute=False)
+# Then the onnx_model has converted raw data as external data and saved to specific directory
+```
+
 
 ## Manipulating TensorProto and Numpy Array
 ```python
--- a/onnx/__init__.py
+++ b/onnx/__init__.py
@@ -8,7 +8,7 @@ from __future__ import unicode_literals
 import os
 
 from .onnx_cpp2py_export import ONNX_ML
-from onnx.external_data_helper import load_external_data_for_model, write_external_data_tensors
+from onnx.external_data_helper import load_external_data_for_model, write_external_data_tensors, convert_model_to_external_data
 from .onnx_pb import *  # noqa
 from .onnx_operators_pb import * # noqa
 from .onnx_data_pb import * # noqa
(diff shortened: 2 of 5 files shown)


askhade:
Update "https://github.com/onnx/onnx/blob/master/docs/PythonAPIOverview.md" with new API

onnx/external_data_helper.py
```diff
@@ -256,3 +258,42 @@ def write_external_data_tensors(model, filepath):  # type: (ModelProto, Text) ->
...
+              If not specified, will use the model name.
+    size_threshold: Threshold for size of data. Only when tensor's data is >= the size_threshold
+    it will be converted to external data. To convert every tensor with raw data to external data set size_threshold=0.
+    convert_attribute: If true, convert all tensors to external data
+                       If false, convert only non-attribute tensors to external data
+    """
+    if convert_attribute:
+        convert_model_to_external_data(model, all_tensors_to_one_file, location, size_threshold)
+    else:
+        if all_tensors_to_one_file:
```
askhade, line 283: why not update the convert_model_to_external_data to take convert_attribute... this way you dont need to duplicate code

mkbhanda: I like this idea, does not require the new function save_model_to_external_data. Could introduce a local variable that takes value _get_all_tensors(model) or _get_initializer_tensors(model) based on convert_attribute.
My sense was the original convert_model_to_external_data did the honors for all tensors .. should that not translate to a default value of True for convert_attribute?

mkbhanda (commented):
Newbie question, where does this new optional parameter for  convert_model_to_external_data need to be documented? Also where is it invoked, to change if we want the default behavior not to save attributes?

onnx/external_data_helper.py
askhade: The idea behind introducing external data was to enable models whose size exceeds the 2G max limit set by protobuf also using external data for tensors can help with sharing weights between models... however serializing attributes to external data defeats the purpose, usually attributes consume little memory and serializing them will require more memory as now you will be storing all the external data info in the tensor proto as opposed to the actual data which will most likely be an int/float/bool
This also causes issues in runtimes as attributes are required for shape inferencing etc... serializing attributes by default is a bug and we should fix that and make it optional. Ideally the option to serialize attributes is not required for the reasons stated above but we planned to add it to support any existing user scenarios which we may not be aware of.

askhade:
> Newbie question, where does this new optional parameter for convert_model_to_external_data need to be documented? Also where is it invoked, to change if we want the default behavior not to save attributes?

https://github.com/onnx/onnx/blob/master/docs/PythonAPIOverview.md and https://github.com/onnx/onnx/blob/master/docs/ExternalData.md

onnx/external_data_helper.py
annajung: Modified to remove the duplicate code and update the existing code to take the convert_attribute and file path.

docs/PythonAPIOverview.md
```diff
@@ -33,6 +33,16 @@ load_external_data_for_model(onnx_model, 'data/directory/path/')
 # Then the onnx_model has loaded the external data from the specific directory
 ```
 
+## Converting an ONNX Model to External Data
+```python
+import onnx
+from onnx.external_data_helper import convert_model_to_external_data
+
+onnx_model = ... # Your model in memory as ModelProto
+convert_model_to_external_data(onnx_model, all_tensors_to_one_file=True, location='filename', size_threshold=1024, convert_attribute=False, model_file_path='path/to/save/the/model.onnx')
```
annajung, line 42: Not sure if specifying what each parameter means is required here

onnx/external_data_helper.py
```diff
@@ -99,11 +101,11 @@ def set_external_data(tensor,  # type: TensorProto
             entry.value = str(v)
 
 
-def convert_model_to_external_data(model, all_tensors_to_one_file=True, location=None, size_threshold=1024):
-    # type: (ModelProto, bool, Optional[Text], int) -> None
+def convert_model_to_external_data(model, all_tensors_to_one_file=True, location=None, size_threshold=1024, convert_attribute=False, model_file_path=None):
```
annajung, line 104: I set `convert_attribute` to False, this will change the behavior of the end-user. But, I didn't think we wanted to set it to True

onnx/external_data_helper.py
```diff
@@ -112,22 +114,32 @@ def convert_model_to_external_data(model, all_tensors_to_one_file=True, location
...
     else:
-        for tensor in _get_all_tensors(model):
+        for tensor in tensors:
             if tensor.HasField("raw_data") and sys.getsizeof(tensor.raw_data) >= size_threshold:
                 tensor_location = tensor.name
                 if not _is_valid_filename(tensor_location):
                     tensor_location = Text(uuid.uuid1())
                 set_external_data(tensor, tensor_location)
 
+    if model_file_path:
```
annajung, line 140: I didn't make model_file_path a requirement to make sure it doesn't break the behavior of end-user

annajung: cc @askhade

annajung: thinking about it more, it makes sense to remove this and make model_file_path a requirement. Why convert if not save?

annajung: I ended up removing this check and making the file path required parameter. Let me know if this works and if we want to go back to making it optional.

annajung: After speaking with Ashwini, we decided to move the model_file_path out of the existing API to remove the requirements for the model file path.

onnx/external_data_helper.py
annajung: After speaking with Ashwini, we decided that setting the `convert_attribute` value to false is okay even if that will change the current behavior as converting attribute was not an intended behavior in the first place

onnx/external_data_helper.py
```diff
@@ -112,23 +114,50 @@ def convert_model_to_external_data(model, all_tensors_to_one_file=True, location
...
-        for tensor in _get_all_tensors(model):
+        for tensor in tensors:
             if tensor.HasField("raw_data") and sys.getsizeof(tensor.raw_data) >= size_threshold:
                 tensor_location = tensor.name
                 if not _is_valid_filename(tensor_location):
                     tensor_location = Text(uuid.uuid1())
                 set_external_data(tensor, tensor_location)
 
 
+def covert_and_save_model_to_external_data(model, model_file_path, all_tensors_to_one_file=True, location=None, size_threshold=1024, convert_attribute=False):
```
askhade, line 140: can we rename this to save_model and add it to onnx/__init__.py along with the original save_model API? So basically we will have 2 save_model APIs ... @gramalingam thoughts?

annajung: So, remove it from `external_data_helper.py` and add to `init.py`? 
I can do that, but if we rename it to `save_model`, wouldn't there be a conflict since I'm calling the existing `save_model` in the new `save_model` API?

gramalingam: I agree, it would be nicer to have them together. One possibility is to have a single API, with extra-parameters determining whether to convert data to external-data.

annajung: ack, I can add an extra parmeter `convert_data=False` in addition to parameters from the `convert_and_save_model_to_external_data` in the existing `save_model` API in the `init.py` file

We can use the `convert_data` parameter to determine if convert data to external data is needed? 
I will work on this later today, let me know if you have any concerns @askhade @gramalingam thanks!

askhade: yes makes sense...  1 suggestion rename convert_data to save_as_external_data and set defualt value to False
Something like this: 1 single save_model api : 
save_model(proto, f, format=None, save_as_external_data=False, all_tensors_to_one_file=True, location=None, size_threshold=1024, convert_attributes=False)

annajung: done, PTAL again when you can!

onnx/__init__.py
```diff
@@ -172,18 +172,29 @@ def load_tensor_from_string(s, format=None):  # type: (bytes, Optional[Any]) ->
     return _deserialize(s, TensorProto())
 
 
-def save_model(proto, f, format=None):  # type: (Union[ModelProto, bytes], Union[IO[bytes], Text], Optional[Any]) -> None
+def save_model(proto, f, format=None, save_as_external_data=False, all_tensors_to_one_file=True, location=None, size_threshold=1024, convert_attribute=False):
+    # type: (Union[ModelProto, bytes], Union[IO[bytes], Text], Optional[Any], bool, bool, Optional[Text], int, bool) -> None
     '''
-    Saves the ModelProto to the specified path.
+    Saves the ModelProto to the specified path and optionally, set tensors with raw data as external data before saving.
```
askhade, line 178: nit: set -> serialize

askhade (approved):

onnx/__init__.py
```diff
@@ -172,18 +172,29 @@ def load_tensor_from_string(s, format=None):  # type: (bytes, Optional[Any]) ->
...
 
     @params
-    proto should be a in-memory ModelProto
-    f can be a file-like object (has "write" function) or a string containing a file name
-    format is for future use
+    proto: should be a in-memory ModelProto
+    f: can be a file-like object (has "write" function) or a string containing a file name format is for future use
```
yufenglee, line 182: typo: please remove #is#

yufenglee (approved):

PR #3284, opened 2021-02-18
update model model load

askhade:


linkerzhang:
Don't quite get the motivation of making this change. Looks to me the original logic/codes are good.

linkerzhang (changes requested):
Looks to me this is moving some "utility" codes from ORT to ONNX? The original logic of loading/parsing a model is good. I'd suggest keeping this repo compact and don't introduce unnecessary utility/helper codes into it. Thanks!

PR #3288, opened 2021-02-21
remove optimizer related codes

daquexian:
Remove optimizer-related codes including `onnx.optimizer`, `onnx.utils.polish_model`, their docs and jupyter notebooks


onnx/utils.py
```diff
@@ -10,24 +10,11 @@
 
 import onnx.checker
 import onnx.helper
-import onnx.optimizer
 import onnx.shape_inference
 
 from onnx import ModelProto, NodeProto, TensorProto, ValueInfoProto
 
 
-def polish_model(model):  # type: (ModelProto) -> ModelProto
```
askhade, line 19: you can keep this method just remove optimize and the checker call after optimize...

askhade (approved):

jcwchen (approved):
LGTM. Thank you for the transfer!

onnx/utils.py
daquexian: If so this method will only run strip_doc_string and shape_inference. It makes this method not so useful. What's your opinion? Thanks!

linkerzhang (approved):
Thank you very much! Jianhao.

@houseroad please double check this change. The optimizer now is moved to a separate repo under onnx org.

gramalingam (approved):

PR #3317, opened 2021-03-08
Fix concurrent remove of .proto3.* files during build

SpaceIm:
Due to how CMakeLists in written, `gen_proto.py` might be called by generators concurrently in the same folder, and therefore the `glob.glob` expression used to delete `*.proto3.*` files might see files that this specific call of `gen_proto.py` doesn't own.

I've seen this issue randomly in builds with Unix Makefile generator on Linux.


linkerzhang (approved):

chausner:
@linkerzhang Anything speaking against merging this?

jcwchen (approved):

Issue #2728, opened 2020-04-16
request for HardSwish


haifenghan:
HardSwish is an important activation function in MobileNetV3.
I see the TensorFlow lite has added it as an operator.
Could ONNX add it into the new opset?

linkerzhang:
do you want to propose a PR for it?

linkerzhang:
and also, with this op, will you have a MobileNetV3 model in onnx format so that it may be contributed into the model zoo? thank you!

haifenghan:
OK. Let me try to generate a MobileNetV3 model in onnx format.
BTW, could you guide me about how to file a ONNX PR?

linkerzhang:
Thank you! @haifenghan .

For proposing a new OP to ONNX, please refer to https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md.

For contributing new models into ONNX model zoo,  @vinitra may offer some guidance please.

vinitra:
Hi @haifenghan! You can submit a new model by creating a PR and filling out the default pull request template in the onnx/models repository. More details on the information needed are included here: https://github.com/onnx/models/blob/master/PULL_REQUEST_TEMPLATE.md

glenn-jocher:
Any progress on adding hardswish support? Exporting from nn.Hardswish() in pytorch 1.6 produces an error now with onnx 1.7 in our YOLOv5 repo:
https://github.com/ultralytics/yolov5
```python
ONNX export failure: Exporting the operator hardswish to ONNX opset version 12 is not supported. Please open a bug to request ONNX export support for the missing operator.
```

askhade:
@ebarsoum @postrational fyi

wschin:
Is the requested `HardSwish` the same as the one in https://paperswithcode.com/method/hard-swish? If yes, I guess we can compose it using existing ONNX ops and we don't need a new op.

glenn-jocher:
@wschin yes, hardswish is essentially x * hardsigmoid(x). We are using this as a current workaround with YOLOv5 exports. The main downside is complication of the viewable model (i.e. with netron viewer).

glenn-jocher:
Current hardswish op in onnx using 4 separate operations, repeat about a hundred times in a YOLOv5 model.

```python
        return x * F.hardtanh(x + 3, 0., 6.) / 6.  # for torchscript, CoreML and ONNX
```

[...]

PR #3332, opened 2021-03-15
Add HardSwish operator 

etusien:
This PR proposes a new operator `HardSwish`. 

`HardSwish` was introduced by Andrew Howard et al. in [Searching for MobileNetV3](https://arxiv.org/abs/1905.02244v5).

`HardSwish` is a type of activation function which can be represented as the following formula:

h-swish(x) = x * ReLU6(x+3) /6

This PR proposes `HardSwish` composed by other ONNX operators:
 `HardSwish(x) = Mul(x, HardSigmoid(x, alpha=1/6, beta=1/2))`

`HardSwish` is used in the following models:
- YOLOv5
- MobileNetV3

The operator was requested in the issue https://github.com/onnx/onnx/issues/2728


onnx/defs/math/defs.cc
```diff
@@ -1029,6 +1029,48 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain input and output types to float tensors.")
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
+static const char* HardSwish_ver14_doc = R"DOC(
+HardSigmoid takes one input data (Tensor<T>) and produces one output data
+(Tensor<T>) where the HardSwish function, y = x * max(0, min(1, alpha * x + beta)),
```
gramalingam, line 1034: It would be helpful to relate this documentation with the function-body used below by adding an equivalent expression using HardSigmoid here. E.g., " = x * HardSigmoid<alpha=1/6, beta=0.5>(x)" or something like that.

onnx/defs/math/defs.cc
```diff
@@ -1029,6 +1029,48 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Constrain input and output types to float tensors.")
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
+static const char* HardSwish_ver14_doc = R"DOC(
+HardSigmoid takes one input data (Tensor<T>) and produces one output data
```
gramalingam, line 1033: HardSigmoid => HardSwish ?

onnx/backend/test/case/node/hardswish.py
```diff
@@ -0,0 +1,32 @@
...
+    def export():  # type: () -> None
+        node = onnx.helper.make_node(
+            'HardSwish',
+            inputs=['x'],
+            outputs=['y'],
+        )
+        x = np.random.randn(3, 4, 5).astype(np.float32)
+        alfa = float(1 / 6)
+        beta = 0.5
+        y = x * np.maximum(0, np.minimum(1, alfa * x + beta))
```
gramalingam, line 29: Lifting these 3 lines out as a separate function "hard_swish" would be helpful to other users as a reference implementation. (Eg., like here: https://github.com/onnx/onnx/blob/03e6c8f2ad599dda614a22ee494e982ffd0bccc7/onnx/backend/test/case/node/logsoftmax.py#L15 )

onnx/defs/math/defs.cc
etusien: Done

onnx/defs/math/defs.cc
etusien: Fixed

onnx/backend/test/case/node/hardswish.py
etusien: Done

gramalingam:
FYI: the ONNX Release 1.9 deadline is next week (https://github.com/onnx/onnx/wiki/Logistics-for-ONNX-Release-1.9 ). There will be ONNX operator SIG meeting tomorrow (10am Pacific time), in case this PR needs to be included in 1.9 and discussed. See the ONNX operators slack channel for meeting details (https://lfaifoundation.slack.com/archives/C018Y2RAY4C/p1615909210002800 )

postrational (approved):

onnx/defs/math/defs.cc
```diff
@@ -1029,6 +1029,48 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain input and output types to float tensors.")
+        .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput)
+        .FunctionBody(FunctionBodyHelper::BuildNodes({
+            // nodes: {outputs, op, inputs, attributes}
+            {{"HS_X"},
+             "HardSigmoid",
+             {"X"},
+             {MakeAttribute("alpha", float(1/6)), MakeAttribute("beta", 0.5f)}},
```
gramalingam, line 1071: I think ```float(1/6)``` evaluates to ```0``` in C++? Could change it to ```1.0f/6.0f```.

gramalingam: Is there some test to validate this? Or catch any such issue?

gramalingam: What I mean is: the "expanded" test-cases generated will help test the function-body. Running ORT or some other backend with those expanded test-cases would help validate correctness of the function body. (Unfortunately, I don't think this can be done in CI currently.)

etusien: I fixed the bug in type casting.
I'm not aware of any tests which are run on ORT to check correctness of functions' body.
However, I ran the expanded model with the bugfix on ORT by myself and it returned correct output. 
(One thing that I had to change was the opset version. Currently,  ORT provides  support for domain ai.onnx till opset 13.)

gramalingam (approved):

PR #3333, opened 2021-03-15
Updating BatchNorm to add training_mode

neginraoof:
Updating BatchNorm to add training_mode as input.


neginraoof:
@gramalingam @askhade @jcwchen @hariharans29 
Please help us review this PR. Thanks!

onnx/backend/test/case/node/batchnorm.py
```diff
@@ -12,19 +12,31 @@
...
+    bias = bias.reshape(-1, *dim_ones)
+    mean = mean.reshape(-1, *dim_ones)
+    var = var.reshape(-1, *dim_ones)
+    return s * (x - mean) / np.sqrt(var + epsilon) + bias
+
+
+def _batchnorm_training_mode(x, s, bias, mean, var, momentum=0.9, epsilon=1e-5):  # type: ignore
+    axis = np.arange(len(x.shape))
+    axis = np.delete(axis, 1)
+    axis = tuple(axis)
```
jcwchen, line 28: nit: perhaps L26-L28 can be merged into 1-2 lines.

jcwchen (commented):
Thank you for redoing this change! Backend test data looks good to me.

I don't have the history context here -- last time this change has been reverted due to a lack of verification. Have we verified it now? It would be great if it can be reviewed again from training team before check-in.

docs/Changelog.md
```diff
@@ -18587,6 +18587,94 @@ This version of the operator has been available since version 13 of the default
...
+  
+  The output and statistics are updated as follows when training_mode=True:
+  ```
+  saved_mean = ReducedMean(X, axis=all_except_channel_index)
+  saved_var =  ReducedVar(X, axis=all_except_channel_index)
+  
+  output_mean = mean * momentum + saved_mean * (1 - momentum)
+  output_var = var * momentum + saved_var * (1 - momentum)
+  
+  Y = (X - saved_mean) / sqrt(var + saved_epsilon) * scale + B
```
hariharans29, line 18612: Should this read: sqrt(saved_var + epsilon) ?

docs/Changelog.md
```diff
@@ -18587,6 +18587,94 @@ This version of the operator has been available since version 13 of the default
...
+<dt><tt>training_mode</tt> (optional, non-differentiable) : T1</dt>
+<dd>If set to true then it indicates BatchNormalization is being used for training. It is an optional value hence unless specified explicitly, it is false.</dd>
+</dl>
+
+#### Outputs (1 - 5)
+
+<dl>
+<dt><tt>Y</tt> (differentiable) : T</dt>
+<dd>The output tensor of the same shape as X</dd>
+<dt><tt>mean</tt> (optional, non-differentiable) : T</dt>
```
hariharans29, line 18659: Is it an error if optional outputs are present and training mode input flows through as false (defaulted value) ?

docs/Changelog.md
```diff
@@ -18587,6 +18587,94 @@ This version of the operator has been available since version 13 of the default
...
+  Carries out batch normalization as described in the paper
+  https://arxiv.org/abs/1502.03167. Depending on the mode it is being run,
+  There is three required inputs 'X', 'mean' and 'var', in addition to one
+  optional input 'training_mode'.
+  Note that 'mean' and 'var' are expected to be the estimated statistics in
+  inference mode (training_mode=False, default),
+  and the running statistics in training mode (traning_mode=True).
+  There are multiple cases for the number of outputs, which we list below:
+  
+  Output case #1: Y, mean, var, saved_mean, saved_var (training_mode=True)
```
hariharans29, line 18601: Clarification: Is the output output_mean and output_var or is it saved_mean and saved_var ? Looking at the terminology below, I would have thought that the output should be output_mean and output_var. 

Suggestion: Would current_mean and current_var be more appropriate than saved_mean and saved_var ?

docs/Changelog.md
```diff
@@ -18587,6 +18587,94 @@ This version of the operator has been available since version 13 of the default
 </dl>
 
 ## Version 14 of the default ONNX operator set
+### <a name="BatchNormalization-14"></a>**BatchNormalization-14**</a>
+
+  Carries out batch normalization as described in the paper
+  https://arxiv.org/abs/1502.03167. Depending on the mode it is being run,
+  There is three required inputs 'X', 'mean' and 'var', in addition to one
```
hariharans29, line 18594: Aren't inputs `B` and `scale` required too ?

onnx/test/shape_inference_test.py
```diff
@@ -3455,6 +3455,39 @@ def test_trilu_lower_one(self):  # type: () -> None
...
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.INT32, (3, 1, 5))])  # type: ignore
 
+    def test_batch_norm_train(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, (3, 4, 5, 6, 7)),
+             ('scale', TensorProto.FLOAT, (4,)),
+             ('b', TensorProto.FLOAT, (4,)),
+             ('mean', TensorProto.FLOAT, (4,)),
+             ('var', TensorProto.FLOAT, (4,)),
+             ('training_mode', TensorProto.BOOL, ())],
```
hariharans29, line 3465: How about adding one test without the optional training_mode input ?

onnx/defs/nn/defs.cc
```diff
@@ -1718,10 +1749,102 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            {"tensor(bool)"},
+            "Constrain input training_mode to boolean tensors.")
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           propagateShapeAndTypeFromFirstInput(ctx);
-          // TODO in training mode, it may be possible to infer some of
-          // the other outputs as well.
+          if (hasInputShape(ctx, 0)) {
+            propagateShapeFromInputToOutput(ctx, 0, 0);
+            auto& x_input_shape = getInputShape(ctx, 0);
+            int num_channels = 1;
```
hariharans29, line 1761: Jus thinking out aloud - is this safe to assume num channels is onehere and tin the next step, suppose the first ndex of the shape doesn't have a dim value, the initialized value will still remain one ?

onnx/defs/nn/defs.cc
```diff
@@ -1718,10 +1749,102 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            if (static_cast<int>(x_input_shape.dim_size()) > 1 &&
+                x_input_shape.dim(1).has_dim_value()) {
+              num_channels = static_cast<int>(x_input_shape.dim(1).dim_value());
+            }
+
+            if (hasInputShape(ctx, 1)) {
+              auto& scale_input_shape = getInputShape(ctx, 1);
+              if (static_cast<int>(scale_input_shape.dim_size()) != 1 ||
+                  !scale_input_shape.dim(0).has_dim_value() ||
+                  static_cast<int>(scale_input_shape.dim(0).dim_value()) !=
```
hariharans29, line 1771: Why throw if it doesn't have a dim_value ? What if it has a dim_param ? Should we check to see if it matches the dim_param of the channel index in the input shape (if that is a dim_param too) ?

onnx/defs/nn/defs.cc
```diff
@@ -1718,10 +1749,102 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
           propagateShapeAndTypeFromFirstInput(ctx);
-          // TODO in training mode, it may be possible to infer some of
-          // the other outputs as well.
+          if (hasInputShape(ctx, 0)) {
+            propagateShapeFromInputToOutput(ctx, 0, 0);
+            auto& x_input_shape = getInputShape(ctx, 0);
+            int num_channels = 1;
+            if (static_cast<int>(x_input_shape.dim_size()) > 1 &&
+                x_input_shape.dim(1).has_dim_value()) {
```
hariharans29, line 1763: Should handle dim_param for the channel index ?

onnx/test/shape_inference_test.py
```diff
@@ -3455,6 +3455,39 @@ def test_trilu_lower_one(self):  # type: () -> None
...
+            [make_node('BatchNormalization', ['x', 'scale', 'b', 'mean', 'var', 'training_mode'],
+                       ['out', 'output_mean', 'output_var', 'saved_mean', 'saved_var'])],
```
hariharans29, line 3467: How about adding more tests to test robustness of the shape inference method :

1) Shapes with dim_params (along with dim_values)
2) Shapes with no dim_values and no dim_params ?
3) Missing shapes for some inputs

docs/Changelog.md
gramalingam: +1 on more meaningful and consistent naming.

gramalingam: In addition, can we rename inputs 'mean' and 'var' to 'input_mean' and 'input_var'? Agree with Hari that "current" is more understandable than "saved"

docs/Changelog.md
gramalingam: +1

docs/Changelog.md
gramalingam: Agree we should explicitly indicate whether these extraneous outputs are allowed or not (in the inference mode). If they are allowed, we need to specify what their values will be (or whether they have an undefined value)

onnx/defs/nn/defs.cc
gramalingam: Agree, we need to handle unknown dimensions carefully. The utility function ```unifyInputDim``` handles all cases carefully and will help simplify the code. See https://github.com/onnx/onnx/blob/490895703c82f68e723a8b308369cac4ed0abc0c/onnx/defs/object_detection/defs.cc#L117 for an example of how it is used.

gramalingam: Roughly speaking, I think something like below should work (I haven't checked if the input-numbers and dim-numbers are the correct ones):
```cpp
   Dim C;
   unifyInputDim(ctx, 0, 1, C);
   unifyInputDim(ctx, 1, 0, C);
   unifyInputDim(ctx, 2, 0, C);
   etc.
```

gramalingam: Basically call ```unifyInputDim``` for every input/dim that should have the same value C

gramalingam: And you can use C to set the corresponding output dimension. This will pick up a known-value from any of the input shapes if at least one of the input shapes has a known value for that dimension.

docs/Changelog.md
```diff
@@ -18587,6 +18587,94 @@ This version of the operator has been available since version 13 of the default
...
+  
+  Y = (X - saved_mean) / sqrt(var + saved_epsilon) * scale + B
+  ```
+  
+  When training_mode=False:
+  ```
+  Y = (X - mean) / sqrt(var + epsilon) * scale + B
+  ```
+  
+  For previous (depreciated) non-spatial cases, implementors are suggested
```
gramalingam, line 18620: I don't understand this either. If the inputs are required to have some shape, can we explicitly document it here? Is it required to be 2D?

docs/Changelog.md
neginraoof: How about using runnning_mean and running_var instead of output_mean and output_var?

docs/Changelog.md
neginraoof: I think this is a note about how to get the old (deprecated) behavior for the op. There was an attribute in older versions:
'Spatial': If true, compute the mean and variance across per activation. If false, compute the mean and variance across per feature over each mini-batch.
And I think this note specifies how to achieve this behavior.

Input X can have N dimensions.
I can clarify the lines above.

docs/Changelog.md
neginraoof: @hariharans29 @gramalingam Do you have a suggestion here? Does it make sense to allow these, but fill with garbage?

onnx/test/shape_inference_test.py
neginraoof: Such a test is there in line 1239 for batch_norm.

onnx/test/shape_inference_test.py
neginraoof: What is the difference between 2 and 3?

docs/Changelog.md
hariharans29: Sure - how about "if the training_mode is false and the optional training related outputs are present, the contents of it are undefined...." ?

docs/Changelog.md
hariharans29: running_mean and running_var seems okay to me and I assume the other set of mean/variance will be current_mean and current_var ?

onnx/test/shape_inference_test.py
hariharans29: 2 -> shapes are present for inputs but they contain no dim_value or dim_param.
3 -> Shapes are missing altogether for certain inputs.

Does this make sense ?

onnx/defs/nn/defs.cc
```diff
@@ -1718,10 +1750,58 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+          unifyInputDim(ctx, 0, 1, num_channels);
+          unifyInputDim(ctx, 1, 0, num_channels);
+          unifyInputDim(ctx, 2, 0, num_channels);
+          unifyInputDim(ctx, 3, 0, num_channels);
+          unifyInputDim(ctx, 4, 0, num_channels);
+
+          if (ctx.getNumInputs() > 5 && hasInputShape(ctx, 5)) {
+            auto& mode_input_shape = getInputShape(ctx, 5);
+            // if mode is not scalar or tensor of rank 1, fail shape inference
```
gramalingam, line 1771: I think that for new op extensions, it is preferable to restrict scalars to be of rank 0.

@postrational what do you think? This is one of the "unified op interface" items. Traditionally, there has been some divergence with some ops using or allowing rank-1 tensors of size 1 for a scalar, with some requiring rank-0 tensors.

docs/Changelog.md
```diff
@@ -18587,6 +18587,95 @@ This version of the operator has been available since version 13 of the default
...
+  Output case #1: Y, running_mean, running_var, current_mean, current_var (training_mode=True)
+  Output case #2: Y (training_mode=False)
+  
+  When training_mode=False, extra outputs are undefined and the user should not depend on those.
+  The outputs are updated as follows when training_mode=True:
+  ```
+  current_mean = ReducedMean(X, axis=all_except_channel_index)
+  current_var =  ReducedVar(X, axis=all_except_channel_index)
+  
+  running_mean = mean * momentum + current_mean * (1 - momentum)
```
hariharans29, line 18610: Sorry - still not sure if this should be `input_mean` and `input_var` instead of `mean` and `var` ?

docs/Changelog.md
```diff
@@ -18587,6 +18587,95 @@ This version of the operator has been available since version 13 of the default
...
+  'input_var', in addition to one optional input 'training_mode'.
+  Note that 'input_mean' and 'input_var' are expected to be the estimated
+  statistics in inference mode (training_mode=False, default),
+  and the running statistics in training mode (training_mode=True).
+  There are multiple cases for the number of outputs, which we list below:
+  
+  Output case #1: Y, running_mean, running_var, current_mean, current_var (training_mode=True)
+  Output case #2: Y (training_mode=False)
+  
+  When training_mode=False, extra outputs are undefined and the user should not depend on those.
```
gramalingam, line 18604: A note about this as it relates to ORT implementation: if the extra outputs are specified and ignored by the ORT kernel, it is possible that even the memory for these output tensors are not allocated. This can cause subsequent errors with unallocated memory usage (depending on how the memory-planner reuses these buffers). So, it may be preferable to flag the use of extra outputs as error when training_mode=False. It would be helpful to get some input from ORT side on this. (@askhade any thoughts?)

onnx/defs/nn/defs.cc
```diff
@@ -1718,10 +1750,58 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            if (static_cast<int>(mode_input_shape.dim_size()) != 0) {
+              if (static_cast<int>(mode_input_shape.dim_size()) > 1 ||
+                  !mode_input_shape.dim(0).has_dim_value() ||
```
gramalingam, line 1774: We can't produce an error-message if the input-dimension is not statically known. (But as commented above, it would be even better to just allow rank-0 tensors here.)

docs/Changelog.md
neginraoof: Here is the ORT implementation: https://github.com/microsoft/onnxruntime/blob/8892ee4b6d343109699ab292e66c2c7a5e41925a/onnxruntime/core/providers/cpu/nn/batch_norm.h#L78

I don't think the backend needs to access memory location of such outputs in inference.

neginraoof: cc @pranavsharma

onnx/defs/nn/defs.cc
```diff
@@ -1660,13 +1682,23 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             "T",
             OpSchema::Single,
             true,
             1,
             OpSchema::Differentiable)
+        .Input(
+            5,
+            "training_mode",
+            "A 0-D tensor. If set to true then it indicates BatchNormalization is being used for training. It is an "
+            "optional value hence unless specified explicitly, it is false.",
```
SherlockNoMad, line 1696: when training_mode is true, output 3 and 4 should be populated.

docs/Changelog.md
pranav-prakash: @neginraoof While the batch_norm kernel doesn't access the optional tensors in inference mode, I believe the concern was over the allocation planner. I'm in favor of making it invalid to use the extra outputs when in inference mode, since it reduces edge-cases and is generally confusing anyway (why introduce the notion of "undefined behavior" when you can just prevent i t entirely)?

pranav-prakash: Btw the discussion in https://github.com/onnx/onnx/issues/1042 might be relevant here. I think the consensurs (at least for n=2, @gramalingam and I) was that there is not really any use-case for the optional outputs during inference. While this means that we could technically change the spec so that training vs. inference mode is determined purely based on presence of these optional outputs, I don't think changing the behavior of an op based on # of outputs is as clean as just having an attribute for it.

gramalingam: One challenge is that the "mode" is an input (as opposed to an attribute). I guess this was chosen to allow a training-engine to use the same graph/node, but change the mode dynamically? If we want this ability, then we will have to allow the multiple outputs to be specified. This also means that the issue mentioned above for allocation-planner buffer-reuse is a problem and ORT will need to solve it someway. A trivial fix would be for the kernel to ensure that the output is allocated by calling "Output" even if it's value is not initialized, but this would mean some unnecessary memory allocation in some situations.

pranav-prakash: Oh I missed that `mode` was an input instead of attribute. IIRC in the older (reverted) PR it was an attribute instead. @neginraoof can you clarify the motivation for it being an input?

> allow a training-engine to use the same graph/node, but change the mode dynamically
I don't think this is an issue because ORT already has a pre-training graph transform pass that can modify attributes or add outputs. Being able to change mode by setting a graph input doesn't seem like it provides any concrete advantage over just changing the attribute.

gramalingam: Yes, an attribute would simplify things. It would be great to clarify the motivation to have it as an input, thanks!

neginraoof: This is the revert PR: https://github.com/onnx/onnx/pull/2750/files
This one shows that training_mode has been an input for the op before, and I kept it based on the old implementations.
But from your feedback, looks like we can do 2 things:
1- Make training_mode an attribute.
2- Indicating extra outputs are invalid when training_mode=False

neginraoof:
@pranav-prakash 
Would you please also review the documentation on statistic updates? Thanks.

docs/Changelog.md
pranav-prakash: Ah I see. If it were solely up to me, I think that having it as an attribute is cleaner – and for consistency it's dropout that should be updated to move the "training_mode" to an attribute (there have been various other ops in the past that have had inputs moved as attributes, so this is not really anything unprecedented – and seems to indicate to me that having things as attributes is what schemas generally evolve towards anyway). But I'm just a user so I'll defer to the others on whatever they think is better.

docs/Changelog.md
```diff
@@ -18587,6 +18587,95 @@ This version of the operator has been available since version 13 of the default
...
+  The outputs are updated as follows when training_mode=True:
+  ```
+  current_mean = ReducedMean(X, axis=all_except_channel_index)
```
pranav-prakash, line 18607: Do you mean `ReduceMean` instead of `ReducedMean`? (same for var)

Otherwise the formulas look good to me (I especially like the use of `running_mean` which is less confusing than the `saved_mean` that the older spec had).

docs/Changelog.md
pranav-prakash: I think input `X` is expected to have at least rank 2. In the most common case input will be NCHW, but as noted in the comment to replicate the older non-spatial behavior you can flatten it to `N x CHW`.

docs/Changelog.md
neginraoof: Thanks. I can update this.
I updated the part for optional outputs, mentioning this is invalid. What do you think?

pranav-prakash: Nice! I think in addition the shape inference could be updated to enforce that no optional outputs are present if `training_mode=true`

neginraoof: Sure. Will do. 
@pranav-prakash Do you think you can help us test this spec with you ORT implementation? Just to check spec works fine.

pranav-prakash: Sure, I can do so (next week though, I'm a bit busy this week). Even with the change in spec I don't there would be be any changes in my  PR for ORT BatchNorm training though (aside from checking the newly-added `training_mode` attribute).

 (There's also a separate issue where the pre-existing CUDA BatchNorm training implementation in ORT deviates from the spec in outputting inverse-standard-deviation instead of variance – whose behavior I copied so the tests could remain the same – but that's a discussion for that PR).

gramalingam (approved):

neginraoof:
@postrational @linkerzhang 
Would you please help us review this PR? Thanks.

postrational (approved):

postrational:
This PR is approved. Please resolve merge conflicts.

neginraoof:
@postrational Thanks! Conflict is resolved.

neginraoof:
@postrational CI looks fine. Please help us merge the PR. Thanks again.

PR #3334, opened 2021-03-15
Extend dtype support for math ops

BowenBao:
Extend uint8, uint16, int8, int16 dtypes for math ops.
To support HuggingFace/transfo_xl from https://github.com/huggingface/transformers/blob/d41dd5359b4e8e4c3805427a7785401e62836de4/src/transformers/models/transfo_xl/modeling_transfo_xl.py#L937

cc @gramalingam @askhade @spandantiwari


gramalingam (approved):

askhade (changes requested):
How to handle overflow for these ops? both input and outputs are 8 bits

askhade:
@yufenglee and @tracysh FYI

onnx/backend/test/case/node/mul.py
```diff
@@ -34,6 +34,12 @@ def export():  # type: () -> None
         expect(node, inputs=[x, y], outputs=[z],
                name='test_mul')
 
+        x = np.random.randint(24, size=(3, 4, 5), dtype=np.uint8)
+        y = np.random.randint(24, size=(3, 4, 5), dtype=np.uint8)
+        z = x * y
```
askhade, line 39: not sure if this can be treated as a valid test case
1. max allowed value here is 24 and 24 * 24 will overflow. In this case different hardware can produce different results and it may not match the expected result in output
2. I think because we don't handle overflow we should not add a test case with expected result when there is an overflow

BowenBao: good point, updating the test case.

BowenBao:
cc @askhade @gramalingam PR updated for opset version. Can we merge the PR now?

gramalingam (approved):

gramalingam:
One minor comment (optional): I think, going forward, it would be good if the operator documentation included a line or two explaining the change made with respect to the previous version of the op. E.g., in this case, add a line like
```
(Opset 14 change): Extend supported types to include uint8, int8, uint16, and int16.
```

askhade (approved):

postrational (approved):

PR #3337, opened 2021-03-16
Reduce binary size for Linux and Mac package

jcwchen:
**Description**
- Instead of building C++ protobuf with default configuration, release pipelines should build it with release mode. It can remarkably reduce binary size for Linux package and slightly reduce for Mac package.

Linux package -- 18.9MB ->11.5MB
Mac package -- 12MB -> 11.3MB

Meanwhile; the total running time will be much faster as well: (Linux 59mins -> 8 mins; Mac 26mins -> 9mins)

- Deprecate Python 3.5 to prevent failure caused by Python 3.5 incompatibility

**Motivation**
ONNX 1.8.1 PyPI package becomes much larger than 1.8.0, especially Linux package. The main reason is ONNX upgraded its protobuf-compiler version from 2.6 to 3.11. We need to figure out a reasonable way to shrink protobuf dependency even though it needs to be upgraded.


.github/workflows/release_mac.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [master, rel-*]
   pull_request:
-    branches: [rel-*]
+    branches: [rel-*, master]  # TODO: remove it before merge
```
jcwchen, line 8: Mac CI fails because of this https://github.com/pypa/setuptools_scm/issues/542. Will try to reduce binary size as well.

jcwchen: ONNX decided to deprecate python 3.5 in the following 1.9. Will remove it in this PR to prevent failure caused by Python 3.5 incompatibility

askhade:
can we separate this in 2 PRs... to keep things clean.... Create a new 1 for python deprecation and keep this for binary size reduction. Thanks!

.github/workflows/release_mac.yml
```diff
@@ -39,6 +39,8 @@ jobs:
     - name: Install Python dependencies
       run: |
         python -m pip install -q --upgrade pip
+        # Latest setuptools-scm-6.0.1 (dependency of setuptools) does not support Python 3.5 anymore
```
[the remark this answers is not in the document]
jcwchen: Good catch. Removed. Thanks

jcwchen:
> can we separate this in 2 PRs... to keep things clean.... Create a new 1 for python deprecation and keep this for binary size reduction. Thanks!

Sure thing. A new PR (https://github.com/onnx/onnx/pull/3353) decouple from this PR

askhade (approved):

etusien (approved):

postrational (approved):

PR #3338, opened 2021-03-17
Move Sparse Tensor type out of ML

yuslepukhin:
Make SparseTensor type first class citizen as we now want to support it and feed inputs/outputs to the model.

We are careful to introduce the type only at this time.

Signed-off-by: Dmitri Smirnov <dmitrism@microsoft.com>


prasanthpul:
Does this require a IR version bump?

onnx/onnx.proto
```diff
@@ -671,6 +671,14 @@ message TypeProto {
     optional TypeProto value_type = 2;
   };
 
+  message SparseTensor {
+    // This field MUST NOT have the value of UNDEFINED
+    // This field MUST have a valid TensorProto.DataType value
+    // This field MUST be present for this version of the IR.
+    optional int32 elem_type = 1;
+    optional TensorShapeProto shape = 2;
```
pranavsharma, line 679: Can we also add the sparse tensor format to the type?

yuslepukhin: The problem with the sparse tensor type format is that it would be encoded into the model. However, different backends may favor different ways to implement sparsity and, thus, may have different format sparsity. In fact, depending on the libraries used, different formats may be preferable on different devices.

Thus, it becomes a runtime property. The customer will still need to know what is the optimal format under the circumstances, however, hard-coding it into the model may not have the desired effect.

gramalingam:
Hi: As discussed offline, I think this does require IR version bump. (@pranavsharma is this a concern for ORT?)

gramalingam:
@postrational : any opinion or feedback on this? No ops are being added at this stage. However, adding the type will enable experimentation with custom ops for sparse-tensors before they are promoted into ops in the ONNX standard.

pranavsharma:
> Hi: As discussed offline, I think this does require IR version bump. (@pranavsharma is this a concern for ORT?)

I don't recall updating the IR version when introducing the sequence type.

gramalingam:
```
I don't recall updating the IR version when introducing the sequence type.
```
I think it was part of either IR 6 or 7 update.

onnx/shape_inference/implementation.h
```diff
@@ -71,8 +73,15 @@ struct InferenceContextImpl : public InferenceContext {
       const auto inputDataIter = inputDataByName.find(input);
       if (inputDataIter != inputDataByName.cend()) {
         allInputData_.push_back(inputDataIter->second);
+        allInputSparseData_.push_back(nullptr);
       } else {
         allInputData_.push_back(nullptr);
+        const auto inputSparseDataIter = inputSparseDataByName.find(input);
```
askhade, line 79: This means typeAndShapeInferenceFunction will have to check the input type and based on the type call the right method of InferenceContext getInputData or getInputSparseData?

onnx/shape_inference/implementation.cc
```diff
@@ -130,12 +175,43 @@ void mergeShapesAndTypes(const TypeProto_Tensor& inferredType, TypeProto_Tensor*
...
+    return;
+  }
+
+  if (!existingType->has_shape()) {
+    // Ensure the shape is initialized. Note that this must be done
+    // even for (zero-dimensional) scalars.
+    existingType->mutable_shape();
+
+    for (int j = 0; j < inferredType.shape().dim_size(); ++j) {
+      existingType->mutable_shape()->add_dim();
```
askhade, line 193: if existingType does not have shape info then you can simply copy the shape info from inferredType there and return... no need to enter the for loop on line 197 right?

onnx/defs/shape_inference.h
```diff
@@ -61,6 +60,9 @@ struct InferenceContext {
   virtual TypeProto* getOutputType(size_t index) = 0;
   virtual GraphInferencer* getGraphAttributeInferencer(const std::string& attribute_name) = 0;
   virtual ~InferenceContext() {}
+  /// XXX: Putting at the end of vtable in case it matters?
+  /// XXX: Do we need default implementation here?
+  virtual const SparseTensorProto* getInputSparseData(size_t index) const = 0;
```
askhade, line 65: we don't need default implementation here...

onnx/shape_inference/implementation.h
yuslepukhin: > This means typeAndShapeInferenceFunction will have to check the input type and based on the type call the right method of InferenceContext getInputData or getInputSparseData?

I think, each kernel would actually query what is required. Currently, all kernels know they would be recieiving Tensors both as input and initializers. The upcoming kernels would query only Sparse Data. In general, it might be both.
The issue here and through out the code, is that we initially have two different fields, initializer and sparse_initializer. And Tensor and SparseTensor. Both of these tuples have the same high level semantic meaning but different types in proto and implementation.
So we are bound to have two different types and at times we would be checking both.

onnx/shape_inference/implementation.cc
yuslepukhin: I agree, but the existing code favors this form.

onnx/defs/shape_inference.h
yuslepukhin: Do we care to preserve the order of vtable entries?

gramalingam: Hmmm ... I don't know if there are users who want that level of compatibility or that ONNX guarantees that. But I guess this is safer anyway.

PR #3341, opened 2021-03-18
Fix release failure with Python 3.5: downgrade setuptools_scm

jcwchen:
**Description**
Downgrade setuptools_scm for now. Will remove it if ONNX deprecates Python 3.5 in the future

**Motivation**
https://github.com/pypa/setuptools_scm/issues/542 Latest setuptools does not support Python 3.5 anymore

### DIFF
--- a/.github/workflows/manylinux/entrypoint.sh
+++ b/.github/workflows/manylinux/entrypoint.sh
@@ -6,7 +6,9 @@ set -e -x
 PY_VERSION=$1
 PLAT=$2
 GITHUB_EVENT_NAME=$3
-BUILD_REQUIREMENTS='numpy==1.16.6 protobuf==3.11.3'
+# Latest setuptools-scm-6.0.1 (dependency of setuptools) does not support Python 3.5 anymore
+# TODO: Remove this line if Python 3.5 is deprecated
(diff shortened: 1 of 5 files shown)


postrational:
Python 3.5 has already reached end-of-life. Should we still support it with upcoming releases?

jcwchen:
> Python 3.5 has already reached end-of-life. Should we still support it with upcoming releases?

Thanks for bringing this up. According to the discussion in today's meeting, Python 3.5 deprecation  will be announced in ONNX 1.9 and be done after 1.9. This PR is still needed for now and it's ready for review.

askhade:
> > Python 3.5 has already reached end-of-life. Should we still support it with upcoming releases?
> 
> Thanks for bringing this up. According to the discussion in today's meeting, Python 3.5 deprecation will be announced in ONNX 1.9 and be done after 1.9. This PR is still needed for now and it's ready for review.

In today's meeting we reached the conclusion that we will deprecate 3.5 in 1.9 since keeping it will require us to fix the current issues

jcwchen:
> In today's meeting we reached the conclusion that we will deprecate 3.5 in 1.9 since keeping it will require us to fix the current issues

Sorry I misunderstood. Close this PR then

PR #3243, opened 2021-01-27
Add missing adapters + new version converter tests

matteosal:
This adds all missing version adapters from opset X to opset X+1 (upgraders) from version 7 to the latest, and also adds a new kinds of version converter tests for all operators. The idea behind the new tests is explained here: https://github.com/onnx/onnx/issues/2931

New tests are only about upgraders for now, but they can be easily modified to include testing of downgraders when those are mature enough.

PR #3344, opened 2021-03-19
New version converter tests

matteosal:
A subset of this PR which was split in two parts as requested: #3243

This part contains the new version converter tests
The other part containing the adapters is: https://github.com/onnx/onnx/pull/3343

### DIFF
--- a/onnx/test/new_version_converter_test.py
+++ b/onnx/test/new_version_converter_test.py
@@ -1,0 +1,998 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import onnx
+from onnx import helper, TensorProto, shape_inference, version_converter
+from typing import Text, List, Dict, Any, Union, Callable, Optional, cast
+import string
+import numpy as np  # type: ignore
+import unittest
+
+#####################################################################################
+# Every test creates a model containing a single operator from the lowest possible
+# opset version, upgrades it to the most recent opset version and then runs checker +
+# shape inference on the upgraded model.
+####################################################################################
+
+target_opset = onnx.defs.onnx_opset_version()
+tested_ops = []
+
+
+class TestVersionConverterNew(unittest.TestCase):
+
+    def _test_op_upgrade(
+        self,
+        op,  # type: Text
+        from_opset,  # type: int
+        input_shapes=[[3, 4, 5]],  # type: List[Union[List[Optional[int]], Text]]
+        output_shapes=[[3, 4, 5]],  # type: List[List[Optional[int]]]
+        input_types=None,  # type: Union[List[Any], None]
+        output_types=None,  # type: Union[List[Any], None]
+        initializer=[],  # type: List[Any]
+        attrs={},  # type: Dict[Text, Any]
+        seq_inputs=[],  # type: List[int]
+        seq_outputs=[]  # type: List[int]
+    ):  # type: (...) -> None
+        global tested_ops
+        tested_ops.append(op)
+
+        n_inputs = len(input_shapes)
+        letters = list(string.ascii_lowercase)[:n_inputs]
+        input_names = [
+            letter if shape != '' else '' for (letter, shape) in zip(letters, input_shapes)
+        ]
+        if input_types is None:
+            input_types = [TensorProto.FLOAT] * n_inputs
+        is_sequence = [0 if id not in seq_inputs else 1 for id in range(n_inputs)]
+        # turn empty strings into [0] to ease type analysis, even though those entries
+        # will be ignored
+        input_shapes_cast = cast(List[List[int]],
+                [[0] if isinstance(shape, str) else shape for shape in input_shapes]
+        )
+        inputs = [
+            helper.make_tensor_value_info(name, ttype, shape) if is_seq == 0
+            else helper.make_sequence_value_info(name, ttype, shape)
+            for (name, ttype, shape, is_seq)
+            in zip(input_names, input_types, input_shapes_cast, is_sequence) if name != ''
+        ]
+
+        n_outputs = len(output_shapes)
+        output_names = list(string.ascii_lowercase)[n_inputs:n_inputs + n_outputs]
+        if output_types is None:
+            output_types = [TensorProto.FLOAT] * n_outputs
+        is_sequence = [0 if id not in seq_outputs else 1 for id in range(n_outputs)]
+        output_shapes_cast = cast(List[List[int]],
+                [[0] if isinstance(shape, str) else shape for shape in output_shapes]
+        )
+        outputs = [
+            helper.make_tensor_value_info(name, ttype, shape) if is_seq == 0
+            else helper.make_sequence_value_info(name, ttype, shape)
+            for (name, ttype, shape, is_seq)
+            in zip(output_names, output_types, output_shapes_cast, is_sequence)
+        ]
+
+        node = helper.make_node(op, input_names, output_names, **attrs)
+        graph = helper.make_graph([node], op, inputs, outputs, initializer)
+        original = helper.make_model(
+            graph,
+            producer_name='test',
+            opset_imports=[helper.make_opsetid('', from_opset)]
+        )
+        onnx.checker.check_model(original)
+        shape_inference.infer_shapes(original)
+
+        converted = version_converter.convert_version(original, target_opset)
+        onnx.checker.check_model(converted)
+        shape_inference.infer_shapes(converted)
+
+    def test_Abs(self):  # type: () -> None
+        self._test_op_upgrade('Abs', 1, attrs={'consumed_inputs': [0]})
+
+    def test_Acosh(self):  # type: () -> None
+        self._test_op_upgrade('Acosh', 9)
+
+    def test_Acos(self):  # type: () -> None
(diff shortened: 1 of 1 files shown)


onnx/test/new_version_converter_test.py
```diff
@@ -0,0 +1,1005 @@
...
+        )
+        onnx.checker.check_model(original)
+        shape_inference.infer_shapes(original)
+
+        if to_opset is None:
+            to_opset = latest_opset
+
+        converted = version_converter.convert_version(original, to_opset)
+        onnx.checker.check_model(converted)
+        shape_inference.infer_shapes(converted)
```
jcwchen, line 89: To strengthen the check, shall we use a stricter shape_inference? like `shape_inference.infer_shapes(converted, strict_mode=True)`

matteosal:
@askhade I managed to write the adapters for the deprecated operators (`Scatter` and `Upsample`), removing the deprecated node and creating a new one as you suggested. But it required quite a bit of work, as I had to make all adapters return the current node and change how the converter loops through the nodes.

This PR should be fully ready now.

onnx/test/new_version_converter_test.py
matteosal: Good point, added it

matteosal:
I have encountered another issue with the version converter: https://github.com/onnx/onnx/issues/3529

It would be nice to discuss a solution for that while this PR is under review, so that a fix could be hopefully added here.

postrational:
Please add instructions about creating a converter adapter to the docs. 
Preferably here:
https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md
Or here:
https://github.com/fdwr/onnx/blob/master/docs/Versioning.md#operator-versioning

matteosal:
> Please add instructions about creating a converter adapter to the docs.
> Preferably here:
> https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md
> Or here:
> https://github.com/onnx/onnx/blob/master/docs/Versioning.md#operator-versioning

Sorry, I forgot about that. Done

matteosal:
Any news on this?

matteosal:
@askhade @postrational @gramalingam another ping on this

matteosal:
@askhade @postrational @gramalingam can we please move this forward? It's been more than 6 months since the original PR

docs/AddNewOp.md
```diff
@@ -49,10 +49,11 @@ Once the criteria of proposing new operator/function has been satisfied, you wil
     1. The testing examples will be extracted to the doc.
     2. We also generate binary data for it.
     3. Example: https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/abs.py
-5. Update the documentation and generate the test data.
+5. Add at least one automatic upgrade test for your operator in https://github.com/onnx/onnx/blob/master/onnx/test/automatic_upgrade_test.py
```
askhade, line 52: Is this step required for new operator?
In case of op updates, what happens if automatic upgrade is not possible?

askhade: I see some details are present in Versioning.md... can you add them here as well as people generally use this doc to understand all the steps involved

onnx/version_converter/adapters/adapter.h
```diff
@@ -26,7 +26,7 @@ class Adapter {
         target_version_(target_version) {
     }
 
-    virtual void adapt(std::shared_ptr<Graph> /*graph*/, Node* node) const = 0;
+    virtual Node* adapt(std::shared_ptr<Graph> /*graph*/, Node* node) const = 0;
```
askhade, line 29: does this create a new Node and returns it? Can you add comments for this mtd

onnx/version_converter/adapters/argmax_argmin_12_11.h
```diff
@@ -22,8 +22,9 @@ class ArgMaxArgMin_12_11 final : public Adapter {
 	    }
     }
 
-    void adapt(std::shared_ptr<Graph> , Node* node) const override {
+    Node* adapt(std::shared_ptr<Graph> , Node* node) const override {
 	    adapt_argmax_argmin_12_11(node);
+      return node;
```
askhade, line 27: I dont understand the point of this change.... input Node* node is not const, it is mutable then why return the same node?

docs/AddNewOp.md
matteosal: > Is this step required for new operator?

Yes because the new tests check that all operators are tested at least once (`test_ops_tested`)

> In case of op updates, what happens if automatic upgrade is not possible?

What do you mean by this? As long as the old op's functionality is a subset of the new one's, upgrade is always possible. Maybe the only case in which this got violated is `batch_normalization_13_14.h` which I added in this PR (opset 14 disabled outputs 4 and 5). Currently the automatic upgrade tests are not testing this upgrade path (they are FAR from testing all the possible upgrade paths from all possible settings for all ops anyways). In the future we can extend these new tests to also add upgrade paths which are expected to fail at some step.

> Can you add more clarification in the doc
> I see some details are present in Versioning.md... can you add them here as well as people generally use this doc to understand all the steps involved

Versioning.md says to add an adapter whenever a change is made (and don't add a test because the automatic tests will test the adapter), while this is about adding a test whenever a new op is introduced (and don't add an adapter because there's nothing to adapt), so I don't see much overlap. But I have added more information about the basic idea of the new tests.

onnx/version_converter/adapters/adapter.h
matteosal: I had to make this change because in case of deprecated operators we have to destroy the current node and replace it with another one. Since the new node is referenced via its pointer even after `adapt` is called, `adapt` has to return the pointer. But 99% of the time (when the node is preserved and only its attributes/inputs/outputs are changed) the returned pointer is the same as the argument.

onnx/version_converter/adapters/argmax_argmin_12_11.h
matteosal: See the other reply. All adapters return the same node except adapters for deprecated operators (upsample_9_10.h, scatter_10_11.h).

docs/AddNewOp.md
askhade: thanks for the clarification.
How are we enforcing\gating PRs on these tests... Will these tests run as part of python test suite and fail if someone did not add a test for new op?

onnx/version_converter/adapters/adapter.h
askhade: you have graph and mutable Node* can you not create a new node and point Node* to this new node?

docs/AddNewOp.md
matteosal: > thanks for the clarification.
> How are we enforcing\gating PRs on these tests... Will these tests run as part of python test suite and fail if someone did not add a test for new op?

Yes. The CI will fail in these 2 cases:
1) A new op is introduced but an automatic upgrade test is not added
2) An op is changed but an upgrade adapter is not added

onnx/version_converter/adapters/adapter.h
matteosal: Yes but why bother to create a new node, copy all the attributes/inputs/outputs of the old node to the new one and destroying the old node in all adapters when we can almost always modify the old node in place? That's how the adapters have always worked up to just before this PR. I have added the possibility of returning a different node only to deal with deprecated operators.

onnx/version_converter/adapters/argmax_argmin_12_11.h
askhade: what is the difficulty in changing say Upsample Node to Resize Node in place? I remember you mentioned it was not possible... cant remember what is the exact reason

matteosal: I just couldn't find a way to do it. Which doesn't mean it can't be done, but `kind_` is private in the definition of `Node` and it didn't seem designed to be changed to me. If that's possible I'll revert this change and modify the nodes in place for the deprecated ops.

gramalingam: So, I guess it is a matter of changing kind_ to be non-constant, and adding a method to SetKind in a Node?

gramalingam: It seems likely that they didn't anticipate a need to change the operation. But generalizing it to allow it seems reasonable, but I understand your concern whether it might cause some unanticipated problem.

askhade: Looking at the  (existing) code it is not clear to me what NodeKind is exactly meant for... It is not exactly similar to op_type since it includes popular attribute names and such... and while it includes most of the ops it does not include all... will need to check the usage of Kind through out version converter and optimizer code to understand whether we should add a setter for kind_. 

I am OK with the current change now. There are some fundamental changes that need to happen to version converter (updates post IRv3 .i.e allow initializers not included as inputs) soon and we can reconsider this when we take up that task.

gramalingam: I am fine with using the current API to play it safe.

gramalingam (approved):

PR #3348, opened 2021-03-21
Make checker with full_check do shape_inference with strict_mode

jcwchen:
**Description**
ONNX checker should either

1. Enable shape_inference with strict_mode when full_check enabled
2. Change full_check to strict_model for consistency and do strict shape_inference
3. Also extend a new argument called strict_mode to checker 

This PR is doing 1. Any feedback is welcome. Thank you.

**Motivation**
https://github.com/onnx/onnx/pull/3169 This brings onnx.shape_inference takes strict_mode. If strict_mode is enabled, onnx.shape_inference will further throw shape inference errors if happened. However, it did not apply to onnx.checker. onnx.checker with full_check should consider all shape_inference errors.


gramalingam (approved):

jcwchen:
> I think full_check and strict_mode serve slightly different purpose right now? Full_check controls whether type/shape inference check is invoked or not, while strict_mode indicates further how strict the type/shape checker should be. So, I don't think option 2 is a good choice. Option 1 seems okay for now. Option 3 could be considered, but I think we may also want this mode flag to become an enumeration in the future, instead of boolean. BTW, what is the default-value for strict-mode? Isn't it true by default?

Agreed. Option 1 is the best choice for now... The default value of strict_mode is False since enabling it will cause some failures from old models (e.g., pytorch-onnx CIs). The current behaviour with strict_mode=False is suspending shape_inference without throwing error if it bumps into a shape inference error, and the inferred shape will be kept.

PR #3359, opened 2021-03-25
Enable protobuf C++ arena

take-cheeze:
Ref: https://developers.google.com/protocol-buffers/docs/reference/arenas


mkbhanda:
Curious, any possible tests for this to track performance improvement?

take-cheeze:
Sorry for latency.

The main reason I want arena is to make destructor call of ModelProto faster.

I've read https://github.com/protocolbuffers/protobuf/blob/master/docs/performance.md but seems it doesn't have dramatic performance increase.
And they haven't open sourced string/bytes improvements: https://github.com/protocolbuffers/protobuf/issues/4327

I've tried some timeit micro benchmark with few binding function using `ParseProtoFromBytes` like: https://gist.github.com/take-cheeze/4a456ed04aaf7ca0bb24ff7ad978d815
Got about 10% speed up with [mask-rcnn onnx file from model zoo](https://github.com/onnx/models/tree/master/vision/object_detection_segmentation/mask-rcnn)
It will require protobuf3 which isn't supported in Ubuntu 16.04(CI failure reason)

mkbhanda:
Thank you @take-cheeze for the detailed response on potential performance improvements

linkerzhang:
Dumb question, is it doable to enable arena as a build option? as you mentioned, adding the option in current way will fail in CI.

PR #3361, opened 2021-03-25
Resize op test update to opset 13

tomdol:
The changes in this PR are about one of the Resize op tests: `test_resize_downsample_sizes_nearest_tf_half_pixel_for_nn`

The current version of this test uses a model with opset_import=11. Given [the specification of opset 11 for Resize](https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Resize-11) the model isn't entirely correct. Only one of the last two inputs of Resize should be specified but in the model both are connected to the respective graph inputs. `scales` input is not optional in opset 11 and the test relies on the values of `sizes`. The workaround in the current version of the test is that the `scales` input is connected to a graph input with a zero-dimension. There's also a corresponding input value in the test's directory containing an empty array. A similar trick was applied to the non-optional `roi` input which is supposed to be ignored in this test because of the coordinate transformation mode.

I propose the following changes to the test:
- switch the test model to opset 13 because of the updated specification which lets a correct model to be created for this test
- set both `roi` and `scales` inputs in the test model to empty strings because they do not affect this test - this way the op will be given to relevant inputs
- remove the obsolete input values `test_data_set_0/input_1.pb` and `test_data_set_0/input_1.pb`
- rename `test_data_set_0/input_3.pb` to `test_data_set_0/input_1.pb` - this way the test runner will provide 2 input values to an operator with 2 inputs


postrational (approved):

postrational:
@jcwchen could you please verify this updated test using ONNX Runtime?

jcwchen:
> @jcwchen could you please verify this updated test using ONNX Runtime?

Yes, they pass ORT backend test and onnx.checker. Thanks.

jcwchen (approved):

fdwr:
@tomdol: We see some failures after adopting this change into ONNX runtime, and I found it's because some tensors have duplicate names:

test_resize_downsample_sizes_nearest_tf_half_pixel_for_nn\test_data_set_0\

input_0.pb - Tensor "X", datatype: float32, dimensions: 1,1,4,4, (64 bytes): 1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16
input_1.pb - *__Tensor "sizes"__*, datatype: int64, dimensions: 4, (32 bytes): 1,1,3,2
input_2.pb - Tensor "scales", datatype: float32, dimensions: 0, (0 bytes)
input_3.pb - *__Tensor "sizes"__*, datatype: int64, dimensions: 4, (32 bytes): 1,1,3,2

Can this be fixed, updating the name in "input_1.pb" to "roi"? I'll just disable that test in meantime. Thanks.

tomdol:
@fdwr will do, sorry I missed that. I'll let you know when the PR is ready.

tomdol:
@fdwr I've re-checked the test and it looks correct to me. The tested model expects the "X" and "sizes" inputs to be passed in. "roi" and "scales" are set to empty strings according to the documentation to indicate that they are optional and should be disregarded for this model.

The input data in `test_resize_downsample_sizes_nearest_tf_half_pixel_for_nn/test_data_set_0` now contains 2 files instead of 4:

```
>>> onnx.TensorProto.FromString(open("input_0.pb", "rb").read())
dims: 1
dims: 1
dims: 4
dims: 4
data_type: 1
name: "X"
raw_data: "\000\000\200?\000\000\000@\000\000@@\000\000\200@\000\000\240@\000\000\300@\000\000\340@\000\000\000A\000\000\020A\000\000 A\000\0000A\000\000@A\000\000PA\000\000`A\000\000pA\000\000\200A"

>>> onnx.TensorProto.FromString(open("input_1.pb", "rb").read())
dims: 4
data_type: 7
name: "sizes"
raw_data: "\001\000\000\000\000\000\000\000\001\000\000\000\000\000\000\000\003\000\000\000\000\000\000\000\002\000\000\000\000\000\000\000"
```

The input names match the operator's input names in the model. Can you please verify on your end?

fdwr:
@tomdol : Sorry for the false alarm :(. Our test copies the new data, but the copying script left behind the existing old tensors which were deleted (as this test case was updated inline). The test passes now 👍. Thanks for your change - it was kinda silly before that the tensor was empty.

PR #3371, opened 2021-03-30
onnx_proto symbols visibility clean-up

mbencer:
The problem was described in https://github.com/onnx/onnx/issues/3319
In short, currently, there is no possibility for import `onnx_proto` symbols on Windows (when we build `onnx_proto` as shared lib).

The basic solution was introduced in https://github.com/onnx/onnx/pull/3335, but there are many doubts around this topic.
The main problem seems to be specifying `onnx_proto` visibility both in [onnx_pb.h header](https://github.com/onnx/onnx/blob/master/onnx/onnx_pb.h) and in [cmake](https://github.com/onnx/onnx/blob/478124d60eb97f5a2818e92e84456cab171b81e7/CMakeLists.txt#L379).

My proposition is unification and one place to defining symbols visibility strategy - cmake file.

`ONNX_BUILD_SHARED_LIBS` flag seems to be unuseful - a way of building is determined by `BUILD_SHARED_LIBS ` so it was removed, but please let me know if there are some scenarios to use it.


mbencer:
@askhade Do you maybe know why `ONNX_BUILD_SHARED_LIBS` is needed?
I don't see a scenario when it is needed (it seems to be equivalent to `BUILD_SHARED_LIBS` in this case)

linkerzhang (approved):

askhade:
@mbencer: Some of the cmake options are legacy and I am not sure why they were added. I need to take a thorough look at this PR before I can approve it. Will add my comments by today or tomorrow

ilyachur:
@askhade Do you have an update for this PR?

CMakeLists.txt
```diff
@@ -414,7 +429,8 @@ if(CMAKE_SYSTEM_NAME STREQUAL "AIX")
   # So, create a object library
   add_library(onnx OBJECT ${ONNX_SRCS})
 else()
-  add_library(onnx ${ONNX_SRCS})
+  # onnx target doesn't export symbols and should be built as static lib
+  add_library(onnx STATIC ${ONNX_SRCS})
```
askhade, line 433: can we keep this default meaning let BUILD_SHARED_LIBS control whether we use STATIC or SHARED

what is the problem with building shared lib when build_shared_lib is set?

CMakeLists.txt
```diff
@@ -381,18 +381,33 @@ if (MSVC)
...
-    set(ONNX_API_DEFINE "-DONNX_API=")
+    # Windows static libs don't need symbols visibility decorators
+    target_compile_definitions(onnx_proto PUBLIC "ONNX_API=")
   endif()
 else()
   # On non-Windows, hide all symbols we don't need
-  set(ONNX_API_DEFINE "-DONNX_API=__attribute__\(\(__visibility__\(\"default\"\)\)\)")
-  set_target_properties(onnx_proto PROPERTIES CXX_VISIBILITY_PRESET hidden)
-  set_target_properties(onnx_proto PROPERTIES VISIBILITY_INLINES_HIDDEN 1)
+    target_compile_definitions(onnx_proto PUBLIC "ONNX_API=__attribute__\(\(__visibility__\(\"default\"\)\)\)")
```
askhade, line 406: what is the difference between original set vs the PR version target_compile_definitions(onnx_proto ?

is set global ?

CMakeLists.txt
```diff
@@ -381,18 +381,33 @@ if (MSVC)
...
     set(EXTRA_FLAGS "")
   endif()
-  if (BUILD_SHARED_LIBS OR ONNX_BUILD_MAIN_LIB)
-    set(ONNX_API_DEFINE "-DONNX_API=__declspec(dllexport)")
+endif()
+
+# onnx_proto symbols visibility
+if(MSVC)
+  if(ONNX_BUILD_MAIN_LIB)
+  # ONNX_BUILD_MAIN_LIB can be be set if onnx_proto is static,
```
askhade, line 389: typo: be be

askhade:
FYI - 
https://github.com/protocolbuffers/protobuf/blob/master/cmake/README.md#dlls-vs-static-linking

If your project is itself a DLL intended for use by third-party software, we recommend that you do NOT expose protocol buffer objects in your library's public interface, and that you statically link protocol buffers into your library.

gramalingam:
There is a long thread relating to this in  a previous PR: please see: https://github.com/onnx/onnx/pull/1938 . Among other things, there is a comment towards the end of the thread that defining the preprocessors macros in the cmake breaks the pytorch build.

gramalingam:
I did not fully understand the context here. Is the goal to produce an ONNX dll that statically links onnx_proto as well as protobuf? Or, is it something else?

CMakeLists.txt
mbencer: The features (classes/members/functions) from onnx library don't use decorators to control visibility.
It is especially required for Windows - `__declspec`.
So the idea behind it was to block building shared `onnx` if it is not supported.

CMakeLists.txt
mbencer: Previous version set value on directory level - it will be visible in current directory scope.
In the current version, we match it only with the target which uses it (it will be visible for `onnx_proto` code and for targets which use `onnx_proto`).
So from a functional point of view, it's rather not a big difference.

mbencer:
> I did not fully understand the context here. Is the goal to produce an ONNX dll that statically links onnx_proto as well as protobuf? Or, is it something else?

The goal for us was to build: shared `protobuf`, shared `onnx_proto`, and static `onnx`.
We had two separate libraries which depend on onnx-related staff (`protobuf, `onnx_proto`) in one application. More details in: https://github.com/onnx/onnx/issues/3319

Issue #3363, opened 2021-03-26
ProtoDebugString is dangerous when ONNX_USE_LITE_PROTO is set


skottmckay:
# Bug Report

### Describe the bug
https://github.com/onnx/onnx/blob/aafa3bd4046fe7032d05581ef02d51a9338c29db/onnx/proto_utils.h#L20-L30

If using protobuf-lite, SerializeAsString is called and returns the **binary** serialization of the protobuf as a std::string. 

As ProtoDebugString can be called from the checker to create an error message, having the binary serialization of the protobuf in that message is not great (can lead to the entire error message being thrown away if the call is via python, or untold other effects if someone tries to print that string).

Values from SerializeAsString probably shouldn't be returned in an error message. One option would be to sanitize them first to only return printable characters. The below example will work on Windows (as std::isprintf asserts if 'c' is not in the range 0 to 255). There may be better approaches to fix this issue.
e.g. 
```
inline std::string ProtoDebugString(const MessageLite& proto) {
  std::string serialized = proto.SerializeAsString();
  std::replace_if(
      serialized.begin(), serialized.end(), [](std::string::value_type c) { return c < 0 || c > 255 || !std::isprint(c); }, '?');
  return serialized;
}
```

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*):  Windows 10
- ONNX version (*e.g. 1.7*):  master
- Python version: 3.7
- GCC/Compiler version (if compiling from source):
- CMake version:
- Protobuf version:
- Visual Studio version (if applicable): 2019

### Reproduction instructions
Create a model with a subgraph containing a node with an operator with an invalid attribute (e.g. RandomNormal with an empty attribute for 'shape'). This will result in the exception handler in checker.cc:check_graph adding the binary serialization of the node's protobuf to the error message.

https://github.com/onnx/onnx/blob/aafa3bd4046fe7032d05581ef02d51a9338c29db/onnx/checker.cc#L634-L638

### Expected behavior
Error messages shouldn't contain binary data.

PR #3373, opened 2021-03-31
bug fix for proto utils and update checker error messages

askhade:

Addressing issue https://github.com/onnx/onnx/issues/3363


skottmckay (approved):

mkbhanda (commented):
There seem to be no references left to the inline function "ProtoDebugString", do we even need it regardless of whether ONNX_USE_LITE_PROTO is defined or not?

askhade:
> There seem to be no references left to the inline function "ProtoDebugString", do we even need it regardless of whether ONNX_USE_LITE_PROTO is defined or not?

Right! but since this is part of proto utils there is a possibility that runtimes or converters may have a dependency on this. Since we are very close to the release dont want to remove this at this moment... usually we give people ample notice before deprecating any utils or apis

gramalingam (approved):

Issue #3307, opened 2021-03-02
Wrong classification of 'protobuf' as 'purebasic' on github linguist


brymer-meneses:
ONNX uses Protobuf for `.pb` files but PureBasic is being assigned by GitHub linguist to these files.

This can be overridden by adding `.gitattributes` to the repository to tell GitHub linguist to classify `.pb` files as ProtoBuf not PureBasic.

Or, we can just remove `.pb` files from appearing at the Languages section of the repository, since ProtoBuf is not a programming language. This also can be done using `.gitattributes`.

jcwchen:
Hi @brymer-meneses,
https://github.com/github/linguist/issues/3816
So linguist didn't fix it? Actually I can repro this... Feel free to submit a PR to workaround this, but it seems to me that it is something needs to be fixed from linguist side. Thank you.

brymer-meneses:
Hello @jcwchen 

We can use `gitattributes` to directly tell linguist to classify `.pb` into ProtoBuff or we can just remove it entirely from the language section.

jcwchen:
I am OK with it, but I don't know other folks' thought. IMO, I think ONNX should be compact and simple so it would be good if it can be fixed from their side. Meanwhile, I will open an issue in linguist for this question. Thank you for bringing up this.

brymer-meneses:
I will submit a PR soon that will override linguist’s file classification.

brymer-meneses:
1.

brymer-meneses:
Hello @jcwchen 
Can you review my PR?

PR #3377, opened 2021-03-31
Fix GitHub programming language classification for the repo

prasanthpul:
GitHub marks onnx repo as PureBasic due to .pb protobuf data files that are used for tests. This change tells GitHub to ignore these files for programming language classification

Signed-off-by: Prasanth Pulavarthi <prasanth.pulavarthi@microsoft.com>


jcwchen:
Similar to https://github.com/onnx/onnx/pull/3309 and it can solve https://github.com/onnx/onnx/issues/3307

jcwchen (approved):

gramalingam (approved):

prasanthpul:
Thanks. Sorry I did not see #3309 before. Since the .pb files are test data, even though they are technically protobuf, it's better to exclude them rather than classify as protobuf. The .proto files are correctly classified as protobuf.

PR #3379, opened 2021-04-01
Updating BatchNorm outputs for training mode

neginraoof:
Updating BatchNorm outputs for training mode based on the discussions with ORT team.
cc @SherlockNoMad 

Signed-off-by: neginraoof <neginmr@utexas.edu>


onnx/defs/nn/defs.cc
```diff
@@ -1598,20 +1598,24 @@ statistics in inference mode (training_mode=False, default),
...
 running_var = input_var * momentum + current_var * (1 - momentum)
 
 Y = (X - current_mean) / sqrt(current_var + epsilon) * scale + B
+
+where:
+
+current_mean = ReduceMean(X, axis=all_except_channel_index)
+current_var =  ReduceVar(X, axis=all_except_channel_index)
 ```
+Outputs 'saved_tensor_1' and 'saved_tensor_2' are to be used for backend implementation only,
```
gramalingam, line 1617: Can we drop the last two outputs "saved_tensor_1" and "saved_tensor_2" from the spec?

SherlockNoMad (approved):

askhade (approved):

gramalingam (approved):

gramalingam:
Thanks very much for resolving the issue with the extra two outputs

onnx/defs/nn/defs.cc
pranav-prakash: @gramalingam Isn't the batch mean/var needed for the gradient calculation? If it's no longer outputted as part of the forward pass then it will have to be redundantly recomputed in the gradient op.

gramalingam: @pranav-prakash : there was an offline discussion as to what these values should be. The conclusion was that there was no unique-choice and that the actual backends compute something different from what was in the ONNX specification (either the inverse of variance or inverse of std-dev, etc.) So, the conclusion was to let backends transform the graph/node into a custom-op of their choice (for backward-propagation).

pranav-prakash: @gramalingam Hm if the backends need to replace Batchnorm with their own custom-op anyway, then isn't having the `running_mean/var` outputs pointless? Any custom-op that the backend defines would have to duplicate the entire schema def anyway, so there doesn't seem to be any advantage to including the `running_mean` in the schema upfront. (By such an argument, `is_training` could be deemed redundant as well, although I suppose that might have some use purely as a sentinel to indicate that the model was exported in training mode).

gramalingam: Technically, note that the model can use running_mean/var outputs later in the model ... backends that replace BatchNorm will then still be computing these for the subsequent usage, but using other mechanisms. @neginraoof can say more about how they plan to use this, as this was requested for some model export.

PR #3383, opened 2021-04-01
Bump version numbers for 1.9.0 release

postrational:


jcwchen (commented):
We need to bump opset_version here as well: https://github.com/onnx/onnx/blob/f22bde0a01899a4497e98c7fc22ffccf3362b07c/onnx/defs/schema.h#L902

One question: why don't we bump version number in the main branch? (Then we don't need to merge release branch back to the main branch in the end)

jcwchen:
nit: We can only set `1.8.200` in `VERSION_NUMBER` and set `1.9.0` in all other places. (For instance, https://github.com/onnx/onnx/blob/f22bde0a01899a4497e98c7fc22ffccf3362b07c/onnx/common/version.h#L11)

askhade:
> We need to bump opset_version here as well:
> 
> https://github.com/onnx/onnx/blob/f22bde0a01899a4497e98c7fc22ffccf3362b07c/onnx/defs/schema.h#L902
> 
> One question: why don't we bump version number in the main branch? (Then we don't need to merge release branch back to the main branch in the end)

That is because we build onnx-weekly packages off of master branch. IF we update the version number in master then we will see onnx-weekly package version bump before the release

jcwchen:
> That is because we build onnx-weekly packages off of master branch. IF we update the version number in master then we will see onnx-weekly package version bump before the release

That would be a pre-release (1.9.0.devXXX) before the real released 1.9.0. Still, I am OK with bumping it in the release branch. Thank you for the explanation

jcwchen:
We have merged a similar PR:https://github.com/onnx/onnx/pull/3408 to bring ONNX TestPyPI package to final verification. Thank you @postrational !

postrational:
Closed in favor of  #3408

PR #3393, opened 2021-04-02
Enhance test_backend_test.py and simplify backend test CI

jcwchen:
**Description**
- Use `onnx/test/test_backend_test.py` to fully test all backend data (use strict shape_inference)
- Update invalid models: Constant-1 does not allow int64 and it will fail shape_inference; bump the opset_version into Constant-9
  - test_PixelShuffle
  - test_operator_repeat
  - test_operator_repeat_dim_overflow
- Reuse test_backend_test.py to do the backend node test data check; Remove test_generated_backend.py

**Motivation**
test_backend_test.py has the same ability as `workflow_scripts/test_generated_backend.py `. ONNX should simplify backend test CI by enhancing existing test_backend_test.py.


.github/workflows/manylinux/test_package_i686.sh
```diff
@@ -26,12 +26,10 @@ $PIP_INTALL_COMMAND dist/*manylinux2010_i686.whl
...
 $PYTEST_COMMAND
 
-# Test backend test data
-# onnx.checker all existing backend data
-$PYTHON_COMAND workflow_scripts/test_generated_backend.py
-# onnx.checker all generated backend data
-$PYTHON_COMAND onnx/backend/test/cmd_tools.py generate-data
-$PYTHON_COMAND workflow_scripts/test_generated_backend.py
+# Test generated backend test data
+$PYTHON_COMMAND onnx/backend/test/cmd_tools.py generate-data
```
askhade, line 30: what is the motivation behind generating the data in the CI pipeline? why not simply test the data which is already checked in?

jcwchen: For now, it can only make sure models from test generation on various platforms (Windows, Linux, Mac) pass strict onnx.checker. It does not require the models generated from different platforms to be the same as the checked-in ones. We can probably achieve that after https://github.com/onnx/onnx/pull/3855 is merged.

.github/workflows/release_win.yml
```diff
@@ -106,14 +106,13 @@ jobs:
...
+    - name: Test generated backend test data
       run: |
         cd onnx
-        # onnx.checker all existing backend data
-        python workflow_scripts\test_generated_backend.py
         # onnx.checker all generated backend data
         python onnx\backend\test\cmd_tools.py generate-data
-        python workflow_scripts\test_generated_backend.py
+        # Only test generated backend node test data
+        pytest onnx\test\test_backend_test.py -k OnnxBackendNodeModelTest
```
askhade, line 115: what is "-k OnnxBackendNodeModelTest" for?

askhade: Also earlier test script was running both checker and shape_inference checks on models and now we are only doing the shape_inference check?

.github/workflows/manylinux/test_package_i686.sh
askhade: I think we should wait for #3855 and then refine this PR. 
Once #3855 is merged we should use since_version to create all the test data and run a git diff check to verify there are no diffs. This will help us validate the PR is including all the test data that it needs to (based on the unit test cases added) and that the test data is refreshed (in the past we have run into cases where author's made changes to the test case and forgot to regenerate the test data. This resulted in bad test data being checked in)

.github/workflows/release_win.yml
jcwchen: Specify OnnxBackendNodeModelTest here will make onnx\test\test_backend_test only test those models under `onnx/backend/test/case/node/`.

> Also earlier test script was running both checker and shape_inference checks on models and now we are only doing the shape_inference check?

Yes. To keep the same behavior as previous verification, I just added onnx.checker in test_backend_test.py. It can also enhance the test for backend data.

onnx/test/test_backend_test.py
```diff
@@ -41,8 +41,9 @@ def prepare(cls,
                 ):  # type: (...) -> Optional[onnx.backend.base.BackendRep]
         super(DummyBackend, cls).prepare(model, device, **kwargs)
 
-        # test shape inference
-        model = onnx.shape_inference.infer_shapes(model)
+        # test strict shape inference
+        model = onnx.shape_inference.infer_shapes(model, check_type=True, strict_mode=True)
+        onnx.checker.check_model(model, True)
```
askhade, line 46: I suggest moving check_model above infer_shapes. No point in inferring shapes for an invalid model.

.github/workflows/release_win.yml
askhade: Where is "OnnxBackendNodeModelTest " defined? The name suggests it will run tests for both Node and Model

jcwchen: I cannot find the source code about it shortly, but from my understanding the name here is collected as "OnnxBackend-[directory name under [here](https://github.com/onnx/onnx/tree/master/onnx/backend/test/data)]-ModelTest".
For instance, OnnxBackend"Node"ModelTest or OnnxBackend"PyTorchConverted"ModelTest.

liqunfu (approved):

jcwchen:
@onnx/sig-operators-approvers Could you please review/approve this PR? Thank you for your time!

gramalingam (approved):

PR #3394, opened 2021-04-02
[Shape inference] Always set the output of Shape to be rank-1

mrry:
Even when the input shape is unavailable, the output of Shape will always be a rank-1 vector.


gramalingam:
Thanks! May help to apply same fix to older version of Shape (in old.cc) also.

mrry:
Thanks @gramalingam... good idea, done!

gramalingam (approved):

postrational (approved):

PR #3398, opened 2021-04-05
Make SparseTensor type first class citizen.

yuslepukhin:
Adjust tests, checker and shape inference.

Signed-off-by: Dmitri Smirnov <dmitrism@microsoft.com>


askhade:
IR_Version needs to be updated too right?

onnx/shape_inference/implementation.cc
```diff
@@ -77,6 +87,39 @@ void checkShapesAndTypes(const TypeProto_Tensor& inferredType, const TypeProto_T
   }
 }
 
+void checkShapesAndTypes(const TypeProto_SparseTensor& inferredType, const TypeProto_SparseTensor& existingType) {
```
askhade, line 90: Can this be simplified by using a template?

onnx/shape_inference/implementation.cc
```diff
@@ -175,13 +251,14 @@ static void InferShapesImpl(
     }
   }
 
-  std::unordered_map<std::string, const TensorProto*> inputDataByName;
   // save for free memory
-  std::vector<TypeProto*> initializerTypeList;
+  std::vector<std::unique_ptr<TypeProto>> initializerTypeList;
```
askhade, line 255: do we even need a unique_ptr here? Can the vector simply hold the TypeProto object?

gramalingam:
A generic question: is this the same as the closed PR, or is there any significant difference in this PR?

onnx/defs/shape_inference.h
```diff
@@ -154,131 +141,141 @@ multiplyDims(const TensorShapeProto& shape, int from, int upto_exclusive) {
   return dim;
 }
 
-void propagateElemTypeWithValidation(
-    const TypeProto* input_type,
-    TypeProto* output_type);
+inline int32_t getTensorElementType(const TypeProto& type) {
```
gramalingam, line 144: (General comment, not specific to this PR). I wonder if this include file is worth splitting into an include-file and implementation-file (in a separate PR, doesn't have to be in this). It is growing big, and it is better to reduce unnecessary implementation-dependence some users might acquire.

yuslepukhin:
There should be no difference as this was created to deal with DCO issues.

---
In reply to: [814347121](https://github.com/onnx/onnx/pull/3398#issuecomment-814347121) [](ancestors = 814347121)

yuslepukhin:
Will do.

---
In reply to: [814311059](https://github.com/onnx/onnx/pull/3398#issuecomment-814311059) [](ancestors = 814311059)

onnx/defs/shape_inference.h
yuslepukhin: I was thinking along the same lines.

onnx/shape_inference/implementation.cc
yuslepukhin: Will do.

yuslepukhin:
> IR_Version needs to be updated too right?

Will do

onnx/defs/shape_inference.h
askhade: I also agree

yuslepukhin: I think this would fall under refactoring category and we better manage it under a separate PR.

jcwchen (commented):
The CI is failing because https://github.com/onnx/onnx/blob/master/onnx/examples/make_model.ipynb checks ir_version as an old one (7).

Perhaps we can simply get rid of this line: `print('The ir_version in model: {}\n'.format(model_def.ir_version))`? Otherwise every time ONNX bumps ir_version, `make_model.ipynb` needs to be updated as well...

onnx/examples/make_model.ipynb
```diff
@@ -9,7 +9,6 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "The ir_version in model: 7\n",
       "\n",
```
jcwchen, line 12: I think L13 needs to be removed as well. The check here is quite unfriendly for people who want to update it... I will try to improve it in another PR. Thank you for the effort.

gramalingam (approved):

onnx/defs/shape_inference.h
```diff
@@ -61,6 +60,9 @@ struct InferenceContext {
   virtual TypeProto* getOutputType(size_t index) = 0;
   virtual GraphInferencer* getGraphAttributeInferencer(const std::string& attribute_name) = 0;
   virtual ~InferenceContext() {}
+  /// XXX: Putting at the end of vtable in case it matters?
+  /// XXX: Do we need default implementation here?
```
askhade, line 64: no we dont... you can remove this comment

onnx/onnx.in.proto
```diff
@@ -100,7 +100,12 @@ enum Version {
   //   method and training algorithm. The execution of TrainingInfoProto
   //   can modify the values of mutable variables.
   // - Implicitly add inference graph into each TrainingInfoProto's algorithm.
-  IR_VERSION = 0x0000000000000007;
+  IR_VERSION_2020_5_8 = 0x0000000000000007;
+
+  // IR VERSION 8 published on <TBD>
```
askhade, line 105: may be we can change this to published with ONNX release 1.10? This way no one will forget to update the TBD

askhade (approved):
Add 2 nit comments other than that looks good to me. Thanks for waiting till ONNX 1.9 release

jcwchen (approved):

onnx/defs/shape_inference.h
gramalingam: I think a default implementation does help users who have an implementation of InferenceContext use this version without modifying their code (which seems reasonable if they do not want to handle sparse data yet).

PR #3407, opened 2021-04-07
Adding optional type

neginraoof:
As discussed in the document for support of Optional types in ONNX:
https://microsoft.sharepoint.com/:w:/r/teams/ONNX2/_layouts/15/Doc.aspx?sourcedoc=%7BEDD907A1-2D0D-4BB8-A073-C6306BC17748%7D&file=ONNX_Optional_Type.docx&action=default&mobileredirect=true&DefaultItemOpen=1&cid=d6ef719f-d187-46ab-8393-947f6fba215a

Following this, ops for constructing an optional types will be added.

Signed-off-by: neginraoof <neginmr@utexas.edu>


yuslepukhin:
Would be nice to have a general description of the PR as having TypeProto in the attribute is unusual and requires explanation.

onnx/onnx.in.proto
```diff
@@ -121,13 +121,15 @@ message AttributeProto {
     TENSOR = 4;
     GRAPH = 5;
     SPARSE_TENSOR = 11;
+    TYPE_PROTO = 13;
```
yuslepukhin, line 124: Changes in in.proto files require regeneration of the actual proto files. Use
`python onnx/gen_proto.py -l`
`
`

onnx/common/ir_pb_converter.cc
```diff
@@ -145,6 +145,18 @@ void convertAttribute(const ONNX_NAMESPACE::AttributeProto& ap, Node* n) {
       n->ts_(sym, std::move(tensors));
       break;
     }
+    case ONNX_NAMESPACE::AttributeProto_AttributeType_TYPE_PROTO:
```
yuslepukhin, line 148: I have a PR pending that introduces SparseTensor as first class citizen. Should I also add some changes here so sparse attributes are handled or it is something that the converter team would address later?

onnx/common/ir_pb_converter.cc
```diff
@@ -498,6 +510,18 @@ void addAttribute(ONNX_NAMESPACE::NodeProto* n_p, Node* n, Symbol name) {
         encodeGraph(g, v);
       }
       break;
+    case AttributeKind::tp: {
+      attr->set_type(ONNX_NAMESPACE::AttributeProto_AttributeType_TYPE_PROTO);
+      auto tp = attr->mutable_tp();
+//      encodeTypeProto(tp, n->tp(name));
```
yuslepukhin, line 516: Comments must be resolved one way or another.

yuslepukhin (changes requested):

neginraoof:
Thanks @yuslepukhin for the review. This is a WIP that I opened to get some feedback as I'm completing the work. Will add more details on why this is needed.

onnx/onnx-data.in.proto
```diff
@@ -99,3 +99,40 @@ message MapProto {
...
+  optional string name = 1;
+
+  enum DataType {
+    UNDEFINED = 0;
+    TENSOR = 1;
+    SPARSE_TENSOR = 2;
+    SEQUENCE = 3;
+    MAP = 4;
+  }
+
```
neginraoof, line 117: Check if we need oneof here.

onnx/helper.py
```diff
@@ -311,6 +311,22 @@ def make_map(
...
+        name,   # type: Text
+        elem_type,   # type: int
+        values,   # type: Optional[Any]
+):  # type: (...) -> OptionalProto
+    '''
+    Make a Sequence with specified value arguments.
+    '''
+    optional = OptionalProto()
+    optional.name = name
+    optional.elem_type = elem_type
```
neginraoof, line 324: How does element type specify the elements' data type?

onnx/common/ir_pb_converter.cc
neginraoof: @yuslepukhin 
Thanks. I'm trying to find a method or write one to copy or clone onnx::TypeProto. Do you have a suggestion for this part?

onnx/test/helper_test.py
```diff
@@ -134,6 +134,20 @@ def test_attr_repeated_graph_proto(self):  # type: () -> None
...
+        attr = helper.make_attribute("type_proto", type)
+        self.assertEqual(attr.name, "type_proto")
+        self.assertEqual(attr.tp, type)
+        self.assertEqual(attr.type, AttributeProto.TYPE_PROTO)
+        # type_protos
+        types = [TypeProto(), TypeProto()]
+        attr = helper.make_attribute("type_proto", types)
+        self.assertEqual(attr.name, "type_proto")
+        self.assertEqual(list(attr.type_protos), types)
+        self.assertEqual(attr.type, AttributeProto.TYPE_PROTO)
```
jcwchen, line 149: Why it is not AttributeProto.TYPE_PROTOS?

onnx/common/ir_pb_converter.cc
```diff
@@ -498,6 +538,18 @@ void addAttribute(ONNX_NAMESPACE::NodeProto* n_p, Node* n, Symbol name) {
...
+      auto tp = attr->mutable_tp();
+      encodeTypeProto(tp, n->tp(name));
+    } break;
+    case AttributeKind::tps:
+      attr->set_type(ONNX_NAMESPACE::AttributeProto_AttributeType_TYPE_PROTOS);
+      for (auto& v : n->tps(name)) {
+        auto tp = attr->add_type_protos();
+        encodeTypeProto(tp, v);
+      }
+      break;
```
jcwchen, line 552: nit: L490-L552 Could we sync all the bracket and break with the same coding style? It looks a little confusing.

onnx/numpy_helper.py
```diff
@@ -272,6 +272,67 @@ def from_dict(dict, name=None):  # type: (Dict[Any, Any], Optional[Text]) -> Map
...
+    values = getattr(optional, value_field)
+    for value in values:
+        if elem_type == OptionalProto.TENSOR or elem_type == OptionalProto.SPARSE_TENSOR:
+            opt = to_array(value)
+        elif elem_type == OptionalProto.SEQUENCE:
+            opt = to_list(value)
+        elif elem_type == OptionalProto.MAP:
+            opt = to_dict(value)
+        else:
+            raise TypeError("The element type in the input optional is not supported.")
```
jcwchen, line 295: Random thought:
Is it possible to create a new utility function for these if-else conditions? (L288-L295 and L150-L157 is a similar logic) `from_list` and `from_optional` have the same situation.

Since using another new function might lose some readability, honestly I am open to both options (keep the original way or introduce a new function). Thanks.

onnx/common/ir_pb_converter.cc
gramalingam: Run clangformat?

onnx/helper.py
```diff
@@ -311,6 +311,23 @@ def make_map(
     return map
 
 
+def make_optional(
+        name,   # type: Text
+        elem_type,   # type: int
+        values,   # type: Optional[Any]
+):  # type: (...) -> OptionalProto
+    '''
+    Make a Sequence with specified value arguments.
```
gramalingam, line 320: "Sequence" => "OptionalProto" ?

onnx/helper.py
```diff
@@ -311,6 +311,23 @@ def make_map(
     return map
 
 
+def make_optional(
+        name,   # type: Text
+        elem_type,   # type: int
```
gramalingam, line 316: The annotation ```type: int``` doesn't sound correct.

gramalingam: Could be a pre-existing problem with the Sequence also

onnx/common/ir_pb_converter.cc
yuslepukhin: > 
> 
> @yuslepukhin
> Thanks. I'm trying to find a method or write one to copy or clone onnx::TypeProto. Do you have a suggestion for this part?

The method is called Copy()

onnx/helper.py
gramalingam: A couple of related questions: (a) Does the caller have to specify elem_type, or should we infer it from the actual value? (b) What is the purpose of elem_type in the proto, is it to just indicate which variant is being used or to indicate the full-type of the contained value? I think we can only do the former, not the latter. Once the "proto" spec is clarified, I think the helper methods can be adjusted accordingly.

gramalingam: Stated another way: how is a None value represented in the Proto? Is elem_type supposed to be UNDEFINED?

onnx/helper.py
```diff
@@ -311,6 +311,23 @@ def make_map(
...
+        values,   # type: Optional[Any]
```
gramalingam, line 317: "value" may be more appropriate than "values"

onnx/common/ir_pb_converter.cc
yuslepukhin: @neginraoof if using Visual Studio, configure it to automatically pickup .clangformat file in this repo and other related options.

onnx/onnx-data.in.proto
```diff
@@ -99,3 +99,40 @@ message MapProto {
...
+message OptionalProto {
+
+  optional string name = 1;
+
+  enum DataType {
+    UNDEFINED = 0;
+    TENSOR = 1;
+    SPARSE_TENSOR = 2;
+    SEQUENCE = 3;
+    MAP = 4;
```
gramalingam, line 115: (Comment meant mainly for SequenceProto, but also here for completeness) We should add "OPTIONAL = 5" to this enum, and add a field "repeated OptionalProto optional_values" (for SequenceProto) and, similarly, "optional OptionalProto optional_value" here ... even though an optional value of optional type is unlikely to be used, it would help to be complete.

onnx/onnx-data.in.proto
```diff
@@ -99,3 +99,40 @@ message MapProto {
...
+  }
+
+  // The data type of the element.
+  // This field MUST have a valid OptionalProto.DataType value
+  optional int32 elem_type = 2;
+
+  // For TensorProto values.
+  // When this field is present, the elem_type field MUST be TENSOR.
+  optional TensorProto tensor_values = 3;
```
gramalingam, line 124: All these names must be singular "tensor_value", etc.

postrational:
@neginraoof Please add a description to this PR with a motivation for this new type and examples of how it will be used.

postrational:
What operators are we expecting to use this optional type?

onnx/numpy_helper.py
neginraoof: I think we can have two maps and use the key of elem_type to get corresponding the transform. I think this is pretty readable and small for now. I'll leave a todo if the list gets expanded.

onnx/common/ir_pb_converter.cc
neginraoof: Thanks!

onnx/common/ir_pb_converter.cc
neginraoof: Yeah would be great if you can add sparse tensor attributes. Thanks!

onnx/helper.py
neginraoof: @gramalingam 
I think from what we discussed, we decided to mandate setting the elem_type even for None.
Use of UNDEFINED I think is mainly for data serialization purposes.

onnx/onnx-data.in.proto
```diff
@@ -99,3 +104,45 @@ message MapProto {
...
+  // When this field is present, the elem_type field MUST be SEQUENCE.
+  optional SequenceProto sequence_value = 5;
+
+  // For MapProto value.
+  // When this field is present, the elem_type field MUST be MAP.
+  optional MapProto map_value = 6;
+
+  // For OptionalProto values, allowing optional to be of itself.
+  // When this field is present, the elem_type field MUST be OPTIONAL.
+  optional SequenceProto values = 5;
```
jcwchen, line 146: This one should be 7?

neginraoof: Thanks!

neginraoof:
@gramalingam @jcwchen @postrational 
Thanks a lot for your comments. I have addressed your comments. Please help in reviewing the updates.

onnx/defs/schema.h
```diff
@@ -735,6 +735,41 @@ class OpSchema final {
     return all_tensor_sequence_types;
   }
 
+  static const std::vector<std::string>& all_optional_types() {
```
gramalingam, line 738: Is this used anywhere? I think it would be better NOT to introduce this kind of "limited" set of special types or encourage its use in op schemas.

onnx/mapping.py
```diff
@@ -66,5 +66,9 @@
     int(SequenceProto.TENSOR): 'tensor_values',
     int(SequenceProto.SPARSE_TENSOR): 'sparse_tensor_values',
     int(SequenceProto.SEQUENCE): 'sequence_values',
-    int(SequenceProto.MAP): 'map_values'
+    int(SequenceProto.MAP): 'map_values',
+    int(OptionalProto.TENSOR): 'tensor_value',
```
gramalingam, line 70: I think this is not going to work, because SequenceProto.TENSOR and OptionalProto.TENSOR have same (key) value. We will have to use two different maps, one for Sequence, and one for Optional.

onnx/defs/schema.h
neginraoof: It is not used yet, but I guess it would be needed once introducing the ops and specifying input/output types.
What is your suggestion for this use-case?

onnx/helper.py
```diff
@@ -506,6 +533,30 @@ def make_sequence_value_info(
...
+        elem_shape_denotation=None,  # type: Optional[List[Text]]
+):  # type: (...) -> ValueInfoProto
+    """Makes a ValueInfoProto based on the data type and shape for Optional."""
+    value_info_proto = ValueInfoProto()
+    value_info_proto.name = name
+    if doc_string:
+        value_info_proto.doc_string = doc_string
+
+    optional_type_proto = value_info_proto.type.optional_type
+    optional_type_proto.elem_type.tensor_type.elem_type = elem_type
```
askhade, line 550: make_optional_value_info will only work for tensors and not sequences or sparse tensors?

neginraoof: This was replicating the behavior of make_sequence_value_info. I extended this for to support sequences as well.

onnx/defs/schema.h
askhade: Once we introduce ops do we need to support all these types? can we remove this and simply add support for the set of types which are required for the op? Example optional(seq(tensor(complex64))) it seems unlikely that we will need support for this...

onnx/helper.py
askhade: another question regarding the elem_type - similar to Rama's question what is the expected value for elem_type? Is it the type held by Optional say Optional <Tensor, Sequence etc> or the complete type meaning optional(Tensor(float32))

neginraoof:
@askhade @gramalingam 
Do you have any other comments?
Thanks for review.

onnx/helper.py
```diff
@@ -547,13 +547,14 @@ def make_optional_value_info(
...
 
     optional_type_proto = value_info_proto.type.optional_type
-    optional_type_proto.elem_type.tensor_type.elem_type = elem_type
-
-    tensor_value_info = make_tensor_value_info(name, elem_type, shape, doc_string, elem_shape_denotation)
-
-    if shape is not None:
-        optional_type_proto.elem_type.tensor_type.shape.CopyFrom(tensor_value_info.type.tensor_type.shape)
-
+    if elem_type == OptionalProto.TENSOR:
```
gramalingam, line 550: I am slightly bothered by this utility: it is not fully general-purpose, and seems specialized to handle just the common cases (of the optional type). E.g., I think it won't work for creating more complex things like "optional sequence of sequence of tensor" etc., or am I mistaken? I feel it may be better to omit this utility for now (or replace with a general-purpose utility which would be extra work). Otherwise, it would be problematic to update it later on if users start taking a dependence on its current form.

neginraoof: This is very close (and even more generic) than make_sequence_value_info.
This function would be needed for shape inference of an op with optional input, and I have the spec for the new op to send right after this PR.
I currently don't see any of the similar utility functions handling more complex types, and I don't also see a use-case for it.
Anyways, if you think so, I can remove it. But the following PR I have requires this.

neginraoof:
cc @postrational 
Please help us review this PR. Thanks.

onnx/helper.py
gramalingam: I agree that make_sequence_value_info has a similar problem. I am afraid that we are duplicating the problem and spreading it further. My concern is if we end up with lot of code assuming that a "sequence" is only a "sequence of tensor" or that an "optional" is only an "optional tensor" or "optional sequence of tensor" (and then we end up trying to using something more general like a "sequence of sequence of tensor" we may hit issues). When I say remove it, I mean that we can write the equivalent code in the caller. At a call-site, the caller knows that they are only creating an "optional tensor" and so write code for that, and that looks preferable. Anyway, happy to hear other opinions. I don't mind the current form if we can clean it up later, but that is tricky once users start using it widely.

onnx/numpy_helper.py
```diff
@@ -272,6 +272,71 @@ def from_dict(dict, name=None):  # type: (Dict[Any, Any], Optional[Text]) -> Map
     return map
 
 
+def to_optional(optional):  # type: (OptionalProto) -> Optional[Any]
+    """Converts an optional def to a Python list.
```
gramalingam, line 276: nit: "Python list" seems like a copy-paste error. Maybe a "Python optional"?

onnx/numpy_helper.py
```diff
@@ -272,6 +272,71 @@ def from_dict(dict, name=None):  # type: (Dict[Any, Any], Optional[Text]) -> Map
...
+
+    Inputs:
+        optional: an OptionalProto object.
+    Returns:
+        lst: the converted list.
```
gramalingam, line 281: Doesn't return a list (always).

onnx/numpy_helper.py
```diff
@@ -272,6 +272,71 @@ def from_dict(dict, name=None):  # type: (Dict[Any, Any], Optional[Text]) -> Map
...
+    """
+    opt = None  # type: Optional[Any]
+    elem_type = optional.elem_type
+    value_field = mapping.OPTIONAL_ELEMENT_TYPE_TO_FIELD[elem_type]
+    values = getattr(optional, value_field)
+    # TODO: create a map and replace conditional branches
+    for value in values:
```
gramalingam, line 288: We don't really have repeated values. Why is this loop needed?

onnx/numpy_helper.py
```diff
@@ -272,6 +272,71 @@ def from_dict(dict, name=None):  # type: (Dict[Any, Any], Optional[Text]) -> Map
...
+        else:
+            raise TypeError("The element type in the input optional is not supported.")
+    return opt
+
+
+def from_optional(opt, name=None, dtype=None):  # type: (Optional[Any], Optional[Text], Optional[int]) -> OptionalProto
+    """Converts an optional value into a Optional def.
+
+    Inputs:
+        lst: a Python list
```
gramalingam, line 304: Not a list.

onnx/helper.py
gramalingam: Alternatively, we could define two specialized functions with different names like "make_optional_tensor_value_info" and "make_optional_sequence_tensor_value_info" ... at least, it makes it clear it is for a special-case.

onnx/helper.py
```diff
@@ -533,28 +533,46 @@ def make_sequence_value_info(
...
+
+def make_optional_sequence_value_info(
+        name,  # type: Text
+        elem_type,  # type: int
+        shape,  # type: Optional[Sequence[Union[Text, int]]]
+        doc_string="",  # type: Text
+        elem_shape_denotation=None,  # type: Optional[List[Text]]
+):  # type: (...) -> ValueInfoProto
+    """Makes a ValueInfoProto based on the data type and shape for Optional."""
+    assert elem_type == OptionalProto.SEQUENCE
```
gramalingam, line 567: There seems to be some potential confusion between this and the passing of elem_type to ```make_sequence_value_info``` down below ... both elem_types cannot be the same, they must be different, right?

onnx/helper.py
gramalingam: Is "elem_type" supposed to be OptionalProto.TENSOR as in assert a few lines above, or is it supposed to the element type of tensor? I think the latter one. But I am unclear how a test-case would pass that assert above.

onnx/helper.py
neginraoof: Fixed this with another argument for seq_elem_type

askhade: building sequence is a bit complicated. Theoretically sequence can contain a tensor or a sequence of tensors or a sequence of sequence of tensors etc...

Current make_sequence_value_info only support Sequence of Tensors ... Are use cases for Optional type to support anything other Sequence of Tensors? if so then make_sequence_valu_info needs to be fixed. 

Regardless we should add some documentation and checks around this

onnx/helper.py
askhade: actually looking at the caller the elem_type is OptionalProto.Tensor
if ele_type == onnx.mapping.OptionalProto.TENSOR:
            return onnx.helper.make_optional_tensor_value_info(
                name=name,
                elem_type=ele_type,
                shape=None)

what this means is this type will not work for make_tensor_value_info

onnx/numpy_helper.py
```diff
@@ -272,6 +272,70 @@ def from_dict(dict, name=None):  # type: (Dict[Any, Any], Optional[Text]) -> Map
...
+    Returns:
+        optional: the converted optional def.
+    """
+    # TODO: create a map and replace conditional branches
+    optional = OptionalProto()
+    if name:
+        optional.name = name
+
+    if dtype:
+        elem_type = dtype
```
askhade, line 316: add some error checking around this? as in whether dtype falls within the expected values

onnx/numpy_helper.py
```diff
@@ -272,6 +272,70 @@ def from_dict(dict, name=None):  # type: (Dict[Any, Any], Optional[Text]) -> Map
...
+    elif isinstance(opt, dict):
+        elem_type = SequenceProto.MAP
+    elif isinstance(opt, list):
+        elem_type = SequenceProto.SEQUENCE
+    else:
+        elem_type = SequenceProto.TENSOR
```
askhade, line 322: is it OK to default to tensor?

askhade, line 322: a bit confused... why is this not just a TensorProto?

onnx/numpy_helper.py
```diff
@@ -272,6 +272,70 @@ def from_dict(dict, name=None):  # type: (Dict[Any, Any], Optional[Text]) -> Map
...
+
+    optional.elem_type = elem_type
+
+    if opt is not None:
+        if elem_type == OptionalProto.TENSOR:
```
askhade, line 327: not sure if I am missing something - 317 to 322 set elem_type to SequenceProto.blah yet here there is no option for SequenceProto.blah

onnx/helper.py
gramalingam: Nit: IMHO, we can just drop the "elem_type" parameter, which is used only in the assert. We only need the seq_elem_type.

onnx/numpy_helper.py
neginraoof: Thanks, yeah this part needs to be updated to OptionalProto.

onnx/helper.py
```diff
@@ -446,9 +473,21 @@ def make_tensor_value_info(
...
+    return value_info_proto
+
+
+def make_tensor_type_proto(
+        elem_type,  # type: int
+        shape,  # type: Optional[Sequence[Union[Text, int]]]
+        shape_denotation=None,  # type: Optional[List[Text]]
+):  # type: (...) -> TypeProto
+    """Makes a Tensor TypeProto based on the data type and shape."""
+    value_info_proto = ValueInfoProto()
```
gramalingam, line 487: Seems redundant to create a ValueInfoProto and then throw it away. I think we can just do ```tensor_type_proto = TypeProto()```, right?

onnx/helper.py
```diff
@@ -483,7 +522,7 @@ def make_tensor_value_info(
             if shape_denotation:
                 dim.denotation = shape_denotation[i]
 
-    return value_info_proto
+    return value_info_proto.type
```
gramalingam, line 525: And change this to ```return type_proto``` (see comment above).

onnx/helper.py
```diff
@@ -493,15 +532,27 @@ def make_sparse_tensor_value_info(
...
+    return value_info_proto
+
+
+def make_sparse_tensor_type_proto(
+        elem_type,  # type: int
+        shape,  # type: Optional[Sequence[Union[Text, int]]]
+        shape_denotation=None,  # type: Optional[List[Text]]
+):  # type: (...) -> TypeProto
+    """Makes a SparseTensor TypeProto based on the data type and shape."""
+    value_info_proto = ValueInfoProto()
```
gramalingam, line 552: See previous comment. This seems redundant. We should just create a TypeProto() here.

onnx/helper.py
```diff
@@ -536,30 +587,41 @@ def make_sparse_tensor_value_info(
...
 
-    return value_info_proto
+    return value_info_proto.type
+
+
+def make_sequence_type_proto(
+        inner_type_proto,  # type: TypeProto
+):  # type: (...) -> TypeProto
+    """Makes a sequence TypeProto."""
+    value_info_proto = ValueInfoProto()
```
gramalingam, line 597: As above. Should just create a TypeProto.

onnx/helper.py
```diff
@@ -536,30 +587,41 @@ def make_sparse_tensor_value_info(
...
+    type_proto = value_info_proto.type
+    type_proto.sequence_type.elem_type.CopyFrom(inner_type_proto)
+    return type_proto
+
+
+def make_optional_type_proto(
+        inner_type_proto,  # type: TypeProto
+):  # type: (...) -> TypeProto
+    """Makes an optional TypeProto."""
+    value_info_proto = ValueInfoProto()
```
gramalingam, line 607: As above. Should just create a TypeProto.

gramalingam:
I notice that the IR version has already been incremented due to the addition of SparseTensor. Can you please add a comment there, documenting the new IR change (addition of Optional type)?

gramalingam (approved):

onnx/helper.py
```diff
@@ -493,15 +532,27 @@ def make_sparse_tensor_value_info(
...
         value_info_proto.doc_string = doc_string
 
-    sparse_tensor_type_proto = value_info_proto.type.sparse_tensor_type
-    sparse_tensor_type_proto.elem_type = elem_type
+    sparse_tensor_type_proto = make_sparse_tensor_type_proto(elem_type, shape, shape_denotation)
+    value_info_proto.type.sparse_tensor_type.CopyFrom(sparse_tensor_type_proto.sparse_tensor_type)
+    return value_info_proto
 
+
+def make_sparse_tensor_type_proto(
```
askhade, line 546: nit: move this above "make_sparse_tensor_value_info"

askhade (approved):

PR #3412, opened 2021-04-08
Add Additional bfloat16 Support to Pow

### LINKED ISSUE #3409: Pow support for bfloat16
I see that with opset13 bfloat16 support has been added to [Pow](https://github.com/onnx/onnx/blob/master/docs/Operators.md#pow
), however only the first operand `T` has been updated to support bfloat16 ; should the second operand `T1` be updated as well?

Reproduced here from https://github.com/onnx/onnx/blob/master/docs/Operators.md#pow

```
T : tensor(int32), tensor(int64), tensor(float16), tensor(float), tensor(double), tensor(bfloat16)
Constrain input X and output types to float/int tensors.
T1 : tensor(uint8), tensor(uint16), tensor(uint32), tensor(uint64), tensor(int8), tensor(int16), tensor(int32), tensor(int64), tensor(float16), tensor(float), tensor(double)
Constrain input Y types to float/int tensors.
```

manbearian:
# Add Additional bfloat16 Support to Pow 

Allow the second operand of `Pow` (a.k.a. the 'power of the exponent') to be a tensor of bfloat16 values.

The change resolves #3409.


gramalingam:
LGTM, thanks! Please note that the opset will likely need to be adjusted to 15 (unless this PR is cherry-picked for the next release).

postrational:
Yes, this PR wasn't included in our 1.9 release, so we'll need to include it in op set 15.

manbearian:
> Yes, this PR wasn't included in our 1.9 release, so we'll need to include it in op set 15.

Seems like something i should be able to fix. I'll push a change shortly

manbearian:
> > Yes, this PR wasn't included in our 1.9 release, so we'll need to include it in op set 15.
> 
> Seems like something i should be able to fix. I'll push a change shortly

This is proving trickier then i would have expected. The doc generating doesn't seem to be working now that i've changed to opset 15. Do i need to add additional code to add a new opset version?

postrational (approved):

gramalingam (approved):

PR #3402, opened 2021-04-06
Update BatchNorm to specify population variance

neginraoof:
Update BatchNorm to specify population variance.

Signed-off-by: neginraoof <neginmr@utexas.edu>

PR #3415, opened 2021-04-09
Add new type constrains for variance and mean in BatchNorm

jcwchen:
**Description**
- Add another type constrain mentions more: it allows all float type for U, but backend would only impl for fp32
- Add missing bfloat16

**Motivation**
Continue the remain work from: https://github.com/onnx/onnx/pull/3402. During mixed precision training of BatchNorm, for numerical stability, in the current state, we usually keep input_mean, input_var and running_mean and running_var in fp32, while X and Y can be in fp16. Therefore we add a new type constrain for this difference.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -18688,9 +18688,9 @@ This version of the operator has been available since version 14 of the default 
 <dd>Scale tensor of shape (C).</dd>
 <dt><tt>B</tt> (differentiable) : T</dt>
 <dd>Bias tensor of shape (C).</dd>
-<dt><tt>input_mean</tt> (differentiable) : T</dt>
+<dt><tt>input_mean</tt> (differentiable) : U</dt>
 <dd>running (training) or estimated (testing) mean tensor of shape (C).</dd>
-<dt><tt>input_var</tt> (differentiable) : T</dt>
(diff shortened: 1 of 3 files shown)


jcwchen:
cc @SherlockNoMad

docs/Changelog.md
```diff
@@ -18699,17 +18699,19 @@ This version of the operator has been available since version 14 of the default
...
 </dl>
 
 #### Type Constraints
 
 <dl>
-<dt><tt>T</tt> : tensor(float16), tensor(float), tensor(double)</dt>
+<dt><tt>T</tt> : tensor(float16), tensor(float), tensor(double), tensor(bfloat16)</dt>
 <dd>Constrain input and output types to float tensors.</dd>
+<dt><tt>U</tt> : tensor(float16), tensor(float), tensor(double), tensor(bfloat16)</dt>
+<dd>Constrain mean and variance types to float tensors. It allows all float type for U, but backend would only implement for fp32.</dd>
```
gramalingam, line 18714: Drop ", but backend would only implement for fp32." That is not part of the spec.

jcwchen: Dropped. Thanks

gramalingam (approved):

SherlockNoMad (approved):

PR #3426, opened 2021-04-15
Bump VERSION NUMBER to 1.9.0

etusien:
Signed-off-by: Ewa21 <ewa.tusien@intel.com>


jcwchen (approved):

jcwchen (commented):
Sorry that we just found Batchnorm needs more changes to make type inference work well for all types... Probably it needs one more attribute. If you are comfortable with this, could you please hold on the final release now? Thank you for your understanding. cc @gramalingam

Issue #3428, opened 2021-04-16
[Operator] RoiAlign backend test case's output is misaligned


fdwr:
# Bug Report

### Is the issue related to model conversion?
No.

### Describe the bug
The [RoiAlign operator](https://github.com/onnx/onnx/blob/master/docs/Operators.md#RoiAlign), per the [Mask R-CNN paper](https://arxiv.org/abs/1703.06870) and Facebook Research's [Detectron 2 implementation](https://github.com/facebookresearch/detectron2/blob/master/detectron2/layers/roi_align.py) aligns sampling points over the center of the pixels, but the [solitary test case](https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/roialign.py) in the ONNX backend has output that is misaligned by half an input pixel. After comparing to various references (table below under **Notes**) and reading more into the history of this operator, I see this is actually an old bug (which Detectron since fixed and PyTorch also fixed with aligned=True) which applied an offset the output subsample by 0.5 but forgot to adjust the input sample to compensate (see their comment in the [code](https://github.com/facebookresearch/detectron2/blob/master/detectron2/layers/roi_align.py): "_the original roi_align (aligned=False) does not subtract the 0.5 when computing neighboring pixel indices and therefore it uses pixels with a slightly incorrect alignment (relative to our pixel model) when performing bilinear interpolation_" and the [test case](https://github.com/facebookresearch/detectron2/blob/master/tests/layers/test_roi_align.py#L15)).

From the paper, note pixel centers used for interpolation:
![image](https://user-images.githubusercontent.com/1809166/110403224-95ce2500-8031-11eb-80c0-92732fa247d2.png)

We should fix the existing ONNX backend test case accordingly. Otherwise even a simple case of cropping an identity region from the input doesn't match. e.g. a new test case:

```python
    # Crop a 2x2 region from the 4x5 input.
    @staticmethod
    def export_roialign_identity_region():  # type: () -> None
        node = onnx.helper.make_node(
            "RoiAlign",
            inputs=["X", "rois", "batch_indices"],
            outputs=["Y"],
            spatial_scale=1.0,
            output_height=2,
            output_width=2,
            sampling_ratio=1,
        )

        X = np.array(
            [
                [
                    [
                        [ 0, 1, 2, 3],
                        [ 4, 5, 6, 7],
                        [ 8, 9,10,11],
                        [12,13,14,15],
                        [16,17,18,19],
                    ]
                ]
            ],
[...]

PR #3429, opened 2021-04-16
Fix RoiAlign backend test case output, and add new test case for an identity slice.

fdwr:
**Description**:
- See issue [3428](https://github.com/onnx/onnx/issues/3428) for details, but in short...
- Fix `test_roialign` baseline results to align pixel sample to centers. The correct equation is:<br/>`input.x = (output.x + 0.5) * computedOutputToInputScale + scaledRoi.x1 - 0.5`, not<br/>`input.x = (output.x + 0.5) * computedOutputToInputScale + scaledRoi.x1 - 0.0`.
- Add an identity window test case where output should exactly equal the input window.

**Motivation and Context**
- _Why is this change required? What problem does it solve?_ Fixes tolerance issues in failing test (ONNX Runtime DirectML EP). Fixes bug that has been addressed in other frameworks/implementations (Detectron [(1)](https://github.com/facebookresearch/detectron2/blob/master/detectron2/layers/roi_align.py#L24) [(2)](https://github.com/facebookresearch/detectron2/blob/master/tests/layers/test_roi_align.py#L35) [(3)](https://detectron2.readthedocs.io/en/latest/notes/compatibility.html#compatibility-with-detectron-and-maskrcnn-benchmark), Caffe/PyTorch [(1)](https://pytorch.org/vision/0.8/ops.html#torchvision.ops.roi_align) [(2)](https://github.com/pytorch/pytorch/pull/23706), mxnet [(1)](https://github.com/apache/incubator-mxnet/pull/16619)).
- _If it fixes an open issue, please link to the issue here._ https://github.com/onnx/onnx/issues/3428


askhade (approved):

wschin:
This looks very reasonable. Pytorch (and other frameworks) has two versions of alignment. Is is possible to maintain BC by introducing an attribute like this one
```
 torchvision.ops.roi_align(aligned=False…)
```
? We also need to bump operator version because we make `ambiguous` behavior clear.

fdwr:
@wschin: Hi Wei-Sheng. We can add an attribute for compat sake if older models need it, but I recommend that either way (like was done with the resampling operator [`Resize`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Resize) attribute `coordinate_transformation_mode`) we default to the corrected one in future opsets (that is default = `half_pixel`). The only reason frameworks have 2 is because they introduced the buggy one first 😅 (sad to have to encode their mistake too).

Regarding naming, the other frameworks call the attribute which controls the coordinate transforms and pixel offsets names like `aligned`, `align_corners`, or `half_pixel_centers`. For consistency of ONNX with itself, I propose using the existing attribute `coordinate_transformation_mode` with two legal values: "half_pixel", plus one to convey that only the *output* is offset like "output_half_pixel" (alas none of the existing `coordinate_transformation_mode`'s in the resampling operator capture this). We *could* also include `coordinate_transformation_mode` = `align_corners` for the sake of `tf.image.crop_and_resize(…)`, but I am reluctant to propagate that resampling algorithm as it's improper from a computer graphics perspective, and TF fixed the behavior for `resize` but not yet `crop_and_resize` (https://github.com/tensorflow/tensorflow/issues/6720). So I'll defer that until someone else pushes for it.

wschin:
Sounds good to me. Extending `coordinate_transformation_mode` is general enough to cover all unknown issues and default to the correct setting is good too. Agree that we only need to focus on this misalign in this PR.

onnx/backend/test/case/node/roialign.py
```diff
@@ -161,33 +161,76 @@ def export_roialign():  # type: () -> None
             [
                 [
                     [
-                        [0.4664, 0.4466, 0.3405, 0.5688, 0.6068],
-                        [0.3714, 0.4296, 0.3835, 0.5562, 0.3510],
-                        [0.2768, 0.4883, 0.5222, 0.5528, 0.4171],
-                        [0.4713, 0.4844, 0.6904, 0.4920, 0.8774],
-                        [0.6239, 0.7125, 0.6289, 0.3355, 0.3495],
+                        [0.5178, 0.3434, 0.3229, 0.4474, 0.6344],
```
gramalingam, line 164: Instead of having just the numbers, is it possible to have code that generates this? That is, some python code that can act as a reference implementation? I guess the original test-case perhaps pre-dates the requirement to use reference implementations to the extent possible.

fdwr: I'm not sure. Maybe [scipy.signal.resample](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample.html]) or [scipy.ndimage.zoom](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.zoom.html) would work 🤔 - I only compared against other frameworks like Detectron/PyTorch/TF.

For the 2nd smaller test case at least, I rather like having directly visible numbers, which make the test cases clearer and more immediately comprehensible (saying this as someone who's had to implement multiple ONNX operators for DML, and frequently referred back to Operators.md 😅). For the 1st case with the larger tensors, cutting that sea of numbers and replacing it with a numpy equivalent would be nice.

gramalingam: Sure, I agree that for the smaller example, visually matching input/output is useful. But for the larger it is harder. The code could also be specialized for different cases, if it makes the job easier.

fdwr: We only want dependencies on pure numpy for the ONNX backend, right? I couldn't find any function in `numpy` that resamples tensors (and the numpy [interpolation functionality](https://numpy.org/doc/stable/reference/generated/numpy.interp.html) works on a series of *coordinates*, not tensors), and playing with various scipy functions, `scipy.ndimage.zoom` and `scipy.ndimage.affine_transform` were the two that worked as expected, but those are scipy dependencies. :/

```python
import numpy
import scipy
from scipy import ndimage
from scipy import signal
from scipy import misc

numpy.set_printoptions(linewidth=120)

x = numpy.array(
        [[ 0, 1, 2, 3],
         [ 4, 5, 6, 7],
         [ 8, 9,10,11],
         [12,13,14,15]],
        dtype=numpy.float32
    )
print("input:\n", x, sep='')

# ✖numpy.interp - Unsuitable. Takes an array input coordinates, not a tensor.
# https://numpy.org/doc/stable/reference/generated/numpy.interp.html

# ✖scipy.interpolate.interp2d - Unsuitable. Takes an array of input points, not a tensor or 2D image.
# https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp2d.html

# ✖scipy.misc.imresize - Unsuitable. Deprecated and removed in later version. Also has dependency on PIL anyway.
# AttributeError: module 'scipy.misc' has no attribute 'imresize'
# https://docs.scipy.org/doc/scipy-1.2.0/reference/generated/scipy.misc.imresize.html
#y = scipy.misc.imresize(x, (8,8), 'bilinear')
#print("\nscipy.misc.imresize\n", y, sep='')

# ✖scipy.signal.resample - Unsuitable. Only takes a single axis.
y = scipy.signal.resample(x, num=8, axis=1)
print("\nscipy.signal.resample:\n", y, sep='')

# ✖scipy.ndimage.zoom grid_mode = False - Unsuitable. Aligns corners rather than pixel centers.
y = scipy.ndimage.zoom(x, zoom=2, order=1, grid_mode=False)
print("\nscipy.ndimage.zoom grid_mode=False:\n", y, sep='')

# ✔scipy.ndimage.zoom grid_mode = True - Suitable. Aligns half pixels.
y = scipy.ndimage.zoom(x, zoom=2, order=1, mode='nearest', grid_mode = True)
print("\nscipy.ndimage.zoom grid_mode=True:\n", y, sep='')

# ✔Suitable.
y = scipy.ndimage.affine_transform(x, matrix=numpy.array([[0.5,0],[0,0.5]]), offset=[-0.25, -0.25], output_shape=(8, 8), order=1, mode="nearest")
print("\nscipy.ndimage.affine_transform:\n", y, sep='')
```

```
input:
[[ 0.  1.  2.  3.]
 [ 4.  5.  6.  7.]
 [ 8.  9. 10. 11.]
 [12. 13. 14. 15.]]

✖scipy.signal.resample:
[[ 0.          0.08578646  1.          1.5         2.          2.9142137   3.          1.5       ]
 [ 4.          4.0857863   5.          5.5         6.          6.9142137   7.          5.5       ]
 [ 8.          8.085787    9.          9.5        10.         10.914213   11.          9.5       ]
 [12.         12.085787   13.         13.5        14.         14.914213   15.         13.5       ]]

✖scipy.ndimage.zoom grid_mode=False:
[[ 0.          0.42857143  0.85714287  1.2857143   1.7142857   2.142857    2.5714285   3.        ]
 [ 1.7142857   2.142857    2.5714285   3.          3.4285715   3.857143    4.285714    4.714286  ]
 [ 3.4285715   3.857143    4.285714    4.714286    5.142857    5.571429    6.          6.428571  ]
 [ 5.142857    5.571429    6.          6.428571    6.857143    7.285714    7.714286    8.142858  ]
 [ 6.857143    7.285714    7.714286    8.142858    8.571428    9.          9.428572    9.857142  ]
 [ 8.571428    9.          9.428572    9.857142   10.285714   10.714286   11.142858   11.571428  ]
 [10.285714   10.714286   11.142858   11.571428   12.         12.428572   12.857142   13.285714  ]
 [12.         12.428572   12.857142   13.285714   13.714286   14.142858   14.571428   15.        ]]

✔scipy.ndimage.zoom grid_mode=True:
[[ 0.    0.25  0.75  1.25  1.75  2.25  2.75  3.  ]
 [ 1.    1.25  1.75  2.25  2.75  3.25  3.75  4.  ]
 [ 3.    3.25  3.75  4.25  4.75  5.25  5.75  6.  ]
 [ 5.    5.25  5.75  6.25  6.75  7.25  7.75  8.  ]
 [ 7.    7.25  7.75  8.25  8.75  9.25  9.75 10.  ]
 [ 9.    9.25  9.75 10.25 10.75 11.25 11.75 12.  ]
 [11.   11.25 11.75 12.25 12.75 13.25 13.75 14.  ]
 [12.   12.25 12.75 13.25 13.75 14.25 14.75 15.  ]]

✔scipy.ndimage.affine_transform:
[[ 0.    0.25  0.75  1.25  1.75  2.25  2.75  3.  ]
 [ 1.    1.25  1.75  2.25  2.75  3.25  3.75  4.  ]
 [ 3.    3.25  3.75  4.25  4.75  5.25  5.75  6.  ]
 [ 5.    5.25  5.75  6.25  6.75  7.25  7.75  8.  ]
 [ 7.    7.25  7.75  8.25  8.75  9.25  9.75 10.  ]
 [ 9.    9.25  9.75 10.25 10.75 11.25 11.75 12.  ]
 [11.   11.25 11.75 12.25 12.75 13.25 13.75 14.  ]
 [12.   12.25 12.75 13.25 13.75 14.25 14.75 15.  ]]
```

gramalingam: @askhade : is there any issue or concern with using scipy? Having said that, I don't know if it is worthwhile introducing extra dependencies for this one utility. I took a look at the pooling operations, and they have the same issue (and just have inputs/outputs). Thanks @fdwr for the detailed response. I personally feel it may be okay to leave it as is.

gramalingam:
On a different note: I was looking at the documentation of the operator. Why does the "rois" input have a floating type, instead of integer type? Did they use T1 instead of T2 by mistake?

fdwr:
> On a different note: I was looking at the documentation of the operator. Why does the "rois" input have a floating type, instead of integer type? Did they use T1 instead of T2 by mistake?

@gramalingam : Floating point is intentional, as objects are identified at a subpixel alignment, which is why RoiAlign has more accuracy over simple integer-aligned RoiPooling. From the Mask-RCNN paper: "RoIPool first quantizes a floating-number RoI to the discrete granularity ... These quantizations introduce misalignments between the RoI and the extracted features. While this may not impact classification, which is robust to small translations, it has a large negative effect on predicting pixel-accurate masks."

image

gramalingam (approved):

onnx/backend/test/case/node/roialign.py
askhade: usually we either use numpy or custom python functions to generate test data...

PR #3431, opened 2021-04-16
Adding Bernoulli operator

hwangdeyu:
Adding new operator "Bernoulli" to ONNX.
This op is needed for HuggingFace Transformer model export from ORTModule.


onnx/defs/generator/defs.cc
```diff
@@ -896,4 +896,68 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Attr(
+            "seed",
+            "(Optional) Seed to the random generator, if not specified we will auto generate one.",
+            AttributeProto::FLOAT,
+            OPTIONAL_VALUE)
+        .Attr(
+            "dtype",
+            "The data type for the elements of the output tensor. if not specified, we will use "
+            "the data type of the input tensor.",
+            AttributeProto::FLOAT,
```
gramalingam, line 918: dtype should probably be an Int, not float.

gramalingam:
It looks like it might be possible to define this as a function using RandomUniformLike? E.g., using ```Where(random < input, 0, 1)``` or something like that. If so, it would be better to define it as a function.

onnx/defs/generator/defs.cc
hwangdeyu: Thanks for this comment. I tried to run a simple example of bernouli in torch as below:
 ```python
>>> a = torch.Tensor([0, 1])
>>> torch.bernoulli(a).type()
'torch.FloatTensor'
```
So the dtype should be float too?

Comment (gramalingam) [reply onnx/defs/generator/defs.cc:918]: The attribute is used to indicate what the type of the output should be (e.g., float, or double, or float16, etc.) But the attribute-value is an integer value from the enumeration TensorProto::DataType  (see https://github.com/onnx/onnx/blob/08afa047e90c33357bb6a13b7313105859bdd2c4/onnx/onnx.in.proto#L461 ). See, for example, this: https://github.com/onnx/onnx/blob/08afa047e90c33357bb6a13b7313105859bdd2c4/onnx/defs/generator/defs.cc#L411

File: onnx/defs/generator/defs.cc [hunk 2]
```diff
@@ -896,4 +896,76 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            }
+            propagateShapeFromInputToOutput(ctx, 0, 0);
+        })
+        .FunctionBody(FunctionBodyHelper::BuildNodes({
+            // nodes: {outputs, op, inputs, attributes}
+            {{"O1"}, 
+            "RandomUniform", 
+            {"input"},
+            {MakeAttribute("high", 1.0f), MakeAttribute("low", 0.f)}},
+            {{"O2"}, "Greater", {"01", "input"}},
```
gramalingam, line 969: Change "01" to "O1"? Actually, alternative names like X1 or T1 may be better, since it is less likely to be confused with numbers like 01.

onnx/defs/generator/defs.cc
```diff
@@ -896,4 +896,76 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
         }));
 
+static const char* Bernoulli_ver15_doc = R"DOC(
+    Draws binary random numbers (0 or 1) from a Bernoulli distribution. The input tensor should be
+    a tensor containing probabilities to be used for drawing the binary random number.
```
gramalingam, line 901: Please clarify if the input-tensor value specifies the probability of a 1 (I assume) or 0.

askhade:
How did you generate the test data? The <test case>.py should also be checked in along with this PR

askhade:
this op should be registered in operator_sets.h

askhade:
please also add shape inference test for this op

onnx/defs/generator/defs.cc
hwangdeyu: This means the output of Bernoulli is 0 or 1, input tensor is a values from range [0, 1].
As [PyTorch docs description](https://pytorch.org/docs/stable/generated/torch.bernoulli.html).
I completed the description in latest commit.

onnx/defs/generator/defs.cc
hwangdeyu: Thanks for very useful comments!
Added the dtype attribute and cast output as well in FunctionBodyBuilde.

hwangdeyu:
> How did you generate the test data? The .py should also be checked in along with this PR

Thanks for the good catch.
I'm sorry for delete the test and registered code in once a rebase master branch and force-pushed git behavior.
now I added those logic again.

hwangdeyu:
@gramalingam @askhade  
All the CI failures are passed, do you have other comments?
Thanks!

hwangdeyu:
@postrational Would you also convenient to help us review this Bernoulli spec PR? Thanks!

onnx/defs/generator/defs.cc
gramalingam: Sorry I was unclear. I suggest changing the second sentence to
```
The input tensor should be a tensor containing probabilities p (a value in the range [0,1]) to be
used for drawing the binary random number, where an output of 1 is produced with probability
p and an output of 0 is produced with probability (1-p).
```

hwangdeyu: Done, this description is more reasonable and clearer.

hwangdeyu:
@gramalingam @askhade 
Please let me know if we can approve and merge the PR.
Thanks!

onnx/backend/test/case/node/bernoulli.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import unicode_literals
+
+import numpy as np  # type: ignore
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+def bernoulli_reference_implementation(x, dtype):  # type: ignore
```
gramalingam, line 15: Let us add a comment here, something like:
```
This example and test-case is for informational purpose. The operator is non-deterministic and may not produce
the same values in different implementations (even if a seed is specified).
```

onnx/defs/generator/defs.cc
```diff
@@ -896,4 +896,115 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
         }));
 
+static const char* Bernoulli_ver15_doc = R"DOC(
+Draws binary random numbers (0 or 1) from a Bernoulli distribution. The input tensor should be a tensor
+containing probabilities p (a value in the range [0,1]) to be used for drawing the binary random number,
+where an output of 1 is produced with probability p and an output of 0 is produced with probability (1-p).
```
gramalingam, line 902: I suggest adding a line as below here:
```
This operator is non-deterministic and may not produce the same values in different
implementations (even if a seed is specified).
```

hwangdeyu: I'm little bit confused. If we specified the seed, the RandomUniformLike op will generate the same value between 0 to 1. Then we can use Greater op to compare input values to product same results right?

gramalingam: I think the same comment applies to RandomUniformLike. The point is: is there a guarantee that different implementations of random number generation (in GPUs vs CPUs, in different OS platforms, etc.) all use the same pseudo-random-number generator algorithm? That seems hard to guarantee. So, it is better to add the above comment. If you know that all backends use the same pseudo-random-number-generator (PRNG), please let us know what this PRNG is.

gramalingam (approved):

PR #3432, opened 2021-04-16
Fix Batchnorm type inference for mean and variance and add a test

jcwchen:
**Description**
- Type inference bug fix: propagate output mean/variance from input mean/variance instead of input X.
- Add a test for different types of input (X v.s. mean and variance).

**Motivation**
Since the data type of mean and variance can be different from X, the output of mean and variance should be propagated from input mean and variance instead.

### DIFF
--- a/onnx/test/shape_inference_test.py
+++ b/onnx/test/shape_inference_test.py
@@ -3488,8 +3488,8 @@ class TestShapeInference(unittest.TestCase):
     def test_batch_norm_train_with_diff_type(self):  # type: () -> None
         graph = self._make_graph(
             [('x', TensorProto.FLOAT16, (3, 4, 5, 6, 7)),
-             ('scale', TensorProto.FLOAT, (4,)),
-             ('b', TensorProto.FLOAT, (4,)),
+             ('scale', TensorProto.FLOAT16, (4,)),
+             ('b', TensorProto.FLOAT16, (4,)),
              ('input_mean', TensorProto.FLOAT, (4,)),
              ('input_var', TensorProto.FLOAT, (4,))],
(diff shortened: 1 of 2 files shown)


onnx/test/shape_inference_test.py
```diff
@@ -3485,6 +3485,21 @@ def test_batch_norm_train_dim_param(self):  # type: () -> None
                                       make_tensor_value_info('output_var', TensorProto.FLOAT, ('C',)),  # type: ignore
                                       ])
 
+    def test_batch_norm_train_with_diff_type(self):  # type: () -> None
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT16, (3, 4, 5, 6, 7)),
+             ('scale', TensorProto.FLOAT, (4,)),
+             ('b', TensorProto.FLOAT, (4,)),
```
gramalingam, line 3492: Shouldn't scale and b be same type as x?

jcwchen: Yes. I am also thinking about this... For shape_inference, do we need to add a strict check for the consistency among x, scale or b?

gramalingam: Yes. Feel free to add it, but I think we need a more general solution since this issue is much more widespread problem: eg., same problem exists with all arithmetic ops like Add, Mul, etc. (There is a more general solution, used inside ORT. We just need to port it to ONNX.) So, we can also solve that as a separate PR later.

jcwchen: Good point. I was also checking other operators to seek for similar implementation but failed. I agreed with you we should do it globally. I will keep this PR simple and let's solve this general issue in another PR. Thank you @gramalingam!

gramalingam (approved):

PR #3440, opened 2021-04-20
[Dup] Update spec for Convtranspose to make it sync 

jcwchen:
**Description**
** Create a new PR to solve DCO issues with old PR: https://github.com/onnx/onnx/pull/3019
These 2 PRs are totally the same one.

Update the spec to make it synchronous. (Description part vs if-else part)
Description part should be corrected. In addition, onnxruntime should also be corrected.: https://github.com/microsoft/onnxruntime/blob/a5530358c9236f20e2ad3d6ce968219b21426874/onnxruntime/core/providers/cpu/nn/conv_transpose_attributes.h#L184)
The behavior of SAME_UPPER and SAME_LOWER should be same as the same attribute (auto_pad) used by other pooling or conv related operators.

**Motivation and Context**
https://github.com/onnx/onnx/issues/1609,  https://github.com/onnx/onnx/pull/2097
The spec for ConvTranspose is not synchronous.


gramalingam:
While we are updating this, can we clarify the documentation for padding a bit more? I find it difficult to understand all the combinations of attributes that are allowed and their combined effect. Can we explicitly indicate the order in which the attributes are considered to determine the output-size? (Which attributes are ignored or forbidden if some other attribute is specified?)

jcwchen:
> While we are updating this, can we clarify the documentation for padding a bit more? I find it difficult to understand all the combinations of attributes that are allowed and their combined effect. Can we explicitly indicate the order in which the attributes are considered to determine the output-size? (Which attributes are ignored or forbidden if some other attribute is specified?)

I agreed since the combined behaviour in current document is quite ambiguous. However, have this ambiguous been well-defined yet? If not, it will take some time to organize the whole behaviour and actually I would like to forward this PR faster because this bug fix in ORT wants to consume this update (updated test data). In that case, perhaps the document clarification can be done by another PR. Thanks.

gramalingam (approved):

PR #3441, opened 2021-04-20
Remove comment about ONNX being very early stage.

garymm:
This seems misleading to me.


gramalingam:
Thanks for the fix. I think you need to generate the ONNX-ML versions of the proto files as well.

garymm:
> Thanks for the fix. I think you need to generate the ONNX-ML versions of the proto files as well.

Done.

As an aside: running `gen_proto.py` with no flags created a bunch of unexpected diffs, which made it a bit confusing to figure out how to regenerate things properly.
Should we change the default flag values to match what is actually checked in at master?

gramalingam (approved):

PR #3443, opened 2021-04-20
Clarify CONTRIBUTING.md documentation generation steps

fdwr:
Let's make this clearer so people don't waste an hour (like me 😅) wondering why running the script yields no changes to Operators.md. For some reason, the much more rarely updated Operators-ml.md is the default update target when `ONNX_ML` is not defined, rather than the one everyone typically wants to update. 🙃 Also, explicitly relist the install step rather than referring back to an earlier step which is actually executed from a different directory (which confuses things).

FYI @jcwchen

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -32,9 +32,11 @@ Then, after you have made changes to Python and C++ files:
 
 ## Generated operator documentation
 
-[Operator docs in Operators.md](Operators.md) are automatically generated based on C++ operator definitions. To refresh these docs, remember to re-install (see above) and then run the following command from the repo root and commit the results:
+[Operator docs in Operators.md](Operators.md) are automatically generated based on C++ operator definitions and backend Python snippets. To refresh these docs, run the following commands from the repo root and commit the results. Note `ONNX_ML=0` updates Operators.md whereas `ONNX_ML=1` updates Operators-ml.md:
 
 ```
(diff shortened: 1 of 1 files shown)


jcwchen (approved):
LGTM. @fdwr Thank you for the improvement!

fdwr:
@jcwchen : Welcome. Do editorial changes need review from a specific sig reviewer, or anyone? I'd propose also that `python onnx/defs/gen_doc.py` just take an *explicit* parameter, rather than using a wonky environment variable `ONNX_ML`, but I'm unsure who to propose that idea too.

jcwchen:
> @jcwchen : Welcome. Do editorial changes need review from a specific sig reviewer, or anyone? I'd propose also that `python onnx/defs/gen_doc.py` just take an _explicit_ parameter, rather than using a wonky environment variable `ONNX_ML`, but I'm unsure who to propose that idea too.

Since it is related to infra, you can propose it to @onnx/sig-archinfra-approvers . Thanks

askhade (approved):

askhade:
> @jcwchen : Welcome. Do editorial changes need review from a specific sig reviewer, or anyone? I'd propose also that `python onnx/defs/gen_doc.py` just take an _explicit_ parameter, rather than using a wonky environment variable `ONNX_ML`, but I'm unsure who to propose that idea too.

you can create a PR and add arch infra sig

PR #3446, opened 2021-04-22
Python API Overview: minor improvements for clarity

garymm:
Also use repo-relative links so that they'll work as expected locally or
in a branch.

### DIFF
--- a/onnx/examples/check_model.ipynb
+++ b/onnx/examples/check_model.ipynb
@@ -81,14 +81,18 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "The model is checked!\n"
+      "The model is valid!\n"
      ]
     }
    ],
    "source": [
     "# Check the model\n",
-    "onnx.checker.check_model(onnx_model)\n",
-    "print('The model is checked!')"
+    "try:\n",
+    "    onnx.checker.check_model(onnx_model)\n",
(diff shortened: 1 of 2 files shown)


askhade (approved):

onnx/examples/check_model.ipynb
```diff
@@ -92,8 +92,10 @@
    ],
    "source": [
     "# Check the model\n",
-    "onnx.checker.check_model(onnx_model)\n",
-    "print('The model is checked!')"
+    "try:\n",
+    "    onnx.checker.check_model(onnx_model)\n",
+    "else:\n",
+    "    print('The model is valid!')"
```
jcwchen, line 98: Missing the following code?
```
except onnx.checker.ValidationError as e:
    print('The model is invalid: %s' % e)
```

garymm: Previous to this change, and now, the behavior is the same: if the model is invalid, an exception will be raised that makes that obvious.
So I didn't see a need to catch the exception.
If you prefer I can add the catch, but could you please explain why?

jcwchen: Thank you for the explanation. Actually the CIs will validate .ipynb and so a try needs a except.

jcwchen (approved):

PR #3448, opened 2021-04-22
Add requirement.txt for ONNX's dependency

jcwchen:
**Description**
- Add ONNX's dependency as requirement.txt
  - requirement.txt: for build
  - requirement-release.txt: for release (fixed dependency version)
  - requirement-dev.txt : for development (testing)
- To improve maintainability, reuse requirement-release.txt in all release CIs to sync dependency version easily
- Add missing `-q` for each pip install

**Motivation**
requirement*.txt is useful for users to understand ONNX's dependency.


.azure-pipelines/MacOS-CI.yml
```diff
@@ -30,12 +30,26 @@ jobs:
       versionSpec: '$(python.version)'
 
   - script: |
-      git submodule update --init --recursive
-      python -m pip install --upgrade setuptools
-      python -m pip install numpy
-      conda install -y -c conda-forge pybind11 protobuf
```
annajung, line 36: I noticed that pybind11 was removed, is this no longer necessary here?

jcwchen: Yes, actually ONNX uses pybind11 as a submodule so basically it should not need to manually install pybind11 for ONNX

requirements.txt
```diff
@@ -0,0 +1,4 @@
+numpy >= 1.16.6
+protobuf >= 3.16.0
```
rajeevsrao, line 2: Based on discussion with @askhade @jcwchen to remove the minimum version requirement here for consistency with past releases. Also check protobuf version requirements of tf/torch.

rajeevsrao (approved):

requirements.txt
jcwchen: Thanks for pointing out! In some converters (onnx-tensorrt, onnx-mlir) they do have protobuf version requirements... To grant users flexibility for using different versions of protobuf, remove the minimum version requirement for ONNX here.

askhade (approved):

PR #3449, opened 2021-04-22
Minor improvements to versioning doc.

garymm:
Mostly for improved readability.

Biggest changes:
* In the operator set versioning example, remove references to the
  concept of deprecation. The versioning system does not actually have
  a deprecation mechanism. In the example, the operator A is not present
  in operator set version 4. Models that wish to use it will have to
  specify an opset version < 4.
* Clarifying what should be done to model versions
  when accuracy or performance changes. The previous text did not
  include any clear guidance.

### DIFF
--- a/docs/Versioning.md
+++ b/docs/Versioning.md
@@ -1,43 +1,40 @@
 <!--- SPDX-License-Identifier: Apache-2.0 -->
 
-# ONNX versioning
+# ONNX Versioning
 
-This document describes the rules for versioning ONNX. Like the rest of the ONNX
-specification, MUST, SHOULD et al are used consistent with [RFC2119](https://tools.ietf.org/html/rfc2119).
+This document describes the rules for versioning ONNX. MUST, SHOULD et al are used consistent with [RFC2119](https://tools.ietf.org/html/rfc2119).
 
 ## Versioning Principles
 
 ONNX defines the versioning policy and mechanism for three classes of entities:
 
-* The abstract model for graphs and operators and the concrete format that represents them. These are always versioned atomically and are referred to as the *IR version*.
+* The [intermediate representation (IR) specification](IR.md), which is the abstract model for graphs and operators and the concrete format that represents them. These are always versioned atomically and are referred to as the *IR version*.
 * Operator specifications that may be referenced by a given ONNX graph. We refer to this as the *operator version*.
-* A defined/trained model that defines a specific graph in terms of specific operators. We refer to this version as the *model version*.
+* A defined/trained model that defines a specific graph in terms of specific operators. We refer to this as the *model version*.
 
-The versioning of all three of these entity types is distinct and largely independent. That is, the ONNX IR format evolves at a different rate than the set operators defined by ONNX – in which the former will version much slower than the latter.
+The versioning of all three of these entity types is distinct and largely independent. The IR specification evolves at a different (generally slower) rate than the operator specifications. Model versions are entirely independent of the other two versions.
 
-While the versioning mechanisms are clearly specified in this document, specific policies for version management are mandated only for IR version and operator version. For model versioning, they are merely recommendations. For model version, ONNX users and systems MAY follow whichever local customs make sense; however, to facilitate easily managing shared collections of ONNX models, they SHOULD adhere to the policies described under model versioning.
+Specific policies for version management are mandated only for IR version and operator version. For model versioning, they are merely recommendations. For model versioning, ONNX users and systems MAY follow whichever local customs make sense; however, to facilitate easily managing shared collections of ONNX models, they SHOULD adhere to the policies described under model versioning.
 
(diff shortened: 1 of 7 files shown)


garymm:
@gramalingam PTAL

docs/Versioning.md
```diff
@@ -1,44 +1,41 @@
...
-This document describes the rules for versioning ONNX. Like the rest of the ONNX
-specification, MUST, SHOULD et al are used consistent with [RFC2119](https://tools.ietf.org/html/rfc2119).
+This document describes the rules for versioning ONNX. MUST, SHOULD et al are used consistent with [RFC2119](https://tools.ietf.org/html/rfc2119).
 
 ## Versioning Principles
 
 ONNX defines the versioning policy and mechanism for three classes of entities:
 
-* The abstract model for graphs and operators and the concrete format that represents them. These are always versioned atomically and are referred to as the *IR version*.
+* The [interemdiate representation (IR) specification](IR.md), which is the abstract model for graphs and operators and the concrete format that represents them. These are always versioned atomically and are referred to as the *IR version*.
```
gramalingam, line 11: interemdiate => intermediate

gramalingam:
On a related note: the existing description of operator versioning (as a three-tuple ```(domain, op_type, op_version)``` is a bit misleading, as well as the comment that nodes refer to them by the 3-part identifier. It would be helpful to fix it as well.

garymm:
> On a related note: the existing description of operator versioning (as a three-tuple `(domain, op_type, op_version)` is a bit misleading, as well as the comment that nodes refer to them by the 3-part identifier. It would be helpful to fix it as well.

I tried to clarify. LMK if this is what you had in mind.

docs/Versioning.md
```diff
@@ -96,15 +91,15 @@ operator id MUST be greater than any extant `op_version` for the
...
 
 ### Operator sets
 
-ONNX uses operator sets to group together immutable operator specifications. An ONNX operator set specifies both the domain of all operators it includes, as well as a version (referred to as the `opset` version). The opset version is largely independent of the version field of the operators it includes. When the inventory of a given operator set changes either by addition or removal, its opset version MUST increase. Moreover, the opset version MUST be no less than the highest operator version number in the set.
+ONNX uses operator sets to group together immutable operator specifications. An operator set specifies both the domain of all operators it includes, as well as a version (referred to as the `opset` version). The opset version is largely independent of the version field of the operators it includes. When the inventory of a given operator set changes either by addition or removal, its opset version MUST increase. Moreover, the opset version MUST be no less than the highest operator version number in the set.
```
gramalingam, line 100: I find this (pre-existing) description also a bit confusing. How about something like:
```
ONNX uses operator sets to group together immutable operator specifications. An operator set represents
a specific version of a domain, indicated by a pair (domain, version). This represents the set of all operators
belonging to the specified domain with the specified version (referred to as the `opset` version). When the
inventory of a given operator set changes either by addition or removal, its opset version MUST increase.
```

garymm: Done.

garymm:
I took another look at the text in the Operator versioning section and tried to clarify it. In particular, it used to reference `op_version`, but from the protocol buffer files, it seems like op version is not a concept that currently exists in the system. I replaced with `since_version`.

However I think the text may still be incomplete as I'm confused as to how `since_version` works as expected when operators can be updated to have different semantics. In the example, `A` in opset version 1 and `A` in opset version 3 are different.
But maybe that's an implementation detail and doesn't need to be in the spec?

garymm:
Regarding your comment from slack about model versioning guidelines:
Happy to remove it altogether if you think that's better. LMK.

gramalingam:
Re. versioning: a later version of an op implicitly supercedes an earlier version. Thus, if we have an op Foo with since_version 1 and an op Foo with since_version 3, then a model that uses opset-version 2 would resolve Foo to Foo with since_version 1, while a model that uses opset-version 4 would resolve Foo to Foo with since_version 3 (assuming, of course, that there is no Foo with since_version 4). Concepually, we could say (Foo, version 1) and (Foo, version 2) are the same, while (Foo, version 3) is different.

gramalingam:
About model versioning guidelines: this is fine, since the document contains a disclaimer that this part is not normative, but just recommended practice.

gramalingam (approved):

gramalingam:
I think onnx.proto3 need to be generated and included in the commit.

garymm:
> I think onnx.proto3 need to be generated and included in the commit.

This command doesn't produce any diffs:
`./onnx/gen_proto.py --ml --lite`

Which diff to onnx.proto3 do you expect?

gramalingam:
Sorry, I meant onnx-operators.proto3. Same change as for the other proto files in your PR. I think that's why the CI is failing.

garymm:
Ah I see it now. Done! Time to try the auto-merge feature?

garymm:
> @garymm Did you manually change this file? CIs are failing because they are seeing diffs : diff --git a/onnx/onnx-operators.proto b/onnx/onnx-operators.proto

Sorry, I think I managed to regenerate things properly now.
Is there a single command that will do it? I had to string together a bunch of commands, involving different `gen_proto.py` flags `git restore`.

askhade (approved):

PR #3459, opened 2021-04-30
Fix typo in LSTM shape for layout=1

axmat:
Fix the shape of initial_h, Y_h, initial_c and Y_c in the documentation of the LSTM operator for layout=1.


gramalingam:
@matteosal , would be great if you could take a look.

matteosal:
This is right, the documentation was wrong. I've also double checked that other recurrent layers and the shape inference code are good.

axmat:
Thanks @gramalingam @matteosal for the review.

gramalingam (approved):

gramalingam:
Hi @axmat : it looks like the CI is failing. I think may be Changelog also needs to be added after generating the operator documentation.

PR #3461, opened 2021-04-30
Add UnionShape for SparseTensor TypeProto

yuslepukhin:
This is a missing utility for SparseTensor type proto


onnx/defs/shape_inference.cc
```diff
@@ -0,0 +1,101 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "shape_inference.h"
+
+namespace ONNX_NAMESPACE {
+
+void UnionShapeInfo(const TensorShapeProto& source_shape, TensorShapeProto& target_shape) {
```
gramalingam, line 9: Thanks for the refactoring and moving it into a .cc file. May be useful to add a comment above that source_shape and target_shape are assumed to have same dim_size().

gramalingam:
LGTM, thanks! Just a minor comment above.

askhade (approved):
LGTM

onnx/defs/shape_inference.cc
yuslepukhin: I will remove it from the header

gramalingam (approved):

Issue #3456, opened 2021-04-28
Shape inference for 'Squeeze' not working if no axes are provided


thilow:
# Bug Report

There is a well defined behavior for the Squeeze op even when now axes are explicitly provided:

_...  If axes is not provided, all the single dimensions will be removed from the shape.  ..._

see [here](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Squeeze).

The shape inference does not work in this case and doesn't produce any shape for the ops output.

### System information

onnx.__version__: 1.9.0

### Reproduction instructions
The script below illustrates the issue. It creates models with only one Squeeze op. In the first model the axes are explicitly provided and shape inference works. In the second case no axes are provided and shape inference does not work

```
#%%
import onnx
from onnx import helper
from onnx import AttributeProto, TensorProto, GraphProto, shape_inference


#%% Create a model with just one "Squeeze". No 'axis' is explicitly provided
X = helper.make_tensor_value_info('X', TensorProto.FLOAT, [1, 42,42])
Y = helper.make_tensor_value_info('Y', TensorProto.FLOAT,None)

#Create model where axes are explictly set.
node_def = helper.make_node('Squeeze', ['X'], ['Y'], axes=[0])
graph_def = helper.make_graph([node_def],'test-model',[X],[Y])
model_def = helper.make_model(graph_def, producer_name='pytorch')
model_def.opset_import[0].version=12

#run onnx shape inference: 
m_inferred = shape_inference.infer_shapes(model_def)
print("This case works fine: ",m_inferred.graph.output[0])


#%% Create a model with just one "Squeeze". No 'axis' is explicitly provided
X = helper.make_tensor_value_info('X', TensorProto.FLOAT, [1, 42,42])
Y = helper.make_tensor_value_info('Y', TensorProto.FLOAT,None)

#Create model where axes are NOT explictly set.
node_def = helper.make_node('Squeeze', ['X'], ['Y'])
graph_def = helper.make_graph([node_def],'test-model',[X],[Y])
model_def = helper.make_model(graph_def, producer_name='pytorch')
model_def.opset_import[0].version=12

#run onnx shape inference: 
m_inferred = shape_inference.infer_shapes(model_def)
print("No shape is inferred when running shape inference: ",m_inferred.graph.output[0])
```
Below is the output of the script
```
This case works fine:  name: "Y"
type {
  tensor_type {
    elem_type: 1
    shape {
      dim {
        dim_value: 42
      }
      dim {
[...]

PR #3465, opened 2021-05-03
Fix shape inference for Squeeze without axes

jcwchen:
**Description**
Add shape inference for Squeeze without axes: keep dimension if it is not equal to 0.

**Motivation**
https://github.com/onnx/onnx/issues/3456 Currently there is no shape inference for Squeeze without axes.


onnx/defs/tensor/defs.cc
```diff
@@ -1655,8 +1646,12 @@ ONNX_OPERATOR_SET_SCHEMA(
                 return axis < 0 ? axis + input_ndim : axis;
               });
 
-          for (int i = 0, j = 0; i < input_ndim; ++i) {
-            if (std::find(axes.begin(), axes.end(), i) != axes.end()) {
+          for (int i = 0; i < input_ndim; ++i) {
```
gramalingam, line 1649: if-then-else may need to be refactored. If axes is not specified, and input_shape.dim(i) has a symbolic value, it will fall into the else branch, which seems wrong.

gramalingam: The case where (not_specify_axes) is a bit more involved. If there is even one unknown input-dimension, we can't really set any of the other dimensions, since the output rank itself becomes unknown.

onnx/defs/tensor/defs.cc
```diff
@@ -1634,19 +1632,12 @@ ONNX_OPERATOR_SET_SCHEMA(
             axes = ParseData<int64_t>(axes_proto);
```
gramalingam, line 1632: But this may not be a compile-time constant. Need to handle the case where axes_proto is null

jcwchen: I did see there is a `axes_proto == nullptr` condition here... Could you please elaborate your case? Thank you!

gramalingam: Hmmm. I am confused. Not sure what code I was looking at! Please ignore this comment!

onnx/defs/tensor/defs.cc
```diff
@@ -1617,17 +1617,20 @@ ONNX_OPERATOR_SET_SCHEMA(
 
           std::vector<int64_t> axes;
           size_t num_inputs = ctx.getNumInputs();
+          bool not_specify_axes = false;
```
askhade, line 1620: nit: change to axes_not_specified

jcwchen: Updated. Thanks

onnx/defs/tensor/defs.cc
```diff
@@ -1638,9 +1641,19 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          for (int i = 0; i < input_ndim; ++i) {
+            if (!input_shape.dim(i).has_dim_value()) {
+              return;
+            }
+          }
 
-          for (int i = 0, j = 0; i < input_ndim; ++i) {
-            if (std::find(axes.begin(), axes.end(), i) != axes.end()) {
+          for (int i = 0; i < input_ndim; ++i) {
+            if (axes_not_specified && input_shape.dim(i).has_dim_value() &&
```
gramalingam, line 1652: nit: ```input_shape.dim(i).has_dim_value()``` can be dropped since it is already checked in previous loop.

jcwchen: Good point. I have removed two of them. Thanks

onnx/defs/tensor/defs.cc
```diff
@@ -1638,18 +1641,25 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 
-          for (int i = 0, j = 0; i < input_ndim; ++i) {
-            if (std::find(axes.begin(), axes.end(), i) != axes.end()) {
-              if (input_shape.dim(i).has_dim_value() &&
-                  input_shape.dim(i).dim_value() != 1) {
+          for (int i = 0; i < input_ndim; ++i) {
+            if (axes_not_specified && input_shape.dim(i).dim_value() == 1) {
+                // if axes not specified, do not keep shape if the dimension is equal to one
+                continue;
+            } else if (std::find(axes.begin(), axes.end(), i) != axes.end()) {
```
gramalingam, line 1655: minor nit: add ```!axes_not_specified``` as an extra condition above. (I think behavior should be the same since axes will be an empty vector if axes_not_specified ... but just for documentation/robustness.)

jcwchen: Added. Thanks

gramalingam (approved):

PR #3467, opened 2021-05-04
Generate the ONNX and ONNX-ML operator doc in one run

jackwish:
We need to setup env var `ONNX_ML` and run `gen_doc.py` twice to update the operator docs if both _ONNX_ and _ONNX-ML_ doc need to be updated.

With this change, `python onnx/defs/gen_doc.py` generates _ONNX_ and _ONNX-ML_ doc in one run.


onnx/defs/gen_doc.py
```diff
@@ -371,10 +375,5 @@ def main(args):  # type: (Type[Args]) -> None
...
 
 if __name__ == '__main__':
-    base_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
-    docs_dir = os.path.join(base_dir, 'docs')
-
-    class Args(object):
-        output = os.path.join(docs_dir, 'Operators' + ext)
-        changelog = os.path.join(docs_dir, 'Changelog' + ext)
-    main(Args)
+    gen_doc(False)
```
askhade, line 378: for better readability change this to 
gen_doc(is_ml=False)
gen_doc(is_ml=True)

onnx/defs/gen_doc.py
```diff
@@ -222,8 +220,14 @@ def support_level_str(level):  # type: (OpSchema.SupportType) -> Text
         "<sub>experimental</sub> " if level == OpSchema.SupportType.EXPERIMENTAL else ""
 
 
-def main(args):  # type: (Type[Args]) -> None
-    with io.open(args.changelog, 'w', newline='') as fout:
+def gen_doc(is_ml):  # type: (bool) -> None
```
askhade, line 223: nit: change is_ml -> is_ml_domain ?

onnx/defs/gen_doc.py
```diff
@@ -49,9 +45,10 @@ def format_name_with_domain(domain, schema_name):  # type: (Text, Text) -> Text
     return schema_name
 
 
-def format_versions(versions):  # type: (Sequence[OpSchema]) -> Text
+def format_versions(versions, fname_ext):  # type: (Sequence[OpSchema]) -> Text
```
askhade, line 48: update type info
# type: (Sequence[OpSchema], Text) -> Text

askhade (changes requested):
Added a few comments. Most are nit picks... 1 regarding type info needs to be fixed to turn CIs green

jackwish:
Hi @askhade thank you for the quick review. I have fixed your concerns, please help to check.

askhade (approved):

jackwish:
Hmmm... Cannot reproduce the error with python 3.8. When trying to repro with the steps of the pipeline `onnx-lite` build, fail to build. Is it necessary to build protobuf online for the CI test?

onnx/defs/gen_doc.py
```diff
@@ -371,10 +375,5 @@ def main(args):  # type: (Type[Args]) -> None
...
 if __name__ == '__main__':
-    base_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
-    docs_dir = os.path.join(base_dir, 'docs')
-
-    class Args(object):
-        output = os.path.join(docs_dir, 'Operators' + ext)
-        changelog = os.path.join(docs_dir, 'Changelog' + ext)
-    main(Args)
+    gen_doc(is_gen_ml_doc=False)
+    gen_doc(is_gen_ml_doc=True)
```
jcwchen, line 379: Currently the CIs failed with ONNX_ML=0 because ONNX_ML won't create ONNX_ML schema and the produced Operator-ml.md and Changelog-ml.md would be basically empty. 

Therefore I would say the behaviour of gen_doc.py should be as follows:
1. If ONNX_ML=1, it will update both general and *-ml.md documents
2. if ONNX_ML=0, it will only update general document

Then, it should help resolving the current CI failures. Thank you!

jackwish:
Close since run twice for _ONNX_ and _ONNX-ML_ should be by design. Thanks for the review!

askhade:
@jackwish : I think you missunderstood the comment. Your PR is a good enhancement for the current scenario it just needs a little change. 

When ONNX_ML is set to 0 when building onnx we do not build ml ops and therefore if you run the gen doc script it will create an empty operators-ml.md file. This is what needs to be fixed. In CIs we do set ONNX_ML to 0 and after running gen_doc.py it produces as empty operators-ml.md and therefore CIs are failing.

I hope you will reopen it and make this fix.

PR #3468, opened 2021-05-04
Revert #3429

askhade:
Revet ""Fix RoiAlign backend test case output, and add new test case for an identity slice. (#3429)"

This reverts commit 82ddc566b950d05fdccb70c4579930f4014ece7f.

I merged too early


gramalingam:
Is there some pending change required (just to know)?

askhade:
> Is there some pending change required (just to know)?

Wei-Sheng had a few open comments regarding adding an optional attribute for alignment to maintain backward compatibility and bumping the op version. With this change more tests also need to be added which include this attr

gramalingam (approved):

gramalingam:
I got the impression they agreed to decouple that change from this change. May be they can comment.

gramalingam:
@fdwr @wschin for your feedback

wschin (approved):
Thanks! Either way works for me.

fdwr:
>I got the impression they agreed to decouple that change from this change. May be they can comment.
Also, either way works for me. I can just re-cherry-pick what I already had again later.

fdwr (approved):

PR #3470, opened 2021-05-05
Add a PR label to manually trigger release CIs

jcwchen:
**Description**
Add a PR label, run release CIs, to manually trigger release CIs.

**Motivation**
Current PRs in ONNX  are not run by release CIs by default, and cannot be manually triggered for release CIs. ONNX should provide  a way to trigger release CIs on demand.


askhade:
What is the motivation for this? We do run release CIs when PRs are pushed to master right

jcwchen:
> What is the motivation for this? We do run release CIs when PRs are pushed to master right

I think it's good for development related to release CIs. Sometimes we want to test release CIs in certain PR but it is not for merging into rel-* and developers need to change code in release CIs to manually trigger (and then change back before merge)... Adding a label to trigger release  CIs would be more convenient and safer (no need to change CI code in a PR).

Another thought: Since current release CIs are much faster from the beginning, actually I am thinking that perhaps we can move all AzurePipelines (all config combination) into GitHub Action and make it be run by every PR... Then this PR is not needed. Thanks.

jcwchen:
For instance, we can prevent issues like this: https://github.com/onnx/onnx/pull/3485. If a PR does change the release CI, the PR can be tested by manually trigger before merge.

.github/workflows/release_linux_i686.yml
```diff
@@ -7,11 +7,21 @@ on:
   push:
     branches: [master, rel-*]
   pull_request:
-    branches: [rel-*]
+    branches: [master, rel-*]
```
askhade, line 10: why add master here?

jcwchen: Add PR for master here to let the following code (L24) be able to consider which labels that PR is using. If the PR for master has a label "run release CIs", it will run whole release CIs; If not, it won't.

askhade (approved):

PR #3472, opened 2021-05-06
Fix misleading global pooling test-case code

calvinmccarter-at-lightmatter:
While the current code happens to be correct for 2d inputs, because `spatial_shape` happens to equal 2, the logic is incorrect for inputs that are 3d or greater. We must take the average / max over "all-but-the-first-two" dimensions, not "only-the-last-two" dimensions.


calvinmccarter-at-lightmatter:
This is a fix for issue #3478

calvinmccarter-at-lightmatter:
@fumihwh - can you verify this? thanks!

onnx/backend/test/case/node/globalmaxpool.py
```diff
@@ -24,7 +24,7 @@ def export():  # type: () -> None
         )
         x = np.random.randn(1, 3, 5, 5).astype(np.float32)
         spatial_shape = np.ndim(x) - 2
-        y = np.max(x, axis=tuple(range(spatial_shape, spatial_shape + 2)))
+        y = np.max(x, axis=tuple(range(2, spatial_shape + 2)))
```
gramalingam, line 27: Thanks for the fix. Is it possible to use the ```keepdims``` parameter of ```np.max``` and eliminate the code below?

onnx/backend/test/case/node/globalaveragepool.py
```diff
@@ -23,7 +23,7 @@ def export():  # type: () -> None
         )
         x = np.random.randn(1, 3, 5, 5).astype(np.float32)
         spatial_shape = np.ndim(x) - 2
-        y = np.average(x, axis=tuple(range(spatial_shape, spatial_shape + 2)))
+        y = np.average(x, axis=tuple(range(2, spatial_shape + 2)))
```
gramalingam, line 26: If we use ```np.mean```, can we use the ```keepdims``` parameter, and eliminate the code below?

onnx/backend/test/case/node/globalmaxpool.py
calvinmccarter-at-lightmatter: Yes to both - done!

gramalingam (approved):
LGTM, thanks!

jcwchen (commented):
Thank you for the fix! Could you please use `python onnx/defs/gen_doc.py` to update the Operators.md as well? (That document uses the old test script now)

gramalingam:
Also, please do verify that the generated binary data for these test-cases hasn't changed (as expected).

calvinmccarter-at-lightmatter:
> Also, please do verify that the generated binary data for these test-cases hasn't changed (as expected).

@gramalingam - besides verifying that pytest succeeds, is there anything else I should do for this?

jcwchen:
> > Also, please do verify that the generated binary data for these test-cases hasn't changed (as expected).
> 
> @gramalingam - besides verifying that pytest succeeds, is there anything else I should do for this?

I have just checked the produced model after your modification. They are basically identical so you should not need to update them. You could use `python onnx/backend/test/cmd_tools.py generate-data --op_type GlobalAveragePool` to check these updated models.

BTW, sorry that I forgot to mention one more thing last time. TestCoverage.md needs to be updated as well. You could use `python onnx/backend/test/stat_coverage.py` to get the updated one. Hopefully all CIs will pass. Thank you. @calvinmccarter-at-lightmatter !

jcwchen:
Please also regenerate Operators.md and TestCoverage.md after removing those print. Thanks!

calvinmccarter-at-lightmatter:
@jcwchen - done!

jcwchen (approved):

PR #3473, opened 2021-05-06
ldconfig requires sudo

jackwish:
Signed-off-by: Zhenhua Wang (王振华) <zhenhua.wzh@icloud.com>


jackwish:
I found this when investigating the pipeline failure of https://github.com/onnx/onnx/pull/3467

jackwish:
As a follow up, do we really need to build protobuf on the fly? I can see that Mac CI is installing with `brew`. So why not `apt install` for Linux? Going to try if we can merge this.

jcwchen:
> As a follow up, do we really need to build protobuf on the fly? I can see that Mac CI is installing with `brew`. So why not `apt install` for Linux? Going to try if we can merge this.

Ideally, ONNX CIs use the same version of protobuf (3.11.3). For Mac-CI, I tracked brew history and get that version of protobuf. (Currently Mac-CI uses the latest protobuf, but it will be fixed in this PR: https://github.com/onnx/onnx/pull/3448) For Linux-CI, I cannot find a proper pre-installed apt package for 3.11.3 so it needs to build from source now... (Fortunately it does not take a long time). Still, please let me know if you figure out a way to apt install 3.11.3 protobuf directly. Thank you!

jcwchen (commented):
Thank you for catching this! Could you update here as well: https://github.com/onnx/onnx/blob/e1eb6eee9dead1996538e358454e0b10f060c9f5/.github/workflows/manylinux/entrypoint.sh#L30

jackwish:
Thanks for the quick check! Fixed the GitHub CI too. Please help to recheck.

PR #3484, opened 2021-05-11
Update Scatter style ops: Add reduction attribute

shubhambhokare1:
Update the following ops:
- ScatterElements
- ScattterND
by adding the attribute 'reduction' which will add support for duplicate values in indices instead of arbitrarily picking a single one of the indices for these operations


gramalingam:
Thanks for the PR. It may be helpful to join the operator SIG call this Thursday to discuss the PR. (See https://github.com/onnx/onnx/blob/299f83fe6900706c2c798def517566ce058846f1/onnx/defs/math/defs.cc#L671 for meeting info.)

gramalingam:
Hi @shubhambhokare1 : is this intended to be included in the upcoming release? Please be aware that the cutoff for this release is July 15th.

postrational:
Please explain the behaviour of the `accumulate` attribute for each of the operators in the specification.

onnx/defs/tensor/defs.cc
```diff
@@ -1363,9 +1381,17 @@ Example 2:
...
 ONNX_OPERATOR_SET_SCHEMA(
     ScatterND,
-    13,
+    16,
     OpSchema()
-        .SetDoc(ScatterND_ver13_doc)
+        .SetDoc(ScatterND_ver16_doc)
+        .Attr(
+            "reduction",
+            "Type of reduction to apply: none (default), add, multiply. "
```
gramalingam, line 1389: Nit: replacing "multiply" by "mul" will make it a bit more consistent with the names used for the corresponding ONNX ops.

onnx/defs/tensor/defs.cc
```diff
@@ -1427,12 +1453,25 @@ index-value for dimension = axis is obtained from the value of the corresponding
...
   output[i][indices[i][j]] = updates[i][j] if axis = 1,
 ```
+When `reduction` is set to "add", the update corresponding to the [i][j] entry is performed as below:
+```
+  output[indices[i][j]][j] += updates[i][j] if axis = 0,
+  output[i][indices[i][j]] += updates[i][j] if axis = 1,
+```
+When `reduction` is set to "multiply", the update corresponding to the [i][j] entry is performed as below:
+```
+  output[indices[i][j]][j] += updates[i][j] if axis = 0,
```
gramalingam, line 1472: Change "+=" tp "*="

gramalingam:
Hi @shubhambhokare1 : there are CI failures (equality comparison in python, etc.). Also, you need to signoff your commits as mentioned in the DCO above. Thanks!

onnx/version_converter/adapters/scatterelements_15_16.h
```diff
@@ -0,0 +1,27 @@
...
+
+// Adapter for ScatterElements in default domain from version 15 to 16
+
+#pragma once
+
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+struct ScatterElements_15_16 final: public Adapter {
```
gramalingam, line 13: This looks fine. But I wonder if we need it. Isn't this a compatible extension? Wouldn't a "CompatibleAdapter" work? If no value is specified for the reduction attribute, the default value "none" produces the old behavior, right?

shubhambhokare1: Added compatible adapters instead

gramalingam (approved):

PR #3485, opened 2021-05-11
Fix GitHub Action release CI failure: remove sudo ldconfig

jcwchen:
**Description**
GitHub Action does not have pre-installed sudo: https://github.com/nektos/act/issues/269. BTW, actually installing protobuf does not need sudo and ldconfig. Remove sudo in GitHub Action pipelines

**Motivation**
GitHub Action release Linux CI fails.


askhade:
Is this PR reverting the one just checked in https://github.com/onnx/onnx/pull/3473 ?

jackwish:
> installing protobuf does not need sudo and ldconfig.
For the `ldconfig` part, it depends on how the protobuf is used. True for just building, the compiler can find where the libraries are installed where they are the well-known paths (e.g. `/usr/lib`). But we do need `ldconfig` for running the test because the linker uses a cache which is updated by `ldconfig`. We don't encounter the failure (something like libprotobuf not found) here because we [install more packages via `apt`](https://github.com/onnx/onnx/pull/3485/files#diff-b8ce27cfa015392b153122c9a224c2651f31ab00451a6a50571ab7e01cace6e6R62), which should involve the `ldconfig` when configuring the package. I.e. `ldconfig` was performed implicitly.

jcwchen:
> For the `ldconfig` part, it depends on how the protobuf is used.
> 
> * True for just building, the compiler can find where the libraries are installed where they are the well-known paths (e.g. `/usr/lib`).
> * But we do need `ldconfig` for running the test because the linker uses a cache which is updated by `ldconfig`. We don't encounter the failure (something like libprotobuf not found) here because we [install more packages via `apt`](https://github.com/onnx/onnx/pull/3485/files#diff-b8ce27cfa015392b153122c9a224c2651f31ab00451a6a50571ab7e01cace6e6R62), which should involve the `ldconfig` when configuring the package. I.e. `ldconfig` was performed implicitly.

Agreed. Even though it works now, to prevent future issues, I just added ldconfig back. (which is same as protobuf's official guide mentioned). Thank you.

.github/workflows/manylinux/entrypoint.sh
```diff
@@ -26,8 +26,8 @@ mkdir build_source && cd build_source
 
 cmake ../cmake -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_POSITION_INDEPENDENT_CODE=ON -Dprotobuf_BUILD_TESTS=OFF -DCMAKE_BUILD_TYPE=Release
 make -j$(nproc)
-sudo make install
-sudo ldconfig
+make install
+ldconfig
```
snnn, line 30: Probably you should completely remove the "ldconfig" line. 

You have set BUILD_SHARED_LIBS=OFF. If there is no shared libraries installed, then you don't need to run ldconfig.

snnn (approved):

askhade (approved):

PR #3486, opened 2021-05-12
Fix function body for NLL Loss and SCE Loss

ashbhandare:
The function body of NLL Loss assumes a float input type and errors out for MLFloat16 currently.
The SCE Loss function body computes the log_prob with an unstable result(possibly due to the Log op), which this PR fixes by using LogSoftmax instead resulting in a more stable result.
This PR also fixes a bug in test generation.


onnx/defs/math/defs.cc
```diff
@@ -3412,19 +3438,47 @@ bool BuildContextDependentFunctionBodySCE(
...
 
+  // Using stable implementation of LogSoftmax
+  // We need to unsqueeze the inputs on the last axis if they are 2D in order to make the 
+  // further Reshape and transpose valid.
+  body.push_back(
+      {{"X_unsqueezed"},
+        "Unsqueeze",
+       {"scores", "axes"}});
+  body.push_back(
+      {{"3D_shape"},
```
gramalingam, line 3451: Suggest renaming "3D_shape" to a valid C identifier, eg., "ThreeD_shape" or "Shape3D".

onnx/defs/math/defs.cc
```diff
@@ -3412,19 +3438,47 @@ bool BuildContextDependentFunctionBodySCE(
       {{"axes"},
        "Constant",
        {},
-       {MakeAttribute("value", ToDimensionOneInt64Tensor(1))}});
+       {MakeAttribute("value", ToDimensionOneInt64Tensor(-1))}});
 
+  // Using stable implementation of LogSoftmax
+  // We need to unsqueeze the inputs on the last axis if they are 2D in order to make the 
+  // further Reshape and transpose valid.
```
gramalingam, line 3445: Just curious: is this really required? It seems like reshaping a 2D with target-shape {0,0,-1} should have the same effect, if I look at the reshape spec. Wonder if there is some bug/ambiguity in Reshape impl/spec.

onnx/defs/math/defs.cc
```diff
@@ -2847,6 +2853,13 @@ bool BuildContextDependentFunctionBody(
     const FunctionBodyBuildContext& ctx,
     const OpSchema& schema,
     FunctionProto& functionProto) {
+  bool float_input = true;
+  if (ctx.getInputType(0) != nullptr) {
+    float_input = ctx.getInputType(0)->tensor_type().elem_type() == TensorProto_DataType_FLOAT;
```
gramalingam, line 2858: minor nit: ```ctx.getInputType(0)->tensor_type().elem_type()``` is repeated thrice. May be worth making it a variable. Changing the control-flow to ```if (ctx.getInputType(0) == nullptr) return;``` would make it doable.

onnx/defs/math/defs.cc
ashbhandare: you are right, this unsqueeze can be skipped

gramalingam (approved):

askhade (approved):

postrational (approved):

PR #3498, opened 2021-05-19
Add missing sparse tensor helpers

yuslepukhin:
**Description**
Add tests for make_atttr for sparse tensor and repeated.
Add make_sparse_tensor_value_info along with the tests.

**Motivation and Context**
Make sure sparse tensors are fully covered.


onnx/helper.py
```diff
@@ -482,6 +486,59 @@ def make_tensor_value_info(
...
+            if d is None:
+                pass
+            elif isinstance(d, integer_types):
+                dim.dim_value = d
+            elif isinstance(d, text_type):
+                dim.dim_param = d
+            else:
+                raise ValueError(
+                    'Invalid item in shape: {}. '
+                    'Needs to of integer_types or text_type.'.format(d))
```
askhade, line 534: typo

onnx/helper.py
```diff
@@ -482,6 +486,59 @@ def make_tensor_value_info(
...
 
 
+def make_sparse_tensor_value_info(
+        name,  # type: Text
+        elem_type,  # type: int
+        shape,  # type: Optional[Sequence[Union[Text, int]]]
+        doc_string="",  # type: Text
+        shape_denotation=None,  # type: Optional[List[Text]]
+):  # type: (...) -> ValueInfoProto
+    """Makes a ValueInfoProto based on the data type and shape."""
```
askhade, line 496: nit: of type SparseTensor

onnx/helper.py
askhade: can you also fix similar typo on line 481

onnx/helper.py
```diff
@@ -482,6 +486,59 @@ def make_tensor_value_info(
...
+    value_info_proto = ValueInfoProto()
+    value_info_proto.name = name
+    if doc_string:
+        value_info_proto.doc_string = doc_string
+
+    sparse_tensor_type_proto = value_info_proto.type.sparse_tensor_type
+    sparse_tensor_type_proto.elem_type = elem_type
+
+    sparse_tensor_shape_proto = sparse_tensor_type_proto.shape
```
gramalingam, line 505: Should this be inside the if-then branch? If shape is None, I guess it means an unknown shape, right? If we create a shape here, wouldn't that become a scalar of rank zero?

gramalingam, line 505: I think we should move this into the if-then branch? Otherwise, we could create a rank-zero (scalar) instead of an unknown-rank tensor for the shape=None case?

askhade: i dont think so... if shape is None as explained in the comment below this field will be omitted from the resulting protobuf... 

we take a similar approach for tensor proto as well

gramalingam: yes, the fact that is is used for tensor suggests it works. I am still unclear what exactly happens. It is possible that a "shape" object is created (when the above statement is executed), and that when it is saved to protobuf it is omitted because it has no fields? There is a difference between the "shape" object and its field "dims" ... the comment below describes whether a dims field is created or not. But my question is about whether the "shape" field is created or not, since in C++ we do check for the existence of the "shape" field to determine if it has a rank or not.

gramalingam: Nevertheless, wouldn't creating the shape field inside the if-then more accurately reflect what we want?

onnx/helper.py
yuslepukhin: Is `be` missing? `Needs to be`?

onnx/helper.py
yuslepukhin: This comment is not clear to me. `Data type` refers to the `elem_type` and the return type is `ValueProtoInfo`.

onnx/helper.py
gramalingam: Ok, I tried it out. It seems to work. It is a bit mysterious how exactly the protobuf implementation works.

yuslepukhin: Yes, it creates a shape object, that is empty. It is the same as calling mutable_shape() in C++

onnx/helper.py
askhade: I meant something like this: Makes a ValueInfoProto of type Sparse Tensor based on the elem_type and shape

askhade: this is just to improve readability... feel free to ignore if it is confusing... but there are 2 types in play here... 1. type of typeproto and the underlying data type of the tensor proto

yuslepukhin: And the actual return type which is ValueInfoProto

onnx/helper.py
gramalingam: Ok, this seems to work. This line does not actually create a shape field. A bit mysterious, but seems ok (just python-approach to protobuf).

gramalingam (approved):

Issue #3512, opened 2021-06-01
Shape inference crash for NegativeLogLikelihoodLoss


matteosal:
This script crashes at the shape inference test:
```
import onnx
from onnx import helper, TensorProto, shape_inference
import numpy as np

in1 = helper.make_tensor_value_info('in1', TensorProto.FLOAT, [3, 4, 5])
in2 = helper.make_tensor_value_info('in2', TensorProto.INT64, [3, 5])
out = helper.make_tensor_sequence_value_info('out', TensorProto.FLOAT, [])

node = helper.make_node('NegativeLogLikelihoodLoss', ['in1', 'in2'], ['out'])

graph = helper.make_graph([node], 'MyGraph', [in1, in2], [out], [])

model = helper.make_model(
    graph,
    producer_name='test', 
    ir_version=7,
    opset_imports=[helper.make_opsetid('', 12)]
)

onnx.checker.check_model(model)
print('Check done')
shape_inference.infer_shapes(model)
print('Shape inference done')
```

askhade:
the crash is indeed due to a bug in the shape inference logic... Sending a PR with fix in a few
1 more thing : "out = helper._make_tensor_sequence_value_info_('out', TensorProto.FLOAT, [])"
why is the output type defined as sequence[tensor[float]] ? This is not a valid type constraint for this op
https://github.com/onnx/onnx/blob/master/docs/Operators.md#type-constraints-85

PR #3513, opened 2021-06-03
crash fix

askhade:

**Description**
Fixing a crash in shape inference for node NegativeLogLikelihoodLoss 
Caller needs to check the result from InferenceContext::getAttribute for nullptr before dereferencing. Otherwise it results in crash when optional attribute "reduction" is not present.

**Motivation and Context**
This is a bug fix
https://github.com/onnx/onnx/issues/3512

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -2859,7 +2859,9 @@ bool BuildContextDependentFunctionBody(
   }
   auto input_type = ctx.getInputType(0)->tensor_type().elem_type();
   bool float_input = input_type == TensorProto_DataType_FLOAT;
-  std::string reduction_attr = ctx.getAttribute("reduction") == nullptr ? "mean" : ctx.getAttribute("reduction")->s();
+  auto reduction_attr_proto = ctx.getAttribute("reduction");
+  std::string reduction_attr =
+      reduction_attr_proto != nullptr && reduction_attr_proto->has_s() ? reduction_attr_proto->s() : "mean";
   std::vector<FunctionBodyHelper::NodeDef> body;
   body.push_back(
       {{"const_zero"},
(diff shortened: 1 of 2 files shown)


onnx/defs/math/old.cc
```diff
@@ -1484,6 +1485,10 @@ bool BuildContextDependentFunctionBody_opset12(
   return true;
 }
 
+void nlllshape(InferenceContext& ctx) {
+  
+}
+
```
jcwchen, line 1491: Don't forget to remove this

onnx/defs/math/defs.cc
```diff
@@ -2859,6 +2859,7 @@ bool BuildContextDependentFunctionBody(
   }
   auto input_type = ctx.getInputType(0)->tensor_type().elem_type();
   bool float_input = input_type == TensorProto_DataType_FLOAT;
+  std::string reduction_attr = ctx.getAttribute("reduction") == nullptr ? "mean" : ctx.getAttribute("reduction")->s();
```
jcwchen, line 2862: I see the reduction attribute is using mean by default: https://github.com/onnx/onnx/blob/04971f7d26aad7bbc71e2d92433631f985cb96c6/onnx/defs/math/defs.cc#L3143
I am curious that will it still be nullptr if reduction is not set?

gramalingam: minor suggestion: could simplify right-hand-side to ```getAttribute(ctx, "reduction", "mean")```. See: https://github.com/onnx/onnx/blob/04971f7d26aad7bbc71e2d92433631f985cb96c6/onnx/defs/shape_inference.h#L92

onnx/defs/math/old.cc
askhade: thanks!

onnx/defs/math/defs.cc
askhade: the ctx here refers to "FunctionBodyBuildContext" and not "InferenceContext"

I will use this in the type and shape inference function which uses InferenceContext

gramalingam: Sorry, missed that

gramalingam (approved):

onnx/defs/math/defs.cc
askhade: @jcwchen : This code sources the attribute from InferenceContext... if the attribute is not explicitly present it will return nullptr

Issue #3515, opened 2021-06-04
Squeeze shape inference failure with symbolic dimension


matteosal:
```
import onnx
from onnx import version_converter, helper, TensorProto, shape_inference
import numpy as np

nodes = [helper.make_node('Squeeze', ['in', 'axes'], ['out'])]
inputs = [
    helper.make_tensor_value_info('in', TensorProto.FLOAT, [1, 2, 'hello']),
    helper.make_tensor_value_info('axes', TensorProto.INT64,  [1])
]
outputs = [helper.make_tensor_value_info('out', TensorProto.FLOAT, (2, 'hello'))]
initializers = [helper.make_tensor('axes', TensorProto.INT64, [1], [0])]
model = helper.make_model(helper.make_graph(nodes, "g", inputs, outputs, initializers))

shape_inference.infer_shapes(model)
```
Shape inference fails with 
```
(op_type:Squeeze): [ShapeInferenceError] Inferred shape and existing shape differ in rank: (0) vs (2)
```

Seems like this stopped working with https://github.com/onnx/onnx/pull/3465

jcwchen:
Hi @matteosal,
Thank you for catching this. I will propose a general solution soon for symbolic shape inference and yes definitely the logic here needs to be improved... I will keep you in the loop if I have some progress for it.

matteosal:
Thanks. In the meantime, https://github.com/onnx/onnx/pull/3516 has a fix for this particular problem

gramalingam:
Thanks for catching this! Added a couple of comments to the PR.

PR #3516, opened 2021-06-04
Fix Squeeze shape inference

matteosal:
Fixes https://github.com/onnx/onnx/issues/3515

There seem to be 2 problems here:
1) Shape inference should not give up as soon as a symbolic dimension is found, but only if the axes spec wants to act on that dimension, i.e. if it's absent or if it contains that dimension explicitly
2) In case shape inference gives up, it should not predict a rank of 0 

I have fixed (1) but (2) still remains to be addressed.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -1642,8 +1642,9 @@ ONNX_OPERATOR_SET_SCHEMA(
               });
 
           for (int i = 0; i < input_ndim; ++i) {
-            if(!input_shape.dim(i).has_dim_value() && (axes_not_specified || std::find(axes.begin(), axes.end(), i) != axes.end())) {
-                // if dim has a symbolic value and the axes spec want to act on dim, return early because we can't infer the shape
+            if(!input_shape.dim(i).has_dim_value() && axes_not_specified) {
+                // if dim has a symbolic value and the axes spec want to act on all dims, 
+                // return early because we can't infer the shape
                 return;
             }
           }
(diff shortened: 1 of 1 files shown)


onnx/defs/tensor/defs.cc
```diff
@@ -1641,10 +1641,11 @@ ONNX_OPERATOR_SET_SCHEMA(
               [&](int64_t axis) -> int64_t {
                 return axis < 0 ? axis + input_ndim : axis;
               });
-          // if input consists a symbolic value, return early since shape cannot be inferred
+
           for (int i = 0; i < input_ndim; ++i) {
-            if (!input_shape.dim(i).has_dim_value()) {
-              return;
+            if(!input_shape.dim(i).has_dim_value() && (axes_not_specified || std::find(axes.begin(), axes.end(), i) != axes.end())) {
```
gramalingam, line 1646: In principle, I think we can omit the ```|| std::find(axes.begin(), axes.end(), i) != axes.end())``` part. Thus, we assume that the axes specification is correct for shape-inference purpose. If it is incorrect, the runtime execution needs to catch and report it anyway. Does that make sense? However, this would require minor changes to the loop below in line 1657 ... we should check if input_shape_has_dim_value before accessing it.

matteosal: And what should happen in the loop below if `has_dim_value` is false? I still see no option other than giving up the shape inference.
I understand your point about the checking that the symbolic size is actually 1 at runtime, that should happen anyway. But still shape inference should give up (i.e. restrain from specifying anything about the output shape) in this case.

matteosal:
> Can't we move line 1634 to after the first loop to fix (2)?

Yes, that worked

matteosal:
@gramalingam any news on this?

onnx/defs/tensor/defs.cc
gramalingam: This is a fine point of distinction. In the loop below, if the outer condition is true (that is, axes is specified and contains i), then we should omit dim(i) from the output. The idea is that the shape-inference contract is that it indicates what the output shape will be IF the op executes successfully and produces an output. Type and shape inference does NOT guarantee that the op will execute successfully (because of the need for some runtime checks). Does that make sense?

gramalingam: If you are not convinced, we can merge this as is, and we can do this as a separate PR later after discussion.

matteosal: > The idea is that the shape-inference contract is that it indicates what the output shape will be IF the op executes successfully and produces an output. Type and shape inference does NOT guarantee that the op will execute successfully (because of the need for some runtime checks). Does that make sense?

Ok so you are suggesting to omit dim(i) from the output if:
* dim(i) is numerical and equal to 1 (usual case) OR
* dim(i) is symbolic

In the second case, runtimes should obviously check that dim(i) is actually 1 when the Squeeze op is evaluated. Is that correct?

matteosal: I have added the change, please let me know if it's ok

gramalingam (approved):

Issue #3506, opened 2021-05-27
Symbolic shape inference support


jcwchen:
**Current ONNX shape inference limitations**
1. Will stop if it encounters symbolic dimensions. e.g., A (5*2) concat B (N*2) = C (?*2) 
2. Will stop if it bumps into some dynamic scenarios. e.g., Reshape to a dynamically computed shape

**Work items**

1. Extend shape inference with symbol generation and propagation
    - Extend graph-level shape inference
    - e.g., `5*2 + N*2` -> Create a new symbol M to represent its inferred shape
2. Partial data computation and propagation
    - Extend a new optional function `PartialDataPropagationFunction` just like existing `TypeAndShapeInferenceFunction`
    - Targeted operators: Add, Cast, Concat, Gather, Mul, Reshap, Shape, Slice, Size, Squeeze, UnSqueeze
3. (Independent) Propose a new ONNX Op to facilitate shape computation
    - There are examples that they need a few operators to extract certain dimension
    - To simplify these operators, add a new op Dimension for extracting a particular dimension value from the shape of input

I am doing 1. now and we plan to support it in ONNX 1.10.

cc @askhade @gramalingam

askhade:
Closing this as all the identified work has been done. As more items come up in this area lets create a new issue

PR #3518, opened 2021-06-08
Symbolic shape inference support-1: symbol generation

jcwchen:
**Description**
While bumping into a symbolic shape, generate a symbol instead of using shape=0.
~Introduce enable_symbolic option in shape_inference~

**Motivation and Context**
https://github.com/onnx/onnx/issues/3506 First step of symbolic shape inference: symbolic symbol generation


onnx/shape_inference/implementation.cc
```diff
@@ -140,6 +150,10 @@ void mergeShapesAndTypes(const TypeProto_Tensor& inferredType, TypeProto_Tensor*
     auto* existingDim = existingType->mutable_shape()->mutable_dim(i);
     if (!existingDim->has_dim_value()) {
       *existingDim = inferredDim;
+      // If shape cannot be inferred; set it as a symbolic shape
+      if (inferredType.shape().dim(i).dim_value() == 0 && !inferredType.shape().dim(i).has_dim_param()) {
+        existingDim->set_dim_param(SymbolicShape::createNew());
```
askhade, line 155: how do you createNew simply generated a random sumbol right... how do you guarantee this symbol is unique within the graph? You need to maintain a graph level symbol list and make sure every new symbols is indeed unique for this graph

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,39 @@
...
+    const std::string symbol_prefix = "unk_";
+    template<class T>
+    void checkExistingSymbolicShape(T protos) {
+      for (auto& proto : protos) {
+        auto tensorType = proto.mutable_type()->mutable_tensor_type();
+        if (tensorType->has_shape()) {
+          for (int j = 0; j < tensorType->shape().dim_size(); ++j) {
+            if (tensorType->shape().dim(j).has_dim_param()) {
+              unsigned int existingIndex = 0;
+              sscanf(tensorType->shape().dim(j).dim_param().c_str(), (symbol_prefix + "%d").c_str(), &existingIndex);
```
jcwchen, line 37: Although using sscanf here can easily detect "unk_XXX", sscanf might be a dangerous function...

onnx/shape_inference/implementation.cc
```diff
@@ -12,6 +12,8 @@ namespace ONNX_NAMESPACE {
 namespace shape_inference {
 namespace {
 
+SymbolicShape symbolicShape;
```
askhade, line 15: why make it global? This should be in InferShapesImpl

onnx/shape_inference/implementation.h
askhade: I dont exactly get whats happening here... Are you assuming the model will have symbols in a specific numerical range ? Why no simply maintain a list of all symbols in the graph and then when you create a new one just make sure it doesnt already exist

jcwchen: My intention was to save some memory (additional existing shape hashset) and time... For instance, if "unk_1", "unk_2"..."unk_50"  are used, using a list to check will go through "unk_1" to "unk_50" and then finally find "unk_51" is usable. This will take more time... but as you said, to assume "unk_XXX" here, the logic will look quite confusing. I am changing the logic to maintain a shape list and check. Thanks for the review!

askhade: well the time depends on the data structure you use.

onnx/test/symbolic_shape_test.py
```diff
@@ -0,0 +1,88 @@
...
+            nodes=[clip, cast],
+            inputs=[helper.make_tensor_value_info('A', TensorProto.FLOAT, [2, 'A']),
+                helper.make_tensor_value_info('B', TensorProto.FLOAT, [2, 3])],
+            outputs=[helper.make_tensor_value_info('output', TensorProto.FLOAT, [2, 'M'])]
+        )
+
+        onnx_model = make_model(graph_def)
+        inferred_model = onnx.shape_inference.infer_shapes(onnx_model, strict_mode=True)
+        vis = list(inferred_model.graph.value_info)
+        inferred_vis = [make_tensor_value_info("C", TensorProto.FLOAT, (2, 'unk__0'))]
```
gramalingam, line 36: Instead of checking that the dimension has a specific name, it is better to check that it has some name. (We can do this check for 'C', and/or 'output'.) Alternatively, we can check that 'C' and 'output' both have a name and that the name is the same.

onnx/test/symbolic_shape_test.py
```diff
@@ -0,0 +1,88 @@
...
+            inputs=[helper.make_tensor_value_info('A', TensorProto.FLOAT, [2, 'A']),
+                helper.make_tensor_value_info('B', TensorProto.FLOAT, [2, 3]),
+                helper.make_tensor_value_info('D', TensorProto.FLOAT, [2, 'D'])],
+            outputs=[helper.make_tensor_value_info('output', TensorProto.FLOAT, [2, 'M'])]
+        )
+
+        onnx_model = make_model(graph_def)
+        inferred_model = onnx.shape_inference.infer_shapes(onnx_model, strict_mode=True)
+        vis = list(inferred_model.graph.value_info)
+        inferred_vis = [make_tensor_value_info("C", TensorProto.FLOAT, (2, 'unk__0')),
```
gramalingam, line 59: Same comment as above.

onnx/test/symbolic_shape_test.py
```diff
@@ -0,0 +1,88 @@
...
+            inputs=[helper.make_tensor_value_info('A', TensorProto.FLOAT, [2, 'unk__0']),
+                helper.make_tensor_value_info('B', TensorProto.FLOAT, [2, 3])],
+            outputs=[helper.make_tensor_value_info('output', TensorProto.FLOAT, [2, 'unk__1'])]
+        )
+
+        onnx_model = make_model(graph_def)
+        inferred_model = onnx.shape_inference.infer_shapes(onnx_model, strict_mode=True)
+        vis = list(inferred_model.graph.value_info)
+        # unk__0 and unk__1 have been used so it will use unk__2
+        inferred_vis = [make_tensor_value_info("C", TensorProto.FLOAT, (2, 'unk__2'))]
```
gramalingam, line 82: Same comment as above. Can check that the new name is different from the old names.

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,39 @@
...
+      newSymbol = symbol_prefix + std::to_string(index_++);
+    } while(existing_shape_set.count(newSymbol) > 0);
+    return newSymbol;
+  };
+  private:
+    unsigned int index_;
+    std::unordered_set<std::string> existing_shape_set;
+    const std::string symbol_prefix = "unk__";
+    template<class T>
+    void checkExistingSymbolicShape(T protos) {
```
gramalingam, line 34: Nit: rename 'checkExistingSymbolicShape' to 'AddExistingSymbolicDims'.

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,39 @@
...
 namespace shape_inference {
 
+struct SymbolicShape {
+  SymbolicShape(GraphProto* g) : index_(0){
+    existing_shape_set.clear();
+    checkExistingSymbolicShape(*g->mutable_input());
+    checkExistingSymbolicShape(*g->mutable_output());
+    checkExistingSymbolicShape(*g->mutable_value_info());
+  }
+  std::string createNew() {
```
gramalingam, line 22: It would be useful to have the symbol_prefix as an optional parameter to this function (with a default value "unk__"). This allows callers to create more meaningful names if they want.

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,39 @@
 namespace ONNX_NAMESPACE {
 namespace shape_inference {
 
+struct SymbolicShape {
```
gramalingam, line 15: Nit: rename 'SymbolicShape' as 'SymbolTable' (a well-known term in compilers).

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,39 @@
...
+  std::string createNew() {
+    std::string newSymbol;
+    do {
+      newSymbol = symbol_prefix + std::to_string(index_++);
+    } while(existing_shape_set.count(newSymbol) > 0);
+    return newSymbol;
+  };
+  private:
+    unsigned int index_;
+    std::unordered_set<std::string> existing_shape_set;
```
gramalingam, line 31: Nit: rename 'existing_shape_set' to 'existing_symbols'.

askhade:
Is this PR ready to be reviewed?

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,39 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
     fail_type_inference("type case unsupported. existing=", existingTypeCase, " inferred=", inferredTypeCase);
   }
 }
+// TypeProto_Tensor or TypeProto_SparseTensor
+template <typename T>
+void generateSymbolicShape(T inferredType, T* symbolicInferredType, SymbolicShape symbolicShape) {
```
gramalingam, line 120: Unclear why we need the first two parameters, and what the relationship between them is. It looks like in the calling context they differ only in the "const" being cast? Can't we use a single non-const parameter?

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,39 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+  if (!inferredType.has_shape()) {
+    *symbolicInferredType = inferredType;
+    return;
+  }
+  for (int i = 0; i < inferredType.shape().dim_size(); ++i) {
+    const auto& inferredDim = inferredType.shape().dim(i);
+    auto* typeAfterSymbolicShapeDim = symbolicInferredType->mutable_shape()->mutable_dim(i);
+    *typeAfterSymbolicShapeDim = inferredDim;
+    // If shape cannot be inferred; set it as a symbolic shape
+    if (inferredType.shape().dim(i).dim_value() == 0 && !inferredType.shape().dim(i).has_dim_param()) {
```
gramalingam, line 130: Shouldn't first condition be "! has_dim_value()"?

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,39 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+      typeAfterSymbolicShapeDim->set_dim_param(symbolicShape.createNew());
+    }
+  }
+}
+
+const TypeProto* enableSymbolicShape(const TypeProto* inferredType, SymbolicShape symbolicShape) {
+  const auto inferred_val_case = inferredType->value_case();
+  TypeProto* typeProto = const_cast<TypeProto*>(inferredType);
```
gramalingam, line 138: This is odd. I would make the parameter non-const here (and make this the caller's issue).

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,39 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+    auto* typeAfterSymbolicShapeDim = symbolicInferredType->mutable_shape()->mutable_dim(i);
+    *typeAfterSymbolicShapeDim = inferredDim;
+    // If shape cannot be inferred; set it as a symbolic shape
+    if (inferredType.shape().dim(i).dim_value() == 0 && !inferredType.shape().dim(i).has_dim_param()) {
+      typeAfterSymbolicShapeDim->set_dim_param(symbolicShape.createNew());
+    }
+  }
+}
+
+const TypeProto* enableSymbolicShape(const TypeProto* inferredType, SymbolicShape symbolicShape) {
```
gramalingam, line 136: nit: may be nicer to rename 'enableSymbolicShape' to 'materializeSymbolicShape'

onnx/test/shape_inference_test.py
```diff
@@ -59,7 +60,9 @@ def _assert_inferred(self, graph, vis, **kwargs):  # type: (GraphProto, List[Val
         inferred_vis = list(inferred_model.graph.value_info)
         vis = list(sorted(vis, key=lambda x: x.name))
         inferred_vis = list(sorted(inferred_vis, key=lambda x: x.name))
-        if vis == inferred_vis:
+        # remove generated symbolic shape for numerical matching
+        removed_symbolic_inferred_vis = re.sub('\n.*dim_param: \"unk_.*', '', str(inferred_vis))
```
gramalingam, line 64: What is this for? I think an alternative way of checking that doesn't assume what the generated names look like might be better.

onnx/shape_inference/implementation.cc
jcwchen: You are right. Besides this one, shall I add another condition: `inferredType.shape().dim(i).has_dim_param() && inferredType.shape().dim(i).dim_value() == -1` ?

onnx/shape_inference/implementation.cc
jcwchen: Sorry that the function here looks confusing previously. Actually the second argument is a mutable tensor type for updating inferredType after symbolic shape_inference. The first argument is the original inferredType. So I think both of them are needed... To make it clearer, I have changed the variable name and updated some comments here.

onnx/shape_inference/implementation.cc
jcwchen: Makes sense. Updated. Thanks

onnx/shape_inference.py
```diff
@@ -24,7 +24,7 @@
 bug in shape inference), and the result is unspecified.
 
 Arguments:
-    input (Union[ModelProto, Text, bytes], Text, bool) -> ModelProto
+    input (Union[ModelProto, Text, bytes], Text, bool, bool) -> ModelProto
```
askhade, line 27: what is the Text for?

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,42 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+    auto* typeAfterSymbolicShapeDim = existingTensorType->mutable_shape()->mutable_dim(i);
+    *typeAfterSymbolicShapeDim = inferredDim;
+    // set a symbol if it doesn't have dim_value and dim_param
+    if (!inferredType.shape().dim(i).has_dim_value() && !inferredType.shape().dim(i).has_dim_param()) {
+      typeAfterSymbolicShapeDim->set_dim_param(symbolTable.createNew());
+    }
+  }
+}
+
+const TypeProto* materializeSymbolicShapeShape(TypeProto* inferredType, SymbolTable symbolTable) {
```
askhade, line 138: SymbolTable& symbolTable

askhade: typo: materializeSymbolicShape**Shape**

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,42 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
     fail_type_inference("type case unsupported. existing=", existingTypeCase, " inferred=", inferredTypeCase);
   }
 }
+// after symbolic shape_inference from inferredType, assign it to existingTensorType
+// TypeProto_Tensor or TypeProto_SparseTensor
+template <typename T>
+void generateSymbolicShape(T inferredType, T* existingTensorType, SymbolTable symbolTable) {
```
askhade, line 121: this is very confusing... I don't understand what are you trying to do here...

All you need to do is check whether inferredType consists an unknown symbol, if so then generate a new symbol and set dim param... why do you need inferredType and existingTensorType.... + these names are misleading because both are same based on line 142

askhade: SymbolTable& symbolTable

askhade: this method should be simplified

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,42 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+    // set a symbol if it doesn't have dim_value and dim_param
+    if (!inferredType.shape().dim(i).has_dim_value() && !inferredType.shape().dim(i).has_dim_param()) {
+      typeAfterSymbolicShapeDim->set_dim_param(symbolTable.createNew());
+    }
+  }
+}
+
+const TypeProto* materializeSymbolicShapeShape(TypeProto* inferredType, SymbolTable symbolTable) {
+  const auto inferred_val_case = inferredType->value_case();
+  TypeProto* symbolicInferredType = inferredType;
```
askhade, line 140: use auto*

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,42 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+  if (inferred_val_case == TypeProto::kTensorType) {
+    generateSymbolicShape(inferredType->tensor_type(), symbolicInferredType->mutable_tensor_type(), symbolTable);
```
askhade, line 142: how are inferredType and symbolicInferredType different?

onnx/shape_inference/implementation.cc
askhade: you want to update inferredType then why not just do inplace update? why are you returning const TypeProto? + 
symbolicInferredType == inferredType  (line 140) this is very confusing...

onnx/shape_inference/implementation.cc
```diff
@@ -115,6 +115,42 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+  } else if (inferred_val_case == TypeProto::kSparseTensorType) {
+    generateSymbolicShape(inferredType->sparse_tensor_type(), symbolicInferredType->mutable_sparse_tensor_type(), symbolTable);
+  } else if (inferred_val_case == TypeProto::kSequenceType) {
+    TypeProto* sequenceTypeProto = symbolicInferredType->mutable_sequence_type()->mutable_elem_type();
+    TypeProto* innerTypeProto = const_cast<TypeProto*>(&(inferredType->sequence_type().elem_type()));
+    *sequenceTypeProto = *materializeSymbolicShapeShape(innerTypeProto, symbolTable);
+  } else {
+    fail_type_inference("type case unsupported for symbolic shape inference. inferred=", inferred_val_case);
```
askhade, line 150: fail_shape_inferece?

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,37 @@
 namespace ONNX_NAMESPACE {
 namespace shape_inference {
 
+struct SymbolTable {
+  SymbolTable(GraphProto* g) : index_(0){
```
askhade, line 16: GraphProto& g?
if you use a pointer then you should check for nullptr... if nullptr is unacceptable then you should simply use a reference and make this a caller issue...

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,37 @@
...
+    do {
+      newSymbol = symbol_prefix + std::to_string(index_++);
+    } while(existing_symbols.count(newSymbol) > 0);
+    return newSymbol;
+  };
+  private:
+    unsigned int index_;
+    std::unordered_set<std::string> existing_symbols;
+    void AddExistingSymbolicDims(google::protobuf::RepeatedPtrField<ValueInfoProto> protos) {
+      for (auto& proto : protos) {
```
askhade, line 33: const auto& since you are not modifying the proto

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,37 @@
...
+        auto tensorType = proto.mutable_type()->mutable_tensor_type();
```
askhade, line 34: why mutable when you are not modifying the proto? you are simply collecting all the symbols right

askhade: dont you need to check for sparse tensor and sequences?

onnx/shape_inference/implementation.h
```diff
@@ -12,6 +12,37 @@
...
+    AddExistingSymbolicDims(*g->mutable_input());
+    AddExistingSymbolicDims(*g->mutable_output());
+    AddExistingSymbolicDims(*g->mutable_value_info());
+  }
+  std::string createNew(std::string symbol_prefix="unk__") {
+    std::string newSymbol;
+    do {
+      newSymbol = symbol_prefix + std::to_string(index_++);
+    } while(existing_symbols.count(newSymbol) > 0);
+    return newSymbol;
```
askhade, line 27: you should add this new symbol to the list as well... symbol Table should include all the symbols applicable for the current graph

onnx/shape_inference/implementation.h
gramalingam: +1

onnx/shape_inference/implementation.cc
```diff
@@ -446,9 +449,9 @@ static void InferShapesImpl(
             *iter->second = *inferredType;
           }
         }
-        // if enable symbolic, check existing symbolic shape and generate new ssymbolic shape
-        SymbolicShape symbolicShape(g);
-        inferredType = enableSymbolicShape(inferredType, symbolicShape);
+        // Check existing symbolic shape and generate new symbolic shape
+        SymbolTable symbolTable(g);
```
gramalingam, line 453: minor nit: recomputing the symbolTable every time seems less efficient; we need to create it just once.

onnx/test/symbolic_shape_test.py
jcwchen: Good idea. Instead of checking the exact dim_param name, right now it only checks whether a dim_param is produced. I use "-1" as dim_value to represent a symbolic dim. Also, it further checks whether the symbolic dim of 'C' is same as 'output'.

onnx/test/symbolic_shape_test.py
jcwchen: Now it will check the number of unique symbol. If it correctly creates a new one without using the same symbol, the number should increase.

onnx/test/shape_inference_test.py
jcwchen: Initially it removed "dim_param: unk_XXX" for matching the inferred valueinfo and expected valueinfo. Now it will check each dim instead of matching the whole string. If it encounters symbolic shape, it will only check whether a dim_param is produced.

onnx/shape_inference.py
jcwchen: just a nit: correct the comment here that there should be 2 bool (check_type and strict_mode).

jcwchen: Oh. I should also remove Text here. Thanks for catching this!

onnx/shape_inference/implementation.cc
jcwchen: Updated: Only using mutable_tensor_type() can achieve it. Thank you for the comment!

onnx/shape_inference/implementation.cc
jcwchen: Thanks for catching this @askhade and @gramalingam ! I have simplified the arguments for `generateSymbolicShape` and `materializeSymbolicShape` by only using mutable_XXX_type()

onnx/shape_inference/implementation.cc
```diff
@@ -116,6 +116,35 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
 }
 
+// TypeProto_Tensor or TypeProto_SparseTensor
+template <typename T>
+void generateSymbolicShape(T* inferredType, SymbolTable& symbolTable) {
+  if (!inferredType->has_shape()) {
+    return;
+  }
+  for (int i = 0; i < inferredType->shape().dim_size(); ++i) {
+    const auto& inferredDim = inferredType->shape().dim(i);
```
askhade, line 126: inferredDim unused

onnx/shape_inference/implementation.cc
```diff
@@ -116,6 +116,35 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+    auto* typeAfterSymbolicShapeDim = inferredType->mutable_shape()->mutable_dim(i);
+    // set a symbol if it doesn't have dim_value and dim_param
+    if (!inferredType->shape().dim(i).has_dim_value() && !inferredType->shape().dim(i).has_dim_param()) {
```
askhade, line 129: use typeAfterSymbolicShapeDim instead of "inferredType->shape().dim(i)"

onnx/shape_inference/implementation.cc
```diff
@@ -116,6 +116,35 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
   }
 }
 
+// TypeProto_Tensor or TypeProto_SparseTensor
```
askhade, line 119: add sequence to the comment

jcwchen: Actually only TypeProto_Tensor or TypeProto_SparseTensor and will use generateSymbolicShape? Sequence will merely be expanded in materializeSymbolicShape.

onnx/shape_inference/implementation.cc
jcwchen: Good point. Now single InferShapes will only have single symbolTable (among InferShapesImpl, InferShapeForFunctionNode, doInferencing). To make subgraph inference share the same symbolTable from the main graph, I introduced symbolTable in graphInferenceContext. Thanks for the suggestion!

onnx/shape_inference/implementation.cc
```diff
@@ -192,14 +220,14 @@ static void InferShapesImpl(
     const std::unordered_map<std::string, int>& opset_imports,
     const bool check_type, // check the type-equality for input and output
     const int error_mode,
+    SymbolTable symbolTable,
```
askhade, line 223: SymbolTable& symbolTable

onnx/shape_inference/implementation.h
jcwchen: Good catch. I have added sparse_tensor and sequence for AddExistingSymbolicDims. Thank you!

onnx/shape_inference/implementation.cc
askhade: you are right

onnx/shape_inference/implementation.cc
```diff
@@ -504,7 +541,8 @@ void InferShapeForFunctionNode(
     const FunctionProto* func,
     const std::unordered_map<std::string, int>& func_opset_imports,
     const ISchemaRegistry* schema_registry,
-    InferenceContext& ctx) {
+    InferenceContext& ctx,
+    SymbolTable &symbolTable) {
```
askhade, line 545: nit: fix spacing SymbolTable& symbolTable

onnx/defs/shape_inference.h
```diff
@@ -13,6 +13,60 @@ namespace ONNX_NAMESPACE {
 
 using Dim = TensorShapeProto_Dimension;
 
+struct SymbolTable {
+  SymbolTable(GraphProto& g) : index_(0){
+    addFromGraph(g);
+  }
+  void addFromGraph(GraphProto& g) {
```
askhade, line 20: nit: add blank line before mtd start and after mtd end

askhade: nit: do you need addFromGraph? can you directly call AddExistingSymbolicDims from the constructor?

onnx/shape_inference/implementation.h
```diff
@@ -17,23 +17,32 @@ struct GraphInferenceContext {
...
+        symbolTable{symbolTable_in} {}
 
 
   const std::unordered_map<std::string, TypeProto*>*
       outer_scope_value_types_by_name;
   const std::unordered_map<std::string, int> opset_imports;
   const ISchemaRegistry* schema_registry;
+  SymbolTable& symbolTable;
 
+  SymbolTable* mutableSymbolTable() {
```
askhade, line 34: I dont understand the purpose of this... what is this for?

onnx/shape_inference/implementation.h
```diff
@@ -17,23 +17,32 @@ struct GraphInferenceContext {
...
+  }
 };
 
 class GraphInferencerImpl : public GraphInferencer {
  public:
   GraphInferencerImpl(GraphProto& g, const GraphInferenceContext& context)
-      : g_{&g}, context_{&context} {}
+      : g_{&g}, context_{&context} {
+        auto* contextForSymbol = const_cast<GraphInferenceContext*>(context_);
+        symbolTable_ = contextForSymbol->mutableSymbolTable();
```
askhade, line 44: how you are making the symbolTable available to GraphInferenceImpl needs to be simplified

context_ is private var for this class... can you not simply add a public mtd to fetch symbol table from context? So you that can get rid of symbolTable_ from here

onnx/defs/shape_inference.h
jcwchen: addFromGraph has another purpose. If shape_inference bumps into a sub-graph, it will rather use addFromGraph to add symbols from sub graph.

onnx/shape_inference/implementation.h
jcwchen: Actually we can simply use symbolTable above... Just removed it. Thanks!

onnx/shape_inference/implementation.cc
```diff
@@ -620,25 +659,29 @@ void InferShapeForFunctionNode(
...
     opset_imports[opset_import.domain()] = static_cast<int>(opset_import.version());
   }
-  InferShapeForFunctionNode(func, opset_imports, schema_registry, ctx);
+  InferShapeForFunctionNode(func, opset_imports, schema_registry, ctx, symbolTable);
 }
 
 std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
     const std::vector<const TypeProto*>& inputTypes,
     const std::vector<const TensorProto*>& inputData) {
+  SymbolTable symbolTable = getSymbolTable();
```
askhade, line 674: use reference... don't copy

onnx/shape_inference/implementation.h
```diff
@@ -39,6 +41,11 @@ class GraphInferencerImpl : public GraphInferencer {
...
       const std::vector<const TensorProto*>& inputData) override;
 
+  SymbolTable& getSymbolTable() {
+    auto* contextForSymbol = const_cast<GraphInferenceContext*>(context_);
+    return contextForSymbol->symbolTable;
+  }
+
  private:
   GraphProto* g_;
   const GraphInferenceContext* context_;
```
askhade, line 51: I think we should remove the const qualifier... making context_ const suggests that we do not change anything but we do... we do update the symboltable ...

onnx/shape_inference/implementation.cc
```diff
@@ -660,7 +703,7 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
         continue;
       }
     }
-
+    materializeSymbolicShape(inferredInput, symbolTable);
```
askhade, line 706: I think the case of GraphInferencer we should first merge and then cal materializeSymbolicShape on graphInput instead of inferredInput...

onnx/shape_inference/implementation.h
jcwchen: Good point. I have made GraphInferenceContext non-const

onnx/shape_inference/implementation.cc
jcwchen: The original expected flow is TypeandshapeInferencefunction -> generated symbols for unknown types -> attempt Partial DataPropagation -> merge.

Are you saying that moving generated symbols and partial DataPropagation after merge since they can manipulate  non-const graphInput after merge?

I have a concern: do we need `checkTensorShapesAndTypes` for checking the shape after Partial DataPropagation? Currently shape_inference does checkTensorShapesAndTypes in merge function. If we move merge before generated symbols and partial DataPropagation, it won't check the final propagated shape with existing shape.

onnx/defs/shape_inference.h
gramalingam: Technically, we need to decide on the scope of the dimension-symbols like M and N. The existing spec is
```
Currently, dimension variables are not scoped. A dimension variable "N" represents the same value
across the entire graph in a model. For example, if the graph has two inputs X and Y each with
shape ["N"], then at runtime the values passed in for X and Y MUST be tensors of rank 1 with the
same dimension. Nested sub-graphs currently share the same scope for dimension variables as the
main-graph. This allows a model to relate the dimensions of tensors inside the subgraph to the
dimensions of tensors in the outer graph.
```
(see under https://github.com/gramalingam/onnx/blob/master/docs/IR.md#static-tensor-shapes ). If we stick to this, we should visit all subgraphs and add their symbols right when constructing the top model-level symbol table.

onnx/test/symbolic_shape_test.py
```diff
@@ -35,27 +35,34 @@ def _assert_valueInfo_shape(self, onnx_model, valueInfo):
                     else:
                         assert dim.dim_value == nameShape[v.name][dim_i], '%s' % (onnx_model)
 
-    def _count_unqiue_dim_param_number(self, onnx_model):
+    def _count_unqiue_dim_param_number(self, onnx_model):  # type: (ModelProto) -> int
```
gramalingam, line 38: nit: spelling "unqiue" => "unique"

onnx/test/symbolic_shape_test.py
```diff
@@ -35,27 +35,34 @@ def _assert_valueInfo_shape(self, onnx_model, valueInfo):
...
 
-    def test_clip_enable_symbolic(self):  # type: () -> None
+    def _get_shape_from_name(self, onnx_model, name):  # type: (ModelProto, Text) -> Optional[Sequence[Union[Text, int]]]
+        shape = []
+        inputs = list(onnx_model.graph.input)
+        outputs = list(onnx_model.graph.output)
+        valueinfos = list(onnx_model.graph.value_info)
+        for v in inputs + outputs + valueinfos:
+            if v.name == name:
+                for dim_i in range(len(v.type.tensor_type.shape.dim)):
```
gramalingam, line 56: may be better to return ```v.type.tensor_type.shape``` here and have a separate equality-check method for two shapes, instead of creating another alternative shape representation. Also, would be useful to document assumptions for use of this method (e.g., whether the type is expected to be a tensor_type, or should we be checking that also here).

onnx/test/symbolic_shape_test.py
```diff
@@ -16,9 +16,9 @@
 
 class TestSymbolicShape(unittest.TestCase):
 
-    def _assert_valueInfo_shape(self, onnx_model, valueInfo):
+    def _assert_valueinfo_shape(self, onnx_model, valueinfo):  # type: (ModelProto, List[ValueInfoProto]) -> None
```
gramalingam, line 19: Doesn't seem to handle the case where inference did not infer a type for a name specified in ```valueinfo``` correctly. I think it may be better to treat this method as a shorthand for invoking ```get_shape_from_name``` and equality check for every entry in ```valueinfo```.

jcwchen: Good catch. I have considered the case if the name does not exist and reuse get_shape_from_name here.

onnx/test/symbolic_shape_test.py
jcwchen: Finally I chose to return shape without other check method and directly compare whether the TensorShapeProto is the same. Or I checked sparse_tensor_proto as well. Thanks!

onnx/shape_inference/implementation.cc
jcwchen: After offline discussion, I agreed with you. Since the inferredInput here isn't inferred in this level, it has generated symbol previously and it should be OK for us to do merge first then materializeSymbolicShape. In that case, we also don't need to use const_cast to break the const... Updated. Thanks!

jcwchen:
Updates:
- Eliminate unnecessary const_cast
- Improve Python test utilities
- Traverse all graphs first to add existing symbols
- Make SymbolTable as an interface

onnx/defs/shape_inference.h
```diff
@@ -13,6 +13,13 @@ namespace ONNX_NAMESPACE {
 
 using Dim = TensorShapeProto_Dimension;
 
+class SymbolTable {
+ public:
+    virtual void addFromGraph(const GraphProto& g) = 0;
```
askhade, line 18: add comments for every mtd

onnx/shape_inference/implementation.cc
```diff
@@ -620,17 +659,21 @@ void InferShapeForFunctionNode(
...
-  InferShapeForFunctionNode(func, opset_imports, schema_registry, ctx);
+  InferShapeForFunctionNode(func, opset_imports, schema_registry, ctx, symbolTable);
 }
 
 std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
     const std::vector<const TypeProto*>& inputTypes,
     const std::vector<const TensorProto*>& inputData) {
+  SymbolTableImpl& symbolTable = getSymbolTable();
+  // add existing symbolic shape into symbol table from sub graph
+  symbolTable.addFromGraph(*g_);
```
askhade, line 676: you dont need this anymore right? since you are already traversing the whole graph including subgraphs in the begining

onnx/shape_inference/implementation.h
```diff
@@ -12,36 +12,101 @@
...
+ public:
+  SymbolTableImpl() : index_(0){}
+
+  void addFromGraph(const GraphProto& g) {
+    AddExistingSymbolicDims(g.input());
+    AddExistingSymbolicDims(g.output());
+    AddExistingSymbolicDims(g.value_info());
+  }
+
+  std::string createNew(const std::string& symbol_prefix="unk__") {
```
askhade, line 25: Add comment something like:
Creates a new unique symbol with the given prefix and adds it to the SymbolTable
Returns the newly created symbol

onnx/test/symbolic_shape_test.py
```diff
@@ -0,0 +1,135 @@
...
+                helper.make_tensor_value_info('D', TensorProto.FLOAT, [2, 'D'])],
+            outputs=[helper.make_tensor_value_info('output', TensorProto.FLOAT, [2, 'unk__1'])]
+        )
+
+        onnx_model = make_model(graph_def)
+        original_count = self._count_unique_dim_param_number(onnx_model)
+        inferred_model = onnx.shape_inference.infer_shapes(onnx_model, strict_mode=True)
+        inferred_count = self._count_unique_dim_param_number(inferred_model)
+        # new symbol 'unk__2' should be generated to prevent duplicate so the count will be count + 1
+        assert inferred_count == original_count + 1, '%s%s' % (inferred_model, onnx_model)
```
askhade, line 131: shouldn't this be original_count + 2?

Concatenation of A[2, 'unk__0'], B[2, 3] should result in output C of shape [2, 'unk__2'] 
Concatenation of C[2, 'unk__2'] , D[2, 'D'] should result in output of shape [2, 'unk__3']
what am I missing?

onnx/test/symbolic_shape_test.py
```diff
@@ -0,0 +1,135 @@
...
+        valueinfos = list(onnx_model.graph.value_info)
+        for v in inputs + outputs + valueinfos:
+            if v.name == name:
+                if v.type.HasField('tensor_type'):
+                    return v.type.tensor_type.shape
+                elif v.type.HasField('sparse_tensor_type'):
+                    return v.type.sparse_tensor_type.shape
+        return None
+
+    def test_concat_enable_symbolic(self):  # type: () -> None
```
askhade, line 69: can you add a test case for unknown shape... in all these tests there is a symbol present... can we add something where 

(!has_dim_value() && !has_dim_param()) == true

onnx/test/symbolic_shape_test.py
jcwchen: Good catch. I think my previous example here is not a proper one (looks confusing):
- The original symbol table is {'unk__0', 'unk__1', 'D'}
- After inference symbol table is {'unk__2', 'unk__0', 'unk__3', 'D'}

Although original output shape is 'unk__1', but the inferred output shape will be 'unk__3', which replaces original 'unk__1'. 

To resolve confusion, I will use 'unk__0' as output shape. Then the inferred count will be count + 2 as you said.

onnx/test/symbolic_shape_test.py
```diff
@@ -5,37 +5,42 @@
...
+                expected_shape = expected_vi.type.tensor_type.shape
+            elif expected_vi.type.HasField('sparse_tensor_type'):
+                expected_shape = expected_vi.type.sparse_tensor_type.shape
+            assert len(shape.dim) == len(expected_shape.dim), '{}'.format(onnx_model)
+            for dim_i in range(len(shape.dim)):
+                dim = shape.dim[dim_i]
+                expected_dim = expected_shape.dim[dim_i]
+                # -1 means it's a symbolic shape
+                if expected_dim.dim_value == -1:
+                    # symbolic shape must exist
```
gramalingam, line 35: nit: => "symbolic dimension must exist"

gramalingam (approved):

askhade (approved):

Issue #3523, opened 2021-06-11
Typo in MaxPool Operator documentation


GuanLuo:
### Question
There seems to be a typo in formula to [calculate output spatial shape](https://github.com/onnx/onnx/blob/master/docs/Operators.md#maxpool):
where `... dilations[i] + 1 ...` should be `...dilations[i] - 1...` (see the shape section in PyTorch's [documentation](https://pytorch.org/docs/stable/generated/torch.nn.MaxPool2d.html))

### Further information
- Relevant Area (*e.g. model usage, backend, best practices, converters, shape_inference, version_converter, training, test, operators*):  maxpool operator

- Is this issue related to a specific model?  
**Model name** (*e.g. mnist*):  
**Model opset** (*e.g. 7*):

### Notes
Any additional information, code snippets.

jcwchen:
Hi @GuanLuo,
I also think this is a typo... Just propose a PR to fix it. Thank you for catching this!

gramalingam:
Hmmm ... I don't think so. Isn't there a parentheses around, like ``` ... - ( ... dilations[i] + 1)``` ? If you remove the parentheses, the minus of +1 becomes minus 1.

jcwchen:
> Hmmm ... I don't think so. Isn't there a parentheses around, like ` ... - ( ... dilations[i] + 1)` ? If you remove the parentheses, the minus of +1 becomes minus 1.

I think you are right. Thank you @gramalingam.

gramalingam:
Ok, I will close this. Please reopen in case there is a problem.

PR #3524, opened 2021-06-11
Correct MaxPool documentation

jcwchen:
**Description**
The formula for output shape of MaxPool was wrong...  Should be as follows: `output_spatial_shape[i] = floor((input_spatial_shape[i] + pad_shape[i] - {kernelSpatialShape}) / strides_spatial_shape[i] - 1)`

Reference:
PyTorch: https://pytorch.org/docs/stable/generated/torch.nn.MaxPool2d.html
ONNXRuntime: https://github.com/microsoft/onnxruntime/blob/c99aa3a3f3834adcbb888ce4b964f2695c524eae/onnxruntime/core/providers/cpu/nn/pool_attributes.h#L178

**Motivation and Context**
https://github.com/onnx/onnx/issues/3523

### DIFF
--- a/onnx/defs/nn/defs.cc
+++ b/onnx/defs/nn/defs.cc
@@ -227,11 +227,11 @@ std::function<void(OpSchema&)> PoolOpSchemaGenerator(
  subset of the input tensor according to the kernel size and downsampling the
  data into the output tensor Y for further processing. The output spatial shape will be following:
  ```
- output_spatial_shape[i] = floor((input_spatial_shape[i] + pad_shape[i] - {kernelSpatialShape}) / strides_spatial_shape[i] + 1)
+ output_spatial_shape[i] = floor((input_spatial_shape[i] + pad_shape[i] - {kernelSpatialShape}) / strides_spatial_shape[i] - 1)
  ```
  or
  ```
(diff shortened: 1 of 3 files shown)


onnx/defs/nn/defs.cc
```diff
@@ -227,11 +227,11 @@ std::function<void(OpSchema&)> PoolOpSchemaGenerator(
  subset of the input tensor according to the kernel size and downsampling the
  data into the output tensor Y for further processing. The output spatial shape will be following:
  ```
- output_spatial_shape[i] = floor((input_spatial_shape[i] + pad_shape[i] - {kernelSpatialShape}) / strides_spatial_shape[i] + 1)
+ output_spatial_shape[i] = floor((input_spatial_shape[i] + pad_shape[i] - {kernelSpatialShape}) / strides_spatial_shape[i] - 1)
```
gramalingam, line 230: Doesn't sound right. Consider the simple case where there is no padding, stride is 1, and kernel-size is same as input-size. We expect the output size to be 1, right? So, the original "+1" sounds right, not "-1". Can you double-check? Thanks!

gramalingam: I responded to the original issue, please take a look.

jcwchen:
Close it now since it is not correct.

PR #3531, opened 2021-06-16
Add model local functions

askhade:
**Description**
Extending model proto to include a list of model local function protos. 
FYI - all .proto and .proto3 files except onnx.in.proto and onnx-operators.in.proto are auto generated


askhade:
Closing this PR in favor of : https://github.com/onnx/onnx/pull/3532 to resolve DCO failures

PR #3532, opened 2021-06-16
extend model proto to include model local functions

askhade:

**Description**
Extending model proto to include a list of model local function protos.
FYI - all .proto and .proto3 files except onnx.in.proto and onnx-operators.in.proto are auto generated


onnx/onnx.in.proto
```diff
@@ -389,6 +390,15 @@ message ModelProto {
   //
   // If this field is empty, the training behavior of the model is undefined.
   repeated TrainingInfoProto training_info = 20;
+
+  // A list of function protos local to the model.
+  //
+  // The opset imports in FunctionProto should not conflict with the ones in 
+  // model proto and should be mergeable. Example, if the function proto imports domain
```
mrry, line 397: Nit: Can two different `FunctionProto`s in this list have conflicting imports? (I'm guessing they can't, but the wording here is ambiguous.)

docs/IR.md
```diff
@@ -83,6 +83,7 @@ Each model has the following components:
 |graph|Graph|The parameterized graph that is evaluated to execute the model.|
 |metadata_props|map<string,string>|Named metadata values; keys should be distinct.|
 |training_info|TrainingInfoProto[]|An optional extension that contains information for training.|
+|functions|FunctionProto[]|An optional list functions local to the model.|
```
askhade, line 86: will fix the typo "list of"

onnx/onnx.in.proto
mrry: Also, is it useful to be able to control the opset imports on a per-function basis, given this restriction? Would it be more ergonomic for the functions to inherit the imports from the containing ModelProto?

onnx/onnx.in.proto
```diff
@@ -755,3 +765,53 @@ message OperatorSetIdProto {
...
+
+  // This field indicates whether the syntax, semantics, or presence
+  // of this function is in an experimental or stable stage. Once an
+  // function is published as STABLE, its syntax and semantics MUST NOT
+  // change in subsequent versions of the operator set.
+  // When a function is published as EXPERIMENTAL, the syntax and semantics
+  // of the function MAY change across operator set versions.
+  // Functions "become" stable by deprecating the experimental version and
+  // introducing a new stable function with the same name.
+  optional OperatorStatus status = 3;
```
mrry, line 792: Is this field meaningful for model-local functions?

onnx/onnx.in.proto
```diff
@@ -755,3 +765,53 @@ message OperatorSetIdProto {
...
+  repeated string attribute= 6;
+
+  // The nodes in the function.
+  repeated NodeProto node = 7;
+  // A human-readable documentation for this function. Markdown is allowed.
+  optional string doc_string = 8;
+
+  // The OperatorSets this function body (graph) relies on.
+  // A FunctionProto body (graph) may implicitly rely on the OperatorSet that
+  // this function belongs to. It can also explicitly rely on more OperatorSets
```
mrry, line 808: Does the set of functions in a ModelProto count as an "OperatorSet" for the purposes of this comment? (i.e. Can a model-local function refer to other model-local functions?)

onnx/onnx.in.proto
```diff
@@ -389,6 +390,15 @@ message ModelProto {
...
   // If this field is empty, the training behavior of the model is undefined.
   repeated TrainingInfoProto training_info = 20;
+
+  // A list of function protos local to the model.
+  //
+  // The opset imports in FunctionProto should not conflict with the ones in 
+  // model proto and should be mergeable. Example, if the function proto imports domain  
+  // 'A' opset 10 whereas model proto also imports domain 'A' but with a different opset
+  // then this is a merge conflict.
+  repeated FunctionProto functions = 25;
```
mrry, line 400: How does the model refer to these functions? If they were in an operator set, presumably there would be a domain, but I don't see a domain here.

Should this field be an OperatorSetProto (or some subset of that proto)?

onnx/onnx.in.proto
```diff
@@ -755,3 +765,53 @@ message OperatorSetIdProto {
...
+  // Functions "become" stable by deprecating the experimental version and
+  // introducing a new stable function with the same name.
+  optional OperatorStatus status = 3;
+
+  // The inputs and outputs of the function.
+  repeated string input = 4;
+  repeated string output = 5;
+
+  // The attributes of the function.
+  repeated string attribute= 6;
```
mrry, line 799: Nit: missing space before `=`.

onnx/onnx.in.proto
gramalingam: Yes, this originates from the use of FunctionProto defined outside of models (in defining standard ops as functions). We could potentially distinguish between the two usages and restrict some fields to be used only in FunctionProtos outside of models.

onnx/onnx.in.proto
gramalingam: Agree, it does not seem relevant to model-local functions. It is unclear if this is even relevant for standard-defined functions any longer.

onnx/onnx.in.proto
gramalingam: I think we should allow a model-local function to refer to other model-local functions.

onnx/onnx.in.proto
askhade: ONNX does not allow importing different versions of the same domain so there cannot be conflicting imports... 

opset imports were added to FunctionProto to allow functions to easily import opsets other than what the model imports... the restriction is only in the version of the domain being unique within the model...

onnx/onnx.in.proto
askhade: FunctionProto has a "name" field. The model compares this with the op_type field from NodeProto.

Rama and I were discussing whether we want domain qualified names, meaning should we add domain field to the functionproto as well... we can do that if we think just the name field wont be enough...

onnx/onnx.in.proto
askhade: I dont think so but we share the same schema for both model local functions and function ops, so I did not  remove this.

onnx/onnx.in.proto
gramalingam: Two possible options: 
(a) We add a domain field to FunctionProto. (We then need to decide whether we want to allow these in-model functions to override out-of-model functions defined in opschema registries.)
(b) Reserve a domain for in-model functions (like 'onnx.ai.local') if we want to avoid the above.

onnx/onnx.in.proto
askhade: Yes one model local function can refer to another model local function...

more and more I think about this I am leaning towards adding domain field to FunctionProto...

onnx/onnx-ml.proto
```diff
@@ -754,6 +775,61 @@ message OperatorSetIdProto {
...
+    STABLE = 1;
+}
+
+message FunctionProto {
+  // The name of the function, similar usage of op_type in OperatorProto.
+  // Combined with FunctionProto.domain, this forms the unique identity of
+  // the FunctionProto.
+  optional string name = 1;
+
+  // Deprecated since IR Verison 8
```
mrry, line 790: Typo: s/Verison/Version/ (in a few other places too)

onnx/onnx-ml.proto
```diff
@@ -754,6 +775,61 @@ message OperatorSetIdProto {
...
+  // All FunctionProtos implicitly import the operator set which the function operator is part of.
+  //
+  // All nodes in the function body (graph) will bind against the operator
+  // with the same-domain/same-op_type operator with the HIGHEST version
+  // in the referenced operator sets. This means at most one version can be relied
+  // for one domain.
+  // 
+  // The opserator sets imported by FunctionProto should be compatible with the ones
+  // imported by ModelProto. Example, if same operator set say 'A' is imported by FunctionProto
+  // and ModelProto then versions for the operator set may be different but,
```
mrry, line 822: I'm struggling to understand why allowing different versions is useful here. I could probably contrive two functions that imported different opset versions for a domain without using ops that are different across the versions, but this seems like it would be fragile.

Would it be better to be more restrictive at first (all versions must be equal for a given domain) and then relax it in a later version?

onnx/checker.cc
```diff
@@ -676,39 +676,85 @@ void check_graph(const GraphProto& graph, const CheckerContext& ctx, const Lexic
...
+int get_version_for_domain(const std::string& domain, const std::unordered_map<std::string, int>& opset_imports) {
+  auto it = opset_imports.find(domain);
+  if (it == opset_imports.end()) {
+    return -1;
+  }
+
+  return it->second;
+}
+
+// Utility function to check compatibility of schema for 2 opsett versions for a given node.
```
mrry, line 690: Typo: s/opsett/opset/

onnx/checker.cc
```diff
@@ -676,39 +676,85 @@ void check_graph(const GraphProto& graph, const CheckerContext& ctx, const Lexic
...
+
+  if (schema_for_function_import && schema_for_model_import &&
+      schema_for_function_import->since_version() == schema_for_model_import->since_version()) {
+    // The since versions of schema for both opset imports match. This means they are compatible.
+    return;
+  }
+
+  fail_check(
+      "Opset import for domain " + node.domain() + " in function op " + node.op_type() +
+      "is not compatible with the version imported my model. FunctionOp imports version " +
```
mrry, line 726: Typo(?): s/my/by/

onnx/checker.cc
```diff
@@ -676,39 +676,85 @@ void check_graph(const GraphProto& graph, const CheckerContext& ctx, const Lexic
...
+      ONNX_NAMESPACE::to_string(func_opset_version) + "whereas model imports version " +
+      ONNX_NAMESPACE::to_string(model_opset_version));
+}
+
 void check_function(const FunctionProto& function, const CheckerContext& ctx, const LexicalScopeContext& parent_lex) {
   enforce_non_empty_field(function, name);
-  enforce_has_field(function, since_version);
+  
+  if (ctx.get_ir_version() >= 0x00000008) {
+    enforce_has_field(function, domain);
```
mrry, line 735: Is there a sensible/backwards-compatible default value for `domain` if it is unspecified?

onnx/checker.cc
```diff
@@ -676,39 +676,85 @@ void check_graph(const GraphProto& graph, const CheckerContext& ctx, const Lexic
...
-            std::to_string(version),
-            " for domain ",
-            (domain.empty() ? "ai.onnx" : domain),
-            " where as the model imports opset version ",
-            std::to_string(it->second));
-      }
+    func_opset_imports[domain] = version;
+
+    auto it = model_opset_imports.find(domain);
+    if (it != model_opset_imports.end() && version > it->second) {
```
mrry, line 748: Is this allowing the model to import opsets at any version less than the model imports for the same domain? (This seems different from the comment on the protobuf, but I haven't internalized all the logic yet, so I might be missing something....)

mrry: I'm also not quite sure why this logic is here rather than in `check_opset_compatibility()`. Might be worth adding a comment to explain.

onnx/defs/function.cc
```diff
@@ -60,9 +60,20 @@ void FunctionExpandHelper(
 
   // For undefined attributes of the function node
   // add default values obtained from the function schema.
+  // get the domain verison for function schema
+  int domain_version = -1;
+  for(const auto& opset_import : func.opset_import()){
```
mrry, line 65: Nit: Run `clang-format` over the PR before submitting.

onnx/defs/function.cc
```diff
@@ -60,9 +60,20 @@ void FunctionExpandHelper(
 
   // For undefined attributes of the function node
   // add default values obtained from the function schema.
+  // get the domain verison for function schema
```
mrry, line 63: Typo: s/verison/version/

onnx/defs/function.cc
```diff
@@ -60,9 +60,20 @@ void FunctionExpandHelper(
...
+  int domain_version = -1;
+  for(const auto& opset_import : func.opset_import()){
+    if(opset_import.domain() == node.domain()) {
+      domain_version = static_cast<int>(opset_import.version());
+    }
+  }
+  if(domain_version == -1) {
+    ONNX_THROW("No opset import registered for domain '" + node.domain() + "' in function proto");
```
mrry, line 71: IIUC, the check here means that every function in a domain `"foo"` must have an entry for `"foo"` in its `opset_import` with a particular version, which becomes `domain_version` here. Is that right?

This logic seems a little tortuous... would it make more sense to have a `domain_version` attribute on the `FunctionProto`? Or - perhaps better - the version could be inferred from the `opset_import` on the containing model (or inferred from the op schema, if the function is not a model-local function)?

At the very least, we need to document this requirement, because it was not obvious to me from reading the protobuf spec.

[EDIT: I see code in schema.cc that appears to add this information if it's missing. Is it still required for model-local functions?]

onnx/defs/schema.cc
```diff
@@ -703,8 +702,20 @@ void OpSchema::BuildFunction(FunctionProto& function_body, const std::vector<Ope
     function_body.add_attribute(a.first);
   }
 
-  for (auto& relied_opset : relied_opsets) {
-    *(function_body.mutable_opset_import()->Add()) = relied_opset;
+  // If the funciton proto does not import the
```
mrry, line 705: Typo: s/funciton/function/

onnx/onnx-ml.proto
askhade: restricting the version to be same as the version imported by model will not work for FunctionOps. Consider this example: Function Foo belongs to domain custom_domain and imports op "Add" from onnx domain in its function body... At the time of authoring this function "Foo", onnx latest released opset version was 13 so Foo imports onnx domain opset 13. Now, with this restriction on opset version the function Foo can only be used in models which import opset 13.... later opset wont work even if schema for Add remains same... Allowing different yet compatible version will allow models of higher opset to use function Foo.

onnx/onnx.in.proto
askhade: removed it in latest iteration

onnx/onnx.in.proto
askhade: added domain field for FunctionProto

onnx/defs/function.cc
askhade: Right every function in a domain "Foo" must have an entry for "foo" in its opset_import with a particular version.

For function ops - If this info is missing then it is inferred ( from op schema ) and implicitly added during schema finalization.

For model local functions - this info is required to be present. We can infer the version for the function from model opset imports but I wanted to reduce the differences in construction or interpretation of FunctionProto for FunctionOps vs model local functions so I decided not to make an exception for model local functions in this case. 

Let me know your thoughts...

onnx/checker.cc
gramalingam: Agree. It feels like this might be redundant, given check_opset_compatibility. If not, having the logic in one place seems better.

onnx/onnx-ml.proto
mrry: Thanks for clarifying! I think there might be a more precise way to state this. Something like:

> The operator sets imported by a FunctionProto must be "compatible" with the ones imported by the ModelProto that invokes them. Two operator sets are compatible if (i) they have the same domain and version, or (ii) they have the same domain and different versions, and all operators from the earlier version are present, with the same schema, in the later version. Intuitively condition (ii) means that a graph using the earlier version of a domain can be transparently "upgraded" to the later version by changing the version number.

I might be mangling the distinction between operator set and domain-version pair here. I'm also not sure if compatibility between model/function imports should be symmetric - as in the text here - or asymmetric. I think the checker implementation is asymmetric, and that might make sense, since IIUC we'd only consider "upgrading" a function (by inlining it into a model), and not a model itself.

onnx/onnx-ml.proto
```diff
@@ -396,10 +397,20 @@ message ModelProto {
...
+  // if a model local function exists for op_type + domain combination
+  // then it will be preferred over the operator or function present in 
+  // standard defined or cutom operator sets.
+  // 
+  // The operator sets imported by FunctionProto should be compatible with the ones
+  // imported by ModelProto and other model local FunctionProtos. 
+  // Example, if same operator set say 'A' is imported by a FunctionProto and ModelProto 
+  // or by 2 FunctionProtos then versions for the operator set may be different but, 
+  // the operator schema returned for op_type, domain, version combination
+  // for both the versions should be same.
```
gramalingam, line 410: "should be same" => "should be same for every node in the function body".

onnx/onnx-ml.proto
askhade: can you elaborate on symmetric vs asymmetric compatibility

onnx/defs/function.cc
gramalingam: There seem to be some corner cases where it is worth documenting what the standard requires. (a) If we have a model-local function "bar" in a domain "foo", is the model/graph required to have an import for "foo" (even though the model does not contain any calls to anything other than "foo")? (b) If so, if the FunctionProto for "bar" also includes an import for "foo", is it allowed to be different from what the graph specifies? The opset-compatibility discussion does not clarify this special-case. It doesn't seem important to allow them to be different in this case. (c) If we don't allow to be different, forcing the model to include the same information in two places seems less desirable. So, inferring it from the model's imports may be reasonable.

onnx/onnx-ml.proto
mrry: Symmetric would be more permissive. e.g. It would be legal to have model-imports-ONNX-10 and function-imports-ONNX-9 **or** model-imports-ONNX-9 and function-imports-ONNX-10. 

Whereas an asymmetric policy would only allow the larger version number in one case... most likely the model. This approach might make sense for an implementation based on inlining, because you could implement this as "upgrading" the function when it is inlined.

IIUC the logic here is asymmetric:

https://github.com/onnx/onnx/pull/3532/files/8a4cdd1d04c002f8db20e2a3ec32a0258515d99c#diff-27b8667be1a3c2c7ce2a2d324700c2f7bf400916c6b0e3508c8b27f85399ac6aR748

I suppose you could "downgrade" the function to an earlier opset when it is inlined, if you wanted to support a symmetric policy. But this doesn't seem necessary for the example you gave in your comment.

onnx/defs/function.cc
gramalingam: In general, it seems that model-local functions should not need to use opset-imports and should be able to get away with the model's own opset-imports (for the ops called from the function body). So, doing the same for the function's own domain (look it up from the model's imports) seems reasonable. It is also important to clarify which of the three options we go with (i) required, (ii) allowed but not required, or (iii) not allowed. In functions outside models, we do need to allow opset imports.

onnx/onnx-ml.proto
askhade: We want asymmetric compatibility and not symmetric. This is in sync with what we do for operators as well... A model can import a version <= model opset import version but never the other way...

mrry: That makes sense. In that case, we'll definitely need to change the comment to make that clear: the current "the versions for the operator set may be different" implies that the function version > model version is allowed.

gramalingam: I am unclear why we need asymmetric compatibility. Symmetric ("downgrading" the function as Derek says) would work just as well, without imposing any extra work or challenges, right? That way, when I define a function, I don't need to go to the trouble of finding the earliest opset it will work with ... I define it for the current opset, and transparent downgrading makes it as general as possible.

askhade: When defining a function the author needs to import the domain and version which they are using for defining the function body. For example if the function body contains op 'A' from domain 'X' then the author should import {'X', 'then_latest_ version'} in the function body. This way any model which imports version equal to or greater than the one imported by the function proto should work. If the schema for op 'A' changes then the function op needs to be updated with the new schema for op 'A' and corresponding import.

gramalingam: The author has several choices in above setting: import {'X', 'latest version of X'} (assuming this contains op 'A') or import {'X', 'earliest version of X that contains the corresponding op 'A''}. Since the opset-compatibility test can figure all this out, why do we need the extra constraint 'model which imports version equal to or greater than the one imported by the function proto' ?

askhade: import {'X', 'latest version of X'} (assuming this contains op 'A') => if <latest version> -1 contains op A then <latets version> will also contain op A. I dont see any motivation for the author to use anything other than latest version... By using a version earlier than latest version you restrict the usage of the function. 
Opset compatibility test only validates whether opsets are compatible... the upgrades are automatic meaning there is no schema change because the schema has not changed.

askhade: after our discussion I updated the checker to reflect symmetric compatibility

onnx/checker.cc
```diff
@@ -717,13 +714,24 @@ void check_opset_compatibility(
...
       ctx.get_schema_registry()->GetSchema(node.op_type(), model_opset_version, node.domain());
 
-  if (schema_for_function_import && schema_for_model_import &&
+  // If func_opset_version > model_opset_version then it is possible that the opschema
+  // for this node is not available. This is acceptable as long as schema is present for the
+  // version imported by function opset imports.
+  // Example: If customdomain.Foo is converted (in place .i.e without version bump) from
+  // primitive op to function op containing nodes from onnx domain then an older model which
+  // contains Foo can run into this scenario. This is bcecause Foo will import a later
+  // version of onnx domain than this older model.
```
gramalingam, line 726: I am not sure about this. I think we can allow func_opset_version > model_opset_version as long as both schemas and exist and we can check the equality of since_version for both as below.

onnx/onnx-ml.proto
```diff
@@ -407,12 +407,11 @@ message ModelProto {
   // Example, if same operator set say 'A' is imported by a FunctionProto and ModelProto 
   // or by 2 FunctionProtos then versions for the operator set may be different but, 
   // the operator schema returned for op_type, domain, version combination
-  // for both the versions should be same.
+  // for both the versions should be same for every node in the function body".
```
gramalingam, line 410: Extraneous " at end.

gramalingam (approved):

PR #3539, opened 2021-06-22
Add shape inference function for dynamic quantize linear

### LINKED ISSUE #3505: DynamicQuantizeLinear function does not have shape inference function
# Bug Report

### Describe the bug
DynamicQuantizeLinear function op does not have shape inference function defined. In absence of shape inference, function body is used to get the shape inference for the function op and although it works as a fallback option it hurts perf.

### Expected behavior
Add shape inference function for DynamicQuantizeLinear 

### Notes

Any additional information

annajung:

**Description**
- Add shape inference function to dynamic quantize linear

**Motivation and Context**
explicitly add shape inference function

fixes #3505

### DIFF
--- a/onnx/defs/quantization/defs.cc
+++ b/onnx/defs/quantization/defs.cc
@@ -179,6 +179,20 @@ ONNX_OPERATOR_SET_SCHEMA(
            {{"Zeropoint"}, "Cast", {"Rounded_ZeroPoint_FP"}, {MakeAttribute("to", int64_t(2))}},
            {{"y_scale"}, "Identity", {"Scale"}},
            {{"y_zero_point"}, "Identity", {"Zeropoint"}},
-           {{"y"}, "QuantizeLinear", {"x", "Scale", "Zeropoint"}}})));
+           {{"y"}, "QuantizeLinear", {"x", "Scale", "Zeropoint"}}}))
+        .TypeAndShapeInferenceFunction(
+            [](ONNX_NAMESPACE::InferenceContext& ctx) {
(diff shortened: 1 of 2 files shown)


annajung:
@askhade Please review when you can. I also wasn't sure if I needed to update any doc or run any script to update anything. If needed, please let me know! thanks

yufenglee (approved):

askhade (approved):
LGTM

onnx/defs/quantization/defs.cc
```diff
@@ -179,6 +179,20 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              updateOutputElemType(ctx, 0, TensorProto::UINT8);
+              updateOutputElemType(ctx, 1, TensorProto::FLOAT);
+              updateOutputElemType(ctx, 2, TensorProto::UINT8);
+
+              if (!hasInputShape(ctx, 0))
+                return;
+
+              auto& input_shape = getInputShape(ctx, 0);
+              updateOutputShape(ctx, 0, input_shape);
+              ctx.getOutputType(1)->mutable_tensor_type()->mutable_shape();
```
gramalingam, line 194: I think that these two lines for scale and zero-point can be moved up before line 189 (since this is valid even if we don't know the input shape).

annajung: Sure that makes sense. Will update

annajung: done! PTAL

gramalingam (approved):

PR #3542, opened 2021-06-24
Add xfail marking support to backend test runner

take-cheeze:
**Description**
- Marks test with [`unittest.expectedFailure`](https://docs.python.org/3/library/unittest.html#unittest.expectedFailure) when a test is marked with `xfail` method

**Motivation and Context**
- Skipping test is not enough to check new operator support so I would like to have `pytest.mark.xfail(strict=True)` like behavior


askhade:
@take-cheeze : Can you elaborate on what do you mean by "Skipping test is not enough to check new operator support"
In what situations are you expecting to use this behavior. Thanks.

take-cheeze:
On my local env I'm adding these lines to backend tests to check I'm not adding support to new operator unexpectedly:
```
NODE_FAILLIST = [
    # Failed tests:
    'test_resize_downsample_scales_linear',
    'test_resize_downsample_scales_linear_align_corners',
    # Failures from 1.9
    '_batchwise_',
]
NODE_FAIL_PATTERN = re.compile('|'.join(NODE_FAILLIST))

test_cases = backend_test.enable_report().test_cases
for cat, tcs in test_cases.items():
    for n, tc in tcs.__dict__.items():
        if NODE_FAIL_PATTERN.search(n):
            mark = pytest.mark.xfail(strict=True)
            setattr(tcs, n, mark(tc))
```
Commenting out these line would do what I expected but this check should run on CI.
I just want more maintained failed list instead of skipping with `exclude`.

After this PR I found this doc: https://github.com/onnx/backend-scoreboard/blob/master/ADD-BACKEND.md
And maybe using this feature it but I don't know how it's generated so I'll investigate later.

askhade:
@take-cheeze : Can you fix the DCO issue

take-cheeze:
@askhade Rebased and signed: https://github.com/onnx/onnx/pull/3542/commits/80653ecfbaf469a855933b9772877aa5e8e9d636

askhade (approved):

PR #3545, opened 2021-06-25
add additional type constraints for scale and bias in BN

mindest:
**Description**
- Add additional type constraints (`T1`) for scale and bias of BatchNormalization

**Motivation and Context**
- Current type binding constrains input `X` and scale/bias of the same type `T`, while in PyTorch there are cases where `X` is of `float16` and scale/bias of `float`, for example. Such cases should be handled.


mindest:
@SherlockNoMad Please help review this PR, thanks!

mindest:
@gramalingam @askhade Could you please help review when you have the time?

gramalingam:
Two questions:
(a) For the remaining type-variables that are shared (e.g., scale and bias; similarly for T2), are we sure that constraining them to be the same is okay? Or, is there any chance we will need to allow them to be of different types in the future?
(b) Can you please add a line of documentation to clarify the semantics when different precision is used? Eg., is the implementation required to cast everything to higher-precision upfront and perform the operation, and then downcast results as required? Or, is it more complex than that?

mindest:
@gramalingam 

> (a) For the remaining type-variables that are shared (e.g., scale and bias; similarly for T2), are we sure that constraining them to be the same is okay? Or, is there any chance we will need to allow them to be of different types in the future?

In common practice the scale/bias (which is like weight/bias in `torch.nn.Linear()`) and mean/var are of the same type, respectively. Loosening more of the type binding should also work, but its necessity is not clear from our practice and PyTorch/CuDNN usage of BatchNorm.

> (b) Can you please add a line of documentation to clarify the semantics when different precision is used? Eg., is the implementation required to cast everything to higher-precision upfront and perform the operation, and then downcast results as required? Or, is it more complex than that?

In training mode, if `T2` is float16, we require that mean/var be cast to float type because calculation of mean/variance can cause overflow in float16 due to summation. I've added this piece in the documentation, could you please check if it is okay?

docs/Operators.md
```diff
@@ -1891,6 +1891,8 @@ expect(node, inputs=[x], outputs=[y], name='test_averagepool_3d_default')
   Y = (X - input_mean) / sqrt(input_var + epsilon) * scale + B
   ```
   
+  It should also be noted that in training mode, if `T2` is float16, the mean/var should be cast to float when implementing, because calculation of mean and variance may cause overflow in float16 due to summation.
```
askhade, line 1894: did you manually edit this?
This change should be added to the defs.cc file, this way when you generate docs it will automatically pick it up

askhade: CIs are failing for the same reason...

askhade:
this change will need a version bump as well

docs/Operators.md
gramalingam: Yes, please update the defs.cc and regenerate the documents. I would change the comment to something like
```
The computation of ReduceMean and ReduceVar uses float to avoid overflow for float16 inputs.
```
Does this mean that it is not required for bfloat16 (only for float16)? Just trying to understand, even if it is an implementation-dependent feature.

mindest: Regenerated the docs, sorry for the mistake.
@gramalingam for bfloat16 case, since bfloat16 has the same range as float32 (not as precise because its fraction bits are truncated), it does not cause overflow. There could be precision issue with bfloat16, but I think this is not unique to BatchNormalization.

gramalingam:
As Ashwini mentioned, the opset version needs to be increased. The old definition is copied to the nn/old.cc file. The opset number is increased to 15 for the new definition and defs/operator_sets.h must be updated. See https://github.com/onnx/onnx/pull/3412 for an example PR.

gramalingam (approved):

PR #3548, opened 2021-06-30
honor ONNX_WERROR for all platforms and all libs

### LINKED ISSUE #3530: Multiple libraries (such as onnx) set WX when building with MSVC even when ONNX_WERROR is set to OFF
I understand from https://github.com/onnx/onnx/issues/1936 that this may be "by design", however, I see two problems with this:

1) It is inconsistent across platform AND a cursory look at the options leads one to believe that by setting ONNX_WERROR to OFF, warnings won't be treated as errors. This results (at least in my case) in a tracking down of where the issue that came from and time lost. At the very least, ONNX_WERROR should be documented correctly to indicate that even when it is OFF, WX will be passed to the compiler for some libraries.

2) Warnings are treated as errors that the user doesn't really have control over. For example, onnx-mlir includes onnx and uses the same version of protobuf as onnx or newer. When building onnx-mlir, we get the following warnings treated as errors (among others):

```
third_party\onnx\onnx\onnx-data.pb.cc(104): error C2220: the following warning is treated as an error
third_party\onnx\onnx\onnx-data.pb.cc(104): warning C4125: decimal digit terminates octal escape sequence
```

Since this is building unmodified onnx (1.9.0), this should be green, but it is not. So as the user of onnx, onnx-mlir has to explicitly disable those warnings or have build failures with MSVC.

askhade:

**Description**
With this change when ONNX_WERROR is set to OFF none of the libs will treat warnings as error.
Removing a few explicitly disabled warnings which don't show up any more

**Motivation and Context**
Fixes #3530 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -485,22 +485,12 @@ if(BUILD_ONNX_PYTHON)
     target_link_libraries(onnx_cpp2py_export PRIVATE ${PYTHON_LIBRARIES})
     target_compile_options(onnx_cpp2py_export
                            PRIVATE /MP
-                                   /WX
-                                   /wd4800 # disable warning type' : forcing
-                                           # value to bool 'true' or 'false'
-                                           # (performance warning)
-                                   /wd4503 # identifier' : decorated name length
-                                           # exceeded, name was truncated
-                                   /wd4146 # unary minus operator applied to
-                                           # unsigned type, result still
-                                           # unsigned from include\google\protob
-                                           # uf\wire_format_lite.h
-                                 /wd4244 # 'argument': conversion from 'google::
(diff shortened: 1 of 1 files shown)


gramalingam (approved):

jcwchen (approved):

CMakeLists.txt
```diff
@@ -545,16 +535,6 @@ endif()
 if(MSVC)
   target_compile_options(onnx_proto
                          PRIVATE /MP
-                                 /WX
-                                 /wd4800 # disable warning type' : forcing value
```
sstamenova, line 549: Does the onnx code generate any of the warnings that are no longer disabled?

sstamenova:
LGTM. Thanks!

PR #3551, opened 2021-06-30
Symbolic shape inference support-2: data propagation-1

jcwchen:
**Description**
- Introduce DataPropagationContext, PartialDataPropagationFunction
- Implement selected operators for PartialDataPropagationFunction as below:

  - [x] Shape 
  - [x] Squeeze 
  - [x] Unsqueeze
  - [x] Cast
 
- Add tests 

TODO in another PR:

Add, Concat, Gather,  Mul, Reshape, Slice, Size

**Motivation and Context**
https://github.com/onnx/onnx/issues/3506 Second step of symbolic shape inference: partial data propagation


onnx/defs/data_propagators.h
```diff
@@ -0,0 +1,29 @@
...
+  // propogate input data
+  const auto input_data = ctx.getGeneratedShapeData(idx);
+  if (input_data != nullptr) {
+    TensorShapeProto tp;
+    tp.CopyFrom(*input_data);
+    ctx.addGeneratedShapeData(0, std::move(tp));
+  }
+}
+
+inline void ShapeDataPropagator(DataPropagationContext& ctx) {
```
askhade, line 15: what is the purpose of this function?

askhade: after reading the rest of the PR now I understand this is the data propagator for shape op... Please update the function name to make it more clear and add some comments

onnx/shape_inference/implementation.cc
```diff
@@ -426,7 +427,9 @@ static void InferShapesImpl(
         materializeSymbolicShape(inferredType, symbolTable);
         // Now we can merge pre-existing and inferred info
         mergeShapesAndTypes(*inferredType, existingType);
-
+        DataPropagationContextImpl dataPropagationCtx(
+            n, valueTypesByName, inputDataByName, generatedShapeDataByName, &graphInferenceContext);
+        schema->GetDataPropagationFunction()(dataPropagationCtx);
```
askhade, line 432: every op will not have a data propagator defined... add a check something like
schema->has_data_propagation_function() before creating dataPropagationCtx and calling schema->GetDataPropagationFunction()(dataPropagationCtx);

jcwchen: Good catch. Added in both Python and C++.

onnx/defs/schema.h
```diff
@@ -388,6 +388,11 @@ class OpSchema final {
                                       : dummyInferenceFunction;
   }
 
+  OpSchema& PartialDataPropagationFunction(DataPropagationFunction partialDataProgationFunction);
```
gramalingam, line 391: minor nit: perhaps we can just call it ```DataPropagationFunction``` ... shorter, and consistent with the getter function below.

onnx/defs/shape_inference.h
```diff
@@ -74,11 +74,27 @@ struct InferenceContext {
   virtual const SparseTensorProto* getInputSparseData(size_t index) const = 0;
 };
 
+struct DataPropagationContext {
```
gramalingam, line 77: What about ```getAttribute``` ? Logically, it could be relevant for some data-propagation.

onnx/defs/shape_inference.h
```diff
@@ -74,11 +74,27 @@ struct InferenceContext {
...
+  virtual size_t getNumInputs() const = 0;
+  virtual const TypeProto* getInputType(size_t index) const = 0;
+  virtual size_t getNumOutputs() const = 0;
+  virtual TypeProto* getOutputType(size_t index) = 0;
+  virtual std::string getOutputName(size_t index) const = 0;
+  virtual ~DataPropagationContext() {}
+  virtual void addGeneratedShapeData(size_t index, TensorShapeProto&& tp) = 0;
+  virtual const TensorShapeProto* getGeneratedShapeData(size_t index) const = 0;
+  virtual const TensorShapeProto* getGeneratedShapeDataFromName(std::string name) const = 0;
```
gramalingam, line 86: Why do we need something name-based? I think index-based should be sufficient, and is more efficient.

gramalingam: Positional access is what we use in the other Contexts.

jcwchen: Right now it is solely used by test (verifying the generated propagation data by  given output name). Current getPropagatedData can only get data from input index. Perhaps I should to add another get function to get from output index instead of string name.

gramalingam: Yes, I think so. Further, I would suggest names like ```getInput``` or ```getInputData```.  It is also worth documenting that what is propagated are actual (partial) values, encoded using a TensorShapeProto, as opposed to shapes of inputs and outputs. E.g., for first input, we get whatever we know about the first input value (not just its shape).

gramalingam: The "by-name" ability is not required for the node-level propagation-method, which doesn't even know the name of the actual-variable passed in as input or output. The tester is more like the caller, which can use the actual implementation class (which exposes more methods than the Context interface).

onnx/defs/shape_inference.h
```diff
@@ -75,15 +75,16 @@ struct InferenceContext {
...
   virtual size_t getNumInputs() const = 0;
   virtual const TypeProto* getInputType(size_t index) const = 0;
   virtual size_t getNumOutputs() const = 0;
   virtual TypeProto* getOutputType(size_t index) = 0;
   virtual std::string getOutputName(size_t index) const = 0;
   virtual ~DataPropagationContext() {}
   virtual void addGeneratedShapeData(size_t index, TensorShapeProto&& tp) = 0;
-  virtual const TensorShapeProto* getGeneratedShapeData(size_t index) const = 0;
-  virtual const TensorShapeProto* getGeneratedShapeDataFromName(std::string name) const = 0;
+  virtual const TensorShapeProto* getInputGeneratedShapeData(size_t index) const = 0;
```
gramalingam, line 86: nit: I still think it is better to replace ```getInputGeneratedShapeData``` by ```getInputData```. It is shorter, and more accurate. I am afraid the ```Shape``` part will likely confuse some developers. It is *not* related to the shape of the input ... it is actually information about the value of the input, (which may be used subsequently to infer the shape of some other tensor).

gramalingam: Similarly, better to change ```getOutputGeneratedShapeData``` to ```getOutputData``` and ```addGeneratedShapeData``` to ```setOutputData```.

onnx/shape_inference/implementation.cc
```diff
@@ -620,6 +621,11 @@ void InferShapeForFunctionNode(
       }
       materializeSymbolicShape(inferred_output_type, symbolTable);
       mergeShapesAndTypes(*inferred_output_type, existingType);
+      DataPropagationContextImpl temp_dataPropagationCtx(
```
gramalingam, line 624: Couldn't we move this constructor into the then-branch below?

onnx/defs/shape_inference.h
askhade: agree with all the 3 recommendations

onnx/defs/shape_inference.h
```diff
@@ -74,11 +74,28 @@ struct InferenceContext {
   virtual const SparseTensorProto* getInputSparseData(size_t index) const = 0;
 };
 
+struct DataPropagationContext {
+  virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
+  virtual size_t getNumInputs() const = 0;
+  virtual const TypeProto* getInputType(size_t index) const = 0;
+  virtual size_t getNumOutputs() const = 0;
+  virtual TypeProto* getOutputType(size_t index) = 0;
+  virtual std::string getOutputName(size_t index) const = 0;
```
askhade, line 83: when do we need output name? and what about input name? do we need that too?

gramalingam: Agree. I don't think it needs to be in this interface. (Not needed for the node-level propagation function, even though the *implementation* may use the names.)

onnx/defs/shape_inference.h
gramalingam: Another suggestion is to add a ```typedef TensorShapeProto PartialInputData``` or ```typedef TensorShapeProto SymbolicInputData``` and use those new type names in this interface. We are reusing TensorShapeProto to avoid defining a new data-structure to store either constant-values or unknown-values. E.g., we could imagine a more sophisticated representation in the future.

askhade: why is getOutputData required? When will this be used? You only setoutput data ... the next node calls getInputdata to fetch this data right?

jcwchen: Just simplified the function name. GetOutputData is just for testing and won't be used in the propagation.

Yes it's still doable for testing without getOutputData... I have removed that function and add one more node behind the tested node to get the input data to verify.

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -0,0 +1,218 @@
...
+  final_node.set_op_type("Cast");
+  final_node.set_domain(ONNX_DOMAIN);
+  final_node.add_input(tested_output_name);
+  final_node.add_output(output_name);
+  AttributeProto* to = final_node.add_attribute();
+  to->set_name("to");
+  to->set_type(AttributeProto::INT);
+  to->set_i(1);
+
+  // Constructs graph
```
gramalingam, line 90: Just a note for the future: using the parser will make writing such test-cases much simpler. E.g., see the examples here: https://github.com/onnx/onnx/blob/56065a0decdafee066a448d72dd8d4fd17bf131d/onnx/test/cpp/parser_test.cc#L213 ... no need to change this now since you have already written it.

onnx/defs/tensor/defs.cc
```diff
@@ -369,6 +376,9 @@ ONNX_OPERATOR_SET_SCHEMA(
           ctx.getOutputType(0)->mutable_tensor_type()->set_elem_type(
               TensorProto::INT64);
           ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+        })
+        .PartialDataPropagationFunction([](DataPropagationContext& ctx) {
+          PropagateShapeDataFromInputToOutput(ctx, 0);
```
gramalingam, line 381: This is not correct. If all dimensions of the input are known values, we should return the actual size, otherwise an unknown value.

onnx/test/cpp/data_propagation_test.cc
jcwchen: Thanks for the info. I will utilize it with future tests

onnx/defs/tensor/defs.cc
jcwchen: Thanks for catching this. Just removed size in this PR

onnx/defs/data_propagators.h
```diff
@@ -0,0 +1,32 @@
+#include "onnx/defs/shape_inference.h"
+
+namespace ONNX_NAMESPACE {
+
+// Please note that TensorShapeProto only supports integer shape
```
jcwchen, line 5: Not sure whether this comment is needed or it should be improved

onnx/defs/shape_inference.h
jcwchen: Originally InferenceContext has `getInputData` function to get "TensorProto". I think DataPropagationContext will need this function for future propagation functions as well... To sync the naming convention, I will make `getInputData` to get "TensorProto", `getInputShapeData"`to get the produced "TensorShapeProto", and `addOutputShapeData` to propagate the "TensorShapeProto". This PR is ready for review again @gramalingam @askhade. Thank you!

onnx/defs/shape_inference.h
```diff
@@ -74,11 +74,27 @@ struct InferenceContext {
...
+struct DataPropagationContext {
+  virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
+  virtual size_t getNumInputs() const = 0;
+  virtual const TypeProto* getInputType(size_t index) const = 0;
+  virtual size_t getNumOutputs() const = 0;
+  virtual TypeProto* getOutputType(size_t index) = 0;
+  virtual ~DataPropagationContext() {}
+  virtual const TensorProto* getInputData(size_t index) const = 0;
+  virtual void addOutputShapeData(size_t index, TensorShapeProto&& tp) = 0;
+  virtual const TensorShapeProto* getInputShapeData(size_t index) const = 0;
```
askhade, line 86: Instead of having 
virtual const TensorProto* getInputData(size_t index) const = 0;
 virtual const TensorShapeProto* getInputShapeData(size_t index) const = 0;

can you simply keep one like this:
virtual const TensorShapeProto* getInputData(size_t index) const = 0;

The implementation of DataPropagationContext can convert the TensorProto to TensorShapeProto... otherwise in every data propagation function you will need to call 1 and if it returns nullptr then call the other...

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,121 @@ struct InferenceContextImpl : public InferenceContext {
...
+      ONNX_THROW("Input " + ONNX_NAMESPACE::to_string(index) + " is out of bounds.");
+    }
+    auto result = generatedShapeData_.insert({outputIndexToNameMap_.at(index), std::move(tp)});
+    if (!result.second) {
+      fail_shape_inference("Data for input  " + ONNX_NAMESPACE::to_string(index) + " already exists.");
+    }
+  }
+
+  const TensorShapeProto* getInputShapeData(size_t index) const override {
+    if (index >= allInputData_.size()) {
```
askhade, line 347: use inputIndexToNameMap_ instead of allInputData_? The check is good as both these containers will be of same size but using inputIndexToNameMap_  is cleaner

onnx/shape_inference/implementation.cc
```diff
@@ -426,7 +427,11 @@ static void InferShapesImpl(
         materializeSymbolicShape(inferredType, symbolTable);
         // Now we can merge pre-existing and inferred info
         mergeShapesAndTypes(*inferredType, existingType);
-
+        if (schema->has_data_propagation_function()) {
```
askhade, line 430: whether to enable data propagation or not needs to be controlled by an option... we can talk about it offline

onnx/defs/data_propagators.h
gramalingam: Attaching a more complete comment below. But I think a better place for this would be where
DataPropagationContext is defined.
```
We use data propagation to perform partial evaluation of the model, to compute statically
known information about tensor values. It is intended to improve the precision of shape
inference. We reuse TensorShapeProto to represent the statically known values. One
limitation of this is that TensorShapeProto can represent only integer values.
As an example, data-propagation is intended to handle code-fragments like below:
   shape = Shape(X)
   batchsize = Slice(shape, [0], [1])
   newshape = Concat (batchsize, [1024, 1024])
   Z = Reshape(Y, newshape)
```

onnx/shape_inference/implementation.cc
jcwchen: I added it as "data_prop" (any other idea?). Since the function now is not complete, setting it as an option (default is False) makes more sense to me. Please let me know if you have other concern. Thanks

onnx/defs/data_propagators.h
jcwchen: Updated. @gramalingam Thank you for the detailed explanation!

onnx/defs/shape_inference.h
jcwchen: Merging it as single `virtual const TensorShapeProto* getInputData(size_t index) const = 0;` sounds good. The updated getInputData will first check whether `allInputTypes_` exists a valid Shape inside TypeProto. If not, it will further check TensorShapeProto from `generatedShapeData_`. Please review it again

onnx/shape_inference/implementation.h
jcwchen: I decided to use `allInputTypes_` now because it will access allInputTypes_[index] in the next lines anyway so using it here will look more simpler...

onnx/shape_inference/implementation.cc
```diff
@@ -458,18 +464,20 @@ void InferShapes(
     const std::unordered_map<std::string, int>& opset_imports,
     const bool check_type,
     const ISchemaRegistry* schema_registry,
-    const int error_mode) {
+    const int error_mode,
+    const int data_prop) {
```
askhade, line 468: it would be good to create an options struct
struct ShapeInferenceOptions {
bool check_types;
enum error_mode; or bool strict_error_mode;
bool enable_data_propagation;
}

onnx/shape_inference/implementation.cc
```diff
@@ -493,7 +502,8 @@ void InferShapes(
     const bool check_type,
     const std::string& save_path,
     const ISchemaRegistry* schema_registry,
-    const int error_mode) {
+    const int error_mode,
+    const int data_prop) {
```
askhade, line 506: making this bool makes more sense... something like bool enable_data_prop
reason for error_mode being int is tomorrow we can introduce more modes easily but data propagation is binary yes or no

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,117 @@ struct InferenceContextImpl : public InferenceContext {
...
+      ONNX_THROW("Output " + ONNX_NAMESPACE::to_string(index) + " is out of bounds.");
+    }
+    return &allOutputTypes_[index];
+  }
+
+  const TensorShapeProto* getInputData(size_t index) const override {
+    if (index >= allInputTypes_.size()) {
+      ONNX_THROW("Input " + ONNX_NAMESPACE::to_string(index) + " is out of bounds.");
+    }
+    if (allInputTypes_[index] != nullptr && allInputTypes_[index]->has_tensor_type()
```
askhade, line 333: this is incorrect.... 
here you are returning the shape of the input and not data... you need to convert TensorProto into TensorShapeProto
For shape computations TensorProto ->TensorShapeProto is possible...

onnx/shape_inference/implementation.cc
jcwchen: Good idea. Added

onnx/shape_inference/implementation.h
jcwchen: Thanks for catching this. I used ParseData to convert allInputData_ in getInputData. Please review it again

onnx/shape_inference/implementation.h
```diff
@@ -325,15 +325,34 @@ struct DataPropagationContextImpl : public DataPropagationContext {
...
     }
-    return allInputData_[index];
+    // Gets it from initializer if it exists
+    const auto* input_data = allInputData_[index];
+    if (input_data != nullptr) {
+      std::vector<int64_t> input_vals;
+      if (input_data->data_type() == TensorProto_DataType_INT64) {
+        input_vals = ParseData<int64_t>(input_data);
+      } else if (input_data->data_type() == TensorProto_DataType_INT32) {
+        std::vector<int32_t> input_vals = ParseData<int32_t>(input_data);
```
gramalingam, line 339: This is a different variable from the outer-scope variable. This won't work since it won't be used in the loop below in line 342. I think you may have to duplicate the loop lines 342-344 separately for the then branch and else branch.

jcwchen: Updated. Thanks for the catch

onnx/defs/shape_inference.h
```diff
@@ -13,6 +13,21 @@ namespace ONNX_NAMESPACE {
 
 using Dim = TensorShapeProto_Dimension;
 
+struct ShapeInferenceOptions {
+  // Checks the type-equality for input and output
+  bool check_type;
+  // 1: Stricter shape inference; will throw errors if any
+  // 0: Simply stop if any error
+  int strict_error_mode;
```
askhade, line 21: I suggest change this to error_mode and let the value determine whether it is strict or something else
per current behavior : when the mode is strict aka 1 shape inf will throw any node level shape inf errors whereas when it is 0 it does not.... other errors like merging existing shape with inferred etc are thrown regardless...

onnx/shape_inference/implementation.cc
```diff
@@ -218,14 +218,14 @@ static void InferShapesImpl(
     GraphProto* g,
     const std::unordered_map<std::string, TypeProto*>& outer_scope_value_types_by_name,
     const std::unordered_map<std::string, int>& opset_imports,
-    const bool check_type, // check the type-equality for input and output
-    const int error_mode,
+    const ShapeInferenceOptions options,
```
askhade, line 221: please use reference: const ShapeInferenceOptions& options

onnx/shape_inference/implementation.cc
```diff
@@ -481,19 +483,17 @@ void InferShapes(
...
       m.ir_version());
 }
 
 void InferShapes(
     const std::string& model_path,
-    const bool check_type,
     const std::string& save_path,
     const ISchemaRegistry* schema_registry,
-    const int error_mode) {
+    const ShapeInferenceOptions options) {
```
askhade, line 496: const ShapeInferenceOptions& options

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,131 @@ struct InferenceContextImpl : public InferenceContext {
...
+      ONNX_THROW("Input " + ONNX_NAMESPACE::to_string(index) + " is out of bounds.");
+    }
+    // Gets it from initializer if it exists
+    const auto* input_data = allInputData_[index];
+    if (input_data != nullptr) {
+      TensorShapeProto* tsp = nullptr;
+      if (input_data->data_type() == TensorProto_DataType_INT64) {
+        std::vector<int64_t> input_vals = ParseData<int64_t>(input_data);
+        for (unsigned int i = 0; i < input_vals.size(); ++i) {
+          tsp->mutable_dim()->Add()->set_dim_value(input_vals[i]);
```
askhade, line 339: tsp is set to nullptr and tsp->mutable_dim() is being called before mem allocation. There is a deeper problem here... even if you initialize who will free it? the ownership for this is not clear.

You can do 1 thing - create TensorShapeProto and add it to generatedShapeData ... this way generatedShapeData  will own it and if the same initializer is being used else where you can save some time in reconverting TensorProto to TensorShapeProto.

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,131 @@ struct InferenceContextImpl : public InferenceContext {
...
+    }
+    return &allOutputTypes_[index];
+  }
+  const TensorShapeProto* getInputData(size_t index) const override {
+    if (index >= allInputData_.size()) {
+      ONNX_THROW("Input " + ONNX_NAMESPACE::to_string(index) + " is out of bounds.");
+    }
+    // Gets it from initializer if it exists
+    const auto* input_data = allInputData_[index];
+    if (input_data != nullptr) {
```
askhade, line 334: also check for shape of input... allow only 1D or scalars

gramalingam: Is that strictly necessary (for correctness)? We treat the returned value as representing the underlying sequence of data-values in the tensor, without saying anything about its shape. (The shape info, if any, is captured in the type.) I understand that in common use-cases this may not arise.

gramalingam: Restricting this to 1D or scalars is okay (it should work for the common-case), but I still think it is not necessary. Allowing other cases enables a more general implementation without much of a cost. (But, given the deadline, I am okay with this restriction for now.)

onnx/defs/shape_inference.h
```diff
@@ -74,11 +90,37 @@ struct InferenceContext {
...
+// If the shape of X is statically known, then data-propagation should be able to determine
+// the value of newshape, as well as the shape of Z.
+struct DataPropagationContext {
+  virtual const AttributeProto* getAttribute(const std::string& name) const = 0;
+  virtual size_t getNumInputs() const = 0;
+  virtual const TypeProto* getInputType(size_t index) const = 0;
+  virtual size_t getNumOutputs() const = 0;
+  virtual TypeProto* getOutputType(size_t index) = 0;
+  virtual ~DataPropagationContext() {}
+  virtual const TensorShapeProto* getInputData(size_t index) = 0;
```
gramalingam, line 111: Why is the ```const``` missing in ```getOutputType``` and ```getInputData``` ?

onnx/shape_inference/implementation.h
askhade: I suggest going with the restriction ... once we come across the models which require a more generic implementation we can update this...

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,145 @@ struct InferenceContextImpl : public InferenceContext {
...
+    // Gets it from previous data propagation
+    auto iter = generatedShapeData_.find(inputIndexToNameMap_.at(index));
+    if (iter != generatedShapeData_.end()) {
+      return &iter->second;
+    }
+    // Otherwise, gets it from initializer if it exists
+    const auto* input_data = allInputData_[index];
+    if (input_data != nullptr) {
+      // Only scalar (0D tensor) or 1D tensor can be converted
+      if (input_data->dims_size() == 0 || input_data->dims_size() == 1) {
```
askhade, line 352: nit: you can combine the if on line 350 and 352

onnx/defs/shape_inference.h
jcwchen: Current getInputData will add initializer into generatedShapeData_.
getOutputType returns the reference of &allOutputTypes_.

So both of them will potentially modify its member.

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,145 @@ struct InferenceContextImpl : public InferenceContext {
...
+
+    size_t output_idx = 0;
+    for (const auto& output : n.output()) {
+      outputIndexToNameMap_.insert({output_idx++, output});
+    }
+
+    allOutputTypes_.resize(n.output_size());
+  }
+
+  std::vector<const TensorProto*> allInputData_;
```
askhade, line 288: nit: move all the variables below the methods?

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,145 @@ struct InferenceContextImpl : public InferenceContext {
...
+        if (input_data->data_type() == TensorProto_DataType_INT64) {
+          tsp = vectorToTensorShapeProto(ParseData<int64_t>(input_data));
+        } else if (input_data->data_type() == TensorProto_DataType_INT32) {
+          tsp = vectorToTensorShapeProto(ParseData<int32_t>(input_data));
+        }
+
+        // Adds this TensorShapeProto from initializer into generatedShapeData
+        // for future use
+        generatedShapeData_.insert({inputIndexToNameMap_.at(index), std::move(tsp)});
+        TensorShapeProto* tsp_ptr = &tsp;
```
askhade, line 367: this wont work... you moved tsp on line 364. you should return &generatedShapeData_[inputIndexToNameMap_.at(index)]

onnx/defs/shape_inference.h
gramalingam: Can ```getOutputType``` return a ```const TypeProto*```? If so, it can be marked ```const```, right?

onnx/shape_inference/implementation.h
jcwchen: For now I still added the restriction since it's the most common case. Still, I added some comments here as TODO for future reference. Thanks for pointing out.

onnx/defs/shape_inference.h
jcwchen: Yes for `DataPropagationContext`'s `getOutputType`, it is doable. I have updated it as a const function. Thanks. 

Just FYI: `getOutputType` from `InferenceContext` is not a const function either and there are quite many places directly using getOutputType and then modify it (like Loop op or shape_inference.h)...

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,148 @@ struct InferenceContextImpl : public InferenceContext {
...
+  const TypeProto* getOutputType(size_t index) const override {
+    if (index >= allOutputTypes_.size()) {
+      ONNX_THROW("Output " + ONNX_NAMESPACE::to_string(index) + " is out of bounds.");
+    }
+    return &allOutputTypes_[index];
+  }
+
+  // Convert integer vector into TensorShapeProto
+  template <typename INTEGER>
+  TensorShapeProto vectorToTensorShapeProto(const std::vector<INTEGER>& input_vals) const {
```
askhade, line 321: nit: you can take in TensorShapeProto& tsp as an input param... this way you dont have to create and copy this obj

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,148 @@ struct InferenceContextImpl : public InferenceContext {
...
+    // Otherwise, gets it from initializer if it exists
+    const auto* input_data = allInputData_[index];
+    // Only scalar (0D tensor) or 1D tensor can be converted for now
+    // TODO: It should support tensors with more dimension on demand
+    if (input_data != nullptr &&
+        (input_data->dims_size() == 0 || input_data->dims_size() == 1)) {
+        TensorShapeProto tsp;
+
+        if (input_data->data_type() == TensorProto_DataType_INT64) {
+          tsp = vectorToTensorShapeProto(ParseData<int64_t>(input_data));
```
askhade, line 348: like I mentioned above you can change this to 

vectorToTensorShapeProto(ParseData<int64_t>(input_data), tsp);

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,148 @@ struct InferenceContextImpl : public InferenceContext {
...
+    }
+  }
+
+  std::vector<const TensorProto*> allInputData_;
+  std::unordered_map<size_t, std::string> inputIndexToNameMap_;
+  std::unordered_map<size_t, std::string> outputIndexToNameMap_;
+  std::vector<const TypeProto*> allInputTypes_;
+  std::vector<TypeProto> allOutputTypes_;
+  std::unordered_map<std::string, TensorShapeProto>& generatedShapeData_;
+  const GraphInferenceContext* graphInferenceContext_;
```
askhade, line 381: will we even need GraphInferenceContext in DataPropagation?

jcwchen: For now no, I think only control ops like `Loop` or `If` will need it and they are not included in the supported list for data propagation. I have removed it. Thanks.

onnx/shape_inference/implementation.h
```diff
@@ -241,6 +241,144 @@ struct InferenceContextImpl : public InferenceContext {
...
+  }
+
+  std::vector<const TensorProto*> allInputData_;
+  std::unordered_map<size_t, std::string> inputIndexToNameMap_;
+  std::unordered_map<size_t, std::string> outputIndexToNameMap_;
+  std::vector<const TypeProto*> allInputTypes_;
+  std::vector<TypeProto> allOutputTypes_;
+  std::unordered_map<std::string, TensorShapeProto>& generatedShapeData_;
+  std::unordered_map<std::string, const AttributeProto*> attributesByName_;
+  std::unordered_map<std::string, GraphProto*> graphProtoAttributesByName_;
```
askhade, line 379: since you removed graphinferencecontext - I dont think you need this anymore

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -0,0 +1,155 @@
...
+    ->mutable_type()->mutable_tensor_type()->mutable_shape();
+  return CompareShape(propagatedShape, outputShape);
+}
+
+TEST(DataPropagationImplTest, ShapeTest) {
+  const char* code = R"ONNX(
+agraph (int32[2,5] x) => (int32[2,5] z)
+{
+    y = Shape(x)
+    z = Cast(y)
```
gramalingam, line 109: I think "Cast" must specify the target of Cast, like "Cast<to = ...>(y)"

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -0,0 +1,155 @@
...
+
+  RunDataPropagationTest(code);
+}
+
+TEST(DataPropagationImplTest, UnsqueezeTest) {
+  const char* code = R"ONNX(
+agraph (int32[2,5] x) => (int32[2,5] w)
+{
+    y = Shape(x)
+    z = Unsqueeze(y)
```
jcwchen, line 146: Adding required input "axes" for Unsqueeze

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -131,23 +129,21 @@ agraph (int32[2,5] x) => (int32[2,5] w)
...
 TEST(DataPropagationImplTest, UnsqueezeTest) {
   const char* code = R"ONNX(
-agraph (int32[2,5] x) => (int32[2,5] w)
+agraph (int32[2,5] x, int32[1] y) => (int32[2,5] t)
 {
-    y = Shape(x)
-    z = Unsqueeze(y)
-    w = Cast(z)
+    z = Shape(x)
+    w = Unsqueeze(z, y)
```
gramalingam, line 143: I think we can make y a constant as ```y = Constant <...>()``` to test this out.

gramalingam: Of course, that would require a data-propagation for Constant. May be that is the issue (until we allow initializers in the parser).

jcwchen: Added. Since the data propagation of Unsqueeze won't use axes, not sure whether Constant works as expected in this case.

> Of course, that would require a data-propagation for Constant. May be that is the issue (until we allow initializers in the parser).

Actually I think we can just skip Constant for data propagation? Constant data can still go into inputDataByName for other op's data propagation function to use, right?

gramalingam: You are right, my mistake, sorry. Please ignore my comment.

gramalingam: I mean it is not required for Unsqueeze

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -0,0 +1,156 @@
...
+        fail_check("The compared shapes should not be nullptr.");
+        return false;
+    }
+    if (A->dim_size() != B->dim_size()) {
+        fail_check("The compared sizes of dim are different.");
+        return false;
+    }
+    for (int i = 0; i < A->dim_size() ; ++i) {
+        if (A->dim(i).has_dim_value() != B->dim(i).has_dim_value() || 
+                A->dim(i).dim_value() != B->dim(i).dim_value()) {
```
askhade, line 32: this check does not seem right... 
if A->dim(i).has_dim_value() is false then you should not fetch dim value... also why is there no check for dim_param
from the function name it seems like you are simply comparing the shapes but you are actually just checking the dim values

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -0,0 +1,156 @@
...
+  EXPECT_TRUE(CompareShape(propagatedShape, outputShape));
+}
+
+TEST(DataPropagationImplTest, ShapeTest) {
+  const char* code = R"ONNX(
+agraph (int32[7,4,1] x) => (int32[7,4,1] y)
+{
+    xs = Shape(x)
+    y = Cast<to = 7>(xs)
+}
```
askhade, line 114: this graph is invalid...
Shape of x is [7, 4, 1] this is correct however the shape of xs and y will not be [7, 4, 1] it will be [3] because shaoe node will output [7, 4,3] and its shape will be [3]... similarly cast will simply cast the data without changing the shape so shape of y will be [3]

Your test passes because you are not calling TypeAndShapeInferenceFunction for the node before calling DataPropagationFunction

Also check on line 105 is not right... that function should simply return the computed shape data... every individual node test should run comparision... for nodes like shape you will need to hard code the expected output

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -0,0 +1,156 @@
...
+  RunDataPropagationTest(code);
+}
+
+TEST(DataPropagationImplTest, CastTest) {
+  const char* code = R"ONNX(
+agraph (int32[2,5] x) => (int32[2,5] y)
+{
+    xs = Shape(x)
+    y = Cast<to = 7>(xs)
+}
```
askhade, line 125: same comment as above

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -0,0 +1,156 @@
...
+TEST(DataPropagationImplTest, SqueezeTest) {
+  const char* code = R"ONNX(
+agraph (int32[2,5] x) => (int32[2,5] z)
+{
+    xs = Shape(x)
+    y = Squeeze(xs)
+    z = Cast<to = 7>(y)
+}
+)ONNX";
+  RunDataPropagationTest(code);
```
askhade, line 139: same comment as above .... the shape changes after Shape node...

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -0,0 +1,166 @@
...
+    }
+    for (int i = 0; i < A->dim_size() ; ++i) {
+      if (A->dim(i).has_dim_value() && B->dim(i).has_dim_value()) {
+        if (A->dim(i).dim_value() != B->dim(i).dim_value()) {
+        fail_check("The compared dim values are different.(",
+          A->dim(i).dim_value(), ") vs (", B->dim(i).dim_value(), ").");
+        return false;
+        }
+      } else if (A->dim(i).has_dim_param() && B->dim(i).has_dim_param()) {
+        continue;
```
jcwchen, line 38: Do we need to further compare the data_params here?

askhade (approved):

postrational:
Unfortunately code from this change breaks building on older compilers (break occurs on CentOS 7.6)

jcwchen:
> Unfortunately code from this change breaks building on older compilers (break occurs on CentOS 7.6)

@postrational Do you have detailed errors that I can look into? Thanks!

take-cheeze:
Now I've enabled data propagation in onnxsim: https://github.com/onnxsim/onnxsim/pull/467
Good news that I don't need complex optimization pass like eliminate_shape_gather

PR #3558, opened 2021-07-06
Add Cast-like operator

gramalingam:

**Description**
Introduces an operator CastLike(X,Y) that casts X to the same type as Y.

**Motivation and Context**
A common use-case in the definition of function-bodies is to produce a constant, like 0.5 or 1, of the same type/precision as an input Y. This enables such functionality (without requiring the use of a type-context and a context-dependent-function). This will also be useful in defining similar functions that are stored in a model (aka model-local functions).


jcwchen (commented):
Please note that TestCoverage.md needs to be updated as well

onnx/defs/tensor/defs.cc
```diff
@@ -103,6 +104,101 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+		"output",
+		"Output tensor produced by casting the first input tensor to have the same type as the second input tensor.",
+		"T2",
+		OpSchema::Single,
+		true,
+		1,
+		OpSchema::Differentiable)
+	.TypeConstraint(
+		"T1",
+		{ "tensor(float16)",
```
askhade, line 147: use OpSchema::all_tensor_types_with_bfloat() instead? for better readability

onnx/defs/tensor/defs.cc
```diff
@@ -103,6 +104,101 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+		 "tensor(uint16)",
+		 "tensor(uint32)",
+		 "tensor(uint64)",
+		 "tensor(bool)",
+		 "tensor(string)",
+		 "tensor(bfloat16)" },
+		"Constrain input types. Casting from complex is not supported.")
+	.TypeConstraint(
+		"T2",
+		{ "tensor(float16)",
```
askhade, line 164: same as above use : OpSchema::all_tensor_types_with_bfloat()

onnx/defs/tensor/defs.cc
gramalingam: That includes the complex types, but cast doesn't support complex. I have copied the same types as in cast.

onnx/backend/test/case/node/castlike.py
```diff
@@ -0,0 +1,100 @@
...
+                    u'0.47031248', u'0.816468', u'0.21087195', u'0.7229038',
+                    u'NaN', u'INF', u'+INF', u'-INF'], dtype=np.dtype(np.object)).reshape([3, 4])
+                output = input.astype(TENSOR_TYPE_TO_NP_TYPE[getattr(TensorProto, to_type)])
+            like = output.flatten()[0:1]
+            node = onnx.helper.make_node(
+                'CastLike',
+                inputs=['input', 'like'],
+                outputs=['output'],
+            )
+            if input_type and output_type:
```
askhade, line 93: not sure I understand why would these 2 be None?

gramalingam: I copied it from Cast. My guess is that minor adjustments were needed for types like bfloat16 because of the use an int-representation-type for the float-value (so, the call to ```expect``` is different here, and the representation-type is explicitly mentioned).

askhade (approved):

jcwchen (approved):

askhade (approved):

Issue #3562, opened 2021-07-08
onnx-tensorflow onnx backend test of case test_identity_sequence_cpu failed


chudegao:
# Bug Report

### Is the issue related to model conversion?
No

### Describe the bug
When run onnx backend test case of test_identity_sequence_cpu with onnx-tensorflow, hit below error:

======================================================================
ERROR: test_identity_sequence_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/anaconda3/envs/otf/lib/python3.7/site-packages/onnx-1.9.0-py3.7-linux-x86_64.egg/onnx/backend/test/runner/__init__.py", line 250, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/opt/anaconda3/envs/otf/lib/python3.7/site-packages/onnx-1.9.0-py3.7-linux-x86_64.egg/onnx/backend/test/runner/__init__.py", line 308, in run
    atol=model_test.atol)
  File "/opt/anaconda3/envs/otf/lib/python3.7/site-packages/onnx-1.9.0-py3.7-linux-x86_64.egg/onnx/backend/test/runner/__init__.py", line 172, in assert_similar_outputs
    np.testing.assert_equal(outputs[i].dtype, ref_outputs[i].dtype)
AttributeError: 'list' object has no attribute 'dtype'

----------------------------------------------------------------------
Ran 1 test in 0.970s

FAILED (errors=1)

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): 
- ONNX version 1.9.0 and master
- Python version:3.7

### Reproduction instructions
- Describe the code to reproduce the behavior.
```
import onnx
from onnx_tf.backend import TensorflowBackend
backend_test = onnx.backend.test.runner.Runner(TensorflowBackend, __name__)
if __name__ == '__main__':
  unittest.main()

...
```
- Attach the ONNX model to the issue (where applicable)

### Expected behavior
The test case should pass (when onnx_tf implemented the operator correctly)

### Notes

The root cause is function assert_similar_outputs() in onnx/backend/test/runner/__init__.py does not handle sequence/list as model/node's output, it always check dtype of the output. 

```
np.testing.assert_equal(outputs[i].dtype, ref_outputs[i].dtype)
```

chudegao:
I will submit a PR for this issue.

jcwchen:
[...]

PR #3563, opened 2021-07-08
fix test_identity_sequence issue

chudegao:
Signed-off-by: degaochu <degaochu@cn.ibm.com>

**Description**
- update function assert_similar_outputs() in onnx/backend/test/runner/init.py to support sequence/list as model/node's output.

**Motivation and Context**
- To fix onnx backend test failed for sequence output operators like identity operator
- fix issue https://github.com/onnx/onnx/issues/3562

### DIFF
--- a/onnx/backend/test/runner/__init__.py
+++ b/onnx/backend/test/runner/__init__.py
@@ -170,7 +170,8 @@ class Runner(object):
         np.testing.assert_equal(len(outputs), len(ref_outputs))
         for i in range(len(outputs)):
             if isinstance(outputs[i], (list, tuple)):
-                np.testing.assert_array_equal(outputs[i], ref_outputs[i])
+                for j in range(len(outputs[i])):
+                    cls.assert_similar_outputs(ref_outputs[i][j], outputs[i][j], rtol, atol)
             else:
                 np.testing.assert_equal(outputs[i].dtype, ref_outputs[i].dtype)
                 if ref_outputs[i].dtype == np.object:
--- a/onnx/backend/test/runner/__init__.py
+++ b/onnx/backend/test/runner/__init__.py
@@ -169,15 +169,18 @@ class Runner(object):
     def assert_similar_outputs(cls, ref_outputs, outputs, rtol, atol):  # type: (Sequence[Any], Sequence[Any], float, float) -> None
         np.testing.assert_equal(len(outputs), len(ref_outputs))
         for i in range(len(outputs)):
-            np.testing.assert_equal(outputs[i].dtype, ref_outputs[i].dtype)
-            if ref_outputs[i].dtype == np.object:
+            if isinstance(outputs[i], (list, tuple)):
                 np.testing.assert_array_equal(outputs[i], ref_outputs[i])
             else:
-                np.testing.assert_allclose(
-                    outputs[i],
-                    ref_outputs[i],
-                    rtol=rtol,
-                    atol=atol)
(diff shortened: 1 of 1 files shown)


onnx/backend/test/runner/__init__.py
```diff
@@ -169,15 +169,18 @@ def tests(self):  # type: () -> Type[unittest.TestCase]
     def assert_similar_outputs(cls, ref_outputs, outputs, rtol, atol):  # type: (Sequence[Any], Sequence[Any], float, float) -> None
         np.testing.assert_equal(len(outputs), len(ref_outputs))
         for i in range(len(outputs)):
-            np.testing.assert_equal(outputs[i].dtype, ref_outputs[i].dtype)
-            if ref_outputs[i].dtype == np.object:
+            if isinstance(outputs[i], (list, tuple)):
```
jcwchen, line 172: Just out of curiosity: is there any example for tuple output?

onnx/backend/test/runner/__init__.py
```diff
@@ -169,15 +169,18 @@ def tests(self):  # type: () -> Type[unittest.TestCase]
...
                 np.testing.assert_array_equal(outputs[i], ref_outputs[i])
```
gramalingam, line 173: I don't understand how this is supposed to work. In particular, I wonder if this assumes some restrictions on the types of the values stored in the list. Will it work for a sequence of sequence of tensors, for example? I think it may be better to recursively call assert_similar_outputs for every pair of corresponding elements.

onnx/backend/test/runner/__init__.py
chudegao: Acutally I did not hit tuple output so far. Just add tuple in case for some special case. Let me know I should keep it or remove it.

onnx/backend/test/runner/__init__.py
chudegao: > I don't understand how this is supposed to work. In particular, I wonder if this assumes some restrictions on the types of the values stored in the list. Will it work for a sequence of sequence of tensors, for example? I think it may be better to recursively call assert_similar_outputs for every pair of corresponding elements.

I checked onnx operator output, it only support seq(tensor) . But it's good idea to to recursively call assert_similar_outputs(). I will update. Thanks.

chudegao: @gramalingam I updated to call assert_similar_outputs() recursively, please review.

onnx/backend/test/runner/__init__.py
```diff
@@ -169,15 +169,19 @@ def tests(self):  # type: () -> Type[unittest.TestCase]
...
+                for j in range(len(outputs[i])):
+                    cls.assert_similar_outputs(ref_outputs[i][j], outputs[i][j], rtol, atol)
             else:
-                np.testing.assert_allclose(
-                    outputs[i],
-                    ref_outputs[i],
-                    rtol=rtol,
-                    atol=atol)
+                np.testing.assert_equal(outputs[i].dtype, ref_outputs[i].dtype)
+                if ref_outputs[i].dtype == np.object:
```
gramalingam, line 177: It would be nice to have a comment to explain what case is handled by the then branch. (I understand that this is a problem with the pre-existing code, but it would help readers/reviewers/maintainers in the future.)

gramalingam: In case you know.

gramalingam (approved):

jcwchen (approved):

PR #3566, opened 2021-07-08
Sync protobuf version in every CI: Install protobuf 3.11.3 in LinuxRelease CI

jcwchen:
**Description**
Previously apt-get version of protobuf should be an old version... To sync protobuf version in every CI, install protobuf 3.11.3 in LinuxRelease CI

**Motivation and Context**
Sync protobuf version in every CI

### DIFF
--- a/.github/workflows/release_linux_x86_64.yml
+++ b/.github/workflows/release_linux_x86_64.yml
@@ -34,9 +34,22 @@ jobs:
         python-version: ${{ matrix.python-version }}
         architecture: ${{ matrix.architecture }}
 
+    - name: Install python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install numpy protobuf==3.11.3
+
+    - name: Build manylinux2010_x86_64
+      uses: docker://quay.io/pypa/manylinux2010_x86_64
+      with:
+        entrypoint: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/release_linux_x86_64.yml
```diff
@@ -34,11 +34,24 @@ jobs:
         python-version: ${{ matrix.python-version }}
         architecture: ${{ matrix.architecture }}
 
+    - name: Install protobuf
```
askhade, line 37: why are you installing protobuf here?
entrypoint.sh does this right?
https://github.com/onnx/onnx/blob/master/.github/workflows/manylinux/entrypoint.sh#L18

jcwchen: entrypoint.sh installs protobuf inside the docker for building the wheel. The protobuf installation here is in the GitHub Action environment for testing the wheel.

askhade: can you add this comment to the yml to make this more clear. Thanks!

askhade: also since this is relevant for testing can we move this below just before testing starts?

jcwchen: Updated. Thanks for the suggestions!

askhade (approved):

PR #3567, opened 2021-07-08
Adding Optional ops to opset 15

neginraoof:
**Description**
Adding operators for optional type to opset 15:
Optional (constructor)
OptionalHasElement
OptionalGetElement

**Motivation and Context**
These ops are necessary for enabling export of customer models with optional type.


askhade:
@postrational can you help review this. We need to get this in onnx 1.10

onnx/backend/test/case/node/optionalgetelement.py
```diff
@@ -0,0 +1,48 @@
...
+            'OptionalGetElement',
+            inputs=['optional_input'],
+            outputs=['output']
+        )
+        output = optional_get_element_reference_implementation(optional)
+        expect(node, inputs=[optional], outputs=[output],
+               name='test_optional_get_element')
+
+    @staticmethod
+    def export_empty():  # type: () -> None
```
askhade, line 38: nit: rename to export_get_sequence

onnx/backend/test/case/node/optionalhaselement.py
```diff
@@ -0,0 +1,49 @@
...
+from typing import Optional, List, Any
+
+import onnx
+from ..base import Base
+from . import expect
+
+
+def optional_has_element_reference_implementation(optional):  # type: (Optional[np.ndarray]) -> np.ndarray
+    if optional is None:
+        return np.array(False)
```
gramalingam, line 18: Does this map to a one-dimensional tensor in ONNX? I think we want a scalar (zero-dimensional tensor), right?

onnx/defs/optional/defs.cc
```diff
@@ -0,0 +1,139 @@
...
+          const size_t numInputs = ctx.getNumInputs();
+          const auto* attr_proto = ctx.getAttribute("type");
+
+          if ((numInputs == 0) && (attr_proto != nullptr)) {
+            if (!attr_proto->has_tp())
+              fail_type_inference("Attribute 'type' should be a TypeProto and it should specify a type.");
+            auto attr_tp = attr_proto->tp();
+
+            ctx.getOutputType(0)->mutable_optional_type()->mutable_elem_type()->CopyFrom(attr_tp);
+          } else if (numInputs == 1) {
```
gramalingam, line 54: Add ```&& (attr_proto == nullptr)``` to the condition.

onnx/defs/optional/defs.cc
```diff
@@ -0,0 +1,139 @@
...
+#include "onnx/defs/function.h"
+#include "onnx/defs/schema.h"
+
+#include <algorithm>
+#include <numeric>
+
+namespace ONNX_NAMESPACE {
+
+static const char* Optional_ver1_doc = R"DOC(
+Construct an optional type containing either an empty optional of a certain type specified by the attribute,
```
gramalingam, line 14: change ```optional type``` to ```optional type value```

onnx/defs/optional/defs.cc
```diff
@@ -0,0 +1,139 @@
...
+or an optional type containing the 'input' element.
```
gramalingam, line 15: change ```optional type``` to ```non-empty value```

onnx/defs/optional/defs.cc
```diff
@@ -0,0 +1,139 @@
...
+        .Output(0, "output", "The optional output enclosing the input element.", "O")
+        .TypeConstraint(
+            "V",
+            [](){
+              auto t = OpSchema::all_tensor_types();
+              auto s = OpSchema::all_tensor_sequence_types();
+              t.insert(t.end(), s.begin(), s.end());
+              return t;
+            }(),
+            "Constrains input type to all tensor and sequence types.")
```
gramalingam, line 34: General comment: We are forced to enumerate the set of allowed types here by ONNX's current design. This makes these ops more restrictive than necessary. We should generalize the implementation to allow unconstrained types as discussed in https://github.com/onnx/onnx/issues/2324. However, I doubt if that can be done by the deadline for current release, so I feel it is okay to have this in this PR for the short term.

onnx/numpy_helper.py
```diff
@@ -86,50 +86,52 @@ def from_array(arr, name=None):  # type: (np.ndarray[Any], Optional[Text]) -> Te
...
-        dtype = mapping.NP_TYPE_TO_TENSOR_TYPE[arr.dtype]
-    except KeyError:
-        raise RuntimeError(
-            "Numpy data type not understood yet: {}".format(str(arr.dtype)))
-    tensor.data_type = dtype
-    tensor.raw_data = arr.tobytes()  # note: tobytes() is only after 1.9.
-    if sys.byteorder == 'big':
-        # Convert endian from big to little
-        convert_endian(tensor)
+    if arr is not None:
```
gramalingam, line 92: I don't understand why None is allowed here. The python type signature above is inconsistent with this. I think there might be a better way to handle this at the caller, instead of changing this function. Where is it being called with None?

onnx/test/shape_inference_test.py
```diff
@@ -60,16 +60,28 @@ def _assert_inferred(self, graph, vis, **kwargs):  # type: (GraphProto, List[Val
...
+                    assert inferred_dim.dim_param, '\n%s\n%s\n' % (vi, inferred_vi)
                 else:
-                    assert dim.dim_value == inferred_dim.dim_value, '\n%s\n%s\n' % (vis, inferred_vis)
+                    assert dim.dim_value == inferred_dim.dim_value, '\n%s\n%s\n' % (vi, inferred_vi)
+        elif vi.HasField('sequence_type'):
+            vi = vi.sequence_type.elem_type
+            inferred_vi = inferred_vi.sequence_type.elem_type
+            self._compare_value_infos(vi, inferred_vi)
+        elif vi.HasField('optional_type'):
+            vi = vi.optional_type.elem_type
```
gramalingam, line 82: May be we should ```assert inferred_vi.HasField('optional_type')``` here.  (Similarly above for other cases.)

onnx/backend/test/case/node/optionalhaselement.py
neginraoof: This is a zero dimension numpy scalar. The ONNX model output value info does not have shape (only boolean dtype).

onnx/defs/optional/defs.cc
neginraoof: Makes sense. Thanks.

onnx/backend/test/case/node/__init__.py
```diff
@@ -103,27 +103,19 @@ def function_testcase_helper(node, input_types, name):  # type: (NodeProto, List
...
-            optional_type_proto = onnx.helper.make_optional_type_proto(tensor_type_proto)
-            return onnx.helper.make_value_info(name=name, type_proto=optional_type_proto)
-    return onnx.helper.make_tensor_value_info(
-        name=name,
-        elem_type=ele_type if ele_type else onnx.mapping.NP_TYPE_TO_TENSOR_TYPE[input.dtype],
-        shape=input.shape)
+def _extract_value_info(input, name, type_proto=None):  # type: (Union[List[Any], np.ndarray, None], Text, TypeProto) -> onnx.ValueInfoProto
+    if type_proto is None:
+        if isinstance(input, list):
+            elem_type = onnx.mapping.NP_TYPE_TO_TENSOR_TYPE[input[0].dtype]
```
askhade, line 109: please add this comment back :

 # TODO: Account for recursive sequence case. Right now, this function supports
 # Sequences of Tensors.

we are still only supporting Sequence of tensors

askhade: In the case where type_proto is None there is no else block to set type_proto... meaning if control does not enter 
if isinstance(input, list) or elif isinstance(input, np.ndarray) or np.isscalar(input) then type_proto will be None when the control reaches line 118.

askhade: how will this work in the case when optional proto holds None value? type_proto will be None?

onnx/numpy_helper.py
```diff
@@ -86,10 +86,11 @@ def from_array(arr, name=None):  # type: (np.ndarray[Any], Optional[Text]) -> Te
         tensor_def: the converted tensor def.
     """
     tensor = TensorProto()
-    tensor.dims.extend(arr.shape)
     if name:
         tensor.name = name
 
+    tensor.dims.extend(arr.shape)
```
gramalingam, line 92: Is this an accidental change? Moving the line around doesn't seem to change anything.

onnx/defs/optional/defs.cc
gramalingam: I don't understand why this is marked as resolved. What happens if both an input is supplied and the type attribute is specified? I think it is safer to flag it as an error.

onnx/backend/test/case/node/__init__.py
gramalingam: Agree. Better to have an assertion/check if the value is None.

neginraoof: @askhade I don't think we need the comment back. The function now does support sequence of sequence or other recursive types if the type_proto for such type is provided. So users can create this kind of type_proto and then use _extract_value_info to create the value info.

neginraoof: @askhade In the case where optional_proto holds None, we rely on the user to provide a corresponding type_proto. For a None tensor, the type_proto has elem_type (depending on the None tensor data type).
I'll add a check to fail if both input and type_proto are None.

onnx/numpy_helper.py
neginraoof: Yes, thanks. This is not intended.

onnx/backend/test/case/node/__init__.py
```diff
@@ -103,27 +103,21 @@ def function_testcase_helper(node, input_types, name):  # type: (NodeProto, List
...
-        if ele_type is None:
-            raise NotImplementedError("extract_value for optional tensor with no data type not "
-                                      "supported.")
+def _extract_value_info(input, name, type_proto=None):  # type: (Union[List[Any], np.ndarray, None], Text, Optional[TypeProto]) -> onnx.ValueInfoProto
+    if type_proto is None:
+        if input is None:
+            raise NotImplementedError("_extract_value_info: both input and type_proto arguments cannot be None.")
+        elif isinstance(input, list):
+            elem_type = onnx.mapping.NP_TYPE_TO_TENSOR_TYPE[input[0].dtype]
+            shape = input[0].shape
```
gramalingam, line 112: shape should be None as before. We can't assume all elements in the sequence have the same shape (unless we check and verify this).

docs/Operators.md
```diff
@@ -12761,6 +12767,200 @@ expect(node, inputs=[indices, depth, values], outputs=[y], name='test_onehot_wit
...
+<dd>Constrains output type to all optional tensor or optional sequence types.</dd>
+</dl>
+
+
+### <a name="OptionalGetElement"></a><a name="optionalgetelement">**OptionalGetElement**</a>
+
+  Outputs the element in the optional-type input'. It is an error if the input value does not have an element "
+  "and the behavior is undefined in this case."
```
hariharans29, line 12813: very minor nit: is the single quote and the superfluous double quotes intentional ?

gramalingam (approved):

askhade (approved):

hariharans29 (approved):
Thanks for this PR !

PR #3569, opened 2021-07-09
checker updates for model local functions

askhade:

**Description**
Adding checks for opset compatibility between different model local functions and relaxing schema check in opset compatibility when the ops do not belong to onnx standard domains.
- Describe your changes.

**Motivation and Context**

### DIFF
--- a/onnx/checker.cc
+++ b/onnx/checker.cc
@@ -720,14 +720,23 @@ void check_opset_compatibility(
     return;
   }
 
+  if (func_opset_version == model_opset_version) {
+    // both versions are same, no need to verify schema.
+    return;
+  }
+
   const auto* schema_for_model_import =
       ctx.get_schema_registry()->GetSchema(node.op_type(), model_opset_version, node.domain());
 
   const auto* schema_for_function_import =
       ctx.get_schema_registry()->GetSchema(node.op_type(), func_opset_version, node.domain());
 
-  if (schema_for_model_import && schema_for_function_import &&
-      schema_for_function_import->since_version() == schema_for_model_import->since_version()) {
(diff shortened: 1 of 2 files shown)


onnx/checker.cc
```diff
@@ -720,14 +720,23 @@ void check_opset_compatibility(
...
+
   const auto* schema_for_model_import =
       ctx.get_schema_registry()->GetSchema(node.op_type(), model_opset_version, node.domain());
 
   const auto* schema_for_function_import =
       ctx.get_schema_registry()->GetSchema(node.op_type(), func_opset_version, node.domain());
 
-  if (schema_for_model_import && schema_for_function_import &&
-      schema_for_function_import->since_version() == schema_for_model_import->since_version()) {
+  if (!schema_for_model_import || !schema_for_function_import) {
```
gramalingam, line 734: The checker can be called by ONNX's own whole-model-checker but also by others (like ORT). I wonder if this is the behavior we want when it is called by ORT? We could consider the following options: (a) check if domain == standard domain directly, (b) replace the "||" by "&&", or (c) Have the caller specify via an option whether they want a strict check here or not. I think (b) may be a reasonable compromise for now. (If one is available and the other is not, treat it as error?)

gramalingam: A combination of (a) and (b) would be to add an extra-check that the domain is not the onnx standard-domain.

askhade: combination of a and b makes sense for now... we can take up option c later and apply it for check_node function as well

askhade: actually just b should suffice... right after this check_node is called and it does check if domain is one of the standard domains... adding (a) would make the check duplicate

gramalingam (approved):

PR #3574, opened 2021-07-12
Dim op

gramalingam:
(Please note that there is a newer PR https://github.com/onnx/onnx/pull/3580 proposed as an alternative to this PR, which just updates the existing ```Shape``` op instead of introducing a new op. Only one needs to be selected. I think the other PR is preferable.)

**Description**
Introduces a new operator Dim, that returns the size of the input tensor along a specified axis.

**Motivation and Context**
This is a commonly useful primitive operation, eg., used to extract the batchsize from the input.
This is defined as a function-op.
Creating this PR as per discussion and agreement in the last operator SIG meeting.


onnx/backend/test/case/node/dim.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @staticmethod
+    def export_axis_0():  # type: () -> None
+        x = np.random.randn(3, 4, 5).astype(np.float32)
+        test_dim(x, 0, 'axis_0')
+
+    @staticmethod
+    def export_axis_minus1():  # type: () -> None
+        x = np.arange(24, dtype=np.int32).reshape((2, 3, 4))
+        test_dim(x, -1, 'axis_minus1')
```
jcwchen, line 44: nit: perhaps axis_negative_1? negative is more commonly used as test name under onnx/backend/test/data/node/

onnx/defs/tensor/defs.cc
```diff
@@ -371,6 +373,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+Takes an input tensor and the index (axis) of the dimension to extract (as an attribute) and
+returns the dimension value at the requested index (axis) in the input shape.
+The op returns the output as a 1-dimensional tensor of size 1,
+for the common usage scenario of creating other shapes (as 1-dimensional tensors).
+The axis specified must be a value in the range [-r, r-1] where r = rank(input).
+Otherwise, the output value is undefined.
+ 
+For example: 
+Input tensor with shape: [2, 3, 4] 
+Index: 1
```
askhade, line 387: nit: use axis instead of Index?

onnx/defs/tensor/defs.cc
```diff
@@ -371,6 +373,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+ONNX_OPERATOR_SET_SCHEMA(
+    Dim,
+    15,
+    OpSchema()
+        .SetDoc(Dim_ver15_doc)
+        .Input(0, "input", "An input tensor.", "T", OpSchema::Single, true, 1, OpSchema::NonDifferentiable)
+        .Output(
+            0,
+            "output",
+            "Dimension of the input tensor at the specified axis.",
```
askhade, line 405: nit: elaborate this sentence similar to:  Dimension of the given input tensor's shape at the specified axis.

onnx/defs/tensor/defs.cc
```diff
@@ -371,6 +373,80 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          auto* type = ctx.getOutputType(0)->mutable_tensor_type();
+          type->set_elem_type(TensorProto::INT64);
+          // output of this operator is a 1D tensor of length 1
+          auto* shape = type->mutable_shape();
+          shape->add_dim()->set_dim_value(1);
+
+          if (hasInputShape(ctx, 0)) {
+            // Check that axis attribute has a proper value.
+            int rank = getInputShape(ctx, 0).dim_size();
+            int axis = static_cast<int>(getAttribute(ctx, "axis", 0));
```
askhade, line 427: the spec does not mention a default value... should this be changed to 1. first verify attr has a value and then validate the value?

gramalingam: There is no default-value, it is a required attribute. The default-checker would already have validated that the attribute is specified. I used ```getAttribute(...)``` as a short-cut (that does the checks anyway), but the default-value won't be used. I could replace this code by ```ctx.getAttribute("axis")->i()``` which is equivalent, if that seems preferable. I am inclined to avoid redundant checks here.

onnx/defs/tensor/defs.cc
gramalingam: Sure. Perhaps something like "The slice of the input tensor's shape at the specified axis"? This started me thinking about the name of the op. The word "dimension" is itself overloaded, since it is also used, like "axis", to indicate a specific axis. But I can't think of a better name for the operator itself. MLIR does use the same opname "Dim", for what it is worth. (We could overload "Shape", making axis an optional parameter ... we can't define it as a function then, but this could be an option.)

onnx/defs/tensor/defs.cc
askhade: if checker is run on the model then yes this should be caught... my point was current code makes you think default value is 0 which is not right... maybe add a comment 

My preference is to use getAttribute, validate not nullptr (fail if nullptr) and fetch value like we do for other required attrs

gramalingam:
Closing this in favor of #3580 .

PR #3580, opened 2021-07-14
Extend shape op to return a slice

gramalingam:

**Description**
Extend the shape op by adding optional attributes start/end so that the return value is input_tensor.shape[start:end].

**Motivation and Context**
Helps abbreviate a common pattern, eg., of extracting a batch-size from a tensor. Note that this extension is an alternative to introducing a separate "Dim" op (see PR: https://github.com/onnx/onnx/pull/3574 ), so only one of the two PRs is required.


rajeevsrao:
ONNX-TensorRT can support either this or the original (`Dim` op) proposal, with a slight preference for using this approach (using shape op and return slice).

onnx/defs/tensor/defs.cc
```diff
@@ -378,53 +378,72 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-          ctx.getOutputType(0)->mutable_tensor_type()->set_elem_type(
-              TensorProto::INT64);
-          auto* output_shape = 
-              ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+          ctx.getOutputType(0)->mutable_tensor_type()->set_elem_type(TensorProto::INT64);
+          auto* output_shape = ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
           auto* output_length = output_shape->add_dim();
-		
+
           if (!hasNInputShapes(ctx, 1)) {
```
askhade, line 434: should this check be before line 432?
If input shape is not known do we want to add dim?

onnx/defs/tensor/defs.cc
```diff
@@ -378,53 +378,72 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             return;
           }
 
           if (ctx.getInputType(0)->tensor_type().has_shape()) {
```
askhade, line 438: is this if required? if (!hasNInputShapes(ctx, 1)) is checking the same thing right?

onnx/defs/tensor/defs.cc
```diff
@@ -378,53 +378,72 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            output_length->set_dim_value(
-                ctx.getInputType(0)->tensor_type().shape().dim_size());
+            int64_t rank = static_cast<int64_t>(ctx.getInputType(0)->tensor_type().shape().dim_size());
+            int64_t start = getAttribute(ctx, "start", 0);
+            if (start < 0)
+              start += rank;
+            int64_t end = getAttribute(ctx, "end", rank);
+            if (end < 0)
+              end += rank;
+            output_length->set_dim_value(end - start);
```
askhade, line 446: validate the end-start before setting dim_value?
make sure end -start is positive > 0

askhade:
I think this approach is better over introducing Dim (PR #3574 )

askhade (approved):

onnx/defs/tensor/defs.cc
hariharans29: @gramalingam @askhade : Is it worth explicitly calling out the behavior in the spec that by setting start > end (and start == end because of the exclusive nature of end) the resultant shape slice will be empty (over it being an error) ? Just by reading the spec, this portion was unclear to me until I saw the shape inference. (Sorry if I overlooked something obvious in the spec)

PR #3583, opened 2021-07-15
Use 0D tensors for scalar inputs in testcases

shinh:
**Description**

Testcases of OneHot, SequenceInsert, and NonMaxSuppression were using 1D tensor as scalar inputs but their documents say they accept scalars.

**Motivation and Context**

These tests look inconsistent with the document and I could not find any document which mentions 1D tensor with a single element can be treated as a scalar. Especially, the document of SequenceInsert clearly states `position` should be "tensor of empty shape".

I felt the second input of TopK op should be also a scalar, but the document says it takes 1D tensor with a single element.


askhade (commented):
PR looks good to me. Please resolve conflicts and then we can merge this

PR #3587, opened 2021-07-16
Update ONNX release, IR, and opset versions for v1.10.0

rajeevsrao:
**Description**
- Update ONNX release version to 1.10.0
- Update ONNX IR version to 8
- Update ONNX opset version to 15

**Motivation and Context**
- Preparation for ONNX release v1.10.0


VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.9.0
+1.10.0
```
jcwchen, line 1: Maybe something like 1.9.2 for VERSION_NUMBER? Typically we use 1.10.0 for the final verified package

rajeevsrao: @jcwchen should I use the 1.9.x version only on release branch (after this is merged to `master`) and prior to generating the test wheels as the workflow document suggested or do that even for `master`?

rajeevsrao: Updated as suggested - downgraded release version to 1.9.100 for test package.

jcwchen: @rajeevsrao Formerly ONNX <= 1.8 did bump the version to the target version directly in the main branch. However, ONNX has an auto-generated weekly package from main branch now so if the release is not completed, there might be a 1.10 onnx-weekly package first before the real 1.10 release... which is quite confusing.

@askhade Do you have any preference for it (bumping to which VERSION_NUMBER)?

rajeevsrao: I changed it to v1.9.100 to be safe.

VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.9.0
+1.9.100
```
askhade, line 1: I suggest changing this to 1.10.0 in master.... In release branch create a new PR to just change the version in this file to 1.9.100

The idea is any new PR going into master post this PR is merged will be towards the next release.

onnx/common/version.h
```diff
@@ -8,6 +8,6 @@
 namespace ONNX_NAMESPACE {
 
 // Represents the most recent release version. Updated with every release.
-constexpr const char* LAST_RELEASE_VERSION = "1.9.0";
+constexpr const char* LAST_RELEASE_VERSION = "1.9.100";
```
askhade, line 11: this should be 1.10.0

rajeevsrao: Switched it back to 1.10 based on comments.

VERSION_NUMBER
rajeevsrao: Thanks - will use 1.9.100 only on the release branch.

askhade (approved):

jcwchen (approved):

PR #3590, opened 2021-07-17
[cmake] Avoid installing empty directories

stephenneuendorffer:
install(DIRECTORY ... FILES_MATCHING) will install files
matching a given pattern, but will not exclude empty directories.
Because of the huge number of tests in backend/test that do not
contain header files, this results in many empty directories,
significantly slowing down the installation process and causes
unnecessary file system pressure.  This patch explicitly excludes
the worst offending directories.


CMakeLists.txt
```diff
@@ -669,7 +669,9 @@ include(GNUInstallDirs)
 install(DIRECTORY ${ONNX_ROOT}/onnx
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
         FILES_MATCHING
-        PATTERN "*.h")
+        PATTERN "*.h"
+        PATTERN "backend/test/case" EXCLUDE
+        PATTERN "backend/test/data" EXCLUDE)
```
jcwchen, line 674: tiny nit: it would be helpful if we can have a comment about why we exclude it for future reference.

jcwchen (approved):
LGTM. Thank you for the PR!

PR #3592, opened 2021-07-19
update reshape shape inference

askhade:

**Description**
- Updating reshape shape inference to pick up shape input generated from partial data propagation.
- Added tests for reshape changes
- Change usage of SymbolTableImpl... all the functions should use SymbolTable interface and not SymbolTableImpl
- Small fixes\minor updates to data propagation tests

**Motivation and Context**


onnx/shape_inference/implementation.cc
```diff
@@ -355,7 +355,7 @@ static void InferShapesImpl(
     }
     auto domain_version = dit->second;
     const auto schema = schema_registry->GetSchema(n.op_type(), domain_version, n.domain());
-    InferenceContextImpl ctx(n, valueTypesByName, inputDataByName, inputSparseDataByName, &graphInferenceContext);
+    InferenceContextImpl ctx(n, valueTypesByName, inputDataByName, inputSparseDataByName, {}, & graphInferenceContext);
```
askhade, line 358: will update once #3551 is checked in

onnx/shape_inference/implementation.cc
```diff
@@ -588,7 +588,7 @@ void InferShapeForFunctionNode(
     }
 
     InferenceContextImpl temp_ctx(
-        copy_n, temp_valueTypesByName, temp_initializersByName, temp_SparseInitializersByName);
+        copy_n, temp_valueTypesByName, temp_initializersByName, temp_SparseInitializersByName, {});
```
askhade, line 591: will update once #3551 is checked in

onnx/shape_inference/implementation.cc
```diff
@@ -122,20 +122,20 @@ void checkShapesAndTypes(const TypeProto& inferredType, const TypeProto& existin
...
+void generateSymbolicShape(TensorTypeProto* inferredType, SymbolTable& symbolTable) {
   if (!inferredType->has_shape()) {
     return;
   }
   for (int i = 0; i < inferredType->shape().dim_size(); ++i) {
     // set a symbol if it doesn't have dim_value and dim_param
     auto* dim = inferredType->mutable_shape()->mutable_dim(i);
     if (!dim->has_dim_value() && !dim->has_dim_param()) {
-      dim->set_dim_param(symbolTable.createNew());
+      dim->set_dim_param(symbolTable.createNew("unk__"));
```
jcwchen, line 133: nit: "unk__" is the default value for createNew already?

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -18,37 +18,29 @@ namespace ONNX_NAMESPACE {
...
+    const TensorShapeProto& inferredShape,
+    const TensorShapeProto& expectedShape,
+    bool checkSameParam = false) {
+  EXPECT_TRUE(inferredShape.dim_size() == expectedShape.dim_size(), "Dim size for inferred and expected shape is different.");
+
+  for (int i = 0; i < inferredShape.dim_size(); i++) {
+    EXPECT_TRUE(
+        (inferredShape.dim(i).has_dim_value() && expectedShape.dim(i).has_dim_value()) ||
+            (inferredShape.dim(i).has_dim_param() && expectedShape.dim(i).has_dim_param()),
+        "Inferred and expected dim values are different.");
```
jcwchen, line 31: Should be `EXPECT_TRUE(
        (inferredShape.dim(i).has_dim_value() && expectedShape.dim(i).has_dim_value()) ||
            (inferredShape.dim(i).has_dim_param() && expectedShape.dim(i).has_dim_param())) <<
        "Inferred and expected dim values are different.";` ?

onnx/defs/shape_inference.h
```diff
@@ -88,6 +88,8 @@ struct InferenceContext {
   virtual GraphInferencer* getGraphAttributeInferencer(const std::string& attribute_name) = 0;
   virtual ~InferenceContext() {}
   virtual const SparseTensorProto* getInputSparseData(size_t index) const = 0;
+  // Gets the shape inputs computed by partial data propagation.
+  virtual const TensorShapeProto* getShapeInput(size_t index) const = 0;
```
jcwchen, line 92: `getShapeInput` here in InferenceContext is basically same as `TensorShapeProto* getInputData` in DataPropagationContext, right?

Actually I am thinking that whether to change `getInputData` into `getShapeInput` in DataPropagationContext to prevent confusion. Otherwise, `getInputData` functions in InferenceContext and DataPropagationContext are doing different things...

onnx/defs/tensor/defs.cc
```diff
@@ -257,23 +258,27 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            targetShape.insert(
-                targetShape.end(),
-                reinterpret_cast<const int64_t*>(bytes.c_str()),
-                reinterpret_cast<const int64_t*>(bytes.c_str() + bytes.size()));
-          } else {
-            const auto& data = targetShapeInitializer->int64_data();
-            targetShape.insert(targetShape.end(), data.begin(), data.end());
+
+          if (targetShapeInitializer) {
+            auto targetShape = ParseData<int64_t>(targetShapeInitializer);
```
jcwchen, line 274: Do we need to consider int32_t for ParseData?

onnx/defs/tensor/defs.cc
```diff
@@ -257,23 +258,27 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            for (auto val : targetShape) {
+              targetShapeProto.add_dim()->set_dim_value(val);
+            }
+          }
+
+          if (shapeInput) {
```
jcwchen, line 280: Perhaps else if for (shapeInput) ?

I thought targetShapeInitializer and shapeInput cannot be valid at the same time.

onnx/shape_inference/implementation.cc
askhade: It is for SymblTableImpl but not for for SymbolTable Interface... 
We should be using SymbolTable and not SymbolTableImpl in all function interfaces otherwise we are defeating the purpose of having SymbolTable and SymbolTableImpl....

onnx/test/cpp/data_propagation_test.cc
askhade: yes updated

onnx/defs/tensor/defs.cc
```diff
@@ -286,69 +291,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                // If allowzero is set however, do not propagate values, since output
+                // dimension is explicitly zero.
+                if (allowzero == 0) {
+                  unresolvedZeros[i] = true;
+                  if (dataInputTensorType.has_shape()) {
+                    if (i >= dataInputTensorType.shape().dim_size()) {
+                      fail_shape_inference("Invalid position of 0.");
+                    }
+                    if (dataInputTensorType.shape().dim(i).has_dim_value()) {
+                      const auto& dim_value =
```
jcwchen, line 335: nit: Redefinition for dim_value

onnx/defs/shape_inference.h
gramalingam: I would recommend ```getSymbolicInput``` to avoid confusion.

gramalingam: Logically, this subsumes ```getInputData```. Eg., in the long term, we could conceivably replace ```getInputData``` by this function (just like in DataPropagationContext), but that would require updating all current uses. If possible, I would like to avoid a potential confusion between ```Get input data encoded as a TensorShapeProto``` (which is what this is) and ```Get shape of input data``` (which this is not).

onnx/defs/tensor/defs.cc
```diff
@@ -286,69 +291,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-              if (negativeOneDim) {
-                fail_shape_inference(
-                    "Target shape may not have multiple -1 dimensions.");
+            if (targetShapeProto.dim(i).has_dim_param()) {
+              new_dim->set_dim_param(targetShapeProto.dim(i).dim_param());
+              // if there is a -1 we cannot use output product to infer the value
+              outputProductValid = false;
+            } else {
+              if (!targetShapeProto.dim(i).has_dim_value()) {
+                fail_shape_inference("Invalid target shape. Neither dim value nor dim param exists.");
```
gramalingam, line 309: A more general solution would be to treat this as an unknown dimension, instead of failing. I think that is preferable.

onnx/defs/tensor/defs.cc
```diff
@@ -286,69 +291,75 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             // Add a new dimension to outputShape
             auto* new_dim = outputShape->add_dim();
-            if (targetShape[i] == -1) {
-              // Check if multiple -1's. If not, set negativeOneDim, marking
-              // this dimension to potentially be filled in later.
-              if (negativeOneDim) {
-                fail_shape_inference(
-                    "Target shape may not have multiple -1 dimensions.");
+            if (targetShapeProto.dim(i).has_dim_param()) {
+              new_dim->set_dim_param(targetShapeProto.dim(i).dim_param());
```
gramalingam, line 304: There is a tricky problem here, in case the actual value at runtime happens to be either 0 or -1. In that case, propagating the dimension in the output-shape is problematic.

onnx/defs/tensor/defs.cc
askhade: Nope only int64 is allowed 
for shape input

From spec:
shape (non-differentiable) : tensor(int64)
Specified shape for output.

onnx/defs/tensor/defs.cc
```diff
@@ -286,69 +273,78 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              // There is a tricky edge case here. It is possible that the value of
+              // symbolic dim can be -1 or 0 at runtime. In that case simply propgating this
+              // symbol can be erroneous. This should be a very rare scenario and in such a
+              // case an option is to turn off data propagation during shape inference.
+              new_dim->set_dim_param(targetShapeProto.dim(i).dim_param());
+              outputProductValid = false;
+            } else {
+              if (!targetShapeProto.dim(i).has_dim_value()) {
+                // treat this dim as unknown dim
+                continue;
```
gramalingam, line 295: Also ```outputProductValid = false``` here?

gramalingam (approved):

PR #3593, opened 2021-07-19
Symbolic shape inference support-3: more ops for data propagation 

jcwchen:
**Description**
Support more ops for data propagation:
- [x] Add
- [x]  Sub
- [x]  Mul
- [x]  Gather
- [x]  Slice
- [x]  Concat

Also add tests for them.

**Motivation and Context**
https://github.com/onnx/onnx/pull/3551 ONNX has supported data propagation. Implement more data propagation functions for more ops


onnx/defs/data_propagators.h
```diff
@@ -16,4 +49,160 @@ inline void PropagateShapeDataFromInputToOutput(DataPropagationContext& ctx, int
...
+}
+
+inline void MathOpDataPropagator(DataPropagationContext& ctx, std::string op_type) {
+  const auto input_0 = ctx.getInputData(0);
+  const auto input_1 = ctx.getInputData(1);
+  if (input_0 == nullptr || input_1 == nullptr) {
+      return;
+  }
+  TensorShapeProto tsp;
+  if (input_0->dim_size() < input_1->dim_size()) {
```
askhade, line 93: are you supporting broadcasting? I suggest for simplicity we only support dim_size == 1... the dim size of shape will always be 1 .i.e is 1D tensor

onnx/defs/data_propagators.h
```diff
@@ -16,4 +49,160 @@ inline void PropagateShapeDataFromInputToOutput(DataPropagationContext& ctx, int
...
+    tsp.CopyFrom(input_shape);
+    ctx.addOutputData(0, std::move(tsp));
+  }
+}
+
+inline void SizeOpDataPropagator(DataPropagationContext& ctx) {
+  const auto input_data = ctx.getInputData(0);
+  if (input_data != nullptr) {
+    TensorShapeProto tsp;
+    tsp.mutable_dim()->Add()->set_dim_value(input_data->dim_size());
```
askhade, line 70: this is not correct ... size returns total number of elements... you should iterator over the dims and calculate the number of elements

onnx/defs/data_propagators.h
```diff
@@ -16,4 +49,160 @@ inline void PropagateShapeDataFromInputToOutput(DataPropagationContext& ctx, int
   }
 }
 
+// Data propagation function for Shape op
+// Propagates input shape to output shape
+inline void ShapeOpDataPropagator(DataPropagationContext& ctx) {
```
askhade, line 54: Rama updated this... merge from master?

onnx/defs/data_propagators.h
```diff
@@ -16,4 +49,160 @@ inline void PropagateShapeDataFromInputToOutput(DataPropagationContext& ctx, int
...
+      if (steps->dim_size() != 1) {
+        return;
+      }
+      step = steps->dim(0).dim_value();
+      if (step == 0) {
+        fail_shape_inference("Step cannot be 0 for Slice");
+      }
+    }
+    TensorShapeProto tsp;
+    for (int i = starts->dim(0).dim_value(); i < ends->dim(0).dim_value(); i += step) {
```
askhade, line 199: need to handle for negative axis... you can refer to Rama's implementation of data propagator for Shape-15

onnx/defs/data_propagators.h
jcwchen: After offline discussion, the minor broadcasting here should be valid. I have further considered failed broadcasting if the dim sizes are incompatible

onnx/defs/data_propagators.h
jcwchen: After offline discussion, the logic here should be good. I have updated Size unit test to prevent confusion.

onnx/defs/data_propagators.h
jcwchen: Thanks for catching this! I have reused the same logic from `TypeAndShapeInferenceFunction` as a `processSliceInputs` function to handle negative start or end. Also, I added a unit test for testing negative start and end.

onnx/defs/math/defs.cc
```diff
@@ -11,6 +11,68 @@
...
+    }
+  } else {
+    for (int i = 0; i < input_0->dim_size(); ++i) {
+      if (input_0->dim(i).has_dim_param() || input_1->dim(i).has_dim_param()) {
+        return;
+      }
+      tsp.mutable_dim()->Add()->set_dim_value(
+          MathOpTwoIntegers(op_type, input_0->dim(i).dim_value(), input_1->dim(i).dim_value()));
+    }
+  }
```
jcwchen, line 72: Trying to refactor L32-L72... Too much duplicate

onnx/defs/tensor/defs.cc
```diff
@@ -1026,6 +1052,49 @@ ONNX_OPERATOR_SET_SCHEMA(
                 ->mutable_dim((int)axis)
                 ->set_dim_value(temp);
           }
+        })
+        .PartialDataPropagationFunction([](DataPropagationContext& ctx) {
+          const auto input_data = ctx.getInputData(0);
+          const auto starts = ctx.getInputData(1);
+          const auto ends = ctx.getInputData(2);
+          const auto axes = ctx.getNumInputs() >= 4 ? ctx.getInputData(3) : nullptr;
```
gramalingam, line 1060: I think this could potentially confuse the situation where the input is specified (but data-propagation has no value for it) and the case where the input is not specified (and has a default value). May be better to use ```const bool axes_specified = ctx.getNumInputs() >= 4``` instead of  ```axes == nullptr``` down below.

onnx/defs/tensor/defs.cc
```diff
@@ -1026,6 +1052,49 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            if (steps != nullptr) {
+              if (steps->dim_size() != 1) {
+                return;
+              }
+              step = steps->dim(0).dim_value();
+            }
+
+            processSliceInputs(input_data->dim_size(), start, end, step);
+            TensorShapeProto tsp;
+            if (start < end) {
```
gramalingam, line 1085: nit: using ```step > 0``` here may be slightly better and preferable. (Similarly in shape inference method above.)

onnx/defs/tensor/defs.cc
```diff
@@ -1595,6 +1664,28 @@ ONNX_OPERATOR_SET_SCHEMA(
                                             : // i - axis < q
                     data_shape.dim(i - q + 1); // i < out_rank < q + r - 1
           }
+        })
+        .PartialDataPropagationFunction([](DataPropagationContext& ctx) {
+          if (!axisIsZero(ctx, true)) {
```
gramalingam, line 1669: Actually, I wonder if we should restrict ourselves to the case where input(0) is 1-dimensional tensor? If it is multidimensional, then it is more complicated.

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -197,5 +198,170 @@ agraph (int32[2,5] x) => (int32[1,2] w)
...
+agraph (int32[2,4,6] x, int32[5] y) => (int32[3] w)
+{
+    xs = Shape(x)
+    ys = Shape(y)
+    z = Sub(xs, ys)
+    w = Cast<to = 7>(z)
+}
+)ONNX";
+  TensorShapeProto expected_tsp;
+  expected_tsp.mutable_dim()->Add()->set_dim_value(-3);
```
askhade, line 245: -3 is not a valid value... This is an invalid model... can you change the input shapes so that that values are valid... 

invalid model test case will make sense if we have some logic to catch\handle such errors and we want to test that...

onnx/defs/tensor/defs.cc
```diff
@@ -1595,6 +1664,28 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          const auto input_data = ctx.getInputData(0);
+          const auto input_indices = ctx.getInputData(1);
+          if (input_data == nullptr || input_indices == nullptr) {
+            return;
+          }
+          TensorShapeProto tsp;
+          for (int i = 0; i < input_indices->dim_size(); ++i) {
+            if (input_indices->dim(i).has_dim_value()) {
+              int index = input_indices->dim(i).dim_value();
+              if (index < input_data->dim_size()) {
```
gramalingam, line 1681: I think that in the else case we should return (instead of creating an incorrect OutputData). (While we could produce an error/exception, it may be simpler to not bother about that for now.

gramalingam: The index value is allowed to be negative to count backwards.

onnx/defs/math/defs.cc
```diff
@@ -11,6 +11,68 @@
...
+    return a + b;
+  } else if (op_type == "Sub") {
+    return a - b;
+  } else if (op_type == "Mul") {
+    return a * b;
+  }
+  fail_shape_inference("Wrong op_type name for running propagation: ", op_type);
+}
+
+inline void MathOpDataPropagator(DataPropagationContext& ctx, std::string op_type) {
```
gramalingam, line 25: I think we can handle dim_params slightly better (e.g., if some are known and some are unknown). We should also handle the case where neither dim_param nor dim_value is set. I suggest something like below (also avoiding code duplication):
```cpp
inline void MathOpDataPropagator(DataPropagationContext& ctx, std::string op_type) {
  const auto input_0 = ctx.getInputData(0);
  const auto input_1 = ctx.getInputData(1);
  if (input_0 == nullptr || input_1 == nullptr) {
      return;
  }

  auto size0 = input_0->dim_size();
  auto size1 = input_1->dim_size();

  if ((size0 != size1) && (size0 != 1) && (size1 != 1) {
        fail_shape_inference("Invalid rank for ", op_type, " broadcasting: (",
            size0, ") vs (", size1, ").");
  }

  TensorShapeProto tsp;
  for (int i = 0; i < std::max(size0, size1); ++i) {
    auto& val0 = input_0->dim(size0 == 1 ? 0 : i);
    auto& val1 = input_1->dim(size1 == 1 ? 0 : i);
    auto* result = tsp.mutable_dim()->Add();
    if (val0.has_dim_value() && val1.has_dim_value())
       result->set_dim_value(MathOpTwoIntegers(op_type, val0.dim_value(), val1.dim_value()));
  }

  ctx.addOutputData(0, std::move(tsp));
}
```

jcwchen: Thank your for the suggestion! I just finished something like this.

onnx/defs/tensor/defs.cc
```diff
@@ -778,6 +803,36 @@ Example 2:
   ]
 )DOC";
 
+inline int64_t clamp(int64_t val, int64_t low, int64_t high) {
```
askhade, line 806: this can be converted in 1 line:
(val < low) ? low : (val > high) ? high : val;

also I suggest change low/high to min/max for better readability

another suggestion is you can simply use this 1 line instead of creating a inline function

onnx/defs/tensor/defs.cc
askhade: agree

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -193,5 +193,170 @@ agraph (int32[2,5] x) => (int32[1,2] w)
...
+  EXPECT_TRUE(CompareShape(propagated_tsp, expected_tsp));
+}
+
+TEST(DataPropagationImplTest, GatherTest) {
+  const char* code = R"ONNX(
+agraph (int32[1,2,3,4,5,6] x) => (int32[3] w)
+{
+    xs = Shape(x)
+    indices = Constant<value = int64[3] {0,3,5}>()
+    z = Gather<axis = 0>(xs, indices)
```
askhade, line 290: suggest adding a test with axis -ve

onnx/defs/math/defs.cc
jcwchen: Done.

onnx/defs/tensor/defs.cc
jcwchen: Good catch! Updated. Also, I have applied the same idea to steps.

onnx/test/cpp/data_propagation_test.cc
jcwchen: Did you mean negative axis for indices?

onnx/defs/tensor/defs.cc
jcwchen: Thanks for catching these 2 issues. Both updated

onnx/defs/tensor/defs.cc
jcwchen: We do have this restriction now for every data propagation, right? (only 0D or 1D tensor can be getInputData)
For other N-D tensors (which N > 1), their getInputData() will be nullptr and it will stop data propagation.

askhade (approved):

onnx/defs/math/defs.cc
gramalingam: It is okay with me. But if we try to add [1, 2, M] and [3, 4, 6], this will fail to compute anything, whereas we can actually compute [4, 6, ?] (where ? means neither dim_param nor dim_value set) easily enough (as indicated in my code sketch).

onnx/defs/math/defs.cc
```diff
@@ -39,11 +39,15 @@ inline void MathOpDataPropagator(DataPropagationContext& ctx, std::string op_typ
...
     auto& input_dim_0 = input_0->dim(size_0 == 1 ? 0 : i);
     auto& input_dim_1 = input_1->dim(size_1 == 1 ? 0 : i);
-    if (!(input_dim_0.has_dim_value() && input_dim_1.has_dim_value())) {
+    if (input_dim_0.has_dim_value() && input_dim_1.has_dim_value()) {
+      tsp.mutable_dim()->Add()->set_dim_value(
+          MathOpTwoIntegers(op_type, input_dim_0.dim_value(), input_dim_1.dim_value()));
+    } else if (input_dim_0.has_dim_param() || input_dim_1.has_dim_param()) {
+      tsp.mutable_dim()->Add()->set_dim_param("?");
+    } else {
+      // Dim is not set by value or param
```
gramalingam, line 48: Hi: I suggest adding a dim in all cases, and setting its value only if both have values, like below:
```cpp
auto* new_dim = tsp.mutable_dim()->Add();
if (input_dim_0.has_dim_value() && input_dim_1.has_dim_value()) {
   new_dim->set_dim_value(
          MathOpTwoIntegers(op_type, input_dim_0.dim_value(), input_dim_1.dim_value()));
}
```

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -26,8 +26,8 @@ inline bool CompareShape(
 
   for (int i = 0; i < inferredShape.dim_size(); i++) {
     EXPECT_TRUE(
-        (inferredShape.dim(i).has_dim_value() && expectedShape.dim(i).has_dim_value()) ||
-        (inferredShape.dim(i).has_dim_param() && expectedShape.dim(i).has_dim_param()))
+        (inferredShape.dim(i).has_dim_value() == expectedShape.dim(i).has_dim_value()) &&
+        (inferredShape.dim(i).has_dim_param() == expectedShape.dim(i).has_dim_param()))
```
jcwchen, line 30: Update here to test with "?" shape (which means no value and no param because it is not computable) -- if the dim does not have value or parameter, it is still a valid dim.

gramalingam (approved):

PR #3599, opened 2021-07-21
make symbol generation optional

askhade:

**Description**
Makes symbol generation optional.
- Describe your changes.

**Motivation and Context**


onnx/shape_inference/implementation.cc
```diff
@@ -658,20 +666,20 @@ void InferShapeForFunctionNode(
     const FunctionProto* func,
     const ISchemaRegistry* schema_registry,
     InferenceContext& ctx,
-    SymbolTable& symbolTable,
-    std::unordered_map<std::string, TensorShapeProto>& generatedShapeDataByName,
     const ShapeInferenceOptions& options) {
```
jcwchen, line 669: Why don't we keep symbolTable* as default nullptr for this function?

gramalingam (approved):

jcwchen (approved):

PR #3602, opened 2021-07-23
Update build instructions

snnn:
**Description**

Update build instructions.

1.
When you install protobuf on CentOS/RHEL/Fedora, you need to add:
"-DCMAKE_INSTALL_LIBDIR=lib64"

2. 
When you set PATH on Windows, do not add include/lib dirs in.
Instead of running
```
set PATH=<protobuf_install_dir>\bin;<protobuf_install_dir>\include;<protobuf_install_dir>\libs;%PATH%
```
You should:
```
set PATH=<protobuf_install_dir>\bin;%PATH%
```

3. Tell people how to set ONNX_USE_PROTOBUF_SHARED_LIBS  properly. Especially for Ubuntu Linux users who get protobuf from apt-get.

**Motivation and Context**


README.md
```diff
@@ -72,9 +72,52 @@ You can also use the [onnx-dev docker image](https://hub.docker.com/r/onnx/onnx-
...
+```bash
+export CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+```
+Windows:
+```bat
+set CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+```
+
+The ON/OFF depends on whether your protobuf was built as a shared library or static library, or both. If you have both, then we recommend you setting it to OFF for Windows and ON for Linux.
+
```
askhade, line 88: why ON for Linux?

askhade: what about Mac?

README.md
```diff
@@ -83,30 +126,52 @@ git submodule update --init --recursive
...
 git clone https://github.com/protocolbuffers/protobuf.git
 cd protobuf
 git checkout v3.16.0
-cd cmake
-cmake -G "Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX=<protobug_install_dir> -Dprotobuf_MSVC_STATIC_RUNTIME=ON -Dprotobuf_BUILD_SHARED_LIBS=OFF -Dprotobuf_BUILD_TESTS=OFF -Dprotobuf_BUILD_EXAMPLES=OFF .
-msbuild protobuf.sln /m /p:Configuration=Release
-msbuild INSTALL.vcxproj /p:Configuration=Release
+git submodule update --init --recursive
+mkdir build_source && cd build_source
+cmake ../cmake -Dprotobuf_BUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_POSITION_INDEPENDENT_CODE=ON -Dprotobuf_BUILD_TESTS=OFF -DCMAKE_BUILD_TYPE=Release
```
askhade, line 148: when -DONNX_USE_PROTOBUF_SHARED_LIBS=ON, -Dprotobuf_BUILD_SHARED_LIBS should also be ON

README.md
```diff
@@ -83,30 +126,52 @@ git submodule update --init --recursive
...
-* **Linux**
+CentOS/RHEL/Fedora:
 ```
 git clone https://github.com/protocolbuffers/protobuf.git
 cd protobuf
 git checkout v3.16.0
 mkdir build_source && cd build_source
-cmake ../cmake -Dprotobuf_BUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_POSITION_INDEPENDENT_CODE=ON -Dprotobuf_BUILD_TESTS=OFF -DCMAKE_BUILD_TYPE=Release
+cmake ../cmake  -DCMAKE_INSTALL_LIBDIR=lib64 -Dprotobuf_BUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_POSITION_INDEPENDENT_CODE=ON -Dprotobuf_BUILD_TESTS=OFF -DCMAKE_BUILD_TYPE=Release
```
askhade, line 160: when -DONNX_USE_PROTOBUF_SHARED_LIBS=ON, -Dprotobuf_BUILD_SHARED_LIBS should also be ON

README.md
snnn: Because most shared libraries are built without "-fPIC". And this sentence is for people have both static and shared libraries. In that case, pretty much the static one won't work.

README.md
snnn: See: https://github.com/onnx/onnx/blob/master/CMakeLists.txt#L51 

Now Protobuf_USE_STATIC_LIBS  is fully controlled by ONNX_USE_PROTOBUF_SHARED_LIBS. Users can't set it.

README.md
snnn: I don't know much about Mac.

snnn: Maybe I should change the wording to:
```
What if you have both:
Window: we recommend  you using the static one.
Linux: if your static library was built without "-fPIC"(the most common case), then it can't be used by onnx. If you don't know how it was built, let assume it doesn't have the flag.  Otherwise we recommend you using the static one.
```

snnn: Or just remove the line:
~If you have both, then we recommend you setting it to OFF for Windows and ON for Linux.~

README.md
```diff
@@ -72,9 +72,52 @@ You can also use the [onnx-dev docker image](https://hub.docker.com/r/onnx/onnx-
...
+set CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+```
+
+The ON/OFF depends on whether your protobuf was built as a shared library or static library, or both. If you have both, then we recommend you setting it to OFF for Windows and ON for Linux.
+
+ONNX_USE_PROTOBUF_SHARED_LIBS is default OFF. You don't need to run the commands above if you want to use a static protobuf library. 
+
+
+### Windows
+If you are building ONNX from source, it is recommended that you also build Protobuf locally as a static library. The version distributed with conda-forge is a DLL, but ONNX expects it to be a static library. Building protobuf locally also let's you control the verison of protobuf. The tested and recommended version is 3.16.0.
```
jcwchen, line 93: nit typo: also **let** you instead of also **let's** you

README.md
```diff
@@ -83,30 +126,52 @@ git submodule update --init --recursive
...
+### Linux
+
+First, you need to install protobuf. 
+
+Ubuntu users: the quickest way to install protobuf is to run
 
-### Building Protobuf from source
-* **Windows**
+```bash
+apt-get install python3-pip python3-dev libprotobuf-dev protobuf-compiler
```
jcwchen, line 136: Shall we add comments to reminder users that `libprotobuf-dev` and `protobuf-compiler` from apt-get would be quite old protobuf versions since we do recommend to use protobuf 3.16.0?

snnn: They are old but they should work without problems. I think we don't need to mention 3.16.0 specifically. Would you mind help me testing the command on Ubuntu 18.04? Supposedly it should work without problems.  You may need to add "cmake  git " to the apt-get install command line.

jcwchen: I tested this updated document (`apt-get install python3-pip python3-dev libprotobuf-dev protobuf-compiler`) with Ubuntu 18.04 and the installation went well. Thanks for the update!

README.md
askhade: lets*

README.md
askhade: what I meant is
On line 137 we are asking users to export CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"

Then the following instructions should match this setting.

README.md
askhade: if the shared lib is built without -fPIC meaning it is NOT position independent then this is a problem right... and in this case using static lib should be the right way

snnn: Shared lib won't be.

askhade: Meaning? Are you saying we can assume shared lib will be build with -fPIC becasue this is the default these days? then yes I agree. 
Now I understand the instructions better... earlier I did not quite understand what was the intention.

May be we can change the instructions to: 

What if you have both:
Window: we recommend  you using the static one.
Linux: When building from source we recommend using static library following the instructions in this doc. If using prebuilt libs, we recommend using shared lib.

snnn:
I'll update this PR shortly and address the comments.

jcwchen (approved):

README.md
```diff
@@ -72,9 +72,50 @@ You can also use the [onnx-dev docker image](https://hub.docker.com/r/onnx/onnx-
...
+Linux:
+```bash
+export CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+```
+Windows:
+```bat
+set CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+```
+
+The ON/OFF depends on what kind of protobuf library you have. Shared libraries are files ended with \*.dll/\*.so/\*.dylib. Static libraries are files ended with \*.a/\*.lib. This option depends on how you get your protobuf library and how it was built. And it is default OFF. You don't need to run the commands above if you'd prefer to use a static protobuf library.
```
askhade, line 87: nit: ending* with (there are 2 occurrences)

snnn: Thank you for point it out.

PR #3603, opened 2021-07-23
make_sequence_value_info alias for backwards compatibility

rajeevsrao:
Signed-off-by: Rajeev Rao <rajeevrao@nvidia.com>

**Description**
- `make_sequence_value_info` was renamed to `make_tensor_sequence_value_info` but partner packages still use the old API.

**Motivation and Context**
- Maintain backwards API compatibility for ONNX minor version update by creating an alias.


TomWildenhain-Microsoft (approved):

gramalingam (approved):

onnx/helper.py
```diff
@@ -905,3 +905,6 @@ def make_training_info(algorithm, algorithm_bindings, initialization, initializa
             binding.value = v
 
     return training_info
+
```
jcwchen, line 908: nit: need 2 lines as spaces between two functions to fix flake8 issue. You can also check it in your local

askhade (approved):

jcwchen:
Shall we add some warning message to let users know it is a deprecated function? Especially it only works for TensorProto

onnx/helper.py
```diff
@@ -905,3 +905,7 @@ def make_training_info(algorithm, algorithm_bindings, initialization, initializa
...
+
+# For backwards compatibility
+make_sequence_value_info = make_tensor_sequence_value_info
```
jcwchen, line 911:
```suggestion
def make_sequence_value_info(
        name,  # type: Text
        elem_type,  # type: int
        shape,  # type: Optional[Sequence[Union[Text, int, None]]]
        doc_string="",  # type: Text
        elem_shape_denotation=None,  # type: Optional[List[Text]]
):  # type: (...) -> ValueInfoProto
    """Makes a Sequence[Tensors] ValueInfoProto based on the data type and shape."""
    print("Warning: make_sequence_value_info has been deprecated since ONNX 1.10.0.",
        "Please now use make_tensor_sequence_value_info for TensorProto sequence. ")
    make_tensor_sequence_value_info(name, elem_type, shape, doc_string, elem_shape_denotation)
```

jcwchen (changes requested):

onnx/helper.py
TomWildenhain-Microsoft: Python warnings might be better: https://docs.python.org/dev/library/warnings.html#module-warnings

TomWildenhain-Microsoft: Deprecation module will do it all automatically: https://deprecation.readthedocs.io/en/latest/#api-documentation

rajeevsrao: @TomWildenhain-Microsoft would prefer not to introduce another package dependency (deprecation) and will use the python warnings instead.
@jcwchen I'll update the PR - should we mark this for removal in 1.11.0 or honor the semver guarantee and hold on to it until 2.0.0?

jcwchen: I personally vote for 1.11.0 since I even don't know when ONNX 2.0.0 will be and we don't want users to rely on old deprecated functions for a longer unexpected time.

rajeevsrao: Given the lack of consensus and clarity on the ONNX API deprecation schedule/policy I updated the warning to: 
"Deprecated in ONNX v1.10.0; `onnx.helper.make_sequence_value_info` alias will be removed in an upcoming release."

jcwchen (approved):

PR #3607, opened 2021-07-27
Extending If, Loop, and Identity ops to support optional type

neginraoof:
Extending If, Loop, and Identity ops to support optional type


docs/Changelog.md
```diff
@@ -19645,6 +19645,253 @@ This version of the operator has been available since version 15 of the default
...
+<dd>A boolean termination condition. Optional. Pass empty string to skip.</dd>
+<dt><tt>v_initial</tt> (variadic, heterogeneous) : V</dt>
+<dd>The initial values of any loop-carried dependencies (values that change across loop iterations)</dd>
+</dl>
+
+#### Outputs (1 - &#8734;)
+
+<dl>
+<dt><tt>v_final_and_scan_outputs</tt> (variadic, heterogeneous) : V</dt>
+<dd>Final N loop carried dependency values then K scan_outputs. Scan outputs must be Tensors.</dd>
```
hariharans29, line 19881: For the target models to be exported, can the constraint that scan outputs needs to be tensors still hold true ? Or should it possibly support optional(tensors) as well ? If it can support optional (tensors), the spec probably needs some update as to how 'None' is dealt with while concatenating the scan outputs (ignored, I assume).

onnx/defs/shape_inference.h
```diff
@@ -354,6 +354,56 @@ inline void propagateElemTypeFromSequenceInputToOutput(InferenceContext& ctx, si
...
+	    setTensorElementType(input_tensor_elem_type, output_value_case, *output_tensor_type);
+	  } else if (output_value_case == TypeProto::VALUE_NOT_SET) {
+	    // Assume output will have the same type
+	    setTensorElementType(input_tensor_elem_type, input_value_case, *output_tensor_type);
+	  } else {
+	    // This is not expected to happen
+	    fail_type_inference(
+	        "Optional tensor output ", outputIndex, " expected to have tensor or sparse tensor type. Got: ", output_value_case);
+	  }
+  } else if (input_opt_type.elem_type().has_sequence_type()) {
```
hariharans29, line 387: Should there be an 'else' that says propagating non-tensors and non-sequence optionals are unsupported ? (Otherwise the elem_type of the output is undefined/dangling isn't it ? )

onnx/defs/controlflow/defs.cc
```diff
@@ -289,6 +289,16 @@ void LoopInferenceFunction(InferenceContext& ctx) {
       if (seq_type.has_elem_type() && seq_type.elem_type().has_tensor_type()) {
         seq_type.mutable_elem_type()->mutable_tensor_type()->clear_shape();
       }
+    } else if (input_type.has_optional_type()) {
```
gramalingam, line 292: It may be helpful to extract lines 285 to 301 or so as a utility function ```ClearShape(TypeProto&)```. If the function is made recursive, it can handle arbitrary types a bit more easily. (Will help in the future if we expand the types even more).

onnx/defs/shape_inference.h
```diff
@@ -354,6 +354,56 @@ inline void propagateElemTypeFromSequenceInputToOutput(InferenceContext& ctx, si
   }
 }
 
+inline void propagateElemTypeFromOptionalInputToOutput(InferenceContext& ctx, size_t inputIndex, size_t outputIndex) {
```
gramalingam, line 357: I think this is more specialized/complex than it needs to be. I think it is better if the utility function works for arbitrary optional types. Further, the extra checks on the output-type is unnecessary. (I realize that this comes from the template used for other types, but it would be useful to start cleaning this up for the new code at least.) It should basically be something like "*outputType = inputType" (or as required by the protobuf generated classes to copy/clone).

gramalingam: I think this include file is becoming too large. It would be better to have just the interface/signature in the shape_inference.h file and move implementation to shape_inference.cc

askhade:
have you tested whether shape inference and checker works for these ops? A simple test can be done by using the existing test framework to generate a test model with optional outputs and run shape inference and checker on this model

onnx/defs/shape_inference.h
askhade: +1

onnx/defs/shape_inference.h
```diff
@@ -354,6 +354,56 @@ inline void propagateElemTypeFromSequenceInputToOutput(InferenceContext& ctx, si
...
+	    setTensorElementType(input_tensor_elem_type, input_value_case, *output_tensor_type);
+	  } else {
+	    // This is not expected to happen
+	    fail_type_inference(
+	        "Optional tensor output ", outputIndex, " expected to have tensor or sparse tensor type. Got: ", output_value_case);
+	  }
+  } else if (input_opt_type.elem_type().has_sequence_type()) {
+
+	  auto input_seq_type = input_opt_type.elem_type().sequence_type();
+	  if (input_seq_type.has_elem_type() && input_seq_type.elem_type().has_tensor_type()) {
```
askhade, line 390: this code only allows Seq[Tensors] is Seq[Seq[Tensor]] not valid?

docs/Changelog.md
neginraoof: Scan outputs are very special, and they can be handled with sequence type inputs instead. I don't think extending the scan output type is necessary.

onnx/defs/shape_inference.h
neginraoof: I refactored and deleted some part of the code. I think the code looks simpler now.

docs/Changelog.md
hariharans29: Agreed. I see that even when sequence type support was added for Loop, scan outputs were still just limited to tensors only. I think we can follow the same pattern for now.

onnx/defs/shape_inference.h
neginraoof: @askhade  I updated the optional type propagation for optional and sequence types. This should work for nested types as well now.

onnx/defs/controlflow/defs.cc
```diff
@@ -427,14 +435,16 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              t.insert(t.end(), o.begin(), o.end());
               return t;
             }(),
-            "All Tensor and Sequence types")
+            "All Tensor, Sequence, and optional types")
         .TypeConstraint("B", {"tensor(bool)"}, "Only bool")
         .TypeAndShapeInferenceFunction(IfInferenceFunction));
 
-static const char* Loop_ver13_doc = R"DOC(
+static const char* Loop_ver15_doc = R"DOC(
```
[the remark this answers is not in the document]
neginraoof: Thanks, yes this was missed in the updated.

gramalingam (approved):

neginraoof:
> have you tested whether shape inference and checker works for these ops? A simple test can be done by using the existing test framework to generate a test model with optional outputs and run shape inference and checker on this model

Yes, we have an exported model passing checker using these ops. Shared the example T5 with you and Hari.

PR #3611, opened 2021-07-28
Update cmake files

snnn:
**Description**

1. Delete add_msvc_runtime_flag. It is not needed anymore.  People can use [CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html)  variable instead. 
2. Don't disable VC++ warnings when ONNX_WERROR is not set. 
3. find_package Python has replaced PythonInterp and PythonLibs since cmake 3.12. We will use the newer one when it's available. 

**Motivation and Context**


askhade:
Current PR does not explicitly select the MSVC option MT\MD and based on the documentation (https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html) default is MD whereas for ONNX we need MT  as default.

snnn:
Both MD/MT work well in most of cases. If you want to make the default to MD, I think we can put such things in setup.py: detect if current os is windows. If it is true, then add "-DCMAKE_MSVC_RUNTIME_LIBRARY=xxx" to the build flags.

askhade:
Yes please add CMAKE_MSVC_RUNTIME_LIBRARY build flag to make sure the default settings dont change

PR #3625, opened 2021-08-03
Add RoiAlign opset 16

jiafatom:
Signed-off-by: David <jiafa@microsoft.com>

**Description**
- Update RoiAlign spec to opset 16: add an attribute `coordinate_transformation_mode`, change the default behavior to the correct one: 
-- torch roi_align aligned=True ---> onnx RoiAlign coordinate_transformation_mode='half_pixel'
-- torch roi_align aligned=False ---> onnx RoiAlign coordinate_transformation_mode='output_half_pixel'
- Bump onnx version to opset 16.
- In the test, I moved the original RoiAlign test into `export_roialign_aligned_false`, and add one test `export_roialign_aligned_true`. The test result aligns pytorch result.
- Added a conversion adapter.

**Motivation and Context**
- onnx RoiAlign need sync with torch roi_align. And torch roi_align has an input `aligned` now and used in several models. We need update onnx spec for it.
- It fixes the issue https://github.com/onnx/onnx/issues/3428
- There is also open discussion on the PR https://github.com/onnx/onnx/pull/3468, and this PR resolves it.


jiafatom:
@askhade @wschin @gramalingam Could you please take a look? Thanks
cc @fdwr

fdwr:
@jiafatom, this is great - thanks.

Regarding the coordinate transformation mode, we had [discussed](https://github.com/onnx/onnx/pull/3429#issuecomment-825076299) using the same attribute name as `Resize` for consistency `coordinate_transformation_mode`, but only supporting two modes (proper `half_pixel` and the old output-only `output_half_pixel`). Additionally, for anyone opting into the newer opset, they would get the correct behavior *by default*, just like later versions of `Resize` default to the correct `half_pixel` behavior by default. I'd like to see that here.

jiafatom:
> @jiafatom: This is great - thanks.
> 
> Regarding the coordinate transformation mode, we had [discussed](https://github.com/onnx/onnx/pull/3429#issuecomment-825076299) using the same attribute name as `Resize` for consistency `coordinate_transformation_mode`, but only supporting two modes (proper `half_pixel` and the old output-only `output_half_pixel`). We should not just copy attributes and names from other API's without holistically considering existing ONNX precedent.
> 
> Additionally, for anyone opting into the newer opset, they would get the correct behavior _by default_, just like `Resize` defaulted to the correct `half_pixel` behavior in later opsets (convertors could still convert to the old mode using the old enum value). I'd like to see that here.

Thanks for your reply. I feel that this current PR is enough for `RoiAlign`.
The onnx `RoiAlign` is in sync with torchvision `roi_align`. There is no `coordinate_transformation_mode` in torch `roi_align`, so we should not add that.
`Resize` (with `CropAndResize`) are different. Although `tf.image.crop_and_resize` (onnx `Resize`) does something similar as `RoiAlign`, but the algorithm is completely different, so it is better not to unify them. Thanks!

jcwchen (commented):
https://github.com/onnx/onnx/blob/391847ff287fd1a494d94783771c04cabd347315/onnx/version_converter/convert.h#L756 Please create a section for 15->16 and add an adapter here. CompatibleAdapter should be sufficient.

onnx/version_converter/convert.h
```diff
@@ -754,6 +754,10 @@ class DefaultVersionConverter : public BaseVersionConverter {
           "Pow", OpSetID(14), OpSetID(15)));
       registerAdapter(make_unique<CompatibleAdapter>(
           "Shape", OpSetID(14), OpSetID(15)));
+
+      /******** 15 -> 16 ********/
+      registerAdapter(make_unique<CompatibleAdapter>(
```
gramalingam, line 759: From what I understand in the discussion, this is not a compatible extension. Don't we have to use a non-default value for the new attribute to get the old behavior? If so, the adapter should add that attribute to the node.

jcwchen: Sorry that I don't have the context here... So previously RoiAlign-10's behaviour is same as using output_half_pixel. instead of half_pixel (default for RoiAlign-16). Then yes please add a new adapter for it. Thanks @gramalingam for catching this!

onnx/defs/object_detection/defs.cc
```diff
@@ -60,6 +60,13 @@ ONNX_OPERATOR_SET_SCHEMA(
             "Default is 'avg'.",
             AttributeProto::STRING,
             std::string("avg"))
+        .Attr(
+            "coordinate_transformation_mode",
+            "The coordinate transformation mode with the value 'half_pixel' or 'output_half_pixel'. "
```
gramalingam, line 65: How about rewording it as below:
```
The coordinate transformation mode. Allowed values are 'half_pixel' and 'output_half_pixel'.
Use the value 'half_pixel' to pixel shift the input coordinates by -0.5 (the recommended behavior).
Use the value 'output_half_pixel' to omit the pixel shift for the input (use this for a
backward-compatible behavior).
```

onnx/version_converter/convert.h
jiafatom: Added an adapter `RoiAlign_15_16`.

onnx/version_converter/adapters/roialign_15_16.h
```diff
@@ -0,0 +1,27 @@
...
+// Adapter for RoiAlign in default domain from version 15 to 16
+
+#pragma once
+
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE { namespace version_conversion {
+
+struct RoiAlign_15_16 final: public Adapter {
+  explicit RoiAlign_15_16() : Adapter("Upsample", OpSetID(15), OpSetID(16)) {}
```
gramalingam, line 14: "Upsample" looks like a copy-paste error. Is it supposed to be "RoiAlign"?

onnx/defs/object_detection/defs.cc
jiafatom: done

jiafatom:
@postrational Could you please review this PR as well? Thanks!

onnx/defs/object_detection/old.cc
```diff
@@ -8,6 +8,126 @@ using namespace ONNX_NAMESPACE;
 
 namespace ONNX_NAMESPACE {
 
+static const char* RoiAlign_ver1_doc = R"DOC(
```
askhade, line 11: nit: ver10

askhade:
LGTM

jcwchen:
nit suggestion: Since RoiAlign-16 default behavior has changed from RoiAlign-10, it would be helpful to avoid confusion if there is also a note about the behavior difference in the old document for the old op (RoiAlign-10).

jiafatom:
> nit suggestion: Since RoiAlign-16 default behavior has changed from RoiAlign-10, it would be helpful to avoid confusion if there is also a note about the behavior difference in the old document for the old op (RoiAlign-10).

synced offline, issue addressed.

jiafatom:
Did the version conversion test locally. Just have a model (opset 15) with RoiAlign. When I run the model conversion (target opset 16), it outputs the model with attribute `coordinate_transfomation_mode="output_half_pixel"`. So it works.

gramalingam (approved):

fdwr (approved):
Thanks again David. After yours goes in, I may resurrect one of my backend cases (the one with whole integers and perfectly aligned slice) and add a zero-size ROI too.

jcwchen (approved):

askhade (approved):

Issue #2813, opened 2020-06-04
SIGBUS due to memory alignment issues in TypeAndShapeInferenceFunction for Slice and ConstantOfShape operators


muxyrmuc:
There is an issue with some models on ARM devices related to unaligned memory access.

I found the function which causes the problem:
https://github.com/onnx/onnx/blob/master/onnx/defs/generator/defs.cc

Function:
```
TypeAndShapeInferenceFunction() for ConstantOfShape
```

Lines:
```
std::vector<int64_t> targetShape;
if (targetShapeInitializer->has_raw_data()) {
  const std::string& bytes = targetShapeInitializer->raw_data();
  targetShape.insert(
    targetShape.end(),
    reinterpret_cast<const int64_t*>(bytes.c_str()),
    reinterpret_cast<const int64_t*>(bytes.c_str() + bytes.size()));
  }
```
We receive SIGBUS in targetShape.insert() because bytes.c_str() pointer must be aligned to int64_t size to safe reinterpret_cast<int64_t*> conversion.

Also I found similar issue in TypeAndShapeInferenceFunction() for Slice operator (checked on different model file) which uses ParseData() from https://github.com/onnx/onnx/blob/master/onnx/defs/tensor_proto_util.cc:
```
res.insert(                                                            \
      res.end(),                                                         \
      reinterpret_cast<const type*>(bytes),                              \
      reinterpret_cast<const type*>(bytes + raw_data.size()));           \
```
These lines also cause SIGBUS due to unaligned access.

Related issue in ONNX Runtime repo: https://github.com/microsoft/onnxruntime/issues/4103

Could you take a look?

PR #3626, opened 2021-08-03
Fix arm crash due to unaligned raw_data buffer of initializer tensors

gwang-msft:
**Description**
- Fix arm crash due to unaligned raw_data buffer of initializer tensors

**Motivation and Context**
- This issue was first reported #2813, and was reported by internal users of onnxruntime
- For now this issue only repros on arm32 but not arm64, this does not repro on x86 or x86_64
- The initializer tensors for slice are usually tensors with a single int64_t element, if the data is store in the raw_data, which is represented as a std::string, where the internal data (`std::string::c_str()`) is aligned as char*, so the reinterpret_cast<const int64_t*> of the raw_data will crash due to misalignment, (`Fatal signal 7 (SIGBUS), code 1 (BUS_ADRALN)`)
- The fix is to copy the raw_data by byte instead of copying by the destination type (`int64_t` in this crash)

### DIFF
--- a/onnx/defs/generator/defs.cc
+++ b/onnx/defs/generator/defs.cc
@@ -267,16 +267,8 @@ ONNX_OPERATOR_SET_SCHEMA(
           // In this case, we extract the shape values from input tensor
           // and create output tensor of that shape.
           // First, extract target shape value.
-          std::vector<int64_t> targetShape;
-          if (targetShapeInitializer->has_raw_data()) {
-            const std::string& bytes = targetShapeInitializer->raw_data();
-            const size_t raw_data_size = bytes.size();
-            targetShape.resize(raw_data_size / sizeof(int64_t));
-            memcpy(reinterpret_cast<char*>(targetShape.data()), bytes.c_str(), raw_data_size);
-          } else {
-            const auto& data = targetShapeInitializer->int64_data();
-            targetShape.insert(targetShape.end(), data.begin(), data.end());
-          }
+          std::vector<int64_t> targetShape = ParseData<int64_t>(targetShapeInitializer);
+
           // Next, set output shape to the target shape.
           auto final_output_shape =
               ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
--- a/onnx/defs/generator/defs.cc
+++ b/onnx/defs/generator/defs.cc
@@ -270,10 +270,9 @@ ONNX_OPERATOR_SET_SCHEMA(
           std::vector<int64_t> targetShape;
           if (targetShapeInitializer->has_raw_data()) {
             const std::string& bytes = targetShapeInitializer->raw_data();
-            targetShape.insert(
-                targetShape.end(),
-                reinterpret_cast<const int64_t*>(bytes.c_str()),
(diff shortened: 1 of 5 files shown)


onnx/defs/generator/defs.cc
```diff
@@ -270,10 +270,9 @@ ONNX_OPERATOR_SET_SCHEMA(
           std::vector<int64_t> targetShape;
           if (targetShapeInitializer->has_raw_data()) {
             const std::string& bytes = targetShapeInitializer->raw_data();
-            targetShape.insert(
-                targetShape.end(),
-                reinterpret_cast<const int64_t*>(bytes.c_str()),
-                reinterpret_cast<const int64_t*>(bytes.c_str() + bytes.size()));
+            const size_t raw_data_size = bytes.size();
```
askhade, line 273: can you use ParseData here?

onnx/defs/tensor/old.cc
```diff
@@ -142,10 +142,9 @@ ONNX_OPERATOR_SET_SCHEMA(
           std::vector<int64_t> targetShape;
           if (targetShapeInitializer->has_raw_data()) {
             const std::string& bytes = targetShapeInitializer->raw_data();
-            targetShape.insert(
-                targetShape.end(),
-                reinterpret_cast<const int64_t*>(bytes.c_str()),
-                reinterpret_cast<const int64_t*>(bytes.c_str() + bytes.size()));
+            const size_t raw_data_size = bytes.size();
```
askhade, line 145: same as above if you use ParseData it will replace the entire if else here

askhade:
Looks like you did update the Upsample adapters. Remove the comment "The code here are not updated, because it is not common to convert Opset version on an arm32 device, and so far the crash is only in int64_t but not float" from the description ?

gwang-msft:
> Looks like you did update the Upsample adapters. Remove the comment "The code here are not updated, because it is not common to convert Opset version on an arm32 device, and so far the crash is only in int64_t but not float" from the description ?

These was pushed accidentally, but changed to use ParseData, so these are changed as well

askhade (approved):

onnx/defs/tensor_proto_util.cc
```diff
@@ -60,10 +60,9 @@ namespace ONNX_NAMESPACE {
         }                                                                  \
       }                                                                    \
     }                                                                      \
-    res.insert(                                                            \
-        res.end(),                                                         \
-        reinterpret_cast<const type*>(bytes),                              \
-        reinterpret_cast<const type*>(bytes + raw_data.size()));           \
+    const size_t raw_data_size = raw_data.size();                          \
```
skottmckay, line 63: Would be good to capture why we do this in a comment so it isn't accidentally removed in the future.

snnn: @gwang-msft
The culprit is reinterpret_cast, which is a warn sign. 
Your change is good, but the reinterpret_cast is unnecessary.

gwang-msft: Yes, the reinterpret_cast here is redundant, can remove in a future change

Issue #3598, opened 2021-07-21
Error when checking model with external data


yjydfnhc:
# Bug Report

### Describe the bug
I saved a simple model with save_as_external_data=True, then load the model into memory with load_external_data=False, will get an error when using checker.check_model() for this model. 

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): win10
- ONNX version (*e.g. 1.7*):  1.9.0
- Python version: 3.8.0

### Reproduction instructions
- Describe the code to reproduce the behavior.
```
import numpy as np
import onnx
from onnx import helper
from onnx import AttributeProto, TensorProto, GraphProto
import json
from google.protobuf.json_format import MessageToJson

#======== 1. create an onnx model
dtype = np.float32
x = np.random.randn(3, 5).astype(dtype)

data_w = np.random.randn(2, 5).astype(dtype)
w = helper.make_tensor(name = 'w', data_type = TensorProto.FLOAT, dims = data_w.shape, vals = data_w.flatten().astype(dtype).tobytes(), raw=True)

Z = helper.make_tensor_value_info('Z', TensorProto.FLOAT, [2, 3])

# Create node (NodeProto)
X = helper.make_node(
    'Constant',
    inputs=[],
    outputs=['X'],
    value=onnx.helper.make_tensor(
        name='const_tensor_x',
        data_type=onnx.TensorProto.FLOAT,
        dims=x.shape,
        vals=x.flatten().astype(float),
    ),
)
node_gemm = helper.make_node(
    'Gemm',                  # name
    ['w', 'X'], # inputs
    ['Z'],                  # outputs
    name = 'gemm1'
)
# Create the graph (GraphProto)
graph_def = helper.make_graph(
    [X, node_gemm],        # nodes
    'test-model',      # name
    [],  # inputs
    [Z],  # outputs
    initializer = [w],
)
model_def = helper.make_model(graph_def, producer_name='onnx-example')

#======== 2. save an onnx model
model_path = '../outs/model.onnx'
# save model with separate data files
onnx.save_model(model_def, model_path
                , save_as_external_data=True
                , all_tensors_to_one_file=False
                , size_threshold=0)

#======== 4. load and try save
# (1) load with data, then check: will success
model_full = onnx.load(model_path)
onnx.checker.check_model(model_full)

# (2) load without data, then check: will fail
model_sep = onnx.load(model_path, load_external_data=False)
[...]

Issue #3596, opened 2021-07-20
Error when saving model with external data


yjydfnhc:
# Bug Report

### Describe the bug
For an in-memory onnx model loaded with load_external_data=False, when trying to save it with onnx.save(model, path), it will raise error because set_external_data() and save_external_data() will always check the existence of "raw_data" field.

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): win10
- ONNX version (*e.g. 1.7*):  1.9.0
- Python version: 3.8.0

### Reproduction instructions
- Example code:
```
import numpy as np
import onnx
from onnx import helper
from onnx import AttributeProto, TensorProto, GraphProto
import json
from google.protobuf.json_format import MessageToJson

#======== 1. create an onnx model

dtype = np.float32
x = np.random.randn(3, 5).astype(dtype)

data_w = np.random.randn(2, 5).astype(dtype)
w = helper.make_tensor(name = 'w', data_type = TensorProto.FLOAT, dims = data_w.shape, vals = data_w.flatten().astype(dtype).tobytes(), raw=True)

Z = helper.make_tensor_value_info('Z', TensorProto.FLOAT, [2, 3])

# Create node (NodeProto)
X = helper.make_node(
    'Constant',
    inputs=[],
    outputs=['X'],
    value=onnx.helper.make_tensor(
        name='const_tensor_x',
        data_type=onnx.TensorProto.FLOAT,
        dims=x.shape,
        vals=x.flatten().astype(float),
    ),
)
node_gemm = helper.make_node(
    'Gemm',                  # name
    ['w', 'X'], # inputs
    ['Z'],                  # outputs
    name = 'gemm1'
)
# Create the graph (GraphProto)
graph_def = helper.make_graph(
    [X, node_gemm],        # nodes
    'test-model',      # name
    [],  # inputs
    [Z],  # outputs
    initializer = [w],
)
model_def = helper.make_model(graph_def, producer_name='onnx-example')

# #======== 2. save an onnx model
model_path = '../outs/model.onnx'
# save model with separate data files
onnx.save_model(model_def, model_path
                , save_as_external_data=True
                , all_tensors_to_one_file=False
                , size_threshold=0)

#========= 3. for observe: convert model to json to find rawData
# (1) for model with data in one file
model_full = onnx.load(model_path)
onnx_json = json.loads(MessageToJson(model_full))
output_json = '../outs/model_full.json'
with open(output_json, 'w') as file:
[...]

Issue #3594, opened 2021-07-20
Error in numpy_helper.to_array(tensor) for tensor with external data


yjydfnhc:
# Bug Report

### Describe the bug
numpy_helper.to_array(tensor) will get error for tensor whose data is saved in an external file. Current implement of to_array() only tried to get data from 'raw_data' field or storage_field(e.g. 'float32_data' field), but never tried to read data from 'external_data' field.  When data of the tensor haven't been loaded into memory and are still in 'external_data', this function will raise an error since it gets empty data from the wrong storage field. 

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*):  win10
- ONNX version (*e.g. 1.7*):  1.9.0
- Python version: 3.8.0

### Reproduction instructions
- Example code:
```
import numpy as np
import onnx
from onnx import  helper, TensorProto, numpy_helper

#======== 1. create an onnx model
dtype = np.float32
x = np.ones((3, 3), dtype)

data_w = np.ones((3, 3), dtype)
w = helper.make_tensor(name = 'w', data_type = TensorProto.FLOAT, dims = data_w.shape, vals = data_w.flatten().astype(dtype).tobytes(), raw=True)

V = helper.make_tensor_value_info('V', TensorProto.FLOAT, [3, 3])
Z = helper.make_tensor_value_info('Z', TensorProto.FLOAT, [3, 3])

# Create node (NodeProto)
X = helper.make_node(
    'Constant',
    inputs=[],
    outputs=['X'],
    value=onnx.helper.make_tensor(
        name='const_tensor_x',
        data_type=onnx.TensorProto.FLOAT,
        dims=x.shape,
        vals=x.flatten().astype(float),
    ),
)
node_cf = helper.make_node(
    'Gemm',
    ['w', 'X'], # inputs
    ['Z1'],                  # outputs
    name = 'tocf'
)
node_sum = helper.make_node(
    'Sum',                  # name
    ['Z1', 'V'], # inputs
    ['Z'],                  # outputs
)
# Create the graph (GraphProto)
graph_def = helper.make_graph(
    [X, node_cf, node_sum],        # nodes
    'test-model',      # name
    [V],  # inputs
    [Z],  # outputs
    initializer = [w],
)
model_def = helper.make_model(graph_def, producer_name='onnx-example')

# #======== 2. save the onnx model: save model with separate data files
path_sep = '../outs/model_sep.onnx'
onnx.save_model(model_def, path_sep
[...]

Issue #3322, opened 2021-03-12
Improve loading ONNX model with external data


jcwchen:
# Feature Request

### System information
ONNX version (you are using): 1.8.1

### What is the problem that this feature solves?
Currently onnx.load will stop loading external data if raw_data exists. Also, it will load whole data from external tensor is the length is not specified. With them, onnx.load and onnx.save the ONNX model will pollute it.

### Describe the alternatives you have considered
1. Perhaps if raw_data exists but its `data_location` is TensorProto.EXTERNAL, `load_external_data_for_tensor` should ignore current raw_data and load it from external tensor anyway.
https://github.com/onnx/onnx/blob/4f8ebec58c9ef1bc9aaf295847260f4f44a02e58/onnx/external_data_helper.py#L43
2. Even though the size of external tensor is not provided by the tensor, ONNX should calculate a reasonable size to read instead of reading all of them for safety.
https://github.com/onnx/onnx/blob/4f8ebec58c9ef1bc9aaf295847260f4f44a02e58/onnx/external_data_helper.py#L54
3. For save_external_data, more consider a situation that tensor's raw_data and data from external data exist at the same time. It should throw some warnings about overlapping existing external data.
https://github.com/onnx/onnx/blob/4f8ebec58c9ef1bc9aaf295847260f4f44a02e58/onnx/external_data_helper.py#L146

### Describe the feature
Why is this feature necessary? What does it accomplish?
https://github.com/onnx/onnx/blob/master/docs/ExternalData.md ONNX does not have strict rule about external tensor. It's better to improve ONNX's functionality to handle more conditions to prevent future issues.

### Will this influence the current api?
Yes. It will improve current onnx.load and onnx.save with external data

### Feature Area
external data

jcwchen:
@annajung since you are working with external data, feel free to pick this up if you are interested. Otherwise I will work with this. Thank you!

annajung:
Thanks for the ping @jcwchen! I'll be happy to pick this up and work on this :)

annajung:
@jcwchen Totally my bad for dropping the ball on this one! I started working on it but looks like you already got it covered.

jcwchen:
Hey @annajung No worries! I should let you know first before working on it... Since my PR regarding to external data is quite a big one so I think it's fine to decouple this work to you. Do you prefer to submit your own PR for it or review my existing PR? I am open to both option. Thanks!

annajung:
[...]

PR #3630, opened 2021-08-03
Improve external tensors handling in ONNX functions/helpers

jcwchen:
**Description**

- [x] ~(1) Make ParseData consider external tensor~
- [x] (2) onnx.save: skip loading external data; only save those have raw_data
- [x] (3) onnx.numpy_helper.to_array: Extend to_array with an optional parameter (base_dir) to access external tensor
- [x] (4) Improve external_data_helper.py: https://github.com/onnx/onnx/issues/3322
- [x] ~(5) Shape inference related: old reshape, ConstantOfShape~
- Add tests for above items
     - [x]  (1)
     - [x]  (2)
     - [x]  (3)
     - [x]  (4)
     - [x]  (5)

**Motivation and Context**
https://github.com/onnx/onnx/issues/3594, https://github.com/onnx/onnx/issues/3596, https://github.com/onnx/onnx/issues/3598. There are issues related to ONNX model without loading external tensors.

ONNX has the ability to load a ONNX model without loading external tensors. Initially this feature is designed to be used with load_external_data_for_model (load external data from somewhere else). So simply using onnx.load('model.onnx', load_external_data=False) will bump into many issues because most functions/helpers assume data exist in the tensor's raw_data. At this moment, these functions/helpers do not support models without load_external_data. 

However, we saw there are more and more large ONNX models have larger tensors. Loading all tensors from external data will easily reach 2GB Protobuf limitation. Besides, actually main ONNX functions like checker or shape inference do not need to know what the real raw_data is. Those external data should be loaded on demand. Therefore, we do want these ONNX functions to have this ability to handle models without loading external data.

(Updated) Nevertheless, actually shape_inference or version converter do not really need external tensors because in most situations, they just use "Shape" data through ParseData, which is always not a big file. Ideally users use onnx.save with proper threshold and those shape data shouldn't be external tensors... Therefore, ParseData/shape_inference won't consider external tensors. Instead, it should throw a warning to let users know these shape data should not be external tensors... It will be done in another PR


onnx/test/test_external_data.py
```diff
@@ -342,49 +342,86 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
+                        location=None,
+                        size_threshold=0,
+                        convert_attribute=False)
+        # Load the saved model again; Only works if the saved path is under the same directory
+        model = onnx.load_model(model_file_path)
 
-# The following test will fail in some platforms
-# because >2GB proto python object is not allowed
-# Disable it for now and it should be fixed after 1.8 Release
-"""
```
jcwchen, line 349: Remove this test for good because in some environments loading model > 2GB will crash and it cannot be tested.

onnx/external_data_helper.py
```diff
@@ -257,9 +262,18 @@ def write_external_data_tensors(model, filepath):  # type: (ModelProto, Text) ->
...
-            save_external_data(tensor, filepath)
-            tensor.ClearField(str('raw_data'))
+            if tensor.HasField("raw_data"):
+                save_external_data(tensor, filepath)
+                tensor.ClearField(str('raw_data'))
+            else:
+                # If the model did not load external data, it's possible there is no raw_data
+                any_not_save_external_tensor = True
+
+    if any_not_save_external_tensor:
```
askhade, line 275: not sure if this is needed

onnx/numpy_helper.py
```diff
@@ -18,11 +19,12 @@ def combine_pairs_to_complex(fa):  # type: (Sequence[int]) -> Sequence[np.comple
     return [complex(fa[i * 2], fa[i * 2 + 1]) for i in range(len(fa) // 2)]
 
 
-def to_array(tensor):  # type: (TensorProto) -> np.ndarray[Any]
+def to_array(tensor, base_dir=""):  # type: (TensorProto, Text) -> np.ndarray[Any]
```
askhade, line 22: adding base_dir is complicating all the APIs... specially ParseData... I wonder if we should just try and read from current dir

onnx/external_data_helper.py
```diff
@@ -41,8 +41,13 @@ def load_external_data_for_tensor(tensor, base_dir):  # type: (TensorProto, Text
     tensor: a TensorProto object.
     base_dir: directory that contains the external data.
     """
-    if tensor.HasField("raw_data"):  # already loaded
-        return
+    # If it has already loaded, override the raw_data from external tensor
+    if tensor.HasField("raw_data"):
+        # Empty raw_data for the following load
+        print("Warning: raw_data and external data exist simultaneously."
```
askhade, line 47: not sure why we need this warning? This is confusing... we should never have both

askhade: Why do we land in this situation? Maybe we are not clearing the raw_data field after serializing the raw data to external data... we should fix that

onnx/defs/tensor/defs.cc
```diff
@@ -2473,7 +2473,7 @@ ONNX_OPERATOR_SET_SCHEMA(
 
 ONNX_OPERATOR_SET_SCHEMA(
     Identity,
-    14,
+    16,
```
askhade, line 2476: why is this change part of this PR? bad merge?

onnx/defs/tensor_proto_util.cc
```diff
@@ -30,45 +33,49 @@ namespace ONNX_NAMESPACE {
...
-    /* misalignment issues on certain platforms, such as arm32-v7a */      \
-    const size_t raw_data_size = raw_data.size();                          \
-    res.resize(raw_data_size / sizeof(type));                              \
-    memcpy(reinterpret_cast<char*>(res.data()), bytes, raw_data_size);     \
-    return res;                                                            \
+#define DEFINE_PARSE_DATA(type, typed_data_fetch)                              \
+  template <>                                                                  \
+  const std::vector<type> ParseData(const TensorProto* tensor_proto) {         \
+    std::vector<type> res;                                                     \
+    if (!tensor_proto->has_raw_data() && tensor_proto->has_data_location() &&  \
```
askhade, line 40: I suggest you check in the following order
If (ExternalData) { raise error}
Else if (RawData) {return raw data}
Else {return typed data}

External data should be checked based on data type != undefined and data location == external
Raw data should be checked using data type != undefined and has_raw_data() == true

onnx/defs/tensor_proto_util.cc
```diff
@@ -30,45 +33,49 @@ namespace ONNX_NAMESPACE {
...
+      tensor_proto->data_location() == TensorProto_DataLocation_EXTERNAL) {    \
+      fail_shape_inference("Cannot parse data from external tensors. Please ", \
+        "load external data into raw data for tensor: ", tensor_proto->name());\
+    } else if (!tensor_proto->has_raw_data()) {                                \
+      const auto& data = tensor_proto->typed_data_fetch();                     \
+      res.insert(res.end(), data.begin(), data.end());                         \
```
askhade, line 46: some error checks here would be good... you can calculate expected size based on dims and then check it matches the <type>_data_size

onnx/external_data_helper.py
```diff
@@ -36,13 +36,11 @@ def __init__(self, tensor):  # type: (TensorProto) -> None
 def load_external_data_for_tensor(tensor, base_dir):  # type: (TensorProto, Text) -> None
     """
     Load data from an external file for tensor.
-
+    Even if the tensor has raw_data, it will load from external data and override existing raw_data anyway.
```
askhade, line 39: May be change to something like: Ideally TensorProto should not hold any raw data but if it does it will be ignored.

askhade:
I think we should convert "set_external_data" to private method

onnx/external_data_helper.py
```diff
@@ -258,8 +256,9 @@ def write_external_data_tensors(model, filepath):  # type: (ModelProto, Text) ->
     The modified model object.
     """
     for tensor in _get_all_tensors(model):
-        if uses_external_data(tensor):
+        if uses_external_data(tensor) and tensor.HasField("raw_data"):
```
askhade, line 259: add some comments explaining why you are checking for uses_external_data. It is not clear

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,124 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
+        # Save without load_external_data
+        model = onnx.load_model(model_file_path, load_external_data=False)
+        onnx.save_model(self.model,
+                        model_file_path,
+                        save_as_external_data=True,
+                        location=None,
+                        size_threshold=0,
+                        convert_attribute=False)
+        # Load the saved model again; Only works if the saved path is under the same directory
+        model = onnx.load_model(model_file_path)
```
askhade, line 362: where will the second save, save the data? In a different file or same as original?

onnx/defs/tensor_proto_util.cc
jcwchen: Good point. Added. IIUC, it's possible that the tensor's dims is empty, right? If so, the check will simply skip here: `tensor_proto->dims_size() != 0`

onnx/test/test_external_data.py
jcwchen: I think it's my typo. I corrected it to use model for second save. Thanks

onnx/test/cpp/data_propagation_test.cc
```diff
@@ -324,7 +324,7 @@ TEST(DataPropagationImplTest, GatherNegativeIndicesTest) {
 agraph (int32[1,2,3,4,5,6] x) => (int32[2] w)
 {
     xs = Shape(x)
-    indices = Constant<value = int64[3] {-2,-1}>()
+    indices = Constant<value = int64[2] {-2,-1}>()
```
jcwchen, line 327: The enhanced size check caught this error written by me

onnx/external_data_helper.py
jcwchen: After offline discussion, remove this warning

onnx/external_data_helper.py
jcwchen: After offline discussion, remove this warning because the override behavior is obvious when using load_external_data_for_tensor. Added a comment in the function instead

onnx/defs/tensor/defs.cc
jcwchen: Yes.. reverted

onnx/defs/tensor_proto_util.cc
jcwchen: I put checking defined data type in the first condition to simplify the cases here... (I assume all valid tensor should have defined data type) Please take a look at these new if conditions. Thank you!

onnx/defs/tensor_proto_util.cc
```diff
@@ -30,45 +31,63 @@ namespace ONNX_NAMESPACE {
...
-    /* misalignment issues on certain platforms, such as arm32-v7a */      \
-    const size_t raw_data_size = raw_data.size();                          \
-    res.resize(raw_data_size / sizeof(type));                              \
-    memcpy(reinterpret_cast<char*>(res.data()), bytes, raw_data_size);     \
-    return res;                                                            \
+#define DEFINE_PARSE_DATA(type, typed_data_fetch)                              \
+  template <>                                                                  \
+  const std::vector<type> ParseData(const TensorProto* tensor_proto) {         \
+    std::vector<type> res;                                                     \
+    if (!tensor_proto->has_data_type() ||                                      \
```
askhade, line 38: these if else conditions are a bit unintuitive
IMO - first check should be if data is external - if yes then raise an error.
Next - check for typed data and fetch types data
Next - check raw for raw data and parse raw data
Lastly - if none of the conditions were satisfied raise an error

onnx/external_data_helper.py
askhade: more elaborate would be nice - something like
Writing to external data happens in 2 passes. In first pass tensors with raw data which pass the necessary conditions (size threshold etc) are marked for serialization. In the second pass, the raw data in these tensors is serialized to a file. This is the second step. Serialize if tensor has raw data and it was marked for serialization.

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,124 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
+        model_file_path = self.get_temp_model_filename()
+        onnx.save_model(self.model,
+                        model_file_path,
+                        save_as_external_data=True,
+                        location=None,
+                        size_threshold=0,
+                        convert_attribute=False)
+        # Save without load_external_data
+        model = onnx.load_model(model_file_path, load_external_data=False)
+        onnx.save_model(model,
```
askhade, line 355: do you have a test where both saves will save some tensors as external data? this will validate whether the write to external file during second save is done properly (meaning offsets are honored etc)

During the second save you can turn on attribute serialization to external data to make the test easy

onnx/defs/tensor_proto_util.cc
```diff
@@ -30,45 +31,63 @@ namespace ONNX_NAMESPACE {
...
+      tensor_proto->data_type() == TensorProto_DataType_UNDEFINED) {           \
+      fail_shape_inference("The type of Tensor: ", tensor_proto->name(),       \
+        " is undefined so it cannot be parsed.");                              \
+    } else if (tensor_proto->has_data_location() &&                            \
+      tensor_proto->data_location() == TensorProto_DataLocation_EXTERNAL) {    \
+      fail_shape_inference("Cannot parse data from external tensors. Please ", \
+        "load external data into raw data for tensor: ", tensor_proto->name());\
+    } else if (!tensor_proto->has_raw_data()) {                                \
+      const auto& data = tensor_proto->typed_data_fetch();                     \
+      int expected_size = 1;                                                   \
```
askhade, line 48: should the default value be 0? if dims is empty then you will not enter the for loop so expected_size will be 1. On line 52 you make an exception for 0 and not 1

onnx/external_data_helper.py
jcwchen: Updated. Thanks

onnx/test/test_external_data.py
jcwchen: Just added a test `test_save_model_with_external_data_and_threshould` to test saving a model with typical tensor and external tensor.

jcwchen:
> I think we should convert "set_external_data" to private method

According to our offline discussion, there might be existing users using it... (e.g., ONNXRuntime) ONNX should keep it as public function at this moment.

onnx/defs/tensor_proto_util.cc
```diff
@@ -30,45 +31,63 @@ namespace ONNX_NAMESPACE {
...
+      for (int i = 0; i < tensor_proto->dims_size(); ++i) {                    \
+        expected_size *= tensor_proto->dims(i);                                \
+      }                                                                        \
+      if (tensor_proto->dims_size() != 0 && data.size() != expected_size) {    \
+        fail_shape_inference("The data size of tensor: ", tensor_proto->name(),\
+        " was expected to have (calculated by dims) ", expected_size,          \
+        " but the actual data size was ", data.size());                        \
```
askhade, line 55: rephrase error message as "Data size mismatch. Tensor: <name> expected size <expected_size> does not match the actual size <data.size()>"

onnx/external_data_helper.py
```diff
@@ -258,8 +256,13 @@ def write_external_data_tensors(model, filepath):  # type: (ModelProto, Text) ->
...
     for tensor in _get_all_tensors(model):
-        if uses_external_data(tensor):
+        # Writing to external data happens in 2 passes:
+        # 1. Tensors with raw data which pass the necessary conditions (size threshold etc) are marked for serialization
+        # 2. The raw data in these tensors is serialized to a file
+        # Thus serialize only if tensor has raw data and it was marked for serialization
+        if uses_external_data(tensor) and tensor.HasField("raw_data"):
             save_external_data(tensor, filepath)
             tensor.ClearField(str('raw_data'))
+        # If raw_data does not exist, simply skip since it should be loaded into external tensors already
```
askhade, line 266: no need for this comment now that you added detailed comments on line 259

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,186 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
         self.assertFalse(attribute_tensor.HasField("data_location"))
         self.assertTrue(np.allclose(to_array(attribute_tensor), self.attribute_value))
 
+    def test_save_model_does_not_load_external_tensor(self):  # type: () -> None
```
askhade, line 345: nit: rename test_save_model_w/o_loading_external_data

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,186 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
+            inputs=[],
+            outputs=['values'],
+            value=w_large
+        )
+        graph = helper.make_graph([constant_node_small, constant_node_large], "test_graph",
+                                  inputs=[], outputs=[],
+                                  initializer=[w_small, w_large])
         model = helper.make_model(graph)
+        # 1st save: save two tensors which have raw_data
+        # Only w_large will be stored as external tensors since it's large than 1024
```
askhade, line 398: nit: larger

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,186 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
+        self.assertTrue(w_large_tensor.HasField("data_location"))
+        self.assertTrue(np.allclose(to_array(w_large_tensor, self.temp_dir), data_w_large))
+
+        # 2nd save: one tensor has raw_data (small); save one external tensor (large)
+        onnx.save_model(model_without_loading_external,
+                        model_file_path,
+                        save_as_external_data=True,
+                        all_tensors_to_one_file=False,
+                        location=None,
+                        size_threshold=1024,
```
askhade, line 422: size_threshold is still 1024 -> 0?

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,186 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
-            constant_array.append(constant_node)
+        attribute_tensor = model.graph.node[0].attribute[0].t
+        self.assertFalse(attribute_tensor.HasField("data_location"))
+        self.assertTrue(np.allclose(to_array(attribute_tensor), self.attribute_value))
 
-        initializers = [self.create_external_data_tensor(self.initializer_value, "input_value")]
-        inputs = [helper.make_tensor_value_info("input_value",
-                                                onnx.TensorProto.FLOAT,
-                                                self.initializer_value.shape)]
+    def test_save_model_with_external_data_and_threshould(self):  # type: () -> None
```
askhade, line 372: nit: rename : test_save_model_with_external_data (dont think threshold is necessary here) the test checks whether multiple saves can work

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,186 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
+                        convert_attribute=True)
+
+        model_without_loading_external = onnx.load(model_file_path, load_external_data=False)
+        w_small_tensor = model.graph.initializer[0]
+        self.assertTrue(not w_small_tensor.HasField("data_location"))
+        self.assertTrue(np.allclose(to_array(w_small_tensor), data_w_small))
+
+        w_large_tensor = model.graph.initializer[1]
+        self.assertTrue(w_large_tensor.HasField("data_location"))
+        self.assertTrue(np.allclose(to_array(w_large_tensor, self.temp_dir), data_w_large))
```
askhade, line 432: suggest testing that external data can be loaded for these tensors... this will verify the file writes did not mess anything...

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,186 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
+        model = onnx.load_model(model_file_path, load_external_data=False)
+        initializer_tensor = model.graph.initializer[0]
+        initializer_tensor.raw_data = b'dummpy_raw_data'
+        # If raw_data and external tensor exist at the same time, override existing raw_data
+        load_external_data_for_tensor(initializer_tensor, self.temp_dir)
+        self.assertEqual(initializer_tensor.raw_data, original_raw_data)
+
+    def test_reshape_inference_with_external_data_fail(self):  # type: () -> None
+        reshape_shape = (2, 12)
+        X = helper.make_tensor_value_info('X', TensorProto.FLOAT, [2, 3, 4])
```
askhade, line 489: see if you can move the model creation code to a utility function and call it from all these tests... atleast a couple of tests can share the same model

onnx/test/test_external_data.py
jcwchen: to_array has already loaded the external tensors here. I thought it should be covered?

onnx/test/test_external_data.py
```diff
@@ -342,49 +342,186 @@ def test_save_model_does_convert_and_saves_the_model(self):  # type: () -> None
...
+                        model_file_path,
+                        save_as_external_data=True,
+                        all_tensors_to_one_file=False,
+                        location=None,
+                        size_threshold=1024,
+                        convert_attribute=True)
+
+        model_without_loading_external = onnx.load(model_file_path, load_external_data=False)
+        w_small_tensor = model.graph.initializer[0]
+        self.assertTrue(not w_small_tensor.HasField("data_location"))
```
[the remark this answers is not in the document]
jcwchen: Yes. I should use `w_small_tensor = model_without_loading_externall.graph.initializer[0]` here. Updated

onnx/test/test_external_data.py
askhade: aah my bad... I missed it... looks good then

onnx/test/test_external_data.py
jcwchen: I have made these 3 tests (test_reshape_inference_with_external_data_fail, test_to_array_with_external_data, test_save_model_with_external_data_multiple_times) using the same model, which simplifies the code. Thanks for pointing out.

askhade (approved):

Issue #3493, opened 2021-05-14
PyTorch export crash with onnx >= 1.8.0 unless import onnx first on windows


BowenBao:
Similar to #2808

The script crashes when onnx is imported after pytorch. The script completes successfully when onnx is imported before pytorch, or when onnx installed version <= 1.7.0

```python
# import onnx  # uncomment and pass
import torch
from onnx import ModelProto

class M(torch.nn.Module):
    def forward(self, x, y):
        return x + y

x = torch.randn(2, 3)
y = torch.randn(2, 3)

import io
f = io.BytesIO()
torch.onnx.export(M(), (x, y), f, verbose=True, input_names=['x', 'y'])
```

Both packages are installed like below
```
pip install torch==1.8.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
pip install onnx==1.8
```

Based on previous issues, I suspect this is also related to different protobuf versions. Can someone explain what is going on underneath, and how to resolve & prevent this in the future?

jcwchen:
Hi @BowenBao,
On Mac I solved a similar issue by upgrading protobuf-compiler (to 3.11.3) and MacOS (10.15). However, on Windows, basically ONNX==1.7 and ONNX>=1.8 are using the same way to install protobuf (3.11.3 static protobuf) as dependency so it looks weird to me that why only ONNX 1.7 works... I have questions:
1. Did you only see this crash on Windows?
2. Will you see this crash if you pip install torch then install ONNX from source (current main branch)?

Thank you!

jcwchen:
BTW, I can repro it with ONNX 1.9.0 as well..

BowenBao:
hi @jcwchen, 
1. yes only on windows.
2. yes observe same behavior with ONNX built from source, with libprotobuf & protobuf installed by `conda install -c conda-forge  libprotobuf=3.11.3 protobuf`.

askhade:
@BowenBao : how are you installing onnx? from pypi or conda?

BowenBao:
> 
> 
> @BowenBao : how are you installing onnx? from pypi or conda?

pypi via `pip install onnx==1.8`

jcwchen:
Hi @BowenBao,
When trying to figure out whether this issue still exists in upcoming ONNX 1.10.0, I have some new findings:

[...]

PR #3644, opened 2021-08-11
Use MSVC Runtime as dll for official ONNX Windows release

jcwchen:
**Description**
To make the official ONNX package more compatible, set USE_MSVC_STATIC_RUNTIME=0 in the release Windows CI.

**Motivation and Context**
https://github.com/onnx/onnx/issues/3493 PyTorch exporter bumped into an unknown crash with ONNX PyPI package >= 1.8 in a conda environment. Building MSVC Runtime as dll can resolve that issue. Meanwhile @snnn will reach out VC team to dig deeper into it. 

~One thing to note is: although I don't know the context before ONNX 1.6, it seems to me that previously official ONNX was not built with USE_MSVC_STATIC_RUNTIME=1: https://github.com/onnx/wheel-builder/blob/master/appveyor.yml There is no environment variable about it and by default MSVC is dynamic linked to ONNX. This might be the reason why ONNX <= 1.6 won't bump into this crash.~

I didn't see strong drawback if ONNX dynamically links to MSVC. Please let me know if anyone has any concern. Thanks!

Reference -- static/dynamic linking MSVC discussions: https://stackoverflow.com/questions/14749662/microsoft-visual-studio-c-c-runtime-library-static-dynamic-linking

### DIFF
--- a/.azure-pipelines/Linux-CI.yml
+++ b/.azure-pipelines/Linux-CI.yml
@@ -4,7 +4,7 @@ trigger:
 jobs:
 - job: 'Test'
   pool:
-    vmImage: 'Ubuntu-16.04'
+    vmImage: 'Ubuntu-18.04'
   strategy:
     matrix:
       py36:


snnn:
@BowenBao could you please help validate this package?

askhade:
@jcwchen : You mention for ONNX <= 1.6 this was not a issue because onnx was not linking to msvc runtime libs statically but actually it was see: https://github.com/onnx/onnx/blob/rel-1.6.0/setup.py#L176

IMO this is a good change because we have run into other issues as well because of static linking (like when onnx throws exceptions we see crashes in some situations) plus other benefits as well.

I don't know the exact rationale behind choosing static linking but it could be that if we link statically then it is contained within onnx binaries and users don't have to worry about installing\no versioning issue etc...

askhade:
Please update other files like README.md and setup.py

setup.py
```diff
@@ -54,8 +54,8 @@
 ONNX_BUILD_TESTS = bool(os.getenv('ONNX_BUILD_TESTS') == '1')
 ONNX_DISABLE_EXCEPTIONS = bool(os.getenv('ONNX_DISABLE_EXCEPTIONS') == '1')
 
-DEBUG = bool(os.getenv('DEBUG'))
-COVERAGE = bool(os.getenv('COVERAGE'))
+DEBUG = bool(os.getenv('DEBUG', '1') == '1')
+COVERAGE = bool(os.getenv('COVERAGE', '1') == '1')
```
jcwchen, line 58: It's not related to the main issue in this PR, but I think it would be better if we can sync how to set config variable to prevent confusion. I thought DEBUG=0 should build a release build yet it did not..

jcwchen:
> @jcwchen : You mention for ONNX <= 1.6 this was not a issue because onnx was not linking to msvc runtime libs statically but actually it was see: https://github.com/onnx/onnx/blob/rel-1.6.0/setup.py#L176

Thank you for catching this! I wasn't aware that setup.py has been updated about USE_MSVC_STATIC_RUNTIME after ONNX 1.6.

> Please update other files like README.md and setup.py

I just have updated README.md about default USE_MSVC_STATIC_RUNTIME=0; setup.py has already used USE_MSVC_STATIC_RUNTIME=0 as default. Thanks!

jcwchen:
One question: Is it OK to let ONNX change the default value in setup.py? Some users might already build ONNX in their project and always use the default value for USE_MSVC_STATIC_RUNTIME. They will possibly bump into issue due to this change (Although it will be an easy fix to them).

askhade (approved):

snnn:
Let me add a summary:  the static one would work if no C++ exception came across DLL boundaries. Who threw an exception, who must process it. You can't throw an exception from ONNX to pytorch. If ONNX needs to tell pytorch that something went wrong,  it should use error code, returned values or python exceptions instead of C++ exceptions.

Issue #3613, opened 2021-07-28
ONNX checker miscounts number of subgraph inputs when optional inputs are present in IR <= 3 model format


TomWildenhain-Microsoft:
# Bug Report

ONNX checker miscounts number of subgraph inputs when optional inputs are present in IR <= 3 model format

Details here: https://github.com/onnx/tensorflow-onnx/issues/1623#issuecomment-887004847

PR #3664, opened 2021-08-18
Enable checker/shape_inference to handle IR <= 3 models with subgraph

jcwchen:
**Description**
- Introduce ir_version in GraphInferencer to enable it to handle different cases for different IR versions
- For IR <=3, it does not require the number of graph input needs to be the same as the number of node input. (It can be more, but it can not be less)
- Add a test

**Motivation and Context**
https://github.com/onnx/onnx/issues/3613 ONNX checker miscounts number of subgraph inputs when optional inputs are present in IR <= 3 model format. It was not caught because checker of ONNX <= 1.9.0 does not use strict onnx.shape_inference. Models which use old IR <= 3 require all initializers to be optional inputs. So the logic needs to be updated to handle IR <=3.


onnx/shape_inference/implementation.cc
```diff
@@ -687,9 +687,18 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
...
   SymbolTable* symbolTable = getSymbolTable();
   int numInputs = int(inputTypes.size());
+  if (getIRVersion() >= 4) {
+    if (g_->input_size() != numInputs) {
+      fail_shape_inference("Graph has ", g_->input_size(), " inputs but ", numInputs, " were provided");
+    }
+  } else {
+    // IR < 4 requires all initializers to be optional inputs
+    // So the number of graph input can be larger than the number of node input 
+    if ( numInputs > g_->input_size()) {
```
gramalingam, line 697: I suggest that if (numInputs < g_input_size()), then we check to ensure that the missing inputs all have an entry in the initializers? That is, if the graph has inputs [x, y, z] while numInputs is 2, then we must check to make sure that z is in the initializer list.

gramalingam:
I suggest updating the docs/IR.md file as well to document this. See the section https://github.com/onnx/onnx/blob/master/docs/IR.md#nodes, the line "In nested subgraphs used as attribute values, users MUST NOT use the same name etc.". We should add "In models with IR version >= 4" before this line, and add "In models with IR version <= 3, users MAY use the same name as both a subgraph initializer and subgraph input.". I think a bit more is required, but I will bring it up in the original issue discussion to check with the converter usage.

gramalingam:
Ok, I suggest adding the following lines to IR.md after the lines mentioned above: "Note that in models with IR version <= 3, the actual inputs expected by the control-flow operator and supplied by the node containing the invocation of the control-flow operator must appear as the initial inputs of the subgraph, with extraneous initializers following subsequently, so that the actual inputs and formal inputs can be matched positionally."

onnx/shape_inference/implementation.cc
jcwchen: Good idea. I have added this case and written a test for it. Thanks!

jcwchen:
Have updated the docs. Thanks @gramalingam for the sentences

gramalingam:
LGTM, thanks! @TomWildenhain-Microsoft , can you please take a look? Thanks!

onnx/shape_inference/implementation.cc
```diff
@@ -687,9 +687,30 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
     const std::vector<const TensorProto*>& inputData) {
   SymbolTable* symbolTable = getSymbolTable();
   int numInputs = int(inputTypes.size());
+  std::unordered_set<std::string> initializerNameSet;
 
-  if (g_->input_size() != numInputs) {
-    fail_shape_inference("Graph has ", g_->input_size(), " inputs but ", numInputs, " were provided");
+  for (const auto& tp : g_->initializer()) {
```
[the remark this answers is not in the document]
jcwchen: Good catch! Just updated

gramalingam (approved):

onnx/shape_inference/implementation.cc
```diff
@@ -688,8 +688,28 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
   SymbolTable* symbolTable = getSymbolTable();
   int numInputs = int(inputTypes.size());
 
-  if (g_->input_size() != numInputs) {
-    fail_shape_inference("Graph has ", g_->input_size(), " inputs but ", numInputs, " were provided");
+  if (getIRVersion() >= 4) {
+    if (g_->input_size() != numInputs) {
+      fail_shape_inference("Graph has ", g_->input_size(), " inputs but ", numInputs, " were provided");
```
TomWildenhain-Microsoft, line 693: Does this properly enforce that none of these inputs have initializers?

docs/IR.md
```diff
@@ -237,7 +237,7 @@ A name belonging to the Value namespace may appear in multiple places, namely as
...
+When a name appears in both the initializer list and the graph input list, a runtime MAY allow a caller to specify a value for this (input) name overriding the value specified in the initializer and a runtime MAY allow users to omit specifying a value for this (input) name, choosing the value specified in the initializer. Names of constants that are not meant to be overridden by the caller should appear only in the initializer list and not in the graph input list. In models with IR version >= 4, in nested subgraphs used as attribute values, users MUST NOT use the same name as both a subgraph initializer and subgraph input unless the corresponding op's specification explicitly allows it. In models with IR version <= 3, users MAY use the same name as both a subgraph initializer and subgraph input. Note that in models with IR version <= 3, the actual inputs expected by the control-flow operator and supplied by the node containing the invocation of the control-flow operator must appear as the initial inputs of the subgraph, with extraneous initializers following subsequently, so that the actual inputs and formal inputs can be matched positionally.
```
TomWildenhain-Microsoft, line 240: I like the clarification, but for IR version <= 3 this is not 100% clear as to whether inputs with initializers (optional inputs) can be used as initial, actual inputs (provided they line up correctly).  If you are disallowing it, you should maybe change
`the actual inputs expected by the control-flow operator and supplied by the node containing the invocation of the control-flow operator must appear as the initial inputs of the subgraph, with extraneous initializers following subsequently, so that the actual inputs and formal inputs can be matched positionally.`
to
`the actual inputs expected by the control-flow operator and supplied by the node containing the invocation of the control-flow operator cannot have initializers and must appear as the initial inputs of the subgraph, so that the actual inputs and formal inputs can be matched positionally. Extraneous initializers may follow subsequently.`

onnx/shape_inference/implementation.cc
```diff
@@ -688,8 +688,28 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
...
+      initializerNameSet.insert(tp.name());
+    }
+    // IR < 4 requires all initializers to be optional inputs
+    // So the number of graph input can be larger than the number of node input 
+    if (numInputs > g_->input_size()) {
+      fail_shape_inference("Graph has ", g_->input_size(), " inputs but ", numInputs, " were provided.",
+        "The number of graph input cannot be smaller than the number of node input" );
+    } else if (numInputs < g_->input_size()) {
+      // Since the number is different, further check whether the additional input is in initializers
+      for (int i = numInputs; i < g_->input_size(); ++i) {
```
TomWildenhain-Microsoft, line 707: You also need to check that the initial inputs do not have initializers, if you are disallowing them. (see first comment)

docs/IR.md
gramalingam: How about changing the two sentences to
```
In models with IR version <= 3, users MAY use the same name as both a subgraph initializer and subgraph input,
but this is restricted to support constants via initializers that are not intended to correspond to any
actual inputs expected by the control-flow operator and supplied by the node containing the invocation
of the control-flow operator. In particular, the control-flow operator semantics determines the set of inputs
supplied to the execution of the subgraph, and these input names MUST NOT appear as subgraph initializers.
Subgraph initializer names must appear in the graph input list _after_ the actual inputs. This allows the actual
inputs and formal inputs to be matched positionally.
```

TomWildenhain-Microsoft: Yes that's good.  nit: I'd revise `not intended to correspond to any
actual inputs expected by the control-flow operator and supplied by the node containing the invocation of the control-flow operator.` 
to 
`not intended to correspond to any actual inputs passed from the node into the subgraph.`

onnx/shape_inference/implementation.cc
jcwchen: Good point. I added the check (initial inputs do not have initializers) here and there (your first comment). Thanks!

docs/IR.md
jcwchen: I took both Rama's new sentences and Tom's revision since it looks clearer to me

onnx/shape_inference/implementation.cc
jcwchen: No it does not and I have added a check for it. Thanks

TomWildenhain-Microsoft (approved):

onnx/shape_inference/implementation.cc
```diff
@@ -687,25 +687,34 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
...
     // IR < 4 requires all initializers to be optional inputs
     // So the number of graph input can be larger than the number of node input 
     if (numInputs > g_->input_size()) {
       fail_shape_inference("Graph has ", g_->input_size(), " inputs but ", numInputs, " were provided.",
         "The number of graph input cannot be smaller than the number of node input" );
     } else if (numInputs < g_->input_size()) {
-      // Since the number is different, further check whether the additional input is in initializers
       for (int i = numInputs; i < g_->input_size(); ++i) {
-        if (initializerNameSet.count(g_->input(i).name()) == 0) {
+        if (i < numInputs && initializerNameSet.count(g_->input(i).name()) > 0) {
```
gramalingam, line 713: I think this may also require changing the for-loop lower bound to ```i = 0``` ?

gramalingam: Or, split it into 2 separate loops.

gramalingam (approved):

PR #3666, opened 2021-08-18
Test source distribution in CI; Add a label for manually trigger testing ONNX Model Zoo CI

jcwchen:
**Description**
- Build and test source distribution in CI weekly (in one environment should be sufficient); Also onnx-weekly TestPyPI package will have source distribution as well
- Add a new label "test ONNX Model Zoo" to manually trigger CI testing the PR with all models from ONNX Model Zoo (use it with caution because it will waste git-lfs quota in onnx/models)

**Motivation and Context**
Enhance CIs

### DIFF
--- a/.github/workflows/release_mac.yml
+++ b/.github/workflows/release_mac.yml
@@ -106,7 +106,7 @@ jobs:
         
         # Test the uploaded source distribution
         python -m pip uninstall -y onnx
-        python -m pip install -i https://test.pypi.org/simple/ --no-binary :all: onnx
+        python -m pip install -i https://test.pypi.org/simple/ --no-binary :all: onnx-weekly
         pytest
 
     - name: Verify ONNX with the latest numpy
--- a/.github/workflows/release_mac.yml
+++ b/.github/workflows/release_mac.yml
@@ -94,10 +94,20 @@ jobs:
         name: wheels
         path: dist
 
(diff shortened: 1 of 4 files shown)


.github/workflows/release_mac.yml
```diff
@@ -94,10 +94,20 @@ jobs:
...
+        twine upload --verbose onnx/dist/*.whl --repository-url https://test.pypi.org/legacy/ -u ${{ secrets.TESTPYPI_USERNAME }} -p ${{ secrets.TESTPYPI_PASSWORD }}
+        
+        # Build and upload source distribution to TestPyPI
+        git clean -xdf
+        python setup.py sdist
+        twine upload dist/* --repository-url https://test.pypi.org/legacy/ -u ${{ secrets.TESTPYPI_USERNAME }} -p ${{ secrets.TESTPYPI_PASSWORD }}
+        
+        # Test the uploaded source distribution
+        python -m pip uninstall -y onnx
+        python -m pip install -i https://test.pypi.org/simple/ --no-binary :all: onnx
```
askhade, line 109: this is tetsing the weekly package right? install onnx-weekly?

jcwchen: Oh it should be onnx-weekly. Updated. Thank you for catching this

PR #3667, opened 2021-08-18
CMakeLists: Do not write generated files to the source dir

geofft:
This makes sure that out-of-source builds can be performed cleanly, and
in particular that multiple out-of-source builds can be performed from
the same directory.

After this commit, you can run cmake with a read-only source directory.

Signed-off-by: Geoffrey Thomas <geofft@twosigma.com>


geofft:
Is this required? https://github.com/onnx/onnx/blob/master/CONTRIBUTING.md says that you're moving away from the CLA, and signing a CLA is a bunch of corporate paperwork, so I'd prefer to avoid it if possible. (DCO is of course fine :) )

prasanthpul:
CLA is not required. just DCO. Will remove the CLA bot.

geofft:
The CLA bot is still asking for a CLA. Should I start the process on my side to do CLA paperwork?

Also, can someone take a look at the actual change? It's very tiny :)

geofft:
Hello, can someone please take a look at this PR and also why the CLA bot is asking for a CLA?

jcwchen (approved):
LGTM. I agree the philosophy that ideally we should not write generated files to the source directory.

PR #3604, opened 2021-07-23
update mypy to latest (WIP)

stillmatic:
**Description**
- Updates the mypy version from 0.600 to 0.910
- Removes the `protobuf` stubs since latest mypy includes them
- Removes `six` usages as Python 2 support is deprecated
- Fixes issues as they arise

**Motivation and Context**
- Allows us to use modern mypy versions with onnx 
- Previous works and more context: #3211

PR #3668, opened 2021-08-19
Upgrade mypy to 0.620 to retire stubs/google/protobuf

jcwchen:
**Description**
Upgrade mypy to 0.620 to retire stubs/google/protobuf in ONNX repo
In the future, ONNX should upgrade it to more recent version although many codes need to be updated.

**Motivation and Context**
mypy 0.600 is too old and it has a few limitations


jcwchen:
Another PR does a better job with this (upgrade to higher version): https://github.com/onnx/onnx/pull/3604. Close this now.

PR #3676, opened 2021-08-24
Fix IR gap for version_converter: update node input from initializer directly 

jcwchen:
**Description**
Similar to an old PR https://github.com/onnx/onnx/pull/3007, but instead of adding initializer into input and removing (sort of hacky), this PR creates a Value directly.
-  Enable node to find input from initializers: create a tensor map from initializers, convert it (Tensor) into a Value if the name is matched and add it into node inputs
- (WIP) Add a test

**Motivation and Context**
Solve input not found issue for version converter since initializers are not included in inputs for newer model (IR >= 4): https://github.com/onnx/onnx/issues/2873

### DIFF
--- a/onnx/common/ir.h
+++ b/onnx/common/ir.h
@@ -542,6 +542,10 @@ public:
     return inputs_.at(i);
   }
 
+  int inputSize() const {
+    return inputs_.size();
+  }
+
   // Graphs
 
   // Note [Topological invariant]
--- a/onnx/common/ir_pb_converter.cc
+++ b/onnx/common/ir_pb_converter.cc
@@ -11,7 +11,8 @@
 namespace ONNX_NAMESPACE {
 
 // Part 1: convert ONNX Protobuf to IR
-std::unique_ptr<Graph> graphProtoToGraph(const GraphProto& gp, bool nested);
+std::unique_ptr<Graph> graphProtoToGraph(const GraphProto& gp, bool nested,
+  const int ir_version);
 
 Tensor tensorProtoToTensor(const ONNX_NAMESPACE::TensorProto& tp) {
   Tensor ret;
@@ -94,7 +95,8 @@ Tensor tensorProtoToTensor(const ONNX_NAMESPACE::TensorProto& tp) {
   return ret;
 }
 
-void convertAttribute(const ONNX_NAMESPACE::AttributeProto& ap, Node* n) {
+void convertAttribute(const ONNX_NAMESPACE::AttributeProto& ap, Node* n,
+  const int ir_version) {
   Symbol sym = Symbol(ap.name());
   switch (ap.type()) {
     case ONNX_NAMESPACE::AttributeProto_AttributeType_FLOAT:
@@ -158,13 +160,13 @@ void convertAttribute(const ONNX_NAMESPACE::AttributeProto& ap, Node* n) {
       break;
     }
     case ONNX_NAMESPACE::AttributeProto_AttributeType_GRAPH:
-      n->g_(sym, graphProtoToGraph(ap.g(), true));
+      n->g_(sym, graphProtoToGraph(ap.g(), true, ir_version));
       break;
     case ONNX_NAMESPACE::AttributeProto_AttributeType_GRAPHS: {
       std::vector<std::shared_ptr<Graph>> graphs;
       graphs.reserve(ap.graphs_size());
       for (int i = 0; i < ap.graphs_size(); i++) {
-        graphs.push_back(graphProtoToGraph(ap.graphs(i), true));
+        graphs.push_back(graphProtoToGraph(ap.graphs(i), true, ir_version));
       }
       n->gs_(sym, std::move(graphs));
       break;
@@ -179,9 +181,9 @@ void convertAttribute(const ONNX_NAMESPACE::AttributeProto& ap, Node* n) {
   }
 }
 
-void convertAttributes(ONNX_NAMESPACE::NodeProto& np, Node* n) {
+void convertAttributes(ONNX_NAMESPACE::NodeProto& np, Node* n, const int ir_version) {
   for (int i = 0; i < np.attribute_size(); i++) {
-    convertAttribute(np.attribute(i), n);
+    convertAttribute(np.attribute(i), n, ir_version);
   }
 }
 
@@ -202,7 +204,8 @@ std::vector<Dimension> tensorShapeProtoToDimensions(const ONNX_NAMESPACE::Tensor
   return dims;
 }
 
-std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp, bool nested) {
+std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp,
+  bool nested, const int ir_version) {
   std::unique_ptr<Graph> g(new Graph());
 
   if (gp.has_name()) {
@@ -227,6 +230,11 @@ std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp, b
   // objects out of them, and equal strings must be mapped to the same
   // Value object.
   std::unordered_map<std::string, Value*> value_by_name_of;
+
+  // Maintain a tensor map from initializer. If a node cannot find a input
+  // from graph inputs, it will look into graph initializers by this map.
+  std::unordered_map<std::string, Tensor> tensor_by_name_of;
+
 
   // We initialize Node inputs in a separate pass from the Nodes
   // themselves. To do so, we need to have access to the names of the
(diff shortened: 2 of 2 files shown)


jcwchen:
Since it can help to resolve the IR issue (input from initializer is not found) for version converter, the main contributor of current version converter @matteosal could you please help to review this? Thank you for your contribution!

Friendly ping ONNX IR experts @linkerzhang @daquexian @fumihwh Feel free to review this new PR for fixing the old IR issue.  Thanks!

onnx/common/ir.h
```diff
@@ -950,6 +953,17 @@ friend struct Value;
...
+  Value* addInitializerNode(Tensor initializer, std::string name) {
+    Value* init_value = initializerNode_->addOutput();
+    std::vector<Dimension> dim_sizes{initializer.sizes().cbegin(),
+                                    initializer.sizes().cend()};
+    init_value->setUniqueName(name);
+    init_value->setSizes(dim_sizes);
+    init_value->setElemType(initializer.elem_type());
+    return init_value;
+  }
   void eraseInitializer(const std::string &name) {
```
jcwchen, line 967: Need to erase created init node here... Doing

onnx/common/ir.h
```diff
@@ -950,6 +953,17 @@ friend struct Value;
     initializers_.push_back(std::move(initializer));
     initializer_names_.push_back(std::move(name));
   }
+  // For IR >= 4, initializer is not required to exist in input
+  // Add initializer into initializer node list and return its Value
+  Value* addInitializerNode(Tensor initializer, std::string name) {
```
askhade, line 958: did you check addInitializerAndInput line 1122. It does something similar

onnx/common/ir.h
askhade: Check eraseInitializerAndInput

onnx/common/ir_pb_converter.cc
```diff
@@ -288,6 +291,17 @@ std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp, b
...
   }
 
+  for (int i = 0; i < gp.initializer_size(); ++i) {
+    auto init = tensorProtoToTensor(gp.initializer(i));
+    g->addInitializer(init, init.name());
+    // If ir_version >= 4, initializer does not have to be included in input
+    // Create a Value from initializer by addInitializerNode
+    // and save it into value_by_name_of for later use (node input)
+    if (ir_version >= 4) {
+      value_by_name_of[init.name()] = g->addInitializerNode(init, init.name());
```
askhade, line 301: did you test this?
This is what happens in addInitializer: 
```
initializers_.push_back(std::move(initializer));
initializer_names_.push_back(std::move(name));
```
you are using initializer and name after move.

onnx/common/ir.h
```diff
@@ -879,6 +879,8 @@ friend struct Value;
   // having corner cases where the list is empty.
   Node * const output_;
   Node * const input_;
+  // Create an independent node list for initializer to prevent same name issue in inputs (SSA)
+  Node * const initializerNode_;
```
askhade, line 883: please use naming convention initializer_node_

askhade: also suggest renaming this to something like input_as_initializer_

onnx/common/ir_pb_converter.cc
gramalingam: I think it should be okay since addInitializer's parameters are passed by-value, so no impact on the values here?

onnx/test/version_converter_test.py
```diff
@@ -1527,6 +1527,34 @@ def test_if_subgraph_10_11(self):  # type: () -> None
...
+        graph = helper.make_graph(
+            nodes,
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (1, 2, 2, 3))],
+            [helper.make_tensor_value_info("Y", TensorProto.FLOAT, (1, 2, 2, 3))],
+            [scale_tensor, b_tensor, mean_tensor, var_tensor])
+
+        converted_model = self._converted(graph, helper.make_operatorsetid(
+            "", 11), 12)
+        # Assert equality of graph and converted_model
```
gramalingam, line 1554: Where is the test that an initializer is not in input?

gramalingam:
Sorry, I don't understand the context for this change. I have a feeling this is not doing the right thing. In particular, why should we change the conversion of GraphProto to IR? What's the issue with this conversion? It simply maps the inputs/initializers of the graph-proto to the IR, and, hence, should do the right thing whatever the IR version is. The likely issues are:
(a) Checking whether a particular node-input X has a constant-value: I think some of the existing transformers may actually be wrong if they pickup the constant-value from initializers, but don't check whether X is an input. (But I don't think this is the original issue people had. But this is a different soundness concern.)
(b) If the transformer creates an initializer, without a corresponding input, does the conversion of the IR to GraphProto complain? Where exactly is the original conversion failure happening?

jcwchen:
Thank you for the review! Let me give you more context from my understanding:
> why should we change the conversion of GraphProto to IR? What's the issue with this conversion? 

Previously the conversion only considers graph input for node input and does not consider initializer. And while loading node input, it won't consider initializer and complain input is undefined. To add initializer into node input, initializer needs to be converted to `Value` (I also stored them into `value_by_name_of` to simplify code). Original conversion cannot achieve that because the old IR only stores initializer as a vector of Tensor. That's why I additionally created a node list of initializer (just like what input did) for storing `Value` from initializer.

To prevent confusion, actually I was considering to remove `initializers_` because the new node list of initializer seems to cover all information and `initializers_` is not used internally from ONNX. (but others might have already relied on it and it needs more investigation)

> (a) Checking whether a particular node-input X has a constant-value: I think some of the existing transformers may actually be wrong if they pickup the constant-value from initializers, but don't check whether X is an input. (But I don't think this is the original issue people had. But this is a different soundness concern.)

Actually current IR node just does not consider initializer at all because old IR (< 4) always assumes initializer exist input.

> (b) If the transformer creates an initializer, without a corresponding input, does the conversion of the IR to GraphProto complain? Where exactly is the original conversion failure happening?

Yes it will complain [here](https://github.com/onnx/onnx/blob/a9c17dae3cd7a7948d9a813c76039487411aab20/onnx/common/ir_pb_converter.cc#L308) because the node input cannot find the corresponding "Value" from `value_by_name_of`. Original `value_by_name_of` only considers graph input. 

The discussion here leads me to think about another thing: I should further test it with certain operator's adapter for version_converter which involves initializer to verify this change.

onnx/common/ir_pb_converter.cc
gramalingam: @jcwchen : thanks for the explanation (in the other thread). In ir_version >= 4, initializer may or may not be in the input list. I think you need to check here whether initializer X already has a value (because it is in the input list) also. We should addInitializerNode only if X is not an input. (If you want value_by_name_of for all initializers, setting it should probably be moved out of the conditional. But do we need it?

gramalingam: Eg., change the condition to
```cpp
   if ( (ir_version >= 4) && (value_by_name_of.count(init.name()) > 0)) {
```

onnx/common/ir.h
```diff
@@ -946,10 +951,29 @@ friend struct Value;
     doc_string_ = std::move(doc_string);
   }
 
-  void addInitializer(Tensor initializer, std::string name) {
-    initializers_.push_back(std::move(initializer));
-    initializer_names_.push_back(std::move(name));
+  void addInitializer(Tensor& initializer) {
+    if (initializer.name().empty()) {
```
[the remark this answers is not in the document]
jcwchen: For initializer comes from real world models, no it cannot be empty.
 
However, the reason why I made the condition here is it can be useful for creating a new initializer during version_converter or testing. In that case users don't need to worry about how to create a unique name. Also this is how existing addInitializerAndInput is doing.

workflow_scripts/test_model_zoo.py
```diff
@@ -72,14 +73,18 @@ def main():
...
+                # stricter onnx.checker with onnx.shape_inference
+                onnx.checker.check_model(model, True)
+                # remove the model to save space in CIs
+                if os.path.exists(model_path):
+                    os.remove(model_path)
+                # clean git lfs cache
+                run_lfs_prune()
+                print('[PASS]: {} is checked by onnx. '.format(model_name))
+            else:
+                print('[SKIP]: {}.'.format(model_name))
```
jcwchen, line 87: This is for testing version_converter with models from ONNX model zoo. Will remove this update before merge.

matteosal:
@jcwchen are there any news for this?

jcwchen:
> @jcwchen are there any news for this?

Before merging it, I was doing more verification because it's a fundamental change. Thank you for letting know yolov4 was the model which cannot be converted. I have verified that model can be converted correctly by this PR and also made sure the output by ONNXRuntime remains the same. Besides, I tested this PR with all models (IR <= 3) from ONNX model Zoo and all models seems converted successfully as before (same failure number due to other reasons).

Please let me know if there are other models which have this issue as well. We hope we can verify this PR as much as we could. If everything goes well, we will forward this PR before the next ONNX release. Thank you!

matteosal:
@jcwchen we have some new unittests which are very suitable to test this fix but I've realized that this PR only contains a change in test_model_zoo.py and nothing about the IR. Am I missing something? Where is the fix?

jcwchen:
> @jcwchen we have some new unittests which are very suitable to test this fix 

@matteosal Good to know! Thank you for testing.

> but I've realized that this PR currently only contains a change in test_model_zoo.py and nothing about the IR. Am I missing something? Where is the fix?

Sorry that last time I somehow messed up the commit log. I just recovered it. Thank you for the reminder.

matteosal:
@jcwchen 
I have ran our tests and found no problem at all. They consist in the following steps for 700+ operators:
* Export an op from our framework to old ONNX opset versions (without including initializers in the input list)
* Use the version converter to upgrade the ONNX model to opset 15
* Make a roundtrip by importing the operator back and run consistency tests between the original and the roundtripped op

Everything went smooth by using this branch, so I would consider it to be ready

gramalingam (approved):

askhade (approved):

PR #3689, opened 2021-08-31
Default to donot check model

jackwish:
Some dynamic shape models may fail due to lack of shape information,
don't make the script crash by default.
(Note: run shape inference can fix the shape info.)


onnx/utils.py
```diff
@@ -133,7 +133,8 @@ def extract_model(
         input_path,  # type: Text
         output_path,  # type: Text
         input_names,  # type: List[Text]
-        output_names  # type: List[Text]
+        output_names,  # type: List[Text]
+        check_model=False,  # type: Boolean
```
askhade, line 137: make this true by default so that existing scenarios are unaffected.

onnx/utils.py
```diff
@@ -148,6 +149,7 @@ def extract_model(
         output_path (string): The path to save the extracted ONNX model.
         input_names (list of string): The names of the input tensors that to be extracted.
         output_names (list of string): The names of the output tensors that to be extracted.
+        check_model (bool): The names of the output tensors that to be extracted.
```
askhade, line 152: Typo: Update comment. "Whether to run model checker on the extracted model"

onnx/utils.py
jackwish: Fixed. Thanks!

onnx/utils.py
jackwish: Fixed. Thanks!

askhade (approved):

Issue #3675, opened 2021-08-20
Onnx shape inference incorrect for ConstantOfShape


TomWildenhain-Microsoft:
# Bug Report

The following model is a ConstantOfShape op with a constant input [4, 3].  The output shape should match the value of the constant input, or at least be unknown.  Onnx reports the output shape as [12884901892].  Similar strange results appear for values other than [4, 3].

```python
from onnx import helper, numpy_helper, TensorProto

import onnx
import numpy as np

model = helper.make_model(
    opset_imports=[helper.make_operatorsetid('', 14)],
    ir_version=8,
    graph=helper.make_graph(
        name='infer-graph',
        inputs=[],
        outputs=[helper.make_tensor_value_info('tfl.scatter_nd__12:0', TensorProto.FLOAT, shape=None)],
        initializer=[numpy_helper.from_array(np.array([4, 3], dtype=np.int32), name='std.constant')],
        nodes=[helper.make_node('ConstantOfShape', inputs=['std.constant'], outputs=['tfl.scatter_nd__12:0'], name='tfl.scatter_nd__12')],
    ),
)

model_shapes = onnx.shape_inference.infer_shapes(model)
onnx.checker.check_model(model_shapes)
print(model_shapes.graph.output)
```

Output:
```
[name: "tfl.scatter_nd__12:0"
type {
  tensor_type {
    elem_type: 1
    shape {
      dim {
        dim_value: 12884901892
      }
    }
  }
}
]
```

Onnx version 1.10.1

TomWildenhain-Microsoft:
In certain cases, this shape inference can cause a Segmentation Fault.

jcwchen:
Hi @TomWildenhain-Microsoft,
I can repro this (int32 input does not work), but I tried it with np.int64 initializer and the output shape worked as expected. According to the [spec](https://github.com/onnx/onnx/blob/master/docs/Operators.md#type-constraints-24) of ConstantOfShape-9, it only supports int64 as input so the [shape inference](https://github.com/onnx/onnx/blob/2e947e5a3fb8be172cae549d03ae5aa7059f8747/onnx/defs/generator/defs.cc#L270) only handles the case of parse int64.

[...]

PR #3702, opened 2021-09-03
Check ParseData type to prevent segfault if parsing different types

jcwchen:
**Description**
Add a type check before parsing data to prevent unexpected behavior/segfault issues.
I only see ONNX uses int32_t/int64_t/float/double for ParseData. Please let me know if more types should be supported. Thanks!

**Motivation and Context**
https://github.com/onnx/onnx/issues/3675 If strict_mode shape_inference is silent, shape inference will still propagate the shape with an invalid type and unexpected behavior will happen (can be a wrong number or even a segfault). ParseData should stop early if the types mismatch.


onnx/defs/tensor_proto_util.cc
```diff
@@ -34,12 +34,29 @@ namespace ONNX_NAMESPACE {
 #define DEFINE_PARSE_DATA(type, typed_data_fetch)                              \
   template <>                                                                  \
   const std::vector<type> ParseData(const TensorProto* tensor_proto) {         \
-    std::vector<type> res;                                                     \
     if (!tensor_proto->has_data_type() ||                                      \
       tensor_proto->data_type() == TensorProto_DataType_UNDEFINED) {           \
       fail_shape_inference("The type of Tensor: ", tensor_proto->name(),       \
         " is undefined so it cannot be parsed.");                              \
-    } else if (tensor_proto->has_data_location() &&                            \
+    } else if (std::is_same<type, int64_t>::value  &&                          \
```
askhade, line 41: you should simplify this check
You can change the macro to take the enum type field something like this:

DEFINE_PARSE_DATA(int32_t, int32_data, TensorProto_DataType_INT32)

and then directly compare tensor_proto->data_type() == enum_type

askhade (changes requested):
Added comments

onnx/defs/tensor_proto_util.cc
jcwchen: Good idea. Updated.

onnx/defs/tensor_proto_util.cc
```diff
@@ -31,15 +31,19 @@ namespace ONNX_NAMESPACE {
...
+#define DEFINE_PARSE_DATA(type, typed_data_fetch, tensorproto_datatype)        \
   template <>                                                                  \
   const std::vector<type> ParseData(const TensorProto* tensor_proto) {         \
-    std::vector<type> res;                                                     \
     if (!tensor_proto->has_data_type() ||                                      \
       tensor_proto->data_type() == TensorProto_DataType_UNDEFINED) {           \
       fail_shape_inference("The type of Tensor: ", tensor_proto->name(),       \
         " is undefined so it cannot be parsed.");                              \
-    } else if (tensor_proto->has_data_location() &&                            \
+    } else if (tensor_proto->data_type() != tensorproto_datatype) {            \
```
askhade, line 41: do you need the if check above? Can you combine if and else if into 1 if loop?
You can update the error message here to say something like 
```
fail_shape_inference("ParseData type mismatch for tensor: ", tensor_proto->name(), "Expected", <>, "Actual", <>);
```

jcwchen: I think these two if are required since TensorProto_DataType_UNDEFINED is not in the existing type-string map? Two different error messages look clear to me for two different kinds of errors. Let me know if you still think they should be merged since we can still extend the map with TensorProto_DataType_UNDEFINED. ("undefined" to TensorProto_DataType_UNDEFINED)

onnx/defs/data_type_utils.h
```diff
@@ -47,6 +47,7 @@ class DataTypeUtils final {
   // If the DataType input is invalid, this function will throw std::invalid_argument exception.
   // If ONNX_NO_EXCEPTIONS is set it will abort.
   static const TypeProto& ToTypeProto(const DataType& data_type);
+  static std::string ToDataTypeString(int32_t tensor_data_type);
```
jcwchen, line 50: I moved this utility function to public function to get type string from tensor_data_type for ParseData

onnx/defs/tensor_proto_util.cc
askhade: No need to extend the map... as ToDataTypeString will throw an error if type is not in its mapping... Maybe it is OK to keep these 2 checks... Can you make sure we have enough tests for ParseData

jcwchen: Added one test to test invalid type for ParseData: strict shape inference should catch it and normal shape inference should not produce unexpected shape

jcwchen: Added another test to test undefined Tensor type

askhade (approved):

PR #3706, opened 2021-09-07
Spec clarification for QuantizeLinear and DequantizeLinear: 0d tensor without optional axis attribute

jcwchen:
**Description**
Add more explanation in the doc for QuantizeLinear and DequantizeLinear about 0D input tensors (y_scale and y_zero_point) without optional attribute (axis).

**Motivation and Context**
QuantizeLinear and DequantizeLinear allows 0d tensor as inputs (y_scale and y_zero_point), but it has an optional axis (by default it is 1). It is quite confused that whether 0d tensor without specified axis is a valid node. The doc needs to be clarified.


gramalingam (approved):

askhade (approved):

jcwchen:
cc @TomWildenhain-Microsoft

docs/Operators.md
```diff
@@ -4632,7 +4632,7 @@ Other versions of this operator: <a href="Changelog.md#DequantizeLinear-10">10</
 
 <dl>
 <dt><tt>axis</tt> : int (default is 1)</dt>
-<dd>(Optional) The axis of the dequantizing dimension of the input tensor. Negative value means counting dimensions from the back. Accepted range is [-r, r-1] where r = rank(input)</dd>
+<dd>(Optional) The axis of the dequantizing dimension of the input tensor. Negative value means counting dimensions from the back. Accepted range is [-r, r-1] where r = rank(input). If either y_zero_point or y_scale is a 1D tensor, then an axis must be specified. If both are scalars, axis is ignored.</dd>
```
TomWildenhain-Microsoft, line 4635: I like the line `If both are scalars, axis is ignored.`.  The line `If either y_zero_point or y_scale is a 1D tensor, then an axis must be specified.` seems to conflict with axis having a default value (and technically being 1 if it is unspecified).

gramalingam: Good catch. My fault, I didn't notice the default value and suggested a more complete description to Jacky.

TomWildenhain-Microsoft: I actually wish the default value didn't exist so we could do it your way but unfortunately it's too late to change I think.

jcwchen: Just removed `If either y_zero_point or y_scale is a 1D tensor, then an axis must be specified.` to prevent confusion. No worries @gramalingam  I didn't notice it in the first place and thank @TomWildenhain-Microsoft for catching it!

jcwchen: Just removed If either y_zero_point or y_scale is a 1D tensor, then an axis must be specified. to prevent confusion. No worries @gramalingam I didn't notice it in the first place and thank @TomWildenhain-Microsoft for catching it!

gramalingam: Minor nit: we should change ```both``` to ```both y_zero_point and y_scale``` since we deleted the preceding line. Otherwise, it is confusing what ```both``` refers to.

jcwchen: Updated. Thanks!

Issue #3715, opened 2021-09-14
TensorProto.FLOAT16 : TypeError: 0.0 has type float, but expected one of: int, long


iperov:
### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*):  Windows 10
- ONNX version (*e.g. 1.7*):  1.10.1
- Python version: 3.7.9
- Protobuf version: 3.17.3

### Reproduction instructions
 ```
onnx_helper.make_tensor(name='asd', data_type=TensorProto.FLOAT16, dims=(3,), vals=[0.0, 0.0, 0.0] )
```

```
TypeError: 0.0 has type float, but expected one of: int, long
```

### Expected behavior

expected no error

**Comment by jcwchen** [comment]
Hi @iperov,
Similar to https://github.com/onnx/onnx/issues/3204. ONNX uses int32_data to store `float16`: https://github.com/onnx/onnx/blob/a5e7ee51176bf78a60c118758174e13d85a87b46/onnx/mapping.py#L41 So you need a conversion from float16 to uint16 like:
```
original_val = np.array([0.0, 0.0, 0.0]).astype(np.float16)
converted_val = original_val.view(dtype=np.uint16).flatten()
onnx_helper.make_tensor(name='asd', data_type=TensorProto.FLOAT16, dims=converted_val.shape, vals=converted_val.tolist())
```

iperov:
but why? can you fix this shit?

PR #3716, opened 2021-09-14
Handle float16/bfloat16 to uint16 properly in helper functions

jcwchen:
**Description**
- Add conversion float16->uint16 in make_tensor so users can simply give it a float list or numpy array
- Do the same thing for bloat16
- Add tests

**Motivation and Context**
https://github.com/onnx/onnx/issues/3715 Currently users need additional code to convert float16 tensor. `make_tensor` should handle the conversion to simplify the usage.

### DIFF
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -251,17 +251,16 @@ def make_tensor(
         raise ValueError("Number of values does not match tensor's size. Expected {}, but it is {}. "
             .format(expected_size, len(vals)))
 
-    if (data_type == TensorProto.COMPLEX64
-            or data_type == TensorProto.COMPLEX128):
-        vals = split_complex_to_pairs(vals)
-    # floa16/bfloat16 are stored as uint16
-    elif (data_type == TensorProto.FLOAT16
-            or data_type == TensorProto.BFLOAT16):
-        vals = np.array(vals).astype(np.float16).view(dtype=np.uint16).flatten().tolist()
-
     if raw:
         tensor.raw_data = vals
     else:
+        if (data_type == TensorProto.COMPLEX64
+                or data_type == TensorProto.COMPLEX128):
+            vals = split_complex_to_pairs(vals)
+        # floa16/bfloat16 are stored as uint16
+        elif (data_type == TensorProto.FLOAT16
+                or data_type == TensorProto.BFLOAT16):
+            vals = np.array(vals).astype(np.float16).view(dtype=np.uint16).flatten().tolist()
         field = mapping.STORAGE_TENSOR_TYPE_TO_FIELD[
             mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE[data_type]]
         getattr(tensor, field).extend(vals)
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -240,15 +240,24 @@ def make_tensor(
         assert not raw, "Can not use raw_data to store string type"
 
     # Check number of vals specified equals tensor size
-    size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[data_type].itemsize)
+    expected_size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[data_type].itemsize)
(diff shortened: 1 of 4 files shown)


onnx/helper.py
```diff
@@ -240,15 +240,24 @@ def make_tensor(
         assert not raw, "Can not use raw_data to store string type"
 
     # Check number of vals specified equals tensor size
-    size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[data_type].itemsize)
+    expected_size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[data_type].itemsize)
+    # Flatten a numpy array if its rank > 1
+    if type(vals) is np.ndarray and len(vals.shape) > 1:
```
askhade, line 245: what is the current behavior for ndarrays?

onnx/numpy_helper.py
```diff
@@ -57,23 +57,23 @@ def to_array(tensor):  # type: (TensorProto) -> np.ndarray[Any]
         if (tensor_dtype == TensorProto.COMPLEX64
                 or tensor_dtype == TensorProto.COMPLEX128):
             data = combine_pairs_to_complex(data)
-        # F16 is stored as int32; Need view to get the original value
-        if tensor_dtype == TensorProto.FLOAT16:
+        # float16/bfloat16 is stored as int32 (uint16 type); Need view to get the original value
```
askhade, line 60: nit: structure this a little different 
```
        # float16/bfloat16 is stored as int32 (uint16 type); Need view to get the original value
        if (tensor_dtype == TensorProto.FLOAT16
                or tensor_dtype == TensorProto.BFLOAT16):
            return (
                np.asarray(
                    tensor.int32_data,
                    dtype=np.uint16)
                .reshape(dims)
                .view(np.float16))
        data = getattr(tensor, storage_field),  # type: Sequence[np.complex64]
        if (tensor_dtype == TensorProto.COMPLEX64 
          or tensor_dtype == TensorProto.COMPLEX128):
            data = combine_pairs_to_complex(data)
   
        return (
            np.asarray(
                data,
                dtype=storage_np_dtype)
            .astype(np_dtype)
            .reshape(dims)
        )
```

onnx/mapping.py
```diff
@@ -19,6 +19,7 @@
     int(TensorProto.INT64): np.dtype('int64'),
     int(TensorProto.BOOL): np.dtype('bool'),
     int(TensorProto.FLOAT16): np.dtype('float16'),
+    int(TensorProto.BFLOAT16): np.dtype('float16'),
```
askhade, line 22: add a comment explaining why you are mapping bfloat16 to np float16 something like (numpy does not support bfloat16...)

onnx/numpy_helper.py
```diff
@@ -59,27 +59,27 @@ def to_array(tensor, base_dir=""):  # type: (TensorProto, Text) -> np.ndarray[An
...
-        # Otherwise simply use astype to convert; e.g., int->float, float->float
-        else:
-            return (
-                np.asarray(
-                    data,
-                    dtype=storage_np_dtype)
-                .astype(np_dtype)
-                .reshape(dims)
-            )
+        data = getattr(tensor, storage_field),  # type: Sequence[np.complex64]
```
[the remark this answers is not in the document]
jcwchen: Yes it is incorrect... Just removed it. Thanks

onnx/test/helper_test.py
```diff
@@ -391,6 +391,30 @@ def test_make_tensor(self):  # type: () -> None
         )
         self.assertEqual(string_list, list(tensor.string_data))
 
+    def test_make_float16_tensor(self):  # type: () -> None
```
askhade, line 394: can you add a test with raw=True for both FLOAT16 and BFLOAT16

jcwchen: Added. To make it work, I made split_complex_to_pairs and numpy_array.view go into the else condition (raw=False). IIUC, the vals for raw=True should be the final binary data after conversion so no more conversion is needed inside this helper function (make_tensor).

askhade (approved):

PR #3717, opened 2021-09-14
Simplify definition of context-free function body definitions

gramalingam:
**Description**

(1) Simplify the definition of the context-free functions (LessOrEqual, DynamicQuantizeLinear, Range, MeanVarianceNormalization, HardSwish) using the parser.

(2) Extend the parser to enable this
(a) Allow reference-attributes (which are permitted in functions), and
(b) Allow untyped inputs/outputs in graphs (which are permitted in subgraphs used in loops, etc.)


onnx/defs/schema.cc
```diff
@@ -676,6 +677,11 @@ bool OpSchema::BuildContextDependentFunction(const FunctionBodyBuildContext& ctx
     return false;
 }
 
+OpSchema& OpSchema::FunctionBody(const char *func_body) {
+  OnnxParser::Parse(*function_body_.mutable_node(), func_body);
```
askhade, line 681: check status.IsOK before returning?

onnx/defs/nn/defs.cc
```diff
@@ -2466,24 +2466,21 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-             {{"Variance"}, "Sub", {"E_Xsquared", "EX_squared"}},
-             {{"STD"}, "Sqrt", {"Variance"}},
-             {{"X_variance"}, "Sub", {"X", "X_RM"}},
-             {{"Processed_STD"}, "Add", {"STD", "Epsilon"}},
-             {{"Y"}, "Div", {"X_variance", "Processed_STD"}}})));
+        .FunctionBody(R"ONNX(
+        {
+          Exponent = Constant <value = float {2.0}>()
+          Epsilon = Constant <value = float {0.000000001}>()
+          X_RM = ReduceMean <axes : ints = @axes> (X)
```
askhade, line 2473: I think we have a limitation in current code which requires function attr name to match node attr name.... can we get rid of this limitation now? If we change the attr name for function from axes to something different will it work?

gramalingam: I don't think there was a limitation in proto. The builder utility earlier had a limitation (that is, it had only one name parameter that it used for both). Even that was subsequently removed. See https://github.com/onnx/onnx/blob/a5e7ee51176bf78a60c118758174e13d85a87b46/onnx/defs/attr_proto_util.h#L45 and https://github.com/onnx/onnx/blob/a5e7ee51176bf78a60c118758174e13d85a87b46/onnx/defs/attr_proto_util.h#L54 ... unless I misunderstood your question.

gramalingam: On a different note: I don't think the "type" field is required in an attribute-reference ... however, all the examples I saw set the type, so I did the same to not change the behavior. But it would be useful to check and clarify whether this field is required or not. I don't think it is used. It can lead to inconsistent models (where the actual attribute is of type int and the reference says float etc., which we don't check for, I think.) It is simpler to pick up the type from the attribute passed into the function.

askhade: I don't remember what was the exact issue this was at least 2 years back and since then we have made quite a bit of improvements in this area.

I agree adding a type here can lead to inconsistent models. Can we move to picking the type from attr instead?

onnx/defs/schema.cc
gramalingam: Done.

onnx/defs/nn/defs.cc
gramalingam: Not sure I understood "pick the type from attr instead". I think we are doing that already. I think we just need to clarify in the spec that setting the type for this case in not needed (and that this type is ignored). And decide what to do in the checker. And update the current uses to drop this. But the issue is compatibility, if other users have custom functions that do this. In any case, I think this is a separate discussion, and we should consider it separately from the current PR.

gramalingam: I tried removing the " : ints". However, the checker is currently checking this type, so this causes a failure. I am rolling back that change, and adding the ": ints" back.

askhade (approved):

Issue #3696, opened 2021-09-02
How to remove warning from shape_inference


mengniwang95:
# Ask a Question

### Question
![image](https://user-images.githubusercontent.com/65003808/131789984-b923d6eb-ef50-4332-a158-e3a3ff5062f2.png)
When I use python API shape_inference, there will be many warning infos, how can I remove these warnings?

jcwchen:
Hi @mengniwang95,
Since FusedConv is not an official ONNX op (it should a contrib_ops from ONNXRuntime), there is no schema definition for it in ONNX and shape_inference will throw this warning and stop shape inference anyway. This is an expected warning and you should probably just use onnxruntime.InferenceSession because onnx.shape_inference does not work with a model consists of unofficial ops. Thanks.

askhade:
@jcwchen: One of the problem is we are writing to std::cerr which is not desirable for all users. We should fix this... others have reported this too example: https://github.com/onnx/onnx/issues/2239

fdwr:
Yeah, writing directly to `std::cerr` is in appropriate for a low-level library, as it's the prerogative of the *calling* library (which actually does properly handle this in the case of ONNX Runtime) or of the application to emit user-visible messages.

jcwchen:
IIUC, after this PR: https://github.com/onnx/onnx/pull/3722 (which is included by ONNX 1.11) was merged, such a warning by std:cerr won't be thrown anymore. That is to say, if there is an unsupported operator, it will just stop shape inference and be silent. I will close this issue now. Thanks.

PR #3718, opened 2021-09-14
add shape inference for model local functions

askhade:

**Description** Add shape inference for model local functions. Also includes some name formatting.
- Describe your changes.
For schema less model local functions where a type and shape inference function is not available, fetch the function proto from model local functions container and traverse the function body to infer the shapes for the function node.

Adding tests

**Motivation and Context**


onnx/shape_inference/implementation.cc
```diff
@@ -223,245 +193,283 @@ void mergeShapesAndTypes(const TypeProto& inferredType, TypeProto* existingType)
...
+      fail_type_inference("Cannot infer type and shape for node name ", n.name(), ". No opset import for domain",
+          n.domain(), " optype ", n.op_type());
     }
     auto domain_version = dit->second;
     const auto schema = schema_registry->GetSchema(n.op_type(), domain_version, n.domain());
-    InferenceContextImpl ctx(n, valueTypesByName, inputDataByName,inputSparseDataByName,
-      &generatedShapeDataByName, &graphInferenceContext);
-    if (!schema) {
-      std::cerr << "Warning: Unsupported operator " << n.op_type() << ". No schema registered for this operator."
-                << std::endl;
```
jcwchen, line 362: So ONNX finally decides not to print such warnings anymore (due to https://github.com/onnx/onnx/issues/3696 and https://github.com/onnx/onnx/issues/2239)?

I am afraid without those warnings, users will complain the shape inference is incomplete and there is no any warning/error... Could ONNX at least throw a warning about unsupported error when enabling strict_mode?

jcwchen: Or perhaps ONNX can add a new "mode" for shape_inference to detect such warning/error (unsupported schema)

askhade:
Closing this in favor of #3722. I mistakenly created this branch in onnx instead of using a fork.

PR #3719, opened 2021-09-15
Fix bug in pad converter, value only needed in constant mode.

LiuYu198701:
Signed-off-by: Yu Liu <yu.liu@enflame-tech.com>

- fix bug in pad converte, value only needed in constant mode

### DIFF
--- a/onnx/version_converter/adapters/pad_10_11.h
+++ b/onnx/version_converter/adapters/pad_10_11.h
@@ -27,15 +27,17 @@ class Pad_10_11 final : public Adapter {
       node->addInput(v_pads);
       node->removeAttribute(kpads);
       // Turn value attribute into input
-      Tensor t_value;
-      t_value.elem_type() = TensorProto_DataType_FLOAT;
-      auto& data_value = t_value.floats();
-      data_value.emplace_back(node->f(kvalue));
-      Node* constant = graph->create(kConstant);
-      constant->insertBefore(node);
-      constant->t_(kvalue, t_value);        
-      node->addInput(constant->output());
-      node->removeAttribute(kvalue);
+      if (node->s(kmode) == "constant") {
+        Tensor t_value;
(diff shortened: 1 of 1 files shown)


onnx/version_converter/adapters/pad_10_11.h
```diff
@@ -27,15 +27,17 @@ class Pad_10_11 final : public Adapter {
...
-      Tensor t_value;
-      t_value.elem_type() = TensorProto_DataType_FLOAT;
-      auto& data_value = t_value.floats();
-      data_value.emplace_back(node->f(kvalue));
-      Node* constant = graph->create(kConstant);
-      constant->insertBefore(node);
-      constant->t_(kvalue, t_value);        
-      node->addInput(constant->output());
-      node->removeAttribute(kvalue);
+      if (!node->hasAttribute(kmode) || node->s(kmode) == "constant") {
```
askhade, line 30: should this be node->hasAttribute(kmode) && || node->s(kmode) == "constant"
Can you elaborate a little more about this check. Thanks!

askhade (changes requested):
Thanks for the PR. I added a comment please take a look.

onnx/version_converter/adapters/pad_10_11.h
LiuYu198701: the default mode is 'constant' and _node->hasAttribute(kmode)_ is false
so, if **node->hasAttribute(kmode) == false**, it is a DEFAULT 'constant' mode

gramalingam (approved):

onnx/version_converter/adapters/pad_10_11.h
gramalingam: I wonder if the IR automatically takes care of returning the default-value? If not, then shouldn't we also check if the "value" attribute is present down below, before using?

askhade (approved):

PR #3721, opened 2021-09-15
Add documentation for symbolic shape inf and partial data propagation

askhade:

**Description**
- Adding the proposal for symbolic shape inference. This proposal was already accepted and has been implemented in onnx 1.10 adding this for documenting the proposal in the repo.

**Motivation and Context**

### DIFF
--- a/docs/proposals/SymbolicShapeInfProposal.md
+++ b/docs/proposals/SymbolicShapeInfProposal.md
@@ -1,0 +1,163 @@
+# Proposal - Symbolic Shape Inference And Partial Data Propagation  
+
+*Note: This proposal was accepted and implemented in ONNX 1.10. Following PRs implemented this proposal: 3518, 3551, 3593, 3580*
+
+## Introduction 
+ONNX provides an implementation of shape inference on ONNX graphs. Shape inference is computed using the operator level shape inference functions. The inferred shape of an operator is used to get the shape information without having to launch the model in a session. Such static shape inference can be used to catch obvious errors before runtime, eliminate run-time checks which are otherwise guaranteed to pass, improve static memory planning and improve model visualization experience. For pytorch exporter and compiler-based execution providers like Nuphar, shape inference is required (rank inference is minimum requirement), and they cannot work with unknown shapes. 
+
+This document explains the limitations of shape inference and lays out a proposal for addressing these limitations.  
+
+## Current onnx shape inference limitations (Pre ONNX 1.10) 
+Today, ONNX shape inference is not guaranteed to be complete. Wherever possible we fall back to rank inference however, there are scenarios when rank inference is not possible either. Here are the various limitations which block the completion of shape inference: 
+
+* Some dynamic behaviors block the flow of shape inference, and the shape inference stops. For example, reshape to a dynamically computed shape.
+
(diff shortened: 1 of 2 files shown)


gramalingam (approved):

docs/proposals/SymbolicShapeInfProposal.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+### Partial data computation and propagation 
+When shape inputs are computed dynamically, shape inference post a reshape node stops. This can be prevented by making this data available to the reshape node during shape inference. We propose computation and propagation of data for operators which are used in shape computation. 
+
+It is called “partial” data computation and propagation because this will only be done for shape computations. It is not meant to be a full-fledged kernel for the operator. For the same reasons data computations will be implemented for a limited set of operators. While we will increase the coverage in the future iterations it is important to note that for some operators like LSTM, convolution ops, pooling ops etc. this will never be added because such ops are not used in shape computations.
```
jcwchen, line 76:
```suggestion
It is called “partial” data computation and propagation because this will only be done for shape computations. It is not meant to be a full-fledged kernel for the operator. For the same reasons data computations will be implemented for a limited set of operators. While we will increase the coverage in the future iterations it is important to note that for some operators like LSTM, convolution ops, pooling ops etc. These will never be added because such ops are not used in shape computations. 
```

docs/proposals/SymbolicShapeInfProposal.md
```diff
@@ -0,0 +1,163 @@
...
+If we have a broadcast between two unknown dimensions “M” and “N” we cannot infer that both M and N should have the same value. The runtime semantics allows for one of the two symbols to have the value 1 and the other to have a value different from 1. So, merging M and N and treating them as the same value is potentially unsound. In this case, a new symbol will be generated for the output shape and the shape inference will continue.  
+
+
+### Inferred shape does not match output shape 
+Inferred and existing shapes can be mismatched. Although failing shape inference in such cases seems like the correct approach it may not always be practical. By default, shape inference will fail when such a case is encountered however callers will have an option to override existing types with inferred types. When this option is enabled, shape inference will continue with the inferred type.  
+
+
+### Handling symbolic dimensions with data propagation 
+When the shape contains symbolic dimensions, we try and propagate them downstream, however in cases where some arithmetic operations are performed on these symbolic dims we create a new symbols and propagate them instead.
```
jcwchen, line 159:
```suggestion
When the shape contains symbolic dimensions, we try and propagate them downstream, however in cases where some arithmetic operations are performed on these symbolic dims we create new symbols and propagate them instead.   
```

jcwchen:
Thanks for the effort! Two more things:
- Since ONNX has improved its shape inference, shall we further mention this document in the original shape inference document (limitation section): https://github.com/onnx/onnx/blob/master/docs/ShapeInference.md#limitations?
- Shall we add a brief demo about how to enable [data propagation](https://github.com/onnx/onnx/blob/a5e7ee51176bf78a60c118758174e13d85a87b46/onnx/shape_inference.py#L39)? We can mention it either in this new data propagation doc or [PythonAPIOverview.md](https://github.com/onnx/onnx/blob/master/docs/PythonAPIOverview.md)

docs/proposals/SymbolicShapeInfProposal.md
askhade: this refers to data propagation function. Updated to use "data propagation function"

askhade:
> Shall we add a brief demo about how to enable data propagation? We can mention it either in this new data propagation doc or PythonAPIOverview.md

Yes lets add it to backlog items.

PR #3729, opened 2021-09-21
Fix bfloat16 issue: remove it from numpy to TensorProto type map

jcwchen:
**Description**
(Updated)
Fix an issue that this PR brought: https://github.com/onnx/onnx/pull/3716: hardcode the map of (numpy to TensorProto type) as np.dtype('float16') to TensorProto.FLOAT16.

**Motivation and Context**
tensor_type to numpy [map](https://github.com/onnx/onnx/blob/316c9dc52916219316f51b3852a456b4ffcd97ab/onnx/mapping.py#L12) should be 1:1; otherwise ONNX does not know a float16 array comes from float16 or bfloat16... At this moment native numpy does not support bfloat16 so right now a numpy float16 will only reversed to TensorProto.FLOAT16.


onnx/mapping.py
```diff
@@ -29,6 +29,9 @@
 }
 
 NP_TYPE_TO_TENSOR_TYPE = {v: k for k, v in TENSOR_TYPE_TO_NP_TYPE.items()}
+# Currently native numpy does not support bfloat16
+# Numpy float16 array is only reversed to TensorProto.FLOAT16
+NP_TYPE_TO_TENSOR_TYPE[np.dtype('float16')] = int(TensorProto.FLOAT16)
```
askhade, line 34: this seems hacky... can you simply remove bfloat16 from the map?
You can take care of bfloat16 type in the calling code...

askhade: This is where you need bfloat16 support from this map right:
expected_size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[data_type].itemsize)

You can handle bfloat16 separately in this code...

jcwchen: I thought about that as well, but actually helper make_tensor is not the only function uses mapping. If we decide to add bfloat16 condition whenever using mapping, there will be quite a few places need this update. (`helper.py`, `numpy_helper.py`, `test/case/code/node/castlike.py`, `test/case/code/node/cast.py`, `test/case/node/__init__.py`) It will be hard to maintain in the future (when new functions need to know the mapping relationship for bfloat16 or bfloat16 is eventually supported by numpy) and also seem hacky...

Merely update mapping.py at least people will clearly know the relationship that
- TensorProto.FLOAT16 -> numpy.float16
- TensorProto.BFLOAT16 -> numpy.float16
- numpy.float16 -> TensorProto.FLOAT16 (not TensorProto.BFLOAT16)

Please let me know if you still think handle bfloat16 separately in other codes is better. Thanks!

askhade: You add "int(TensorProto.BFLOAT16): np.dtype('float16')," to the map only in your last PR right?

gramalingam: Regardless of the number of places where it is used, the important thing is to understand *why* it is used in all those places to make sure the right thing is happening in all those places. And see if we can have different methods that are less likely to be wrongly used in the future (e.g., a "sizeof" method would be a better way to encapsulate its use to determine size).

jcwchen: After discussion, we decided to add a if condition to exclude TensorProto.BFLOAT16 while constructing NP_TYPE_TO_TENSOR_TYPE. It looks clearer for future reference and less hacky.

askhade (approved):

onnx/mapping.py
```diff
@@ -28,7 +28,12 @@
     int(TensorProto.STRING): np.dtype('object')
 }
 
-NP_TYPE_TO_TENSOR_TYPE = {v: k for k, v in TENSOR_TYPE_TO_NP_TYPE.items()}
+NP_TYPE_TO_TENSOR_TYPE = {}
+for tensor_type, numpy_type in TENSOR_TYPE_TO_NP_TYPE.items():
+    # Currently native numpy does not support bfloat16 so TensorProto.BFLOAT16 is ignored for now
+    # Numpy float16 array is only reversed to TensorProto.FLOAT16
+    if tensor_type != TensorProto.BFLOAT16:
+        NP_TYPE_TO_TENSOR_TYPE[numpy_type] = tensor_type
```
gramalingam, line 36: minor nit/suggestion: I think you can just add the if condition to the original construct as below:
```py
   NP_TYPE_TO_TENSOR_TYPE = {v: k for k, v in TENSOR_TYPE_TO_NP_TYPE.items() if k != TensorProto.BFLOAT16}
```

jcwchen: Looks good... I will update in another PR: https://github.com/onnx/onnx/pull/3674 Thanks!

PR #3731, opened 2021-09-22
Support read optional type pb data in test case

chudegao:
**Description**
support read optional type pb data in test case

**Motivation and Context**
When run backend node test with onnx-tf for optional operators, it can not load pb data. Check the runner, it does not handle optional data type. 
And function to_optional() in numpy_helper.py will hit error if optional data is empty.


onnx/numpy_helper.py
```diff
@@ -289,6 +289,8 @@ def to_optional(optional):  # type: (OptionalProto) -> Optional[Any]
     opt = None  # type: Optional[Any]
     elem_type = optional.elem_type
     value_field = mapping.OPTIONAL_ELEMENT_TYPE_TO_FIELD[elem_type]
+    if not optional.HasField(value_field):
```
gramalingam, line 292: The previous implementation does seem incomplete/broken, thanks for catching it. In particular, it does not handle two cases properly, the case where elem_type is UNDEFINED and the case where elem_type is OPTIONAL. The representation does not seem well-documented. In particular, what is elem_type supposed to be for a None value? I think elem_type should be UNDEFINED (rather than OPTIONAL). My guess is that there are two issues, requiring two different fixes. First is to add
```
   if elem_type == OptionalProto.UNDEFIND:
      return None
```
_before_ line 291. The second is to add
```
   elif elem_type == OptionalProto.OPTIONAL
      return to_optional(value)
```
before line 300/302. The proposed fix makes me wonder if we are choosing to represent a None value by setting the elem_type to OPTIONAL and leaving the optional field empty, which seems a bit problematic.

chudegao: Thanks for comments. I added the two fixes in comments and still keep the original fix as there's case that elem_type is tensor/seq/map but no value filed.

>>> proto_filename='onnx/backend/test/data/node/test_optional_has_element_empty/test_data_set_0/input_0.pb'
>>> f=open(proto_filename, 'rb')
>>> protobuf_content = f.read()
>>> optional = onnx.OptionalProto()
>>> optional.ParseFromString(protobuf_content)
18
>>> optional
name: "optional_input"
elem_type: 1

>>>

gramalingam: Thanks for the test-case where it fails. I think the problem is in the creation of the test-case data (in the function ```from_optional``` below). I think we should add the following before line 335/341 (old and new line numbers):
```
   elif (opt is None)
      elem_type = OptionalProto.UNDEFINED
```
I think this should also be cleaned up more (to omit the parameter dtype), but it can be done separately, assuming that the other fixes work and do the right thing.

chudegao: Yes, it's caused from_optional not set elem_type as ptionalProto.UNDEFINED. I added it as your comment and update the checker to skip checking when elem_type is OptionalProto.UNDEFINED.

gramalingam (approved):

PR #3734, opened 2021-09-23
Enable Python 3.10 support/add Python 3.10 in release CIs

jcwchen:
**Description**
- Add Python 3.10 in release CIs to test/produce py 3.10 wheel
- Fix issues caused by Python 3.10

**Motivation and Context**
ONNX should support 3.10 next release


.github/workflows/win_no_exception_ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10.0-rc.2']
```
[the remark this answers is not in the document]
jcwchen: Python 3.10 support is still experimental in GitHub Action... 3.10 is not found

snnn:
BTW, I just found the latest protobuf release doesn't work.  It needs this fix: https://github.com/protocolbuffers/protobuf/pull/9437 

It doesn't impact your this change.  Just FYI.

.github/workflows/manylinux/test_package_i686.sh
```diff
@@ -36,17 +34,13 @@ $PYTEST_COMMAND
...
 # onnx.checker all generated backend data
-$PYTHON_COMAND onnx/backend/test/cmd_tools.py generate-data
-$PYTHON_COMAND workflow_scripts/test_generated_backend.py
-
-# Verify ONNX with the latest numpy
-if [ "$PY_VERSION" == "3.8" ] || [ "$PY_VERSION" == "3.9" ]; then
-    $PIP_UNINSTALL_COMMAND numpy onnx && $PIP_INSTALL_COMMAND numpy==1.21.5
-else
-    $PIP_UNINSTALL_COMMAND numpy onnx && $PIP_INSTALL_COMMAND numpy
+# FIXME: Python 3.10 wheel runs forever during generate-data; skip for now
```
jcwchen, line 37: https://github.com/numpy/numpy/issues/20277 NumPy didn't provide prebuilt Python 3.10 i686 wheel. Since now ONNX uses NumPy 1.21.5 in general, it needs to build NumPy from source for Python 3.10, which is probably not a good idea. Besides, this built Python 3.10 i686 ONNX wheel seems problematic: it hangs a long time while generating data. It took me quite a while to investigate what the root cause is, but I still don't know why.

And in newer NumPy (>1.22), they even don't provide i686 wheel for all Python versions anymore. I think we should reconsider to deprecate manylinux i686 packages since people are retiring them and the release CI for i686 will be unstable. If we decide to do so, it needs to be done early before next release to let users be aware.

jcwchen: Created an issue https://github.com/onnx/onnx/issues/4049 for tracking. I tend to skip Python 3.10+i686 for now in this PR and then we can at least let CIs cover Python 3.10 in other environments.

liudonghua123:
Recently I install onnx in my windows python 3.10 x64, I have to build protobuf before build onnx locally. I would like to install via a prebuild whl file. 😄

.github/workflows/release_mac.yml
```diff
@@ -106,7 +106,7 @@ jobs:
         pytest
 
     - name: Verify ONNX with ort-nightly
-      if: matrix.python-version != '3.6' # Do not test ort-nightly for Python 3.6 due to missing (unsupported) package
+      if: matrix.python-version != '3.10' # FIXME Do not test ort-nightly for Python 3.10 due to missing (unsupported) package
```
gramalingam, line 109: What's the FIXME? Is there a reason 3.10 cannot be tested here? Thanks!

jcwchen: Sorry for the confusion. The reason is ort-nightly has not supported Python 3.10 yet: https://test.pypi.org/project/ort-nightly/#files. I think I should use TODO instead, which is more proper. Just updated. Please review it again. Thank you!

gramalingam (approved):

jcwchen:
@liudonghua123 FYI you will see Python 3.10 wheel [here](https://test.pypi.org/project/onnx-weekly/) from onnx-weekly package after this weekend. Have fun!

PR #3738, opened 2021-09-24
Extend Where op to permit bfloat16 types

gramalingam:
**Description**
Update the Where op to allow bfloat type.

**Motivation and Context**
Looks like this op was overlooked when bfloat type was added.


askhade:
Is BFLoat16 support required for this OP? Meaning is any runtime blocked currently or is this simply to close the gap?

PR #3741, opened 2021-09-27
Add Signal Operator definitions

### LINKED ISSUE #3573: Operator for fft_rfft2.                ( Exporting the operator fft_rfft2 to ONNX opset version 11 is not supported )
# New Operator

I was implementing this new paper - [Global Filter Networks for Image Classification](https://arxiv.org/abs/2107.00645)
([GitHub](https://github.com/raoyongming/GFNet)) which uses 2D Discrete Fourier Transform and 2D Inverse Fourier transform but it seems that the current onnx version does not have support for this. It might be a huge plus point to include this operator as the use of FFTs is increasing steadily and the result seem pretty promising! considering the runtime-accuracy tradeoff

smk2007:
This PR adds signatures for the following signal operator definitions.
- **DFT**
  PyTorch: https://pytorch.org/docs/stable/generated/torch.fft.fft.html
  TensorFlow: https://www.tensorflow.org/api_docs/python/tf/signal/fft
- **IDFT**
  PyTorch: https://pytorch.org/docs/stable/generated/torch.fft.ifft.html
  TensorFlow: https://www.tensorflow.org/api_docs/python/tf/signal/ifft
- **STFT**
  PyTorch: https://pytorch.org/docs/stable/generated/torch.stft.html
  TensorFlow: https://www.tensorflow.org/api_docs/python/tf/signal/stft
- **ISTFT**
  PyTorch: https://pytorch.org/docs/stable/generated/torch.istft.html
  TensorFlow: https://www.tensorflow.org/api_docs/python/tf/signal/inverse_stft
- **HannWindow**
  PyTorch: https://pytorch.org/docs/stable/generated/torch.hann_window.html
  TensorFlow: https://www.tensorflow.org/api_docs/python/tf/signal/hann_window
- **HammingWindow**
  PyTorch: https://pytorch.org/docs/stable/generated/torch.hamming_window.html
  TensorFlow: https://www.tensorflow.org/api_docs/python/tf/signal/hamming_window
- **BlackmanWindow**
  PyTorch: https://pytorch.org/docs/stable/generated/torch.blackman_window.html
  TensorFlow: _none_
- **MelSpectrogram**
  PyTorch: https://pytorch.org/audio/stable/transforms.html?highlight=melscale#torchaudio.transforms.MelScale
  TensorFlow: https://www.tensorflow.org/api_docs/python/tf/signal/linear_to_mel_weight_matrix

Fixes #3573


onnx/defs/math/defs.cc
```diff
@@ -3660,4 +3660,534 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+    default:
+      fail_shape_inference("Unsupported input data type of ", data_type);
+  }
+}
+
+static const char* DFT_ver13_doc =
+    R"DOC(Computes the discrete Fourier transform of input.)DOC";
+
+    ONNX_OPERATOR_SET_SCHEMA(DFT,
+        13,
```
jcwchen, line 3690: The latest opset version is 16 so please replace all 13 with 16.

You need to run `tools/update_doc.sh` or `tools/update_doc.bat` to get updated documents (Changelog.md, Operators.md, TestCoverage.md) and please include them in this PR. Thanks!

jcwchen (commented):
Thank you for proposing this! Three more things --
1. Don't forget to signoff your commits since DCO is needed in ONNX
2. Add test scripts as `onnx/backend/test/case/node/XXX.py` and models with input/output in `onnx/backend/test/data/node/test_XXX/`
3. Add simple version converter tests for new ops in `onnx/test/automatic_upgrade_test.py `. You can check this PR as an example: https://github.com/onnx/onnx/pull/3558/files

FYI: more details in this document: https://github.com/onnx/onnx/blob/master/docs/AddNewOp.md

onnx/defs/math/defs.cc
```diff
@@ -3660,4 +3660,534 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                OpSchema::Single,
+                true,
+                1,
+                OpSchema::NonDifferentiable)
+            .Output(
+                0,
+                "output",
+                "The discrete Fourier transform of input. "
+                "If onesided is 1, the output has the shape: [batch_size][floor(n_fft/2)+1][2]. " 
+                "If onesided is 0, the output has the shape: [batch_size][n_fft][2]"
```
gramalingam, line 3720: Adding a single line to indicate the order of elements for [n_fft] may be helpful, just like in line 3695 above.

onnx/defs/math/defs.cc
```diff
@@ -3660,4 +3660,534 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                // To support multidimensional DFT the signal_ndim may != 1,
+                // but this attribute is not included on the schema yet
+                const auto signal_ndim = 1;
+
+                // Add 1 dimensions to the rank for the batch size dimension
+                auto real_inputs_ndim = signal_ndim + batch_ndim;
+                // Add 2 dimensions to the rank for the batch size dimension AND component dimension
+                auto complex_inputs_ndim = signal_ndim + batch_ndim + component_ndim;
+
+                auto& input_shape = getInputShape(ctx, 0);
```
gramalingam, line 3742: Best to make all the code from here conditional on ```hasInputShape(ctx,0)```, and check that the shape has the expected dimensions.

onnx/defs/math/defs.cc
```diff
@@ -3660,4 +3660,534 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                "If onesided is 0, the output has the shape: [batch_size][n_fft][2]"
+                "For all types of input, the last dimension of the output represents the components of a complex number.",
+                "T")
+            .TypeConstraint(
+                "T",
+                {"tensor(float16)", "tensor(float)", "tensor(double)"},
+                "Constrain input and output types to float tensors.")
+            .TypeAndShapeInferenceFunction([](ONNX_NAMESPACE::InferenceContext& ctx) {
+                propagateElemTypeFromInputToOutput(ctx, 0, 0);
+                // To support multidimensional DFT the signal_ndim may != 1
```
gramalingam, line 3729: Is the goal to support multidimensional? If so, my recommendation would be an attribute to indicate if the input is complex or real (and then the number of dimensions is automatically determined by the input rank).

onnx/defs/math/defs.cc
```diff
@@ -3660,4 +3660,534 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                "T",
+                OpSchema::Single,
+                true,
+                1,
+                OpSchema::NonDifferentiable)
+            .Output(
+                0,
+                "output",
+                "The inverse 1-dimensional discrete Fourier transform of the input. "
+                "The output tensor has the same shape and type as the input.",
```
gramalingam, line 3793: This does not seem consistent with the shape-inference function below, which may add an extra dimension for complex output.

onnx/defs/math/defs.cc
```diff
@@ -3660,4 +3660,534 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                   1,
+                   OpSchema::NonDifferentiable)
+            .TypeConstraint(
+                "T1",
+                {"tensor(int64)"},
+                "Constrain the input size to int64_t.")
+            .TypeConstraint(
+                "T2",
+                {"tensor(float)",
+                 "tensor(float16)",
```
gramalingam, line 3873: Is there a reason for not including bfloat16? Suggest also replacing this with one of the predefined sets, like ```all_numeric_types_with_bfloat()``` : https://github.com/onnx/onnx/blob/5282dd787b2c913110c26f34de7b624cf7ab9c01/onnx/defs/schema.h#L633

gramalingam:
It would be helpful to add the motivation for these ops (eg., if these are commonly required for audio processing).

onnx/defs/math/defs.cc
```diff
@@ -3660,4 +3660,534 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                } else {
+                  fail_shape_inference(
+                        "The input_shape must [batch_size][n_fft] for real values or [batch_size][n_fft][2] for complex values.")
+                }
+
+                updateOutputShape(ctx, 0, result_shape);
+            }));
+
+static const char* HannWindow_ver13_doc =
+    R"DOC(Generate a Hann window.)DOC";
```
gramalingam, line 3836: Can we add any link to help explain what a Hann window is?

gramalingam: If all the window ops are mostly identical, I would suggest factoring it out to avoid duplication. This will help making updates/changes while keeping them in sync. (E.g., see https://github.com/onnx/onnx/blob/5282dd787b2c913110c26f34de7b624cf7ab9c01/onnx/defs/math/defs.cc#L53 )

napohou:
any update?

onnx/defs/math/defs.cc
wuerflts: Please extend it to multi-dimensional now before adding it to ONNX. Otherwise image processing applications like magnetic resonance image reconstruction (https://fastmri.org/) will have to resort to weird workarounds which will break in upates.

wuerflts:
Besides audio processing the FFT is an integral component of MRI reconstruction. See this open challenge by NYU / Meta: https://fastmri.org/
One important caveat is that the N-dimensional FFT is needed for those applications which can be implemented by successive 1D FFTs. However, it would make for a cleaner interface in ONNX to simply define only one Operator which is N-dimensional by default since tensorflows and pytorchs special 1D, 2D, 3D and ND cases can be represented by that. This would provide a more concise ONNX graph.

onnx/backend/test/data/real/test_bvlc_alexnet/data.json
```diff
@@ -1 +1 @@
-{"atol": 1e-07, "model_name": "bvlc_alexnet", "rtol": 0.001, "url": "https://s3.amazonaws.com/download.onnx/models/opset_9/bvlc_alexnet.tar.gz"}
\ No newline at end of file
+{"atol": 1e-07, "model_name": "bvlc_alexnet", "rtol": 0.001, "url": "https://s3.amazonaws.com/download.onnx/models/opset_10/bvlc_alexnet.tar.gz"}
```
askhade, line 1: what are these changes for?

napohou:
Hi @smk2007 
Did you implement fft and ifft in your code of https://github.com/smk2007/onnx? If so, which branch should we use, master or user/sheilk/signal_ops?

piEsposito:
Are you also adding the N-dimensional FFT? This was not clear per the discussion, and @wuerflts has some good points.

garymm:
@xadupre do you think any of these can be expressed as functions (similar to what you did in https://github.com/onnx/tensorflow-onnx/pull/1371)?

xadupre:
@garymm they can be expressed as functions. One question before that is about complex. Right now, ONNX operators do not fully support complex types (Add, Sub, ...). It is possible to do without, that's what we did in tf2onnx but the implementation would be more simple with than without.

garymm:
Let's discuss with @gramalingam on Monday. Perhaps we should wait until ONNX Script is ready and then we can use it to define these functions.

onnx/defs/math/defs.cc
```diff
@@ -3412,4 +3412,594 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .FunctionBody(R"ONNX(
+        {
+          A0 = Constant <value = float {0.54347826087}>()
+          A1 = Constant <value = float {0.45652173913}>()
+          A2 = Constant <value = float {0.0}>()
+          Zero = Constant <value = float {0.0}>()
+          One = Constant <value = float {1.0}>()
+          Two = Constant <value = float {2.0}>()
+          Tau = Constant <value = float {6.2831853}>()
+          AngularIncrement = Div (Tau, size)
```
xadupre, line 3693: size is expected to be a float here. I assume size is the only input defined as an integer.

docs/Changelog.md
```diff
@@ -20366,6 +20600,50 @@ This version of the operator has been available since version 16 of the default
...
+
+#### Version
+
+This version of the operator has been available since version 16 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>onesided</tt> : int (default is 0)</dt>
+<dd>If onesided is 1, only values for w in [0, 1, 2, ..., floor(n_fft/2) + 1] are returned because the real-to-complex Fourier transform satisfies the conjugate symmetry, i.e., X[m, w] = X[m,w]=X[m,n_fft-w]*. Note if the input or window tensors are complex, then onesided output is not possible. Enabling onesided with real inputs performs a Real-valued fast Fourier transform (RFFT).When invoked with real or complex valued input, the default value is 0. Values can be 0 or 1.</dd>
```
fdwr, line 20615: (nit) Space after "(RFFT).When".

garymm:
Can you please update the PR description to include `Fixes #3741`?

ternaus:
@smk2007 Any updates? This PR should unlock a set of very important models to be converted to ONNX.

docs/Changelog.md
fdwr: https://github.com/smk2007/onnx/pull/1

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,591 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                "Note if the input or window tensors are complex, then onesided output is not possible. "
+                "Enabling onesided with real inputs performs a Real-valued fast Fourier transform (RFFT)."
+                "When invoked with real or complex valued input, the default value is 0. "
+                "Values can be 0 or 1.",
+                AttributeProto::INT,
+                static_cast<int64_t>(0))
+            .Attr(
+                "axis",
+                "The axis on which to perform the DFT. By default this value is set to 0, which corresponds to the first dimension after the batch index."
+                "This value must be less than signal_dimN, where signal_dimN is the number of dimensions in the signal.",
```
gramalingam, line 3438: Ideally, it should allow negative values, just like other ops.

gramalingam, line 3438: Ideally, it should allow negative values, just like other ops.

onnx/defs/math/defs.cc
gramalingam: Hi, the shape inference code has issues that must be addressed, since we cannot assume that the rank/dimensions of an input are known during shape-inference time. Please see this PR who updated documentation on shape-inference to help clarify a few things. Please do let me know if you have any questions/suggestions. Thanks!

gramalingam: Forgot the PR link: https://github.com/onnx/onnx/pull/4163

onnx/defs/math/defs.cc
xadupre: I would choose to support multidimensional. We need to choose how to handle complex. Most of these operators return complex. Three options for the return: complex tensor, real tensor + 1 dimension, 2 tensors (real + imaginary parts). Using complex seems the best choice but we need to add new operators (real part, imaginary part, conj). Otherwise we may keep real for now and extend it later to complex.

gramalingam: @wuerflts : do you see a need for 2D or 3D, or generic N-D? Do you favor separate ops, or a single op for all dimensions?

postrational:
Shouldn't the window functions have the `periodic` attribute to match the PyTorch and Tensorflow implementations?

postrational:
I believe the `ISTFT` is currently missing from the Pull Request.

smk2007:
working on an update this week.

---
In reply to: [964875867](https://github.com/onnx/onnx/pull/3741#issuecomment-964875867) [](http://example.com/codeflow?ancestors=964875867,964875867)

smk2007:
Hi. To accomodate the n-dimensional FFT I have added the axis attribute to allow chaining 1d FFTs.
While the graph may be cleaner with a single nFFT, I think that this can be easily handled by converters as most do not author onnx graphs from scratch.
Unless there is a pressing performance reason to rationalize nFFT as a single operator, I am postpone nDFT to a later time when needed.

---
In reply to: [970120447](https://github.com/onnx/onnx/pull/3741#issuecomment-970120447) [](http://example.com/codeflow?ancestors=970120447,970120447)

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,591 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                auto has_component_dimension = dim_size > 2; 
+
+                ONNX_NAMESPACE::TensorShapeProto result_shape_proto = input_shape;
+                
+                bool axis = static_cast<bool>(getAttribute(ctx, "axis", 0));
+                bool is_onesided = static_cast<bool>(getAttribute(ctx, "onesided", 0));
+                if (is_onesided) {
+                    // Since signal_ndim = 1, and multidimensional DFT is not supported,
+                    // only the single signal dim (1) needs to be updated
+                    auto n_fft = input_shape.dim(1 + axis).dim_value();
```
gramalingam, line 3485: Here, the specific `dim_value` may not be known.

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,591 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                OpSchema::NonDifferentiable)
+            .TypeConstraint(
+                "T",
+                {"tensor(float16)", "tensor(float)", "tensor(double)", "tensor(bfloat16)"},
+                "Constrain input and output types to float tensors.")
+            .TypeAndShapeInferenceFunction([](ONNX_NAMESPACE::InferenceContext& ctx) {
+                propagateElemTypeFromInputToOutput(ctx, 0, 0);
+                const int64_t batch_ndim = 1;
+                
+                auto& input_shape = getInputShape(ctx, 0);
```
gramalingam, line 3545: Here too: must check `hasInputShape` first.

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,591 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                // but this attribute is not included on the schema yet
+                const auto signal_ndim = 1;
+
+                // Add 1 dimensions to the rank for the batch size dimension
+                auto real_inputs_ndim = signal_ndim + batch_ndim;
+                // Add 2 dimensions to the rank for the batch size dimension AND component dimension
+                auto complex_inputs_ndim = signal_ndim + batch_ndim + component_ndim;
+
+                // Get inputs
+                auto& input_shape = getInputShape(ctx, 0);
```
gramalingam, line 3911: Check `hasInputShape`

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,591 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                if (window_input == nullptr && frame_length_input == nullptr)
+                {
+                    fail_type_inference("STFT expects to have at least one of these inputs set: [window, frame_length].");
+                } else if (window_input != nullptr && frame_length_input != nullptr)
+                {
+                    if (window_input->dim_size() != 1)
+                    {
+                        fail_type_inference("STFT's window input, must have rank = 1.");
+                    }
+                    auto window_length = window_input->dim(0).dim_value();
```
gramalingam, line 3945: `dim_value` may be unknown here

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,591 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                }
+
+                bool is_onesided = static_cast<bool>(getAttribute(ctx, "onesided", 0));
+                if (is_onesided) {
+                    dft_size = is_onesided ? ((dft_size >> 1) + 1) : dft_size;
+                }
+
+                // Since signal_ndim = 1, and multidimensional DFT is not supported,
+                // only the single signal dim (1) needs to be updated
+                auto signal_size = input_shape.dim(1).dim_value();
```
gramalingam, line 3971: Here too `dim_value` may be unknown

onnx/defs/math/defs.cc
smk2007: Updated to include hasInputShape checks.

onnx/defs/math/defs.cc
smk2007: I think multidim fft can be added if need is there in a separate PR. I am okay either way.

onnx/defs/math/defs.cc
smk2007: cast to float

docs/Changelog.md
smk2007: Update in the code, this will be generated at the end and reflected in this file.

onnx/defs/math/defs.cc
smk2007: Updated to allow negative values. The axis is defaulted to 1.

smk2007: done

onnx/defs/math/defs.cc
smk2007: added check.

onnx/defs/math/defs.cc
smk2007: added check.

onnx/defs/math/defs.cc
smk2007: added check.

onnx/defs/math/defs.cc
smk2007: Added check.

onnx/defs/math/defs.cc
smk2007: added check.

onnx/defs/math/defs.cc
```diff
@@ -3415,6 +3415,72 @@ static T get_scalar_value_from_tensor(const ONNX_NAMESPACE::TensorProto* t) {
...
+    }
+
+    // In general the output shape will match the input shape exactly
+    // So initialize the output shape with the input shape
+    auto& input_shape = getInputShape(ctx, 0);
+    ONNX_NAMESPACE::TensorShapeProto result_shape_proto = input_shape;
+
+    // Get the axis where the DFT will be performed.
+    auto axis = static_cast<int>(getAttribute(ctx, "axis", 1));
+    auto axis_idx = axis % input_shape.dim_size();
```
gramalingam, line 3435: This is a bit different from the usual handling (eg.: see: https://github.com/onnx/onnx/blob/dad79aca26a276a7b8175b9add0614b5ebaa6f89/onnx/defs/controlflow/old.cc#L156 ). Also, doesn't look like modulo will work as needed here for C++? (https://stackoverflow.com/questions/11630321/why-does-c-output-negative-numbers-when-using-modulo )

onnx/defs/math/defs.cc
```diff
@@ -3415,6 +3415,72 @@ static T get_scalar_value_from_tensor(const ONNX_NAMESPACE::TensorProto* t) {
...
+    }
+
+    auto dim_size = static_cast<int64_t>(result_shape_proto.dim_size());
+    auto has_component_dimension = dim_size > 2;
+    if (has_component_dimension) {
+        // If the rank is greater than 2, then the last dimension of the input is always the component dimension.
+        // Since the output is always complex, we can just coerce the last dimension to 2.
+        result_shape_proto.mutable_dim(static_cast<int>(dim_size - 1))->set_dim_value(2);
+    } else {
+        // If equal to 2, the last dimensions is missing, and this is the single dimension, real-valued DFT case
```
gramalingam, line 3476: Why this special case handling? Is there some compatibility or convenience reason? Whether the input is real or complex is typically independent of the rank of the input tensor. We could have 2D or 3D real-values, for example. Using different conventions for 2D reals and 3D reals is not a good idea.

gramalingam: Uniform convention would be always use one extra dimension (with a value of 1 for real-inputs and 2 for complex-inputs), regardless of whether rank is 2 or greater.

onnx/defs/math/defs.cc
smk2007: This should have been
    auto axis_idx = (input_shape.dim_size() + axis % input_shape.dim_size()) % input_shape.dim_size();
this will allow for arbitrary index with wrapping.

Is this behavior expected?

onnx/defs/math/defs.cc
```diff
@@ -3427,72 +3493,63 @@ static const char* DFT_ver16_doc =
...
-                "The final dimension represents the real and imaginary parts of the value."
-                "For real multi-dimensional input, the following shape is expected: [batch_idx][signal_dim1][signal_dim2]...[signal_dimN][1]."
-                "For complex multi-dimensional input, the following shape is expected: [batch_idx][signal_dim1][signal_dim2]...[signal_dimN][2]."
-                "The first dimension is the batch dimension.",
-                "T",
+                "For real input, the following shape is expected: [batch_idx][signal_dim1][signal_dim2]...[signal_dimN][1]. "
+                "For complex input, the following shape is expected: [batch_idx][signal_dim1][signal_dim2]...[signal_dimN][2]. "
+                "The first dimension is the batch dimension. "
+                "The following N dimentions correspond to the signal's dimensions. "
+                "The final dimension represents the real and imaginary parts of the value.",
```
gramalingam, line 3513: Please explicitly specify the order. Eg., if real comes first, then I suggest adding the words `in that order`

onnx/defs/math/defs.cc
smk2007: This is here because the existing contirib op in onnxruntime allows the special case of 1d reals that do not have a complex dimension. The general rule however is to always have a complex dimension.

gramalingam: This will likely cause confusion. I think we should try to make the spec simple. I suggest that we choose between the two simple options we have:

Option 1: Always use 1 extra-dimension as discussed above (no special cases).
Option 2: Always use a (new) attribute to indicate if the input is real or complex. If the input is real, no extra-dimension is used, and an extra-dimension is used only for complex.

Will neither of these options work?

onnx/defs/math/defs.cc
gramalingam: No. All other ops require axis to be in the range [-rank, rank-1] and reject an axis greater than rank. So, it is preferable that this op do the same. Thanks.

onnx/defs/math/defs.cc
smk2007: Option1 works. I only documented the behavior with the extra dimension.
I can remove the shape inference special case - which is a little hidden easter egg right now.

onnx/defs/math/defs.cc
smk2007: Im assuming that this is a numpy convention?
Seems kind of arbitrary to restrict the range.

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,715 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        {
+            // Clear the value and param (which would otherwie be inherited from the input).
+            result_shape_proto.mutable_dim(axis_idx)->clear_dim_value();
+            result_shape_proto.mutable_dim(axis_idx)->clear_dim_param();
+        }
+    }
+
+    auto dim_size = static_cast<int64_t>(result_shape_proto.dim_size());
+    auto has_component_dimension = dim_size > 2;
+    if (has_component_dimension) {
```
gramalingam, line 3479: The if-condition should also be dropped, right?

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,676 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                        rank);
+                }
+                
+                auto axis_idx = (axis >= 0 ? axis : axis + rank);
+                
+                // If dft_length is specified, then we should honor the shape.
+                // Set the output dimension to match the dft_length on the axis.
+                // If onesided this will be adjusted later on...
+                const TensorProto* dft_length = nullptr;
+                if (hasInputShape(ctx, 1))
```
gramalingam, line 3525: I don't think we need this condition and can drop it?

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,676 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                    {
+                        dft_length = ctx.getInputData(1);
+                    } else {
+                        // If we cannot read the dft_length, we cannot infer shape
+                        // return...
+                        return;
+                    }
+                }
+
+                if (nullptr != dft_length) {
```
gramalingam, line 3539: I think that 
(a) this if-statement can be moved into the then-branch above, after line 3531.
(b) we should add an else-branch to this condition, clearing value and param (like lines 3561/62)
(c) we can drop the else-branch above in line 3532 (because in that case, we use the input-dim as-is).
Do the changes make sense?

onnx/defs/math/defs.cc
```diff
@@ -3392,4 +3392,676 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        fail_shape_inference("size input must be a scalar.");
+      }
+
+      auto size_value = get_scalar_value_from_tensor<int64_t>(size); 
+      if (size_value <= 0) {
+        fail_shape_inference("size input must be greater than 0.");
+      }
+
+      ONNX_NAMESPACE::TensorShapeProto result_shape;
+      result_shape.add_dim()->set_dim_value(size_value);
```
gramalingam, line 3641: IIUC, the `periodic` attribute affects the size by 1?

onnx/defs/math/defs.cc
smk2007: dft_length is optional, it may not be supplied.

So i figured there were 3 cases.
1) Not specified -> will wont have InputShape
2) Specified, but not constant -> getInputData will fail
3) Specified, and constant -> getInputData will succeed

Is that not right?

gramalingam: You are right, we must distinguish between the three cases. But I was confused by the use of `hasInputShape`. I think that is not equivalent to checking if a particular input is present or not (an input might be specified, but its shape may not be available; however, a type is always available for a specified input in a valid program). I suggest using the condition `(ctx.getNumInputs() >= 2) && (ctx.getInputType(1) != nullptr)` as a way to check if this input is specified. @jcwchen : Does this sound reasonable? Can we add a method `hasInput(i)` to `InferenceContext` that is defined to be equivalent to a check like above?

onnx/defs/math/defs.cc
gramalingam: Since these suggestions correlate to the previous comment, maybe it is better if I describe the issues I think could be addressed (regardless of how the logic is structured). 
(a) The `return` in like 3535 could be made more precise. We cannot infer the dimension for the fft-axis, but the other dimensions can be propagated as is, correct?

docs/Changelog.md
```diff
@@ -19875,6 +19875,90 @@ This version of the operator has been available since version 15 of the default
...
+  Computes the discrete Fourier transform of input.
+
+#### Version
+
+This version of the operator has been available since version 16 of the default ONNX operator set.
+
+#### Attributes
+
+<dl>
+<dt><tt>axis</tt> : int (default is 1)</dt>
```
xadupre, line 19929: Should we expose fft_length as well?

onnx/defs/math/defs.cc
smk2007: Updated!

onnx/defs/math/defs.cc
smk2007: I updated the previous check.
I left the return in there because it seemed weird to set a partially specified tensor shape.

docs/Changelog.md
smk2007: it is an input

onnx/backend/test/case/node/stft.py
```diff
@@ -0,0 +1,62 @@
...
+from . import expect
+
+
+class STFT(Base):
+
+    @staticmethod
+    def export() -> None:
+        signal = np.arange(0, 128, dtype=np.float32).reshape(1, 128, 1)
+        length = np.array((16))
+        step = np.array((8))
```
jcwchen, line 21:
```suggestion
        step = np.array((8)).astype(np.int64)
```

onnx/backend/test/case/node/stft.py
```diff
@@ -0,0 +1,62 @@
...
+from ..base import Base
+from . import expect
+
+
+class STFT(Base):
+
+    @staticmethod
+    def export() -> None:
+        signal = np.arange(0, 128, dtype=np.float32).reshape(1, 128, 1)
+        length = np.array((16))
```
jcwchen, line 20:
```suggestion
        length = np.array((16)).astype(np.int64)
```
NumPy will use numpy.int32 by default on Windows and use numpy.int64 by default on Mac. Specifying the variable type should resolve this issue.

onnx/backend/test/case/node/stft.py
jcwchen: ditto

gramalingam (approved):

averkij:
Thank you for that.

I have a question though, — is it possible to use this version now for exporting pytorch model somehow?

jcwchen:
> I have a question though, — is it possible to use this version now for exporting pytorch model somehow?

I don't think so -- since ONNX just introduces Singal operators, it will take some time to let converters consume this update and be able to convert signal-related operators.

urinieto:
Is there a way to use something like [Custom Operators](https://pytorch.org/docs/stable/onnx.html#custom-operators) to be able to export pytorch models that use some of these audio operators?

(I know the default pytorch converter does not consume this update yet, but I am trying to figure out possible workarounds)

gramalingam:
@urinieto : possibly. You might get a more complete answer if you ask in the pytorch-converter repo.

justinchuby:
@urinieto Yes. That is possible. Happy to discuss more in the pytorch repo. Contributions are always welcome and appreciated!

PR #3745, opened 2021-09-29
Add Python 3.10 in AzurePipelines for testing

jcwchen:
**Description**
Add Python 3.10 in AzurePipelines

**Motivation and Context**
Before adding Python 3.10 in release pipelines: https://github.com/onnx/onnx/pull/3734, first test it in Azurepipelines.


askhade:
Let's revive this work?

garymm:
Don't need this because we have https://github.com/onnx/onnx/pull/3734

PR #3754, opened 2021-10-04
Clarify that shapes of scale and zero point must match for QuantizeLinear and DequantizeLinear

TomWildenhain-Microsoft:


docs/Operators.md
```diff
@@ -13020,6 +13023,11 @@ values = np.array([off_value, on_value], dtype=output_type)
 y = one_hot(indices, depth, axis=axisValue, dtype=output_type)
 y = y * (on_value - off_value) + off_value
 expect(node, inputs=[indices, depth, values], outputs=[y], name='test_onehot_negative_indices')
+
+# print(y)
+# [[3. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
```
TomWildenhain-Microsoft, line 13028: I didn't make this change. Not sure why it is here.

jcwchen: Hmmm it seems like an interesting bug. I cannot repro it from my end (Neither can the CIs).  I guess with certain Python version and platform `inspect` package behaves differently... (whether parse the comments in the end or not) Which Python version are you using?

Since I cannot repro it, could you please try the following commands from your end?
```
import inspect
def test_parse_comment():
   # Begin comment
   print('this is a test')
   # End comment
inspect.getsource(test_parse_comment)
```
My output is:
```
"def test_parse_comment():\n   # Begin comment\n   print('this is a test')\n"
```

I guess yours will include the last comment `# End comment\n`

A quick workaround will be moving these comments from the end of the files to the line before `expect()` function, just like what other comments do

TomWildenhain-Microsoft: Yes, this is what is happening.

PR #3761, opened 2021-10-07
Simplify common version converter adapter design patterns

gramalingam:
**Description**
Introduce generic transformers to help simplify a class of version converter adapters (based on adding, removing, and setting attributes). Replace existing adapters that fall into these categories.

**Motivation and Context**
- Eliminates duplicate/redundant code
- More declarative and readable transformers (clarify the exact change in a particular version)
- Removes a bug in AveragePool adapter that was missing a remove-attribute.
- This is part 1 of simplifying the version-converter that will also make it easier to maintain. (We can even eliminate the dependence on the IR if desired in the future.)

### DIFF
--- a/.azure-pipelines/Linux-CI.yml
+++ b/.azure-pipelines/Linux-CI.yml
@@ -4,7 +4,7 @@ trigger:
 jobs:
 - job: 'Test'
   pool:
-    vmImage: 'Ubuntu-16.04'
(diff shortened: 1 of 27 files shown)


gramalingam:
@matteosal : thanks for the suggestions. Yes, I am planning on doing the extensions you suggest, probably as a separate PR.

PR #3762, opened 2021-10-07
Fix shape in LSTM documentation

matteosal:
Reissue of https://github.com/onnx/onnx/pull/3459 to fix LSTM doc


gramalingam:
Thanks for the fix ... I guess it is a redo of the outdated PR: https://github.com/onnx/onnx/pull/3459 ... we can close the other one, once this is merged in.

Issue #3641, opened 2021-08-10
Convolution Op: are both Channel In & Out multiple of group number?


AlexandreEichenberger:
## Question

Are both the Channel In (C) and Channel Out (M) multiple of the group number for the Conv Op? In the doc, Channel In is specified as being a multiple of the group number. Nothing is said about Channel Out.

## Question Details

Here is what the doc say:
> W (differentiable) : T [...]  X.shape[1] == (W.shape[1] * group) == C
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Conv

Both Pytorch and Keras are specific, indicating that both channel in & out are to be multiple of the number of groups, for grouped convolutions.

> Pytorch requires both channel in (CI) and channel out (CO) to be multiple of group number (G).
https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html

> Keras also specifies it: Input channels and filters must both be divisible by groups.
https://www.tensorflow.org/api_docs/python/tf/keras/layers/Conv2D

### Further information

Answer to this matters as code needs to be different if supporting channel out (M) that are non-multiple of the number of groups.

I propose that we add the clarification that both are multiple of group number. I can make the change if there is agreement over this.

AlexandreEichenberger:
Conversation on slack, reported here for completeness.

>G. Ramalingam  5 days ago
My guess is that this is just a limitation of the documentation. I don't know enough about the semantics of a grouped convolution, but I guess the shape-inference function should indicate what the output dimensions are expected to be. But the documentation in the op itself is quite limited. It would be great if someone knowledgeable about the op could expand on the documentation.

>Alexandre Eichenberger  4 days ago
@G. Ramalingam thanks for your response. I added this constraint in our onnx-mlir code and thus I will know empirically if someone uses models where this does not hold. We run subsets of the model zoo, so I will know for sure if there is a case that break this assert. I will submit a PR to fix the doc if this holds true for a while.

kraiskil:
Reading the current ONNX documentation there doesn't seem to be any reason to require output feature maps being a multiple of groups. The only place where groups is documented to be used is in sharing the weights tensor on the input channel axis. I.e.`W[M][C/groups][...]`, so the number of feature maps(output channels) is independent of the number of groups.

[...]

PR #3763, opened 2021-10-07
Specify the CONV behavior in the presence of groups more precisely

AlexandreEichenberger:
This PR implements the changes suggested in #3641. Namely, Tensorflow and Pytorch conv2D requires the input and output channels to be multiple of the number of groups. Output channels are also referred to feature map. This is essentially because groups are used to reduce the size of the filter. Currently the document only states that the input channel must be a multiple of the number of groups.

Additional investigation by @kraiskil shows in the same issue #3761 that the model accept cases where the output channels is not a multiple of the number of groups, but fails to execute on the ONNX runtime.

In addition, the original text is plainly wrong on the input channel constraint:
```
X.shape[1] == (W.shape[1] * group) == C (assuming zero based indices for the shape array). Or in other words FILTER_IN_CHANNEL should be equal to DATA_CHANNEL.
```

The equation is correct, but the text is not: The correct sentence is
```
 Or in other words FILTER_IN_CHANNEL multiplied by the number of groups should be equal to DATA_CHANNEL.
```

This PR does both: fix the above error, and add the constraints on the output channels.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -19108,7 +19108,7 @@ This version of the operator has been available since version 14 of the default 
 <dt><tt>input_forget</tt> : int (default is 0)</dt>
 <dd>Couple the input and forget gates if 1.</dd>
 <dt><tt>layout</tt> : int (default is 0)</dt>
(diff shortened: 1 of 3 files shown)


AlexandreEichenberger:
@gramalingam Its my first contribution to this repo, and I am a bit surprised by the number of failure associated with a doc change. Is there something that appears like an obvious error on my part?

gramalingam:
@askhade or @jcwchen, any other suggestions?

gramalingam:
The generated documentation suggests that the generator is accessing older and out-of-date opschemas for other ops ... either because of build issues or because the source is not the latest.

jcwchen:
I saw a lot of old changes -- like removing Castlike, moving bool too np.bool. It seems that it was produced by an old ONNX because the update from `defs/nn/defs.cc` was not included either.

@AlexandreEichenberger could you please make sure you have uninstalled all pre-installed ONNX, `python setup.py install` and `python onnx/defs/gen_doc.py`? Then `gen_doc.py` should take your update from the right ONNX you just built. Thanks!

PR #3784, opened 2021-10-19
Use symbolic input for shape inference of ConstantOfShape

xuzijian629:
[Symbolic shape inference and partial data propagation](https://github.com/onnx/onnx/blob/master/docs/proposals/SymbolicShapeInfProposal.md) is available since 1.10. 

When the input of ConstantOfShape has symbolic input, i.e., propagated shape data, we use it as the output shape.

See a similar inference in [Reshape](https://github.com/onnx/onnx/blob/master/onnx/defs/tensor/defs.cc#L249).


onnx/defs/generator/defs.cc
```diff
@@ -235,6 +235,14 @@ ONNX_OPERATOR_SET_SCHEMA(
           // Shape inference based on input shape
           const TensorProto* targetShapeInitializer = ctx.getInputData(0);
           if (!targetShapeInitializer) {
+            // Shape inference via partial data propagation
+            const TensorShapeProto* shapeInput = ctx.getSymbolicInput(0);
+            if (shapeInput) {
+              TensorShapeProto targetShapeProto;
+              targetShapeProto.CopyFrom(*shapeInput);
```
gramalingam, line 242: I think the above 2 lines can be dropped.

jcwchen (approved):
Basically LGTM. Thank you for the enhancement!
Only a nit comment: I am thinking perhaps we can put all data propagation related tests to a new python file under onnx/test since current shape_inference_test.py focuses on pure (without data propagation) shape inference and it's quite huge.

gramalingam (approved):
LGTM, thanks for the improvement!

xuzijian629:
> Basically LGTM. Thank you for the enhancement! Only a nit comment: I am thinking perhaps we can put all data propagation related tests to a new python file under onnx/test since current shape_inference_test.py focuses on pure (without data propagation) shape inference and it's quite huge.

Thank you for the reviews!!
I agree with your opinion to separate tests with data propagation. Since the modification of the test file could be large and unrelated to ConstantOfShape, I'll create another PR (todo: https://github.com/onnx/onnx/issues/3944).

askhade (approved):

PR #3787, opened 2021-10-21
Consider multiple types for ParseData according to DataType definition

jcwchen:
**Description**
Make ParseData can parse more types correctly according to the map as below:
string_data -- string
int32_data -- int32, int16, int8, uint16, uint8, bool, float16 
int64_data -- int64
float_data -- float, complex64
double_data -- double, complex128
uint64_data -- uint32, uintt64

**Motivation and Context**
https://github.com/onnx/onnx/pull/3702 This PR made ParseData check the type for parsing. It needs the type exactly the same. However, according to [DataType definition](https://github.com/onnx/onnx/blob/be76ca7148396176784ba8733133b9fb1186ea0d/onnx/onnx.proto3#L549), multiple types can be stored as the same type. For instance, INT32, INT16, INT8, UINT16, UINT8, BOOL, or FLOAT16 can be stored as int32_data. ONNX did not bump into any issue because ParseData in ONNX only parse data for int32, int64, float, double. However, other projects like ONNXRuntime might rely on ParseData for other types. For example, AttentionFusion uses ParseData for uint8. Therefore, to resolve that, this PR makes ParseData in ONNX accept multiple types according to DataType definition.

### DIFF
--- a/onnx/defs/tensor_proto_util.cc
+++ b/onnx/defs/tensor_proto_util.cc
@@ -32,77 +32,72 @@ namespace ONNX_NAMESPACE {
     return t;                                                   \
   }
 
-#define DEFINE_PARSE_DATA(type, typed_data_fetch, supported_tensorproto_datatypes)      \
-  template <>                                                                           \
-  const std::vector<type> ParseData(const TensorProto* tensor_proto) {                  \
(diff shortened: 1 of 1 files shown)


jcwchen:
cc @hariharans29. Thank you for catching this issue!

PR #3789, opened 2021-10-21
Add shape inference for Expand using symbolic shape input

xuzijian629:
#### Description
Even when the shape input is not available as an initializer, we can infer the output shape using symbolic shape input.

#### Motivation
Existing shape inference cannot infer `dim_value` in some case like

```
agraph (float[3, 1, 2] x, float[1, 4, 2] y) => (float[3, 4, 2] z)
{
    ys = Shape(y)
    z = Expand(x, ys) 
}
```


xuzijian629:
@askhade I followed #3807 and simplified the implementation. Sorry for sooo late response..

onnx/defs/math/defs.cc
```diff
@@ -2050,15 +2050,23 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                    int64_t dim_value = shape_input_shape.dim(0).dim_value();
+                    for (int64_t i = 0; i < dim_value; ++i) {
+                        second_shape.add_dim();
+                    }
+                }
             }
             bidirectionalBroadcastShapeInference(
                 input_shape, second_shape, *getOutputShape(ctx, 0));
+          } else {
+              // Use data propagation for shape inference when shape_initializer
```
jcwchen, line 2069: What is this else for?

xuzijian629: Sorry, this was just a mistake when I manually merged main

jcwchen (approved):
LGTM. Thank you!

gramalingam:
LGTM, thanks! A minor comment: I think we would want to use the same logic for any input that denotes a shape (in any op). So, extracting this as a generic utility may be useful. Eg., a function like:
```
   TensorShapeProto getShapeInput(InferenceContext& ctx, size_t input_index);
```

gramalingam (approved):
LGTM. Added a minor suggestion.

xuzijian629:
> LGTM, thanks! A minor comment: I think we would want to use the same logic for any input that denotes a shape (in any op). So, extracting this as a generic utility may be useful. Eg., a function like:
> 
> ```
>    TensorShapeProto getShapeInput(InferenceContext& ctx, size_t input_index);
> ```

Thank you for the suggestion. I have a question about the implementation.
How do you handle the case when the shape input has neither initializer nor symbolic data? When the shape input does not even have shape, we cannot do anything, i.e., we cannot return a `TensorShapeProto`.
Maybe it's better to prepare a function like `setShapeInputData(InferenceContext& ctx, size_t input_index, TensorShapeProto* target_shape)` to avoid such situation?

gramalingam:
Good question! Your suggestion is fine, but I think we will need to return a boolean value to indicate whether a value was found or not.

I think this case is not being handled correctly by the existing code, so it probably needs to be fixed anyway. I think the existing code uses the default value assigned to ```TensorShapeProto second_shape;``` which has zero dimensions, which is incorrect. If no information is available, we should skip setting the output shape.

xuzijian629:
@gramalingam Sorry for late response! I implemented `getShapeInput` and used it in the shape inference of Expand :+1:

gramalingam:
Hi @xuzijian629 : thanks for the update! Would it be possible to fix the conflicts, so we can merge this in? Thanks!

xuzijian629:
Hi @gramalingam,
I resolved the conflicts and merged main.

memo: https://github.com/onnx/onnx/pull/4019/commits/37ae3126ce5e480931d5e52089230cec80ab5b5d returns early if symbolic input does not have dim_value. In such case, `GetShapeInput` returns with `found == false`, so already handled.

gramalingam:
Great, thanks very much @xuzijian629

jcwchen:
@gramalingam Could you please sign-off approval for this PR again and let's merge this one? Thanks!

PR #3802, opened 2021-10-25
update test_if_opt

chudegao:
Signed-off-by: degaochu <degaochu@cn.ibm.com>

**Description**
- update node test test_if_opt: change then_body output to optional of seq(empty seq as []) instead of optional of tensor(null). This makes then body's output consist to else body's output type and consist to the ground truth "[]" in function compute_if_outputs.

**Motivation and Context**
- Current test case will make onnx-tf test failed as then_body node output is not same as expected.


chudegao:
@neginraoof can you help review?

docs/Operators.md
```diff
@@ -8538,15 +8539,21 @@ x = [np.array([1, 2, 3, 4, 5]).astype(np.float32)]
...
+    inputs=[],
+    outputs=['seq_empty_out']
+)
+
 opt_empty_in = onnx.helper.make_node(
     'Optional',
-    inputs=[],
+    inputs=['seq_empty_out'],
     outputs=['optional_empty'],
-    type=seq_in_tp
```
hariharans29, line 8545: Sorry, I am not following what this change is achieving.

Previously the type attribute contains a TypeProto which is basically `sequence(tensor(float))` and there are no inputs to the `Optional` node. Looking at the spec - https://github.com/onnx/onnx/blob/master/docs/Operators.md#Optional, this is valid and this operator will produce an empty `optional(sequence(tensor(float)))` as is right ? Why do we need an empty sequence input flowing into the Optional node and also change the type to reflect this (I would have thought the type attribute is moot after the input comes into the picture). It seems to me like even as is, the Optional node will produce an empty optional sequence of float tensor and after this change it will still do the same.

And also not following the part of making the then and else output types match - I would have thought the types of `then_out` and `else_out` are the same even now - `optional(sequence(tensor(float)))`

Sorry if I am overlooking something very obvious.

chudegao:
```
Previously the type attribute contains a TypeProto which is basically sequence(tensor(float)) and there are no inputs to the Optional node. Looking at the spec - https://github.com/onnx/onnx/blob/master/docs/Operators.md#Optional, this is valid and this operator will produce an empty optional(sequence(tensor(float))) as is right ? 
```
I don't think it's right. The output is still seqence as the node definition.
```
>>> ten_in_tp = onnx.helper.make_tensor_type_proto(onnx.TensorProto.FLOAT, shape=[5])
>>> seq_in_tp = onnx.helper.make_sequence_type_proto(ten_in_tp)
>>> opt_empty_in = onnx.helper.make_node(
            'Optional',
            inputs=[],
            outputs=['optional_empty'],
            type=seq_in_tp
        )
>>> 
>>> opt_empty_in
output: "optional_empty"
op_type: "Optional"
attribute {
  name: "type"
  tp {
    sequence_type {
      elem_type {
        tensor_type {
          elem_type: 1
          shape {
            dim {
              dim_value: 5
            }
          }
        }
      }
    }
  }
  type: TYPE_PROTO
}

>>> 
```

After the update, the output is optional(sequence):
```
>>> ten_in_tp = onnx.helper.make_tensor_type_proto(onnx.TensorProto.FLOAT, shape=[5])
>>> seq_in_tp = onnx.helper.make_sequence_type_proto(ten_in_tp)
>>> opt_in_tp = onnx.helper.make_optional_type_proto(seq_in_tp)
>>> seq_empty_node = onnx.helper.make_node(
            'SequenceEmpty',
            inputs=[],
            outputs=['seq_empty_out']
        )
>>> opt_empty_in = onnx.helper.make_node(
            'Optional',
            inputs=['seq_empty_out'],
            outputs=['optional_empty'],
            type=opt_in_tp
        )
>>> opt_empty_in
input: "seq_empty_out"
output: "optional_empty"
op_type: "Optional"
attribute {
  name: "type"
  tp {
    optional_type {
      elem_type {
        sequence_type {
          elem_type {
            tensor_type {
              elem_type: 1
              shape {
                dim {
                  dim_value: 5
                }
              }
            }
          }
        }
      }
    }
  }
  type: TYPE_PROTO
}


```

hariharans29: Not quite following. The type attribute in the Node contains a sequence but the output type of the Node will still be an optional right ? 

See how the output type is created for the Optional node:
https://github.com/onnx/onnx/blob/34dbc3e3a8f4709d96e7ceba3ac38510f4ed2ff3/onnx/defs/optional/defs.cc#L53

chudegao: @hariharans29 You are right. I misunderstand the usage of attr type. I revert the change. Please review. Thanks

hariharans29: The change to feed an empty sequence into Optional is redundant because the behavior is still the same with or without it. Without it, an "empty" optional is created from the type attribute

chudegao: Ok. I will update the backend implement to use the type to generate the right output. I will close this PR.

hariharans29: Makes sense to me, thanks. It isn't that there is something wrong with your test case, it is just that nothing seems to be "broken" in the existing test that needs a "fix". AFAIK  per the operator spec, this test is valid. I implemented this op in onnxruntime and I was able to run the test fine.

PR #3806, opened 2021-10-26
Enhance shape inference: ParseData/Transpose/QuantizeLinear

### LINKED ISSUE #3709: ONNX checker doesn't check repeated permutation for Transpose operator
# Bug Report

### Is the issue related to model conversion?
If the ONNX checker reports issues with this model then this is most probably related to the converter used to convert the original framework model to ONNX. Please create this bug in the appropriate converter's GitHub repo (pytorch, tensorflow-onnx, sklearn-onnx, keras-onnx, onnxmltools) to get the best help.

### Describe the bug
The ONNX checker doesn't check if `perm` attribute contains duplicate entry, which is checked in numpy. The sample code below repeats 0 axis and passed the check.

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): Ubuntu 20.04
- ONNX version (*e.g. 1.7*):  1.10.1
- Python version: 3.8.10

### Reproduction instructions
```
from onnx import helper
from onnx import TensorProto as tp
from onnx import checker
from onnx import save
from onnx import shape_inference
inputs = [
    ("input", tp.FLOAT, [2, 3, 4]),
]
weights = []
outputs = [
    ("output", tp.FLOAT, [4, 2, 2]),
]
node = helper.make_node('Transpose',
                        inputs=[input[0] for input in (inputs + weights)],
                        outputs=[output[0] for output in outputs],
                        perm=[2, 0, 0])
graph = helper.make_graph(
    [node], 'test_graph',
    [helper.make_tensor_value_info(*input) for input in inputs],
    [helper.make_tensor_value_info(*output) for output in outputs],
    [helper.make_tensor(*weight) for weight in weights])
model = helper.make_model(graph, producer_name='model')
checker.check_model(model, full_check=True)
inferred_shape = shape_inference.infer_shapes(model)
print(inferred_shape)
```

### Notes

Any additional information

### LINKED ISSUE #3744: ONNX shape inference for QuantizeLinear fails on empty string as zero_point
# Bug Report

### Describe the bug
An ONNX model with QuantizeLinear (opset 13) passes shape inference when it has no input for zero_point, but fails if the input is the empty string.  Trailing optional inputs should be able to be expressed in either way.

onnx version 1.10.1

jcwchen:
**Description**
1. Enhance  ParseData check: having raw_data cannot be String
2 Transpose should check repeated perm to prevent issues
3 If third input (y_zero_point) of QuantizeLinear is an empty string, shape inference should still work because y_zero_point is an optional input

**Motivation and Context**
1. Continues a sub task from https://github.com/onnx/onnx/pull/3787#discussion_r733854228.
2. Closes #3709 
3. Closes #3744


jcwchen:
> The change looks good... please add tests

Good point. I added 2 tests for checking repeated perm for Transpose and empty string for QuantizeLinear. 

Regarding to test for ParseData with raw_data and string type, actually ParseData [does not support parse string type now](https://github.com/onnx/onnx/blob/1faae95520649c93ae8d0b403816938a190f4fa7/onnx/defs/tensor_proto_util.cc#L116) so the condition cannot be tested... If that is the case, do we still need to add the check here?

askhade (approved):

PR #3810, opened 2021-10-28
Check negative index for attributes of Slice-1

### LINKED ISSUE #3565: infer_shapes fails but onnxruntime works
# Bug Report

### Is the issue related to model conversion?
onnx raises an exception while running ``infer_shapes`` (``onnx.onnx_cpp2py_export.shape_inference.InferenceError: [ShapeInferenceError] (op_type:Sqrt, node name: ComplexAbsoutput__19): [ShapeInferenceError] Inferred shape and existing shape differ in dimension 1: (4) vs (3)``).

### Describe the bug

The following code fails (onnx graph: [bug_shape_inference.zip](https://github.com/onnx/onnx/files/6785879/bug_shape_inference.zip)):
```
import onnx
with open('bug_shape_inference.onnx', 'rb') as f:  #see attached file
    model_proto = onnx.load(f)
onnx.checker.check_model(model_proto, full_check=True)
```

Exception:
```
onnx.onnx_cpp2py_export.shape_inference.InferenceError: [ShapeInferenceError] (op_type:Sqrt, node name: ComplexAbsoutput__19): [ShapeInferenceError] Inferred shape and existing shape differ in dimension 1: (4) vs (3)
```

However, onnxruntime is able to compute the output:

```
from onnxruntime import InferenceSession
import numpy
mat = numpy.random.randn(3, 4).astype(numpy.float32)
sess = InferenceSession('bug_shape_inference.onnx')
print(sess.run(None, {'input:0': mat}))
```

It produces a matrix (3,3) as mentioned in the onnx graph:

```
  output {
    name: "output:0"
    type {
      tensor_type {
        elem_type: 1
        shape {
          dim {
            dim_value: 3
          }
          dim {
            dim_value: 3
          }
        }
      }
    }
  }
```

The error message seems to say onnx expects an output with shape (4,3) or (3,4) when it should be (3,3).

### System information
- OS Platform and Distribution (*Windows, Ubuntu 20.04*): 
- ONNX version (*e.g. 1.7*):  1.9.0
- Python version: 3.9.5
- Visual Studio version (if applicable): 2019

### Reproduction instructions
See the bug description.

### Expected behavior
I would expect infer_shapes to succeed and find a shape equal to (3,3).

jcwchen:
**Description**
Add negative index check for attributes start, end, axes of Slice-1. To prevent breaking existing old models, instead of throwing an error, just simply finish rank inference and return.

**Motivation and Context**
Closes #3565 
The model uses opset 9 so it uses Slice-1. The inferred output shape is wrong because its Slice-1 uses a negative index for axes. Please note that negative index support was added after opset-1: https://github.com/onnx/onnx/pull/2281 Thus Slice-1 does not support it. Still, shape inference for Slice-1 should catch this negative index error instead of propagating a wrong shape.


onnx/defs/tensor/old.cc
```diff
@@ -3192,7 +3192,14 @@ ONNX_OPERATOR_SET_SCHEMA(
...
                 "Incorrect or missing attribute value for starts and ends");
-            ;
+          }
+          auto is_negative = [](int64_t index) {
+            return index < 0;
+          };
+          if (!std::none_of(starts.begin(), starts.end(), is_negative)) {
+            fail_shape_inference("Attribute start cannot have negative value.");
+          } else if (!std::none_of(ends.begin(), ends.end(), is_negative)) {
+            fail_shape_inference("Attribute end cannot have negative value.");
```
askhade, line 3202: Instead of throwing an error can we either work with the negative value (when possible) or fallbacl to rank inference? This is because some runtimes (example ORT) did allow negative values even in cases where the standard did not explicitly state it... I am afraid throwing an error here will break existing scenarios...

jcwchen: Good point. After offline discussion, to prevent breaking existing old models, ONNX should at least perform rank inference instead of throwing an error. Just updated. Also it can resolve the shape inference error from #2281. Please review it again. Thank you!

askhade: You said: Also it can resolve the shape inference error from #2281. Please review it again.
Which error?

askhade: Is this change complete? I only see you switched from throwing error for attr axis and not for start and end

jcwchen: Oh sorry. I meant this one: https://github.com/onnx/onnx/issues/3565 Also update start and end. Thank you!

onnx/defs/tensor/old.cc
```diff
@@ -3192,21 +3192,40 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
           std::vector<int64_t> axes;
           if (!getRepeatedAttribute(ctx, "axes", axes)) {
             for (int i = 0; (size_t)i < starts.size(); ++i) {
               axes.push_back(i);
             }
           } else if (axes.size() != starts.size()) {
             fail_shape_inference("Attribute axes has incorrect length");
-            ;
+          } else if (!std::none_of(axes.begin(), axes.end(), is_negative)) {
```
askhade, line 3213: adding this check as else if does not seem right... both the if condition above and this else if condition regarding negative axes check can be true at once right?

askhade (changes requested):
added comment

onnx/defs/tensor/old.cc
gramalingam: yes. I think we should replace `!std::none_of` by `std::any_of`

gramalingam: sorry, my comment is a different from one Ashwini's. We may need to change the nested-else, but the condition also needs to be changed, I think

onnx/defs/tensor/old.cc
```diff
@@ -3192,21 +3192,40 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            has_negative_axis = true;
           } else if (!std::is_sorted(axes.begin(), axes.end())) {
             // TODO support shape inference for unsorted axes
             return;
           }
-
+          if (has_negative_axis) {
+            // Since negative axes are not supported before opset-10
```
gramalingam, line 3220: In light of the discussion in the thread, can we expand this comment and say
```
   Negative axes were not explicitly discussed in the spec before opset-10. Hence, they are
   officially not part of the spec, but some models/runtimes may use them. Hence, we 
   perform simple rank inference in this case.
```

onnx/defs/tensor/old.cc
gramalingam: It seems to work ok (since we fail in the previous case).

onnx/defs/tensor/old.cc
```diff
@@ -3192,21 +3192,40 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             fail_shape_inference(
                 "Incorrect or missing attribute value for starts and ends");
-            ;
           }
+
+          auto is_negative = [](int64_t index) {
+            return index < 0;
+          };
+          bool has_negative_axis = false;
+          if (!std::none_of(starts.begin(), starts.end(), is_negative) ||
```
gramalingam, line 3201: Replacing `! none_of` by `any_of` may make it slightly easier to read.

onnx/defs/tensor/old.cc
askhade: aah right. I think only changing !none_of to any_of (for better readability) should be enough

jcwchen:
Thanks for the review. I have updated comments, used any_of and moved the negative axes condition into one place to simplify the code.

askhade (approved):

gramalingam (approved):

PR #3815, opened 2021-10-29
Shape inference should not propagate missing optional outputs

### LINKED ISSUE #3439: Shape inference for Unique assigns shape to empty string when optional outputs are provided.
A value_info entry is added for empty string, which is not correct.

Onnx version 1.8.1.

https://www.dropbox.com/s/ljyo821c0b2miv5/test_unsorted_segment_sum.onnx?dl=1

jcwchen:
**Description**
~- Add outputIsEmpty in InferenceContext to make it know which outputs are missing optional outputs (do not have name)~
~- Catch it in caller: getOutputType will get dummy empty TypeProto if the output is missing name. In that case graph-level shape inference won't propagate it into value_info~
~- Fix original Unique bug: should not propagate missing optional outputs~

Shape inference will now detect empty output in graph-level and won't produce value_info in that case.

**Motivation and Context**
Shape inference for Unique assigns shape to empty string when optional outputs are provided.
Closes #3439


askhade:
Why not simply skip propagating the empty output shapes downstream in graph level shape inference? Wouldn't that be a much simpler approach?

jcwchen:
> Why not simply skip propagating the empty output shapes downstream in graph level shape inference? Wouldn't that be a much simpler approach?

After offline discussion, for now we decide not to extend InferenceContext for detecting empty output during node-level shape inference. To keep it simple without changing interface, shape inference will detect it in graph-level and won't infer a shape for empty output.

onnx/defs/tensor/defs.cc
```diff
@@ -3390,6 +3390,8 @@ ONNX_OPERATOR_SET_SCHEMA(
 
           // 'indices', 'inverse_indices', and 'counts' are 1-D tensors of
           // unknown dimension.
+          // For now it is possible that an empty output infers a shape here
+          // but graph-level shape inference will catch it and won't produce a inferred shape
```
askhade, line 3394: // Shape inference will happen even in case of empty optional outputs, 
// graph-level shape inference should not propagate the shape downstream for empty optional outputs.

onnx/shape_inference/implementation.cc
```diff
@@ -426,7 +426,9 @@ static void InferShapesImpl(
       }
       for (int i = 0; i < n.output_size(); ++i) {
         auto* inferredType = ctx.getOutputType(i);
-        if (inferredType->value_case() == TypeProto::ValueCase::VALUE_NOT_SET) {
+        // If an output is missing (lacks name), shape inference should not infer shape
+        if (inferredType->value_case() == TypeProto::ValueCase::VALUE_NOT_SET ||
+          n.output(i).empty()) {
```
askhade, line 431: suggest making this a separate condition and adding it above line 428. We should not even fetch the inferredType with output is empty.

On a separate note why is there no check for inferredType != nullptr? Can we safely assume ctx.getOutputType(i) never returns a nullptr?

jcwchen: > suggest making this a separate condition and adding it above line 428. We should not even fetch the inferredType with output is empty.

Sounds good and I just moved the condition 

> On a separate note why is there no check for inferredType != nullptr? Can we safely assume ctx.getOutputType(i) never returns a nullptr?

For now I think it should be fine because the vector of output type will be initialized here: https://github.com/onnx/onnx/blob/master/onnx/shape_inference/implementation.h#L184

askhade (approved):

onnx/shape_inference/implementation.cc
```diff
@@ -425,6 +425,10 @@ static void InferShapesImpl(
         schema->CheckInputOutputType(ctx);
       }
       for (int i = 0; i < n.output_size(); ++i) {
+        // If an output is missing (lacks name), shape inference should not infer shape
```
askhade, line 428: change to : // skip type and shape propagation for missing optional outputs.

PR #3828, opened 2021-11-08
Append dim even both dim value and param are not set

xuzijian629:
**Description**
- In data propagation, fix `appendDimToTensorShapeProto` so that a dimension is appended even when both dim value and param are not known.

**Motivation and Context**
- I got a ShapeInferenceError caused by [here](https://github.com/onnx/onnx/blob/master/onnx/defs/tensor/defs.cc#L624), where the `input_data` is not nullptr but its dims have neither dim_values nor dim_params.


askhade:
Curious - Are you using ONNX's graph level shape inference? meaning are you using onnx shape inference apis? When using onnx graph level shape inference I don't know how you could land in this situation this because - either the input data is an initializer in which case all dimensions are dim values or the input data is dynamic in which case graph level shape inference will create a symbol for every unknown dim... (https://github.com/onnx/onnx/blob/master/onnx/shape_inference/implementation.cc#L141)

However, not all runtimes implement symbol generation and therefore it may be best to include your fix.

onnx/defs/data_propagators.h
```diff
@@ -7,11 +7,7 @@
 namespace ONNX_NAMESPACE {
 
 inline void appendDimToTensorShapeProto(TensorShapeProto& tsp, const TensorShapeProto_Dimension& dim) {
-  if (dim.has_dim_value()) {
-    tsp.mutable_dim()->Add()->set_dim_value(dim.dim_value());
-  } else if (dim.has_dim_param()) {
-    tsp.mutable_dim()->Add()->set_dim_param(dim.dim_param());
-  }
+  tsp.mutable_dim()->Add()->CopyFrom(dim);
```
askhade, line 10: can you simply add a dim?  tsp.mutable_dim()->Add() instead of Copy

askhade (changes requested):
Added a comment

askhade:
@xuzijian629 : The PR looks good to me. I added a minor comment. As soon as you address it we can merge this. Thanks for your contribution.

askhade:
@xuzijian629: Any updates?

askhade:
Any updates?

xuzijian629:
Sorry, I missed the notification. I will soon make an update :bow:

xuzijian629:
> https://github.com/onnx/onnx/pull/3828#issuecomment-963380020

Yes, I encountered this problem when I was using ONNX's graph level shape inference.

I got the bug in the following part of a model.



Since I cropped the graph, the inputs of `Gather` and `Sub` are hidden. Note that the inputs to `Unsqueeze`s are all 0-dimensional (i.e., scalar).

xuzijian629:
I summarized the above explanation into code.

Note that before running this, you may need to merge https://github.com/onnx/onnx/pull/3784 because the following code uses data propagation for ConstantOfShape.

```python
import onnx
from onnx import parser, helper, shape_inference

input = '''
   agraph (float[3, n] x) => ()
   {
        shape = Shape(x)
        one = Constant<value = int64 {1}>()
        scalar = Gather(shape, one)
        subbed = Sub(scalar, one)
        unsqueeze_dim = Constant<value = int64[1] {0}>()
        unsqueezed = Unsqueeze(subbed, unsqueeze_dim)
        concat_out = Concat<axis = 0>(shape, unsqueezed)
        out = ConstantOfShape(concat_out)
   }
'''
graph = parser.parse_graph(input)
original_model = helper.make_model(graph, producer_name='onnx-examples')
onnx.checker.check_model(original_model)

# Apply shape inference on the model
inferred_model = shape_inference.infer_shapes(original_model, data_prop=True)
onnx.checker.check_model(inferred_model)
print(inferred_model.graph.value_info)
```

With this PR, the output shape of `out` is `[3, n, unk__0]`, where as the existing ONNX infers as `[3, n]`, which has incorrect rank.

xuzijian629:
Please take a look! @askhade :bow:

onnx/defs/data_propagators.h
```diff
@@ -11,6 +11,8 @@ inline void appendDimToTensorShapeProto(TensorShapeProto& tsp, const TensorShape
     tsp.mutable_dim()->Add()->set_dim_value(dim.dim_value());
   } else if (dim.has_dim_param()) {
     tsp.mutable_dim()->Add()->set_dim_param(dim.dim_param());
+  } else {
+    tsp.mutable_dim()->Add();
```
gramalingam, line 15: Thanks for fixing this! This looks correct. But I think the whole function can be simplified to
```
   * tsp.add_dim() = dim;
```

xuzijian629: Thank you!

gramalingam (approved):

askhade (approved):

PR #3855, opened 2021-11-19
Make test/node generation produce models with since_version by default

### LINKED ISSUE #2629: [CI] Check backend tests have been generated
### Problem
It's currently easy to merge a PR with missing or out-of-date backend test files.

CI can find issues if those files are present, particularly for functions, where the expanded graph can fail the checker or shape inference, exposing bugs in the body definition. (e.g. #2607 #2623)

### Suggested resolution
It would be good to check in CI: generate backend test data & models and compare with what's checked into git.

### Questions
It's worth noting that model files can have older IR and opset versions, but running [`tools/update_doc.sh`](https://github.com/onnx/onnx/blob/master/tools/update_doc.sh) generates the same files with newer versions.
⇒ _Question: presumably we want to keep these older versions the same -- should we change the python that generates them to explicitly set the versions? Atm it seems like a manual process of choosing which of the many changed files should actually be added to git._

jcwchen:
**Description**
- By default, test generation should produce test models with since_version
- Enable test generation to specify opset_version if needed
- Add test generation in Windows-CI, Linux-CI and Mac-CI to verify consistent node models
- Update some tests to use make_model_gen_version with specified opset_version (e.g., simple tests)

P.S. since_version is the version of the operator set that this operator was initially declared in.

More details about changed behaviors please see the comment [here](https://github.com/onnx/onnx/pull/3855#issuecomment-1024800124).

**Motivation and Context**
Closes #2629 Currently test generation will always produce models with the latest opset_version. However, if we want to test  "test generation" in CIs with the same situation when they were uploaded, test generation should be always consistent  regardless of which platform. To achieve that, make_model_gen_version with specified since_version is needed.


onnx/backend/test/case/node/__init__.py
```diff
@@ -155,6 +156,14 @@ def expect(node,  # type: onnx.NodeProto
         inputs=inputs_vi,
         outputs=outputs_vi)
     kwargs[str('producer_name')] = 'backend-test'
+
+    if 'opset_imports' not in kwargs:
+        # To make sure the model will be produced with the same opset_version after opset changes
+        # By default, it uses since_version as opset_version for produced models
+        since_version = onnx.defs.get_schema(node.op_type, node.domain).since_version
```
askhade, line 163: use max_inclusive_version for fetching schema to make sure we fetch the right schema for tests IF they target an older opset. Today none of the tests do but it is best if we add this check.

onnx/backend/test/case/node/__init__.py
```diff
@@ -155,6 +156,14 @@ def expect(node,  # type: onnx.NodeProto
...
+        produce_opset_version = int(_TargetOpsetVersion) if _TargetOpsetVersion is not None else since_version  # type: ignore
```
askhade, line 164: Instead of a random target opset version the option should be to use the max inclusive version for the test... If we use a random opset version most likely we won't be able to generate test data for all tests this is because if users choose an older opset version for which the tests are not applicable we have to skip those tests.

askhade: Also add comments on line 129 regarding this change

askhade (changes requested):
added comments

onnx/backend/test/case/node/__init__.py
jcwchen: Good point. I have updated get_schema with max_inclusive_version and make target opset_version as use_max_opset_version, which is more accurate. It takes some time for to inspect whether there is any updated model. Will update them later.

askhade: > If you run your script and do a git diff that should tell you right?

jcwchen: There might be other changes due to system difference... To be careful, I will go through all of them (perhaps write a simple script to verify it).

jcwchen: Eventually I wrote a script for only updating models/output.pb if their opset_version have been changed by this PR. However, there are 566 models which need to be updated (864 models in total). This PR will impact projects which are relying on backend test data.

askhade: in that case we can change the test to include max_inclusive_version parameter to restrict the model to older opset

askhade: going forward this should not be a problem... we should ask the authors to update all the models for the changing op... infact it will be gated by the git diff check in the PR.

onnx/backend/test/data/real/test_zfnet512/data.json
```diff
@@ -1 +1 @@
-{"atol": 1e-07, "model_name": "zfnet512", "rtol": 0.001, "url": "https://s3.amazonaws.com/download.onnx/models/opset_9/zfnet512.tar.gz"}
```
jcwchen: Based on the specified opset version here: https://github.com/onnx/onnx/blob/7f12395a87cc85d9c6138a8b99530571014d3307/onnx/backend/test/case/model/__init__.py#L40, updating opset_10 here should be correct. I was surprise that these haven't been updated for a while. Since now we do test all models from ONNX model Zoo weekly, perhaps we can consider to remove/refactor these old models.

jcwchen:
This PR will forward after 1.11 release since it involves too many files and makes a fundamental change about data generation.

Here is the summary that data generation behavior will be changed as follows:


**Previous**

1. `python onnx/backend/test/cmd_tools.py generate-data` always produces node models with the "latest" opset version. There is no way to generate a model with older opset version by newer ONNX.
2. Because of 1., newer ONNX always updates many unrelated models by bumping their opset version to the latest one (although they are not updated)
3. Because of 1., newer ONNX cannot created models with older opset version. Now users need to get older models by older ONNX package.
4. Many old models were not updated when its opset version has bumped. 

**New**

1. By default, `python onnx/backend/test/cmd_tools.py generate-data` will now produce the node models with "since_version".
2. Users can specify a desired opset version for generate-data. It will find the latest version ("since_version") before desired opset version.
3. Due to 2., now generate-data is able to create a whole set of node models with an older opset version.
4. Developers can specify older opset_version by giving `opset_imports` while writing a node test. It will use since_version based on specified opset_version.
5. Node models are all updated by their opset's "since_version".
6. Due to 5., now CI should be able to check the consistency of all generated data.

askhade (commented):
removing "changes requested" status to unblock the PR

jcwchen:
Sorry I was wrong. Actually there are three test failures from ORT due to this update. They are
- onehot_negative_indices
- onehot_with_axis
- onehot_with_negative_axis

The error messages are all the same, which is: `[ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid rank for input: depth Got: 0 Expected: 1 Please fix either the inputs or the model.` I will take a closer look at this issue.

onnx/backend/test/case/node/__init__.py
```diff
@@ -150,6 +160,31 @@ def expect(node: onnx.NodeProto,
...
+        max_inclusive_versions = []
+        onnx_opset_version = None
+        for opset in kwargs[str('opset_imports')]:
+            produce_opset_version = onnx.defs.get_schema(node.op_type, opset.version, opset.domain).since_version
+            max_inclusive_versions.append(onnx.helper.make_operatorsetid(opset.domain, produce_opset_version))
+            if opset.domain == '' or opset.domain == 'ai.onnx':
+                onnx_opset_version = produce_opset_version
+        kwargs[str('opset_imports')] = max_inclusive_versions
+        if onnx_opset_version is not None:
+            kwargs[str('ir_version')] = _OpsetVersionToIRVersionMap[onnx_opset_version]
```
jcwchen, line 186: By default helper.make_model will produce a model with the latest IR_VERSION, which means giving an older opset_version make_model will create a model with older opset_version but with the latest ir_version. This looks problematic to me. Therefore, I use the version map here to get the proper ir_version according to the opset_version.

However, it even brings more updates than before. The reason is typically when a contributor introduced a new node test data, at that moment the opset_version has been already bumped, but the ir_version has not been bumped before the release. Thus most cases those node test data were using opset_version with (corresponding ir_version -1). This PR will make them use corresponding ir_version instead, which makes more sense to me.

jcwchen (commented):
> The error messages are all the same, which is: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid rank for input: depth Got: 0 Expected: 1 Please fix either the inputs or the model. I will take a closer look at this issue.

This is because of this ONNX PR: https://github.com/onnx/onnx/pull/3774. The depth should be scalar. Perhaps the issue is in ORT that it expects the depth for OneHot is 1D. I will dig deeper into it.

jcwchen:
> This is because of this ONNX PR: https://github.com/onnx/onnx/pull/3774. The depth should be scalar. Perhaps the issue is in > ORT that it expects the depth for OneHot is 1D. I will dig deeper into it.

I found these failures only happened when new input+old ONNX node model so it is a false alarm. With new input and new ONNX node model, all tests for node model from ONNX backend all passed by ONNXRuntime (I tested them locally by `onnx_backend_test_series.py` in ORT). This PR is ready for review.

onnx/helper.py
```diff
@@ -52,6 +52,10 @@ def process(release_version: Text, ir_version: int, *args: Any) -> None:
         for pair in zip(['ai.onnx', 'ai.onnx.ml', 'ai.onnx.training'], args):
             if (pair not in result):
                 result[pair] = ir_version
+                if pair[0] == 'ai.onnx':
+                    result[('', pair[1])] = ir_version
+                if pair[0] == 'ai.onnx.training':
+                    result['ai.onnx.preview.training', pair[1]] = ir_version
```
jcwchen, line 58: Some training testing models originally used ai.onnx.preview.training and didn't have IR_VERSION. In this PR, these models are updated with specified IR_VERSION.

onnx/backend/test/case/node/__init__.py
```diff
@@ -187,8 +208,27 @@ def expect(node: onnx.NodeProto,
...
+        # By default, it uses since_version as opset_version for produced models
+        if _TargetOpsetVersion is None:
+            produce_opset_version = onnx.defs.get_schema(node.op_type, node.domain).since_version
+        # If given targeted opset version
+        # it will generate test data for the latest opset vesion that supports before targeted opset version
+        else:
+            produce_opset_version = onnx.defs.get_schema(node.op_type, int(_TargetOpsetVersion), node.domain).since_version
+        kwargs[str("opset_imports")] = [onnx.helper.make_operatorsetid(node.domain, produce_opset_version)]
+    else:
+        # Also, for the same reason above, convert given opset.version (from opset_imports) to max_inclusive_version
```
gramalingam, line 224: Do we really need to do this? If the user explicitly specifies opset version in the test-case, why not just use that? Automatically adjusting it can be problematic in some edge cases (eg., what if the user really wants a specific opset-version? Even though the test-case is for a single node, that node could be a control-flow node with a subgraph of operators which may require specific opset version, for example). On the flip side, does adjusting it give us any benefit or advantage?

jcwchen: Thank you @gramalingam for catching it. I agree with your point -- If specifying opset version here, we should simply use it instead of using max_inclusive_version. It provides more flexibility for users if they really want specific opset version, especially for some control-flow cases. I will update here soon to use the opset version directly.

onnx/backend/test/case/node/__init__.py
```diff
@@ -206,47 +246,43 @@ def expect(node: onnx.NodeProto,
...
-def collect_testcases() -> List[TestCase]:
-    '''Collect node test cases defined in python/numpy code.
-    '''
-    import_recursive(sys.modules[__name__])
-    return _NodeTestCases
-
-
-def collect_testcases_by_operator(op_type: Text) -> List[TestCase]:
-    '''Collect node test cases which include specific operator
+def collect_testcases(op_type: Text, opset_version: int) -> List[TestCase]:
```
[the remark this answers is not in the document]
gramalingam: Is there a pressing need for this? Or, is this something that can be addressed separately later? May be we can talk offline in the meeting tomorrow. Just trying to understand what the spec should be for this.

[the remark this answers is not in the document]
jcwchen: There is no urgent need for it. I am fine to decouple this part to another new PR since this PR is already a huge one. I will do it later. Thanks!

jcwchen:
Update: I am trying to test the consistency of test generation in CIs. It seems that only Windows CI can have exactly the same node tests as this PR. I suspect there is an issue like this one: https://github.com/onnx/onnx/pull/3120. Still investigating.

onnx/backend/test/case/node/__init__.py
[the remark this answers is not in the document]
jcwchen: Updated. Please review it again. Thank you!

jcwchen:
> Update: I am trying to test the consistency of test generation in CIs. It seems that only Windows CI can have exactly the same node tests as this PR (I updated these node test models on Windows). I suspect there is an issue like this one: #3120. Still investigating.

This issue exists because NumPy functions might behave differently on different platforms. Therefore skip all output.pb and some input.pb for now. I have added test generation in Windows CI, Linux CI and Mac CI in this PR. At least all ONNX models produced by test generation are consistent.

gramalingam (approved):

askhade:
overall the PR looks good to me. I suggest using this branch in ort CIs and make sure all pipelines are green.

jcwchen:
Thanks for the review! I have verified these updated models once in ORT CIs and there isn't any related failure: https://github.com/microsoft/onnxruntime/pull/11214

Issue #2449, opened 2019-11-12
Spec unclear: Can a scalar be Reshaped to a 1D tensor and vice versa?


mathisdon:
A scalar has shape [ ] and rank 0. Can the Reshape operator be applied to it with a shape of [ 1 ]? And Can a 1D tensor of shape [ 1 ] have Reshape applied with a shape of [ ]?

linkerzhang:
Thanks a lot for bringing this up! I think both should be supported. @ebarsoum  @postrational

PR #3861, opened 2021-11-24
Version converter for Softmax should not produce empty shape

### LINKED ISSUE #3853: Version converter for Softmax 12 to 13 should not produce a Reshape node with empty shape
# Bug Report

Original issue: https://github.com/onnx/onnx/issues/3728

The code for repro:
```
import onnx
from onnx import version_converter
model = onnx.load('bertsquad-8.onnx')
model_opset_15 = version_converter.convert_version(model, 15)  # from onnx/models
# onnx.save(model_opset_15, "bertsquad-8_opset_15.onnx") # For C++ side to load
inferred_model = onnx.shape_inference.infer_shapes(model_opset_15, True, True, True)
```

Model link: https://github.com/onnx/models/blob/master/text/machine_comprehension/bert-squad/model/bertsquad-8.onnx

The error message:
```
InferenceError: [ShapeInferenceError] Shape inference error(s): (op_type:MatMul, node name: bert/encoder/layer_0/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_1/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_2/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_3/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_4/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_5/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_6/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_7/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_8/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_9/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_10/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
(op_type:MatMul, node name: bert/encoder/layer_11/attention/self/MatMul_1): [ShapeInferenceError] Input tensors of wrong rank (0).
```

After version conversion, the shape for Reshape node should not have an empty shape:
Screen Shot 2021-11-18 at 2 40 54 PM

### LINKED ISSUE #3854: Shape inference for Reshape op should catch empty shape error
# Bug Report
Original issue: https://github.com/onnx/onnx/issues/3728

### Reproduction instructions
```
import onnx
from onnx import version_converter
model = onnx.load('bertsquad-8.onnx')
model_opset_15 = version_converter.convert_version(model, 15)  # onnx/models
# onnx.save(model_opset_15, "bertsquad-8_opset_15.onnx") # For C++ side to load
inferred_model = onnx.shape_inference.infer_shapes(model_opset_15, True, True)
...
```
https://github.com/onnx/models/blob/master/text/machine_comprehension/bert-squad/model/bertsquad-8.onnx

### Expected behavior
Shape inference should complain Reshape node cannot have empty shape

jcwchen:
**Description**
- If the version conversion fail with Softmax due to unknown input shape, throw an error
- Update spec about empty shape for Reshape

**Motivation and Context**
Close #3853, ~#3854~, #2449.


onnx/version_converter/adapters/softmax_12_13.h
```diff
@@ -48,6 +48,10 @@ class Softmax_12_13 final : public Adapter {
 
         // Set shape input of Reshape
         const std::vector<Dimension>& target_shape = flatten->inputs()[0]->sizes();
+
+        ONNX_ASSERTM(target_shape.size() != 0, "Version conversion for Softmax failed because "
+          "input shape is unknown.");
+
```
jcwchen, line 54: Not sure whether the version conversion here for Softmax can be done without knowing the input shape due to dynamic situation... At least version converter should catch this failure and not produce an invalid model. (has a Reshape node with an empty Shape)

onnx/defs/tensor/defs.cc
```diff
@@ -260,6 +260,10 @@ ONNX_OPERATOR_SET_SCHEMA(
             return;
           }
 
+          if (targetShapeProto.dim_size() == 0) {
```
gramalingam, line 263: Why can't we allow an empty shape? That's what we need for a scalar. So, if we want to convert a tensor with shape [1] into a scalar with shape [], I think that should be permitted?

jcwchen: Thank you for catching it! I wasn't sure for the case of empty shape because the spec is not very clear. But, it makes sense to me that using empty shape for reshaping 1D to scalar. Will correct it and focus on the fix in version_converter.

gramalingam: Thanks for clarifying the documentation. May be it will help to also say that "The input tensor's shape and the output tensor's shape are required to have the same number of elements."

gramalingam (approved):

onnx/version_converter/adapters/softmax_12_13.h
gramalingam: True. This problem also appears elsewhere, since the version-converter seems to assume that the shapes are known in many places. How does the underlying IR represent the cases where the rank is unknown or a dimension is unknown?

jcwchen: IIUC, it will create a node with empty sizes() (empty vector of Dimension).

PR #3879, opened 2021-12-04
Extend InferShapes to expose result of data propagation

jcwchen:
**Description**
- Extend existing InferShapes to expose the result of data propagation
- Add data propagation functions for Shape-1, Gather-11, Gather-1
- Refactored several shape inference function to make APIs' parameter order consistent
-  Improve data propagation function for Gather

**Motivation and Context**
- Background:
ONNX introduced data propagation in ONNX 1.10. However, those data are only used internally and they can only be recollected by Reshape and ConstantOfShape (getSymbolicInput).
PyTorch exporter are planning to utilize ONNX symbolic shape inference to simplify its implementation. Having a way to get those data propagation would be really helpful (it relies on graph-level shape inference).

- Advantages:
1.	Make the result of data propagation more applicable for more use cases (not only obtain from Reshape or ConstantOfShape)
2.	Current tests for the new proposed data propagation are few. If we enable it in PyTorch exporter, it can relatively have more tests for it
3.	The result of data propagation can be useful for other ONNX-related projects as well
4.	Engage more users for using data propagation with the new API

### More context why PyTorch exporter needs this
Current PyTorch exporter relies on ONNX shape inference to infer the tensors' types and shapes. However, it does not utilize the new ONNX symbolic shape inference. Instead, additional symbolic shape inference was implemented. Utilizing symbolic shape inference from ONNX can help to simplify the implementation. For the same reason, exposing the result of data propagation can benefit other ONNX-related projects as well. Then there won't be duplicate work for symbolic shape inference.

### Why PyTorch exporter cannot utilize symbolic shape inference by current ONNX's API?
PyTorch exporter now uses ONNX shape inference on a **subgraph** of ONNX model ([code](https://github.com/pytorch/pytorch/blob/26b8a009a53a57b3c5d4e4020c02b093de9d4ae3/torch/csrc/jit/passes/onnx/shape_type_inference.cpp#L2076)). Such a ONNX subgraph was converted by certain PyTorch op so it can consist one or a few ONNX ops. In most cases, there won't be Reshape or ConstantOfShape node to collect those data propagation result. Thus, having such an API like InferShapesAndDataPropagation to expose the result of data propagation is required. In that case, any kind of ONNX graph can benefit from it.

### Why don't PyTorch exporter use ONNX node-level shape inference to utilize data propagation?
Although it's doable, it will be duplicate work for the exporter to implement basically the same code of `onnx::shape_inference::InferShapes`. I think having another API to expose data propagation result can easily reduce a lot of work for other projects which rely on ONNX shape inference. It seems more profitable to achieve it from ONNX side. Other projects can benefit from it.


jcwchen:
Update: add more context in the PR description about why I think ONNX should add this API.

onnx/shape_inference/implementation.cc
```diff
@@ -684,6 +688,40 @@ void InferShapes(
   }
 }
 
+std::unordered_map<std::string, TensorShapeProto> InferShapesAndDataPropagation(
+    ModelProto& m,
+    const std::unordered_map<std::string, TensorShapeProto>& generated_shape_data_by_name,
+    const ISchemaRegistry* schema_registry,
+    const ShapeInferenceOptions& options) {
+  std::unordered_map<std::string, int> opset_imports;
```
gramalingam, line 696: I wonder why some of the code below doesn't happen in the constructor of ShapeInferenceImpleBase? May be with a separate, different, constructor from the existing one?

onnx/defs/data_propagators.h
```diff
@@ -49,4 +53,29 @@ inline void PropagateShapeDataFromInputToOutput(DataPropagationContext& ctx, int
   }
 }
 
+inline void GatherOp13DataPropagator(DataPropagationContext& ctx) {
```
gramalingam, line 56: nit: seems like this PR is doing two different things? Just wondering why it also has updates to specific data. prop. logic?

onnx/shape_inference/implementation.cc
```diff
@@ -684,6 +688,40 @@ void InferShapes(
   }
 }
 
+std::unordered_map<std::string, TensorShapeProto> InferShapesAndDataPropagation(
+    ModelProto& m,
+    const std::unordered_map<std::string, TensorShapeProto>& generated_shape_data_by_name,
```
gramalingam, line 693: So, is this used because the caller iteratively uses results from previous calls to do inference for other subgraphs later?

onnx/shape_inference/implementation.cc
```diff
@@ -563,6 +563,10 @@ class ShapeInferenceImplBase {
             ir_version,
             model_local_functions_map} {}
 
+  std::unordered_map<std::string, TensorShapeProto> getGeneratedShapeData() {
+    return generated_shape_data_by_name;
```
gramalingam, line 567: Seems okay for now. But this may end up creating unnecessary copies of a large data-structure. It may be useful to consider a design that uses a `std::move` or some other approach.

jcwchen: Good point. Actually I was considering to modify `generated_shape_data_by_name` in place without returning it by `InferShapesAndDataPropagation` since another input parameter `ModelProto` is also called by reference and can be modified in place. So here are two options:
1. Make `generated_shape_data_by_name` call by reference as `InferShapesAndDataPropagation`'s input.
2. Return `generated_shape_data_by_name` additionally by `InferShapesAndDataPropagation`. (What this PR currently does and it needs some improvement for moving it as you mentioned)

@gramalingam do you have a preference about it? Thanks.

onnx/shape_inference/implementation.cc
jcwchen: I think the code below are only common in 2 places: one is in ` InferShapes` and another one is `InferShapesAndDataPropagation`, because only these 2 take ModelProto as input. Although we cannot put them into ShapeInferenceImpleBase (it's for GraphProto), perhaps at least we can make them as a common function to reduce duplicate code.

Still, it seems that `TraverseGraphsToAddExistingSymbol` can be included in the ShapeInferenceImpleBase constructor. I will give it a try.

onnx/shape_inference/implementation.cc
gramalingam: There are two interfaces where copying can happen: when the top-level `Infer...` API is called, and second when the Base class constructor is called. So, both will need to be addressed. 

I think that if option 1 works, that should be good.

gramalingam: Specifically, the question is: can we also change the base-class constructor to take a non-const reference, and avoid copying there?

onnx/shape_inference/implementation.cc
jcwchen: I introduced two functions, `GetOpsetImportsFromProto` and `GetModelLocalFunctionsMapFromModel`, to reduce duplicate code between `InferShapes` and `InferShapesAndDataPropagation`. In addition, I moved TraverseGraphsToAddExistingSymbol into process(GraphProto& graph) since it's required step to process a graph.

BTW, I might find a potential bug that it does not consider symbols in a FunctionProto. I created an issue to track that: https://github.com/onnx/onnx/issues/4191.

onnx/shape_inference/implementation.cc
jcwchen: Thanks for your opinion. I made base-class constructor take a non-const reference and `InferShapesAndDataPropagation` use `generated_shape_data_by_name` by reference, which can address some potential copy issues. Please review it again.

onnx/shape_inference/implementation.h
```diff
@@ -459,6 +459,16 @@ void InferShapes(
     const ISchemaRegistry* schema_registry = OpSchemaRegistry::Instance(),
     const ShapeInferenceOptions& options = {}
     );
+///
+/// Forces shape inference to run with data propagation and data propagation result will be updated in place
```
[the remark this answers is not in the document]
gramalingam: Would it make sense to then just extend the previous method with a 
```cpp
   std::unordered_map<std::string, TensorShapeProto>* generated_shape_data_by_name = nullptr
```
?

onnx/shape_inference/implementation.h
```diff
@@ -86,22 +86,25 @@ struct GraphInferenceContext {
...
+        ir_version{ir_version_in} {}
 
   const std::unordered_map<std::string, TypeProto*>* outer_scope_value_types_by_name;
   const std::unordered_map<std::string, int> opset_imports;
   SymbolTable* symbol_table;
+  const ModelLocalFunctionsMap& model_local_functions;
   const ISchemaRegistry* schema_registry;
+  std::unordered_map<std::string, TensorShapeProto>* generated_shape_data_by_name;
   const int ir_version;
-  const ModelLocalFunctionsMap& model_local_functions;
```
jcwchen, line 104: FYI: I changed the order here for `ModelLocalFunctionsMap& model_local_functions` to make it consistent to other APIs.

onnx/shape_inference/implementation.h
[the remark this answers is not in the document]
jcwchen: Thank you for the great idea! Yes, I think we can simply extend existing `InferShapes` instead of introducing more similar API. I have just removed new `InferShapesAndDataPropagation` and use generated_shape_data_by_name as pointer globally. Please review it again.

onnx/shape_inference/implementation.cc
```diff
@@ -629,37 +654,35 @@ void InferShapes(
...
-      opset_imports,
-      options,
-      &symbol_table,
-      model_local_functions_by_id,
-      schema_registry,
-      m.ir_version());
+void InferShapes(ModelProto& m,
+    const ISchemaRegistry* schema_registry,
+    const ShapeInferenceOptions& options,
+    std::unordered_map<std::string, TensorShapeProto>* generated_shape_data_by_name) {
```
gramalingam, line 660: What is the specification/contract for this API? I think that to keep this backwards-compatible, the implementation below should handle the case where `(generated_shape_data_by_name == nullptr) && (options.enable_data_propagation )` by creating a local `map` and passing its address to the constructor. I realize that one of the other functions expects the user to do this, but here we are constrained by backwards-compatibility.

gramalingam: Okay, never mind, I see that this is happening in `InferShapesImpl` !

gramalingam (approved):

PR #3880, opened 2021-12-06
[Testing] Test all models from model Zoo with current version converter

jcwchen:
**Description**
 Test all models from model Zoo with current version converter

**Motivation and Context**
 Test all models from model Zoo with current version converter and compare the result with the updated version converter from this PR: https://github.com/onnx/onnx/pull/3676.


jcwchen:
Close here since I finished the experiment.

PR #3882, opened 2021-12-08
Simplify function definition of context-dependent functions

gramalingam:

**Description**
Simplify the definition of context-dependent functions, using the parser and with a convenient function-builder API.

**Motivation and Context**
The API simplifies defining functions in the future. The function definitions are also more compact and more readable, reducing the chances of unexpected errors.


onnx/defs/function.h
```diff
@@ -106,4 +107,83 @@ class FunctionBodyHelper {
...
+    while (!parser.EndOfInput()) {
+      auto status = parser.Parse(*nodes.Add());
+      if (!status.IsOK())
+        ONNX_THROW_EX(std::logic_error("Error parsing node:" + status.ErrorMessage()));
+    }
+
+    return *this;
+  }
+
+  FunctionBuilder& Add(const char* node_txt, const AttributeProto& attr) {
```
askhade, line 127: Is this function needed? Can the above one not handle a case with node + attr
For example Cast<to=1>(input) is supported right?

askhade: if we remove this then we can make the function body construction code more uniform... for cases where attr value is static all nodes can be constructed with 1 Add call others can use "Add(const char* node_txt, const std::string& attr_name, T attr_value)"

gramalingam: For the first question, (as your 2nd message implies), there are cases where the value is not static and cannot be embedded in the string. For the second comment, the ```Add(const char* node_txt, const std::string& attr_name, T attr_value)``` variant is syntactic shorthand for invoking this and is defined in terms of this. So, not sure what we achieve by eliminating this. Since there are other utility functions to create AttributeProto (other than just name/value), having the two variants doesn't hurt and could be useful.

askhade: I was just suggesting that we can keep 2 methods public since they cover all the scenarios: 
Add(const char* node_txt)
Add(const char* node_txt, const std::string& attr_name, T attr_value)

Add(const char* node_txt, const AttributeProto& attr) can be made private.

The reason I am suggesting this is - too many options can confuse the author and usually these utilities are used by a large number of folks (who are contributing to onnx functions) and everyone does not have a thorough understanding of this.

askhade (approved):

PR #3896, opened 2021-12-15
Honor existing dim_param in shape inference

shinh:
**Description**

Before this change, existing `dim_param` was overwritten by generated symbols like `unk__`. `NonZero` is used to test the behavior.

**Motivation and Context**

We'd like to annotate output values of ops which produce unknown dims (e.g., NonZero) by pre-setting `dim_param`s of such values. However, current ONNX overwrites them.

Reproduce script:

```
import onnx
import onnx.shape_inference
import torch


class NonZero(torch.nn.Module):
    def forward(self, x):
        return torch.nonzero(x)


x = torch.rand(7)
torch.onnx.export(NonZero(), x, "nonzero.onnx",
                  output_names=["y"],
                  dynamic_axes={"y": {0: "NZ"}})
model = onnx.load("nonzero.onnx")
model = onnx.shape_inference.infer_shapes(model)
print(model.graph.output)  # "NZ" is overwritten by "unk__0"
```

Previous behavior:

| inferred\existing | dim_value | dim_param | N/A |
| - | - | - | - |
| dim_value | error | inferred | inferred |
| dim_param | existing | inferred | inferred |
| unk__ | existing | inferred | inferred |

This PR:

| inferred\existing | dim_value | dim_param | N/A |
| - | - | - | - |
| dim_value | error | inferred | inferred |
| dim_param | existing | **existing** (1) | inferred |
| unk__ | existing | **existing** (2) | inferred |

I guess no one would object the change for (2) in the above table, which is the reason why I created this PR. I'm not sure about (1). To me, it is more natural to check if the existing `dim_param` is equal to the inferred one like what we do for `dim_value`, but it'd break backward compatibility. What do you think?

### DIFF
--- a/onnx/shape_inference/implementation.cc
+++ b/onnx/shape_inference/implementation.cc
@@ -172,13 +172,8 @@ void mergeShapesAndTypes(const TypeProto_Tensor& inferredType, TypeProto_Tensor*
   }
 
   if (!existingType->has_shape()) {
-    // Ensure the shape is initialized. Note that this must be done
-    // even for (zero-dimensional) scalars.
-    existingType->mutable_shape();
-
-    for (int j = 0; j < inferredType.shape().dim_size(); ++j) {
-      existingType->mutable_shape()->add_dim();
-    }
+    *existingType->mutable_shape() = inferredType.shape();
+    return;
   }
 
   for (int i = 0; i < inferredType.shape().dim_size(); ++i) {
@@ -201,13 +196,8 @@ void mergeShapesAndTypes(const TypeProto_SparseTensor& inferredType, TypeProto_S
   }
 
   if (!existingType->has_shape()) {
-    // Ensure the shape is initialized. Note that this must be done
-    // even for (zero-dimensional) scalars.
-    existingType->mutable_shape();
-
-    for (int j = 0; j < inferredType.shape().dim_size(); ++j) {
-      existingType->mutable_shape()->add_dim();
-    }
+    *existingType->mutable_shape() = inferredType.shape();
+    return;
   }
 
   for (int i = 0; i < inferredType.shape().dim_size(); ++i) {
--- a/onnx/shape_inference/implementation.cc
+++ b/onnx/shape_inference/implementation.cc
@@ -184,7 +184,8 @@ void mergeShapesAndTypes(const TypeProto_Tensor& inferredType, TypeProto_Tensor*
   for (int i = 0; i < inferredType.shape().dim_size(); ++i) {
     const auto& inferredDim = inferredType.shape().dim(i);
     auto* existingDim = existingType->mutable_shape()->mutable_dim(i);
-    if (!existingDim->has_dim_value()) {
+    if ((!existingDim->has_dim_value() && !existingDim->has_dim_param()) ||
+        inferredDim.has_dim_value()) {
       *existingDim = inferredDim;
     }
   }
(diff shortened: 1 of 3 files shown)


onnx/test/symbolic_shape_test.py
```diff
@@ -104,9 +104,8 @@ def test_two_symbolic_concat(self):  # type: () -> None
         inferred_model = onnx.shape_inference.infer_shapes(onnx_model, strict_mode=True)
         self._assert_valueinfo_shape(inferred_model, [
             make_tensor_value_info("C", TensorProto.FLOAT, (2, -1)),
-            make_tensor_value_info("E", TensorProto.FLOAT, (2, -1))])
-        # the symbolic shape of E and output should be the same
-        assert self._get_shape_from_name(inferred_model, 'E') == self._get_shape_from_name(inferred_model, 'output')
+            make_tensor_value_info("E", TensorProto.FLOAT, (2, -1)),
```
gramalingam, line 107: Maybe it would be better to change line 100 above and replace 'M' by None? And leave the current check?

onnx/test/symbolic_shape_test.py
```diff
@@ -81,9 +81,9 @@ def test_concat_enable_symbolic(self):  # type: () -> None
 
         onnx_model = make_model(graph_def)
         inferred_model = onnx.shape_inference.infer_shapes(onnx_model, strict_mode=True)
-        self._assert_valueinfo_shape(inferred_model, [make_tensor_value_info("C", TensorProto.FLOAT, (2, -1))])
-        # the symbolic shape of C and output should be the same
-        assert self._get_shape_from_name(inferred_model, 'C') == self._get_shape_from_name(inferred_model, 'output')
+        self._assert_valueinfo_shape(inferred_model, [
```
gramalingam, line 84: Should we change line 79, to specify output shape as [2, None] or [None, None], and leave the current check as-is?

onnx/shape_inference/implementation.cc
```diff
@@ -184,7 +184,8 @@ void mergeShapesAndTypes(const TypeProto_Tensor& inferredType, TypeProto_Tensor*
   for (int i = 0; i < inferredType.shape().dim_size(); ++i) {
     const auto& inferredDim = inferredType.shape().dim(i);
     auto* existingDim = existingType->mutable_shape()->mutable_dim(i);
-    if (!existingDim->has_dim_value()) {
+    if ((!existingDim->has_dim_value() && !existingDim->has_dim_param()) ||
+        inferredDim.has_dim_value()) {
```
askhade, line 188: I think the logic should be as follows:
```
if ((!existingDim->has_dim_value() && !existingDim->has_dim_param()) ||
        inferredDim.has_dim_value() && !existingDim->has_dim_value()) {
// prefer inferred dim value when 1. existingDim is unknown and 2. existingDim does not have dim_value
*existingDim = inferredDim;
}
```

askhade, line 188: condition
```
(!existingDim->has_dim_value() && !existingDim->has_dim_param()
```
can be folded in for loop on line 179. This way when existing shape is empty we dont have to run through 2 for loops (1 on line 179 and 1 on line 184)

onnx/test/symbolic_shape_test.py
shinh: Ah, nice. Done!

onnx/test/symbolic_shape_test.py
shinh: Ditto. Done!

onnx/shape_inference/implementation.cc
shinh: Is this what you meant? https://github.com/onnx/onnx/pull/3896/commits/be56060267a7ade6f420152e11c742f8d15a00ab

Note I think we still need `(!existingDim->has_dim_value() && !existingDim->has_dim_param()` here for the case where `existingType` has a known rank (i.e., `existingType->has_shape()` is true) with unknown dimensions.

askhade: yes thanks for adding line 175. And yes, I agree we still need this check here for the case where existingType has a known rank.

askhade (approved):

gramalingam (approved):

PR #3897, opened 2021-12-15
Update TreeEnsembleClassifier and TreeEnsembleRegressor to support tensor as attributes

xadupre:
**Description**

This PR changes FLOATS attribute into TENSOR for operator TreeEnsembleClassifier and TreeEnsembleRegressor and update ML OPSET to 3.

**Motivation and Context**

TreeEnsembleRegressor supports double as input type but only supports float as attribute. This introduces discrepancies when comparing ONNX to lightgbm. Lightgbm uses double, `x < th` becomes `(float)x < (float)th` after the conversion into ONNX. The result of this comparison is usually the same as the inner one but the probability to be different grows with the number of predictions to compute. When it is, lightgbm and ONNX produce results coming from different leave in the same tree. The discrepancy is usually significant.


gramalingam:
Just wondering: should we permit both floats as well as tensor attributes? (That is, support both the old style and new style.) That will be backwards-compatible, making it easier for exporters and the version-converter. This would mean introducing the tensor attributes with a new attribute name, and requiring that the model specify either the original or new attribute.

gramalingam:
Need to also update the version-converter (and version-conversion tests). This depends on how the above comment is addressed

xadupre:
I looked into that but it does not seem to support other domains than the main one: https://github.com/onnx/onnx/blob/master/onnx/version_converter/convert.h.

xadupre:
> Just wondering: should we permit both floats as well as tensor attributes? (That is, support both the old style and new style.) That will be backwards-compatible, making it easier for exporters and the version-converter. This would mean introducing the tensor attributes with a new attribute name, and requiring that the model specify either the original or new attribute.

In that case, which name do you suggest, some thing like `class_weights_tensor`?

gramalingam:
> > Just wondering: should we permit both floats as well as tensor attributes? (That is, support both the old style and new style.) That will be backwards-compatible, making it easier for exporters and the version-converter. This would mean introducing the tensor attributes with a new attribute name, and requiring that the model specify either the original or new attribute.
> 
> In that case, which name do you suggest, something like `class_weights_tensor`?

Yes, for every FLOATS attribute named X, we could add an attribute named X_tensor or X_as_tensor

gramalingam:
Hi: I think extending the version-converter infrastructure to support ONNXML would be useful. It requires extending the structure to include "domain" as an additional field/parameter (with a default value of "", so that none of the existing ONNX converters need to be changed). Would it make sense to do that as part of this PR? Thanks!

xadupre:
I would prefer to do it in another PR. It seems a different topic to me.

onnx/defs/traditionalml/defs.cc
```diff
@@ -1002,9 +1024,32 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+          auto* nodes_hitrates = ctx.getAttribute("nodes_hitrates");
+          auto* nodes_hitrates_as_tensor = ctx.getAttribute("nodes_hitrates_as_tensor");
+          auto* class_weights = ctx.getAttribute("class_weights");
+          auto* class_weights_as_tensor = ctx.getAttribute("class_weights_as_tensor");
+          auto* base_values = ctx.getAttribute("base_values");
+          auto* base_values_as_tensor = ctx.getAttribute("base_values_as_tensor");
+
+          if (nullptr != nodes_values && nullptr != nodes_values_as_tensor) {
+            fail_shape_inference(
+                "Only one of the attributes 'nodes_values', 'nodes_values_as_tensor'.");
```
gramalingam, line 1039: minor nit: could improve the error message to say "Only one of the attributes ... and ... should be specified."

onnx/defs/traditionalml/defs.cc
```diff
@@ -1002,9 +1024,35 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+          auto* base_values = ctx.getAttribute("base_values");
+          auto* base_values_as_tensor = ctx.getAttribute("base_values_as_tensor");
+
+          if (nullptr != nodes_values && nullptr != nodes_values_as_tensor) {
+            fail_shape_inference(
+                "Only one of the attributes 'nodes_values', 'nodes_values_as_tensor' should be specified.");
+          }
+          if (nullptr != nodes_hitrates && nullptr != nodes_hitrates_as_tensor) {
+            fail_shape_inference(
+                "Only one of the attributes 'nodes_hitrates', 'nodes_hitrates' should be specified.");
```
gramalingam, line 1043: second "nodes_hitrates" => "nodes_hitrates_as_tensor"

onnx/defs/traditionalml/defs.cc
```diff
@@ -1115,9 +1180,41 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
```
gramalingam, line 1205: Same as above: second "nodes_hitrates" => "nodes_hitrates_as_tensor"

onnx/defs/traditionalml/defs.cc
xadupre: done

onnx/defs/traditionalml/defs.cc
gramalingam: Hi, looks like the other one was fixed, but this still needs to be fixed. Thanks!

xadupre: Just fixed. Sorry for that. I commit before it was saved.

gramalingam (approved):

PR #3908, opened 2021-12-24
Clearer description of Slice

diyessi:
Signed-off-by: Scott Cyphers <cyphers@lightmatter.co>

**Description**
- Reworked the description of `Slice`.

**Motivation and Context**
- The intended behavior for negative values is unclear; the referenced NumPy documentation is equally unclear, as is Python's documentation.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -18040,19 +18040,37 @@ This version of the operator has been available since version 13 of the default 
 ### <a name="Slice-13"></a>**Slice-13**</a>
 
   Produces a slice of the input tensor along multiple axes. Similar to numpy:
-  https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html
-  Slices uses `starts`, `ends`, `axes` and `steps` inputs to specify the start and end
-  dimension and step for each axis in the list of axes, it uses this information to
-  slice the input `data` tensor. If a negative value is passed for any of the
-  start or end indices, it represents number of elements before the end of that
-  dimension. If the value passed to start or end is larger than the `n` (the
-  number of elements in this dimension), it represents `n`. For slicing to the
-  end of a dimension with unknown size, it is recommended to pass in `INT_MAX`
-  when sclicing forward and 'INT_MIN' when slicing backward.
-  If a negative value is passed for step, it represents slicing backward.
-  However step value cannot be 0.
-  If `axes` are omitted, they are set to `[0, ..., ndim-1]`.
+  https://numpy.org/doc/stable/user/basics.indexing.html?highlight=slice#slicing-and-striding
+
+  Slice uses the `starts`, `ends`, `axes` and `steps` inputs to select a sub-tensor
+  of its input `data` tensor.
+
+  An effective `start[i]`, `end[i]`, and `step[i]` must be computed for each `i`
+  in `[0, ... r-1]` where `r = rank(input)` as follows:
+
+  If `axes` are omitted, they are set to `[0, ..., r-1]`.
   If `steps` are omitted, they are set to `[1, ..., 1]` of length `len(starts)`
+
+  The effective values are initialized as `start[i] = 0`, `end[i] = dims[i]` where
+  `dims` are the dimensions of `input` and `step[i] = `1.
+
+  All negative elements of `axes` are made non-negatve by adding `r` to them, where
+  `r =rank(input)`.
+
+  All negative values in `starts[i]` and `ends[i]` have `dims[axes[i]]` added to them,
+  where `dims` are the dimensions of `input`. Then `start[axes[i]]` is the adjusted
+  `starts[i]` clamped into range of valid indices, i.e. `[0, dims[axes[i]]-1]`.
+
+  The clamping for the adjusted `ends[i]` depends on the sign of `steps[i]` and must
+  accommodate copying 0 through `dims[axes[i]]` elements, so for positive stepping
+  `end[axes[i]]` is clamped to `[0, dims[axes[i]]]`, while for negative stepping it
+  is clamped to `[-1, ends[i]-1]`.
+
+  Finally, `step[axes[i]] = steps[i]`.
+
+  For slicing to the end of a dimension with unknown size, it is recommended to pass
+  in `INT_MAX` when slicing forward and 'INT_MIN' when slicing backward.
+
   Example 1:
     data = [
         [1, 2, 3, 4],
@@ -18090,9 +18108,9 @@ This version of the operator has been available since version 13 of the default 
 <dt><tt>ends</tt> (non-differentiable) : Tind</dt>
 <dd>1-D tensor of ending indices (exclusive) of corresponding axis in `axes`</dd>
 <dt><tt>axes</tt> (optional, non-differentiable) : Tind</dt>
-<dd>1-D tensor of axes that `starts` and `ends` apply to. Negative value means counting dimensions from the back. Accepted range is [-r, r-1] where r = rank(data).</dd>
+<dd>1-D tensor of axes that `starts` and `ends` apply to. Negative value means counting dimensions from the back. Accepted range is [-r, r-1] where r = rank(data). Behavior is undefined if an axis is repeated.</dd>
 <dt><tt>steps</tt> (optional, non-differentiable) : Tind</dt>
-<dd>1-D tensor of slice step of corresponding axis in `axes`. Negative value means slicing backward. 'steps' cannot be 0. Defaults to 1.</dd>
+<dd>1-D tensor of slice step of corresponding axis in `axes`. Negative value means slicing backward. 'steps' cannot be 0. Defaults to 1s.</dd>
 </dl>
 
 #### Outputs
@@ -19795,7 +19813,7 @@ This version of the operator has been available since version 15 of the default 
   The end axis, if specified, is exclusive (and the returned value will not include the size of that axis).
   If the end axis is omitted, the axes upto the last one will be included.
   Negative axes indicate counting back from the last axis.
-  Note that axes will be clipped to the range [0, r-1], where r is the
+  Note that axes will be clamped to the range [0, r-1], where r is the
   rank of the input tensor if they are out-of-range (after adding r in the case of
   negative axis). Thus, specifying any end value > r is equivalent to specifying an end
   value of r, and specifying any start value < -r is equivalent to specifying a start
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -382,7 +382,7 @@ If start axis is omitted, the slice starts from axis 0.
 The end axis, if specified, is exclusive (and the returned value will not include the size of that axis).
 If the end axis is omitted, the axes upto the last one will be included.
 Negative axes indicate counting back from the last axis.
-Note that axes will be clipped to the range [0, r-1], where r is the
+Note that axes will be clamped to the range [0, r-1], where r is the
 rank of the input tensor if they are out-of-range (after adding r in the case of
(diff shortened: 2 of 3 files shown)


onnx/defs/tensor/defs.cc
```diff
@@ -769,18 +769,31 @@ ONNX_OPERATOR_SET_SCHEMA(
 static const char* Slice_ver13_doc = R"DOC(
 Produces a slice of the input tensor along multiple axes. Similar to numpy:
 https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html
-Slices uses `starts`, `ends`, `axes` and `steps` inputs to specify the start and end
```
gramalingam, line 772: I wonder if https://numpy.org/doc/stable/user/basics.indexing.html?highlight=slice#slicing-and-striding is a better link?

onnx/defs/tensor/defs.cc
```diff
@@ -769,18 +769,31 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+The effective values are initialized as `start[i] = 0`, `end[i] = dims[i]` where
+`dims` are the dimensions of `input` and `step[i] = `1.
+
+All negative elements of `axes` are made non-negatve by adding `r` to them, where
+`r =rank(input)`.
+
+All negative values in `starts[i]` and `ends[i]` have `dims[axes[i]]` added to them,
+where `dims` are the dimensions of `input`. Then `start[axes[i]]` is the adjusted
+`starts[i]` clamped into range `[0, dims[axes[i]]]`. The clamping for the adjusted
```
gramalingam, line 789: Should that be ```dims[axes[i]]-1``` ?

gramalingam: Thanks for the clarification. The spec is somewhat ambiguous. One interpretation would be that the sliced indices are all the values "start + j*step" that lie within the valid range obtained by clamping. The other interpretation would be to clamp start first, and generate values "clamped(start) + j*step". Do we know which is intended? Thanks!

onnx/defs/tensor/defs.cc
diyessi: A link to NumPy would make more sense for something preceded by "Similar to numpy".

onnx/defs/tensor/defs.cc
diyessi: It's `dim[axes[i]]` since you want `-1` to go to the last element.

The problem with negative steps is that there is really no clean way to stop at 0 since `end` is exclusive and negative values get shifted by the dimension size; you need to pick a really big negative end that is still below 0 when shifted by the size.

There is some code just below the doc that does some of this stuff; it looked similar to what I had implemented in our fork of glow based on experiments with numpy.

gramalingam: Thanks. Everything seems fine, except one point I am confused about: what happens if `starts[i] = dims[axes[i]]` and steps is negative ? Your new version says it will get clamped into the value `dims[axes[i]] - 1`. I tried it is onnxruntime, and this is what happens. But, then, this line in the shape-inference code doesn't seem to be consistent with that: https://github.com/onnx/onnx/blob/50a1981dc3d50af13075cec33b08b4c87fb0e41f/onnx/defs/tensor/defs.cc#L824

diyessi: In the PR I had changed that to `input_rank - 1`.

BTW, when looking for that I noticed a one-word typo/omitto which I just fixed.

gramalingam: Sorry, my fault. I saw that, but it didn't register, I thought the changes were only in the documentation, not the code. LGTM, thanks!

gramalingam (approved):

jcwchen (commented):
Thanks for the clarification! Please use [update_doc.sh](https://github.com/onnx/onnx/blob/main/tools/update_doc.sh) or [update_doc.bat](https://github.com/onnx/onnx/blob/main/tools/update_doc.bat) to update Operator.md as well.

gramalingam:
Thanks Jacky for the update!

jcwchen:
@diyessi Do you have bandwidth to finish this PR? If not, please allow me to send a duplicate PR of yours and go forward.  Thank you for your contribution.

diyessi:
> 

I might today; I need to re-figure out how to run the ONNX stuff (the README needs some updates). If I don't, feel free to jump in.

Do I just need to merge from main, run `update_doc.sh` and push back?

jcwchen:
> Do I just need to merge from main, run update_doc.sh and push back?

Thank you for the quick reply. Other than those, you will also need to fix the DCO issue because some of your commits were not sign-offed. A faster way to solve it is to squash all your old commits into one (for instance, git reset softly to the base and make/sign-off a single commit with your whole change).

diyessi:
@jwchen I think it's ready now.

jcwchen:
> @jwchen I think it's ready now.

Thank you for the quick update! Sorry that I might not be clear: actually using update_doc.sh will update `Operators.md` and `Changelog.md`. The CI failed because `Changelog.md` was not updated. Please use update_doc.sh to update `Changelog.md` as well and the PR should be ready to go.

diyessi:
> > @jwchen I think it's ready now.
> 
> Thank you for the quick update! Sorry that I might not be clear: actually using update_doc.sh will update `Operators.md` and `Changelog.md`. The CI failed because `Changelog.md` was not updated. Please use update_doc.sh to update `Changelog.md` as well and the PR should be ready to go.

It actually updates a lot of things when I run it. Hopefully this time it will all be good.

onnx/defs/tensor/defs.cc
```diff
@@ -878,7 +894,7 @@ ONNX_OPERATOR_SET_SCHEMA(
             "steps",
             "1-D tensor of slice step of corresponding axis in `axes`. "
             "Negative value means slicing backward. 'steps' cannot be 0. "
-            "Defaults to 1.",
+            "Defaults to 1s.",
```
jcwchen, line 897: Thanks for the update. The CIs now all passed! 

Is this `1s` here (L897) intended or a typo?

diyessi: A slice can have more than one thing in it, so "1s" as in `(1, 1, ...)`

jcwchen (approved):

docs/Changelog.md
```diff
@@ -18040,19 +18040,37 @@ This version of the operator has been available since version 13 of the default
...
+  `r =rank(input)`.
+
+  All negative values in `starts[i]` and `ends[i]` have `dims[axes[i]]` added to them,
+  where `dims` are the dimensions of `input`. Then `start[axes[i]]` is the adjusted
+  `starts[i]` clamped into range of valid indices, i.e. `[0, dims[axes[i]]-1]`.
+
+  The clamping for the adjusted `ends[i]` depends on the sign of `steps[i]` and must
+  accommodate copying 0 through `dims[axes[i]]` elements, so for positive stepping
+  `end[axes[i]]` is clamped to `[0, dims[axes[i]]]`, while for negative stepping it
+  is clamped to `[-1, ends[i]-1]`.
```
hariharans29, line 18067: Should this read `dims[axes[i]]-1` or am I mis-understanding something ?

diyessi: Starts are inclusive, ends are exclusive, so the slice covers indices `{i | start <= i < end}`.
If it still seems wrong or you see a way to make it clearer, let me know.

hariharans29: Sorry for not being clear enough. I think the logic is right. Looking at it, it seems inconsistent with the way it is `worded`. 

`...so for positive stepping `ends[axes[i]]` is clamped to `[0, dims[axes[i]]]`

Looking at the above line, I would have thought, the last set of words should be `.... is clamped to [-1, dims[axes[i]] - 1]` instead of the `ends` you are referencing. Does this make sense or am I on the wrong track ? :)

diyessi: Let's say we have a tensor X[5]. Then a `slice(X, start=0, end=0)` is empty, `slice(X, start=0, end=5)` is all of X, and `slice(X, start=0, end=101086)` is the same as `slice(X, start=0, start=5)`.
In ranges, `[` and `]` are inclusive, `(` and `)` are exclusive.

hariharans29: I agree with everything above. 

I was just pointing out the clamping wording while negative stepping:
`[-1, ends[i]-1]`   ->   `[-1, dims[axes[i]] - 1]` ?

gramalingam: @hariharans29 : I agree with your comment about the wording, which seems like a typo. More importantly, I was wondering about the issue you created https://github.com/onnx/onnx/issues/4115 ... do you still have a concern about this? It seemed like an important point, but I haven't checked it out yet.

hariharans29: Thanks Rama, yes, it seems like a typo.

Re: #4115 - To have the link to Numpy's slicing in the op's description and prescribe a different clamping for `start` is a bit misleading. ORT implements the Numpy clamping (which does the clamping based on the slicing direction) for `start` since the time `steps` input was introduced in Slice-10. This is because the spec initially didn't have the clamping logic called out explicitly and I used Numpy's clamping as the de-facto (Numpy's Slice was also linked to in the op's spec).  FWIW- The "real-world" impact of this should be minimal. It is just that ORT will need to make changes if it needs to "honor" this newly called out clamping logic. As much as this discussion is more pedantic than a burning issue, it would be nice to clarify this.

diyessi: I hadn't seen #4115 and thought you were referring to the positive stepping case. Your change looks correct and agrees with what I implemented in our glow fork (although I used the equivalent adjusted start instead of `dim[axes[i]]-1`, so that's two implementations.

gramalingam: The original textual description was unclear. However, how about the original shape-inference code?
(a) Is the original shape-inference code consistent with the ORT/glow implementation or not?
(b) Does the change in shape-inference code in this PR change the behavior or does it not? This line of change seems relevant to the kind of counterexample that Hari discusses in the above issue, but I feel there may be edge-cases where the shape-inference code may not be consistent, and it would be helpful to check this.

hariharans29: This was the original shape inference code:
 
 if (step < 0)
    start = clamp(start, 0, input_rank - 1);
  else
    start = clamp(start, 0, input_rank);

This is how ORT implements it: https://github.com/microsoft/onnxruntime/blob/26fceca90f35fe94f39864ebaec0f2d51e367f10/onnxruntime/core/providers/cpu/tensor/slice_helper.h#L132

So, the old shape inference code matches with ORT's implementation.

gramalingam: So, I think we should revert to the old shape-inference code. Is there any case where the old shape-inference code is incorrect?

gramalingam: And change the textual description accordingly. (Since we are not bumping the opset version, this PR should be only a clarification, not a change in behavior.)

gramalingam: Does that sound reasonable? Thanks!

diyessi: I'm not seeing how the step > 0 case above is correct; start is inclusive, which is [0, input_rank), not [0, input_rank+1].

diyessi: NumPy does behave like the original.

hariharans29: @gramalingam : I think reverting to the old shape inference will be a good idea. It atleast maintains status quo and as you rightly say changing any behavior without an opset bump up is not a good idea.

@diyessi : Try this example on Numpy (this is for step > 0),
x[5:6:1]
array([], dtype=int32)

If `start` had been clamped to an inclusive value as you are suggesting, with the `ends` clamping the way we all agree it to be, the result would actually be non-empty

PR #3915, opened 2022-01-03
add a local function test case

liqunfu:
**Description** 
add a local function test case

- Describe your changes.
refactor make_function helper method
add test case for models with local functions

**Motivation and Context**
- improve runtime test coverage of onnx local function specification


askhade:
@liqunfu : DCO check is failing for this PR because the commits are not signed. including "-s" option in every git commit cmd enables sign off. 
```
git commit -s -m "my commit"
```

You can check your local logs to make sure the commit is signed. If you push more than 1 unsigned commits there is no way to go back and sign the only option I know is to resend a new PR. Unfortunately, this is the current limitation.

onnx/backend/test/case/model/local-functions.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @staticmethod
+    def export():  # type: () -> None
+
+        ops = [
+            helper.make_opsetid("", 10),
+            helper.make_opsetid("local", 10)
+        ]
+
+        g = GraphProto()
```
askhade, line 30: I suggest using onnx parser api for test model creation. It is more readable and easier to use.

askhade:
Can you add details around what exactly this test covers.

liqunfu:
close this pr.
Local function is better tested be runtime (e.g. https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/test/ir/onnx_model_test.cc).

PR #3939, opened 2022-01-13
mypy: update to 0.760 and remove vendored protobuf stubs

stillmatic:
**Description**
- We can remove the stubs in favor of types-protobuf package
- `mypy==0.600` doesn't install properly for me (see https://github.com/python/typed_ast/issues/97). The newer version of mypy here does install but doesn't change much logic otherwise.

**Motivation and Context**
- See https://github.com/onnx/onnx/pull/3604 for more context, broadly seeking to update mypy to a more recent version
- There is quite a lot of changes that have to be done in order to do the upgrade properly.
- This is a small piece of that work, broken out here as a smaller chunk and easier to merge individually.
- I also believe that it will be more reasonable to upgrade slowly since each new version of mypy brings new behavioral changes.


setup.py
```diff
@@ -330,7 +330,7 @@ def run(self):
 
 if sys.version_info[0] == 3:
```
askhade, line 331: can remove this if since onnx does not support python 2 anymore

stillmatic: yep - was removed in the other commit, let me fix the merge conflict

askhade (approved):

PR #3941, opened 2022-01-13
Bump to 10.15 in AzurePipeline because 10.14 was deprecated

jcwchen:
**Description**
- Bump to 10.15 in AzurePipeline 
- There are two more places need branch renaming

**Motivation and Context**
Mac-CI in AzurePipeline is failing due to unsupported 10.14

### DIFF
--- a/.azure-pipelines/MacOS-CI.yml
+++ b/.azure-pipelines/MacOS-CI.yml
@@ -4,7 +4,7 @@ trigger:
 jobs:
 - job: 'Test'
   pool:
-    vmImage: 'macOS-10.14'
+    vmImage: 'macOS-10.15'
   strategy:
     matrix:
       py36:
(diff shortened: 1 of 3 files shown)


askhade (approved):
Thanks for this change!

askhade:
Wondering how we can catch this sooner next time... meaning before the pipeline fails. I would expect to see some warning in the CI pipelines to let the users know a certain image is going to be deprecated.

jcwchen:
> Wondering how we can catch this sooner next time... meaning before the pipeline fails. I would expect to see some warning in the CI pipelines to let the users know a certain image is going to be deprecated.

Actually it did in previous CI logs: `##[warning]The macOS-10.14 environment is deprecated and will be removed on December 10, 2021. For more details see https://devblogs.microsoft.com/devops/hosted-pipelines-image-deprecation/`

Mac-OS upgrading in CI was in my TO-DO list, but I was OOF at that time and then just forgot... We should take an action for upgrade in time next time if we notice the warning. Thanks for the reminder this time!

PR #3954, opened 2022-01-19
Bugfix extractor misses local functions 

liqunfu:
**Description**
- Fix issue 3938 extract a model with local functions
- Add function test cases

**Motivation and Context**
- bug fix
- improve test coverage


onnx/utils.py
```diff
@@ -79,6 +79,14 @@ def _collect_reachable_nodes(
         nodes = [n for n in self.graph.node if n in reachable_nodes]
         return nodes
 
+    def _collect_referred_local_functions(
+            self,
+            nodes,  # type: List[NodeProto]
+    ):  # type: (...) -> List[FunctionProto]
+        node_op_types = {n.op_type for n in nodes}
+        local_functions = [f for f in self.model.functions if f.name in node_op_types]
```
gramalingam, line 87: I think we may need to check both (domain, op_type) to match calling-node with called-function?

onnx/utils.py
```diff
@@ -79,6 +79,14 @@ def _collect_reachable_nodes(
...
+        return local_functions
```
gramalingam, line 88: I think we may need a loop to accumulate functions called from other functions. Main-graph may call local function foo1, and function foo1 may contain a call to local function foo2, and so on. We should collect foo2 also.

liqunfu: good point. coded for this case

onnx/helper.py
```diff
@@ -171,6 +172,24 @@ def make_opsetid(domain, version):  # type: (Text, int) -> OperatorSetIdProto
     return opsetid
 
 
+def make_function(
+    domain,  # type: Text
+    fname,  # type: Text
+    inputs,  # type: List[Text]
+    outputs,  # type: List[Text]
+    nodes,  # type: List[NodeProto]
+    opset_imports,    # type: Sequence[OperatorSetIdProto]
```
gramalingam, line 181: Suggest adding other fields in FunctionProto, with default-values? Default empty list may make sense for opset_imports also.
```python
   attributes = [],
   doc_string = ''
```

gramalingam (approved):
LGTM, thanks! Just have a minor suggestion for `make_function`

Issue #3842, opened 2021-11-16
Add bfloat16 to all relevant ops


fdwr:
# Bug Report

### Is the issue related to model conversion?
No.

Looks like an oversight, as it's inconsistent with both [`Less`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Less) and [`Equal`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Equal) which support bfloat16 but their combination does not, especially given [`LessOrEqual`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#LessOrEqual) can be implemented as an aggregate operator ("function") of those two. The same applies to [`GreaterOrEqual`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#GreaterOrEqual).

### Describe the bug
LessOrEqual and GreaterOrEqual should support bfloat16. Pertinent PR https://github.com/onnx/onnx/pull/2770.

### System information
- ONNX version:  master 1.10

fdwr:
FYI @gramalingam, @souptc.

fdwr:
Just noticed also that [`Sigmoid`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Sigmoid) has `float16` but not [`HardSigmoid`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#HardSigmoid), even though all the constituent operators (max, min, mul, add) do support bfloat16.

gramalingam:
Makes sense. Looks like an oversight. We should add the missing bfloat16 type to all such ops. Would be nice to find all such ops in one go! (Eg., we had to do this for Pow and Where previously I think.)

gramalingam:
[...]

PR #3960, opened 2022-01-21
Add bfloat16 type to a few ops missing it

gramalingam:
**Description**
Add bfloat16 type to a few ops missing it (LessOrEqual, GreaterOrEqual, Loop, If, Scan, LeakyRelu, PRelu).

See issue 3842: https://github.com/onnx/onnx/issues/3842

### DIFF
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -5,7 +5,7 @@ jobs:
 
 - job: 'Test'
   pool:
-    vmImage: 'vs2017-win2016'
+    vmImage: 'windows-2019'
   strategy:
     matrix:
       py37:
--- a/README.md
+++ b/README.md
@@ -18,27 +18,27 @@ ONNX is [widely supported](http://onnx.ai/supported-tools) and can be found in m
 * [Pre-trained ONNX models](https://github.com/onnx/models)
 
 # Learn about the ONNX spec
-* [Overview][overview]
-* [ONNX intermediate representation spec][ir]
-* [Versioning principles of the spec][versioning]
-* [Operators documentation][operators]
-* [Python API Overview][python_api]
+* [Overview](docs/Overview.md)
+* [ONNX intermediate representation spec](docs/IR.md)
+* [Versioning principles of the spec](docs/Versioning.md)
+* [Operators documentation](docs/Operators.md)
+* [Python API Overview](docs/PythonAPIOverview.md)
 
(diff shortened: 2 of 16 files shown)


onnx/defs/math/defs.cc
```diff
@@ -536,18 +536,21 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
 
-static const char* LeakyRelu_ver6_doc = R"DOC(
+static const char* LeakyRelu_ver16_doc = R"DOC(
 LeakyRelu takes input data (Tensor<T>) and an argument alpha, and produces one
 output data (Tensor<T>) where the function `f(x) = alpha * x for x < 0`,
 `f(x) = x for x >= 0`, is applied to the data tensor elementwise.
+
+**History**
+- Version 16 adds bfloat16 to the types allowed.
```
jcwchen, line 545: I like providing a history section, which makes the change clearer. Just out of curiosity: Why only LeakyRelu and PRelu add history section?

jcwchen (approved):
LGTM. Thanks!

askhade (approved):

onnx/defs/math/defs.cc
gramalingam: Good question! For the control-flow ops, especially, I was thinking about another issue: they have very long documentation, and we end up copy-pasting it when we update the op even though the documentation doesn't change much, which also presumably increases binary size etc. I was thinking a slightly better solution might be needed, if we wanted to avoid this kind of increase. Something to think about. Thanks!

fdwr: Oh my goodness, a history section in the actual op would be so useful. I spend so much time digging through dozens of PR's each ONNX update and comparing Changelog.md entries to see what actually changed between ops for a given version.

askhade: @fdwr : I agree it is cumbersome to find out op updates. In short term you can refer to the release wiki pages that we create to track op validation. This one is from last release :  https://github.com/onnx/onnx/wiki/Logistics-for-ONNX-Release-1.10#changelog
We will create a similar one for 1.11 soon. This table has a short description on what change and link to onnx PR which made the update.

fdwr: @askhade : 👍 That wiki looks nice. I read the generic notes on the Release page, but the wiki has more notes beyond just the PR title and a link.

PR #3962, opened 2022-01-22
Use Python type annotations rather than comments

garymm:
These have been supported since Python 3.5.

ONNX doesn't support Python < 3.6, so we can use the annotations.

Diffs generated by https://pypi.org/project/com2ann/.


garymm:
@liqunfu could you please review?

garymm:
@gramalingam could you please approve?

gramalingam (approved):

PR #3966, opened 2022-01-24
mypy: update to 0.782 and enable it with Python 3.8 and 3.9

### LINKED ISSUE #3968: mypy typecheck (mypy version now is 0.760 in main branch) cannot work with Python 3.8 and 3.9
# Bug Report
mypy typecheck (mypy 0.760 in main branch) cannot work with Python 3.8 and 3.9. It will throw few check errors. It is not caught by CI (AzurePipeline) because it only uses Python 3.6 and 3.7 with mypy typecheck. We should add 3.8 and 3.9 in AzurePipeline to cover it.

### Reproduction instructions
```
python setup.py typecheck
...
```

### Expected behavior
Should not produce any error.

jcwchen:
**Description**
- Update to mypy 0.782
- Fix mypy failures for Python 3.8 and 3.9
- Add Python 3.8 and 3.9 in AzurePipeline to cover this mypy typecheck for Python 3.8 and 3.9

**Motivation**
Closes #3968. Current mypy version 0.760 cannot pass by Python 3.8 and Python 3.9

### DIFF
--- a/onnx/backend/test/report/coverage.py
+++ b/onnx/backend/test/report/coverage.py
@@ -159,7 +159,8 @@ class Coverage(object):
         if os.path.isfile(nodes_path):
             with open(nodes_path, 'r') as nodes_file:
                 reader = csv.DictReader(nodes_file)
-                frameworks = list(reader.fieldnames)  # type: ignore
+                assert reader.fieldnames
+                frameworks = list(reader.fieldnames)
                 for row in reader:
                     op = row[str('Op')]
                     del row[str('Op')]
--- a/onnx/backend/test/report/coverage.py
+++ b/onnx/backend/test/report/coverage.py
@@ -20,11 +20,11 @@ _all_schemas = defs.get_all_schemas()
 
 
 class AttrCoverage(object):
-    def __init__(self):  # type: () -> None
-        self.name = None  # type: Optional[Text]
-        self.values = set()  # type: Set[Text]
+    def __init__(self) -> None:
+        self.name: Optional[Text] = None
+        self.values: Set[Text] = set()
 
-    def add(self, attr):  # type: (onnx.AttributeProto) -> None
+    def add(self, attr: onnx.AttributeProto) -> None:
         assert self.name in [None, attr.name]
         self.name = attr.name
         value = helper.get_attribute_value(attr)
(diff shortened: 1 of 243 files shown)


onnx/helper.py
```diff
@@ -865,10 +865,10 @@ def printable_graph(graph, prefix=''):  # type: (GraphProto, Text) -> Text
...
     # body
     for node in graph.node:
-        pn, gs = printable_node(node, indent, subgraphs=True)
-        assert isinstance(gs, list)
-        content.append(pn)
-        graphs.extend(gs)
+        contents_subgraphs = printable_node(node, indent, subgraphs=True)
+        assert isinstance(contents_subgraphs[1], list)
+        content.append(contents_subgraphs[0])
+        graphs.extend(contents_subgraphs[1])
```
jcwchen, line 871: L868 and L800 both bumped into the following error: `error: Unpacking a string is disallowed` so I used single variable to catch it.

tools/protoc-gen-mypy.py
```diff
@@ -120,7 +120,7 @@ def _import_message(self, type_name):
 
     @contextmanager  # type: ignore
     def _indent(self):
-        # type: () -> Generator
+        # type: () -> Generator[None, None, None]
```
jcwchen, line 123: Fix the error: `Missing type parameters for generic type "Generator"`

onnx/backend/test/report/coverage.py
```diff
@@ -159,7 +159,7 @@ def report_csv(self, all_ops, passed, experimental):  # type: (List[Text], List[
         if os.path.isfile(nodes_path):
             with open(nodes_path, 'r') as nodes_file:
                 reader = csv.DictReader(nodes_file)
-                frameworks = list(reader.fieldnames)
+                frameworks = list(reader.fieldnames)  # type: ignore
```
jcwchen, line 162: The error is `Argument 1 to "list" has incompatible type "Optional[Sequence[str]]"; expected "List[str]"`.
I tried to make `reader.fieldnames` as `List Optional[Sequence[str]]` or others, but it still showed error anyway. Please let me know if there is a better way to solve it. Thanks!

garymm: The problem is that `reader.fieldnames` may be None. We can assert that it's not to avoid the type error:

```suggestion
                assert reader.fieldnames
                frameworks = list(reader.fieldnames)
```

garymm (approved):
LGTM with one suggestion, thanks!

onnx/backend/test/report/coverage.py
jcwchen: Nice trick. I have updated and it worked. Thank you for the suggestion!

askhade (approved):

Issue #3906, opened 2021-12-21
Extend QuantizeLinear and DequantizeLinear 


volcacius:
# Feature Request
This should probably be split into many separate issues, but I think it would be worth having a general discussion before to see what's the general sentiment on the topic.

There an ongoing trend towards representing quantization only with QuantizeLinear and DequantizeLinear, which enables greater flexibility in terms of quantization pattern that can be represented compared to the traditional op oriented approach (e.g. QLinearConv). 

At the same time, currently QuantizeLinear and DequantizeLinear exibhit a few limitations. Off the top of my head:

1) There isn't support for quantizing to uint32/int32. 
2) Support for dequantizing 32b is limited to int32 with zero-point == 0 (basically just for biases).
3) There isn't support for representing uint16/int16 quantization.
4) There isn't support for a "narrow-range" representation, e.g. [-127, 127] for int8 instead of [-128, 127].
5) There isn't support for block-wise scale/zero-point.
6) There isn't support for decoupling the container byte-alligned datatype ((u)int8/16/32) from the "virtual" datatype that comes from constraints imposed at quantization time that might not be byte alligned. This is particuarly relevant to below-8b quantization.

### System information
ONNX version (you are using): 1.10.2

### What is the problem that this feature solves?
1) Quantizing to 32b is a very common operation, e.g. ONNX implicitly assumes it for biases. I see value in making it explicit.
2) Dequantizing a 32b integer with non-zero zero point can be useful to e.g. dequantize the output accumulator of a layer.
3) (u)int16 can be used in a lot of interesting way. As a reduced precision datatype for biases. As a datatype for intermediate accumulators. I see value in adding explicit support for it.
4) Narrow range brings symmetry in the signed case, which enables e.g. merging change of signs into weights. It also helps prevent overflow in certain corner cases.
5) Certain hardware backends internally work at a block-of-channels level. Having block-wise quantization brings some of the benefit of channel-wise quantization with a more hardware friendly approach. 
6) Below 8b datatypes are supported in many hardware platforms, and there are many successful algorithmic approach to go below 8b. PyTorch actually deploys 7b quantization ("reduced range" in their codebase) to avoid overflow on certain CPU backends. This kind of information should be captured somehow. 

### Describe the alternatives you have considered
Nothing upstream.

### Describe the feature
See above.

### Will this influence the current api?
Features 1 to 5 wouldn't really change much. 6 is more complicated one since it virtualizes over byte boundaries.  

### Feature Area
Operators.

[...]

PR #3971, opened 2022-01-26
Add support for 16-bit quantization, in QuantizeLinear and DequantizeLinear

frank-ff:
**Description**
- Adds support for signed and unsigned 16-bit quantization in the QuantizeLinear and DequantizeLinear ops
- Generated new documentation via update_doc.sh script

**Motivation and Context**
- Today, ONNX only supports 8-bit quantization. 
- This change hopes to bring ONNX up to speed with the major NN frameworks and libraries that all support 16-bit.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -19860,6 +19860,50 @@ This version of the operator has been available since version 15 of the default 
 </dl>
 
 ## Version 16 of the default ONNX operator set
+### <a name="DequantizeLinear-16"></a>**DequantizeLinear-16**</a>
+
+  The linear dequantization operator. It consumes a quantized tensor, a scale, and a zero point to compute the full precision tensor.
+  The dequantization formula is y = (x - x_zero_point) * x_scale. 'x_scale' and 'x_zero_point' must have same shape, and can be either a scalar
+  for per-tensor / per layer quantization, or a 1-D tensor for per-axis quantization.
+  'x_zero_point' and 'x' must have same type. 'x' and 'y' must have same shape. In the case of dequantizing int32,
+  there's no zero point (zero point is supposed to be 0).
+
+#### Version
+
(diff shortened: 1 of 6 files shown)


yufenglee:
Which are the major NN frameworks and libraries that you mentioned in the description? Do they get better performance?

frank-ff:
Hi @yufenglee 

1. To answer your first question, here are some examples:

Tensorflow allows for arbitrary bit-width quantization (between 2 - 16 bits) : https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/fake-quant-with-min-max-args

TVM allows for 16-bit quantization as well, and in fact, is the default quantization width for certain op types: https://tvm.apache.org/docs/how_to/deploy_models/deploy_prequantized.html?highlight=16 (see the footnote that says "Without fast 8 bit instructions, TVM does quantized convolution in 16 bit")

CMSIS-NN has 16-bit operator support for most of its NN operators: https://www.keil.com/pack/doc/CMSIS/NN/html/group__groupNN.html


2. To answer your second question, yes, having higher granularity bitwidths in quantization helps increase accuracy at the cost of larger model sizes (but still better than full unquantized float32 models) -- this has been incredibly valuable in the models I've worked on personally. And in fact, scanning through the Issues list, it seems I'm not the only one: https://github.com/onnx/onnx/issues/3906

frank-ff:
@gramalingam @yufenglee Any further comments or thoughts on this PR?

sousoux:
Could I suggest that you add an extra levels attribute to the quantize and dequantize operator which matches the definition of levels in https://docs.openvino.ai/latest/openvino_docs_ops_quantization_FakeQuantize_1.html? It can default to the size of the container i.e. pow(2,8) for int8 and pow(2, 16) for int16. This would be very helpful for us.

jywu-msft:
Can we revive this?

PR #3973, opened 2022-01-26
bump version to 1.11.0

liqunfu:

**Description**
- prepare for 1.11.0 release


onnx/test/helper_test.py
```diff
@@ -348,6 +348,7 @@ def test(opset_versions, ir_version):  # type: (List[Tuple[Text, int]], int) ->
         test([("", 13)], 7)
         test([("", 14)], 7)
         test([("", 15)], 8)
+        test([("", 16)], 8)
         # standard opset can be referred to using empty-string or "ai.onnx"
         test([("ai.onnx", 9)], 4)
         test([("ai.onnx.ml", 2)], 6)
```
jcwchen, line 354: nit: perhaps we can add a test for ml-opset like
```
test([("ai.onnx.ml", 3)], 8)
```

onnx/defs/schema.h
```diff
@@ -1004,7 +1004,7 @@ class OpSchemaRegistry final : public ISchemaRegistry {
       // Version corresponding last release of ONNX. Update this to match with
       // the max version above in a *release* version of ONNX. But in other
       // versions, the max version may be ahead of the last-release-version.
-      last_release_version_map_[ONNX_DOMAIN] = 15;
+      last_release_version_map_[ONNX_DOMAIN] = 16;
       last_release_version_map_[AI_ONNX_ML_DOMAIN] = 3;
```
jcwchen, line 1008: Not relevant to this PR: although last release version of AI_ONNX_ML_DOMAIN has been already updated a few days ago, isn't it better to upgrade `last_release_version_map_[AI_ONNX_ML_DOMAIN] = 3` now right before the release?

liqunfu: Similar thing happened with IR version change before 1.10.0 release. IR version was increased to 8 in a PR adding sparse tensor support. Optional data type was added after but before 1.10.0 release. Neither the optional data pr nor the release version bump pr did update the IR version because it was set correct with the first PR. I think it makes sense. A user may use any build from the main branch. Updating the version with first PR will inform the user a possible breaking change in the build.

jcwchen: I agreed that bumping version (like `map_[AI_ONNX_ML_DOMAIN]` above) in the first related PR is good as a notification to users, but bumping the version in `last_release_version_map_` might be quite different. Based on the naming, it seems to me it's better to bump it right before the release. @gramalingam since you are the contributor of it, do you have any suggestion about when to bump the version in `last_release_version_map_`? Thanks!

liqunfu: for example, an op is updated. It needs to be validated in ort with build from onnx main. without increase the op domain last rel version, the validation will fail.

liqunfu:
@onnx/sig-archinfra-approvers @onnx/sig-operators-approvers Kindly remind that this PR is needed for the release cut planned for Jan 31.

gramalingam (approved):

gramalingam (approved):

Issue #3964, opened 2022-01-22
Upgrade mypy to latest version


stillmatic:
# Feature Request

### System information
ONNX version (you are using): latest

### What is the problem that this feature solves?
The current 

### Describe the alternatives you have considered
N/A

### Describe the feature
The current mypy version on main is 0.760 (17 December 2019), we should upgrade to a more recent version and take advantage of the improvements.

### Will this influence the current api?
Unlikely.

### Feature Area
Which area in ONNX does this impact?  (*e.g. model usage, backend, best practices, converters, shape_inference, version_converter, training, test, operators*): 

### Are you willing to contribute it (Y/N):
I will not likely have substantial additional bandwidth for this.

### Notes
Previous PRs:
* https://github.com/onnx/onnx/pull/3604 -- initial attempt at upgrading everything in one go
* https://github.com/onnx/onnx/pull/3926 -- removing six usages and python 2 support (so no more special casing typechecks)
* https://github.com/onnx/onnx/pull/3939 -- upgraded from 0.600 -> 0.760 and removed vendored protobufs

asears:
As part of the upgrade, should the typing documentation for contributors be updated, if not currently supporting python 2?

https://github.com/onnx/onnx/blob/main/docs/TypeAnnotations.md

![image](https://user-images.githubusercontent.com/7253192/151124023-ba68f23a-3b22-4c58-b3e3-48efa45d889a.png)

Might need some blanket issue to improve the typing from comments to code if necessary...

stillmatic:
Yep the docs should definitely be reworked.


> improve the typing from comments to code
Should be able to do so via https://github.com/ilevkivskyi/com2ann

jcwchen:
ONNX should pass the latest mypy (0.971) after this PR: https://github.com/onnx/onnx/pull/4495. I will close this one. Thanks!

PR #3974, opened 2022-01-26
typing: run com2ann

stillmatic:
**Description**
- runs [com2ann](https://github.com/ilevkivskyi/com2ann) to automatically convert typing hints to annotations
- specifically, ran `com2ann onnx/.`

**Motivation and Context**
- Current typehints are done in Python 2 comment style, we only support Python 3 now
- see also https://github.com/onnx/onnx/issues/3964#issuecomment-1021949910

### DIFF
--- a/onnx/__init__.py
+++ b/onnx/__init__.py
@@ -27,7 +27,7 @@ from typing import Union, Text, IO, Optional, cast, TypeVar, Any
 
 
 # f should be either readable or a file path
-def _load_bytes(f):  # type: (Union[IO[bytes], Text]) -> bytes
+def _load_bytes(f: Union[IO[bytes], Text]) -> bytes:
     if hasattr(f, 'read') and callable(cast(IO[bytes], f).read):
         s = cast(IO[bytes], f).read()
     else:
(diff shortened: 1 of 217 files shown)


stillmatic:
```
▶ python onnx/backend/test/cmd_tools.py
  File "onnx/backend/test/cmd_tools.py", line 13, in <module>
    import onnx.backend.test.case.node as node_test
  File "/Users/hua/Development/onnx/onnx/backend/test/__init__.py", line 9, in <module>
    from .runner import Runner as BackendTest # noqa
  File "/Users/hua/Development/onnx/onnx/backend/test/runner/__init__.py", line 23, in <module>
    from onnx import helper, numpy_helper, NodeProto, ModelProto, TypeProto
  File "/Users/hua/Development/onnx/onnx/numpy_helper.py", line 21, in <module>
    def to_array(tensor: TensorProto, base_dir: Text = "") -> np.ndarray[Any]:
TypeError: Type subscription requires python >= 3.9
```

looks related to Type Hinting Generics (https://www.python.org/dev/peps/pep-0585/) so unfortunately this is likely not scalably convertable in the short term

garymm:
FYI did this in https://github.com/onnx/onnx/pull/3962

PR #3981, opened 2022-01-31
Fix LGTM C++ failure: always use Python3 for lgtm C/C++

jcwchen:
**Description**
- Use Python 3 to build ONNX in LGTM.com; Meanwhile it will need building protobuf from source to prevent issue of shared symbol cannot be found

**Motivation and Context**
lgtm.com for C++ is failing. Recently ONNX removed couple of things related to Python 2 because ONNX has deprecated Python 2 for a while. However, LGTM uses Python 2 by default: https://github.com/github/securitylab/discussions/383. ONNX should use Python 3 in all CIs so this PR forces LGTM C/C++ check to use Python 3.


.lgtm.yml
```diff
@@ -0,0 +1,17 @@
+extraction:
```
askhade, line 1: since this is a new file, please add comments why this file is need and why we are building protobuf from source.

.lgtm.yml
```diff
@@ -0,0 +1,17 @@
...
+  cpp:
+    before_index:
+      # always use Python 3
+      - cp `which python3` $LGTM_WORKSPACE/python
+      - PATH=$LGTM_WORKSPACE:$PATH
+      # build protobuf from source with -fPIC
+      - git clone https://github.com/protocolbuffers/protobuf.git
+      - cd protobuf
+      - git checkout v3.16.0
```
askhade, line 10: when we update protobuf to a next version we will need to update all these places... wondering whether there is an easier way to fetch the version from a common place... not implying that this needs to be done in this PR.

jcwchen: Yes it is a burden when every time ONNX bumps protobuf version... Also I was considering to have a common script for building protobuf from source since we have quite many different CIs using the same commands.  I will add this investigation in my TODO list.

askhade (approved):

PR #3982, opened 2022-01-31
Python clean up

garymm:
* Remove `__future__` imports that are no-ops on Python 3. This code
  does not support Python 2 anymore so they're not needed.
* Remove some unused imports.
* Fix a couple things that trigger warnings on pytest:
    * In function_test.py, mark a string as raw so that the invalid
      escape sequence warning is not triggered.
    * In test_case.py, set `TestCase.__test__ = False` so that PyTest
      knows it's not a runnable test.
* Use types rather than strings for type annotations. Possible since we dropped support for Python 3.6.


garymm:
@postrational @linkerzhang would you mind approving?

garymm:
@gramalingam could you please approve?

gramalingam (approved):

PR #3987, opened 2022-02-02
[CI] Simplify CI pipelines: Add common scripts for building protobuf from source

jcwchen:
**Description**

- [x] Add a common script `build_protobuf_unix.sh` for building protobuf from source on Linux and Mac
- [x] Add a common script `build_protobuf_win.ps1` for building protobuf from source on Windows

**Motivation and Context**
ONNX now has quite a few CIs and there are many places using the same code. For instance, build protobuf from source with certain version of protobuf. Having a common script can simplify CIs and also make it easier to bump protobuf version in the future.


workflow_scripts/protobuf/build_protobuf_unix.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+if [[ -z "$INSTALL_PROTOBUF_PATH" ]]; then
+   export INSTALL_PROTOBUF_PATH=/usr
+fi
+
+# Build protobuf from source with -fPIC on Unix-like system
+ORIGINAL_PATH=$(pwd)
+cd ..
+git clone https://github.com/protocolbuffers/protobuf.git
```
snnn, line 19: Fetch the tarball would be simpler, which requires less network traffic than "git clone". When you run "git clone", you download all the versions.

snnn (approved):

askhade (approved):

PR #3988, opened 2022-02-03
Add `tools/style.sh` to run flake8 and mypy

### LINKED ISSUE #3936: Is this concat operation supported by ONNX?
# Source
PaddlePaddle->onnx->onnx-sim 
onnx version:11.0
# Describe
The input feature graph and the two learned feature graphs are spliced in dimension.
# New Operator
concat
# one of the two learned feature graphs info:
a35e8db6072a06bf6ffb120c7e5df03
# Confused point
How does this work? 
Any suggestions if I need to refactor it with traditional operations by caffe？

# Thanks

I hope I can get your help！
Thanks !

xuzijian629:
**Description**
You can run python check and format c++ code just by running ~`./tools/format.sh`~ `./tools/style.sh`.
Since version difference of `clang-format` may cause different format results, I specified it in 704108651649fdc32e360b0da93d19ddaeae9cbf.

**Motivation and Context**
- Resolves #3936 

**Request for comments**
I think we need following things to discuss. Please share your ideas.
* There is already [tools/mypy-onnx.py](https://github.com/onnx/onnx/blob/6a2e9c12e13075ba169d17dcc19073d1a30b1ee7/tools/mypy-onnx.py) that will be internally called by running `python setup.py typecheck`. Do you think it will be better to remove this file and the command? Or, do you think it's better to prepare similar files and commands like `python setup.py lint`?
* There is an awesome lint tools called [pysen](https://github.com/pfnet/pysen). It runs multiple lint tools like flake8, black, isort, clang-format, cmake-format at the same time just by `pysen run lint`. But since we don't currently have such a complex lint settings, I think `format.sh` is enough for now (also, pysen command is usually slow...).

**TODO**
- [ ] Apply clang-format
- [x] Replace CI commands if necessary
- [x] Update CI Pipeline and its docs
- [x] Add explanation to CONTRIBUTING.md


setup.py
```diff
@@ -327,7 +327,12 @@ def run(self):
 tests_require.append('nbval')
 tests_require.append('tabulate')
 
-extras_require['mypy'] = ['mypy==0.782', 'types-protobuf==3.18.4']
+extras_require['lint'] = [
+    'clang-format==13.0.0',
+    'flake8',
+    'mypy==0.782',
+    'types-protobuf==3.18.4',
+]
```
jcwchen, line 335: As you mentioned in the PR description, I think adding a new command like "python setup.py lint" and removing existing typecheck should be a good idea.

jcwchen (commented):
Thank you for the contribution!

To fix CI failures, now these 2 CIs are running mypy typecheck ([Linux-CI.yml](https://github.com/onnx/onnx/blob/main/.azure-pipelines/Linux-CI.yml), [Windows-CI.yml](https://github.com/onnx/onnx/blob/main/.azure-pipelines/Windows-CI.yml)) and ([Windows-CI.yml](https://github.com/onnx/onnx/blob/main/.azure-pipelines/Windows-CI.yml), [Mac-CI.yml](https://github.com/onnx/onnx/blob/main/.azure-pipelines/MacOS-CI.yml)). You can try to utilize new format.sh on these CIs first. 

For others' reference: this PR will probably bring clang-format into CI testing as well which ONNX CIs don't have today. Personally I think it's a good thing (the current code passes .clang-format we have).

xuzijian629:
Thank you for your review!

I'm still not sure what is the best practice, but my current idea is as follows.
* Call `format.sh` in CIs.
* Keep `python setup.py typecheck` command
* Do not introduce `python setup.py lint` or `python setup.py format`.

The reasons are
* `python setup.py typecheck` command is explained several times in docs and some users may already get used to it.
* It's difficult to prepare two scripts, `lint` and `format`, because flake8 is a lint tool (i.e., do not format), and clang-format is a format tool. So I put all checks into single file.
* `python setup.py lint` would be something that calls `format.sh` within subprocess. I think it's redundant and just calling `format.sh` would be nicer.

setup.py
xuzijian629: Thanks for the comments!
For `python setup.py lint` command, I summarized my opinion in https://github.com/onnx/onnx/pull/3988#issuecomment-1031139310

xuzijian629:
Since clang-format is still not applied, CI checks are failing.

Applying clang-format will cause a large amount of changes, so I'll do it later when other code is ready.

jcwchen (commented):
Thank you for the summary -- Yes, then I think only adding `format.sh` should be good. The main purpose here will be making it easy-to-use for users and also for CIs.

> Since clang-format is still not applied, CI checks are failing.
> Applying clang-format will cause a large amount of changes, so I'll do it later when other code is ready.

IIUC, clang-format is never applied in any ONNX CI (although we have such a file in the repo) so it makes sense that it needs a lot of changes to apply clang-format... To go forward, could we first only include mypy typecheck and flake8 into `tools/format.sh` and let's create an issue for further applying clang-format into `tools/format.sh` in the future?

xuzijian629:
Hi @jcwchen !
Thank you for your response!
I removed clang-format calls from CI. Now, I think CIs will pass.

tools/format.sh
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+
+set -eu
+
+
+cd "$(git rev-parse --show-toplevel)"
+
+echo -e "===> check flake8"
+flake8 onnx
```
jcwchen, line 12: Shall we use `flake8` instead of `flake8 onnx` to cover more files under ONNX repo? For instance, the files [here](https://github.com/onnx/onnx/tree/main/workflow_scripts).

jcwchen (commented):
Thank you for the quick update! Overall looks good to me. Only one comment about flake8. It would be great if we can have a similar .bat for Windows as well.

tools/format.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+
+
+set -eu
```
garymm, line 6: use long option names for readability:
```suggestion
set -o errexit
set -o nounset
```

garymm (changes requested):
Thanks for your contribution. A few minor comments.

tools/format.sh
xuzijian629: Since just using `flake8` also affects files in CI environment, I changed this to `flake8 onnx tools workflow_scripts` (also, it will also prevent checking files in `.venv/` in your local environment).

jcwchen (approved):

gramalingam (approved):

garymm (approved):
🍎

PR #3990, opened 2022-02-04
[BUG] Use correct pytest types in backend

kevinch-nv:
Signed-off-by: Kevin Chen <kevinch@nvidia.com>

**Description**
- Fix incorrect type annotations in backend test reporting

**Motivation and Context**
-  https://github.com/onnx/onnx/pull/3962 introduced type annotations, but types `pytest.nodes.Item` and `pytest.terminal.TerminalReporter` belong to `_pytest` instead of `pytest`

### DIFF
--- a/onnx/backend/test/report/__init__.py
+++ b/onnx/backend/test/report/__init__.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import pytest  # type: ignore
-import _pytest
+import _pytest  # type: ignore
 
 from .coverage import Coverage
 from typing import Dict, Text, Sequence, Any, List
--- a/onnx/backend/test/report/__init__.py
+++ b/onnx/backend/test/report/__init__.py
@@ -6,6 +6,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import pytest  # type: ignore
+import _pytest
 
 from .coverage import Coverage
 from typing import Dict, Text, Sequence, Any, List
@@ -23,7 +24,7 @@ def _add_mark(mark: Any, bucket: Text) -> None:
         _coverage.add_proto(proto, bucket, mark.args[1] == 'RealModel')
 
 
-def pytest_runtest_call(item: pytest.nodes.Item) -> None:
+def pytest_runtest_call(item: _pytest.nodes.Item) -> None:
     mark = item.get_closest_marker('onnx_coverage')
     if mark:
         assert item.nodeid not in _marks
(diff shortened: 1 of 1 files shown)


kevinch-nv:
@onnx/sig-archinfra are we testing the reporting functionality in the backend tests in CI? The previous MR that added type annotations introduced a bug that didn't seem to be caught by CI.

A minimal repro:
```
import pytest

def test(inp: pytest.nodes.Item):
    print(inp)

test(0)
```
Running this script gives: 
```
python3 test.py
  File "test.py", line 3, in <module>
    def test(inp: pytest.nodes.Item):
  File "/usr/local/lib/python3.8/dist-packages/pytest/__init__.py", line 171, in __getattr__
    raise AttributeError(f"module {__name__} has no attribute {name}")
AttributeError: module pytest has no attribute nodes
```
In this MR I've updated the type annotation to use the correct one from `_pytest`, but the CI build is failing. Given that `_pytest` holds the internal classes of `pytest`, should these functions even be type annotated?

onnx/backend/test/report/__init__.py
```diff
@@ -6,6 +6,7 @@
 from __future__ import unicode_literals
 
 import pytest  # type: ignore
+import _pytest
```
jcwchen, line 9:
```suggestion
import _pytest  # type: ignore
```

jcwchen: Ignore it for now. In this way it should pass mypy check in CIs

jcwchen (commented):
Thank you for catching this bug! It wasn't caught by CIs because there is no mypy stub (type annotation) for pytest and that's why import pytest was ignored. I have gone through the whole repo looking for pytest and these two should be the only ones which were used incorrectly.

To prevent future bug like it, I am trying to investigate whether there is a pre-built stub (like types-protobuf) for pytest and then we can enable type checking for pytest-related code as well. If no, we have to be very careful about all third-party libraries which are ignored.

onnx/backend/test/report/__init__.py
```diff
@@ -6,6 +6,7 @@
 from __future__ import unicode_literals
 
 import pytest  # type: ignore
+import _pytest  # type: ignore
```
jcwchen, line 9: FYI: I made a commit to ignore _pytest because mypy doesn't have a stub for it. Sorry about directly modifying it because I think your fix is good to be included in the upcoming 1.11 release (so I want to forward this PR ASAP). Thank you for the contribution!

jcwchen, line 9: FYI: I made a commit to ignore _pytest because mypy doesn't have a stub for it. Sorry about directly modifying it because I think your fix is good to be included in the upcoming 1.11 release (so I want to forward this PR ASAP). Thank you for the contribution!

jcwchen (approved):

PR #4006, opened 2022-02-11
Use urlopen from urllib.request to download models instead of wget

jcwchen:
**Description**
- Use urlopen (from Python standard library: urllib) to download a file instead of wget: add a new function _download_file
- Remove wget dependency
- Move master branch to main branch for onnx/models (The migration in onnx/models was just done today)

**Motivation and Context**
wget was brought by ONNX 1.11 new feature: `onnx.hub`. However, wget is not included in standard Python library and it has not been updated for quite a while (Last version was published 6 years ago). 

Another proposal for using requests:
```
def _download_file(url: str, filename: str) -> None:
    """
    Downloads the file with specifed filename from the url
    @param url: a url of download link
    @param filename: a specified file name for the downloaded file
    """
    with requests.get(url, stream=True) as response:
        response.raise_for_status()
        with open(filename, 'wb') as f:
            for chunk in response.iter_content(chunk_size=8192):
                f.write(chunk)
```
I finally chose to simply use urlopen from urllib because it's from Python standard library so ONNX doesn't need to install additional library.


liqunfu (approved):

onnx/test/hub_test.py
```diff
@@ -9,14 +9,10 @@
 import os
 
 
-@pytest.mark.skipif(
-    'TEST_HUB' not in os.environ or not os.environ['TEST_HUB'],
-    reason="Conserving Git LFS quota"
-)
```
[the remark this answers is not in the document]
jcwchen: It passes all CIs and I am adding it back.

onnx/hub.py
```diff
@@ -122,22 +121,38 @@ def _get_base_url(repo: str, lfs: bool = False) -> str:
         return "https://raw.githubusercontent.com/{}/{}/{}/".format(repo_owner, repo_name, repo_ref)
 
 
+def _download_file(url: str, file_name: str) -> None:
+    """
+    Downloads the file with specifed file_name from the url
+    @param url: a url of download link
```
xadupre, line 127: `:param`

xadupre (approved):

onnx/hub.py
jcwchen: Good point. `:param` might be more prevalent than `@param`. Have just updated it globally in the same file. Thanks!

mhamilton723 (approved):

mhamilton723:
Thank you @jcwchen!

PR #4017, opened 2022-02-15
Fix SSL issue by using http in manylinux2020 env

jcwchen:
**Description**
ca-bundles.crt was too old in CentOS 6. As a workaround, use http instead of https for now. After 1.11 release, ONNX will upgrade its manylinux package to manylinux2014 and we can remove this workaround then.

**Motivation and Context**
Current error:
```
+ yum install -y cmake3
Loaded plugins: ovl
Setting up Install Process
https://vault.centos.org/centos/6/os/x86_64/repodata/repomd.xml: [Errno 14] problem making ssl connection
Trying other mirror.
Error: Cannot retrieve repository metadata (repomd.xml) for repository: base. Please verify its path and try again
+ echo 'Installing yum package(s) failed.'
+ exit 1
Installing yum package(s) failed.
```
Similar issue: https://unix.stackexchange.com/questions/109585/yum-update-fails-error-cannot-retrieve-repository-metadata-repomd-xml-for-re


snnn (approved):

askhade (approved):

jcwchen:
Draft this because now release CIs run normally. Won't need this for now and probably after the release we will just upgrade to manylinux2014

Issue #4018, opened 2022-02-15
Some tf2onnx test cases failed with latest 1.11.0rc1 and 1.11.0rc2 version.


fatcat-z:
# Bug Report

### Describe the bug
While running tf2onnx test case BackendTests.test_sparse_fill_empty_rows with latest RC2 version, the test case failed with below exception:

======================================================================
ERROR: test_sparse_fill_empty_rows (__main__.BackendTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "repos/fatcat-z/tensorflow-onnx/tests/test_backend.py", line 4806, in test_sparse_fill_empty_rows
    self._run_test_case(func, [_OUTPUT, _OUTPUT1], {_INPUT: i_val, _INPUT1: v_val, _INPUT2: ds_val, _INPUT3: d_val})
  File "repos/fatcat-z/tensorflow-onnx/tests/test_backend.py", line 178, in _run_test_case
    return self.run_test_case(func, feed_dict, [], output_names_with_port, **kwargs)
  File "repos/fatcat-z/tensorflow-onnx/tests/backend_test_base.py", line 400, in run_test_case
    self.assert_shapes_correct(g, self.config.allow_missing_shapes, not self.config.skip_onnx_checker)
  File "repos/fatcat-z/tensorflow-onnx/tests/backend_test_base.py", line 297, in assert_shapes_correct
    raise e
  File "repos/fatcat-z/tensorflow-onnx/tests/backend_test_base.py", line 291, in assert_shapes_correct
    onnx.checker.check_model(model_proto, full_check=True)
  File "anaconda3/envs/tfonnxdebug/lib/python3.8/site-packages/onnx/checker.py", line 107, in check_model
    onnx.shape_inference.infer_shapes(model, check_type=True, strict_mode=True)
  File "anaconda3/envs/tfonnxdebug/lib/python3.8/site-packages/onnx/shape_inference.py", line 41, in infer_shapes
    inferred_model_str = C.infer_shapes(model_str, check_type, strict_mode, data_prop)
onnx.onnx_cpp2py_export.shape_inference.InferenceError: [ShapeInferenceError] Shape inference error(s): (op_type:Concat, node name: SparseFillEmptyRows/SparseFillEmptyRows_Concat__45): [ShapeInferenceError] All inputs to Concat must have same rank

This exception was thrown by onnx.checker, but only onnx version is changed in the environment so we should take a look.

### System information
[...]

PR #4019, opened 2022-02-15
Fix Expand shape inference: stop rank inference if the shape is symbolic

jcwchen:
**Description**
- Add a case to return early if rank inference for Expand fails
- Update error message for Concat to help debugging

**Motivation and Context**
Closes https://github.com/onnx/onnx/issues/4018. In ONNX 1.11, shape inference for Expand has been improved by supporting more case of rank inference: https://github.com/onnx/onnx/pull/3807. However, there is missing case that if the shape is a symbolic one, Expand should not do rank inference. Otherwise, it will be rank 0.

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -2063,11 +2063,12 @@ ONNX_OPERATOR_SET_SCHEMA(
               for (int64_t i = 0; i < dim_value; ++i) {
                 second_shape.add_dim();
               }
+            } else {
+                return;
             }
             bidirectionalBroadcastShapeInference(
                 input_shape, second_shape, *getOutputShape(ctx, 0));
           }
-          return;
         }));
 
 static const char* Sinh_ver9_doc = R"DOC(
--- a/onnx/defs/math/old.cc
+++ b/onnx/defs/math/old.cc
@@ -1020,11 +1020,12 @@ ONNX_OPERATOR_SET_SCHEMA(
               for (int64_t i = 0; i < dim_value; ++i) {
                 second_shape.add_dim();
               }
+            } else {
+                return;     
             }
             bidirectionalBroadcastShapeInference(
                 input_shape, second_shape, *getOutputShape(ctx, 0));
           }
-          return;
         }));
 
 static const char* Sign_ver9_doc = R"DOC(
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -590,7 +590,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           for (size_t i = 0; i < numInputs; i++) {
             const auto& shape = ctx.getInputType(i)->tensor_type().shape();
             if (shape.dim_size() != rank) {
-              fail_shape_inference("All inputs to Concat must have same rank");
+              fail_shape_inference("All inputs to Concat must have same rank. Input ", i , " has rank ", shape.dim_size(), " != ", rank);
             }
             for (int j = 0; j < rank; j++) {
               if (j == axis) {
--- a/onnx/defs/tensor/old.cc
+++ b/onnx/defs/tensor/old.cc
@@ -559,7 +559,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           for (size_t i = 0; i < numInputs; i++) {
             const auto& shape = ctx.getInputType(i)->tensor_type().shape();
             if (shape.dim_size() != rank) {
-              fail_shape_inference("All inputs to Concat must have same rank");
+              fail_shape_inference("All inputs to Concat must have same rank. Input ", i , " has rank ", shape.dim_size(), " != ", rank);
             }
             for (int j = 0; j < rank; j++) {
               if (j == axis) {


gramalingam (approved):

onnx/test/shape_inference_test.py
```diff
@@ -383,6 +383,18 @@ def test_expand_dynamic_shape(self) -> None:
...
+        graph = self._make_graph(
+            [('x', TensorProto.INT32, (1, 2, None)),
+             ('shape', TensorProto.INT64, ('unk__0',))],
+            [make_node("Expand", ['x', 'shape'], ['y'])],
+            [],
+            initializer=[])
+        # if giving a symbolic shape, Expand should not infer any shape or rank inference
+        self._assert_inferred(
+            graph,
+            [make_tensor_value_info('y', TensorProto.INT32, ())])
```
gramalingam, line 396: Is that `()` mean an empty shape (rank 0) or no shape?

gramalingam: Ideally it should be `None` or something like that

jcwchen: Good question. It's tricky here because it seems that both shape (rank 0) and no shape represent in the same way (empty shape)... Let me investigate more and update the test

gramalingam: I think we should supply `None` here when creating the value_info. But the comparison should be fixed at https://github.com/onnx/onnx/blob/d824ce38f935b83f724628a254cb855a95180778/onnx/test/shape_inference_test.py#L76 to check if one has no shape the other has no shape also.

jcwchen: Yes! I am exactly doing the same thing by adding:
```
assert vi_type.tensor_type.HasField('shape') == inferred_vi_type.tensor_type.HasField('shape')
```
I have tried that rc2 cannot pass this test, but with my fix in this PR, it can pass. Thank you!

onnx/test/shape_inference_test.py
```diff
@@ -72,6 +72,7 @@ def _compare_value_infos(self, vi_type: TypeProto, inferred_vi_type: TypeProto)
             assert inferred_vi_type.HasField('tensor_type')
             assert vi_type.tensor_type.HasField('elem_type')
             assert inferred_vi_type.tensor_type.HasField('elem_type')
+            assert vi_type.tensor_type.HasField('shape') == inferred_vi_type.tensor_type.HasField('shape')
```
gramalingam, line 75: the for-loop below should also be protected by `if (vi_type.tensor_type.HasField('shape'):`

onnx/test/shape_inference_test.py
```diff
@@ -383,6 +384,18 @@ def test_expand_dynamic_shape(self) -> None:
             graph,
             [make_tensor_value_info('y', TensorProto.INT32, (None, 2, None))])
 
+    def test_expand_symbolic_shape(self) -> None:
+        graph = self._make_graph(
+            [('x', TensorProto.INT32, (1, 2, None)),
+             ('shape', TensorProto.INT64, ('unk__0',))],
```
gramalingam, line 390: I believe that this also needs to be `('shape', TensorProto.INT64, None)`, right?

onnx/test/shape_inference_test.py
jcwchen: Although I think even if a tensor_type doesn't have a shape, tensor_type.shape.dim will just be empty. Still, I think adding protection is safer. Added. Thanks!

gramalingam: The problem is that `tensor_type.shape` has the side-effect of initializing `shape` to be an empty list (an unfortunate design choice of protobuf) if the object doesn't have a `shape` field. The checker method should not have such a side-effect.

onnx/test/shape_inference_test.py
```diff
@@ -3881,7 +3895,7 @@ def test_optional_sequence_has_element(self) -> None:
              make_node('OptionalHasElement', ['optional'], ['output'])],
             [])
         self._assert_inferred(graph, [sequence_val_info, optional_val_info,
-                                      make_tensor_value_info('output', TensorProto.BOOL, None)])  # type: ignore
+                                      make_tensor_value_info('output', TensorProto.BOOL, ())])  # type: ignore
```
jcwchen, line 3898: Originally these 3 tests (which involve Optional-related operators) have a tensor whose shape is empty (has the shape field without any content). We should use `()` to represent this case.

A follow-up question: are rank 0 (having a shape field without any content) and empty shape (not having a shape field) always indifferent? From my understanding, rank 0 means it doesn't have (or need) a shape; empty shape means unknown shape. @gramalingam am I correct? Thank you.

gramalingam: Let me look at the tests. But "rank 0" means same as "shape is an empty list" same as "empty shape". But "unknown shape" is different and is represented by "no shape field" and it means that the "rank is unknown".

gramalingam: Using shape `()` or `[]` for "rank 0" makes sense to me. I need to look at the examples to see what is expected in the example.

onnx/test/shape_inference_test.py
```diff
@@ -2111,7 +2125,7 @@ def test_if_with_different_optional_shapes_in_then_else_branches(self) -> None:
             []
         )
 
-        output_tensor_proto = helper.make_tensor_type_proto(elem_type=TensorProto.FLOAT, shape=None)
+        output_tensor_proto = helper.make_tensor_type_proto(elem_type=TensorProto.FLOAT, shape=())
```
gramalingam, line 2128: I don't think `shape=()` in this example. The then branch produces a `shape=[1]` and else branch produces a `shape=[5]`.

gramalingam: I think we should use `shape=(None,)` or `shape=[None]` here

onnx/test/shape_inference_test.py
```diff
@@ -3862,7 +3876,7 @@ def test_optional_tensor_has_element(self) -> None:
              make_node('OptionalHasElement', ['sequence'], ['output'])],
             [])
         self._assert_inferred(graph, [optional_val_info,
-                                      make_tensor_value_info('output', TensorProto.BOOL, None)])  # type: ignore
+                                      make_tensor_value_info('output', TensorProto.BOOL, ())])  # type: ignore
```
gramalingam, line 3879: Yes, here this looks fine, since `output` must be a scalar boolean with empty shape

onnx/test/shape_inference_test.py
jcwchen: Updated since (None, ) is more precise as you suggested. Thank you @gramalingam !

PR #4039, opened 2022-02-24
Add OperatorSetId definition in docs

### LINKED ISSUE #1100: Missing/wrong documentation for opset_import/OperatorSetId
The OperatorSetId is not defined in https://github.com/onnx/onnx/blob/master/docs/IR.md

> opset_import | OperatorSetId | A collection of operator set identifiers made available to the  model. An implementation must support all operators in the set or reject  the model.

`operator set identifiers` may be assumed to be a simple string. In fact it is a `{domain, version}` struct.

sechkova:
**Description**
- Add OperatorSetId definition in IR docs.

**Motivation and Context**
- Fixes #1100

### DIFF
--- a/docs/IR.md
+++ b/docs/IR.md
@@ -114,6 +114,14 @@ Name|Type|Format|Description
 model_author|string|A comma-separated list of names.|The personal name of the author(s) of the model, and/or their organizations.
 model_license|string|Name or URL.|The well-known name or URL of the license under which the model is made available.
 
+### Operator Set Identifiers
+
+Each operator set is uniquely identified by a (domain, version) pair.
+
+Name|Type|Description
+|---|---|---|
+domain|string|The domain of the operator set being identified.
+version|int64|The version of the operator set being identified. Same as 'opset_version' in the operator set.
 
 ### Operator Sets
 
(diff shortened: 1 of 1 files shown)


docs/IR.md
```diff
@@ -114,6 +114,14 @@ Name|Type|Format|Description
...
 model_license|string|Name or URL.|The well-known name or URL of the license under which the model is made available.
 
+### Operator Set Identifiers
+
+Each operator set is uniquely identified by a (domain, version) pair.
+
+Name|Type|Description
+|---|---|---|
+domain|string|The domain of the operator set being identified.
+version|int64|The version of the operator set being identified. Same as 'opset_version' in the operator set.
```
gramalingam, line 124: May be worth adding the following lines (from the subsequent section Operator Set):
```
The operator set version is a simple integer value that is monotonically increased as new versions of the
operator set are published.

Operator sets other than the default operator set MUST specify a domain and SHOULD use reverse domain names
based on the responsible organization's identity, the same convention that is used for
[naming Java packages](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html).
```

sechkova: Thanks for the review, it's added with [9600196](https://github.com/onnx/onnx/pull/4039/commits/96001968cf0e9c5569cc308efe9215b0c707ab4d)

gramalingam (approved):

jcwchen (approved):

PR #4044, opened 2022-02-28
Change the wrong words of second layer `input`

How-Wang:
**Description**
- Change wrong words in file

**Motivation and Context**
- We should sent the `output` of first model to the `input` of second layer instead of `output` of second layer.


jcwchen (commented):
Thanks for catching this. Please fix DCO (check [here](https://github.com/onnx/onnx/pull/4044/checks?check_run_id=5362419496))

jcwchen (approved):

gramalingam (approved):

Issue #4053, opened 2022-03-04
Failed to install latest version of onnx-weekly nightly job.


fatcat-z:
# Bug Report

### Describe the bug
In some case, when we run the command "pip install --index-url https://test.pypi.org/simple/ onnx-weekly" to install latest version of onnx-weekly nightly build, it failed to install those latest ones but only an old version 1.10.1.dev20210830 will be installed. Some logs are shown as below:

```
:~/Desktop/code/tensorflow-onnx$ pip install --index-url https://test.pypi.org/simple/ onnx-weekly
Looking in indexes: https://test.pypi.org/simple/
WARNING: Keyring is skipped due to an exception: Failed to create the collection: Prompt dismissed..
Collecting onnx-weekly
Downloading https://test-files.pythonhosted.org/packages/19/56/84b1eb141b6129ffeac8c0c3a76e4170771d003479857ef1d855fa3b09b9/onnx_weekly-1.11.0.dev20220228-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (12.8 MB)
|████████████████████████████████| 12.8 MB 1.0 MB/s
Downloading https://test-files.pythonhosted.org/packages/3d/13/392cc9c308014001ce5d74e914bdb8d7a917b23f2edb29ac56d4b28c0ed6/onnx_weekly-1.11.0.dev20220221-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (12.8 MB)
|████████████████████████████████| 12.8 MB 1.1 MB/s
Downloading https://test-files.pythonhosted.org/packages/30/af/2a359cc38da90572c000ff10ac4e939d3423e7d1215300afcfe07044acc5/onnx-weekly-1.11.0.dev20220214.tar.gz (9.9 MB)
|████████████████████████████████| 9.9 MB 87 kB/s
ERROR: Command errored out with exit status 1:
[...]

PR #4001, opened 2022-02-09
Remove Python 3.6, upgrade NumPy to 1.21.5, upgrade to manylinux2014

jcwchen:
**Description**
- Remove Python 3.6 in all CIs
- Upgrade NumPy to 1.21.5 to prevent security issue. Also 1.21.5 supports Python 3.10.
- Upgrade to use `Visual Studio 17 2022` because now windows-latest is Windows 2022 server
- Upgrade to manylinux2014 since manylinux2010 will be deprecated soon

TODO: Add Python 3.10 support after NumPy upgrade. It will be anohter PR.

**Motivation and Context**
Closes #3975. [Python 3.6 has reached end-of-life](https://www.python.org/dev/peps/pep-0494/). GitHub Action just dropped Python 3.6 support as well.

PR #4059, opened 2022-03-07
Loose NumPy requirement to grant more flexibility

jcwchen:
**Description**
Loose NumPy requirement to grant more flexibility.

**Motivation and Context**
Closes https://github.com/onnx/onnx/issues/4053. Some old TensorFlow versions are not compatible with NumPy>=1.20. Recently ONNX bumped its NumPy version for building wheels: https://github.com/onnx/onnx/pull/4001 because of Python 3.10 support and resolving vulnerability issue from old NumPy. However, ONNX itself does not have strict dependency on NumPy. IMO, it should be OK to make minimum version requirement for NumPy lower to grant more flexibility for ONNX downstream repos.

cc @fatcat-z

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-numpy >= 1.16.6
+numpy >= 1.16.6 # TODO: update once TensorFlow 2.6 is end of life
 protobuf >= 3.12.2
 typing-extensions >= 3.6.2.1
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,3 +1,3 @@
-numpy >= 1.21.5
+numpy >= 1.16.6
```
garymm, line 1: Add a comment to say
# TODO: update once TensorFlow 2.6 is end of life

gramalingam (approved):

PR #4068, opened 2022-03-16
(WIP) local function inline

liqunfu:

**Description**
- Describe your changes.

**Motivation and Context**


justinchuby:
@titaiwangms

titaiwangms:
@

> @titaiwangms

@justinchuby I think this is exactly what we need for new symbolic_fn. We should proceed on this if ONNX team is busy with other stuff? What do you think?

justinchuby:
I don't know if the same features are being worked on in other prs? @liqunfu @gramalingam could you share some context on where we are in inlining features?

gramalingam:
This does look outdated. Do we really need this, given that we have one implementation of inliner in place already?

Issue #3986, opened 2022-02-02
A couple of issues with shape inference for functions


gramalingam:
# Bug Report

### Describe the bug

Shape inference for functions does not handle "Constants". Specifically, the function (see https://github.com/onnx/onnx/blob/f0c4ee2b46b3d00c005d1ecd9440cfc151f4fd1d/onnx/shape_inference/implementation.cc#L568 ) does not check for `Constant` ops the way shape inference for Graph does it (see https://github.com/onnx/onnx/blob/f0c4ee2b46b3d00c005d1ecd9440cfc151f4fd1d/onnx/shape_inference/implementation.cc#L339 ).

However, rather than duplicating the logic, I think it would be better to unify the implementations of `InferShapesImpl` and `InferShapeForFunctionNode`, since they have a lot of duplication/commonality. Pulling them into a common base-class for inference would help simplify the code and fix the bug as well.

PR #4069, opened 2022-03-17
Refactor shape inference

gramalingam:
**Description**
Refactor graph-level shape inference function into a class. This is the first step in eliminating unnecessary code duplication between graph-level and function-level shape-inference.

**Motivation and Context**

See https://github.com/onnx/onnx/issues/3986


onnx/shape_inference/implementation.cc
```diff
@@ -425,67 +382,172 @@ static void InferShapesImpl(
...
+  }
+
+ public:
+  ShapeInferenceImplBase(
+      GraphProto* gp,
+      const std::unordered_map<std::string, TypeProto*>& outer_scope_value_types_by_name_in,
+      const std::unordered_map<std::string, int>& opset_imports_in,
+      const ShapeInferenceOptions& options_in,
+      SymbolTable* symbol_table_in,
+      const ModelLocalFunctionsMap& model_local_functions_map_in,
```
jcwchen, line 486: It would be great if we can add generated_shape_data_by_name_in into ShapeInferenceImplBase like [here](https://github.com/onnx/onnx/blob/67f21f2095084f96ca1698ed621a125fe8b53fb0/onnx/shape_inference/implementation.cc#L572) for FunctionProto.

jcwchen: After discussion, we think it's fine to do it in another PR.

onnx/shape_inference/implementation.cc
```diff
@@ -236,107 +235,62 @@ std::string GetModelLocalFunctionsMapIdentifier(const std::string& domain, const
...
-    const ISchemaRegistry* schema_registry = OpSchemaRegistry::Instance(),
-    const int ir_version = IR_VERSION // default the latest one
-) {
-  std::unordered_map<std::string, TypeProto*> value_types_by_name{outer_scope_value_types_by_name};
-  std::unordered_map<std::string, TypeProto*> undefined_value_types_by_name{outer_scope_value_types_by_name};
-  std::unordered_map<std::string, TensorShapeProto> generated_shape_data_by_name;
+// For graph-output types that are undefined, update them post-inference.
+// Q: Why exception is not saved as error if experimental op?
+// Document (in IR.md): domain for model-local functions
+// What is sparsetensortype.shape supposed to be?
```
jcwchen, line 241: nit: do we want to keep this for future reference or it can be removed in this PR?

jcwchen (approved):
LGTM. I have just verified this graph-level shape inference change with ONNX model Zoo, PyTorch-ONNX exporter and tf2onnx. I didn't see any failure caused by this fundamental change.

onnx/shape_inference/implementation.cc
gramalingam: Removed the comments, thanks!

askhade (approved):

PR #4076, opened 2022-03-22
Define layer normalization as a function

### LINKED ISSUE #2379: Layer norm operator
LayerNorm is an very important operator in BERT (one of the computation bottleneck). Maybe we should add it as a FunctionProto to have a more meaningful BERT representation and allow runtime to easily write an optimized kernel for it.

wschin:
Layer normalization has been a popular operation in language models, so let's support it in ONNX. `LayerNormalization` is defined as a function, so this PR doesn't increase the number of `primitive` operators. A reference implementation in numpy is provided and verified with Pytorch and Nvidia APEX.

Test script for reference implementation:
```python
import numpy as np
import torch
from torch import nn


def _layer_normalization(X, W, B, axis, epsilon=1e-5):
    X_shape = X.shape
    unsqueezed_rank = len(X_shape) - axis
    # The shape of mean and variance tensors.
    reduction_shape = X_shape[0:axis] + (1,) * unsqueezed_rank

    # Parameter used to convert N-D tensor layer
    # normalization to equivalent 2-D matirx operations.
    X_rank = len(X_shape)
    row_number = 1
    col_number = 1
    for i in range(X_rank):
        if i < axis:
            row_number *= X_shape[i]
        else:
            col_number *= X_shape[i]

    # After reshaping input tensor X into a matrix,
    # layer normalization is equivalent to conducting
    # standardization on each column vector (s.t. each
    # column has zero mean and unit variance).
    x_mat = np.reshape(X, (row_number, col_number))
    # This computes mean for every x_mat's column.
    x_mean = np.sum(x_mat, axis=1, keepdims=True) / col_number
    x_diff = x_mat - x_mean
    x_squared_diff = x_diff * x_diff
    # This computes variance for every x_mat's column.
    variance = np.sum(x_squared_diff, axis=1, keepdims=True) / col_number
    variance_eps = variance + epsilon
    std_dev = np.sqrt(variance_eps)
    inv_std_dev = np.reciprocal(std_dev)
    # Standardlization step. y_mat is zero-mean and unit-variance.
    y_mat = x_diff * inv_std_dev
    # Apply affine transform on normalization outcome.
    # W is linear coefficient while B is bias.
    Y = np.reshape(y_mat, X_shape) * W + B
    # Matrix-level operations' outputs should be reshaped
    # to compensate the initial tensor-to-matrix reshape.
    X_mean = np.reshape(x_mean, reduction_shape)
    X_inv_std_dev = np.reshape(inv_std_dev, reduction_shape)

    return Y, X_mean, X_inv_std_dev


def _apex_layer_normalization(X, W, B, axis, epsilon):
    import apex
    import fused_layer_norm_cuda

    X_shape = X.shape
    # The size of sub-tensor each mean/variance is
    # computed on.
    normalized_shape = X_shape[axis:]
    Y, X_mean, X_inv_std_dev = fused_layer_norm_cuda.forward_affine(
        X, normalized_shape, W, B, epsilon)

    unsqueezed_rank = len(X_shape) - axis
    reduction_shape = X_shape[0:axis] + (1,) * unsqueezed_rank

    return Y, X_mean.reshape(reduction_shape), X_inv_std_dev.reshape(reduction_shape)


def run(shape, axis, epsilon, atol):
    # Compare ONNX's reference implementation with
    # Pytorch's and APEX's layer normalization.

    X = torch.randn(shape)
    normalized_shape = X.shape[axis:]
    # Shape of mean and variance tensors.
    reduced_shape = X.shape[0:axis]
    layer_norm = nn.LayerNorm(normalized_shape)
    Y = layer_norm(X)

    # ONNX reference implementation.
    Y1, X_mean1, X_inv_std_dev1 = _layer_normalization(
        X.detach().numpy(),
        layer_norm.weight.detach().numpy(),
        layer_norm.bias.detach().numpy(),
        axis, epsilon)
    assert np.allclose(Y.detach().numpy(), Y1, atol=atol)
    # NVIDIA reference implementation.
    Y2, X_mean2, X_inv_std_dev2 = _apex_layer_normalization(
        X.detach().to('cuda'),
        layer_norm.weight.detach().to('cuda'),
        layer_norm.bias.detach().to('cuda'),
        axis, epsilon)
    assert np.allclose(Y.detach().numpy(),
                       Y2.detach().to('cpu').numpy(), atol=atol)
    print(X_mean1.shape, X_mean2.shape)
    assert np.allclose(X_mean1, X_mean2.detach().to(
        'cpu').detach().numpy(), atol=atol)
    print(X_inv_std_dev1.shape, X_inv_std_dev2.shape)
    assert np.allclose(X_inv_std_dev1, X_inv_std_dev2.detach().to(
        'cpu').numpy(), atol=atol)


run((2, 7, 5, 27), axis=2, epsilon=1e-5, atol=1e-6)
run((7, 5, 27), axis=0, epsilon=1e-5, atol=1e-6)
run((7, 5, 27), axis=1, epsilon=1e-5, atol=1e-6)
run((7, 5, 27), axis=2, epsilon=1e-5, atol=1e-6)
run((7, 5), axis=0, epsilon=1e-5, atol=1e-7)
run((7, 5), axis=1, epsilon=1e-5, atol=1e-7)
run((12,), axis=0, epsilon=1e-5, atol=1e-7)
```

Fixes #2379 also verified with `https://github.com/microsoft/onnxruntime/pull/11191`


docs/Changelog.md
```diff
@@ -20030,6 +20030,93 @@ This version of the operator has been available since version 16 of the default
...
+        Normalized = Mul(D, InvStdDev)
+        ```
+        where `normalized_axes` is `[axis, ..., rank of X - 1]`.
+        The variables `Var` and `StdDev` stand for variance and
+        standard deviation, respectively. The second output is
+        `Mean` and the last one is `InvStdDev`.
+        Depending on `stash_type` attribute, the actual computation may
+        happen in different floating-point precision.
+        For example, if `stash_type` is 1, this operator may cast
+        all input variables to 32-bit float, does the computation, and
```
chausner, line 20058:
```suggestion
        all input variables to 32-bit float, perform the computation, and
```

onnx/defs/nn/defs.cc
```diff
@@ -2487,4 +2487,201 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                  .Const("FloatEpsilon", ToTensor<float>(epsilon))
+                  .Add("Epsilon = Cast (FloatEpsilon)", "to", U)
+                  .Add("XShape = Shape (X)")                                                    // shape of input tensor: 1D tensor
+                  .Add("Rank = Size (XShape)")                                                  // rank of input tensor: scalar
+                  .Add("Zero1D = Constant()", "value", mktensor(0))                             // [0] : 1D tensor
+                  .Add("Axis1D = Constant()", "value", mktensor(axis))                          // [axis] : 1D tensor
+                  .Add("PrefixShape = Slice (XShape, Zero1D, Axis1D)")                          // [d[0], ..., d[axis-1]]
+                  .Add(axis > 0                                                                 // number of axes that are reduced =
+                          ? "NumReducedAxes = Sub (Rank, Axis1D)"                              // [rank - axis]: 1D tensor
+                          : "NumReducedAxes = Neg (Axis1D)")                                   // [-axis] : 1D tensor
```
chausner, line 2653:
```suggestion
                          ? "NumReducedAxes = Sub (Rank, Axis1D)"                               // [rank - axis]: 1D tensor
                          : "NumReducedAxes = Neg (Axis1D)")                                    // [-axis] : 1D tensor
```

onnx/defs/nn/defs.cc
```diff
@@ -2487,4 +2487,201 @@ ONNX_OPERATOR_SET_SCHEMA(
         )ONNX"
         ));
 
+static const char* LayerNormalization_ver16_doc = R"DOC(
+      This is layer normalization defined in ONNX as function.
+      The overall computation can be split into two stages.
+      The first stage is standardlization, which makes the
+      normalized elements having zero mean and unit variances.
+      The computation required by standardlization can be
```
chausner, line 2495:
```suggestion
      The first stage is standardization, which makes the
      normalized elements have zero mean and unit variances.
      The computation required by standardization can be
```

onnx/backend/test/data/node/test_layer_normalization_2d_axis0/test_data_set_0/input_0.pb
```diff
@@ -0,0 +1 @@
+BXJ0x��?h��>��z?�j@$�?�.z��8s?b��hdӽ�9�>(�>�%�?
```
chausner, line 1: git attributes should be set to declare all .pb files as binary, otherwise it can lead to data corruption when git interprets them as text files and messes up line ending characters.

onnx/backend/test/case/node/layernormalization.py
```diff
@@ -0,0 +1,122 @@
...
+    # This computes mean for every x_mat's column.
+    x_mean = np.sum(x_mat, axis=1, keepdims=True) / col_number
+    x_diff = x_mat - x_mean
+    x_squared_diff = x_diff * x_diff
+    # This computes variance for every x_mat's column.
+    variance = np.sum(x_squared_diff, axis=1, keepdims=True) / col_number
+    variance_eps = variance + epsilon
+    std_dev = np.sqrt(variance_eps)
+    inv_std_dev = np.reciprocal(std_dev)
+    # Standardlization step. y_mat is zero-mean and unit-variance.
```
chausner, line 41:
```suggestion
    # Standardization step. y_mat is zero-mean and unit-variance.
```

docs/Operators.md
```diff
@@ -9320,6 +9321,183 @@ expect(node, inputs=[input, W, R, B, seq_lens, init_h, init_c, P], outputs=[Y_h.
 </details>
 
 
+### <a name="LayerNormalization"></a><a name="layernormalization">**LayerNormalization**</a>
+
+  This is layer normalization defined in ONNX as function.
+        The overall computation can be split into two stages.
+        The first stage is standardlization, which makes the
+        normalized elements having zero mean and unit variances.
+        The computation required by standardlization can be
```
chausner, line 9330:
```suggestion
        The first stage is standardization, which makes the
        normalized elements have zero mean and unit variances.
        The computation required by standardization can be
```

docs/Changelog.md
```diff
@@ -20030,6 +20030,93 @@ This version of the operator has been available since version 16 of the default
 <dd>Only bool</dd>
 </dl>
 
+### <a name="LayerNormalization-16"></a>**LayerNormalization-16**</a>
+
+  This is layer normalization defined in ONNX as function.
+        The overall computation can be split into two stages.
+        The first stage is standardlization, which makes the
+        normalized elements having zero mean and unit variances.
+        The computation required by standardlization can be
```
chausner, line 20039:
```suggestion
        The first stage is standardization, which makes the
        normalized elements have zero mean and unit variances.
        The computation required by standardization can be
```

onnx/backend/test/data/node/test_layer_normalization_2d_axis0/test_data_set_0/input_0.pb
wschin: Is [it](https://github.com/onnx/onnx/blob/2ab133404afce34552aaccd86e7023e1fb9a60d2/.gitattributes#L1) what you're looking for?

wschin: I changed `.gitattributes` at this [commit](https://github.com/onnx/onnx/pull/4076/commits/0fc927cc315b1386a2e15054998de5bfa1b606c1). Then, I [removed ](https://github.com/onnx/onnx/pull/4076/commits/73eeee8e90b05078eb2c91183dfdf1615e639aa4)and [added ](https://github.com/onnx/onnx/pull/4076/commits/ebfbde75451c708f13dea14270ca1f5597857a2a)back my `.pb` files. Does it look ok now?

wschin: I have tried many combinations of `binary`, `-text`, and so on in `.gitattributes` file. Github just keeps showing the diff for `*.pb` files. Since this is a more difficult problem, I'd suggest to address it in another PR.

onnx/defs/nn/defs.cc
```diff
@@ -2487,4 +2487,201 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+      The overall computation can be split into two stages.
+      The first stage is standardization, which makes the
+      normalized elements have zero mean and unit variances.
+      The computation required by standardization can be
+      described by the following equations.
+      ```
+      Mean = ReduceMean<axes=normalized_axes>(X)
+      D = Sub(X, Mean)
+      DD = Mul(Diff, Diff)
+      Var = ReduceMean<axes=axes>(DD)
```
gramalingam, line 2501: Is the second `axes` supposed to be `normalized_axes`?

onnx/defs/nn/defs.cc
```diff
@@ -2487,4 +2487,201 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+      StdDev = Sqrt(VarEps)
+      InvStdDev = Reciprocal(StdDev)
+      Normalized = Mul(D, InvStdDev)
+      ```
+      where `normalized_axes` is `[axis, ..., rank of X - 1]`.
+      The variables `Var` and `StdDev` stand for variance and
+      standard deviation, respectively. The second output is
+      `Mean` and the last one is `InvStdDev`.
+      Depending on `stash_type` attribute, the actual computation may
+      happen in different floating-point precision.
```
gramalingam, line 2512: Regarding the use of "may": would it be possible to be more precise about what is guaranteed/required (in terms of precision) and what is allowed to be dependent on the implementation?

gramalingam: Unless there is a good reason, it is better to be specific.

onnx/defs/nn/defs.cc
```diff
@@ -2487,4 +2487,201 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              "The epsilon value to use to avoid division by zero.",
+              AttributeProto::FLOAT,
+              1e-5f)
+        .Attr("stash_type",
+              "Type of Mean and InvStdDev. This also specifies stage one's computation precision.",
+              AttributeProto::INT,
+              static_cast<int64_t>(ONNX_NAMESPACE::TensorProto_DataType_FLOAT))
+        .AllowUncheckedAttributes()
+        .Input(0, "X", "Tensor to be normalized.", "T")
+        .Input(1, "Scale", "Scale tensor.", "T")
```
gramalingam, line 2551: The definition of BatchNorm went through multiple revisions because there was a need to allow different precisions for different inputs. Is it fine if X/Scale/B are all constrained to be same type?

onnx/defs/nn/defs.cc
```diff
@@ -2487,4 +2487,201 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T",
+            {"tensor(float16)", "tensor(float)", "tensor(double)", "tensor(bfloat16)"},
+            "Constrain input types and output Y type to float tensors.")
+        .TypeConstraint(
+            "U",
+            {"tensor(float)", "tensor(bfloat16)"},
+            "Type of Mean and InvStdDev tensors.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateShapeAndTypeFromFirstInput(ctx);
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
```
gramalingam, line 2566: Seems redundant given previous line.

onnx/defs/nn/defs.cc
```diff
@@ -2487,4 +2487,201 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            axis = axis_proto->i();
+          }
+          if (axis < 0) {
+            // Convert negative axis value to equivalent
+            // positive value.
+            axis += input_ndim;
+          }
+
+          if (ctx.getNumOutputs() > 1) {
+            auto saved_mean_shape = ctx.getOutputType(1)->mutable_tensor_type()->mutable_shape();
```
gramalingam, line 2597: nit: probably simpler to omit the prefix `saved_` in the variable names here: just `mean_shape` and `inv_std_dev_shape`

onnx/backend/test/case/node/layernormalization.py
```diff
@@ -0,0 +1,122 @@
...
+                inputs=['X', 'W', 'B'],
+                outputs=['Y', 'Mean', 'InvStdDev'],
+                axis=axis,
+            )
+
+            expect(node, inputs=[X, W, B], outputs=[Y, mean, inv_std_dev],
+                   name=f'test_layer_normalization_4d_axis{axis}')
+
+        for i in range(len(X.shape)):
+            case(i)
```
gramalingam, line 76: Some testcases with negative axis and unspecified axis may be good.

onnx/defs/nn/defs.cc
wschin: Makes sense. I change it to `must`.

onnx/defs/nn/defs.cc
wschin: Layer normalization is very similar to batch normalization (major difference is the normalized axes). As batch normalization have shown the need of using difference precision for `mean` and `variance` tensors, I prefer to keep this as is for layer normalization's `mean` and `variance`. In addition, mix-precision training for language model is popular enough to support extra types. For inference, I guess mix-precision is also helpful to compress the trained models. Does it sound ok?

gramalingam: Discussed with Wei-Sheng. Summary: the inputs will be of same type for now, as there is no use-case yet of using mixed-precision for inputs.

gramalingam (approved):

postrational (approved):

Issue #4078, opened 2022-03-22
Version converter crashes if a value_info's name does not exist in the graph


jcwchen:
# Bug Report

### Describe the bug

An ONNX model crashes after version conversion without any error. The model is attached. The root cause is [here](https://github.com/onnx/onnx/blob/94e2f64551ded652df53a7e9111031e8aabddaee/onnx/common/ir_pb_converter.cc#L357). It assumes value_info's name has already existed in the graph. Although this model might be an invalid model (the value_info seems unused), version_converter should not crash without any error or warning.

cc @skottmckay 

### Reproduction instructions
- Describe the code to reproduce the behavior.
```
import onnx
from onnx import version_converter
m = onnx.load('pt_mobilenet_v2.ssi.basic.onnx')
version_converter.convert_version(m, 15)
...
```
- Attach the ONNX model to the issue (where applicable)

[pt_mobilenet_v2.ssi.basic.onnx.zip](https://github.com/onnx/onnx/files/8326917/pt_mobilenet_v2.ssi.basic.onnx.zip)

### Expected behavior

The model should be successfully converted without crash.

### Notes

Will submit a PR for it soon.

PR #4079, opened 2022-03-22
Fix crash: Skip unused value_info for version_converter

jcwchen:
**Description**
- Add a common CreateDummyValue function for version_converter
- ~Create dummy value for value_info whose name does not exist in the graph. Or, perhaps it should throw an error or warning since it might be an invalid model~. If a value_info's name does not exist in the graph (unused), simply skip it during version conversion.

**Motivation and Context**
Closes https://github.com/onnx/onnx/issues/4078.


liqunfu (approved):

gramalingam:
I am unclear about the exact invariants of the IR and the design. Would it be possible to add a line or so near https://github.com/onnx/onnx/blob/94e2f64551ded652df53a7e9111031e8aabddaee/onnx/common/ir_pb_converter.cc#L223 to at least document the creation of dummy nodes for valueinfos without a definition?

onnx/common/ir_pb_converter.cc
```diff
@@ -204,6 +204,13 @@ std::vector<Dimension> tensorShapeProtoToDimensions(const ONNX_NAMESPACE::Tensor
   return dims;
 }
 
+void createDummyValue(std::unique_ptr<Graph>& g, const std::string& name, std::unordered_map<std::string, Value*>& value_by_name_of) {
+  auto* undef = g->create(kCaptured, 1);
```
gramalingam, line 208: What exactly does `kCaptured` mean? Maybe it is relevant to the other two original calls, but not sure it means the same for undefined valueinfos.

jcwchen:
After discussion, we think simply skipping such a unused value_info should be fine for now. Then we don't need to create a dummy node for those unused value_infos.

onnx/common/ir_pb_converter.cc
jcwchen: Good question. It seems to me that `kCaptured` is used for an inner (nested) graph that some inputs/outputs might not be visible in value_by_name_of so it creates a dummy node for them with `kCaptured` type.

gramalingam (approved):

Issue #4033, opened 2022-02-21
Spec issue: GridSample


broune:
This is a collection of issues encountered with GridSample in the ONNX spec.

It's not clear to me if this is the proper method/format/venue to report these - let me know if I should be doing something else.

For reference: [current ONNX spec for GridSample](https://github.com/onnx/onnx/blob/main/docs/Operators.md#GridSample) and [PyTorch GridSample](https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html)

I also have [a draft spec change proposal](https://github.com/broune/onnx/blob/main/docs/Operators.md#GridSample), which simply makes some assumptions about how the issues below will be resolved and so would likely need some updates later. Regardless of those decisions, this draft is far more detailed in explaining the behavior of GridSample than anything else I've seen, and may be useful for anyone seeing this who has not spent a lot of time looking at GridSample recently. (I also can't figure out how to display LaTeX equations in the spec, as you will notice if you read that link, is LaTeX possible / allowed in the spec?)

## Errors
1) The description of how the reflection padding_mode works for points close to the border is true for align_corners=0 but not when align_corners=1 (which moves the axis of mirroring, so it’s different). Both the Microsoft ONNX implementation and PyTorch do something different for align_corners=1 and doing so is also consistent with the description in the spec of how this works for points far from the border.

2) The spec allows the grid input to contain complex numbers without defining what that means. I don’t know what it would mean to access an array at a complex number (with or without interpolation) and I think allowing this is an error.

3) The X input is allowed to be a complex number (which is fine), but the output type specification does not include a complex number, which seems like an error, since complex number input would yield complex number output (assuming the obvious interpretation of how complex number input should be handled).

## Missing details
4) The spec does not specify how normalized coordinates are denormalized.

5) The spec does not specify that the minor coordinate comes before the major coordinate, though it can be seen from the first example (which matches PyTorch). This is surprising and opposite from the order in gather_nd, so would be good to specify explicitly.

6) It’s not specified what the element type of the output should be. If the X input is floating point, it makes sense that the output would use the same element type, but this is not specified anywhere. Also, in that case, what precision should internal calculations use? This is a bigger question for integer inputs - it’s really not clear what the output type should then be. Presumably some floating point type.
 
[...]

Issue #3724, opened 2021-09-18
Tensor dimensions of size zero


kraiskil:
How should a tensor dimension of size zero be interpreted? Is that dimension ignored, or does the tensor have zero elements even if other dimensions are of non-zero size? I could not find this mentioned in the documentation.

The refernce output of backend test 'slice start out of bounds' has the shape (20,0,5) 
`onnx/onnx/backend/test/data/node/test_slice_start_out_of_bounds/test_data_set_0/output_0.pb `
The output is generated by slicing a tensor with one dimension reduced to size zero.

This is causing problems when generating C code where arrays are of size float[20][0][5]. I wonder if it is safe to just not produce any output from the node in the above test?

TomWildenhain-Microsoft:
Ah yes, zero-element tensors are interesting and important to get right. Tensors with shapes (20, 0, 5) and (5, 0, 20) might seem the same (they both have no elements) but there are some real situations where the differing shape data for zero element tensors is important.

Suppose someone has a tensor of shape (100, 3, 128, 256) where the first dimension is a batch dimension and the last two are the height/width of an image.  Maybe they filter for entries with some criteria and get no results producing a tensor of shape (0, 3, 128, 256). If they use a Shape op to get the image dimensions, the dimensions 128 and 256 still must be preserved correctly.

In a similar situation, the user might filter the images twice and concatenate the results. If one of those is empty, the dimensions still need to match for the concatenate to work.

That said, zero element tensors are rare and for practical purposes dealing with them correctly often doesn't matter. But when it does, errors can lead to bugs that are hard to find and may be difficult to fix without updating the ONNX spec for an op.

Mathematically, a tensor can be thought of as pair of sequences/arrays (V, S) where S is the shape of the tensor (a list of non-negative integers) and V is a list of values with length equal to the product of the dimensions in S.  Two tensors (V, S) and (V', S') are equal if and only if V = V' and S = S'.  This has some interesting implications:

 - S can contain dimensions of value 0. If any dimensions are 0, V must have length 0.
 - If S has length 0, V must have length 1, since the empty product is 1. In this case, the tensor represents a scalar.
 - If S has length 1, V has length equal to the single dimension in S. In this case, the tensor represents a vector.
 - A vector of length one has shape [1] and a single element. A scalar has shape [] and a single element. Scalars are _not_ vectors of length 1.

To make things additionally complicated, ONNX's Shape op returns a _tensor_, not a list as in numpy, which itself has a shape.

[...]

PR #4089, opened 2022-03-24
GridSample spec change proposal

broune:
**Description**
Proposal for updating spec on GridSample. See https://github.com/onnx/onnx/issues/4033 .

**Motivation and Context**
See https://github.com/onnx/onnx/issues/4033 .

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -10016,19 +10016,200 @@ Other versions of this operator: <a href="Changelog.md#GreaterOrEqual-12">12</a>
 
 ### <a name="GridSample"></a><a name="gridsample">**GridSample**</a>
 
-  Given an input `X` and a flow-field `grid`, computes the output `Y` using `X` values and pixel locations from `grid`.
-  Currently, only spatial (4-D) inputs are supported. For input `X` with shape (N, C, H, W) and `grid` with shape (N, H_out, W_out, 2),
-  the output `Y` will have shape (N, C, H_out, W_out).
+  GridSample is conceptually similar to gather_nd but allows access to data at non-integer
+  coordinates by interpolating from nearby integer coordinates. GridSample also
+  allows access at coordinates beyond the size of the input `X`.
 
-  The tensor `X` contains values at centers of square pixels in a H by W 2-dimensional image.
-  The tensor `grid` describes normalized positions where the output `Y` is to be computed
-  using a specified interpolation method (the mode) and a padding mode (for grid positions falling outside the 2-dimensional image).
+  Given a 4D `X` tensor with dimensions `[N,C,H,W] ` and a 4D `grid` tensor
+  with dimensions `[N,W',H',2]`, the 4D output 'Y' tensor of `GridSample` has
+  dimensions `[N,C,W',H']`. Conceptually, the output element at `[n,c,row',col']`
+  equals `X[n,c,row,col]` where `(row, col)` is calculated from the 2-element
+  vector `(x,y)` that is equal to `grid(n,row',col')`. Conceptually, that is what
+  is going on. The details are more involved as described below.
 
-  Elements in `grid[N, H_out, W_out]` are size-2 vectors specifying positions in the 2-dimensional space of `X`.
-  They are used to interpolate output values of `Y[N, C, H_out, W_out]`.
+  See also in [torch.nn.functional.grid_sample](https://pytorch.org/docs/master/generated/torch.nn.functional.grid_sample.html#torch-nn-functional-grid-sample). Currently, ONNX GridSample requires 4D tensors.
 
-  The GridSample operator is often used in doing grid generator and sampler in the [Spatial Transformer Networks](https://arxiv.org/abs/1506.02025).
-  See also in [torch.nn.functional.grid_sample](https://pytorch.org/docs/master/generated/torch.nn.functional.grid_sample.html#torch-nn-functional-grid-sample).
+  #### Denormalization of coordinates depending on `align_corners`
+
+  A coordinate vector is in-bounds if it stays within the dimension sizes of
+  `X` in the height and width dimensions. A coordinate vector in GridSample is
+  not required to be in-bounds. This section concerns in-bounds coordinate vectors
+  while out-of-bounds coordinate vectors are handled in a later section.
+
+  The in-bounds coordinate vectors in `grid` are normalized to the range `[-1,1]`.
+  The indices of `X` are in a range `[0,W)` and `[0,H)`, respectively, so we
+  need to denormalize the coordinates in `grid` to put them into a range more
+  appropriate for indexing into `X`. The calculation to do the denormalization
+  depends on the value of `align_corners`.
+
+  If `align_corners` is `0` (i.e. `false`), then for the horizontal (left-right)
+  dimension, `-1` refers to the left border of the left-most column of pixels while
+  `1` refers to the right border of the right-most column of pixels. The formula for
+  denormalization of a coordinate `x` is then:
+
+  `denormalize(x, W, false) = ((x + 1) * W - 1) / 2` .
+
+  If `align_corners` is `1` (i.e. `true`), then for the horizontal (left-right)
+  dimension, `-1` refers to the center of the left-most column of pixels while `1`
+  refers to the center of the right-most column of pixels. The formula for
+  denormalization of a coordinate `x` is then:
+
+  `denormalize(x, W, true) = (x + 1) * (W - 1) / 2` .
+
+  The normalized coordinates in `grid` are notated as `(x, y)` while the
+  denormalized coordinates used for lookups are notated as `(row, col)`, matching
+  PyTorch's GridSample. The concept of `x` is horizontal (left-right) which
+  corresponds to `col`, while `y` is vertical (up-down) which corresponds to
+  `row`. So in order to do a lookup in `X` based on `(x, y)`, it is also
+  necessary to swap the two coordinates in addition to denormalization. All
+  together, the coordinate calculation is then:
+
+  `(row, col) = (denormalize(y, H, align_corners), denormalize(x, W, align_corners))`
+
+  where `y` goes first while in `grid` the ordering is `(x, y)` with `x` first. In
+  other words, `grid` lists its coordinates in minor-to-major ordering (`y` and
+  `row` are the major dimensions while `x` and `col` are minor). This is in
+  contrast to e.g. `gather_nd` which uses the opposite ordering. So if `(-1, -1)`
+  is the top-left corner, then `(-1, 1)` refers to the bottom-left corner, not the
+  top-right corner.
+
+  #### Interpolation of coordinates depending on `interpolation_mode`
+
+  If the denormalized coordinates `(row, col)` are integer and in-bounds, then the
+  value in the output at those coordinates is the value of `X` at those
+  coordinates. If the denormalized coordinates are not integer, then the value is
+  interpolated depending on the value of `interpolation_mode`.
+
+  If `interpolation_mode` is `nearest`, then the value at those
+  coordinates are the values in `X` of the nearest integer coordinates. This
+  corresponds to the calculation
+
+  `(row', col') = (round(row), round(col))` .
+
+  If two integer points are equally close in a given dimension,
+  corresponding to a coordinate ending in `.5`, then a rounding to the nearest
+  even integer occurs.
+
+  For example, if `(row, col)=(1.5,3.5)`, then the coordinates used for the lookup
+  could be any one of `(1.0, 3.0)`, `(1.0, 4.0)`, `(2.0, 3.0)`, and `(2.0, 4.0)`.
+  Whichever one of those a given version of an implementation picks, it must
+  always make that same choice when given this same input.
+
+  If `interpolation_mode` is `bilinear`, then the value at the output depends on a
+  linear interpolation between the nearest four integer coordinates. The weight of
+  the nearby integer points depends on the (Manhattan) distance to those points.
+  In the horizontal (left-right) dimension, the weighting of the two nearest
+  points to the left equals `frac(col)` which is the fractional part of `col`
+  while the weighting of the two nearest points to the right equals `1 -
+  frac(col)`.
+
+  For example, if `(row, col) = (1.3, 3.9)` then the interpolated value is the
+  matrix product:
+
+  ```math
+  \text{Y[n,c,row',col'] } =
+  \left(\begin{array}{c}
+  0.9 & 0.1
+  \end{array}\right)
+  \left(\begin{array}{cc}
+  \text{X[n,c,1,3]} & \text{X[n,c,1,4]} \\
+  \text{X[n,c,2,3]} & \text{X[n,c,2,4]} \\
+  \end{array}\right)
+  \left(\begin{array}{cc}
+  0.3 \\
+  0.7 \\
+  \end{array}\right)
+  ```
+
+  If `interpolation_mode` is `bicubic` then it is a similar
+  calculation as for linear interpolation, except the nearest 4x4 integer points
+  (instead of 2x2) are sampled around `(row, col)` and therefore a 4D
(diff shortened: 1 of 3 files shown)


hwangdeyu:
Hi @broune, please follow the DCO required to sign-off all the commits.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
 #### Attributes
 
 <dl>
-<dt><tt>align_corners</tt> : int (default is 0)</dt>
-<dd>If align_corners=1, the extrema (-1 and 1) are considered as referring to the center points of the input's corner pixels. If align_corners=0, they are instead considered as referring to the corner points of the input's corner pixels, making the sampling more resolution agnostic.</dd>
+<dt><tt>align_corners</tt> : int32 (default is 0)</dt>
```
hwangdeyu, line 7829: It should be `int` in Attributes.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
+
+
 #### Inputs
 
 <dl>
-<dt><tt>X</tt> (differentiable) : T1</dt>
-<dd>4-D tensor of shape (N, C, H, W), where N is the batch size, C is the numbers of channels, H and W are the height and width of the input data.</dd>
-<dt><tt>grid</tt> (non-differentiable) : T1</dt>
-<dd>Input offset, 4-D tensor of shape (N, H_out, W_out, 2), where H_out and W_out are the height and width of grid and output, Grid specifies the sampling pixel locations normalized by the input spatial dimensions. Therefore, it should have most values in the range of [-1, 1]. If grid has values outside the range of [-1, 1], the corresponding outputs will be handled as defined by padding_mode.</dd>
+<dt><tt>input</tt> (differentiable) : T1</dt>
```
hwangdeyu, line 7846: We set this input name as `X` in kernel implementation and example. So `X` should be the input1 name.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
+
+<dt><tt>grid</tt> (non-differentiable) : T2</dt>
+<dd>4-D tensor of shape (N, H_out, W_out, 2), where H_out and W_out are the height and width of grid and output. This is the tensor that GridSample uses to determine where to do lookups in input. This parameter is allowed to be integer, but this would typically indicate a mistake as this can only access the input data at its corners. It is not allowed for N, H_out or W_out to be zero.</dd>
 </dl>
 
 #### Outputs
 
 <dl>
-<dt><tt>Y</tt> (differentiable) : T2</dt>
+<dt><tt>output</tt> (differentiable) : T1</dt>
```
hwangdeyu, line 7856: `Y`, same with `X`.

hwangdeyu:
@XiyinOSS @askhade @gramalingam Could you also take a review of this PR that detailed the `GridSample` op? thanks!

docs/Operators.md
```diff
@@ -7611,17 +7611,213 @@ Other versions of this operator: <a href="Changelog.md#GreaterOrEqual-12">12</a>
...
+In the horizontal (left-right) dimension, the weighting of the two nearest
+points to the left equals `frac(col)` which is the fractional part of `col`
+while the weighting of the two nearest points to the right equals `1 -
+frac(col)`.
+
+For example, if `(row, col) = (1.3, 3.9)` then the interpolated value is the
+matrix product:
+
+```math
+\text{output[n,c,row',col'] } =
```
hwangdeyu, line 7713: Not Sure if there are someway to display the `mathematical formula` well in markdown file.

docs/Operators.md
broune: This is one of the issues pointed out in the doc. How many bits and signedness is "int"?

docs/Operators.md
broune: Yep, let me fix that.

docs/Operators.md
broune: On gitlab this example (or something like it, I tried a few different versions before putting this up here) displays as beautiful LaTeX rendered math. Seems like we'd want to be able to describe math in something like a ML op spec since everything is very mathematical in ML?

There are some people with thoughts on that here, there appear to be some workarounds possible on github:

  https://stackoverflow.com/questions/35498525/latex-rendering-in-readme-md-on-github

Maybe we could have an HTML mode or something where this works?

In any case, we should have an answer about how to describe math if we don't have typesetting for it. Maybe some sort of plain ascii-based math?

hwangdeyu: Hi @broune, 
After my confirmation, ONNX spec file operator.md does not support the LaTeX formula so for. But there are some plans to support it in the next half year, and some of the limitations of the operators will be fixed at the same time.
Just as you said, we might have to describe math like the way of [RNN-14](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#rnn-14) describing the math of activation functions.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
 <dt><tt>T2</tt> : tensor(float16), tensor(float), tensor(double)</dt>
-<dd>Constrain output types to float tensors.</dd>
+<dd>Constrain grid operand type to float tensors.</dd>
 </dl>
 
+#### Unresolved
+
+**Some concerns from the issue reported on GridSample that are not resolved here:
+
+1. We need dimensions higher than 4D for GridSample. Can the spec run
```
hwangdeyu, line 7873: Thanks for the detailed and clear issue contributions. That's a very good idea for 5D support. But if we add the 5d spec description, users who are not familiar with this op would be confused if they export their model with it in 5D and failed. So it's not a good time to add it at present. And welcome to contribute the 5D kernel implementation of GridSample.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
+2. Math is not displaying properly on Github. Should it? How should
+math be typeset properly in the ONNX spec? Presumably we want
+LaTeX-type math since ML has a lot of math in it.
+
+3. For nearest_mode, I did not follow resize(). I removed ceil and
+floor, as I don't understand when those would be used. I added
+round_to_even and put it as the default, as it seems the most balanced
+/ unbiased rounding mode of these ones.
+
+4. I set align_corners as int32, instead of int. Many other ops just
```
hwangdeyu, line 7886: The `align_corners` is an attribute type `AttributeProto::INT` in the spec. It will get the Type from AttributeProto, which only includes the `int` by desgin in [here](https://github.com/onnx/onnx/blob/96516aecd4c110b0ac57eba08ac236ebf7205728/onnx/onnx.in.proto#L120). This `int` represents `int64`. Maybe it's ok to keep `int'.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
+
+5. I disallowd integers inputs. They make sense mathematically, but
+logically it would almost always be an errro (maybe actually just
+always) to use this, as you'd hit the grid only on the corners, which
+doesn't make a lot of sense and is almost certainly something the user
+didn't intend to do and probably indicates either a bug in the user
+code or a misunderstanding of how the op works. So I think it's better
+to get an error in this case, but let me know what you think.
+
+6. I disallowed zero dimensions. Not clear what the general thinking
```
hwangdeyu, line 7898: It seems the zero dimensions is rare, but it was supported in onnx https://github.com/onnx/onnx/issues/3724.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
+is in ONNX for zero dimensions. Maybe it should be allowed? I would
+normally allow it, but it's hard to interpolate among no information,
+not clear what e.g. the border padding mode should do in this case.
```
hwangdeyu, line 7901: Hi @broune , basically the spec need to consider the op implementation so we should keep the description in accordance with kernel. Do you need further help and we can complete this change proposal together?

docs/Operators.md
```diff
@@ -7611,17 +7611,213 @@ Other versions of this operator: <a href="Changelog.md#GreaterOrEqual-12">12</a>
...
+  `denormalize(x, W, false) = ((x + 1) * W - 1) / 2` .
+
+If `align_corners` is `1` (i.e. `true`), then for the horizontal (left-right)
+dimension, `-1` refers to the center of the left-most column of pixels while `1`
+refers to the center of the right-most column of pixels. The formula for
+denormalization of a coordinate `x` is then:
+
+  `denormalize(x, W, true) = (x + 1) * (W - 1) / 2` .
+
+The normalized coordinates in `grid` are notated as `(x, y)` while the
```
liqunfu, line 7662: Relation between x, y coordinate and row, col in data matrix are understandable in general. I suggest to either use x, y[, z] in op description. Avoiding using row, col shall reduce document length.

docs/Operators.md
```diff
@@ -7611,17 +7611,213 @@ Other versions of this operator: <a href="Changelog.md#GreaterOrEqual-12">12</a>
...
+coordinates are the values in `input` of the nearest integer coordinates. This
+corresponds to the calculation
+
+`(row', col') = (round(row), round(col))` .
+
+If two integer points are equally close in a given dimension,
+corresponding to a coordinate ending in `.5`, then a rounding occurs
+according to the value of the nearest_mode attribute.
+
+For example, if `(row, col)=(1.5,3.5)`, then the coordinates used for the lookup
```
liqunfu, line 7696: I do not see explicit statement on this from pytorch doc either. in pytorch:
https://github.com/pytorch/pytorch/blob/cda0d5a57b9126c6d244fdd5b02198f05c742615/aten/src/ATen/native/GridSampler.cpp#L177
std::round is used. 
However in onnxruntime, it uses std::nearbyintf which depends on the environment. 
I perfer Pytorch's deterministic approach. 
@hwangdeyu  do you know why ORT uses std::nearbyintf instead of std::round?

liqunfu (commented):
Thanks for the PR. Added my comments. I am looking forward to the final PR (real patch).

docs/Operators.md
gramalingam: I think the description is good in its current form. x and y refer to normalized-coordinates in the range [-1, 1], IIUC, and the denormalize function computes the row, col used to index into the input matrix. So, they are distinct.

docs/Operators.md
```diff
@@ -7611,17 +7611,213 @@ Other versions of this operator: <a href="Changelog.md#GreaterOrEqual-12">12</a>
...
+If `align_corners` is `1` (i.e. `true`), then for the horizontal (left-right)
+dimension, `-1` refers to the center of the left-most column of pixels while `1`
+refers to the center of the right-most column of pixels. The formula for
+denormalization of a coordinate `x` is then:
+
+  `denormalize(x, W, true) = (x + 1) * (W - 1) / 2` .
+
+The normalized coordinates in `grid` are notated as `(x, y)` while the
+denormalized coordinates used for lookups are notated as `(row, col)`, matching
+PyTorch's GridSample. The concept of `x` is vertical (left-right) which
```
gramalingam, line 7664: How is vertical "left-right" and horizontal  "up-down" ? Sounds confusing to me.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
 
+#### Unresolved
+
+**Some concerns from the issue reported on GridSample that are not resolved here:
+
+1. We need dimensions higher than 4D for GridSample. Can the spec run
+ahead of the Microsoft ONNX implementation or does it need to wait for
+implementation to be done first?
+
+2. Math is not displaying properly on Github. Should it? How should
```
gramalingam, line 7877: Good point, but orthogonal to this PR. I think we should handle this aspect separately, and just focus on GridSample in this PR.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
+math be typeset properly in the ONNX spec? Presumably we want
+LaTeX-type math since ML has a lot of math in it.
+
+3. For nearest_mode, I did not follow resize(). I removed ceil and
```
gramalingam, line 7881: Haven't verified this. But I think we shouldn't change the spec unnecessarily.

docs/Operators.md
```diff
@@ -7630,39 +7826,79 @@ This version of the operator has been available since version 16 of the default
...
+floor, as I don't understand when those would be used. I added
+round_to_even and put it as the default, as it seems the most balanced
+/ unbiased rounding mode of these ones.
+
+4. I set align_corners as int32, instead of int. Many other ops just
+say "int". Not sure what is correct here. I think the spec should say
+what the type is precisely.
+
+5. I disallowd integers inputs. They make sense mathematically, but
```
gramalingam, line 7890: Makes sense

docs/Operators.md
broune: I modified this so that the proposal now no longer changes the behavior of existing GridSample, it simply documents what's there better (other than disallowing integer and complex inputs and outputs - not sure if your existing implementation supports that).

docs/Operators.md
broune: It's displaying properly now! :)

docs/Operators.md
broune: Let me know if I need to change this, but rounding towards nearest even makes sense to me as an unbiased rounding mode.

docs/Operators.md
broune: Yes, it's unfortunate that one cannot avoid using both concepts since the op inherently involves both.

docs/Operators.md
```diff
@@ -7611,17 +7611,213 @@ Other versions of this operator: <a href="Changelog.md#GreaterOrEqual-12">12</a>
...
+allows access at coordinates beyond the size of the input.
+
+Given a 4D `input` tensor with dimensions `[N,C,H,W] ` and a 4D `grid` tensor
+with dimensions `[N,W',H',2]`, the 4D output tensor of `GridSample` has
+dimensions `[N,C,W',H']`. Conceptually, the output element at `[n,c,row',col']`
+equals `input[n,c,row,col]` where `(row, col)` is calculated from the 2-element
+vector `(x,y)` that is equal to `grid(n,row',col')`. Conceptually, that is what
+is going on. The details are more involved as described below.
+
+See also in [torch.nn.functional.grid_sample](https://pytorch.org/docs/master/generated/torch.nn.functional.grid_sample.html#torch-nn-functional-grid-sample). Currently, ONNX GridSample requires 4D tensors.
```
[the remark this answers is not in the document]
broune: Agreed, though I left this out for this proposal/pull request for now.

docs/Operators.md
broune: I would like to know how you can handle zero dimensions. I think there's no good way to handle them, so I assume it's an error.

docs/Operators.md
broune: It's not a question of whether zero dimensions should be supported for tensors in general, that's a definite "absolutely yes". The question is whether zero dimensions makes sense for the gridsample op in particular. There I think the answer is no, though if someone really wants to, it could make sense in some of the configurations of the op. I set it to be disallowed in all cases in this proposal / pull request.

docs/Operators.md
broune: I kept it to 4D in this pull request.

docs/Operators.md
broune: This now works.

docs/Operators.md
broune: It was described as input/grid/output in this pull request. Please reopen if the contention is that it's not allowed to be called anything other than X/grid/Y.

docs/Operators.md
broune: Good catch. Fixed.

broune:
Took a while, but this should now be done correctly in the .cc file and ready for merging / further review .

Issue #1934, opened 2019-04-12
Case sensitivity of operator names and attributes


fdwr:
The spec should be clear about case sensitivity when matching names of operators, attributes, and attribute string enums. That includes NodeProto::op_type, AttributeProto::name, and AttributeProto::s/strings.

## Applicable documentation:
https://github.com/onnx/onnx/blob/master/docs/IR.md
https://github.com/onnx/onnx/blob/master/docs/Operators.md#RNN
https://github.com/onnx/onnx/blob/master/onnx/onnx.proto

## Context:
We’ve been matching all of these in WinML kernels atop DirectML with exact case sensitivity (simpler/faster/less ambiguous), but we encountered a failing model exported from PyTorch where the RNN activation attribute “Tanh” was “tanh” instead :/. Granted, I see this as a bug in the exporter because the exporter should use the same case as the spec lists them, but the spec should make it clear in any case whether "tanh" should match too.

PR #4096, opened 2022-03-29
Clarify that op_type is case sensitive

fdwr:
This is consistent with current precedent. e.g. ONNX Runtime fails with "conv ... Error No Op registered for conv with domain_version of 7", as it must be "Conv".

Signed-off-by: Dwayne Robinson <dwayner@microsoft.com>

**Description**
- Update IR.md to clarify op type case sensitivity.

**Motivation and Context**
- *Why is this change required? What problem does it solve?* To clarify behavior.
- *If it fixes an open issue, please link to the issue here.* https://github.com/onnx/onnx/issues/1934

### DIFF
--- a/docs/IR.md
+++ b/docs/IR.md
@@ -160,7 +160,7 @@ The properties of an operator definition are:
 
 Name|Type|Description
 |---|---|---|
-op_type|string|The name of the operator, as used in graph nodes. MUST be unique within the operator set’s domain.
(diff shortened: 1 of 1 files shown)


fdwr:
FYI @jcwchen and @spandantiwari.

jcwchen (approved):
LGTM. Thank you for the spec clarification! Could @onnx/sig-archinfra-approvers  please review it?

PR #4113, opened 2022-04-04
Clarify NonZero behavior for scalar input in spec

jcwchen:
**Description**
- ~Fix shape inference for scalar input: NonZero; Since the behavior is changed, bump opset version to NonZero-17~
- Clarify NonZero behavior for scalar input in spec
- Add tests for scalar input (NonZero, UnSqueeze)

**Motivation and Context**
In the spec of NonZero, it said `NonZero behaves similar to numpy.nonzero`, but current shape inference for NonZero does not follow NumPy's behavior. For NumPy, NonZero(x) produce output shape (1, N) if x is a scalar. ONNX shape inference is doing the proper job here (output shape is (0, N)). So It should be clarified in NonZero's spec to prevent confusion. In that case, ONNX follows PyTorch's behavior instead of deprecated NumPy's behavior, which should make more sense. Further action needs to be done in ONNX Runtime, because the shape computation there follows NumPy's behavior.


onnx/test/shape_inference_test.py
```diff
@@ -69,6 +69,7 @@ def _compare_value_infos(self, vi_type: TypeProto, inferred_vi_type: TypeProto)
             assert vi_type.tensor_type.elem_type == inferred_vi_type.tensor_type.elem_type
             assert vi_type.tensor_type.HasField('shape') == inferred_vi_type.tensor_type.HasField('shape')
             if vi_type.tensor_type.HasField('shape'):
+                assert len(vi_type.tensor_type.shape.dim) == len(inferred_vi_type.tensor_type.shape.dim)
```
jcwchen, line 72: Previously if the dim_size of expected value_info is zero, the dim_size of inferred value_info can be anything because the following loop will never run. Add this line here to enhance the dim comparison.

onnx/test/shape_inference_test.py
```diff
@@ -632,6 +633,15 @@ def test_unsqueeze_negative_axes(self) -> None:
...
         self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (1, 3, 4, 5, 1))])
 
+    def test_unsqueeze_scalar(self) -> None:
+        graph = self._make_graph(
+            [('x', TensorProto.FLOAT, ()),
+            ('axes', TensorProto.INT64, ())],
+            [make_node('Unsqueeze', ['x', 'axes'], 'y')],
+            [],
+            initializer=[make_tensor('axes', TensorProto.INT64, (), (-1,))])
+        self._assert_inferred(graph, [make_tensor_value_info('y', TensorProto.FLOAT, (1,))])
```
jcwchen, line 643: Random thought: I found this test only works for UnSqueeze with opset_version >= 11 because older version does not support negative axes. To enhance test coverage, I suppose perhaps we can make every test in shape_inference_test.py be tested by every opset_version if the opset_version is not specified in the test.

onnx/defs/tensor/defs.cc
```diff
@@ -3144,7 +3144,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           auto* dim = output_shape.add_dim();
           if (hasInputShape(ctx, 0)) {
             const TensorShapeProto& input_shape = getInputShape(ctx, 0);
-            dim->set_dim_value(input_shape.dim_size());
+            dim->set_dim_value(input_shape.dim_size() == 0? 1 : input_shape.dim_size());
```
gramalingam, line 3147: I think the problem here is that the previous code is confusing the rank of the output-tensor and the dimensions of the tensor. The output is not 1-dimensional (based on the ORT implementation and numpy spec), since we return a tuple of indices for each non-zero value. So, the output is 2-dimensional, with one dimension-size being the input-tensor's rank and the other dimension having an unknown-value (== number of non-zero values in input).

This should be clarified better in the spec.

gramalingam: As to what happens for a scalar: The numpy spec deprecates this since version 1.17, asking users to explicitly convert it to at-least 1D. This is sort of related to above issue as well, because the output will have one-dimension of size 0 (rank of input tensor), unless it is special-cased to be at-least 1. I don't think we need to special-case this (unless there is a good reason, that is some model depending on this.)

onnx/test/shape_inference_test.py
gramalingam: I think the issue (before and after version 11) is that axes was converted from attribute to input (not relating to negative values)? Anyway, on the other question: I don't see this test-case mention any opset_verion?

onnx/defs/tensor/defs.cc
jcwchen: Thanks for the investigation and I understand your concern. Another fact I would like to point out is even today np.nonzero is still behaving like `nonzero(atleast_1d(arr))` although there is a deprecation warning. So if ONNX decides not to behave the same as Current NumPy, the spec should clarify it to prevent confusion. Also, the test in ORT will need to be updated.

onnx/test/shape_inference_test.py
jcwchen: > I think the issue (before and after version 11) is that axes was converted from attribute to input (not relating to negative values)?

It's because opset_version before 11 does not support negative axes.

> Anyway, on the other question: I don't see this test-case mention any opset_verion?

Without specified opset_version means using the latest opset_version. Going forward this test case should always be verified.

Regarding to the proposal here (make every test in shape_inference_test.py be tested by every opset_version if the opset_version is not specified in the test) May I understand your opinion about it @gramalingam ? Thanks.

onnx/version_converter/convert.h
```diff
@@ -740,6 +740,10 @@ class DefaultVersionConverter : public BaseVersionConverter {
         OpSetID(15), OpSetID(16)));
       registerAdapter(make_unique<CompatibleAdapter>("PRelu",
         OpSetID(15), OpSetID(16)));
+    
+      /******** 16 -> 17 ********/
+      registerAdapter(make_unique<CompatibleAdapter>("NonZero", OpSetID(16), OpSetID(17)));
```
gramalingam, line 745: This looks like a tricky edge case, since the behavior is not really preserved for the scalar case. If the input is known to be a scalar, we could fail or produce a warning, I suppose. But if the input rank is not known, unclear what we can do.

onnx/test/shape_inference_test.py
gramalingam: Hi Jacky, that's an interesting idea worth discussing (what the default behavior should be if no opset is specified), but I assume you mean that for a different PR, right? No need to mix that with this change for NonZero now, right?

jcwchen: Yes, that would be a big change and should be done by another PR. We can discuss more details about it in our next meeting. I will keep this PR only focuses on fixing shape inference for NonZero.

jcwchen:
Here is the summary for different libraries behavior:
- PyTorch
```
import torch
torch.nonzero(torch.tensor(0))
# tensor([], size=(0, 0), dtype=torch.int64)
torch.nonzero(torch.tensor(1))
# tensor([], size=(1, 0), dtype=torch.int64)
```
- TensorFlow
```
import tensorflow
tensorflow.where(tensorflow.constant(0))
# InvalidArgumentError: WhereOp : Unhandled input dimensions: 0 [Op:Where]
tensorflow.where(tensorflow.constant(1))
# InvalidArgumentError: WhereOp : Unhandled input dimensions: 0 [Op:Where]
```
- NumPy
```
import numpy as np
np.array(np.nonzero(0))
# array([], shape=(1, 0), dtype=int64)
np.array(np.nonzero(1))
# array([[0]]), shape (1, 1)
```

To sum up, it seems that old ONNX follows PyTorch's behavior, TensorFlow does not allow scalar input for Where (NonZero-like operator in TensorFlow). ONNXRuntime follows NumPy's behavior, which is what the ONNX document said.

gramalingam:
ONNX itself had an inconsistency (the documentation said "like numpy", but that was vague, but the shape-inference code was more specific and had the pytorch behavior). So, it seems best to fix the "old" op by updating the documentation to clarify the behavior to be aligned with shape-inference. 

For the "new" op: one question is: Does anyone want the numpy behavior?

Assuming some users want it, the best solution seems an upward compatible extension by adding an attribute "treat_0D_as_1D", which if set to 1 (true), will produce the numpy behavior.

Does this sound reasonable?

jcwchen:
Thank you for the conclusion @gramalingam. I agree with your points -- 

> ONNX itself had an inconsistency (the documentation said "like numpy", but that was vague, but the shape-inference code was more specific and had the pytorch behavior). So, it seems best to fix the "old" op by updating the documentation to clarify the behavior to be aligned with shape-inference.

+1 Since ONNX follows PyTorch and TensorFlow even does not allow scalar input. And also NumPy behavior will be deprecated. A proper direction should be following upstream like PyTorch or TensorFlow to prevent users' confusion. We should update the document to resolve ambiguity.

> For the "new" op: one question is: Does anyone want the numpy behavior?

Yes, it's an open question. Personally I think it shouldn't be a popular case since most of developers would probably follow PyTorch and TensorFlow. Still, this will very likely break backwards compatibility in ORT so it's better to include ORT core team to make the decision.

jcwchen:
@gramalingam before filing an issue in ORT, I just found another issue in ONNX -- it seems to me that current ONNX shape inference for NonZero with scalar input `1` will infer shape as (0, unknown), which is different from PyTorch (1, 0), although with scalar input `0` will infer shape as (0, unknown), which is the same as PyTorch (0, 0). Therefore, current ONNX shape inference does not exactly follow PyTorch's behavior. To fix it, if giving a scalar input, ONNX should infer shape like (unknown, 0) instead of (0, unknown), am I correct?

gramalingam:
@jcwchen : just to clarify: I think you mean that the order of dimensions is different for pytorch (N X D) and numpy (DxN), where D denotes the rank of the input tensor and N denotes the number of non-zero values. Is that correct? It is not specific to scalars. I don't think we should "fix" it, since it changes the spec. I assume that the ORT implementation returns the same as what ONNX shape inference says (DxN) for non-scalars? If so, we should continue to use DxN. We should clarify all this in the op. documentation, but not change the spec or shape-inference, I think. It may be useful to check what the pytorch-to-onnx converter does (it would need to introduce a transpose?). What about tf.where for non-scalars? Does it return NxD or DxN?

jcwchen:
Thank you for the clarification! That's really good to know -- so this difference is a global issue (PyTorch (N X D) v.s. NumPy, ONNX (DxN)) and it's not a specific issue here.

> What about tf.where for non-scalars? Does it return NxD or DxN?

I think TensorFlow behaves like PyTorch -- NxD. For instance
```
import tensorflow as tf
tf.where([2, 4, 0, 5]).shape
# TensorShape([3, 1])
```

> It is not specific to scalars. I don't think we should "fix" it, since it changes the spec. I assume that the ORT implementation returns the same as what ONNX shape inference says (DxN) for non-scalars? If so, we should continue to use DxN. We should clarify all this in the op.

Agree -- if ONNX behaves differently from PyTorch and TensorFlow, it's better to clarify it in the document.

gramalingam:
Hi @jcwchen : would be useful to update the PR description/title, now that this has changed into just adding test-cases

jcwchen:
> Hi @jcwchen : would be useful to update the PR description/title, now that this has changed into just adding test-cases

Thanks for the reminder. I have updated this PR's title, description and content. Please review it again. Thank you!

To summarize, ONNX decides to follow PyTorch's behavior and keep the original shape inference for NonZero. Furthermore, clarify NonZero's behavior for scalar input to prevent confusion.

onnx/defs/tensor/utils.h
```diff
@@ -31,4 +31,12 @@ void resizeShapeInferenceHelper_opset7_to_10(
     const TensorShapeProto& input_shape,
     const std::vector<float>& scales_data,
     TensorShapeProto* output_shape);
+
+static const char* NonZero_ver9_doc = R"DOC(
```
gramalingam, line 35: I guess this should not be `static`?

gramalingam (approved):

onnx/defs/tensor/utils.h
jcwchen: I suppose it must be static here. Otherwise, it will be defined twice in onnx/defs/tensor/def.cc and onnx/defs/tensor/old.cc and bump into a warning.

gramalingam: well, but if the goal is to avoid duplicate copies, we need to move it into one of the .cc files, and leave it as a simple extern declaration in the include file? Otherwise, wouldn't you end up with multiple copies anyway? (In any case, there is a CI failure relating to this, that needs some resolution.)

jcwchen: Thanks for catching this! Yes, I believe extern without static should be the right approach to avoid duplicate copies. Will update it later and let you review.

PR #4117, opened 2022-04-06
Fix typo in slice doc

### LINKED ISSUE #4115: Seeking clarity on Slice operator's clamping
# Ask a Question

### Question

In https://github.com/onnx/onnx/pull/3908/ which seems to have clarified Slice's operation, there seems to be a contradiction with how Numpy does clamping  (the numpy link reference is also added to the op's spec) versus what is prescribed by the clarification. Furthermore, the clarification is without an opset bump for Slice (indicating that this was the behavior all along).

Prescribed start clamping:
Then `start[axes[i]]` is the adjusted `starts[i]` clamped into range of valid indices, i.e. `[0, dims[axes[i]]-1]`.

This seems to indicate that starts clamping is slicing direction oblivious.

Here is an illustrative example from Numpy which indicates starts clamping is slicing direction dependent.

x = np.array([1, 2, 3])

Positive stepping
>>> x[5:6:1]
array([], dtype=int32)

5 is not clamped to a valid range which would be 2 which would have resulted in a non-empty result

Negative stepping
>>> x[5:0:-1]
array([3, 2])

5 is clamped to a valid range (2) and this results in the printed value.

### LINKED ISSUE #4116: Incorrect wording in Slice operator's spec
# Bug Report

### Describe the bug
In https://github.com/onnx/onnx/pull/3908, the following line was added:

  The clamping for the adjusted `ends[i]` depends on the sign of `steps[i]` and must
  accommodate copying 0 through `dims[axes[i]]` elements, so for positive stepping
  `end[axes[i]]` is clamped to `[0, dims[axes[i]]]`, while for negative stepping it
  is clamped to `[-1, ends[i]-1]`.

I think the final `...ends[i]-1` must read `dims[axes[i]] - 1` if I am understanding correctly

diyessi:
Signed-off-by: Scott Cyphers <cyphers@lightmatter.co>

**Description**
- Fix a typo in Slice doc for negative stepping end.

**Motivation and Context**
- Fixes #4115 
- Fixes #4116


hariharans29:
Thanks. This certainly does fix #4116.

For #4115, Numpy's clamping logic looks more like this for the `starts` (based on the shown examples in #4115): https://github.com/microsoft/onnxruntime/blob/26fceca90f35fe94f39864ebaec0f2d51e367f10/onnxruntime/core/providers/cpu/tensor/slice_helper.h#L132

So until that is addressed, #4115 won't be resolved by this PR as is

diyessi:
> Thanks. This certainly does fix #4116.
> 
> For #4115, Numpy's clamping logic looks more like this for the `starts` (based on the shown examples in #4115): https://github.com/microsoft/onnxruntime/blob/26fceca90f35fe94f39864ebaec0f2d51e367f10/onnxruntime/core/providers/cpu/tensor/slice_helper.h#L132. So until that is addressed, #4115 won't be resolved by this PR as is.

Should be okay now. This is also a good time to fix anything that could be worded more clearly.

hariharans29:
> > Thanks. This certainly does fix #4116.
> > For #4115, Numpy's clamping logic looks more like this for the `starts` (based on the shown examples in #4115): https://github.com/microsoft/onnxruntime/blob/26fceca90f35fe94f39864ebaec0f2d51e367f10/onnxruntime/core/providers/cpu/tensor/slice_helper.h#L132. So until that is addressed, #4115 won't be resolved by this PR as is.
> 
> Should be okay now. This is also a good time to fix anything that could be worded more clearly.

LGTM. Thanks for the quick fixes and thanks for taking the time to clarify the spec in the first place !

docs/Operators.md
```diff
@@ -20177,12 +20177,13 @@ expect(node, inputs=[x], outputs=[y],
...
   where `dims` are the dimensions of `input`. Then `start[axes[i]]` is the adjusted
-  `starts[i]` clamped into range of valid indices, i.e. `[0, dims[axes[i]]-1]`.
+  `starts[i]` is clamped into range of `[0, dims[axes[i]]]` for positive stepping
+  and `[0, dims[axes[i]]-1]` for negative stepping.
 
   The clamping for the adjusted `ends[i]` depends on the sign of `steps[i]` and must
   accommodate copying 0 through `dims[axes[i]]` elements, so for positive stepping
   `end[axes[i]]` is clamped to `[0, dims[axes[i]]]`, while for negative stepping it
-  is clamped to `[-1, ends[i]-1]`.
+  is clamped to `[-1, dim[axes[i]]-1]`.
```
hariharans29, line 20186: dim -> dims for consistency

diyessi: Fixed, along with better wording in the starts change.

hariharans29 (approved):
LGTM

hariharans29:
@gramalingam : Any comments ?

gramalingam (approved):
LGTM, thanks for fixing this!

PR #4118, opened 2022-04-07
Fix a bug in _get_initializer_tensors()

wangyems:
**Description**
This change supports getting all initializers and attribute tensors from an onnx model. e.g A model with multiple/multi-layer subgraphs

**Motivation and Context**

Previously saving a model containing subgraphs(protobuf size > 2G) with external data format will fail because tensors in subgraphs are not returned from the _get_all_tensors() so they are not set to external data. This change fix that.


onnx/external_data_helper.py
```diff
@@ -185,10 +185,22 @@ def _get_all_tensors(onnx_model_proto: ModelProto) -> Iterable[TensorProto]:
...
 
+def _get_initializer_tensors_from_graph(onnx_model_proto_graph: GraphProto) -> Iterable[TensorProto]:
+    """Create an iterator of initializer tensors from ONNX model graph."""
+    for initializer in onnx_model_proto_graph.initializer:
+        yield initializer
+    for node in onnx_model_proto_graph.node:
+        for attr in node.attribute:
+            if attr.type == AttributeProto.AttributeType.GRAPH:
+                yield from _get_initializer_tensors_from_graph(attr.g)
+            if attr.type == AttributeProto.AttributeType.GRAPHS:
```
jcwchen, line 196:
```suggestion
            if attr.type == AttributeProto.GRAPHS:
```
Mypy typecheck failed. Using AttributeProto.GRAPHS should help. L194 ditto.

jcwchen (commented):
I think generally ONNX utilities do not handle the subgraph well. Thank you for catching it!

> Ideally, we should yield attribute tensors recursively in a similar way, but I think it's very uncommon to see subgraph attributes tensors are really big so leave _get_attribute_tensors() untouched.

You concern is valid, but IMO logically it will be still better that the utility (_get_attribute_tensors()) can do it with the correct way. (although it is an uncommon case)

onnx/external_data_helper.py
wangyems: Thanks for reviewing it. I'll add this functionality to attribute tensors as well.

jcwchen (approved):
LGTM. Thank you for the PR!

gramalingam (approved):

PR #4072, opened 2022-03-18
[WIP] Refactor function shape inference implementation

gramalingam:
**Description**

This is part 2 of refactoring shape inference.
(A continuation of the PR: https://github.com/onnx/onnx/pull/4069 )

PR #4127, opened 2022-04-11
Shape inference refactoring part 2

gramalingam:

**Description**
This is part 2 of refactoring shape inference.
(A continuation of the PR: https://github.com/onnx/onnx/pull/4069 )

Closing https://github.com/onnx/onnx/pull/4072 and opening this to deal with signoff issues.

Fixes https://github.com/onnx/onnx/issues/4131


onnx/shape_inference/implementation.cc
```diff
@@ -380,7 +380,7 @@ class ShapeInferenceImplBase {
         if (!n.output(i).empty())
           updateType(n.output(i), ctx.getOutputType(i));
       }
-
+
```
jcwchen, line 383: nit: remove unnecessary spaces

onnx/shape_inference/implementation.cc
```diff
@@ -634,140 +722,18 @@ void InferShapeForFunctionNode(
...
+  ShapeInferenceImplBase base(
+      &g,
+      {}, //outer_scope_value_types_by_name,
+      func_opset_imports,
+      options,
+      symbol_table,
+      model_local_functions_map,
+      *generated_shape_data_by_name, // We need to add generated_shape_data_by_name into base
+      schema_registry);
+      // ir_version); Q: Whether IR_VERSION effects functionProto? If so, shall we make it consider ir_version?
```
jcwchen, line 735: Remove Q: comments. Ditto L511

garymm:
Please add `Fixes #4131` to PR description.
Thanks!

onnx/shape_inference/implementation.cc
```diff
@@ -471,6 +487,71 @@ class ShapeInferenceImplBase {
     }
   }
 
+
+ void process(const NodeProto& n, std::unordered_map<std::string, const AttributeProto*> attr_map) {
```
jcwchen, line 491: Duplicate function? (L396 `void process(NodeProto& n, std::unordered_map<std::string, const AttributeProto*> attr_map)`). If both of them are required, is it possible to make them common in certain way?

onnx/shape_inference/implementation.cc
```diff
@@ -471,6 +487,71 @@ class ShapeInferenceImplBase {
...
+        auto type_proto = ctx.getOutputType(i);
+        type_proto->CopyFrom(*(iter->second));
+      }
+    }
+
+ }
+
  public:
   ShapeInferenceImplBase(
       GraphProto* gp,
```
jcwchen, line 557: tiny nit: shall we use `g_in` instead of `gp` to make the naming convention same as others?

jcwchen (commented):
Some nit comments. Before merging it, it would be super great if we can test it in ORT to prevent issues.

onnx/shape_inference/implementation.cc
gramalingam: Good catch. I think it is a duplicate from a merge produced by git. I am mystified why the compiler does not complain, since both have same signature?

gramalingam: I see, they differ in `const` ...

onnx/shape_inference/implementation.cc
gramalingam: Done

jcwchen (approved):
> Before merging it, it would be super great if we can test it in ORT to prevent issues.

Actually currently ORT is improving its function implementation. Therefore, after offline discussion with Rama, we decided to test it in ORT in next stage (shape inference refactoring part 3).

LGTM. Thanks!

PR #4128, opened 2022-04-12
Fix output of `ScatterND` examples in Operators.md

maxstate:
**Description**
- There are wrong output in ScatterND example

**Motivation and Context**
- The wrong example needs to be fixed


gramalingam:
Thanks for your contribution. Please note that the operator documentation is auto-generated. The fix needs to go into the backend test case here: https://github.com/onnx/onnx/blob/1b8a9e3c6565e9a287fec40def0031f944f9c235/onnx/backend/test/case/node/scatternd.py#L73 and the documentation need to be generated (running `gendoc.py` under `onnx\defs`)

PR #4135, opened 2022-04-13
[WIP] Remove all related things about protobuf stub for mypy

jcwchen:
**Description**
Remove all related things about Protobuf stub for mypy

**Motivation and Context**
Follow-up work for https://github.com/onnx/onnx/pull/3939. That PR introduces Protobuf stub from types-protobuf package directly instead of keeping a Protobuf stub code in ONNX repo. We should further remove all related things.


jcwchen:
I was wrong. These files are still needed to create protobuf stub even with types-protobuf package. Close here now.

PR #4136, opened 2022-04-13
Extend parser to handle other types

gramalingam:

**Description**
Extend the parser to handle other types beyond tensor type.

**Motivation and Context**
Fixes https://github.com/onnx/onnx/issues/4043


onnx/defs/parser.h
```diff
@@ -138,6 +142,10 @@ class KeyWordMap {
     map_["model_version"] = KeyWord::MODEL_VERSION;
     map_["doc_string"] = KeyWord::DOC_STRING;
     map_["metadata_props"] = KeyWord::METADATA_PROPS;
+    map_["seq"] = KeyWord::SEQ_TYPE;
```
jcwchen, line 145: tiny nit: it seems that seq is the only abbreviation here for other types. Perhaps we can just use sequence here for consistency, but it's up to you.

jcwchen (approved):
LGTM. Thank you for the enhancement!

onnx/defs/parser.h
gramalingam: Thanks. I am using the abbreviations already used for all these types, in the opschema registrations. See, for example: https://github.com/onnx/onnx/blob/00738d2032ca0a39d9124b02085a90adda96d018/onnx/defs/schema.h#L668

liqunfu (approved):

onnx/test/cpp/parser_test.cc
```diff
@@ -67,6 +68,37 @@ TEST(ParserTest, TypeTest) {
...
+  // optional type:
+  Parse(type, "optional(float)");
+  EXPECT_TRUE(type.has_optional_type());
+  auto& optelttype = type.optional_type().elem_type();
+  EXPECT_TRUE(optelttype.has_tensor_type());
+  EXPECT_EQ(optelttype.tensor_type().elem_type(), float_type);
+  EXPECT_TRUE(optelttype.tensor_type().has_shape());
+
+  // optional type:
+  Parse(type, "sparse_tensor(float[1000])");
```
wschin, line 89: Maybe we can extend the test to cover all cases mentioned in
```
            // Grammar:
            // float indicates scalar (rank 0)
            // float [] indicates unknown rank tensor (not a zero rank tensor)
            // float [one-or-more-dimensions] indicates tensor of known rank > 0.
```
?

onnx/defs/parser.cc
```diff
@@ -86,8 +86,71 @@ Status OnnxParser::Parse(TypeProto& typeProto) {
...
     }
-  } else
-    return ParseError("Unexpected type.");
+  } else {
+    switch (KeyWordMap::Lookup(id)) {
+      case KeyWordMap::KeyWord::SEQ_TYPE: {
+        // Grammar: seq ( type )
+        MATCH('(');
+        auto* seqtype = typeProto.mutable_sequence_type();
+        PARSE(*seqtype->mutable_elem_type());
```
wschin, line 95: Does this support nested types such as `seq(seq(float))`? Maybe throw if we don't want nested types.

onnx/test/cpp/parser_test.cc
```diff
@@ -67,6 +68,37 @@ TEST(ParserTest, TypeTest) {
   Parse(type, "float[N,?,K]");
   EXPECT_FALSE(type.tensor_type().shape().dim(1).has_dim_param());
   EXPECT_FALSE(type.tensor_type().shape().dim(1).has_dim_value());
+
+  // sequence type:
+  Parse(type, "seq(float[])");
```
wschin, line 73: Maybe add tests for `seq(optional(float))` and `seq(float)`?

onnx/test/cpp/parser_test.cc
```diff
@@ -67,6 +68,37 @@ TEST(ParserTest, TypeTest) {
...
+  EXPECT_TRUE(optelttype.tensor_type().has_shape());
+
+  // optional type:
+  Parse(type, "sparse_tensor(float[1000])");
+  EXPECT_TRUE(type.has_sparse_tensor_type());
+  EXPECT_EQ(type.sparse_tensor_type().elem_type(), float_type);
+  EXPECT_EQ(type.sparse_tensor_type().shape().dim_size(), 1);
+
+  // map type:
+  Parse(type, "map(int32, float[N])");
```
wschin, line 95: Similar to other comments, maybe add more basic and fancy tests for `map`?

onnx/defs/parser.cc
gramalingam: We want to allow such nested types, and the parser allows it. It is up to operators to restrict allowed types.

wschin (approved):
This PR is good enough. I only have some comments about tests and clarification on nested types.

onnx/test/cpp/parser_test.cc
gramalingam: Those are already tested separately. Since the parser is recursive, I think the combinations should work fine. Not sure we need to enumerate all possibilities. However, I did add different forms for the different constructors (like seq, map, optional) so overall there is good coverage, I think.

PR #4140, opened 2022-04-14
Fix bug of resizeShapeInference for Resize13

xuzijian629:
**Description**
- Fix `resizeShapeInference` function in `onnx/defs/tensor/utils.cc`.
- It fails when the number of inputs of Resize is at most 2, which can happen for Resize13.
- Also, I removed `bool is_resize_op` because it's always set to `true`.

**Motivation and Context**
- Shape inference of Resize fails when the number of inputs is at most 2.


jcwchen (approved):
LGTM. Thanks for catching this!

gramalingam:
Thanks for the fix. Strictly speaking, the issue is that the interface-definition does not clarify what the contract is for these methods like `getInputData` (https://github.com/onnx/onnx/blob/bd849e86cb47e75b765ad120b1a1a05ef5cf8c10/onnx/defs/shape_inference.h#L85 ). I think it is safer for implementations of this interface to return `nullptr` for missing inputs (in cases like the above). This is preferable because the condition `i < ctx.getNumInputs()` is not sufficient to guarantee that input `i` is present.

gramalingam (approved):

jcwchen:
> I think it is safer for implementations of this interface to return nullptr for missing inputs (in cases like the above). This is preferable because the condition i < ctx.getNumInputs() is not sufficient to guarantee that input i is present in the general case. This particular change is fine, but I think it would help to ensure that implementations of InferenceContext are more robust.

I almost had the same thought about this while reviewing this PR. Then, we can prevent global issues like it. Still, definitely this PR is good to go for fixing current Resize issue. I will merge it soon. @gramalingam Would you mind opening an issue to track the work item you mentioned above? Thank you.

PR #4142, opened 2022-04-15
Check every dim has either `dim_param` or `dim_value` in getShapeInput

xuzijian629:
**Description**
- Check each dim is valid when referring to symbolic input in shape inference.

**Motivation and Context**
- Sometimes, due to shape inference error (which is most likely caused by some hidden bugs), an inferred shape may contain a dimension that has neither `dim_param` or `dim_value`.
- We should fall back to rank inference in getShapeInput in this case, rather than continue to pass this illegal shape.


xuzijian629:
Hmm… since other shape inference functions don’t specially handle these illegal dims, I think  I should work around somewhere else..

PR #4148, opened 2022-04-19
add sequence, map, and optional as model graph initializer

liqunfu:
**Description**
- To include SequenceProto, MapProto, OptionalProto as supported initializer types

**Motivation and Context**
- for the completeness, all data types shall be used as initializer to a model graph.


onnx/onnx.in.proto
```diff
@@ -451,6 +451,9 @@ message GraphProto {
   // Initializers (see above) stored in sparse format.
   repeated SparseTensorProto sparse_initializer = 15;
 
+  // Initializers (see above) stored in sparse format.
+  repeated SequenceProto sequence_initializer = 16;
```
gramalingam, line 455: Should we add something for "maps" and "optional" types at the same time? There is no request for those currently, but it helps to avoid multiple bumps to IR_version if we do them all together. What do you think? Thanks!

PR #4153, opened 2022-04-20
add attribute_proto to function_proto

liqunfu:
**Description**
- Add attribute proto to ONNX function proto so that attributes with default values can be added to an ONNX function

**Motivation and Context**
- ONNX function is used to define ONNX ops to avoid increasing total primary op count. Runtimes are able to execute function ops without implementing the ops' kernels. Runtimes may choose to implement an optimized execution kernel for a function op. An ONNX op can have attributes with optional default values. To enable function ops with the same behavior, in this PR, AttributeProto is added to FunctionProto. ONNX Function authoring can now assign default attribute values to a ONNX function. Default values are stored in AttributeProto for runtimes to consume. Example:
       input = '''
         <
            ir_version: 9,
            opset_import: [ "" : 15, "custom_domain" : 1],
            producer_name: "FunctionProtoTest",
            producer_version: "1.0",
            model_version: 1,
            doc_string: "A test model for model local functions."
          >
         agraph (float[N] x) => (float[N] out)
         {
            out = custom_domain.Selu<gamma=2.0, gamma=3.0>(x)
         }

         <
         domain: "custom_domain",
         opset_import: [ "" : 15],
         doc_string: "Test function proto"
         >
           Selu
           <alpha: float=1.67326319217681884765625, gamma: float=1.05070102214813232421875>
           (X) => (C)
           {
               constant_alpha = Constant<value_float: float=@alpha>()
               constant_gamma = Constant<value_float: float=@gamma>()
               alpha_x = CastLike(constant_alpha, X)
               gamma_x = CastLike(constant_gamma, X)
               exp_x = Exp(X)
               alpha_x_exp_x = Mul(alpha_x, exp_x)
               alpha_x_exp_x_ = Sub(alpha_x_exp_x, alpha_x)
               neg = Mul(gamma_x, alpha_x_exp_x_)
               pos = Mul(gamma_x, X)
               _zero = Constant<value_float=0.0>()
               zero = CastLike(_zero, X)
               less_eq = LessOrEqual(X, zero)
               C = Where(less_eq, neg, pos)
           }
         '''
default values are set for Selu function:
alpha: float=1.67326319217681884765625, gamma: float=1.05070102214813232421875

Selu function can be called without attribute, in which case the default values are used.
Selu function can also be called with attribute values, in which case the default values are overridden.
Function attributes can be mixed with attribute ids, for example:
<alpha: float=1.67326319217681884765625, gamma>
in this case alpha has a default value. It is saved in AttributeProto. gamma does not have a default value. It is saved as a string. 

A function attribute shall be represented either as a string attribute or an AttributeProto, not both.


onnx/onnx.in.proto
```diff
@@ -801,9 +804,16 @@ message FunctionProto {
   repeated string input = 4;
   repeated string output = 5;
 
-  // The attributes of the function.
+  // The attribute identifiers of the function.
```
gramalingam, line 807: nit: change "identifiers" to "parameters"

onnx/onnx.in.proto
```diff
@@ -801,9 +804,16 @@ message FunctionProto {
...
+  // It is for function attributes without default values.
```
gramalingam, line 808: nit: change "attributes" to "attribute parameters"

onnx/defs/parser.cc
```diff
@@ -43,6 +43,35 @@ Status OnnxParser::Parse(char open, IdList& idlist, char close) {
...
+      ParseOptionalIdentifier(id);
+      *idlist.Add() = id;
+    }
+  } while (Matches(','));
+  return Status::OK();
+}
+
+Status OnnxParser::Parse(char open, IdList& idlist, AttrList& attrlist, char close) {
+  idlist.Clear();
+  attrlist.Clear();
```
gramalingam, line 67: nit: The above two lines are duplicated in method above also. I wonder if we should just move this into an else-clause down below.

onnx/defs/parser.cc
```diff
@@ -43,6 +43,35 @@ Status OnnxParser::Parse(char open, IdList& idlist, char close) {
...
 }
 
+Status OnnxParser::Parse(IdList& idlist, AttrList& attrlist) {
+  idlist.Clear();
+  attrlist.Clear();
+  do {
+    std::string id;
+    bool is_assignment = PeekIsIdentifierAssignment(id);
+    if (id.empty())
+      return Status::OK(); // Treat as empty list of identifiers
```
gramalingam, line 53: minor nit: Not sure why we need this here. It has the effect of allowing "< >" (which is okay), as well as "< foo, >" (which is also okay, but we don't do that consistently everywhere else).

onnx/defs/parser.cc
```diff
@@ -43,6 +43,35 @@ Status OnnxParser::Parse(char open, IdList& idlist, char close) {
...
+    if (is_assignment) {
+      PARSE(*attrlist.Add());
```
gramalingam, line 55: just wondering: may be the logic could be simplified if we extend the `OnnxParser::Parse(AttributeProto& attr)` method to take the `id` as an optional parameter?

onnx/defs/parser.h
```diff
@@ -385,6 +385,14 @@ class ParserBase {
     return Status::OK();
   }
 
+  bool PeekIsIdentifierAssignment(std::string& id) {
```
gramalingam, line 388: nit: may be `PeekIsAttributeValue` would be a better name.

onnx/test/cpp/parser_test.cc
```diff
@@ -419,6 +419,35 @@ square (x) => (y) {
...
   CheckModel(code);
+
+  const char* code_function_with_attributes = R"ONNX(
+<
+  ir_version: 9,
+  opset_import: [ "" : 15, "custom_domain" : 1]
+>
+agraph (float[N] x) => (float[N] out)
+{
+  out = custom_domain.foo<gamma=2.0, gamma=3.0>(x)
```
gramalingam, line 430: Why are there 2 values for `gamma`?

gramalingam:
I think that the checker may need to be extended to call the checker on the newly added proto fields

onnx/onnx.in.proto
```diff
@@ -95,13 +95,16 @@ enum Version {
...
   IR_VERSION_2020_5_8 = 0x0000000000000007;
 
-  // IR VERSION 8 published on <TBD>
+  // IR VERSION 8 published on April 23, 2021
   // Introduce TypeProto.SparseTensor
   // Introduce TypeProto.Optional
   // Added a list of FunctionProtos local to the model
   // Deprecated since_version and operator status from FunctionProto
-  IR_VERSION = 0x0000000000000008;
+  IR_VERSION_2021_4_23 = 0x0000000000000008;
```
garymm, line 103: 2022

onnx/onnx.in.proto
```diff
@@ -95,13 +95,16 @@ enum Version {
   // - Implicitly add inference graph into each TrainingInfoProto's algorithm.
   IR_VERSION_2020_5_8 = 0x0000000000000007;
 
-  // IR VERSION 8 published on <TBD>
+  // IR VERSION 8 published on April 23, 2021
```
garymm, line 98: 2022

gramalingam: I believe this is a reference to a previous version (an old TBD that was not updated earlier). But I am not sure what the date should be, April 23, 2021 was not the release date, I guess it was presumably the release date after that?

jcwchen: I think it should be July 30, 2021. This date is the release date of ONNX 1.10 (the ONNX version which bumped IR version to 8).

liqunfu: That makes more sense!

onnx/onnx.in.proto
liqunfu: changed to May 8th, 2020, (I was using the PR date which is less accurate)

onnx/test/cpp/parser_test.cc
liqunfu: my mistake, it should be alpha.

onnx/defs/parser.cc
liqunfu: OnnxParser::Parse(AttributeProto& attr) is called from: (1) OnnxParser::Parse(NodeProto& node); (2) OnnxParser::Parse(FunctionProto& fn). For the first case, id is not used. If we want to merge the 2 calls, we can in the first case call OnnxParser::Parse(IdList& idlist, AttrList& attrlist) and ignore the idList. I feel that makes the code less readable. Maybe I understand it wrong?

gramalingam (approved):

PR #4176, opened 2022-04-30
Extend print utility

gramalingam:
**Description**
Extend the print utility functions to handle some constructs not handled:
* types other then tensor-types
* tensor-proto names
* attribute-references


liqunfu (approved):

onnx/defs/printer.cc
```diff
@@ -39,56 +39,108 @@ std::ostream& operator<<(std::ostream& os, const TypeProto_Tensor& tensortype) {
...
+  else if (type.has_sparse_tensor_type())
+    os << type.sparse_tensor_type();
   return os;
 }
 
 std::ostream& operator<<(std::ostream& os, const TensorProto& tensor) {
   os << PrimitiveTypeNameMap::ToString(tensor.data_type());
   print(os, "[", ",", "]", tensor.dims());
 
-  // TODO: does not yet handle raw_data or FLOAT16 or externally stored data.
```
jcwchen, line 57: Try to understand more context about it: you removed the comment here for FLOAT16, because current ParseData for raw_data does not support FLOAT16?

Issue #4189, opened 2022-05-04
Encoding BFLOAT16 Constant to ONNX Fails 


manbearian:
# Bug Report

### Is the issue related to model conversion?
no

### Describe the bug

BFLOAT16 constants are encoded incorrectly when creating tensor initialization data via ONNX Python support. 

This feature was added in v1.11.0 so you need at least that version to reproduce the problem.

I've been able to trace the problem to the following code in `helper.py`. In `make_tensor` there is this code
```Python
        # floa16/bfloat16 are stored as uint16
        elif (data_type == TensorProto.FLOAT16
                or data_type == TensorProto.BFLOAT16):
            vals = np.array(vals).astype(np.float16).view(dtype=np.uint16).flatten().tolist()
```
Notice that the code is converting the value to `np.float16` before writing it out; this is wrong for BFLOAT16 since it doesn't have the same encoding. The result is that this code will write out data in FLOAT16 format when the type is BFLOAT16.

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*):  N/A
- ONNX version (*e.g. 1.7*):  1.11.0
- Python version: N/A
- GCC/Compiler version (if compiling from source): N/A
- CMake version: N/A
- Protobuf version: N/A
- Visual Studio version (if applicable): N/A

### Reproduction instructions
The following python program reproduces the error when run with ONNX v1.11.0

The program prints the incorrect initialization data and writes out the bad graph to disk (as `test.onnx`)

```Python
import onnx
from onnx import helper, TensorProto

type =  TensorProto.BFLOAT16
dims = [2,2]
input = helper.make_tensor_value_info('input', type, dims, False)

c0 = helper.make_tensor_value_info('c0', type, dims, False)
initializer = helper.make_tensor(c0.name, type, dims, [1.0, 2.0, 3.0, 4.0])

# dims: 2              
# dims: 2              
# data_type: 16        
# int32_data: 15360    | CORRECT_VALUE: 16256 (0x3F80)
# int32_data: 16384    | CORRECT_VALUE: 16384 (0x4000)
# int32_data: 16896    | CORRECT_VALUE: 16448 (0x4040)
# int32_data: 17408    | CORRECT_VALUE: 16512 (0x4080)
# name: "c0"           
print(initializer)

output = helper.make_tensor_value_info('output', type, dims)
add = helper.make_node('Add', [input.name, c0.name], [output.name], 'Add')

[...]

PR #4193, opened 2022-05-10
Fix bf16 Support

manbearian:
**Fix bf16 Support**
Fix bfloat16 support in helper and numpy_helper.py. 

**Motivation and Context**
Without native bfloat16 support in numpy implementing bfloat16 support for ONNX is a bit complicated. Recently support was added by encoding bfloat16 as float16. this is not correct as this is a lossy encoding; it's also a bit confusing as a design and indeed broke some bespoke tools on my end that didn't expect the bfloat16 to be encoded as float16 in the ONNX file.

### Open Questions
I've implemented some changes here that give rudimentary support (not completely broken) but i do have a two questions that i believe need to be answered before committing these changes.
1. what is the proper binary encoding of bfloat16 in ONNX protobuf format (is this documented? should it be?)
2. it appears that "raw" encoding and normal encoding for 16-bit values are of different sizes, is this intended?

### Not implemented Changes
1. support for big-endian encodings is not implemented in my code
2. support for "correct" truncation of NaN payloads is not implemented when converting from f32->bf16 in my code

Addresses #4189

### DIFF
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -14,6 +14,10 @@ from onnx.mapping import STORAGE_TENSOR_TYPE_TO_FIELD
 from typing import Text, Sequence, Any, Optional, Dict, Union, TypeVar, Callable, Tuple, List, cast
 import numpy as np  # type: ignore
 
+from cmath import isnan
+import struct
+import sys
+
 VersionRowType = Union[Tuple[Text, int, int, int], Tuple[Text, int, int, int, int]]
 VersionTableType = List[VersionRowType]
 AssignmentBindingType = List[Tuple[Text, Text]]
@@ -278,6 +282,24 @@ def split_complex_to_pairs(ca: Sequence[np.complex64]) -> Sequence[int]:
             for i in range(len(ca) * 2)]
 
 
+# convert a f32 to bf16 (as int)
+def float32_to_bfloat16(fval: float) -> int:
+    ival = int.from_bytes(struct.pack('<f', fval), 'little')
+    if isnan(fval):
+        # NaN requires at least 1 significand bit set
+        ival16 = 0x7FC0  # sign=0, exp=all-ones, sig=0b1000000
+    else:
+        # drop bottom 16-bits
+        # round remaining bits using round-to-nearest-even
+        round = ((ival >> 16) & 1) + 0x7fff
+        ival16 = (ival + round) >> 16
+    # swap byte order for big-endian
+    if sys.byteorder == 'big':
+        bytes = struct.pack('<h', ival16)
+        ival16 = int.from_bytes(bytes, 'big')
+    return ival16
+
+
 def make_tensor(
         name: Text,
         data_type: int,
@@ -328,10 +350,11 @@ def make_tensor(
         if (data_type == TensorProto.COMPLEX64
                 or data_type == TensorProto.COMPLEX128):
             vals = split_complex_to_pairs(vals)
-        # floa16/bfloat16 are stored as uint16
-        elif (data_type == TensorProto.FLOAT16
-                or data_type == TensorProto.BFLOAT16):
+        # float16/bfloat16 are stored as uint16
+        elif data_type == TensorProto.FLOAT16:
             vals = np.array(vals).astype(np.float16).view(dtype=np.uint16).flatten().tolist()
+        elif data_type == TensorProto.BFLOAT16:
+            vals = list(map(float32_to_bfloat16, np.array(vals).astype(np.float32).flatten().tolist()))
         field = mapping.STORAGE_TENSOR_TYPE_TO_FIELD[
             mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE[data_type]]
         getattr(tensor, field).extend(vals)
--- a/onnx/numpy_helper.py
+++ b/onnx/numpy_helper.py
@@ -3,6 +3,7 @@
 import sys
 
 import numpy as np  # type: ignore
+import numpy.typing as nptyping  # type: ignore
 from onnx import TensorProto, MapProto, SequenceProto, OptionalProto
 from onnx import mapping, helper
 from onnx.external_data_helper import load_external_data_for_tensor, uses_external_data
@@ -11,6 +12,12 @@ from typing import Sequence, Any, Optional, Text, List, Dict
 
 def combine_pairs_to_complex(fa: Sequence[int]) -> Sequence[np.complex64]:
     return [complex(fa[i * 2], fa[i * 2 + 1]) for i in range(len(fa) // 2)]
+
+
+# convert ndarray of bf16 (as uint32) to f32 (as uint32)
+def bfloat16_to_float32(data: np.ndarray, dims: nptyping._ShapeLike) -> np.ndarray:
+    shift = lambda x: x << 16  # noqa: E731
+    return shift(data.astype(np.int32)).reshape(dims).view(np.float32)
 
 
 def to_array(tensor: TensorProto, base_dir: Text = "") -> np.ndarray:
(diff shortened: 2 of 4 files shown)


onnx/helper.py
```diff
@@ -277,6 +277,16 @@ def split_complex_to_pairs(ca: Sequence[np.complex64]) -> Sequence[int]:
             for i in range(len(ca) * 2)]
 
 
+# convert a f32 (as uint32) to bf16 (as uint32)
+# TODO: big-endian
```
gramalingam, line 281: FWIW: there is some code that does a cast here:
https://github.com/onnx/onnx/blob/5b1346e30d66e4ec550f6b63c3883b258a2e8e3e/onnx/backend/test/case/node/cast.py#L37
May be worth unifying into some common code (especially if there is some minor difference).

manbearian: that code is wrong; it's not rounding properly when converting from F32 to BF16. Particularly this:
` np_bfp16 = np_uint16_view[1::2] if little_endisan else np_uint16_view[0::2]`

It looks like this is all tests code though, so its fine for testing as long as the test values are crafted in such a way as to not need rounding.

gramalingam: yes, I saw the difference. That's why I was suggesting unifying both to use a common *correct* implementation.

manbearian:
How can we move forward on this? I have some time this week to continue to hack on this if there's more folks would like to see.

gramalingam:
It would be great if we could make this work for both big-endian and small-endian. Is that hard? Is there some reference for the transformation you are using? I am trying to understand why trunction of 32-bit to 16-bit is incorrect.

onnx/helper.py
```diff
@@ -277,6 +277,16 @@ def split_complex_to_pairs(ca: Sequence[np.complex64]) -> Sequence[int]:
...
+# TODO: check for NaN before rounding to 'preserve' NaN payloads
+# algorithm:
+#  truncate bottom 16-bits, round remaining bits using round-to-nearest-even
```
souptc, line 284: why we want this special handle for trancate? from bfloat16 definition in wiki, it seems just truncate the 16 bits from float32.
https://en.wikipedia.org/wiki/Bfloat16_floating-point_format

manbearian: When converting from FLOAT32 to BF16 truncation can work, since the encodings line up save for the 16 lest significant bits, but that means rounding errors. I don't want to get into the exact math, instead let me use a metaphor. let's say i want to convert a 3-digit number to a 2-digit number. Further, let's say the 3-digit number has the value `1.49`. if i convert via truncation i end up with `1.4` if i convert via _rounding_ i end up with `1.5`. The `1.5` value is closer to my original value and is "more correct". The rounding scheme i chose to implement is round-to-nearest-even. Which is the most common rounding scheme. Values below 0.5 are rounded down, values above 0.5 are rounded up, and values exactly equal to 0.5 are rounded to the nearest even value.

The comment i added here around NaN relates to what happens when the value isn't a number. For infinity values the lower 16-bits are 0s, so things work fine. but for NaN the lower mantissa may contain a "NaN payload", extra information that that value is carrying that further describes it beyond just NaN. Its probably not correct to round this value (since its not a number) but it also probably isn't correct to just truncate it. If detecting NaN was easy i would have not rounded it, but because i converted to integer value first, NaN is harder to detect.

Here is the code in pytorch that implements this: https://github.com/pytorch/pytorch/blob/e81bfffbe1349c5ba568a36ec911404c6e494ea0/c10/util/BFloat16.h#L51

Notice that pytorch uses round-to-nearest-even for converting from fp32 to bf16 and also it doesn't do rounding of NaN payloads.

manbearian:
> It would be great if we could make this work for both big-endian and small-endian. Is that hard? Is there some reference for the transformation you are using? I am trying to understand why trunction of 32-bit to 16-bit is incorrect.

Endianness makes by brain hurt, but i'll see if i can work through the pain and implement it. :)  (i was kind of hoping someone would say it isn't needed ;)

See my [other reply](https://github.com/onnx/onnx/pull/4193#discussion_r878466099) as to why i chose rounding over truncation for converting from fp32 to bf16.

manbearian:
NaN is now implemented correctly (my previous implementation could potentially convert NaN to infinity)
Big-Endian is now 'supported'... at least i believe i have the right implementation, but i have no way to test.

manbearian:
@gramalingam and @souptc thoughts?

souptc:
i see. i think it is correct. I actually wondering how does our cpu/gpu EP handle it.

manbearian:
I'm struggling with getting the Python typing correct for one of my new functions. Is there a mpy expert here that might be able to lend a hand?

onnx/numpy_helper.py
```diff
@@ -3,6 +3,7 @@
 import sys
 
 import numpy as np  # type: ignore
+import numpy.typing as nptyping
```
jcwchen, line 6:
```suggestion
import numpy.typing as nptyping  # type: ignore
```
Doesn't have NumPy stub for mypy so simply skip it like above.

onnx/numpy_helper.py
```diff
@@ -13,6 +14,12 @@ def combine_pairs_to_complex(fa: Sequence[int]) -> Sequence[np.complex64]:
     return [complex(fa[i * 2], fa[i * 2 + 1]) for i in range(len(fa) // 2)]
 
 
+# convert ndarray of bf16 (as uint32) to f32 (as uint32)
+def bfloat16_to_float32(data: np.ndarray, dims: nptyping._ShapeLike) -> np.ndarray:
+    shift = lambda x: x << 16
```
jcwchen, line 19: https://github.com/python/mypy/issues/9590 Mypy might have some issues with lambda... I am still investigating. Will keep you posted.

jcwchen: I am not sure why mypy will complain this lambda. It seems like a false positive to me and it still exists in the latest mypy. You can just simply use `# type: ignore` to ignore it for now. Or, shift the variable without using lambda.

onnx/test/helper_test.py
```diff
@@ -425,20 +440,38 @@ def test_make_bfloat16_tensor(self) -> None:
...
+             struct.unpack('!f', bytes.fromhex('40800000'))[0]],  # 4.0
+            [struct.unpack('!f', bytes.fromhex('3DCC0000'))[0],   # truncated
+             struct.unpack('!f', bytes.fromhex('3DCB0000'))[0]],  # truncated
+            [struct.unpack('!f', bytes.fromhex('3DCC0000'))[0],   # truncated
+             struct.unpack('!f', bytes.fromhex('3DCC0000'))[0]],  # truncated
+            [struct.unpack('!f', bytes.fromhex('7FC00000'))[0],   # NaN
+             struct.unpack('!f', bytes.fromhex('7F800000'))[0]],  # inf
+        ])
+
+        truncate = lambda x: x >> 16
```
jcwchen, line 462: ditto

gramalingam:
Nit: if we decide that rounding is the right behavior instead of truncating, it would be good to fix the "Cast" implementation in the test-case generator also.

gramalingam (approved):

manbearian:
> Nit: if we decide that rounding is the right behavior instead of truncating, it would be good to fix the "Cast" implementation in the test-case generator also.

i'm looking at this code now, and suspect i see how it can be changed, but i'm not sure how it gets used. Are there any pointers to how it works?

manbearian:
> i see. i think it is correct. I actually wondering how does our cpu/gpu EP handle it.

do you have a pointer to where the code might be, i'm happy to take a look.

jcwchen:
> > Nit: if we decide that rounding is the right behavior instead of truncating, it would be good to fix the "Cast" implementation in the test-case generator also.
> 
> @gramalingam i'm looking at this code now, and suspect i see how it can be changed, but i'm not sure how it gets used. Are there any pointers to how it works?

https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/cast.py This file is used to create the node test model (For instance, https://github.com/onnx/onnx/tree/main/onnx/backend/test/data/node/test_cast_FLOAT_to_BFLOAT16). You can update the code with correct behavior in cast.py first and then use [tools/update_doc.sh](https://github.com/onnx/onnx/blob/main/tools/update_doc.sh) to regenerate related node test models.

jcwchen:
> > i see. i think it is correct. I actually wondering how does our cpu/gpu EP handle it.
> 
> do you have a pointer to where the code might be, i'm happy to take a look.

Probably [here](https://github.com/microsoft/onnxruntime/blob/9707181257391323dfe7170d506650ded04dc59c/include/onnxruntime/core/framework/float16.h#L57). However, different EPs seem to have different behavior. We are trying to reach out the code owner to understand the inconsistency and which behavior is correct.

manbearian:
@jwchen and @gramalingam i updated cast.py/castlike.py to use rounding for f32->bf16. Let me know if this is what you were looking for.

I was expecting to see some tests fail, but i believe the tests aren't checking for exact values, so the single bit difference isn't detectable.

gramalingam:
I suspect that the checkin for test-data for ops other than cast / cast-like are spurious, I don't think they should change?

jcwchen:
Thank you @manbearian for the quick update! As @gramalingam mentioned, please remove other irrelevant operators' update (like updating output.pb due to different numpy.random behaviors by different machines). tools/update_doc.sh will update every node test data and we should only take Cast/CastLike related tests.

> I was expecting to see some tests fail, but i believe the tests aren't checking for exact values, so the single bit difference isn't detectable.

You are right -- these models/input.pb/output.pb just exactly follow what onnx/backend/test/case/node/[operator_name].py defines. The CI only checks that whether the uploaded model can be reproduced by CI environments.

manbearian:
after talking with @gramalingam  offline we decided to drop the CAST/CASTLIKE changes for now. I have them stored in a branch if we want to bring them back. There was also a suggestion that the f32->bf16 helper support both rounding and truncation modes. I can add support for that with a future PR if there is interest (please open an issue and assign it to me).

PR #4194, opened 2022-05-10
Handle constant op variants (#3985)

### LINKED ISSUE #3985: Handle Constant op variants in shape inference
# Bug Report

### Describe the bug

Shape-inference does not handle variants of Constant op introduced in opset 12. The opset 12 version introduced the ability to specify constant values using one of multiple attributes (like `value_ints`, for example) but the implementation handles only the original form (using the attribute `value`).

### Reproduction instructions
```
from onnx import shape_inference, parser

def infer(x):
    model = parser.parse_model(x)
    inferred_model = shape_inference.infer_shapes(model)
    return { vi.name : vi.type for vi in inferred_model.graph.value_info }

result = infer('''
<
  ir_version: 7,
  opset_import: [ "" : 15 ]
>
toy (int64[?] X) => (int64[] Y)
{
    S = Constant <value_ints = [10, 20]>()
    Z = Reshape (X, S)
    Y = Identity (Z)
}
''')

print(result['Z'])

result2 = infer('''
<
  ir_version: 7,
  opset_import: [ "" : 15 ]
>
toy (int64[?] X) => (int64[] Y)
{
    S = Constant <value = int64[2] {10, 20}>()
    Z = Reshape (X, S)
    Y = Identity (Z)
}
''')

print(result2['Z'])
```

### Expected behavior
Both `result` and `result2` should have same value

### Notes

This line needs to be fixed: https://github.com/onnx/onnx/blob/f0c4ee2b46b3d00c005d1ecd9440cfc151f4fd1d/onnx/shape_inference/implementation.cc#L341

avinashsai:
Fixes #3985

### DIFF
--- a/onnx/shape_inference/implementation.cc
+++ b/onnx/shape_inference/implementation.cc
@@ -287,7 +287,7 @@ class ShapeInferenceImplBase {
       has_experimental_op = true;
     } else if (n.op_type() == "Constant" && n.output().size() == 1) {
       for (const auto& attr : n.attribute()) {
-        if (attr.name() == "value") {
+        if (attr.name() == "value" || attr.name() == "value_ints") {
           if (attr.type() == AttributeProto::TENSOR && attr.has_t()) {
             input_data_by_name[n.output(0)] = &attr.t();
           } else if (attr.type() == AttributeProto::SPARSE_TENSOR && attr.has_sparse_tensor()) {


onnx/shape_inference/implementation.cc
```diff
@@ -287,7 +287,7 @@ class ShapeInferenceImplBase {
       has_experimental_op = true;
     } else if (n.op_type() == "Constant" && n.output().size() == 1) {
       for (const auto& attr : n.attribute()) {
-        if (attr.name() == "value") {
+        if (attr.name() == "value" || attr.name() == "value_ints") {
```
gramalingam, line 290: Thanks for creating the PR. Unfortunately, I don't think this will be enough to fix the issue. When the name is "value_ints", the attribute type is a list of ints, and not a tensor. So, the lines below also need to be extended to handle it correctly.

avinashsai: Thanks for the suggestions! If i am correct, I need to add similar else if block which handles list of ints right?

gramalingam: Yes. The proto fields are as defined here: https://github.com/onnx/onnx/blob/5b1346e30d66e4ec550f6b63c3883b258a2e8e3e/onnx/onnx.in.proto#L168
In addition, there is a need to convert the value into a TensorProto.

gramalingam: Here is how onnxruntime does it: we need something similar: https://github.com/microsoft/onnxruntime/blob/4c37f15c1ba3e24e0c7e2110f34ab410fadb2340/onnxruntime/core/framework/tensorprotoutils.cc#L821

gramalingam:
This has been completed by a different PR.

PR #4202, opened 2022-05-17
Enable test_driver.cc on Windows and test it in Windows CI

### LINKED ISSUE #4200: onnxifi_test_driver_gtests onnx/backend/test/data/node is missing on Windows
# Bug Report

### Describe the bug
onnxifi_test_driver_gtests onnx/backend/test/data/node is missing on Windows. Currently ONNX only tests it in Linux-CI and MacOS-CI. We should test it to validate input.pb and output.pb under onnx/backend/test/data/node in Windows-CI as well

jcwchen:
**Description**
Currently ONNX only tests test_driver_gtests on Linux and Mac. We should enable the test on Windows as well
- Enable test_driver.cc on Windows 
- Add test command in Windows-CI

**Motivation and Context**
Closes #4200.


onnx/backend/test/cpp/driver/test_driver.cc
```diff
@@ -206,6 +198,11 @@ ResolvedTestCase LoadSingleTestCase(const UnsolvedTestCase& t) {
       std::string input_data;
       ONNX_NAMESPACE::ValueInfoProto input_info;
       LoadSingleFile(input_file, input_data);
+      if (test_data_counter >= st.model_.graph().input().size()) {
```
liqunfu, line 201: nip: shall we take into consideration on feed name and check all model's required inputs get mapped to from the feed?

liqunfu (approved):

onnx/backend/test/cpp/driver/test_driver.cc
jcwchen: I don't quite understand -- what kind of feed name you were saying? If you were saying input/output names, then ONNX uses naming like `input_X.pb` and `output_X.pb` for input/output files. If added .pb files do not follow this naming, other CIs will catch this kind of error (during test generation). For here, I believe it will simply fail due to unfound data. I added the size check here, because previously wrong test data size will lead to segmentation fault (access out of range), which is not acceptable.

jcwchen (commented):
@onnx/sig-operators-approvers Could you please review/approve this PR? Thank you for your time!

jcwchen:
Close it since the files do not exist after the deprecation of ONNXIFI.

PR #4203, opened 2022-05-18
Specify default value for score_threshold attribute of NMS

rajeevsrao:
**Description**
- NonMaxSuppression operator specification does not specify a default value for the `score_threshold` attribute.

**Motivation and Context**
- Default values must be specified for optional attributes in the Operator specifications.

Signed-off-by: Rajeev Rao <rajeevrao@nvidia.com>

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -13127,9 +13127,9 @@ Other versions of this operator: <a href="Changelog.md#NonMaxSuppression-10">10<
 <dt><tt>max_output_boxes_per_class</tt> (optional) : tensor(int64)</dt>
 <dd>Integer representing the maximum number of boxes to be selected per batch per class. It is a scalar. Default to 0, which means no output.</dd>
 <dt><tt>iou_threshold</tt> (optional) : tensor(float)</dt>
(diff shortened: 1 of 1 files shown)


rajeevsrao:
@jcwchen can we include this for 1.12? Trivial change.

jcwchen (commented):
Hi @rajeevsrao,
Is it just a document update for Operators.md without changing shape inference behavior? If yes, perhaps we don't really need to include this into 1.12? I think typically users check Operators.md from the main branch.

BTW, to fix CI failure, instead of updating Operators.md directly, please update the string [here](https://github.com/onnx/onnx/blob/f6f2f588fec9701ae4813aad921ef7fcc015cd19/onnx/defs/object_detection/defs.cc#L162) and [here](https://github.com/onnx/onnx/blob/f6f2f588fec9701ae4813aad921ef7fcc015cd19/onnx/defs/object_detection/old.cc#L154), and then use `python onnx/defs/gen_doc.py` to regenerate Operators.md and Changelog.md. Thanks!

cc @onnx/sig-operators-approvers to review this PR.

PR #4210, opened 2022-05-20
Upgrade python syntax with pyupgrade

justinchuby:
**Description**

Upgrade python syntax with pyupgrade https://github.com/asottile/pyupgrade

```
pyupgrade --py36-plus --keep-runtime-typing **/*.py
```

**Motivation and Context**

Existing code contains obsolete python syntax. Using the tool to modernize the code base.

PR #4212, opened 2022-05-20
Upgrade python syntax with pyupgrade

justinchuby:
**Description**

Upgrade python syntax with pyupgrade https://github.com/asottile/pyupgrade

```
pyupgrade --py36-plus --keep-runtime-typing **/*.py
```

Replaces #4210

**Motivation and Context**

Existing code contains obsolete python syntax. Using the tool to modernize the code base.


justinchuby:
@jcwchen I think this is ready for review. Could you help trigger the checks?

onnx/__init__.py
```diff
@@ -18,31 +18,31 @@
...
 
 # str should be bytes,
 # f should be either writable or a file path
-def _save_bytes(str: bytes, f: Union[IO[bytes], Text]) -> None:
+def _save_bytes(str: bytes, f: Union[IO[bytes], str]) -> None:
     if hasattr(f, 'write') and callable(cast(IO[bytes], f).write):
         cast(IO[bytes], f).write(str)
     else:
-        with open(cast(Text, f), 'wb') as writable:
+        with open(cast(str, f), 'wb') as writable:
```
jcwchen, line 40: Linux-CI failed because mypy typecheck error here: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=24804&view=logs&j=26ad7b42-9ea7-5ad4-891d-e581c481314e&t=36dc6483-53ad-5f31-0c8e-c9104543cd82

justinchuby: Looks like the function param is redefining `str`. I can try changing it. Is there a fix you would recommend?

jcwchen: I don't have a strong preference about the solution as long as mypy typecheck is happy. It's interesting that only L40 complains this error. (I saw similar a change at L29). Actually `with open(cast(str, f), 'wb') as writable:` looks fine to me so it is possible that it is a false alarm? Or, according to the error message, you might try to remove the cast since it says it is unnecessary. (although I think it should be needed)

justinchuby: Renamed the first param to `content`

jcwchen (approved):
LGTM. Thank you for the quick update!

jcwchen:
It still needs approval from onnx/sig-operators. Could anyone in @onnx/sig-operators-approvers please review this PR? Thanks!

gramalingam (approved):

PR #4219, opened 2022-05-24
Use ONNX Runtime PyPI package instead of ort-nightly

jcwchen:
**Description**
Use ONNX Runtime PyPI package instead of ort-nightly in all release CIs to fix current CI failure because of missing ort-nightly packages.

**Motivation and Context**
Previously ONNX repo uses ort-nightly TestPyPI packages to run a simple test. However, https://github.com/microsoft/onnxruntime/issues/11613#issuecomment-1136292597 actually ort-nightly on TestPyPI is not used for external testing. Using ONNX Runtime PyPI package to validate the combined usage of ONNX and ORT should be sufficient.


gramalingam (approved):

linkerzhang (approved):

snnn:
I setup a new feed to replace the testpypi one: https://aiinfra.visualstudio.com/PublicPackages/_artifacts/feed/ORT-Nightly/connect/pip 
The new one is hosted on ADO. It is fully under ONNX Runtime's control. It doesn't have storage limit. So I do not need to delete old packages every week.

jcwchen:
Thank @snnn for the effort and update! That's great to know and I will keep in mind for future use case.

PR #4234, opened 2022-05-31
Fix mapping TensorProto to NumPy for bfloat16

jcwchen:
**Description**
- The map in TENSOR_TYPE_TO_NP_TYPE is the "used" NumPy type instead of "stored" NumPy type. 
- Fix a mapping bug in make_tensor for raw. It should use storage type instead.
- Add comments for confusing maps: TENSOR_TYPE_TO_NP_TYPE and TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE

(*Updated)
There are two similar maps in mapping:
- TENSOR_TYPE_TO_NP_TYPE: which NumPy type you will use to create such a TensorProto. Take float16 as an example, you can directly use numpy.float16 as values for make_tensor.
- TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE: which ONNX Tensor type wou will use to "store" such a TensorProto. Take float16 as an example again, by model proto definition, you should use "TensorProto.UINT16" to store "TensorProto.FLOAT16".

Based on the float16 example, I think the map for bfloat16 should be
- (TENSOR_TYPE_TO_NP_TYPE) TensorProto.FLOAT16: np.dtype('float32')
- (TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE)  TensorProto.BFLOAT16: TensorProto.UINT16

To make a bfloat16 tensor, ONNX should use NumPy float32 values in the first place and then they will be converted to uint16 later by the helper.

**Motivation and Context**
Fix mapping TensorProto to NumPy for bfloat16

### DIFF
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -333,7 +333,8 @@ def make_tensor(
         assert not raw, "Can not use raw_data to store string type"
 
     # Check number of vals specified equals tensor size
-    expected_size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[data_type].itemsize)
+    storage_type = mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE[data_type]
+    expected_size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[storage_type].itemsize)
     # Flatten a numpy array if its rank > 1
     if type(vals) is np.ndarray and len(vals.shape) > 1:
         vals = vals.flatten()
--- a/onnx/mapping.py
+++ b/onnx/mapping.py
@@ -13,7 +13,7 @@ TENSOR_TYPE_TO_NP_TYPE = {
     int(TensorProto.INT64): np.dtype('int64'),
     int(TensorProto.BOOL): np.dtype('bool'),
     int(TensorProto.FLOAT16): np.dtype('float16'),
-    int(TensorProto.BFLOAT16): np.dtype('float32'),  # native Numpy does not support bfloat16
+    int(TensorProto.BFLOAT16): np.dtype('float32'),  # Native numpy does not support bfloat16
     int(TensorProto.DOUBLE): np.dtype('float64'),
     int(TensorProto.COMPLEX64): np.dtype('complex64'),
     int(TensorProto.COMPLEX128): np.dtype('complex128'),
@@ -22,7 +22,7 @@ TENSOR_TYPE_TO_NP_TYPE = {
     int(TensorProto.STRING): np.dtype('object')
 }
 
-# Currently native Numpy does not support bfloat16 so TensorProto.BFLOAT16 is ignored for now
+# Currently native numpy does not support bfloat16 so TensorProto.BFLOAT16 is ignored for now
 # Numpy float32 array is only reversed to TensorProto.FLOAT
 NP_TYPE_TO_TENSOR_TYPE = {v: k for k, v in TENSOR_TYPE_TO_NP_TYPE.items() if k != TensorProto.BFLOAT16}
 
--- a/onnx/mapping.py
+++ b/onnx/mapping.py
@@ -13,7 +13,7 @@ TENSOR_TYPE_TO_NP_TYPE = {
     int(TensorProto.INT64): np.dtype('int64'),
     int(TensorProto.BOOL): np.dtype('bool'),
     int(TensorProto.FLOAT16): np.dtype('float16'),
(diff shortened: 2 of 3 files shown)


jcwchen:
Sorry it looks confusing and I was about to update the PR description later -- There are two similar maps in mapping:
- TENSOR_TYPE_TO_NP_TYPE: which NumPy type you will use to create such a TensorProto. Take float16 as an example, you can directly use numpy.float16 as values for make_tensor.
- TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE: which ONNX Tensor type wou will use to "store" such a TensorProto. Take float16 as an example again, by model proto definition, you should use "TensorProto.UINT16" to store "TensorProto.FLOAT16".

Based on the float16 example, I think the map for bfloat16 should be
- (TENSOR_TYPE_TO_NP_TYPE) TensorProto.FLOAT16: np.dtype('float32')
- (TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE)  TensorProto.BFLOAT16: TensorProto.UINT16

To make a bfloat16 tensor, ONNX should use NumPy float32 values in the first place and then they will be converted to uint16 later by the helper. Please correct me if I am wrong. Thanks!

gramalingam:
Thanks @jcwchen for the explanation. It would be great if you could add this documentation to the mapping.py file. A quick search indicates it is used perhaps for converting TensorProto to numpy array (and it is used in the other direction only for `raw` data?)

onnx/helper.py
```diff
@@ -333,7 +333,8 @@ def make_tensor(
         assert not raw, "Can not use raw_data to store string type"
 
     # Check number of vals specified equals tensor size
-    expected_size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[data_type].itemsize)
+    storage_type = mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE[data_type]
+    expected_size = 1 if (not raw) else (mapping.TENSOR_TYPE_TO_NP_TYPE[storage_type].itemsize)
```
jcwchen, line 337: The given raw type here should be the "storage type". For instance:
- float16: TensorProto.UINT16
- bfloat16: TensorProto.UINT16

This bug was uncaught by float16 because `np.dtype('float16').itemsize == np.dtype('uint16').itemsize == 2`.

onnx/mapping.py
```diff
@@ -23,9 +24,10 @@
 }
 
 # Currently native numpy does not support bfloat16 so TensorProto.BFLOAT16 is ignored for now
-# Numpy float16 array is only reversed to TensorProto.FLOAT16
+# Numpy float32 array is only reversed to TensorProto.FLOAT
 NP_TYPE_TO_TENSOR_TYPE = {v: k for k, v in TENSOR_TYPE_TO_NP_TYPE.items() if k != TensorProto.BFLOAT16}
 
+# This map refers to which TensorPrto type you should store for certain TensorPrto type
```
gramalingam, line 30: How about
```
This map indicates what storage-type is used in the protobuf (serialized) representation for TensorProto.
```

jcwchen:
> 
> No problem, i'm happy to help!

Thank you :)
 
> The changes you have here make total sense to me in the abstract. I'm missing some context i think though. What scenario is broken because the mapping from tensorproto->numpy is wrong? Do we have a test case?

Actually it does not break anything internally previously, because TENSOR_TYPE_TO_NP_TYPE is mainly used for make_tensor with raw, but there is a bug [there](https://github.com/onnx/onnx/pull/4234/files#r887138570) After fixing that bug, the map (TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE) should be corrected for the consistent meaning: it is used for converting TensorProto values into Numpy arrays. Externally ONNX users might inspect this map to understand what NumPy type they should use for `vals` while making a tensor.

Currently there is no specific test to verify this map. The test for make_tensor with raw kind of covers partial use case because it relies on `TENSOR_TYPE_TO_NP_TYPE`. To make more tests cover this map, perhaps we can update `make_tensor` as:
```
np_dtype = mapping.TENSOR_TYPE_TO_NP_TYPE[data_type]
# float16/bfloat16 are stored as uint16
elif data_type == TensorProto.FLOAT16:
    vals = np.array(vals).astype(np_dtype).view(dtype=np.uint16).flatten().tolist()
elif data_type == TensorProto.BFLOAT16:
    vals = list(map(float32_to_bfloat16, np.array(vals).astype(np_dtype).flatten().tolist()))
```

gramalingam (approved):

PR #4235, opened 2022-06-01
Update ir.h to support initializer better

daquexian:
**Description**
I'm the maintainer of [onnx optimizer](https://github.com/onnx/optimizer).

https://github.com/onnx/onnx/pull/3676 supports non-input initializers in IR officially.  In order to make onnx optimizer (and maybe also version converter) happier with this great improvement, I made the following changes:

1. `eraseInitializerAndInput` only erases input when the initializer is actually an input
2. `value.setUniqueName` also updates `initializer_names_` etc.
3. When converting GraphProto to Graph, add initializers before other nodes to avoid the conflict of unique names.
4. `initializer_to_offset_map_` is removed. The reason is described in https://github.com/onnx/onnx/pull/4235/files#r887535784

**Motivation and Context**
I'm going to update onnx optimizer and support non-input initializers better


onnx/common/ir.h
```diff
@@ -972,9 +969,11 @@ struct Graph final {
         initializers_.end());
     initializer_names_.erase(
         std::remove(initializer_names_.begin(), initializer_names_.end(), name), initializer_names_.end());
-    if (initializer_to_offset_map_.count(name) > 0) {
-      initializer_node_->eraseOutput(initializer_to_offset_map_[name]);
-      initializer_to_offset_map_.erase(name);
```
daquexian, line 977: after erasing output of `initializer_node_`, the `initializer_to_offset_map_` should be updated accordingly. However, it requires a full traverse to update every element in the map, which makes maintaining a map even worse than a naive search-and-erase.

onnx/common/ir.h
```diff
@@ -1248,17 +1249,27 @@ inline const Graph* Value::owningGraph() const {
...
+  if (has_unique_name() && update_related_names) {
+    auto old_name = unique_name_;
+    for (size_t i = 0; i < owningGraph()->initializer_names_.size(); i++) {
+      auto &initializer_name = owningGraph()->initializer_names_[i];
+      if (initializer_name == old_name) {
+        initializer_name = name;
+        owningGraph()->initializers_[i].setName(name);
+      }
+    }
+    auto *graph = owningGraph();
```
jcwchen, line 1264: tiny nit: could we declare `auto *graph = owningGraph();` in advance before L1257 for later use (L1257-L1261)? Or perhaps directly use `owningGraph()` all the time without having a variable.

jcwchen (approved):
Thank you for the improvement! Overall looks good to me. Please fix clang-format warning about the pointer (See details [here](https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=25342&view=logs&j=26ad7b42-9ea7-5ad4-891d-e581c481314e&t=36dc6483-53ad-5f31-0c8e-c9104543cd82)). It's great to see the previous IR fix makes onnx optimizer happier!

One question regarding the 3rd point:
> When converting GraphProto to Graph, add initializers before other nodes to avoid the conflict of unique names.

I am curious that what circumstance will cause this name conflict? Can't current setUnique design help to avoid this situation?

onnx/common/ir.h
daquexian: sorry for the late response~ updated

daquexian:
> One question regarding the 3rd point:
> 
> When converting GraphProto to Graph, add initializers before other nodes to avoid the conflict of unique names.
> 
> I am curious that what circumstance will cause this name conflict? Can't current setUnique design help to avoid this situation?

Thanks for your review! `getNextUnique` will find if the candidate already exists in `initializer_names_`, so it works well when the initializers are already added before any normal ir nodes, but fails otherwise: a normal ir node may get its "unique" name by `getNextUnique`, and then an initializer with the same name is added.

daquexian:
@jcwchen is this pr ready to merge? Thanks!

Issue #4237, opened 2022-06-01
Support different modes for float to bfloat conversion


gramalingam:
# Feature Request

As discussed in the recent PR https://github.com/onnx/onnx/pull/4193 , it would be good to extend the utility function for converting float32 to bfloat16 to support different modes (e.g., truncation vs rounding; the treatment of NaN could potentially lead to distinguishing pure-truncation vs. truncation with correct NaN handling).

jcwchen:
I guess this has been handled by Ian's fresh PR: https://github.com/onnx/onnx/pull/4238? If yes, feel free to close it. Thanks!

PR #4238, opened 2022-06-01
Add Truncation Mode to F32->BF16 Conversion Helper (and fix big-endian support)

manbearian:
I'm adding support to do truncation instead of rounding to the float32_to_bfloat16 conversion helper. This new functionality is currently unused.

I've also included a fix for big-endian support

Implements #4237


manbearian:
@gramalingam is there a way i can test big-endian support? I'm not very confident that my implementation is correct.

onnx/helper.py
```diff
@@ -282,22 +282,23 @@ def split_complex_to_pairs(ca: Sequence[np.complex64]) -> Sequence[int]:
...
-    # swap byte order for big-endian
-    if sys.byteorder == 'big':
-        bytes = struct.pack('<h', ival16)
-        ival16 = int.from_bytes(bytes, 'big')
-    return ival16
+        return 0x7FC0  # sign=0, exp=all-ones, sig=0b1000000
+    # drop bottom 16-bits
+    # round remaining bits using round-to-nearest-even
+    round = ((ival >> 16) & 1) + 0x7fff
+    return (ival + round) >> 16
```
gramalingam, line 301: I assume that the endian adjustment was removed intentionally (and that it is not required)?

manbearian:
> @gramalingam is there a way i can test big-endian support? I'm not very confident that my implementation is correct.

@gramalingam (with the help of one of my amazing coworkers) i was able to get an emulation of s390x up and running to verify big-endian support. It turns out that my current implementation was broken and i've added a fix for that to the code.

This fix is to basically disable any special casing for big-endian. I'm not going to say i understand why i don't need to swap the byte order, but i verified that the test passes, a generated ONNX file with BF16 values encodes correctly (same as on my x64 machine), and manual inspection of the in-memory buffer looks correct.

I was able to get the full set of tests running on my emulator and saw this:

```
=============================================== short test summary info ================================================
FAILED onnx/test/helper_test.py::TestHelperTensorFunctions::test_make_bfloat16_tensor_with_raw - AssertionError:
FAILED onnx/test/helper_test.py::TestHelperTensorFunctions::test_make_float16_tensor_with_raw - AssertionError:
FAILED onnx/test/helper_test.py::TestHelperTensorFunctions::test_make_tensor - AssertionError:
FAILED onnx/test/test_external_data.py::TestExternalDataToArray::test_save_model_with_external_data_multiple_times - ...
FAILED onnx/test/test_external_data.py::TestExternalDataToArray::test_to_array_with_external_data - AssertionError: F...
FAILED onnx/test/version_converter_test.py::TestVersionConverter::test_upsample_with_raw_constant_node_9_8 - onnx.onn...
FAILED onnx/test/version_converter_test.py::TestVersionConverter::test_upsample_with_raw_initializer_9_8 - onnx.onnx_...
FAILED onnx/examples/np_array_tensorproto.ipynb::Cell 4
======================== 8 failed, 1904 passed, 1084 skipped, 9 warnings in 1501.56s (0:25:01) =========================
```

i looked into some of these failures, and i have a suspicion that the big-endian support in `numpy_helper.py` is not correct and that there are likely 1 or 2 more lurking issues.

onnx/helper.py
gramalingam: I suspect no special-casing is required for endian because you are using "integer arithmetic operators" (and implementing them correctly is the compiler's responsibility, regardless of endian-ness). I mean, the operators used have a fixed semantics, regardless of the endian-ness.

manbearian: > Out of curiosity: I don't follow the `(ival >> 16) & 1` bit ... we end up adding either 0x7fff or 0x8000 (depending on the 16th bit) ... which seems slightly odd.

this is the "even" part of the "round-to-nearest-even" rounding. In this rounding mode if the fractional part being dropped is half-way between the two potential target values we round toward the nearest even value.

Specifically, we partition ival into two 16-bit parts [A][B]. If B is less than 0x8000 we round A down. if B is greater than 0x8000 we round A up. if B is exactly equal to 0x8000 we round A up or down based on if A is odd or even.


if A is odd `(ival >> 16) & 1` is 1, so `round` will be `0x8000`
if A is even `(ival >> 16) & 1` is 0, so `round` will be `0x7FFFF`

This lines up with what you noticed above, that round is always one of these two values.

Now the key is that we add this to the full 32-bit number.

If B is < 0x8000 and we add either of these values, the result will not overflow into A, so the addition is a no-op. When we drop B, we have effectively "rounded down"

if B is > 08x000 and we add either of these values, the result will *always* oveerflow a single bit into A. When we drop B, we have effectively "rounded up"

The magic is if B is exactly equal to 0x8000. 
- if we add 0x7FFF to this value, we do *not* overflow, the effect is round-down.
- if we add 0x8000 to this value, we do overflow, the effect is round-up.

So, if A is even, we round down (to the nearest even) and if A is odd, we round up (to the nearest even).

A useful analogy, that might be more familiar is rounding values in C when converting from FP to INT. In C this conversion "truncates" by dropping the fractional portion effectively implementing "round-toward-zero". However, if you add 0.5 to the value before converting, you get a "round-to-nearest" rounding. The difference between that technique and what i've implemented here is that we don't want "round-to-nearest" we want "round-to-nearest-even" so we need an extra step to force rounding down when the value is already even.


I hope this explanation makes sense!

manbearian: > I suspect no special-casing is required for endian because you are using "integer arithmetic operators" (and implementing them correctly is the compiler's responsibility, regardless of endian-ness). I mean, the operators used have a fixed semantics, regardless of the endian-ness.

I suspect also that internally python is strong its values in little-endian format regardless of the underlying machine, and that was confusing me. In otherwards i was overcomplicating things by conflating what i would need to do in C vs. what the semantics were of what i wanted. High-level languages FTW win here i suppose.

manbearian:
> i looked into some of these failures, and i have a suspicion that the big-endian support in numpy_helper.py is not correct for raw encodings. I also suspect there are likely 1 or 2 more lurking issues as not everything looked related to raw encodings.

@gramalingam please let me know if you'd like me to open an issue on this. i don't know if you folks are tracking issues on big-endian platforms and/or would prefer to see a customer reported case before opening an issue.

onnx/helper.py
gramalingam: Thanks for the detailed explanation about the rounding.

gramalingam (approved):

gramalingam:
> > i looked into some of these failures, and i have a suspicion that the big-endian support in numpy_helper.py is not correct for raw encodings. I also suspect there are likely 1 or 2 more lurking issues as not everything looked related to raw encodings.
> 
> @gramalingam please let me know if you'd like me to open an issue on this. i don't know if you folks are tracking issues on big-endian platforms and/or would prefer to see a customer reported case before opening an issue.

Yes, that would be helpful, thanks! We should fix/resolve these.

manbearian:
> > > i looked into some of these failures, and i have a suspicion that the big-endian support in numpy_helper.py is not correct for raw encodings. I also suspect there are likely 1 or 2 more lurking issues as not everything looked related to raw encodings.
> > 
> > 
> > @gramalingam please let me know if you'd like me to open an issue on this. i don't know if you folks are tracking issues on big-endian platforms and/or would prefer to see a customer reported case before opening an issue.
> 
> Yes, that would be helpful, thanks! We should fix/resolve these.

Opened https://github.com/onnx/onnx/issues/4253

gramalingam (approved):

gramalingam:
> @gramalingam do you need anything else from me to complete the PR?

No, thanks very much for the explanation and the fix.

garymm:
@etiotto please cherry pick

Issue #4231, opened 2022-05-30
How to dump ONNX model / graph to textual syntax?


Lyken17:
In the https://github.com/onnx/onnx/blob/main/docs/PythonAPIOverview.md#onnx-parser section, there is an example showing that we can convert a textual representation (in ONNX syntax) to ONNX model / graph

```
input = '''
   agraph (float[N, 128] X, float[128, 10] W, float[10] B) => (float[N, 10] C)
   {
        T = MatMul(X, W)
        S = Add(T, B)
        C = Softmax(S)
   }
'''
graph = onnx.parser.parse_graph(input)
```

However, I tried to search docs and test cases and didn't find a way to perform the conversion vice versa. For example, is there any api can achieve following operations?

```
graph.dump_to_text() #? Just a sample API, will not work
'''
   agraph (float[N, 128] X, float[128, 10] W, float[10] B) => (float[N, 10] C)
   {
        T = MatMul(X, W)
        S = Add(T, B)
        C = Softmax(S)
   }
'''
```

jcwchen:
Hi @Lyken17,
[onnx-parser](https://github.com/onnx/onnx/blob/main/docs/PythonAPIOverview.md#onnx-parser) was created for easier ONNX graph creation (text -> ONNX proto). Reverse conversion is not supported now.

Lyken17:
Thanks @jcwchen 's quick response!

Actually stringfy conversion should be easier parser? I feel this would be a nice feature to have as the stringfied text are much more readable than current onnx json-like output and many other frameworks (e.g., tvm) supports such functions. 

How can I submit a feature request for this utility?

gramalingam:
Just to clarify: the reverse (pretty printing) is supported in C++ (see: https://github.com/onnx/onnx/blob/main/onnx/defs/printer.h ), but it has not yet been exposed in Python. I am planning on creating a PR soon to expose at least the top-level print methods (for model/graph) to Python

Lyken17:
Looking forward to it.  Please let me know once the PR is merged!

PR #4246, opened 2022-06-02
Print utility extension

gramalingam:
**Description**

Extend the printer utilities and expose them in Python.
* Create a class to encapsulate printer
* Add support for indentation for more readable output for sub-graphs
* Add support for printing ModelProto
* Expose top-level functionality to Python

**Motivation and Context**

https://github.com/onnx/onnx/issues/4231

**Limitations**

Need support for external tensors. A textual-representation is not suitable for large tensor constants.


onnx/defs/printer.cc
```diff
@@ -3,108 +3,208 @@
...
+  void print(const MetaDataProps& metadataprops) {
+    print ("[", ", ", "]", metadataprops);
   }
-  os << close;
-}
 
-std::ostream& operator<<(std::ostream& os, const TensorShapeProto_Dimension& dim) {
+  void print(const MetaDataProp& metadata) {
+    printQuoted(metadata.key());
+    output << " : ";
```
xadupre, line 64: `": "`?

onnx/defs/parser.h
```diff
@@ -160,6 +160,15 @@ class KeyWordMap {
     return KeyWord::NONE;
   }
 
+  static const std::string& ToString(KeyWord kw) {
+    static std::string undefined("undefined");
+    for (const auto& pair : Instance()) {
```
xadupre, line 165: Maybe it would be worth to have a dictionary value: key to speed it up? I don't know how many times this is called.

onnx/defs/printer.cc
```diff
@@ -115,145 +215,218 @@ std::ostream& operator<<(std::ostream& os, const TensorProto& tensor) {
...
+  print("(", ", ", ")", vilist);
 }
 
-std::ostream& operator<<(std::ostream& os, const AttributeProto& attr) {
+void ProtoPrinter::print(const AttributeProto& attr) {
   // Special case of attr-ref:
   if (attr.has_ref_attr_name()) {
-    os << attr.name() << " : " << AttributeTypeNameMap::ToString(attr.type()) << " = @" << attr.ref_attr_name();
-    return os;
+    output << attr.name() << " : " << AttributeTypeNameMap::ToString(attr.type()) << " = @" << attr.ref_attr_name();
```
xadupre, line 260: `": "`?

onnx/test/printer_test.py
```diff
@@ -0,0 +1,36 @@
+# SPDX-License-Identifier: Apache-2.0
+import onnx
+import unittest
+from onnx import helper, parser, printer, GraphProto
+
+
+class TestBasicFunctions(unittest.TestCase):
```
xadupre, line 7: We should have more unit tests.

gramalingam: I have converted every parser-test (in C++ unit-tests) into also a printer-test. So, we now have better test coverage.

onnx/defs/printer.cc
gramalingam: Done (in all cases below, also).

onnx/printer.py
```diff
@@ -0,0 +1,15 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import onnx
+import onnx.onnx_cpp2py_export.printer as C
+from typing import Text, Union
+
+
+def to_text(proto):  # type: Union[onnx.ModelProto, onnx.FunctionProto, onnx.GraphProto] -> Text
```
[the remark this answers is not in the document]
gramalingam: I am following the existing conventions. One possible reason for the existing style might be that the protobuf types are generated by the protobuf compiler for both C++ and Python. Not sure how easy it is to automatically convert between these representations using pybind11.

onnx/printer.py
```diff
@@ -0,0 +1,15 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import onnx
+import onnx.onnx_cpp2py_export.printer as C
+from typing import Text, Union
+
+
+def to_text(proto: Union[onnx.ModelProto, onnx.FunctionProto, onnx.GraphProto]) -> Text:
+    if isinstance(proto, onnx.ModelProto):
```
xadupre, line 9: It is better this way. There should be a documentation for this function. I still think the call to SerializeToString() should be avoided.

gramalingam: How do we avoid the serialization? (Of course, we can reimplement the pretty-printer in Python, but I assume that's not what you mean? That would lead to a potential divergence of behavior between C++ and Python. Since this usage is mostly for debugging and understanding etc., I think the efficiency issue is not that critical.)

xadupre: The model is serialized here and then deserialized in C++. I think these steps should be avoided. The only change to make that happen is to change the C++ signature to take FunctionProto, ModelProto or a GraphProto and remove the call to ParseFromString. Users may call this function on deep learning models to compare two graphs. Avoiding serialization could make it a lot faster.

gramalingam: Specifically, I don't think that Python serialization and C++ deserialization are inverses of each other (at least, it is not documented to be so, and I doubt it is). So, their composition is not a "no-op", but it serves to convert from the python-representation to C++-representation.

onnx/printer.py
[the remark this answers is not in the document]
xadupre: I looked into the code and you are right. It is not trivial. I saw that other functions are using the same trick as you did. I guess the protobuf parser does not copy data and just creates intermediate structures pointing to the raw data. SerializeToString should also avoid copying the data. Let's keep it that way then.

onnx/defs/parser.h
gramalingam: May be. Doesn't seem critical to me yet. We can do this later, if needed.

onnx/defs/printer.cc
```diff
@@ -3,108 +3,211 @@
...
+    if (addsep)
+      output << "," << std::endl;
+    output << std::setw(indent_level) << ' ' << KeyWordMap::ToString(key) << ": ";
+    print(val);
+  }
+
+  inline void printKeyValuePair(KeyWordMap::KeyWord key, const std::string& val) {
+    output << "," << std::endl;
+    output << std::setw(indent_level) << ' ' << KeyWordMap::ToString(key) << ": ";
+    printQuoted(val);
```
wschin, line 95: Maybe replace `printQuoted` with `print` and implement a `print` for `std::string`? Then, we can merge these  two `printKeyValuePair`s.

onnx/defs/printer.cc
```diff
@@ -3,108 +3,211 @@
...
+    output << open;
+    for (auto& elt : coll) {
+      output << sep;
+      print(elt);
+      sep = separator;
+    }
+    output << close;
+  }
+
+  std::ostream& output;
```
wschin, line 110:
```suggestion
  std::ostream& output_;
```

onnx/defs/printer.h
```diff
@@ -38,4 +38,13 @@ std::ostream& operator<<(std::ostream& os, const GraphProto& graph);
 
 std::ostream& operator<<(std::ostream& os, const FunctionProto& fn);
 
+std::ostream& operator<<(std::ostream& os, const ModelProto& model);
+
+template <typename ProtoType>
+std::string ProtoToString(const ProtoType& proto) {
```
wschin, line 44: Why is this function added given we have all `operator<<`s above?

onnx/onnx_cpp2py_export/parser.pyi
```diff
@@ -7,3 +7,7 @@ def parse_model(model: str) -> Tuple[bool, bytes, bytes]:
 
 def parse_graph(graph: str) -> Tuple[bool, bytes, bytes]:
     ...
+
+
+def parse_function(function: str) -> Tuple[bool, bytes, bytes]:
```
wschin, line 12: Function requires documentation. Personally, an ideal document should contain
- clear description of each argument and returned value. In this pyi file, what are `bool`, `bytes`, and `bytes`?
- Some input-output examples.
Please consider add them if it makes sense. :)

onnx/defs/parser.h
```diff
@@ -269,13 +278,32 @@ class ParserBase {
...
-      result.value = std::string(from + 1, next_ - from - 2); // skip enclosing quotes
+      if (has_escape) {
+        std::string& target = result.value;
+        target.clear();
+        target.reserve(next_ - from - 2); // upper bound
+        // *from is the starting quote. *(next_-1) is the ending quote.
+        // Copy what is in-between, except for the escape character
+        while (++from < next_ - 1) {
+          // Copy current char, if not escape, or next char otherwise.
+          target.push_back(*from != '\\' ? (*from) : *(++from));
```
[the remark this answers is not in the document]
gramalingam: True. On the flip-side, we don't know the string-length upfront. So, the string object will keep growing, via re-allocation of memory, as we keep pushing back chars. Asymptotically, that will also cause a doubling of the processing time, plus extra memory-allocation/fragmentation etc. We could reserve some string-length, if we know what's the likely string-sizes, but I decided this wasn't too bad.

onnx/onnx_cpp2py_export/parser.pyi
gramalingam: Added, thanks!

onnx/defs/printer.cc
gramalingam: Done, thanks!

onnx/defs/printer.h
gramalingam: It's just a convenient short-hand. Further, for Python, we just expose this functionality instead of the operator<<.

onnx/defs/printer.cc
gramalingam: That's how I started. But ran into C++ template-specialization issues (template-specialization is allowed for class-members only by some compilers, not all.) Furthermore, we use `print(str)` as well as `printQuoted(str)`, both are needed.

onnx/defs/parser.h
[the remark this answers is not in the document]
wschin: Makes sense. It sounds like comparing "reallocating time" and "data loading time". We can do a study if it really becomes a problem.

wschin (approved):
LGTM now. Thank you!

gramalingam:
@liqunfu : could you take a look? This requires approval from archinfra sig too. Thanks!

onnx/printer.py
xadupre: A Python ModelProto and a C++ ModelProto are different: the Python ModelProto wraps the C++ ModelProto, probably in a capsule. I expect that the Python and C++ SerializeToString to produce the exact same sequence of bytes. That's what pybind11 does. They use swig to do it for ModelProto. It should work in a similar way.

xadupre: I did some measurements. I was hoping to see that serialization is independant of the size but it does not. So call SerializeToString and then ParseFromString is really inefficient.

Figure_1

gramalingam: > A Python ModelProto and a C++ ModelProto are different: the Python ModelProto wraps the C++ ModelProto, probably in a capsule. I expect that the Python and C++ SerializeToString to produce the exact same sequence of bytes. That's what pybind11 does. They use swig to do it for ModelProto. It should work in a similar way.

Of course, SerializeToString will produce the same sequence of bytes, since that is defined by the Protobuf semantics. But that is not the question. The question is whether the C++ in-memory-representation is available in a Python ModelProto (as you imply). If so, what is the API/call we can use to get the C++ in-memory-representation from a Python ModelProto?

gramalingam: (The question is about the Deserialize part, that is: creating the in-memory-representation used by the two languages).

xadupre: See https://github.com/onnx/onnx/pull/4274.

jcwchen (approved):
Only one nit comment. Exciting to see Python printer has more supports!

onnx/test/printer_test.py
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: Apache-2.0
+import onnx
+import unittest
+from onnx import helper, parser, printer, GraphProto
```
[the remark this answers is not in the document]
gramalingam: I can fix this manually, but out of curiosity, what will happen with/after Justins PR? Is there a specific auto-format tool/command recommended for people to apply? Thanks!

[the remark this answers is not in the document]
jcwchen: Good question. I think all these format tools will be included into style.sh as mentioned [here](https://github.com/onnx/onnx/pull/4296#pullrequestreview-1012856368). Then developer can simply run single script to check whether there is any error before submitting CIs. For auto-apply, perhaps we can also introduce another script if needed.

PR #4249, opened 2022-06-04
Fix incorrect expected output in windowing and stft tests, and window function function def.

smk2007:
Fix incorrect expected output in windowing and stft tests, and window function function def.

Fix failing Window Functions Tests (Hann, Hamming, Blackman)
The ONNX tests have 2 issues.

1. They are generating functions that do not take into account the periodic attribute in the function definition.
The default behavior in ONNX runtime has not been updated to match, and is missing the periodic attribute the contrib op definition always generated the "periodic" (window size instead of window size -1 ) window and not the "symmetric" window.
2. Incorrect expected values are produced by the window functions as constants are incorrectly set.

3. Window Functions' Functional Defs are incorrect (and dont take into account the periodic attribute.

A additional test was also added to validate the period/symmetric attribute.

Fix STFT Tests
The STFT test is also wrong because it doesn't take into consideration that the default output is onesided.
Consequently it incorrectly expects dimension to be 16, instead of half+1 (9).


gramalingam (approved):

gramalingam:
@etiotto : I believe that we will need to cherrypick this PR for the release, since it fixes the failing tests.

etiotto:
@gramalingam ok I will create a PR for that immediately

PR #4256, opened 2022-06-07
Fix: Window functions generate incorrect shape for with symmetric attribute

smk2007:
Fix: Window functions generate incorrect shape for with symmetric attribute

Affects: hannwindow, hammingwindow, and blackmanwindow operators.

Issue: The periodic=0 option to create a symmetric window for these operators was incorrectly causing the size of the output to be reduced by 1 in the function definition.

Test added to verify output shape with periodic attribute set to 0.

STFT test slicing expected output incorrectly. This causes the DFT to be run on a different size, as opposed to be being run on the original size, and sliced after.


garymm:
@etiotto please cherry pick

PR #4259, opened 2022-06-08
Update README with build modifications

Yesha-Thakkar:
**Description**
- Describe your changes.
This is a test PR with documentation clarifications for building Protobuf and ONNX from source.

**Motivation and Context**
This change allows for those debugging build issues to easily resolve some of the issues they might run into. Ensuring that packages are compatible will reduce the number of potential errors in the build process. 


README.md
```diff
@@ -114,6 +115,15 @@ Alternative: if you don't want to change your PATH, you can set ONNX_PROTOC_EXEC
...
 
+Before proceeding, ensure that you have pybind11 installed on your computer. You can do this using pip:
+```bat
+pip install pybind11
+```
+Alternatively, you can use conda-forge:
+```bat
+conda install -c conda-forge pybind11
+```
+
```
jcwchen, line 126: Is manual installation for pybind11 really needed? ONNX has its own pybind11 as submodule. Could you please try to build ONNX from source without manual installation for pybind11? Please let me know if you bump into any issues. Thanks!

README.md
```diff
@@ -90,6 +90,7 @@ The ON/OFF depends on what kind of protobuf library you have. Shared libraries a
 If you are building ONNX from source, it is recommended that you also build Protobuf locally as a static library. The version distributed with conda-forge is a DLL, but ONNX expects it to be a static library. Building protobuf locally also lets you control the version of protobuf. The tested and recommended version is 3.16.0.
 
 The instructions in this README assume you are using Visual Studio.  It is recommended that you run all the commands from a shell started from "x64 Native Tools Command Prompt for VS 2019" and keep the build system generator for cmake (e.g., cmake -G "Visual Studio 16 2019") consistent while building protobuf as well as ONNX.
+If you have any protobuf shared libraries (such as libprotobuf) installed, and you are planning to build Protobuf locally as a static library, ensure that the libprotobuf version is also set to 3.16.0 to ensure compatibility.
```
jcwchen, line 93: Ideally, it would be the best to only have one Protobuf installed in the environment to prevent any possible conflict issues. IMO, perhaps mention something like: If you see Protobuf failures while building Protobuf or ONNX from source, please make sure you only have one installed Protobuf (either static or shared) with recommended version to ensure compatibility.

README.md
Yesha-Thakkar: When I built ONNX from source during my very first installation, I ran into the following issue:

"could not find pybind11 missing: pybind11_dir"

So I assumed the manual installation of pybind11 was necessary. I can try building ONNX from source without this installation again, though!

jcwchen: Please run `git submodule update --recursive` before building ONNX from source if you didn't include `--recursive` while git clone previously.

Yesha-Thakkar: It seems to work this time! It appears that there were multiple versions of some packages in my conda environment, which caused some issues. I can change this to say, "If you receive an error, check the following:" instead of suggesting a manual installation. Thanks!

README.md
```diff
@@ -123,6 +126,9 @@ git submodule update --init --recursive
 set CMAKE_ARGS=-DONNX_USE_LITE_PROTO=ON
 pip install -e .
 ```
+
+Note: if you run into any issues while building ONNX, please ensure that the Python version for your activated conda environment is set to 3.8, and that python38.lib exists in the environment's folder on your device.
```
jcwchen, line 130: Thank you for keeping improving README.md! but the build document should be more generic -- I would suggest:
- No specific Python version (except there is a specific issue for that version)
- Not assume users use Conda environment

May I understand the intention of this note? What shall we do if pythonxx.lib does not exist?

Yesha-Thakkar: Building ONNX and running the pip command returns the following error if there are conflicting versions of python. Since I was working with Anaconda 3.9, but my environment was in python 3.8, I ran into issues. If there are no assumptions that the user is working in a Conda environment or a specific Python environment, these changes are not necessary.

Should I close this PR, or are there any changes you recommend I make? Please let me know what you think is the best move.

jcwchen: Thank you for providing the context. So it seems like a conflict issue about having multiple installed Python in certain environment. I might encounter similar issues before -- typically cleaning all prebuilt files (possibly there are some existing built files were generated by another Python environment) and rebuilding ONNX from source should help. Besides, if you use `pip install -e .` to build ONNX from source, you need to make sure your `pip` and `python` refer to the same installed Python. For instance, if you have one from native Python and another one from Conda Python with difference Python versions, you will probably see failures. If the issue you bumped into is reproducible, you can try whether the methods above help to resolve it.

> Should I close this PR, or are there any changes you recommend I make? Please let me know what you think is the best move.

Perhaps we can reword the sentences to make them more generic and then put them into [the section of common errors](https://github.com/onnx/onnx#common-errors). It can be something like: if you bump into issues like missing pythonXX.lib while install ONNX from source, please clean all existing build files and rebuild ONNX again. Also please make sure you have consistent Python version for common Python commands (e.g., `python` and `pip`).

gramalingam: Yes, the "common errors" seems like a good place for these kind of troubleshooting hints (and FAQs).

jcwchen (commented):
Basically LGTM. One comment about a typo. Thank you for the improvement!

jcwchen (approved):

PR #4262, opened 2022-06-10
handle raw types correctly in helper.make_tensor

garymm:
Before this change, the newly added `test_make_int8_tensor` would fail
because `TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE[INT8] = FLOAT32`. But raw
INT8 values are packed and are not the same size as FLOAT32 values.


gramalingam (approved):

jcwchen (approved):
Thank you so much for the quick fix! I can confirm this is the right way. Please fix flake8 failure.

I misunderstood so my fix (https://github.com/onnx/onnx/pull/4234) was incorrect --

Instead of storage type, actually make_tensor with raw directly uses its original NumPy type (defined in TENSOR_TYPE_TO_NP_TYPE) as bytes. Only TensorProto.FLOAT16 and TensorProto.BFLOAT16 are exceptions. Although now TensorProto.FLOAT16 uses uint16 as raw_data, the itemsize for uint16 and float16 is the same (2) so TensorProto.FLOAT16 is not an issue.

Thus, just like what this PR does, we only need to handle the special case TensorProto.BFLOAT16, because float32's itemsize is 4 and uint16 itemsize is 2 (there is a truncation when processing original float32 NumPy array).

Another finding is for making float16 data with raw, actually there is no need to convert original float16 NumPy array as `np_array.view(dtype=np.uint16).flatten().tobytes()` for raw_data, because their tobytes() are identical (See [here](https://github.com/onnx/onnx/blob/a92870cdf359297495a118184dca2eaecee4b717/onnx/test/helper_test.py#L414)). I will send another PR to improve it. Let's merge this PR first for upcoming ONNX 1.12.

garymm:
@jcwchen please merge

PR #4263, opened 2022-06-10
Fix layer normalization's reference outputs

wschin:
Existing outputs are always generated using default `epsilon`, so the reference outputs with non-default `epsilon` are incorrect. This PR fixes this problem. This change is tested with ONNXRuntime (commit: aceaa0a1f7ebc8919999c7b0e88bb565373c1e6d) and the tests are passing.


gramalingam (approved):

VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.12.0
```
garymm, line 1: Oops :-)
I'll send a fix.

garymm (commented):
Thanks!

VERSION_NUMBER
wschin: Thank you! This is fixed in `main` branch but I didn't touch release branch.

PR #4265, opened 2022-06-10
Revert "Fix layer normalization's reference outputs (#4263)"

wschin:
Reverts onnx/onnx#4263 and re-apply the fix. #4263 shouldn't change version number in master branch.


wschin:
This PR is replaced by #4266.

Issue #4261, opened 2022-06-10
provide a nicer API for accessing the proper member of TensorProto


garymm:
Currently code like this is repeated several times:

```py
field = mapping.STORAGE_TENSOR_TYPE_TO_FIELD[
    mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE[data_type]]
getattr(tensor, field)
```

This is repetitive and can be encapsulated in a helper function.
Also the name "storage tensor type" is misleading and has led to at least one bug.

We should:
1. Add a function that does all this
2. Deprecate `mapping.STORAGE_TENSOR_TYPE_TO_FIELD` and `mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE`. Ideally log a warning when they're accessed.
3. Eventually, then in the next release, remove those fields entirely.

PR #4270, opened 2022-06-13
Improve mapping and add more tests for make_tensor

jcwchen:
**Describe your changes**
- Refactor mapping
- Deprecate TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE
- Use functions instead of more maps to prevent confusions and save binary size
- Add more tests (test_make_tensor) for more tensor types 

**Motivation and Context**
- Relevant to https://github.com/onnx/onnx/issues/4261. Current mapping is quite confusing and there are some duplicate implementation for getting field by a tensor type.
- Now there are missing some tensor types for test_make_tensor w/ and w/o raw.


onnx/mapping.py
```diff
@@ -77,3 +68,24 @@
     int(OptionalProto.MAP): 'map_value',
     int(OptionalProto.OPTIONAL): 'optional_value'
 }
+
+
+def to_np_type(tensor_type: str) -> Any:
```
gramalingam, line 73: Are the `str` annotations correct? (Here and several places below.)

onnx/mapping.py
```diff
@@ -77,3 +68,24 @@
...
+    return TENSOR_TYPE_MAP[tensor_type][0]
+
+
+def to_sotrage_tensor_type(tensor_type: str) -> int:
+    return TENSOR_TYPE_MAP[tensor_type][1]
+
+
+def to_string(tensor_type: str) -> str:
```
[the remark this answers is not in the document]
gramalingam: Suggesting `tensor_dtype` as `tensor_type` is somewhat overloaded already. We mean the element-type of tensors, right?

onnx/test/hub_test.py
```diff
@@ -81,3 +81,7 @@ def test_get_model_info(self) -> None:
         hub.get_model_info("mnist", self.repo, opset=8)
         hub.get_model_info("mnist", self.repo)
         self.assertRaises(AssertionError, lambda: hub.get_model_info("mnist", self.repo, opset=-1))
+
+
+if __name__ == '__main__':
```
garymm, line 86: use double quotes for consistency with rest of file

onnx/test/parser_test.py
```diff
@@ -66,3 +66,7 @@ def test_parse_model_error(self) -> None:
            }
            '''
         self.assertRaises(onnx.parser.ParseError, lambda: onnx.parser.parse_model(input))
+
+
+if __name__ == '__main__':
```
garymm, line 71: use double quotes for consistency with rest of file

onnx/test/helper_test.py
```diff
@@ -660,5 +594,37 @@ def test_initializer_no_matching_graph_input(self) -> None:
...
 
 
+@pytest.mark.parametrize("tensor_type",
+    [t for t in TENSOR_TYPE_MAP if t not in {TensorProto.BFLOAT16, TensorProto.STRING, TensorProto.COMPLEX64, TensorProto.COMPLEX128}],
+    ids=lambda tensor_type: to_string(tensor_type)
+)
+def test_make_tensor_vals(tensor_type: str) -> None:
+    np_array = np.random.randn(2, 3).astype(TENSOR_TYPE_MAP[tensor_type])
+    tensor = helper.make_tensor(
+        name='test',
```
garymm, line 604: use double quotes

onnx/test/helper_test.py
```diff
@@ -660,5 +594,37 @@ def test_initializer_no_matching_graph_input(self) -> None:
...
+
+
+@pytest.mark.parametrize("tensor_type",
+    [t for t in TENSOR_TYPE_MAP if t not in {TensorProto.BFLOAT16, TensorProto.STRING}],
+    ids=lambda tensor_type: to_string(tensor_type)
+)
+def test_make_tensor_raw(tensor_type: str) -> None:
+    np_array = np.random.randn(2, 3).astype(TENSOR_TYPE_MAP[tensor_type])
+    tensor = helper.make_tensor(
+        name='test',
```
garymm, line 619: use double quotes

onnx/test/helper_test.py
```diff
@@ -660,5 +594,37 @@ def test_initializer_no_matching_graph_input(self) -> None:
...
+        data_type=tensor_type,
+        dims=np_array.shape,
+        vals=np_array.tobytes(),
+        raw=True,
+    )
+    np.testing.assert_equal(np_array, numpy_helper.to_array(tensor))
+
+
 if __name__ == '__main__':
```
garymm, line 628: ditto, double quotes

garymm (changes requested):
I think we should go much further and make only a small number of very useful, user-friendly things public and hide the rest.

Everything currently inside mapping.py should be made private, since I think it's all implemtnation details that clients shouldn't rely on.

I propose:
* Add `helper.tensor_field_for_data_type(TensorProto, DataType)`, `helper.tensor_field_for_numpy_dtype(TensorProto, np.dtype)`. These do all of the mapping lookups and the `getattr` call.
* Move `mapping.py` to `_mapping.py`
* Add a new `mapping.py` that does something like:

```python
import warnings
warnings.warn("onnx.mapping is deprecated and will be removed in a future release. Use functions in onnx.helper instead.", ...)

# for temporary backwards compatibility
from _mapping import *
```

If there's something else in mapping.py that is truly needed by users, we can expose that in `helper.py` as well.

WDYT?

onnx/test/helper_test.py
```diff
@@ -306,28 +306,28 @@ def test_model(self) -> None:
...
-        self.assertEqual(model_def.producer_name, 'test')
+        model_def = helper.make_model(graph_def, producer_name="test")
+        self.assertEqual(model_def.producer_name, "test")
 
     def test_model_docstring(self) -> None:
         graph = helper.make_graph([], "my graph", [], [])
-        model_def = helper.make_model(graph, doc_string='test')
-        # models may have their own documentation, but don't have a name
+        model_def = helper.make_model(graph, doc_string="test")
+        # models may have their own documentation, but don"t have a name
```
garymm, line 315: undo

onnx/test/helper_test.py
```diff
@@ -374,73 +374,73 @@ def test_make_string_tensor(self) -> None:
         self.assertEqual(string_list, list(tensor.string_data))
 
     def test_make_bfloat16_tensor(self) -> None:
-        # numpy doesn't support bf16, so we have to compute the correct result manually
+        # numpy doesn"t support bf16, so we have to compute the correct result manually
```
garymm, line 377: undo

onnx/helper.py
```diff
@@ -362,8 +362,7 @@ def make_tensor(
             vals = np.array(vals).astype(np_dtype).view(dtype=np.uint16).flatten().tolist()
         elif data_type == TensorProto.BFLOAT16:
             vals = list(map(float32_to_bfloat16, np.array(vals).astype(np_dtype).flatten().tolist()))
-        field = mapping.STORAGE_TENSOR_TYPE_TO_FIELD[
-            mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE[data_type]]
+        field = mapping.to_field(data_type)
```
garymm, line 365: I would instead have the function take in a tensor and a data type and returns the field. Because every single caller of this function is going to do the same `getattr(tensor, field)`, so it should just be inside the helper function.

onnx/mapping.py
```diff
@@ -2,52 +2,42 @@
...
+    int(TensorProto.UINT64): (np.dtype('uint64'), int(TensorProto.UINT64), 'TensorProto.UINT64'),
+    int(TensorProto.STRING): (np.dtype('object'), int(TensorProto.STRING), 'TensorProto.STRING'),
 }
 
+# This map is used for converting TensorProto values into numpy arrays
+TENSOR_TYPE_TO_NP_TYPE = {tensor_type: value[0] for tensor_type, value in TENSOR_TYPE_MAP.items()}
+# This is only used to get keys into STORAGE_TENSOR_TYPE_TO_FIELD.
+# TODO(https://github.com/onnx/onnx/issues/4261): Remove this.
+warnings.warn(str("`mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE` is deprecated in ONNX 1.13 and will be removed in next release. To silence this warning, please use `to_storage_tensor_type` instead."), DeprecationWarning, stacklevel=2)
+TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE = {tensor_type: value[1] for tensor_type, value in TENSOR_TYPE_MAP.items()}
```
garymm, line 35: Can you figure out a way to have it warn only when the dict is accessed? Maybe something like:

```python
class WarningDict(dict):
   def __getitem__(self, key):
       warnings.warn(...)
       return super().__getitem__(self, key)
```

garymm (changes requested):

onnx/helper.py
jcwchen: I introduced `get_attr_from_sequence_elem_type` and `get_attr_from_optional_elem_type` to achieve that. I will still keep `storage_type_to_field` and `optional_type_to_field` public, because I saw external code has already used them (by `STORAGE_ELEMENT_TYPE_TO_FIELD` and `OPTIONAL_ELEMENT_TYPE_TO_FIELD`). Please let me know if you still think we should hide them as private functions.

onnx/mapping.py
jcwchen: Brilliant! Thanks for sharing. I tried it and it worked well.

onnx/mapping.py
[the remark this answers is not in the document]
jcwchen: Updated. I used `tensor_dtype_to_XXX` for naming. Thanks!

onnx/helper.py
```diff
@@ -1029,3 +1026,48 @@ def make_training_info(algorithm: GraphProto, algorithm_bindings: AssignmentBind
             binding.value = v
 
     return training_info
+
+
+def tensor_dtype_to_np_type(tensor_dtype: int) -> np.dtype:
```
gramalingam, line 1031: I think it will also help to document all of these, given the subtle differences (eg., like storage_tensor_type vs storage_numpy_type, etc.). Given that the functions are inter-related, we could put it into some doc file perhaps,
just to clarify the different types, and when/where they are used or should be used.

onnx/helper.py
```diff
@@ -814,8 +812,7 @@ def str_list(str_elem: Callable[[_T], str], xs: Sequence[_T]) -> str:
             content.append("<Tensor>")
         else:
             # special case to print scalars
-            field = STORAGE_TENSOR_TYPE_TO_FIELD[attr.t.data_type]
-            content.append(f'<Scalar Tensor {str(getattr(attr.t, field))}>')
```
jcwchen, line 818: It seems incorrectly used... We shouldn't use STORAGE_TENSOR_TYPE_TO_FIELD for a TensorProto so I just removed it.

onnx/helper.py
gramalingam: We could also try to reduce the number of helper functions down below, in case it helps. Not sure if all of these are used to a significant degree?

onnx/mapping.py
```diff
@@ -1,53 +1,51 @@
...
-    int(TensorProto.BFLOAT16): np.dtype('float32'),  # Native numpy does not support bfloat16 so now use float32 for bf16 values
-    int(TensorProto.DOUBLE): np.dtype('float64'),
-    int(TensorProto.COMPLEX64): np.dtype('complex64'),
-    int(TensorProto.COMPLEX128): np.dtype('complex128'),
-    int(TensorProto.UINT32): np.dtype('uint32'),
-    int(TensorProto.UINT64): np.dtype('uint64'),
-    int(TensorProto.STRING): np.dtype('object')
+# tensor_dtype: (numpy type, storage type, string name)
+TENSOR_TYPE_MAP = {
+    int(TensorProto.FLOAT): (np.dtype('float32'), int(TensorProto.FLOAT), 'TensorProto.FLOAT'),
```
xadupre, line 10: Why do you need the key in the value tuple?

onnx/mapping.py
```diff
@@ -1,53 +1,51 @@
...
+    int(TensorProto.UINT8): (np.dtype('uint8'), int(TensorProto.INT32), 'TensorProto.UINT8'),
+    int(TensorProto.INT8): (np.dtype('int8'), int(TensorProto.INT32), 'TensorProto.INT8'),
+    int(TensorProto.UINT16): (np.dtype('uint16'), int(TensorProto.INT32), 'TensorProto.UINT16'),
+    int(TensorProto.INT16): (np.dtype('int16'), int(TensorProto.INT32), 'TensorProto.INT16'),
+    int(TensorProto.INT32): (np.dtype('int32'), int(TensorProto.INT32), 'TensorProto.INT32'),
+    int(TensorProto.INT64): (np.dtype('int64'), int(TensorProto.INT64), 'TensorProto.INT64'),
+    int(TensorProto.BOOL): (np.dtype('bool'), int(TensorProto.INT32), 'TensorProto.BOOL'),
+    int(TensorProto.FLOAT16): (np.dtype('float16'), int(TensorProto.UINT16), 'TensorProto.FLOAT16'),
```
xadupre, line 18: not sure about the value here

onnx/mapping.py
xadupre, line 10: maybe a namedtuple would help to remember the meaning of each element

jcwchen:
> That makes sense to me, but I only have one concern that I believe external users have already started to use variables from mapping.py long time ago... IMO, it's safer to move these variables into private after major version change in ONNX. Still, we can introduce helper functions first. I have added a few in helper.py. Please review it. Thanks!

> Even for TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE, it has been used from a few places: https://github.com/search?q=org%3Aonnx+TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE&type=code. Do we still want to deprecate it in next release?

After discussion yesterday, we concluded that at least at this moment we can show deprecated messages for all those we would like to deprecate to warn users in advance. I will add it later.

onnx/helper.py
jcwchen: Good point. Adding more docs for these functions should be very useful. Just added API comments in the functions and two commonly-used functions (`tensor_dtype_to_np_type` and `tensor_dtype_to_field`) in PythonAPIOverview.md. Please review it. Thanks!

> We could also try to reduce the number of helper functions down below, in case it helps. Not sure if all of these are used to a significant degree?

I think perhaps functions like `sequence_type_to_field`, `optional_type_to_field`, `tensor_dtype_to_storage_tensor_type` can be private functions, because ONNX code can directly use outer functions like `get_attr_from_sequence_elem_type`, `get_attr_from_optional_elem_type`, `tensor_dtype_to_storage_numpy_type` correspondingly. I am not very sure whether those functions are helpful to external users though. @gramalingam WDYT?

onnx/mapping.py
jcwchen: Great idea. I have introduced `TensorDtypeMap = NamedTuple(
    "TensorDtypeMap", [("np_type", np.dtype), ("storage_type", int), ("name", str)]
)` to better manage this data structure.

onnx/mapping.py
jcwchen: I merged previous `TENSOR_TYPE_TO_NP_TYPE` and `TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE` with the new naming map into the single map `TENSOR_TYPE_MAP`, because they have the same set of keys (TensorProto data_type). They were the ground truth here and not tested previously either. If the unit test is really needed, I am inclined to have another PR for it since this PR is already a big one and actually I didn't change the original content here.

onnx/mapping.py
```diff
@@ -1,84 +1,167 @@
...
+        return (
+            self._origin_function == other._origin_function
+            and self._future_function == other._future_function
+        )
+
+    def __getitem__(self, key: Union[int, str]) -> Any:
+        warnings.warn(
+            str(
+                f"`mapping.{self._origin_function}` is deprecated in ONNX 1.13 and will "
+                + "be removed in next release. To silence this warning, please use `helper.{_self.future_function}` instead."
```
gramalingam, line 92: Are we definitely planning to remove in release 1.14? If we are not sure, we can change the message to "will be removed soon" or "will be removed in the next release or so."

onnx/mapping.py
```diff
@@ -1,84 +1,167 @@
...
+            stacklevel=2,
+        )
+        return super().__getitem__(key)
+
+
+# This map is used for converting TensorProto values into numpy arrays
+TENSOR_TYPE_TO_NP_TYPE = DeprecatedWarningDict(
+    {tensor_dtype: value.np_type for tensor_dtype, value in TENSOR_TYPE_MAP.items()},
+    "TENSOR_TYPE_TO_NP_TYPE",
+    "tensor_dtype_to_np_typeto_storage_tensor_type",
```
gramalingam, line 104: Looks like there is some type or copy-paste error above?

onnx/helper.py
```diff
@@ -1098,3 +1095,121 @@ def make_training_info(
...
+    """
+    Convert a TensorProto's data_type to corresponding numpy dtype. It can be used while making tensor.
+
+    :param tensor_dtype: TensorProto's data_type
+    :return: numpy's data_type
+    """
+    return mapping.TENSOR_TYPE_MAP[int(tensor_dtype)].np_type
+
+
+def tensor_dtype_to_storage_tensor_type(tensor_dtype: int) -> int:
```
xadupre, line 1111: why dtype sometimes and only type some other types?

docs/PythonAPIOverview.md
```diff
@@ -280,27 +293,27 @@ In this example we merge two models by connecting each output of the first model
...
+model2 = onnx.load("path/to/model2.onnx")
 #   agraph (float[N] X, float[N] Y) => (float[N] Z)
 #   {
 #      Z = Mul(X, Y)
 #   }
 
 combined_model = onnx.compose.merge_models(
     model1, model2,
-    io_map=[('C', 'X'), ('D', 'Y')]
+    io_map=[("C", "X"), ("D", "Y")]
```
xadupre, line 311: I would have one this change in a separate PR. It would be easier to read.

onnx/numpy_helper.py
```diff
@@ -36,10 +36,9 @@ def to_array(tensor: TensorProto, base_dir: str = "") -> np.ndarray:
         raise TypeError("The element type in the input tensor is not defined.")
 
     tensor_dtype = tensor.data_type
-    np_dtype = mapping.TENSOR_TYPE_TO_NP_TYPE[tensor_dtype]
-    storage_type = mapping.TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE[tensor_dtype]
```
xadupre, line 40: In PR #4510, I remove the use of dictionary `TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE`. It think use int8 or uint8 to store bool should not be allowed.

docs/PythonAPIOverview.md
jcwchen: Sounds good. I have create a specific PR for it: https://github.com/onnx/onnx/pull/4530.

onnx/helper.py
jcwchen: Good catch. Updated by always using dtype.

onnx/mapping.py
jcwchen: To be flexible, I changed it into: `XXX is now deprecated and will be removed in the next release or so.`

onnx/mapping.py
jcwchen: Thanks for catching this! Corrected.

onnx/numpy_helper.py
```diff
@@ -156,13 +155,9 @@ def to_list(sequence: SequenceProto) -> List[Any]:
     """
     lst: List[Any] = []
     elem_type = sequence.elem_type
-    value_field = mapping.STORAGE_ELEMENT_TYPE_TO_FIELD[elem_type]
-    values = getattr(sequence, value_field)
+    values = helper.get_attr_from_sequence_elem_type(sequence, elem_type)
```
gramalingam, line 158: IMHO, I think it would be simpler to rewrite the code as below:
```py
   if elem_type == SequenceProto.TENSOR:
      return [to_array(v) for v in sequence.tensor_values]
   elif elem_type == SequenceProto.SPARSE_TENSOR:
      return [to_array(v) for v in sequence.sparse_tensor_values]
   elif elem_type == SequenceProto.SEQUENCE:
      return [to_list(v) for v in sequence.sequence_values]
   etc.
```
I don't think we need to define the helper methods like `get_attr_from_seqience_elem_type`.

onnx/helper.py
```diff
@@ -1098,3 +1095,121 @@ def make_training_info(
...
+    """
+    Convert a SequenceProto's elem_type to corresponding field name.
+
+    :param tensor_dtype: SequenceProto's elem_type
+    :return: field name
+    """
+    return cast(str, mapping.STORAGE_ELEMENT_TYPE_TO_FIELD[elem_type])
+
+
+def get_attr_from_sequence_elem_type(tensor: SequenceProto, elem_type: int) -> Any:
```
gramalingam, line 1177: I don't think we need these utility/helper methods. Please see my comment in `numpy_helper.py` on rewriting the code there. I think splitting the code into these two files actually makes the logic of either one harder to understand. (I understand this is a problem with the pre-existing code introduced by someone else previously, and not this PR.)

onnx/numpy_helper.py
```diff
@@ -318,10 +313,9 @@ def to_optional(optional: OptionalProto) -> Optional[Any]:
     elem_type = optional.elem_type
     if elem_type == OptionalProto.UNDEFINED:
         return opt
```
gramalingam, line 315: Change to `return None`

onnx/numpy_helper.py
```diff
@@ -318,10 +313,9 @@ def to_optional(optional: OptionalProto) -> Optional[Any]:
...
-    value_field = mapping.OPTIONAL_ELEMENT_TYPE_TO_FIELD[elem_type]
-    value = getattr(optional, value_field)
+    value = helper.get_attr_from_optional_elem_type(optional, elem_type)
     # TODO: create a map and replace conditional branches
-    if elem_type == OptionalProto.TENSOR or elem_type == OptionalProto.SPARSE_TENSOR:
+    if elem_type in (OptionalProto.TENSOR, OptionalProto.SPARSE_TENSOR):
         opt = to_array(value)
     elif elem_type == OptionalProto.SEQUENCE:
         opt = to_list(value)
```
gramalingam, line 321: Change to `return to_list(optional.tensor_value)`

gramalingam: Likewise for all the cases. We don't need the helper utility method, whose purpose is only to do this switch/case anyway.

onnx/helper.py
```diff
@@ -1098,3 +1095,121 @@ def make_training_info(
...
+    """
+    Get the name of given TensorProto's data_type.
+
+    :param tensor_dtype: TensorProto's data_type
+    :return: the name of data_type
+    """
+    return mapping.TENSOR_TYPE_MAP[int(tensor_dtype)].name
+
+
+def tensor_dtype_to_storage_numpy_type(tensor_dtype: int) -> np.dtype:
```
gramalingam, line 1131: I wonder if we can omit this function. Users can call the two functions themselves. It is not easy to document/describe, and too many such functions in the API can confuse a reader.

docs/PythonAPIOverview.md
```diff
@@ -148,6 +148,16 @@ Runnable IPython notebooks:
 - [make_model.ipynb](/onnx/examples/make_model.ipynb)
 - [Protobufs.ipynb](/onnx/examples/Protobufs.ipynb)
 
+## Conversion utilities for mapping attributes in ONNX IR
```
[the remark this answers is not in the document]
jcwchen: Moved.

onnx/helper.py
jcwchen: Makes sense to me. Just omitted.

onnx/numpy_helper.py
jcwchen: Done. Thanks!

onnx/helper.py
jcwchen: For readability, yes I think keeping if-else statement makes more sense, but for maintainability, having a common function for the same mapping seems not too bad (we don't need to change the mapping in several places if mapping is changed). However, if the attribute mapping for protos is quite stable, I am fine with having if-else statements. I have removed attribute map for SequenceProto and OptionalProto and used if-else statements instead.

Not sure whether it is a concern -- that also means that previously the attribute mappings for SequenceProto and OptionalProto are public used functions and in the future they will be removed.

gramalingam (approved):

PR #4274, opened 2022-06-14
[WIP] experimental: c export of ModelProto in python package

xadupre:
**Description**
Many functions serialize a proto into bytes then deserialized in the C++ binding (checked, infershapes, ...) It could be faster by avoiding this step. This PR tries to make ModelProto and other available through pybind11.

**Motivation and Context**
- Performance


daquexian:
It is what I've been looking forward to for a long time. Serializing and deserializing between c++ and python waste much time in onnx optimizer. 

I have another idea to discuss: Why do we stick to protobuf and not switch to another more flexible format like flatbuffers? It is so strange that onnx, an op standard, is trying to solve the problems of a specific storage format. But the fact is we even create a mechanism (external data) just to solve the 2GB problem of protobuf.

PR #4277, opened 2022-06-15
Remove reference to numpy.typing

justinchuby:
**Description**

Remove reference to numpy.typing

**Motivation and Context**

onnx supports numpy > 1.16 (from requirements.txt) but `numpy.typing` is introduced in 1.20. The PR removes `numpy.typing` and instead creates the `_ShapeLike` type.

- It also corrects the return type for `combine_pairs_to_complex`
- Removes unused import `helper`

Needed for the 1.12 release


justinchuby:
@jcwchen we prob need a newer version of typing_extensions: https://github.com/python/typing/pull/724

onnx/numpy_helper.py
```diff
@@ -1,20 +1,19 @@
...
-from typing import Sequence, Any, Optional, List, Dict
 
 
-def combine_pairs_to_complex(fa: Sequence[int]) -> Sequence[np.complex64]:
+def combine_pairs_to_complex(fa: Sequence[int]) -> List[complex]:
     return [complex(fa[i * 2], fa[i * 2 + 1]) for i in range(len(fa) // 2)]
 
 
-def bfloat16_to_float32(data: np.ndarray, dims: nptyping._ShapeLike) -> np.ndarray:
+def bfloat16_to_float32(data: np.ndarray, dims: Union[int, Sequence[int]) -> np.ndarray:
```
jcwchen, line 16: Sorry, my typo:
```suggestion
def bfloat16_to_float32(data: np.ndarray, dims: Union[int, Sequence[int]]) -> np.ndarray:
```

justinchuby:
Could you update the rc so that it can be tested in pytorch?

PR #4279, opened 2022-06-15
Verify minimum supported numpy in Windows Release CI

jcwchen:
**Description**
Verify minimum supported numpy in Windows CI. Only apply it in Windows since numpy 1.16.6 does not support Python>=3.9 so we need to build numpy from source in CIs, which is time-consuming and it might bump into some issue unrelated to ONNX.  For now verifying it in Windows CI should be sufficient.

**Motivation and Context**
Since ONNX's requirements.txt has loose restriction for NumPy (1.16.6) and it uses NumPy (1.21.5) for release. To prevent issues with older NumPy version (>=1.16.6; <1.21.5), CI should verify the built wheel can work with the minimum supported numpy. For instance, https://github.com/onnx/onnx/pull/4277 numpy.typing was introduced by NumPy 1.20. The failure with lower numpy version is not caught by any CI.


jcwchen (commented):
This PR should pass all CIs after https://github.com/onnx/onnx/pulls has been merged.

liqunfu (approved):

PR #4296, opened 2022-06-20
Create inline lints with github actions

justinchuby:
**Description**
Create inline lints with github actions

**Motivation and Context**

- Added flake8, pylint, shellcheck, spellcheck and mypy lints to facilitate PR reviews.
- Updated mypy configs to exclude paths

Examples: https://github.com/justinchuby/onnx/runs/6974801964.


jcwchen (commented):
Thank you for proposing this PR. Let me provide more context: currently ONNX uses 3 kinds of format/style tools:
- flake8 (Python)
- mypy typecheck (Python)
- clang-format (C++)

See https://github.com/onnx/onnx/blob/main/tools/style.sh.

I am thinking perhaps we can put all Python related lint/type check into the same Python Lint CI (or maybe decouple typecheck from lint) and then put all C++ related things (now we only have clang-format tho) into another C++ Lint CI. In that case, we can finally remove existing style.sh command in Linux-CI to make it focus on ONNX implementation testing.

For style.sh, I would say we should include all Python/C++ lint/style tools into this script. This way developer can always simply run their code with it to see whether they can pass all of them before submitting to CIs. @justinchuby what's your thought? Thanks!

justinchuby:
@jcwchen that sounds like a good plan to me. In particular, we can (1) split lint checks from the tests so that they can still run in spite of formatting errors. (2) add the mypy checks in an other pr (maybe https://github.com/tsuyoshicho/action-mypy)

The reference workflow is from onnx runtime https://github.com/microsoft/onnxruntime/blob/master/.github/workflows/lint.yml, where we have inline annotations in the `lint-python` job. These are all reviewdog actions. They provided annotations in the PR UI and are non-blocking. So we are still fine even if reviewdog fails. The blocking checks are defined in `lint-python-format`, where we use the official `psf/black@stable` and `isort/isort-action@master` which should be more reliable. I think we can do the same in onnx

justinchuby:
I can update style.sh in the same pr

tools/style.sh
```diff
@@ -3,20 +3,26 @@
...
+err=0
+trap 'err=1' ERR
+
+echo -e "\n::group:: Check flake8..."
 flake8 onnx tools workflow_scripts
+echo -e "::endgroup::"
 
-echo -e "\n===> check mypy"
+echo -e "\n::group:: Check mypy"
 mypy . --no-site-packages
```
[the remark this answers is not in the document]
jcwchen: > Do we need to exclude third_party?

Makes sense to me. Please exclude third_party. Thanks!

justinchuby:
@jcwchen there are new mypy errors. Were they expected?

justinchuby:
Locally mypy yields

```
onnx/onnx_ml_pb2.pyi:10: error: Cannot find implementation or library stub for module named "google.protobuf.message"
onnx/onnx_ml_pb2.pyi:14: error: Cannot find implementation or library stub for module named "google.protobuf.internal.containers"
tools/protoc-gen-mypy.py:33: error: Cannot find implementation or library stub for module named "google.protobuf.descriptor_pb2"
tools/protoc-gen-mypy.py:33: error: Cannot find implementation or library stub for module named "google"
tools/protoc-gen-mypy.py:33: error: Cannot find implementation or library stub for module named "google.protobuf"
tools/protoc-gen-mypy.py:34: error: Cannot find implementation or library stub for module named "google.protobuf.compiler"
onnx/onnx_operators_ml_pb2.pyi:2: error: Cannot find implementation or library stub for module named "google.protobuf.message"
onnx/onnx_operators_ml_pb2.pyi:16: error: Cannot find implementation or library stub for module named "google.protobuf.internal.containers"
onnx/onnx_data_pb2.pyi:2: error: Cannot find implementation or library stub for module named "google.protobuf.message"
onnx/onnx_data_pb2.pyi:14: error: Cannot find implementation or library stub for module named "google.protobuf.internal.containers"
onnx/__init__.py:19: error: Cannot find implementation or library stub for module named "google.protobuf.message"
onnx/__init__.py:19: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
onnx/__init__.py:19: error: Cannot find implementation or library stub for module named "google"
onnx/__init__.py:19: error: Cannot find implementation or library stub for module named "google.protobuf"
onnx/__init__.py:67: error: Returning Any from function declared to return "bytes"
onnx/checker.py:20: error: Cannot find implementation or library stub for module named "google.protobuf.message"
onnx/helper.py:8: error: Cannot find implementation or library stub for module named "google.protobuf.message"
onnx/helper.py:8: error: Cannot find implementation or library stub for module named "google"
onnx/helper.py:8: error: Cannot find implementation or library stub for module named "google.protobuf"
workflow_scripts/test_model_zoo.py:96: error: Incompatible types in assignment (expression has type "str", variable has type "ModelProto")
onnx/test/symbolic_shape_test.py:59: error: Returning Any from function declared to return "Optional[TensorShapeProto]"
onnx/test/symbolic_shape_test.py:61: error: Returning Any from function declared to return "Optional[TensorShapeProto]"
onnx/test/shape_inference_test.py:59: error: Returning Any from function declared to return "Union[SupportsDunderLT, SupportsDunderGT]"
onnx/backend/test/runner/__init__.py:154: error: Value of type variable "SupportsRichComparisonT" of "sorted" cannot be "Type[TestCase]"
onnx/backend/test/runner/__init__.py:167: error: Value of type variable "SupportsRichComparisonT" of "sorted" cannot be "Dict[str, TestItem]"
Found 24 errors in 11 files (checked 254 source files)
```

which is less than in CI. Do we need to update the mypy version for it to read pyproject?

jcwchen:
> which is less than in CI. Do we need to update the mypy version for it to read pyproject?

There is an open issue to track this work item: https://github.com/onnx/onnx/issues/3964. I am happy to see mypy upgrade, but it can be a large PR. You can decouple it to another PR if needed. Thanks!

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,60 @@
+name: Lint
+
+on:
+  push:
+    branches:
+      - main
```
jcwchen, line 6: Please include rel-* branch as well (Or perhaps all branches).

tools/style.sh
```diff
@@ -3,20 +3,28 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-set -o errexit
+set +o errexit
```
jcwchen, line 6: May I understand why it turns off?

jcwchen (commented):
Thanks for the quick update. The lint CI looks good to me overall.

Let me summarize the TODO items:
1. Include new pylint, shellcheck, spellcheck (which are introduced by this PR) into current style.sh. 
2. Remove running style.sh in the Linux-CI after we have covered all check in the new Lint CIs.
3. (Optional) Have another single script auto apply lint formatter and .clang-format

It seems to me 1. can be included in this PR, since we indeed introduce these new things from this PR. It can be helpful to remind users.

tools/style.sh
justinchuby: I unset it so that it will always run all checks in the script. The `$err` will set the exit code if any of them fails

jcwchen (commented):
>1. Include new pylint, shellcheck, spellcheck (which are introduced by this PR) into current style.sh.

After offline discussion with @justinchuby, this work item is actually not necessary because these new checks are not required to pass and more like recommendations.

setup.cfg
```diff
@@ -57,6 +57,18 @@ no_implicit_optional = True
 # TODO disallow_subclassing_any = True
 disallow_untyped_decorators = True
 warn_unused_configs = True
+show_error_codes = True
+show_column_numbers = True
+exclude = (?x)(
```
jcwchen, line 62: image
@justinchuby I am seeing this in the CIs and my local. Is it not supported by older mypy? (ONNX uses 0.782)

justinchuby: Hmm, let me test that with 0.782

PR #4333, opened 2022-06-30
Clarify in the doc that ONNX should use rc as release candidate

jcwchen:
**Description**
Going forward, ONNX should always use `x.x.xrc1` as release candidates. Clarify it in the release document to prevent future confusion for release managers.

**Motivation and Context**
In the past ONNX releases, ONNX did not always use `x.x.xrc1` as release candidates, because some external code from partners do not handle non-numeric version number.  (For instance, 1.8.x and 1.9.x: https://test.pypi.org/project/onnx/#history)
 However, ONNX has used rc in VERSION_NUMBER for 2 recent releases (1.11 and 1.12), which did not cause any problem. Therefore, going forward ONNX should always use `x.x.xrc1` as release candidates.


jcwchen:
cc @garymm explicitly mentioning "ONNX should use rc for release candidates" in the release document should make your ORT PR https://github.com/microsoft/onnxruntime/pull/12039 more reliable.

garymm (approved):
I don't think it will make a difference for that code (it will handle a version string with or without any suffix), but seems OK either way.

docs/OnnxReleases.md
```diff
@@ -13,7 +13,7 @@ The ONNX project, going forward, will plan to release roughly on a four month ca
 * Before creating the release branch, increase `VERSION_NUMBER` in the main branch. The following files will be updated: [VERSION_NUMBER file](/VERSION_NUMBER) and
 [version.h](/onnx/common/version.h)
 
-* Please use a VERSION_NUMBER smaller than the target (release VERSION_NUMBER) and larger than the previous one to test TestPyPI before using the target VERSION_NUMBER. 
+* Please use VERSION_NUMBER with `rc` (e.g., `1.x.0rc1`) to test TestPyPI in advance before using the target VERSION_NUMBER for final release.
```
linkerzhang, line 16: Please use target VERSION_NUMBER with `rc` (e.g., `1.x.0rc1`) to test TestPyPI in advance before using it for final release.

gramalingam (approved):

PR #4351, opened 2022-07-08
Specify pairwise conversion behavior for the Cast op

yuanyao-nv:

**Description**
Add more details for conversion among numerical types for the Cast op

**Motivation and Context**
In the previous version of the spec for Cast, not much detail is given for out-of-range conversion among numerical types. Here we explicitly list all the pair-wise conversions and indicate when to expect undefined behavior.

Suggestions are welcome for conversions to/from string types, from people to whom string conversions are relevant.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -15023,6 +15023,22 @@ This version of the operator has been available since version 13 of the default 
   For example, a 64-bit float 3.1415926459 may be round to a 32-bit float 3.141592. Similarly, converting
   an integer 36 to Boolean may produce 1 because we truncate bits which can't be stored in the targeted type.
 
+  In more detail, the conversion among numerical types should follow these rules:
+
+  * Casting from floating point to:
+    * floating point: +/- infinity if OOR (out of range).
+    * fixed point: undefined if OOR.
+    * bool: +/- 0.0 to False; all else to True.
+  * Casting from fixed point to:
+    * floating point: +/- infinity if OOR. (+ infinity in the case of uint)
+    * fixed point: when OOR, discard higher bits and reinterpret (with respect to two's complement representation for
+  signed types). For example, 200 (int16) -> -56 (int8).
+    * bool: zero to False; nonzero to True.
+  * Casting from bool to:
+    * floating point: `{1.0, 0.0}`.
+    * fixed point: `{1, 0}`.
+    * bool: no change.
+
 #### Version
 
 This version of the operator has been available since version 13 of the default ONNX operator set.
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -2497,6 +2497,22 @@ expect(node, inputs=[size], outputs=[y],
   For example, a 64-bit float 3.1415926459 may be round to a 32-bit float 3.141592. Similarly, converting
   an integer 36 to Boolean may produce 1 because we truncate bits which can't be stored in the targeted type.
 
+  In more detail, the conversion among numerical types should follow these rules:
+
+  * Casting from floating point to:
+    * floating point: +/- infinity if OOR (out of range).
+    * fixed point: undefined if OOR.
+    * bool: +/- 0.0 to False; all else to True.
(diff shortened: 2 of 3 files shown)


yuanyao-nv:
> Thanks for adding the details. It looks good to me ... the only unclear part is the int/uint to int/uint part: I don't fully understand what truncation means with regards to the sign bit, for example. In the case where the value is IN range, it is not a pure truncation because of the sign bit, right? Can the behavior wrt sign bit be clarified? Thanks!

@gramalingam For in-range conversions, there should be no ambiguity because the integer should be exactly representable in the new dtype. 
Are you referring to out-of-range? There, a non-sign bit in a longer int can be interpreted as the sign bit of a shorter int during truncation. For example, 200 (int16) -> -56 (int8)

gramalingam:
@yuanyao-nv : thanks for the clarification, it looks fine. Maybe it is worth adding a line clarifying that the truncation semantics is with respect to a two's complement representation for negative numbers?

gramalingam (approved):

yuanyao-nv:
@gramalingam I've added a sentence to specify two's complement. Please review again. Thanks.

docs/Changelog.md
```diff
@@ -15023,6 +15023,22 @@ This version of the operator has been available since version 13 of the default
   For example, a 64-bit float 3.1415926459 may be round to a 32-bit float 3.141592. Similarly, converting
   an integer 36 to Boolean may produce 1 because we truncate bits which can't be stored in the targeted type.
 
+  In more detail, the conversion among numerical types should follow these rules:
+
+  * Casting from float to:
+    * float: +/- infinity if OOR (out of range).
```
p-wysocki, line 15029: It would be great to specify that it's about a case of casting to a lower-bit float, it may look confusing at first.

gramalingam: I think @p-wysocki means replacing `float` by something like `{double, float, float16, bfloat16}` ... otherwise, a float-to-float conversion sounds confusing (if interpreted as float32).

yuanyao-nv: @gramalingam I see. How about we say "floating point" instead of "float". This should make it clear we are referring to all floating point formats.

gramalingam: Fine with me. If so, we could use fixed point instead of int, uint.

docs/Changelog.md
```diff
@@ -15023,6 +15023,22 @@ This version of the operator has been available since version 13 of the default
...
+    * int, uint: undefined if OOR.
+    * bool: +/- 0.0 to False; all else to True.
+  * Casting from int, uint to:
+    * float: +/- infinity if OOR. (+ infinity in the case of uint)
+    * int, uint: when OOR, discard higher bits and reinterpret with respect to two's complement representation. For
```
gramalingam, line 15034: Thanks for the change. Minor nit: may be add "for signed types" after "two's complement representation", since it is not meant for unsigned.

docs/Changelog.md
yuanyao-nv: I changed them to "floating point" and "fixed point" as suggested.

docs/Changelog.md
yuanyao-nv: Updated.

gramalingam (approved):

postrational (approved):

jcwchen (commented):
Hi @yuanyao-nv,
I can repro the git diff failure from the CI. Did you use `sh tools\update_doc.sh` to update `Operators.md` and `Changelog.md` instead of updating those docs manually?

yuanyao-nv:
> Hi @yuanyao-nv, I can repro the git diff failure from the CI. Did you use `sh tools\update_doc.sh` to update `Operators.md` and "Changelog.md" instead of updating those docs manually?

@jcwchen Thanks. Is it requirement to use `update_doc.sh`? For small changes I just edited manually. Let me try the script.

jcwchen (approved):

PR #4361, opened 2022-07-14
Set up black and isort

justinchuby:
**Description**

- Set up configurations for black and isort. The actual formatting will be in the following pr.
- Create github action to enforce style
- Fix CI errors when having a pyproject.toml file 
- Add lint dependencies to requirements-dev 

**Motivation and Context**

Standardize python code format

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,10 @@
+[tool.black]
+# extend-exclude needs to be a regular expression
+extend-exclude = "third_party/"
+target-version = ["py37", "py38", "py39", "py310"]
+
+[tool.isort]
+profile = "black"
+extend_skip_glob = [
+    "third_party/*",
+]
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,6 +9,7 @@ jupyter
 pyzmq
 setuptools
 twine
+# Dependencies for linting. Versions match those in setup.py.
 flake8
 clang-format==13.0.0
 mypy==0.782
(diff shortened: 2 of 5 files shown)


jcwchen:
Reopen this PR to trigger CI

requirements-dev.txt
```diff
@@ -8,4 +8,8 @@ wheel
 jupyter
 pyzmq
 setuptools
-twine
\ No newline at end of file
+twine
+flake8
+clang-format==13.0.0
+mypy==0.782
+types-protobuf==3.18.4
```
jcwchen, line 15: Shall we add a comment saying that the formatter versions here are corresponding to the versions in setup.py?

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[tool.black]
+# extend-exclude needs to be a regular expression
+extend-exclude = "third_party/"
+target-version = ["py37", "py38", "py39", "py310"]
```
jcwchen, line 4: Just out of curiosity: is the target-version for Python here required? I just would like to reduce the duplicate work to add/remove Python versions in a lot of places if possible.

justinchuby: It not specified, black will assume the python version it is executed in, which may affect compatibility.

requirements-dev.txt
justinchuby: Done

jcwchen (approved):

Issue #4192, opened 2022-05-07
how to save 4bit/2bit models?


xiguadong:
# Ask a Question

### Question
I have a QAT quantized model , but the type of weight is 4bit/2bit. Could  I save it to onnx format? I found that onnx.proto only support type as belows:
![image](https://user-images.githubusercontent.com/55774832/167242388-37dcce97-88bc-43c6-b23b-fcf1e0d2c452.png)

 https://github.com/onnx/onnx/blob/5b1346e30d66e4ec550f6b63c3883b258a2e8e3e/onnx/onnx.proto#L484

jcwchen:
Hi @xiguadong,
You are right -- as you can see in the proto definition, currently 4 bit/2 bit models are not supported.

xiguadong:
so the next question is:
how do I support the 4/2bit? could you tell me the steps? 
thk u very much

jcwchen:
If you are looking for a future official support, this proposal will need to be reviewed by onnx/sig-archinfra and onnx/sig-operators since it's a fundamental change in IR. Then the frontend converter and backend runtime will start to support it. The process will take quite a while. May I understand more about your use case?

To sort of workaround it, you can try to modify ONNX locally and build it from source. An example PR for adding bfloat16: https://github.com/onnx/onnx/pull/1421. However, please note that probably most runtimes for ONNX have not supported 2 bit/4 bit models yet for now since such a type is not in the IR.

xiguadong:
I work for a npu chip company, we have some 2bit/4bit models runs on our chip. We decide to use onnx format save the 2bit/4bit cause the format we used before performs bad. But onnx is very  beautiful and easy to use!  And we can test the new spec soon if official support it.
It seems like 2bit/4bit has been a trend. I noticed that openvino/cuda/torch has support it. And , I guess onnxruntime also support 4bit quantization although save it with 8bit https://github.com/microsoft/onnxruntime/blob/207ad7eef9f91f88312cc28b2528bdc6362eaadb/onnxruntime/python/tools/quantization/onnx_quantizer.py#L51
Btw, we also plan to push our engine to github this week, [this](https://github.com/AXERA-TECH) is our repo（ :）its empty now). And we could send you  some interesting demo board in the future. 

Thanks !

pauldog:
Yes please support 4bit, 2bit and 1bit.

maziyarpanahi:
[...]

PR #4363, opened 2022-07-15
add dtype

bzhang3:
**Description**
- Add low bit data type into onnx.proto (TensorProto Dtype)
- u4/s4/u2/s2

**Motivation and Context**
- In future, more models need to be quantized as low bit model in edge industry field. Engineers need to get data type from tensorproto, however, current proto couldnot support this scenery. This pr would not change data storage method which can be packed in int32_data 

related issues:
https://github.com/onnx/onnx/issues/4192


daquexian:
I believe it's a big change. Adding a new dtype may impact a lot of ops. Related discussion: https://github.com/onnx/onnx/pull/2925.

Could you please introduce the expected usage of int4/int2 dtype? For example, which ops should support int4/int2 to support your use case?

AlexandreEichenberger:
Question: do you intend to be only a data format operation, or the intent is to have many of the current operators to support this new format? Also, is the behavior of u/s 4/2 well defined with respect to saturation?

bzhang3:
@daquexian  the int4/int2 dtype would be used in edge inference framework (low bits quant model). These data formats should be displayed by low bit dtype in tensor info which help engineer to take right data info from tensor. For now, I can only take data info as u8, s8 (all data can be saved in int32_data). If the u4/s4 or u2/s2 used in model storage, users may not take right data format info because u4/s4, u2/s2 are not in tensor info proto. The ops could be Conv, Pool or any other ops.

bzhang3:
@AlexandreEichenberger I  think this data format should be supported by currents operators, since if model uses low bits format, the inference framework should know right data format to do the next action such like computing or malloc, etc. 
The saturation of u/s 4/2 could follow u/s 8's saturation method (like -15 ~ 16 or 0 ~ 31 )

daquexian:
> The ops could be Conv, Pool or any other ops.

@bzhang3 @AlexandreEichenberger I think the support for low-bit quantization is of importance. However, is it ok to update QDQ operators only and keep other operators unchanged? For example, after adding a `bit_width` attribute to QDQ, a 4-bit conv can be represented by
```
input (fp32) -> Q (bit_width=4) -> DQ (bit_width=4) -> Conv -> output (fp32)
                 weight (4 bit integer) -> DQ (bit_width=4)  --/
```

bzhang3:
@daquexian I think this is a reasonable solution for this change. We do not need modify operator's attribute, and only need to modify TensorProto as data signature

daquexian:
> We do not need modify operator's attribute

By operator's attribute you mean the attribute of other operators (conv, matmul, ...) or the attribute of QDQ?

bzhang3:
@daquexian 
the attribute of other operators (conv, matmul, ...)

maltanar:
Hi! Happy to see the interest around <8-bit quantized models. The FINN and hls4ml communities have been using ONNX custom ops & annotations to represent few-bit quantized NNs for a while, we're working on a joint response that I will share shortly.

gramalingam:
As discussed above, a key point is to figure out the ops that are impacted by the change. Aas @daquexian suggests, it would be better if only a few ops (the QDQ ops) need to be modified to support this, instead of many ops. 

I don't understand @bzhang3's comment about "only need to modify TensorProto as data signature" ? I guess you mean the change in this PR?

One possibility is that we specify how the bit values are packed into an uint8 (as part of the Quantize op's spec) and we don't even need to introduce the new types. This seems simpler. (The only possible disadvantage is if different hardware prefers to pack the bits in different orders.)

maltanar:
> One possibility is that we specify how the bit values are packed into an uint8 (as part of the Quantize op's spec) and we don't even need to introduce the new types. This seems simpler. (The only possible disadvantage is if different hardware prefers to pack the bits in different orders.)

@gramalingam I believe this PR proposes to use `int32/uint32` to store the 2- and 4-bit values, and does not specify any packing to save storage space. In a way, this is using the native `int32/uint32` types as a container, and indicating that they actually contain 2/4-bit values.

```
STORAGE_TENSOR_TYPE_TO_FIELD = {
   # ...
    int(TensorProto.BOOL): 'int32_data',
    int(TensorProto.UINT4): 'int32_data',
    int(TensorProto.INT4): 'int32_data',
    int(TensorProto.UINT2): 'int32_data',
    int(TensorProto.INT2): 'int32_data',
}
```

Since these dtypes don't natively exist on most CPUs or in numpy for Python bindings, this approach avoids the complexity of having to deal with various headaches (packing order, alignment/bit padding, packing/unpacking functions...) in ONNX.

bzhang3:
>

bzhang3:
> > One possibility is that we specify how the bit values are packed into an uint8 (as part of the Quantize op's spec) and we don't even need to introduce the new types. This seems simpler. (The only possible disadvantage is if different hardware prefers to pack the bits in different orders.)
> 
> @gramalingam I believe this PR proposes to use `int32/uint32` to store the 2- and 4-bit values, and does not specify any packing to save storage space. In a way, this is using the native `int32/uint32` types as a container, and indicating that they actually contain 2/4-bit values.
> 
> ```
> STORAGE_TENSOR_TYPE_TO_FIELD = {
>    # ...
>     int(TensorProto.BOOL): 'int32_data',
>     int(TensorProto.UINT4): 'int32_data',
>     int(TensorProto.INT4): 'int32_data',
>     int(TensorProto.UINT2): 'int32_data',
>     int(TensorProto.INT2): 'int32_data',
> }
> ```
> 
> Since these dtypes don't natively exist on most CPUs or in numpy for Python bindings, this approach avoids the complexity of having to deal with various headaches (packing order, alignment/bit padding, packing/unpacking functions...) in ONNX.

That was my mistake. The proposal for this pr should similar to UINT8/INT8's storage method. The above codes is not right modification. Until now, I want to find possible solution for this situation through this discussion and then modify code.

What I want to do is to store the low-bit data in the high-bit data space such as int32_data, and then tell the user through the TensorProto file that the data type in the high-bit data space (int32_data) is 4/2 type

yufenglee:
ONNX doesn't have a direct quantized tensor definition. Essentially, it uses QDQ to represent a quantized tensor. Thus we can limit the change to Q/DQ operators only as @daquexian and @gramalingam suggested and we don't really need to add sub-byte data type. For Q/DQ operators, we still keep their input/output type as it is now but add 2 attributes: one for bit width like suggested and the other for packing mode. 

For packing mode, we can simply support 2 mode: no_packing, continuous. If no_packing, a sub-byte is represented with byte. For continous, sub-byte is packed continuously, for example, we use 2 bytes to represent 4 3-bit values like below. The last 4 bits are used for padding.
image

xiguadong:
> ONNX doesn't have a direct quantized tensor definition. Essentially, it uses QDQ to represent a quantized tensor. Thus we can limit the change to Q/DQ operators only as @daquexian and @gramalingam suggested and we don't really need to add sub-byte data type. For Q/DQ operators, we still keep their input/output type as it is now but add 2 attributes: one for bit width like suggested and the other for packing mode.
> 
> For packing mode, we can simply support 2 mode: no_packing, continuous. If no_packing, a sub-byte is represented with byte. For continous, sub-byte is packed continuously, for example, we use 2 bytes to represent 4 3-bit values like below. The last 4 bits are used for padding. image

Hi, thanks for your reply. I think onnx has another representation for quantized ops like [QlinearConv](https://github.com/microsoft/onnxruntime/blob/207ad7eef9f91f88312cc28b2528bdc6362eaadb/onnxruntime/python/tools/quantization/operators/conv.py#L125)(I found in this link[all ops](https://github.com/microsoft/onnxruntime/blob/207ad7eef9f91f88312cc28b2528bdc6362eaadb/onnxruntime/python/tools/quantization/operators/)). And I guess we can't use 2/4 bit in this format directly . 
Also ,I found workaround for 4bit in [QlinearConv](https://github.com/microsoft/onnxruntime/blob/207ad7eef9f91f88312cc28b2528bdc6362eaadb/onnxruntime/python/tools/quantization/onnx_quantizer.py#L51).   As @bzhang3's comment about  "only need to modify TensorProto as data signature", if we want use the format like QlinearConv ,we only need to change the dtype of the output which response to the tensor in graph.initializer. We don't need to define the attributes such as output_dtype/reduce_range etc. Also , I think quantized repsetation(QlinearConv) is more beautiful than QDQ mode cause it could more eaiser to understand the network's structure.

maltanar:
I think we would benefit from having a more structured discussion to consider the various aspects of representing quantization that are brought up in this thread. These aspects are related and they affect each other, but there is benefit in considering them separately. 

Here is my attempt to structure this into three different aspects:

1) **Quantization function:** How is quantization (restricting the set of values each element of a tensor can take) described as an operator in ONNX? From a clean-abstraction perspective, I think having a standalone quantization operator (like QDQ, or [Quant](https://github.com/fastmachinelearning/qonnx/blob/main/docs/qonnx-custom-ops/quant_op.md) in the QONNX proposal) is preferable over having a quantized variant of each standard operator (like QuantConv, QuantLinear etc). In the former case, one can combine the quantization op with any other op in ONNX to obtain a quantized variant of that operator. In the latter case, we need a separate op type in ONNX for each new quantized op type (e.g. in addition to QuantConv and QuantLinear we would also need QuantAdd, QuantMul, QuantGather and so on). 

2) **Datatype indication:** How does a framework or developer know that a given tensor contains quantized values, how many bits each element needs for representation, whether the values are signed or unsigned, whether they represent integer-only quantization, fixed-point (scale factor is a power of two) or even non-uniform quantized values?  Note that it is possible to separate the datatype indication from the storage and packing aspects (point 3 below), e.g. I could have a tensor where each value is stored using uint8 but I know that each value is really uint2. In the current ONNX specification I believe there is a possibility to [store Boolean data as one byte per bit](https://github.com/onnx/onnx/blob/main/onnx/onnx.proto#L574).

3) **In-memory layout/storage and manipulation:** Assuming each element of a quantized tensor can be represented with k bits, how do we store this tensor in memory and on disk as part of the ONNX standard? When accessing the value of this tensor through the protobuf or Python API, do we return packed data or do we transparently expand it into some native type? If packing, do we assume the equivalent of high-endian or low-endian? Do we zero-pad to achieve a multiple of 8-bit, 32-bit or some other native word size? Do we pad with zeros on the upper or lower endian side? Note that k can be arbitrary in the general case: it could be <8-bit for highly-compressed networks, or it can be something like 11-bit or 13-bit for applications using elementary fixed-point quantization methods for post-training quantization that only looks at the max values of a given tensor.

xiguadong:
> I think we would benefit from having a more structured discussion to consider the various aspects of representing quantization that are brought up in this thread. These aspects are related and they affect each other, but there is benefit in considering them separately.
> 
> Here is my attempt to structure this into three different aspects:
> 
> 1. **Quantization function:** How is quantization (restricting the set of values each element of a tensor can take) described as an operator in ONNX? From a clean-abstraction perspective, I think having a standalone quantization operator (like QDQ, or [Quant](https://github.com/fastmachinelearning/qonnx/blob/main/docs/qonnx-custom-ops/quant_op.md) in the QONNX proposal) is preferable over having a quantized variant of each standard operator (like QuantConv, QuantLinear etc). In the former case, one can combine the quantization op with any other op in ONNX to obtain a quantized variant of that operator. In the latter case, we need a separate op type in ONNX for each new quantized op type (e.g. in addition to QuantConv and QuantLinear we would also need QuantAdd, QuantMul, QuantGather and so on).
> 2. **Datatype indication:** How does a framework or developer know that a given tensor contains quantized values, how many bits each element needs for representation, whether the values are signed or unsigned, whether they represent integer-only quantization, fixed-point (scale factor is a power of two) or even non-uniform quantized values?  Note that it is possible to separate the datatype indication from the storage and packing aspects (point 3 below), e.g. I could have a tensor where each value is stored using uint8 but I know that each value is really uint2. In the current ONNX specification I believe there is a possibility to [store Boolean data as one byte per bit](https://github.com/onnx/onnx/blob/main/onnx/onnx.proto#L574).
> 3. **In-memory layout/storage and manipulation:** Assuming each element of a quantized tensor can be represented with k bits, how do we store this tensor in memory and on disk as part of the ONNX standard? When accessing the value of this tensor through the protobuf or Python API, do we return packed data or do we transparently expand it into some native type? If packing, do we assume the equivalent of high-endian or low-endian? Do we zero-pad to achieve a multiple of 8-bit, 32-bit or some other native word size? Do we pad with zeros on the upper or lower endian side? Note that k can be arbitrary in the general case: it could be <8-bit for highly-compressed networks, or it can be something like 11-bit or 13-bit for applications using elementary fixed-point quantization methods for post-training quantization that only looks at the max values of a given tensor.

Thanks, the structure is very very very clear. 
To conclude:
We all agree: **We need a representation of tensor which has dtype of 2/4 bit.**  [Quant](https://github.com/fastmachinelearning/qonnx/blob/main/docs/qonnx-custom-ops/quant_op.md) is also a wonderful solution in QDQ ways. 

But the topic of this pr is about : **Do we really need 2/4 bit in onnx.proto?** 

My answer is **YES**.  For simply, I make some difinitions:
- Qops: quant ops like QlinearConv, QlinearAdd, QlinearMul, QlinearGather etc.
- Qops IR : IR which coustruct with Qops and float ops.
- QDQ IR : IR which coustruct with Q/DQ ops and float ops.
 
1st , Qops IR can't represent 2/4 bit directly until now. As my mention before, [QlinearConv](https://github.com/microsoft/onnxruntime/blob/207ad7eef9f91f88312cc28b2528bdc6362eaadb/onnxruntime/python/tools/quantization/onnx_quantizer.py#L51) support 4bit by workaround. There are still a lot of ops like QlinearGemm/QlinearMatMul lack 2/4 bit representation.

2nd, Qops is a important part of ONNX. Qops IR is simply for deploying and it's stable representation. ONNX has a huge opset that support almost all ops of other training frameworks. But it's a big chanllenge for inference frameworks, we must analysis ervery case of every ops. Buf for quantazation, we only use a subset of ONNX ops and we want it to be stable such like the defintion of [TFLITE quantization spec](https://www.tensorflow.org/lite/performance/quantization_spec). It's easier to analysis for inferece frameworks. 

3rd, as using int32 as save format of uint8, maybe we can save the 2/4bit as int32. For raw data, maybe we can package the 2/4 bit to uint8 and save them as uint8.

So, I think the requirement is real and I hope we could discuss about the 3rd point above to promote this pr.
Thanks.

maltanar:
@xiguadong thanks, and agreed. Regarding your 3rd point, in this sentence:

> 3rd, as using int32 as save format of uint8, maybe we can save the 2/4bit as int32. For raw data, maybe we can package the 2/4 bit to uint8 and save them as uint8.

Could you clarify the difference between the int32 save part and the uint8 raw data part here? Not sure if I understood this.

pauldog:
Please support saving quantized models into 4bit, 2bit and 1bit. 
1) To make smaller file sizes
2) To support quantized models.

justinchuby:
Discussions moved to 
- https://github.com/onnx/onnx/issues/6900

PR #4365, opened 2022-07-18
Test lower protoc version in Linux CI

jcwchen:
**Description**
- Test lower protoc version in Linux CI
- Modify two protobuf usage to make ONNX be compatible with lower protoc version
- Improve script style for manylinux/entrypoint.sh
- Make Linux CI of Azp test ONNX_USE_LITE_PROTO=OFF and Linux CI of GitHub Action test ONNX_USE_LITE_PROTO=ON, because lower protoc version from apt-get cannot work with ONNX_USE_LITE_PROTO=ON.
- Sync option settings for all released wheels: same as Windows release CI, Linux and Mac release CI will use ONNX_USE_LITE_PROTO=ON as well. It can shrink binary size a little bit.

**Motivation and Context**
For now there is no CI in ONNX testing lower protoc version. For instance, using libprotobuf-dev and protobuf-compiler from apt-get. ONNX mention this usage in the document and ideally the CI should cover this case.


gramalingam (approved):

jcwchen:
One important change I would like to highlight is that this PR makes released ONNX wheel for Linux and Mac start to use `ONNX_USE_LITE_PROTO=ON`. The released wheel for Windows has already used it for quite a while (since ONNX 1.7.0) and so I think it's good to sync the option settings for every released wheel to prevent confusions. Also it can save some binary size (~0.5 MB).

It's possible that Linux users use older protoc versions (e.g., 3.0.0) from apt-get. Although using prebuilt wheel should be a fine case, to be safe, I tested the ONNX Linux wheel, which was built with `ONNX_USE_LITE_PROTO=ON` from this PR, on Ubuntu 18.04 with protoc 3.0.0 and it did work well without any issue.

PR #4366, opened 2022-07-18
Add detailed error message for saving 2GB proto

jcwchen:
**Description**
Add detailed error message for saving 2GB proto.

**Motivation and Context**
Currently saving 2GB proto will simply get complaint from Protobuf.  To improve user experience, ONNX should suggest a possible solution in the error message.

cc @BowenBao not sure whether it is the right improvement you are looking for. Please review it. Thanks!

### DIFF
--- a/onnx/__init__.py
+++ b/onnx/__init__.py
@@ -73,7 +73,7 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
         except:
             raise
         return result
-    raise TypeError(f'No SerializeToString method is detected. Neither proto is a str.\ntype is {type(proto)}')
+    raise TypeError(f"No SerializeToString method is detected. Neither proto is a str.\ntype is {type(proto)}")
 
 
 _Proto = TypeVar('_Proto', bound=google.protobuf.message.Message)
@@ -91,10 +91,10 @@ def _deserialize(s: bytes, proto: _Proto) -> _Proto:
         The proto instance filled in by s
     '''
     if not isinstance(s, bytes):
-        raise ValueError(f'Parameter s must be bytes, but got type: {type(s)}')
+        raise ValueError(f"Parameter s must be bytes, but got type: {type(s)}")
 
     if not (hasattr(proto, 'ParseFromString') and callable(proto.ParseFromString)):
-        raise ValueError(f'No ParseFromString method is detected. Type is {type(proto)}')
+        raise ValueError(f"No ParseFromString method is detected. Type is {type(proto)}")
 
     decoded = cast(Optional[int], proto.ParseFromString(s))
     if decoded is not None and decoded != len(s):
--- a/onnx/__init__.py
+++ b/onnx/__init__.py
@@ -63,7 +63,12 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
     if isinstance(proto, bytes):
         return proto
     elif hasattr(proto, 'SerializeToString') and callable(proto.SerializeToString):
-        result = proto.SerializeToString()
+        try:
+            result = proto.SerializeToString()
+        except:
+            if proto.ByteSize() >= 2147483648:
+                print("The single proto is larger than 2GB. Please use save_as_external_data to save proto separately.")
+            raise
         return result
     else:
         raise TypeError('No SerializeToString method is detected. '
--- a/onnx/__init__.py
+++ b/onnx/__init__.py
@@ -67,7 +67,8 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
             result = proto.SerializeToString()
         except ValueError as e:
             if proto.ByteSize() >= onnx.checker.MAXIMUM_PROTOBUF:
-                raise ValueError("The single proto is larger than 2GB. Please use save_as_external_data to save proto separately.") from e
+                raise ValueError("The proto size is larger than the 2 GB limit. "
+                    "Please use save_as_external_data to save tensors separately from the model file.") from e
             raise
         return result
     raise TypeError(f"No SerializeToString method is detected. Neither proto is a str.\ntype is {type(proto)}")
--- a/onnx/__init__.py
+++ b/onnx/__init__.py
@@ -70,8 +70,6 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
                 raise ValueError("The single proto is larger than 2GB. Please use save_as_external_data to save proto separately.") from e
             else:
                 raise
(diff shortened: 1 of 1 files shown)


onnx/__init__.py
```diff
@@ -62,12 +62,15 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
     '''
     if isinstance(proto, bytes):
         return proto
-    elif hasattr(proto, 'SerializeToString') and callable(proto.SerializeToString):
-        result = proto.SerializeToString()
+    if hasattr(proto, 'SerializeToString') and callable(proto.SerializeToString):
+        try:
+            result = proto.SerializeToString()
+        except:
```
justinchuby, line 68: I would catch a more specific exception

onnx/__init__.py
```diff
@@ -62,12 +62,15 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
...
+            if proto.ByteSize() >= 2147483648:
+                print("The single proto is larger than 2GB. Please use save_as_external_data to save proto separately.")
```
justinchuby, line 70: Possible to do `raise ... from ...`?

onnx/__init__.py
```diff
@@ -62,12 +62,18 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
...
+    if hasattr(proto, 'SerializeToString') and callable(proto.SerializeToString):
+        try:
+            result = proto.SerializeToString()
+        except ValueError as e:
+            if proto.ByteSize() >= onnx.checker.MAXIMUM_PROTOBUF:
+                raise ValueError("The single proto is larger than 2GB. Please use save_as_external_data to save proto separately.") from e
+            else:
+                raise
+        except:
+            raise
```
justinchuby, line 74:
```suggestion
```

Fine to not have this branch since errors not caught are raised

onnx/__init__.py
```diff
@@ -62,12 +62,15 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
...
-    elif hasattr(proto, 'SerializeToString') and callable(proto.SerializeToString):
-        result = proto.SerializeToString()
+    if hasattr(proto, 'SerializeToString') and callable(proto.SerializeToString):
+        try:
+            result = proto.SerializeToString()
+        except ValueError as e:
+            if proto.ByteSize() >= onnx.checker.MAXIMUM_PROTOBUF:
+                raise ValueError("The single proto is larger than 2GB. Please use save_as_external_data to save proto separately.") from e
+            raise
         return result
```
jcwchen, line 72: @justinchuby I found we have different mypy version between required check (from style.sh) and Python Lint. That's why this line only passed by style.sh (mypy 0.782) but failed with Python Lint (mypy 0.971). To prevent confusion, is it possible to use mypy 0.782 from reviewdog for now? I believe many of existing ONNX code cannot pass by the latest mypy.

If it is not feasible, do you think we should disable no-any-return in Python Lint's mypy? It seems to complain here due to missing type specified from Protobuf's SerializeToString, which will be represented as Any type.

onnx/__init__.py
jcwchen: All updated. Thanks for the reviews. Please review it again.

onnx/__init__.py
justinchuby: It doesn't seem the check supports this at the moment. disabling no-any-return sounds like a good idea.

justinchuby (approved):
LGTM🚀

onnx/__init__.py
```diff
@@ -62,12 +62,15 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
...
     if isinstance(proto, bytes):
         return proto
-    elif hasattr(proto, 'SerializeToString') and callable(proto.SerializeToString):
-        result = proto.SerializeToString()
+    if hasattr(proto, 'SerializeToString') and callable(proto.SerializeToString):
+        try:
+            result = proto.SerializeToString()
+        except ValueError as e:
+            if proto.ByteSize() >= onnx.checker.MAXIMUM_PROTOBUF:
+                raise ValueError("The single proto is larger than 2GB. Please use save_as_external_data to save proto separately.") from e
```
gramalingam, line 70: Suggest
```
The proto size is larger than the 2 GB limit. Please use save_as_external_data to save tensors
separately from the model file.
```

onnx/__init__.py
gramalingam: You could try changing `result = proto.SerializeToString()' to `return proto.SerializeToString()` and delete the `return result` to see if it helps.

gramalingam: Or, assign an empty string to `result` at the beginning. I wonder if it is just an imprecision in the analysis producing a false positive.

justinchuby: If you really want, I would do `cast(bytes, result` to explicitly make mypy happy. https://docs.python.org/3/library/typing.html#typing.cast

justinchuby: Mypy doesn't know about SerializeToString. It may if we compile the lib first and install all dependencies in the github action

jcwchen: Thank you both for all the suggestions, but unfortunately none of these can resolve the mypy error in two CIs. `cast(bytes, result)` will bring Redundant cast to "bytes" if mypy knows SerializeToString. 

> Mypy doesn't know about SerializeToString. It may if we compile the lib first and install all dependencies in the github action

Good idea. I think we should do that before mypy check to solve this issue globally. Especially I also saw mypy errors in other PRs like:
```
[mypy] reported by reviewdog 🐶 Library stubs not installed for "google.protobuf.message" (or incompatible with Python 3.8) [import] Raw Output: /home/runner/work/onnx/onnx/onnx/helper.py:8:1: error: Library stubs not installed for "google.protobuf.message" (or incompatible with Python 3.8) [import]
```
and
```
[mypy] reported by reviewdog 🐶 Module "onnx" has no attribute "OptionalProto"; maybe "Optional"? [attr-defined] Raw Output: /home/runner/work/onnx/onnx/onnx/mapping.py:5:1: error: Module "onnx" has no attribute "OptionalProto"; maybe "Optional"? [attr-defined]
```

@justinchuby do you have bandwidth to take care of this in added mypy check from Python Lint? Definitely it can help to resolve many unneccessary mypy errors in every PR. Thanks!

gramalingam (approved):

onnx/__init__.py
justinchuby: I may be able to pick this up early next week if that works.

PR #4369, opened 2022-07-20
Fix TopK docs typo

CharlieL7:
Signed-off-by: charlie <charlie.lin@amd.com>

**Description**
- Remove `r` typo from TopK operator docs

**Motivation and Context**
- Does not look like `r` of input shape is used or correct.


andife:
Thank you for your contribution. The answer is not related to TopK in special.

For your information:
The Operators.md is generated by executing https://github.com/onnx/onnx/blob/main/tools/update_doc.sh after changing https://github.com/onnx/onnx/blob/main/onnx/defs/math/defs.cc

PR #4371, opened 2022-07-22
Remove pyflakes from CI

justinchuby:
**Description**

pyflakes does not support diff_context, creating too much noise. So we disable it.

- Set diff_context in all other checks
- We do not need to reenable pyflakes because flake8 is supposed to include pyflakes. We just need to configure properly

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -21,13 +21,7 @@ jobs:
           # Change reporter level if you need.
           # GitHub Status Check won't become failure with a warning.
           level: error
-          filter_mode: file
-      - name: pyflakes
-        uses: reviewdog/action-pyflakes@v1
-        with:
-          github_token: ${{ secrets.github_token }}
-          reporter: github-pr-check
-          level: warning
+          filter_mode: diff_context
       - name: misspell # Check spellings as well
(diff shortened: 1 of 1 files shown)


justinchuby:
@jcwchen

jcwchen (approved):
LGTM. Thank you for the updates! ~In the future we can add pyflakes back after it has supported diff_context.~

justinchuby:
We prob don't need to reenable pyflakes because flake8 is supposed to include pyflakes. We just need to configure flake8 properly (to enable those checks)

jcwchen:
> We prob don't need to reenable pyflakes because flake8 is supposed to include pyflakes. We just need to configure flake8 properly (to enable those checks)

That's good to know. I thought pyflakes involves other linter as well (https://github.com/reviewdog/action-pyflakes#lint---reviewdog-integration), but actually ONNX Python Lint has covered some of them already: action-shellcheck and action-misspell. Then yes we don't need to reenable it. Updated my old comment.

justinchuby:
I think that's just a templated readme that's not updated

jcwchen:
> I think that's just a templated readme that's not updated

You are right 👍

PR #4372, opened 2022-07-22
Extend printer to handle non-empty domain

gramalingam:

**Description**

The printer does not currently handle a non-empty domain (used for ops in domains other than the standard one).
Extend the printer to handle it.


onnx/defs/printer.cc
```diff
@@ -319,7 +319,10 @@ void ProtoPrinter::print(const AttrList& attrlist) {
 void ProtoPrinter::print(const NodeProto& node) {
   output_ << std::setw(indent_level) << ' ';
   printSet("", ", ", "", node.output());
-  output_ << " = " << node.op_type();
+  output_ << " = ";
+  if (node.domain() != "")
+    output_ << node.domain() << ".";
```
daquexian, line 324: "ai.onnx" and "" both mean the default domain. Should ops with domain "ai.onnx" have consistent behavior with those with "" domain?

gramalingam: yes, that's right. (Does it have any implication to this PR?)

daquexian: > Does it have any implication to this PR?

Yes, maybe the line 323 should be changed to `if (node.domain() != "" && node.domain() != "ai.onnx")`

gramalingam: I think it is better in the current form. Both forms are legal and mean the same thing. The parser/printer are just convenience utilities to convert the proto format to/from a readable textual form. It is useful to preserve the distinction between the two different, but equivalent, forms in the parser/printer.

linkerzhang (approved):
LGTM.

wschin (approved):

PR #4377, opened 2022-07-25
Use fully qualified pathname when loading DLL to prevent security vulnerability

jcwchen:
**Description**
To prevent security vulnerability, we should use fully qualified pathname when loading DLL.

**Motivation and Context**
Security vulnerability: The LoadLibrary*() family of functions are used to dynamically load a DLL into your
program. Most versions of the Windows operating system search the current directory first
when attempting to load files. If you do not specify a fully qualified pathname when loading
a Windows dynamic load library (DLL), the loader will attempt to locate the library in the
current directory. If there is a malicious library located in the current directory, it will be
loaded instead of the intended library.


AlexandreEichenberger (approved):
LGTM, not sure if the default `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` is still needed as a fully qualified path is used. But maybe its used for something else... your call.

AlexandreEichenberger:
Can an approved accept this PR as it is a potential vulnerability. Thanks

jcwchen:
cc @onnx/sig-archinfra-approvers PTAL. Thanks!

jcwchen:
> LGTM, not sure if the default `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` is still needed as a fully qualified path is used. But maybe its used for something else... your call.

Thanks for catching this. `LOAD_WITH_ALTERED_SEARCH_PATH` seems more accurate for absolute path and so I will use it instead.

liqunfu (approved):

gramalingam (approved):

PR #4385, opened 2022-07-27
save as external data automatically

daquexian:
**Description**
Following https://github.com/pytorch/pytorch/pull/64382, automatically determine the value of save_as_external_data according to whether the proto byte size is greater than 2GB.

**Motivation and Context**
It is inconvenient if save_as_external_data needs to be determined manually. Without this PR, users must write the following complicated code just to save a model without unexpected error:

    ```python
    try:
      onnx.save(model, f)
    except ValueError:
      onnx.save(model, f, save_as_external_data=True)
    ```


daquexian:
To reviewers: I'm afraid the newly added test cannot run in CI machine due to the memory limit. Is there any advice? Thanks!

jcwchen (commented):
> To reviewers: I'm afraid the newly added test cannot run in CI machine due to the memory limit. Is there any advice? Thanks!

Previously I did write a similar test about loading 2GB proto in ONNX, but I also bumped into some issues dynamically in CIs due to different hardware resource and eventually I decided to remove them.

To solve your question, perhaps we can use `psutil.virtual_memory().available` to detect available memory and `@pytest.mark.skipif` to test it conditionally?

daquexian:
> To solve your question, perhaps we can use psutil.virtual_memory().available to detect available memory and @pytest.mark.skipif to test it conditionally?

It makes sense. However, I'm afraid it introduces a new required package (psutil) for testing. What about changing the line

```python
save_as_external_data = proto.ByteSize() > 2 ** 31
```

to 

```python
large_model_threshold = os.getenv("ONNX_SAVING_LARGE_MODEL_THRESHOLD", 2 ** 31)
save_as_external_data = proto.ByteSize() > large_model_threshold
```

And set ONNX_SAVING_LARGE_MODEL_THRESHOLD to a small number programmatically in test:
```python
def test_save_model_as_external_data_automatically(self) -> None:
  ...
  os.environ["ONNX_SAVING_LARGE_MODEL_THRESHOLD"] = "500000"
  onnx.save_model(model, model_file_path)
  del os.environ["ONNX_SAVING_LARGE_MODEL_THRESHOLD"]
  ...
```

jcwchen:
> It makes sense. However, I'm afraid it introduces a new required package (psutil). What about changing the line

Good catch. Didn't realize it's not included in standard Python.

> And programmatically set ONNX_SAVING_LARGE_MODEL_THRESHOLD to a small number in the test

Thank you for the proposal and yes probably it the best option we can get.

jcwchen (commented):
Please ignore the mypy failures about undefined things under onnx from Python Lint CI for now. ([more context](https://github.com/onnx/onnx/pull/4366/files#r926068941)) Sorry for the confusion.

daquexian:
@jcwchen thanks for your review, updated :)

onnx/test/test_external_data.py
```diff
@@ -368,6 +368,51 @@ def test_save_model_without_loading_external_data(self) -> None:
         self.assertFalse(attribute_tensor.HasField("data_location"))
         self.assertTrue(np.allclose(to_array(attribute_tensor), self.attribute_value))
 
+    def test_save_model_as_external_data_automatically(self) -> None:
```
liqunfu, line 371: Please consider using onnx model script https://github.com/onnx/onnx/blob/main/onnx/test/parser_test.py for better readability.

onnx/__init__.py
```diff
@@ -186,6 +188,11 @@ def save_model(proto: Union[ModelProto, bytes], f: Union[IO[bytes], str], format
     '''
     if isinstance(proto, bytes):
         proto = _deserialize(proto, ModelProto())
+    if save_as_external_data is None:
+        # Save to external data automatically if the model is larger than
+        # ONNX_SAVING_AS_EXTERNAL_DATA_THRESHOLD (default 2GB)
+        threshold = int(os.getenv("ONNX_SAVING_AS_EXTERNAL_DATA_THRESHOLD", 2 ** 31))
```
jcwchen, line 194:
```suggestion
        threshold = int(os.getenv("ONNX_SAVING_AS_EXTERNAL_DATA_THRESHOLD", onnx.checker.MAXIMUM_PROTOBUF))
```
I made onnx.checker.MAXIMUM_PROTOBUF as 2000000000 for lower boundary to detect possible 2GB exceed error. It seems fine to use 2 ** 31 directly: https://github.com/protocolbuffers/protobuf/blob/450d24ca820750c5db5112a6f0b0c2efb9758021/python/google/protobuf/pyext/message.cc#L1693. Feel free to update the number for onnx.checker.MAXIMUM_PROTOBUF. Thanks!

jcwchen (approved):
Thank you for making save_model more intelligent!

Some nit comments:
1. Not sure whether we should remove the warning introduced by this PR: https://github.com/onnx/onnx/pull/4366, but it might be still a useful warning if an user uses save_as_external_data=False intentionally.
2. Similar to this PR, I found typically users will easily see issues if using `all_tensors_to_one_file=True` by default (See [this issue](https://github.com/onnx/onnx/issues/4394)) because all initializers in total can easily be larger than 2GB nowadays. Perhaps we can also make save_model automatically turn off `all_tensors_to_one_file`. We can have another PR to improve it if needed.

onnx/test/test_external_data.py
daquexian: Thanks! Will update soon

onnx/__init__.py
daquexian: Thanks! I will update it soon.

gramalingam (approved):

PR #4386, opened 2022-07-27
Make C++ and Python check_model consistent

### LINKED ISSUE #3884: Make C++ and Python checker API consistent
Python checker API supports `full_check` arg:
https://github.com/onnx/onnx/blob/fa6f8cfdce3d86346e8a7494f3062b98416c85fb/onnx/checker.py#L94

C++ does not.
It'd be nice for them to be consistent.

sechkova:
**Description**
Add the `full_check` argument (enabling shape inference) to the `check_model` functions of the C++ API (checker.h/checker.cc).

**Motivation and Context**
Fixes #3884


onnx/checker.cc
```diff
@@ -910,9 +911,14 @@ void check_model(const ModelProto& model, CheckerContext& ctx) {
   if (ctx.get_ir_version() >= 0x00000008) {
     check_model_local_functions(model, ctx, lex_ctx);
   }
+
+  if (full_check) {
+    ShapeInferenceOptions options{true, 1, true};
```
jcwchen, line 916: To be consistent with previous behavior: data_prop was not enabled.
```suggestion
    ShapeInferenceOptions options{true, 1, false};
```

onnx/checker.h
```diff
@@ -142,8 +142,8 @@ void check_model_local_functions(
     const CheckerContext& ctx,
     const LexicalScopeContext& parent_lex);
 
-void check_model(const ModelProto& model);
-void check_model(const std::string& model_path);
+void check_model(ModelProto& model, bool full_check = false);
```
daquexian, line 145: `void check_model(const ModelProto& model)` -> `void check_model(ModelProto& model)` brings a breaking change for C++ users and also unnecessarily refuses const models when full checking is not needed. How about 

```suggestion
void check_model(const ModelProto& model);
void check_model(ModelProto& model, bool full_check);
```

sechkova: Thanks for the suggestion, not breaking the API is better in any case. I added new commits addressing the comments.

jcwchen: I have a concern: if users call checker with full_check, did they expect the ONNX model will get updated (shape_inference will bring inferred shapes in value_info)?

Current behavior:
- check_model+full_check with model path (Python and C++ API): the original model will be overrode by the inferred model thru model path (it's by default from onnx.shape_inference_path)
- check_model+full_check with ModelProto (Python and C++ API): the original model won't be updated

After this PR:
- check_model+full_check with model path (Python and C++ API): same as current behavior
- check_model+full_check with ModelProto (Python API): the original model won't be updated
- check_model+full_check with ModelProto (C++ API): the original ModelProto will be overrode by the inferred model

IMO, ideally all APIs behaviors for check_model should be consistent to prevent confusion. It seems weird to me that the ONNX model will be updated after onnx.checker with full_check. Still, it's an existing issue (check_model does not update the model, but check_model_path does) so it should not be the blocker to this PR. If both of you also think it's a valid concern, probably we can improve it in another PR.

gramalingam: Yes, it would be helpful to ensure a consistent and orthogonal API. Do we need full-check and const-ness of the model interact or can they be orthogonal/independent? It would also be a good idea to add some comment/documentation for all of this.

jcwchen (approved):
LGTM. Thank you for make them consistent! We can have another PR to fix existing inconsistency: https://github.com/onnx/onnx/pull/4386#discussion_r937159690.

Issue #4188, opened 2022-05-04
Remove Split op’s dependence on number of outputs


garymm:
It is desirable to make the behavior of ONNX ops fully determined by its input parameters as well as attributes. The Split op is one of the rare cases where this does not hold. The split op has an optional second parameter that indicates how the split should happen, e.g., into 3 parts of sizes 10, 10, and 4, for example. However, if this parameter is omitted, the current spec says that the split axis is evenly divided among the specified outputs. The behavior is, hence, dependent on the variadic number of outputs specified. It will be cleaner if, instead, we add a specific attribute, say num_outputs, to indicate how many equal parts that the split axis should be divided into.

gramalingam:
Let me clarify the intent here. It is desirable to make the behavior of ONNX ops fully determined by its input parameters as well as attributes. The Split op is one of the rare cases where this does not hold. The split op has an optional second parameter that indicates how the split should happen, e.g., into 3 parts of sizes 10, 10, and 4, for example. However, if this parameter is omitted, the current spec says that the split axis is evenly divided among the specified outputs. The behavior is, hence, dependent on the variadic number of outputs specified. It will be cleaner if, instead, we add a specific attribute, say `num_outputs`, to indicate how many equal parts that the split axis should be divided into.

garymm:
Don't forget the version converter support. Might be non-trivial for this change.

gramalingam:
Yes, the version-converter is non-trivial, but conceptually simple: we should add a num_outputs attribute with value = number of outputs specified in the Node.

p-wysocki:
WIP PR: https://github.com/onnx/onnx/pull/4358

p-wysocki:
PR: https://github.com/onnx/onnx/pull/4481

p-wysocki:
The PR has been merged. Please reopen the issue if needed.

PR #4358, opened 2022-07-13
Add Split-18: new num_outputs attribute

p-wysocki:
Signed-off-by: Przemyslaw Wysocki [przemyslaw.wysocki@intel.com](mailto:przemyslaw.wysocki@intel.com)

**Description**
- New Split attribute `num_outputs` determining the number of equally sized output tensors. Can be overridden by `split` input.

**Motivation and Context**
- Fixes an issue described at https://github.com/onnx/onnx/issues/4188

PR #4389, opened 2022-07-28
Add Split-18: new num_outputs attribute

p-wysocki:

**Description**
- New optional Split attribute `num_outputs` determining the number of equally sized output tensors.

**Motivation and Context**
- Fixes an issue described at https://github.com/onnx/onnx/issues/4188


onnx/defs/tensor/defs.cc
```diff
@@ -587,15 +587,15 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
         }));
 
-static const char* Split_ver13_doc =
-    R"DOC(Split a tensor into a list of tensors, along the specified
-'axis'. Lengths of the parts can be specified using input 'split'.
-Otherwise, the tensor is split to equal sized parts.
+static const char* Split_ver18_doc =
+    R"DOC(Either input 'split' or the attribute 'num_outputs' should be specified, but not both.
```
postrational, line 591: This sentence should not be removed:

> Split a tensor into a list of tensors, along the specified 'axis'.

docs/CONTRIBUTING.md
```diff
@@ -31,13 +31,23 @@ from onnx root dir should work.
 
 ## Generated operator documentation
 
-[Operator docs in Operators.md](Operators.md) are automatically generated based on C++ operator definitions and backend Python snippets. To refresh these docs, run the following commands from the repo root and commit the results. Note `ONNX_ML=0` updates Operators.md whereas `ONNX_ML=1` updates Operators-ml.md:
+[Operator docs in Operators.md](Operators.md) are automatically generated based on C++ operator definitions and backend Python snippets. To refresh these docs, run the following commands from the repo root and commit the results.
```
postrational, line 34: CONTRIBUTING.md should not change in this PR.

postrational (changes requested):

postrational:
Previously discussed in https://github.com/onnx/onnx/pull/4358

onnx/defs/tensor/defs.cc
p-wysocki: Good catch, I restored it.

docs/CONTRIBUTING.md
p-wysocki: I reverted the changes.

onnx/defs/tensor/defs.cc
```diff
@@ -682,14 +688,21 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                fail_shape_inference("Attribute `num_outputs` value cannot be lower than 1");
+              }
+              if (split_dim_value % num_outputs != 0) {
+                fail_shape_inference("The input is not evenly splittable");
+              }
+              int chunk_size = split_dim_value / num_outputs;
+              split.reserve(ctx.getNumOutputs());
+              for (int i = 0; i < static_cast<int>(ctx.getNumOutputs()); i++) {
+                split.push_back(chunk_size);
+              }
```
daquexian, line 703: Suggestion:
```suggestion
              split.resize(ctx.getNumOutputs(), chunk_size);
```

onnx/backend/test/case/node/split.py
```diff
@@ -17,7 +17,8 @@ def export_1d() -> None:
             'Split',
             inputs=['input'],
             outputs=['output_1', 'output_2', 'output_3'],
-            axis=0
+            axis=0,
+            num_outputs=3
```
gramalingam, line 21: I think we should perhaps have test-cases for both the earlier version and newer version? @jcwchen ?

postrational (approved):

onnx/backend/test/case/node/split.py
jcwchen: After this PR: https://github.com/onnx/onnx/pull/3855, going forward all of node test cases for the same operator will be updated if there is an opset bump. If people would like to keep some old test cases for old opset versions, they need to specify the opset_version while creating node test cases. Take Loop op as an example, see this [line](https://github.com/onnx/onnx/blob/ea7c5204713cce87e8c22e5382833c68483f329b/onnx/backend/test/case/node/loop.py#L127).

p-wysocki: I applied the changes you suggested. Opset13 tests were reverted, new (opset18) tests stayed the same.

jcwchen:
For unknown reason (might be a GitHub server issue), this PR hasn't been merged into the main branch even after 1 hrs... Let's see whether it will be merged toady.

gramalingam:
As per @p-wysocki , it is unclear why this (empty) merge even happened. He says he did nothing to trigger it. And it seems to be an empty merge with no commits. What's happening? Seems mysterious to me.

p-wysocki:
I was 1 commit ahead of `onnx:main`, it contained all my changes, squashed with a sign off. Unfortunately some other commits got in and there were ~350 files changed. Long story short, I reverted this one, single commit from `p-wysocki:add_split18` and pushed it to my fork, essentially syncing `onnx:main` with `p-wysocki:add_split18`, what can be seen as "Files changed: 0". I think this may be just a weird behavior from GitHub, I did not enable merging (merge should not be possible at all anyway), my changes are not on master and it would be just a "visual" glitch. 

We had this happen in OpenVINO too, where an external contributor seemingly merged his changes, but they were not on master and the PR showed 0 changed files.

PR #4391, opened 2022-07-29
fix bugs and add .clang-tidy

daquexian:
**Description**
1. Fix a critical bug in IR
2. Add .clang-tidy to enable clang-tidy in vscode/vim/... when using clangd
3. Fix other bugs reported by clang-tidy

**Motivation and Context**

An [onnxsim](https://github.com/daquexian/onnx-simplifier) user reports that onnxsim fails to simplify some models. After debugging I found that it is because the `bool` and `int` member of `Dimension` class in onnx IR is not initialized, so accessing them is undefined behavior.

To avoid this kind of bug as possible, I created a .clang-tidy config file to enable clang-tidy in vscode/vim/... when using clangd, and fixed other bugs clang-tidy reported.


onnx/common/ir.h
```diff
@@ -65,9 +68,9 @@ class ResourceGuard final {
 };
 
 struct Dimension final {
-  Dimension() : is_unknown(true) {}
-  Dimension(std::string param) : is_unknown(false), is_int(false), dim(-1), param(std::move(param)) {}
-  Dimension(int64_t dim) : is_unknown(false), is_int(true), dim(dim) {}
+  Dimension() : is_unknown(true), is_int(false), dim(-1) {}
```
daquexian, line 71: Critical bug: is_int and dim are not initialized by default, accessing them is undefined behavior before this PR

onnx/common/ir.h
```diff
@@ -1119,7 +1125,7 @@ struct Graph final {
   // Adds to graph initializer list, initializer names list, and as a graph input
   // Also syncs the initializer name, tensor name, and value name
   // Create an initializer whose value is stored in input
-  Value* addInitializerAndInput(const Tensor& initializer, std::string name) {
+  Value* addInitializerAndInput(const Tensor& initializer, const std::string& name) {
```
daquexian, line 1128: Performance improvement: copying std::string may be expensive

onnx/common/ir.h
```diff
@@ -1172,7 +1178,7 @@ struct Graph final {
 
   friend std::ostream& operator<<(std::ostream& out, const Graph& g);
 
-  void forSelfAndEachSubGraph(std::function<void(Graph*)> fn) {
+  void forSelfAndEachSubGraph(const std::function<void(Graph*)>& fn) {
```
daquexian, line 1181: Performance improvement: copying std::function may be expensive

onnx/common/ir_pb_converter.cc
```diff
@@ -256,7 +256,7 @@ std::unique_ptr<Graph> graphProtoToGraph(const ONNX_NAMESPACE::GraphProto& gp, b
   }
 
   for (int i = 0; i < gp.input_size(); i++) {
-    auto vip = gp.input(i);
+    const auto& vip = gp.input(i);
```
daquexian, line 259: Performance improvement: avoid unnecessary copy of ValueInfoProto

onnx/common/tensor.h
```diff
@@ -75,6 +75,9 @@ struct Tensor final {
     name_ = std::string(other.name_.data(), other.name_.size());
     raw_data_ = std::string(other.raw_data_.data(), other.raw_data_.size());
   }
+  Tensor(Tensor&&) = default;
+  Tensor& operator=(Tensor&&) = default;
```
daquexian: Performance improvement: move constructor and move assignment operator are implicit deleted if copy constructor or copy assignment operator is explicitly declared, so they should be defined explicitly to enable moving.

onnx/common/ir.h
```diff
@@ -52,7 +52,10 @@ class ResourceGuard final {
   bool released_;
 
  public:
-  ResourceGuard(std::function<void()> destructor) : destructor_(std::move(destructor)), released_(false) {}
+  ONNX_DISALLOW_COPY_AND_ASSIGN(ResourceGuard);
```
daquexian, line 55: Bug: ResourceGuard should not be copy-able, for the `destructor` function can only be executed once.

onnx/common/tensor.h
```diff
@@ -75,6 +75,8 @@ struct Tensor final {
     name_ = std::string(other.name_.data(), other.name_.size());
     raw_data_ = std::string(other.raw_data_.data(), other.raw_data_.size());
   }
+  Tensor(Tensor&&) = default;
```
daquexian, line 78: Performance improvement: move constructor and move assignment operator are implicit deleted if copy constructor or copy assignment operator is explicitly declared, so move constructor should be defined explicitly to enable moving.

onnx/common/tensor.h
```diff
@@ -75,6 +75,8 @@ struct Tensor final {
...
+  ~Tensor() = default;
```
daquexian, line 79: define destructor explicitly to meet the rule of five

onnx/common/ir_pb_converter.cc
```diff
@@ -190,13 +190,13 @@ std::vector<Dimension> tensorShapeProtoToDimensions(const ONNX_NAMESPACE::Tensor
...
-      dims.push_back(Dimension(static_cast<int>(tsp.dim(i).dim_value())));
+      dims.emplace_back(static_cast<int>(tsp.dim(i).dim_value()));
     } else if (tsp.dim(i).has_dim_param()) {
-      dims.push_back(Dimension(tsp.dim(i).dim_param()));
+      dims.emplace_back(tsp.dim(i).dim_param());
     } else {
       // a dimension that has neither dim_value nor dim_param set
       // represents an unknown dimension unrelated to other unknown dimensions.
-      dims.push_back(Dimension());
+      dims.emplace_back();
```
daquexian, line 199: Performance improvement: `emplace_back(...)` is more effective than `push_back(Dimension(...))` because it constructs the `Dimension` object in-place.

.clang-tidy
```diff
@@ -0,0 +1,46 @@
...
+  modernize-use-emplace
+
+CheckOptions:
+  # `cppcoreguidelines-special-member-functions` is enabled, refer to https://en.cppreference.com/w/cpp/language/rule_of_three
+  - key:             cppcoreguidelines-special-member-functions.AllowSoleDefaultDtor
+    value:           True
+  - key:             performance-move-const-arg.CheckTriviallyCopyableMove
+    value:           False
+  - key:             cppcoreguidelines-special-member-functions.AllowMissingMoveFunctionsWhenCopyIsDeleted
+    value:           True
```
jcwchen, line 46: It looks useful. Do you think we should add this check into CI as well?

jcwchen (approved):
LGTM. Thank you for the improvement!

.clang-tidy
daquexian: @jcwchen sorry for the late response. Yes, it is definitely better if we have this in CI.

PR #4393, opened 2022-07-29
Clarify and add helper for making attribute references

### LINKED ISSUE #4374: Referencing function attributes in body
### Question
Is there currently a standard way (as described in the [IR](https://github.com/onnx/onnx/blob/main/docs/IR.md)) to reference function attributes inside the function body (i.e. the nodes)?

### Further information
- Relevant Area: IR, model definition, functions

### Notes

The internal implementation uses the field `ref_attr_name` (for example used [here](https://github.com/onnx/onnx/blob/b064f0a88cbe2fb75081dc8d25adde942ad9a916/onnx/defs/parser.cc#L386), called from e.g. `MeanVarianceNormalization`'s [implementation](https://github.com/onnx/onnx/blob/64e79b78a410e69a804f4f198bbaf45193fb3c32/onnx/defs/nn/defs.cc#L2181)) in the language parser, but this does not seem to be documented in the standard nor exposed in `onnx.helper`.

I've found a workaround for accessing function attributes when defining a function in Python: Directly use the protobuf interface for creating an `AttributeProto`:
```py
def ref_attr(name: str, ref_name: str, value_type: onnx.AttributeProto.AttributeType):
    attr = onnx.AttributeProto()
    attr.name = name
    attr.ref_attr_name = ref_name
    attr.type = value_type
    return attr
```

This can be used like so, explicitly adding an `AttributeProto` to `NodeProto.attribute`:
```py
fun_shift_node = onnx.helper.make_node(
    'BitShift',
    ['A', 'L'],
    ['B']
)
# onnx.helper does not support referencing function attributes, do it manually (to specify direction):
fun_shift_node.attribute.append(ref_attr("direction", "dir", onnx.AttributeProto.STRING))

fun = onnx.helper.make_function(
    'test',
    'FunBitShift',
    ['A', 'L'],
    ['B'],
    nodes=[fun_shift_node],
    opset_imports=[onnx.helper.make_operatorsetid('', 16)],
    attributes=["dir"]
)  # fun<dir>: (A, L) -> B
```

Then we can specify the attribute when calling the function normally:
```py
call_shift_node = onnx.helper.make_node(
    'FunBitShift',
    ['D', 'K'],
    ['R'],
    domain='test',
    dir="LEFT"
)  # R = FunBitShift<dir="LEFT">(D, K)
```

I found no alternate solution that avoids doing this outside the helper.

KubinGH:
**Description**
- Adds `onnx.helper.make_attribute_ref` to create parent function attribute references.
- Explains the mechanism of passing attribute references in `IR.md`.

**Motivation and Context**
- When developing code generating ONNX models, supporting the function specification requires attribute references. This  PR helps clarify the expected method for doing this.
- Fixes #4374

### DIFF
--- a/docs/IR.md
+++ b/docs/IR.md
@@ -327,8 +327,11 @@ ints|int64[]|A list of 64-bit integer values.
 strings|byte[][]|A list of UTF-8 strings.
 tensors|Tensor[]|A list of tensor values.
 graphs|Graph[]|A list of graphs.
+ref_attr_name|string|The name of a parent function's attribute.
 
 The properties ‘name’ and ‘type’ are required on all attributes, and ‘doc_string’ SHOULD be used on all attributes. An attribute MUST have only one of the value-carrying properties.
+
+In case ‘ref_attr_name’ is set, this attribute does not contain data, and instead it's a reference to the parent function's attribute of the given name. Can only be used within the function body.
 
 
 #### Variadic Inputs and Outputs
--- a/onnx/helper.py
+++ b/onnx/helper.py
@@ -545,7 +545,23 @@ def make_attribute(
     return attr
 
 
+def make_attribute_ref(
+        name: str,
+        attr_type: AttributeProto.AttributeType,
+        doc_string: Optional[str] = None
+) -> AttributeProto:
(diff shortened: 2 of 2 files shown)


gramalingam (approved):
LGTM, thanks!

PR #4395, opened 2022-07-31
use CMAKE_INSTALL_LIBDIR to avoid hardcoded path

daquexian:
**Description**
We can get the default library dir (lib vs lib64) by cmake [built-in functionality](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), so no need to hardcode them and make users edit them manually.


jcwchen (approved):
Thank you for the improvement!

PR #4400, opened 2022-08-01
Do not allow to read tensor's external_data outside the model directory

### LINKED ISSUE #3991: The onnx runtime allow to load the external_data from the file outside the folder
The external_data field of the tensor proto can have a path to the file which is outside the model current directory or user provided directory, for example "../../../etc/passwd". 

There is no validation on this in this function: https://github.com/onnx/onnx/blob/96516aecd4c110b0ac57eba08ac236ebf7205728/onnx/checker.cc#L129

The python library have the _sanitize_path function which has some basic restrictions but it doesn't work when you use the default onnxruntime package to do the model execution.

I can provide POC and create a patch by request.

jnovikov:
Signed-off-by: jnovikov <johnnovikov0@gmail.com>

This PR fixes the vulnerability that allows to read tensor_data outside the model directory.

The logic behind the PR is to first normalize the relative_path inside the tensor and then check that path has no '../' suffix.
Normalization code is based on go-lang path.Clean() but simplified to work correctly with the relative_paths only. 
Provided tests should cover most of the cases.

Please note that 'optimal' solution would be to use boost::filesystem or c++17 filesystem to get the absolute path file and resolve possible symlink problems, but AFAIK the project target is C++11 and the boost is not used at the project. 

fixes #3991


onnx/test/test_external_data.py
```diff
@@ -492,5 +497,24 @@ def test_save_model_with_external_data_multiple_times(self) -> None:
...
 
 
+class TestNotAllowToLoadExternalDataOutsideModelDirectory(TestLoadExternalDataBase):
+    """Essential test to check that onnx (validate) C++ code will not allow to load external_data outside the model
+    directory. """
+    def create_external_data_tensor(self, value: List[Any], tensor_name: str) -> TensorProto:
+        tensor = from_array(np.array(value))
+        tensor.name = tensor_name
+
+        set_external_data(tensor, location="../../file.bin")
```
jcwchen, line 507: nit: could we also have another test for Windows-path like "..\\..\\..\\file.bin"?

onnx/test/test_external_data.py
```diff
@@ -492,5 +497,24 @@ def test_save_model_with_external_data_multiple_times(self) -> None:
...
+
+        tensor.ClearField('raw_data')
+        tensor.data_location = onnx.TensorProto.EXTERNAL
+        return tensor
+
+    def test_check_model(self) -> None:
+        """We only test the model validation as onnxruntime uses this to load the model. """
```
jcwchen, line 514: Just speak aloud: is it possible this PR breaks some tests in ORT? If so, we need to keep this in mind when upadating ONNX in ORT.

Even if it does break some tests there, probably we should fix it from ORT side (making them use inner directory instead), because loading files from outside directory seems like a bad practice in that situation.

jcwchen (commented):
Nit comments. Thank you for the contribution to make ONNX more secure!

onnx/test/test_external_data.py
jnovikov: It might. Speaking about the vulnerability, the vulnerability affects the ORT(I tested the POC using onnxruntime) and this new check should fix it. 

It might break some tests, but I'm not sure how to test it now. 
I do not know how to build ORT from scratch to check it with patched ONNX library, so it would be great if someone experienced will do it or just provide some guide to me how to build ORT and build it with patched onnx lib.

onnx/test/test_external_data.py
jnovikov: Done

onnx/test/test_external_data.py
jcwchen: Yes, testing it in onnxruntime in advance will be super great. Provide some context about how to build onnxruntime with updated ONNX:
- Build onnxruntime from souce: https://onnxruntime.ai/docs/build/inferencing.html#linux
- Update ONNX in onnxruntime: https://github.com/microsoft/onnxruntime/blob/master/docs/How_To_Update_ONNX_Dev_Notes.md

onnx/test/test_external_data.py
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 import tempfile
+import pathlib
```
jcwchen, line 3: Unused pathlib?

onnx/test/test_external_data.py
jcwchen: I have discussed it with onnxruntime team. It seems that they have similar check already so ideally this PR shouldn't bring test failures when updating ONNX there.

jcwchen (approved):
Thank you for making ONNX more secure! Please remove `pathlib` if it is really unused. Otherwise LGTM.

gramalingam (approved):

onnx/test/test_external_data.py
jnovikov: Good if so. As I mentioned I didn't found the checks inside the onnxruntime. At least my poc works for the 
onnxruntime==1.10.0 (pypy) version.

Issue #4367, opened 2022-07-19
RuntimeError: Inferred shape and existing shape differ in dimension 2: (640) vs (320)


philipwan:
# Bug Report


### Describe the bug
shape_inference.infer_shapes api Error: SliceOp attr['step'] =2 , shape diff wrong

### System information
- Linux Ubuntu 18.04: 
- ONNX version 1.7.0:  
- Python version 3.7.8:
- Onnxruntime 1.7.0


### Reproduction instructions
- Describe the code to reproduce the behavior.
```
import numpy as np                                                                                                      
import onnx                                                                                                             
import onnxruntime as ort                                                                                               
from onnx import TensorProto                                                                                            
from onnx.backend.test.case.node import expect                                                                          
from onnx import shape_inference                                                                                        
                                                                                                                        
inp = np.random.randn(1, 3, 640, 640).astype(np.float32)                                                                
x = onnx.helper.make_tensor_value_info('x', TensorProto.FLOAT, [1, 3, 640, 640])                                        
starts = onnx.helper.make_tensor('starts', TensorProto.INT64, [1], [0])                                                 
ends = onnx.helper.make_tensor('ends', TensorProto.INT64, [1], [9223372036854775807])                                   
steps = onnx.helper.make_tensor('steps', TensorProto.INT64, [1], [2])                                                   
axes = onnx.helper.make_tensor('axes', TensorProto.INT64, [1], [2])                                                     
y = onnx.helper.make_tensor_value_info('y', TensorProto.FLOAT, [1, 3, 16, 640])  
                                                                              
node = onnx.helper.make_node(                                                                                           
    'Slice',                                                                                                            
    inputs=['x', 'starts', 'ends', 'axes', 'steps'],                                                                    
    outputs=['y'],                                                                                                      
)                                                                                                                       
                                                                                                                        
graph_def = onnx.helper.make_graph(                                                                                     
    [node],                                                                                                             
    'test_conv_mode',                                                                                                   
    [x],                                                                                                                
    [y],                                                                                                                
    initializer=[starts, ends, axes, steps]                                                                             
)                                                                                                                       
                                                                                                                        
print(node)                                                                                                             
mode_def = onnx.helper.make_model(graph_def, producer_name='onnx-example')                                              
onnx.checker.check_model(mode_def)                                                                                      
onnx.save(mode_def, "./slice.onnx")                                                                                     
                                                                                                                        
onnx_model = onnx.load("./slice.onnx")                                                                                 
[...]

PR #4403, opened 2022-08-02
Fix IR gap for shape inference: update input from initializer (#2901)

philipwan:
* Update input from initializer if missing

* add a test and comments

* remove useless modification

* fix flake8

* do not update shape if exisit in input

* add return type for test

* correct comments

* add input into temp graph to avoid graph modification

* check input and initializer; consider ir_version

* remove merge for input and initializer

* use FLOAT for shape data type

* Revert "use FLOAT for shape data type"

This reverts commit 404fc87e7fdc610fdc083eaa5bf5c655c30f6e1b.

* also consider opset_version, not use pytest lib

* test no using ir_Version

* free correctly

* create a temp valueinfoproto for free

* use delete instead

* consider opset; default using IR_VERSION

* delete useless typeproto

* skip if using ops from other domain

* use unsigned int for comparison

* Add tests for opset<9 and other domain

* add test_infer_initializer_input_consistency

* ignore shape for being None

* refactor tests and fix typecheck

* ignore type for empty list

* remove consdiering opset_version

* remove considering opsets from domain

* remove domain test as well

* free memory for created tensors and use positive dim

* use proper name for deleteCreatedTypes

* Use (none, none) as valid model for a test

**Description**
- Describe your changes.

**Motivation and Context**

### DIFF
--- a/onnx/shape_inference/implementation.cc
+++ b/onnx/shape_inference/implementation.cc
@@ -159,9 +159,10 @@ static void InferShapesImpl(
     GraphProto* g,
     const std::unordered_map<std::string, TypeProto*>&
         outer_scope_value_types_by_name,
-    const std::unordered_map<std::string, int>& opset_imports,
-    bool check_type,
-    const ISchemaRegistry* schema_registry = OpSchemaRegistry::Instance()
+    const std::unordered_map<std::string, int>& opset_imports, 
+    const bool check_type,
+    const ISchemaRegistry* schema_registry = OpSchemaRegistry::Instance(),
+    const int ir_version = IR_VERSION  // default the latest one
     ) {
   std::unordered_map<std::string, TypeProto*> valueTypesByName{
       outer_scope_value_types_by_name};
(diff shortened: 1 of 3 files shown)


philipwan:
@jcwchen 
hi，Due to this known issue: https://github.com/onnx/onnx/issues/4367, and according to your suggestion, I made a pull request, there is no problem in the cherry-pick process, please review this PR, thank you。
By the way，will version 1.7.1 be released if it is merged?

jcwchen (commented):
Hi @philipwan,
Thank you for your time to propose this PR. Sorry that I might be not clear -- I was suggesting if you really need ONNX 1.7.0 somehow, you can add the IR gap fix into the commit of rel-1.7.0 in your local and use it on your own. For other users who need the IR gap fix, they can directly use ONNX>=1.8. IMO, ONNX 1.7 is a really old version (2 years ago) and we should stabilize it without any small changes, unless there is a critical issue.

jcwchen (commented):
Close this one since we do not plan to do an old update. Please let me know if you have any concern/question. Thanks!

Issue #4405, opened 2022-08-02
Exposing a Python interface for calling inference functions


KubinGH:
# Feature Request

### Feature description
Currently, [the only place](https://github.com/onnx/onnx/blob/d6a0fefa83828dfbaab15e5477ce5c95b12b7be9/onnx/shape_inference/implementation.cc#L332) where `TypeAndShapeInferenceFunction`s (as set in `defs.cc`) are called is the `InferShapes` function. However, for one to call it from the Python API we need a well-formed `ModelProto`, for which inputs and outputs have to be set. 

It seems like for many (all?) cases the output types could be inferred from the inputs, and exposing inference functions could be useful to achieve this for libraries building models - avoiding reimplementing the inference routines instead of using the reference implementation.

The inference functions do not have Python bindings, but it seems like they could. This should be somewhat straightforward following the existing `pybind11` code.

### Alternatives
- Allowing `InferShapes` to work without output shapes specified
  - In this case, we can provide an entire finished model and infer types in one go. Could be useful, but an incremental approach where we can call the inference functions at any point sounds more useful and efficient.
- Reimplementing the standard inferences on Python-side in the reference API.
  - Currently this basically ends up happening in converter libraries, but the implementation could be moved to `onnx.shape_inference`. Doesn't seem like a good idea since the implementation already exists in C++.

### API
Most likely `OpSchema` would receive a new method in Python bindings like `infer_types(node: NodeProto, ...)`, taking objects necessary to construct an `InferenceContext`. It would return the updated objects. 
It may have inference errors thrown in the inference functions, unsure what would be the wanted behaviour when exceptions are disabled.

### Feature Area
Which area in ONNX does this impact?: Python API, shape inference

### Are you willing to contribute it
Yes.

gramalingam:
Sounds good to me. Yes, it is unclear what we could do in builds with exceptions turned off, since it will just crash. It would be useful to see the proposed interface/API. Thanks!

KubinGH:
Please see the draft PR for a proposal - it's not finished yet, as I saw some complications.

- Currently I'm passing in arguments to construct an `InferenceContextImpl` from Python. Some of them, however, make things more complex:
  - `generatedShapeData` basically requires also exposing a similar partial data propagation interface. 
  - Additionally, `graphInferenceContext` takes a lot of arguments, too (it's used for type inference when GraphProto attributes are present), which would complicate the interface.
[...]

PR #4409, opened 2022-08-04
Expose a Python interface for inference functions

KubinGH:
Exposes a node-level API for performing type/shape inference to Python bindings.

**Description**
(Updated)

- Adds a low-level function `OpSchema._infer_node_outputs`, to which serialised protobuf bytes are passed - arguments: schema, node, input types, input data/other inference context arguments, returns: inferred output types. 
- Adds a public interface `shape_inference.infer_node_outputs`:
```py
def infer_node_outputs(
     schema: onnx.defs.OpSchema,
     node: onnx.NodeProto,
     input_types: Dict[str, onnx.TypeProto],
     input_data: Optional[Dict[str, onnx.TensorProto]] = None,
     input_sparse_data: Optional[Dict[str, onnx.SparseTensorProto]] = None,
 ) -> Dict[str, onnx.TypeProto]:  # output_types
```
  - Underneath this performs validation of passed in nodes/types, `ValidationError`s and `InferenceError`s are raised.
  - Should be the only used function, `OpSchema._infer_node_outputs` is internal to the implementation.
  - `input_types` keys are named like `node` declares its inputs.
  - `input_data` and `input_sparse_data` are used for passing in known constant inputs (as per C++ usage).
  - Could implement passing in partial data propagation results and `GraphInferenceContext` in the future.
- Adds test cases for some inference calls from the Python side.

**Motivation and Context**
- See #4405


KubinGH:
@gramalingam Would you mind viewing the proposal if this sort of interface would work? In the future it could be extended to support the other arguments, they could be optional so it's backward compatible.

onnx/defs/schema.h
```diff
@@ -818,6 +818,10 @@ class OpSchema final {
     return tensor_inference_function_ ? true : false;
   }
 
+  InferenceFunction get_type_and_shape_inference_function() const {
```
gramalingam, line 821: I think this getter function already exists: see `GetTypeAndShapeInferenceFunction`.

gramalingam:
Updating the pyi definition for [OpSchema](https://github.com/onnx/onnx/blob/main/onnx/onnx_cpp2py_export/defs.pyi#L7) would be useful.

onnx/test/inference_function_test.py
```diff
@@ -0,0 +1,64 @@
+import unittest
+from typing import Iterable, Optional, List
+
+import onnx
+import onnx.shape_inference
+
+
+def call_infer_types(
```
gramalingam, line 8: I think it might be useful to promote an API like this as a wrapper around the raw C++ method (that is, make if part of the public API instead of something in just this test-file). But, for that, I don't think we would want `node` to be optional. And am not sure about `num_outputs` either.

onnx/cpp2py_export.cc
```diff
@@ -39,6 +39,47 @@ static std::string ProtoBytesToText(const py::bytes& bytes) {
...
+  size_t i = 0;
+  for(auto kv : bytesMap) {
+    ParseProtoFromPyBytes(&values[i], kv.second);
+    result[kv.first] = &values[i];
+    i++;
+  }
+  return make_pair(move(values), result);
+}
+
+std::vector<py::bytes> CallInferenceFunction(
```
gramalingam, line 57: nit: for various functions it might be useful to explicitly indicate it is a node-level inference (vs. graph-level inference) in the name, eg: `CallNodeInferenceFunction` and `infer_types_for_node`.

onnx/test/inference_function_test.py
KubinGH: Sure, that definitely makes sense. We could always require a `NodeProto`, `num_outputs` was a temporary idea.

KubinGH:
> I think it might be useful to promote an API like this as a wrapper around the raw C++ method (that is, make if part of the public API instead of something in just this test-file). But, for that, I don't think we would want `node` to be optional. And am not sure about `num_outputs` either.

When it comes to the Python-side interface: A node has input/output names, operator name & domain, attributes. Own name and docstring don't matter. We are missing basically only the operator version it seems, and the input types. I see some variations of the possible signature:

- Inputs & outputs
  - `list[TypeProto]` - we already have an ordering of fields inside the node. Node inputs/outputs are positional anyway, so making it explicit with ordering in a `list` seems okay.
  - `dict[str, TypeProto]` - since the node names the fields in definition. On the other hand since the node has to get created explicitly first anyway in this approach we might as well use the names that are within the node. I think this shouldn't mess with generics etc. since if we use a name multiple times is has to have the same type - we get some very simple consistency check.
- Operator version
  - `version: int` - get the current operator schema from a list at this current version based on what is in the node.
  - `OpSchema`- make the user explicitly provide the schema themselves. Is more redundant, since the Node already has the domain and identifier defined. 


Also, where would we put this function? `shape_inference` would be fine, but really it should have been called `inference` (as it also infers types and other stuff?). Could also put it in `helper`. Any ideas?

Let me know your thoughts and I'll get started on some improvements in the Python-interface area.

onnx/cpp2py_export.cc
KubinGH: That's a fair idea, I'll adapt to this in the next iteration.

onnx/defs/schema.h
KubinGH: You're right, I think I must have thought it was private for some reason. Fixed!

KubinGH:
@gramalingam I pushed some changes:

```cpp
std::unordered_map<std::string, py::bytes> CallNodeInferenceFunction(
    OpSchema* schema,
    const py::bytes& nodeBytes,
    std::unordered_map<std::string, py::bytes> valueTypesByNameBytes,
    std::unordered_map<std::string, py::bytes> inputDataByNameBytes,
    std::unordered_map<std::string, py::bytes> inputSparseDataByNameBytes) {
```

- Now returns a map, since the inputs are a map already. Doesn't return entries were the output `TypeProto` was not initialised (inference failed implicitly?).
- Added `OpSchema::Verify` before the call to the inference (checks if the count of inputs/outputs is correct, etc.), and `OpSchema::CheckInputOutputType` afterwards (checks if the types are consistent with the signature). Both should throw `ValidationError`.
- At the same time, inference itself may throw `InferenceError`. Interestingly, this may also throw errors related to attributes (so things like `Cast` will check here if exactly one of the right attributes is set).
- Let me know if I should handle exceptions there in some more involved way. I think what is thrown already reflects most of what `check_model` and `infer_shapes` do.

```py
# class OpSchema:
    def infer_node_outputs(self, node_proto: bytes, value_types: Dict[str, bytes],
                           input_data: Dict[str, bytes], input_sparse_data: Dict[str, bytes]
                           ) -> Dict[str, bytes]: ...
```

- This is just a binding for the C++ function. I added it to `defs.pyi` as above. 
- My general feeling is that the attributes after `value_types` should have empty defaults so that in the future `generatedShapeData` and `graphInferenceContext` may be supported as well (but this involves, I think, significant work).

```py
# shape_inference.py
def infer_node_outputs(
    schema: onnx.defs.OpSchema, node: onnx.NodeProto, input_types: Dict[str, onnx.TypeProto],
    input_data: Optional[Dict[str, onnx.TensorProto]] = None,
    input_sparse_data: Optional[Dict[str, onnx.SparseTensorProto]] = None
) -> Dict[str, onnx.TypeProto]: ...
```

- The public-facing Python function does all of the conversions and expects proper types.
- Already has default arguments, which we may change in the future. Reflect the actual underlying implementation.
- Does some checks if all the expected keys are present and only copies over what is actually required on C++ side for some performance.


Let me know what you think of these changes! I also added some more tests.

onnx/test/inference_function_test.py
```diff
@@ -1,63 +1,73 @@
...
-        assert call_infer_types(ADD_SCHEMA, [
-            onnx.helper.make_tensor_type_proto(1, (None, 2)),
-            onnx.helper.make_tensor_type_proto(1, (2,))
-        ], 1) == [onnx.helper.make_tensor_type_proto(1, (None, 2))]
-        assert call_infer_types(ADD_SCHEMA, [
-            onnx.helper.make_tensor_type_proto(1, (None, 2)),
-            onnx.helper.make_tensor_type_proto(1, (1, 2))
-        ], 1) == [onnx.helper.make_tensor_type_proto(1, (None, 2))]
+        cases = [
+            ({'A': (1, ()), 'B': (1, ())}, {'C': (1, ())}),
```
gramalingam, line 45: nit: I assume the 1 is the type encoding of float. If so, better to use the symbolic names like `TensorProto.FLOAT` instead.

onnx/cpp2py_export.cc
```diff
@@ -39,6 +39,57 @@ static std::string ProtoBytesToText(const py::bytes& bytes) {
...
+
+  // Convert arguments to C++ types, allocating memory
+  const auto& valueTypes = ParseProtoFromBytesMap<TypeProto>(valueTypesByNameBytes);
+  const auto& inputData = ParseProtoFromBytesMap<const TensorProto>(inputDataByNameBytes);
+  const auto& inputSparseData = ParseProtoFromBytesMap<const SparseTensorProto>(inputSparseDataByNameBytes);
+
+  // Construct inference context and get results - may throw InferenceError
+  shape_inference::InferenceContextImpl ctx(node, valueTypes.second, inputData.second, inputSparseData.second);
+  schema->GetTypeAndShapeInferenceFunction()(ctx);
+  // Verify the inference succeeded - may also throw ValidationError
```
gramalingam, line 75: A question: wondering what happens in builds that don't support exceptions. I assume that the C++ code with abort. What does pybind return in this case? (I realize that this is a pre-existing issue, but just wondering whether we should mark/exclude tests in some special way for this case.)

KubinGH: Looks like this isn't really going to be an issue, since `cmake_build` under Python requires the bindings to be built with exceptions:

https://github.com/onnx/onnx/blob/f6cb158986cf65feae64774d9a851e2649b94073/setup.py#L207-L208

onnx/test/inference_function_test.py
KubinGH: I'll fix this (and the Windows CI, looks like it's failing due to a different default int dtype for a numpy array?) tomorrow.

gramalingam: I heard that the Windows CI are failing across PRs, may be unrelated to this PR. Thanks!

onnx/cpp2py_export.cc
```diff
@@ -85,6 +136,10 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
             return py::bytes(bytes);
           })
       .def_property_readonly("has_context_dependent_function", &OpSchema::HasContextDependentFunction)
+      .def("infer_node_outputs", CallNodeInferenceFunction,
+          py::arg("nodeBytes"), py::arg("valueTypesByNameBytes"),
+          py::arg("inputDataByNameBytes") = std::unordered_map<std::string, py::bytes>{},
```
xadupre, line 141: None is better than an empty container as a default value.

JakubBachurskiQC: What do you mean by None here? A null pointer? 
I think that not having any information in the map makes sense as a default for inference.

KubinGH: I don't think I know what you mean here. This is C++ side, so there's no None. Do you mean to make it a pointer and use nullptr/use std::optional? 
I think that this wouldn't be a lot better as the map isn't mutated and also in this it makes sense that the default is "no additional information" (for the inference context we have to pass in _some_ map, and it makes sense to make it empty as there's no input data we know).
Could you provide a code example for the change?

gramalingam (approved):
Looks great, thanks very much!

KubinGH:
Thanks! 
Should be ready for merge, then?

onnx/shape_inference.py
```diff
@@ -82,4 +82,42 @@ def infer_shapes_path(
         )
 
 
+def infer_node_outputs(
+    schema: onnx.defs.OpSchema,
+    node: onnx.NodeProto,
+    input_types: Dict[str, onnx.TypeProto],
```
justinchuby, line 88: Do we explicitly require a dictionary for these params? If not prefer a more generic type like Mapping. 

——-

Use more generic types for input parameters and specific types for return values. For example, a function may take a `Sequence` and return a `List`.

Diagram for reference: https://gist.github.com/justinchuby/4021cebe9e093f636759a88de325c85f

Issue #4404, opened 2022-08-02
No shipped stubs and using mypy


KubinGH:
### Question
It looks like default ONNX (both installed through pip and conda-forge) does not ship with stub files that mypy can use for typing types like `TensorProto`, `AttributeProto`, etc. `*Proto`.  Is there any way to generate them when using the library, or is this a bug?

When I tried using `python setup.py develop` myself, it correctly generated the stubs (`.pyi`s for all the protobuf modules), but it would be nice if this worked when installing through a manager. I tracked down that the `cmake` generates the stubs with the `ONNX_GEN_PB_TYPE_STUBS` flag:

 https://github.com/onnx/onnx/blob/66480ce6ab0f4c56e27bb120f246c3ddd0d1e19e/CMakeLists.txt#L274-L296

However, they don't seem to be generated on a default installation, and I'm not sure why.

### System information
- OS Platform and Distribution: macOS Monterey 12.4
- ONNX version: 1.12.0 
- Python version: 3.10.5

### Further information
- Relevant Area: best practices

### Notes

The main issue is that types like `AttributeProto` end up as `Any`, even though we know that they are concrete.
Additionally, information on what fields they have is lost. 

```py
# test.py
import onnx
reveal_type(onnx.helper.AttributeProto)
```

```py
test.py:1: error: Skipping analyzing "onnx": module is installed, but missing library stubs or py.typed marker
test.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
test.py:3: note: Revealed type is "Any"
Found 1 error in 1 file (checked 1 source file)
```

jcwchen:
Hi @KubinGH,
Good question -- thank you for catching this. I don't have the history context that why onnx did not ship stubs in PyPI, but it seems something good to have as long as it does not have an impact on performance or binary size. I propose a possible fix for it: https://github.com/onnx/onnx/pull/4412. Please review it. Thanks!

KubinGH:
Are you sure this fully fixes the issue? It still types `AttributeProto` as `Any` for me:

```py
import onnx
reveal_type(onnx.helper.AttributeProto)
```

```
$ mypy test.py 
test.py:2: note: Revealed type is "Any"
Success: no issues found in 1 source file
```

Note: I had to build this myself as I'm on Apple Silicon (sorry for not specifying this), using `python setup.py bdist_wheel`. 

[...]

PR #4412, opened 2022-08-04
Add py.typed to let mypy get stub for type information

jcwchen:
**Description**
- Add py.typed to let mypy get stub for type information
- Ship onnx/py.typed in the wheel to make type information available from PyPI.

**Motivation and Context**
Fixes https://github.com/onnx/onnx/issues/4404.


setup.py
```diff
@@ -340,6 +340,7 @@ def run(self):
     packages=packages,
     license="Apache License v2.0",
     include_package_data=True,
+    package_data={"onnx": ["py.typed"]},
```
KubinGH, line 343:
```suggestion
    package_data={"onnx": ["py.typed", "*.pyi"]},
```

KubinGH (changes requested):
I think this proposed change both universally fixes the typing error on my setup & properly types proto objects and their fields (see my comment: https://github.com/onnx/onnx/issues/4404#issuecomment-1207447210)

KubinGH (approved):
LGTM, issue seems to be fixed!

onnx/py.typed
```diff
@@ -0,0 +1 @@
+# Marker file for PEP-561 (inline types)
```
gramalingam, line 1: nit: Do we need to add an end-of-line?

gramalingam (approved):

onnx/py.typed
jcwchen: Sounds good. Added.

setup.py
```diff
@@ -340,6 +340,7 @@ def run(self):
     packages=packages,
     license="Apache License v2.0",
     include_package_data=True,
+    package_data={"onnx": ["py.typed", "*.pyi"]},
```
jcwchen, line 343: One nit update: I removed adding `py.typed` in MANIFEST.in, because I think this line L343 will do the same thing here. I have tested the updated ONNX wheel can still produce correct type information.

PR #4413, opened 2022-08-05
Add an assert to make sure the quantity is reasonable for vsnprintf

jcwchen:
**Description**
~Add an assert to make sure the quantity is reasonable for vsnprintf.~ Add comments about why current vsnprintf is fine without vulnerability issue.

**Motivation and Context**
To resolve potential vulnerability issue for vsnprintf: Potentially vulnerable format argument for printf style function. The printf type functions format argument does not appear to be hard-coded. Never place user data in the format argument as an attacker can influence a buffer overflow through the format string itself.


onnx/common/assertions.cc
```diff
@@ -6,20 +6,23 @@
...
 
 namespace ONNX_NAMESPACE {
 
 std::string barf(const char* fmt, ...) {
-  char msg[2048];
+  char msg[BUFFER_MAX_SIZE];
   va_list args;
-
+  // To resolve potential vulnerability issue for vsnprintf, add an assert to make sure the quantity is reasonable.
+  assert(strlen(fmt) <= BUFFER_MAX_SIZE && "The string length for vsnprintf is larger than the buffer size 2048.");
```
daquexian, line 22: Could you please explain a bit more about why checking the length of fmt resolves potential vulnerability issues? Thanks!

BTW, should `<` be used instead of `<=`? Because `strlen(fmt)` doesn't include the terminating null but the `n` of `vsnprintf` includes it.

AlexandreEichenberger: We did the same fixes in onnx-mlir for the same type of vulnerability:

from https://cplusplus.com/reference/cstdio/vsnprintf/
```
Write formatted data from variable argument list to sized buffer
Composes a string with the same text that would be printed if format was used on [printf](https://cplusplus.com/printf), but using the elements in the variable argument list identified by arg instead of additional function arguments and storing the resulting content as a C string in the buffer pointed by s (taking n as the maximum buffer capacity to fill).

If the resulting string would be longer than n-1 characters, the remaining characters are discarded and not stored, but counted for the value returned by the function.

Internally, the function retrieves arguments from the list identified by arg as if [va_arg](https://cplusplus.com/va_arg) was used on it, and thus the state of arg is likely to be altered by the call.

In any case, arg should have been initialized by [va_start](https://cplusplus.com/va_start) at some point before the call, and it is expected to be released by [va_end](https://cplusplus.com/va_end) at some point after the call.

Parameters
s
Pointer to a buffer where the resulting C-string is stored.
The buffer should have a size of at least n characters.
n
Maximum number of bytes to be used in the buffer.
The generated string has a length of at most n-1, leaving space for the additional terminating null character.
[size_t](https://cplusplus.com/cstdio:size_t) is an unsigned integral type.
format
C string that contains a format string that follows the same specifications as format in [printf](https://cplusplus.com/printf) (see [printf](https://cplusplus.com/printf) for details).
arg
A value identifying a variable arguments list initialized with [va_start](https://cplusplus.com/va_start).
[va_list](https://cplusplus.com/va_list) is a special type defined in [<cstdarg>](https://cplusplus.com/cstdarg).
```

The key phrase is "If the resulting string would be longer than n-1 characters, the remaining characters are discarded and not stored, but counted for the value returned by the function."

So in your case, the code is already safe as the buffer is sized `BUFFER_MAX_SIZE` and you limit the writing in the buffer by `BUFFER_MAX_SIZE` bytes -1.

The format string is kind of irrelevant, as any `%f` or similar input will be expanded in multi-characters. And the `vsnprintf` already will truncate the output string written to the buffer to `BUFFER_MAX_SIZE`.

So you can safely remove the assert, which does not protect anything, and I would just write a lengthy comment explaining why this is safe.

jcwchen: Thank you @AlexandreEichenberger for the elaboration! You suggestion makes sense to me. I just reverted the unnecessary check and add some comments about why the current vsnprintf is fine.

gramalingam (approved):

daquexian (approved):

Issue #3839, opened 2021-11-11
Singular Value Decomposition (SVD)


coltonpeltierSE:
# New Operator

### Describe the operator
The Singular Value Decomposition (SVD) is a popular linear algebra technique to decompose any matrix into 3 matrices. The SVD is popular in many fields and is particularly useful in ML for use cases like compression (one of my use cases), or on denoising gathered data before performing inference (another of my use cases).

### Can this operator be constructed using existing onnx operators?
I'm not sure of a clear way.

### Is this operator used by any model currently? Which one?
My team within Schneider Electric is currently working on developing models which will utilize SVD in Tensorflow. We have use cases for exporting and potentially embedding these models in our IoT devices using ONNX.

### Are you willing to contribute it? (Y/N)
No, I do not have a deep enough understanding of ONNX development.

### Notes
My original request to the onnxruntime team is [here](https://github.com/onnx/tensorflow-onnx/issues/1720).
I've found another request for the SVD to be implemented in ONNX [here](https://github.com/pytorch/pytorch/issues/63861).

askhade:
@coltonpeltierSE : Can this op be represented as a composition of existing primitive onnx ops? List of available onnx ops can be found here: https://github.com/onnx/onnx/blob/master/docs/Operators.md 

FYI @gramalingam , @postrational

coltonpeltierSE:
@askhade - I didn't think so initially, but I've done some more digging... Looking at [Eigen's library](https://eigen.tuxfamily.org/dox/group__SVD__Module.html), there seems to be two main techniques to solving for the SVD in a computational setting:
- Jacobi SVD
- Bidiagonal Divide and Conquer SVD (This seems like it divides the matrix into smaller sections and performs Jacobi SVD). 

I found an interesting reference implementation in python of the Jacobi Eigenvalue algorithm [here](https://github.com/mateuv/MetodosNumericos/blob/master/python/NumericalMethodsInEngineeringWithPython/jacobi.py).

Looking at the operators within the python code (len, range, abs, sqrt, identity) I think it may be possible to put this together using the operators you've linked resulting in the Jacobi Eigenvalue algorithm. This could then be extended to perform the SVD on small matrices. I think this could then be extended to handle larger matrices via Bidiagonal Divide and Conquer.

xadupre:
[...]

PR #4416, opened 2022-08-06
SVD operator

williamberman:
### Semantics:

The SVD operator covers pytorch, numpy, and tensorflow’s SVD semantics. 

Numpy and tensorflow use the same `compute_uv` flag for computing just the singular values. Pytorch uses two different operations, `svd` and `svdvals`.

Pytorch and numpy return the same conjugate transpose, `Vh`. Tensorflow returns `V` directly.

Tensorflow returns in the same order, `S U Vh` because S is the only non-optional return value. Pytorch and numpy return in the order of the factorization, `U S Vh`.  

### Derivative
"thin"/"partial" vs "full", computing only singular values vs the whole factorization, and real vs complex inputs all change the derivative, impacting both its value and its numerical stability. 

There are different resources documenting the different derivative variants. The implementations in well known AD codebases (pytorch, tensorflow, and jax) all have slightly different implementations of the derivative.

I consolidated the documentation of the different cases and provided example implementations in this [python notebook](https://github.com/williamberman/svd-derivative/blob/main/svd-derivative.ipynb)

### Existing docs
- [Numpy](https://numpy.org/doc/stable/reference/generated/numpy.linalg.svd.html)
- [Pytorch](https://pytorch.org/docs/stable/generated/torch.linalg.svd.html)
- [Tensorflow](https://www.tensorflow.org/api_docs/python/tf/linalg/svd)

### Previous discussion on adding an SVD operator to ONNX
https://github.com/pytorch/pytorch/issues/81084
https://github.com/onnx/onnx/issues/3839 

### Example models

[An Analysis of SVD for Deep Rotation Estimation](https://proceedings.neurips.cc//paper/2020/file/fec3392b0dc073244d38eba1feb8e6b7-Paper.pdf)

SVD is used as a layer in a neural net for predicting rotation matrices. The layer is defined as $\mathrm{SVDO^+}(M) := U \Sigma ' V^\top$ where $\Sigma ' = diag(1, ..., 1, det(U V^\top))$ (See equation 2). There are two models, SVD-Train and SVD-Inference. SVD-Train uses $\mathrm{SVDO^+}$ as the final layer for both training and inference. SVD-Inference omits $\mathrm{SVDO^+}$ as the final layer in training but it is used as the final layer during inference (See section 4 methods).

The full network definition can be found on [github](https://github.com/google-research/google-research/blob/f36474f2e18c3b79a59f7d82d604360f7781e770/special_orthogonalization/main_point_cloud.py#L133). See `regress_from_features` for the pre-SVD layer definitions. 

[Training Deep Networks with Structured Layers by Matrix Backpropagation](https://arxiv.org/pdf/1509.07838.pdf)

The image recognition layer called second-order pooling computes $log(F F^\top + \epsilon I)$ where F is a matrix of image features. Given the SVD of F, the layer can be simplified so log is computed element wise over a diagonalized matrix. Given $F = U \Sigma V^\top$, the second-order pooling layer simplifies to $V log(Σ^\top Σ+ \epsilon I)V^\top $. See section 5.2

##### [Improving training of deep neural networks via Singular Value Bounding](https://arxiv.org/pdf/1611.06013.pdf) and [Orthogonal Deep Neural Networks](https://arxiv.org/abs/1905.05929)

Training proceeds by standard SGD except that weight matrices are maintained as near orthogonal by bounding/clipping their singular values near 1. Weight matrix singular values are bounded within the range $[\frac{1}{1 + \epsilon},1 + \epsilon]$ every $T_{svb}$ iterations where $\epsilon$ and $T_{svb}$ are hyperparameters. See Algorithm 1.

##### [SVD-Softmax: Fast Softmax Approximation on Large Vocabulary Neural Networks](https://proceedings.neurips.cc/paper/2017/file/4e2a6330465c8ffcaa696a5a16639176-Paper.pdf)

SVD-Softmax is a fast approximation of softmax that can be used during inference. The decomposition of the softmax weight matrix, $A = U \Sigma V^\top$, is used to create the matrix $B = U \Sigma$. A subset W of the columns of B are used to estimate the result of the softmax where W is a hyperparameter. The complete softmax is computed for the top N approximations where N is a hyperparameter. See Algorithm 1.

##### [SVD-Embedded Deep Autoencoder for MIMO Communications](https://arxiv.org/abs/2111.02359)

This model embeds the SVD factorization of the channel matrix into the DAE.

The singular values of the channel matrix are used as inputs to create part of the feature vector, $v_\gamma$ (equation 4). $v_\gamma$ is concatenated with the bit input to create the complete input to the Transmitter DAE (section III.A.2). $v_\gamma$ is also concatenated with the output of the Receiver Pre-processor to create the input to the Receiver DAE (section III.F).

The Transmitter Precoding adds one layer of non-trainable weights composed of the right singular vectors of the channel matrix (section III.C).

The Receiver Pre-processing adds two layers of non-trainable weights. One is the left-singular vectors of the channel matrix. The other is the pseudo-inverse of the matrix containing the singular values as its diagonal (section III.E).


onnx/defs/math/defs.cc
```diff
@@ -3337,4 +3337,220 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            OpSchema::Unknown)
+        .TypeConstraint(
+            "T",
+            OpSchema::all_numeric_types_with_bfloat_and_complex(),
+            "Constrain input and output types to all numeric tensor types")
+        .Attr(
+            "full_matrices",
+            "If true (1), compute the full SVD. If false (0), compute the reduced SVD. Defaults to true (1).",
+            AttributeProto::INT,
+            false)
```
williamberman, line 3513: I don't see a boolean type for AttributeProto. Is it acceptable to use an AttributeProto::INT? We treat 0 as false and everything else as true

onnx/defs/math/defs.cc
```diff
@@ -3337,4 +3337,220 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  for (; dim_idx < A_dim_size - 2; ++dim_idx) {
+    const auto dim = A_shape.dim(dim_idx);
+
+    if (compute_uv) {
+      *U_shape->add_dim() = dim;
+      *Vh_shape->add_dim() = dim;
+    }
+
+    *S_shape->add_dim() = dim;
+  }
```
williamberman, line 3420: Confirming dimension ordering here is correct. I.e. dimension at index 0 is highest dimension and dimension at index `A_shape.dim_size() - 1` is the lowest dimension.

onnx/defs/math/defs.cc
```diff
@@ -3337,4 +3337,220 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Input(
+            0,
+            "A",
+            "Input tensor to factor. Must have at least 2 dimensions.",
+            "T",
+            OpSchema::Single,
+            true,
+            1,
+            // TODO
+            OpSchema::Unknown)
```
williamberman, line 3474: x-posting from LFAI slack

 My basic understanding is that SVD (and other matrix factorizations) are usually differentiable but sometimes are numerically unstable. I'm not sure exactly how this maps to specifying differentiability in the onnx operator inputs/outputs -- should all input and output differentiability be unknown?
From the pytorch docs re: SVD differentiability
Gradients computed using U or Vh will only be finite when A does not have repeated singular values. If A is rectangular, additionally, zero must also not be one of its singular values. Furthermore, if the distance between any two singular values is close to zero, the gradient will be numerically unstable ...
torch.linalg.svdvals() computes only the singular values. Unlike torch.linalg.svd(), the gradients of svdvals() are always numerically stable.
https://pytorch.org/docs/stable/generated/torch.linalg.svd.html#torch.linalg.svd
https://pytorch.org/docs/stable/generated/torch.linalg.svdvals.html
https://pytorch.org/docs/stable/generated/torch.linalg.eigh.html
Previous discussion in pytorch:
https://github.com/pytorch/pytorch/issues/49886
https://github.com/pytorch/pytorch/issues/57272
https://github.com/pytorch/pytorch/issues/47599#issuecomment-812539850

onnx/defs/math/defs.cc
```diff
@@ -3337,4 +3337,220 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+See SVD operator doc for general SVD factorization and dimension information.
+SVDVals only computes and outputs S, the vector of singular values.
+
+Full vs partial SVD does not apply as the vector of singular values will always have the same dimensions.
+
+TODO - add doc about differentiability. According to pytorch, svdvals is always differentiable
+)DOC";
+
+ONNX_OPERATOR_SET_SCHEMA(
```
xadupre, line 3527: Why not having an optional output instead of two operators?

onnx/defs/schema.h
```diff
@@ -612,6 +612,25 @@ class OpSchema final {
     return all_numeric_types_with_bfloat;
   }
 
+  static const std::vector<std::string>& all_numeric_types_with_bfloat_and_complex() {
+    static const std::vector<std::string> all_numeric_types_with_bfloat_and_complex = {
+        "tensor(uint8)",
+        "tensor(uint16)",
+        "tensor(uint32)",
+        "tensor(uint64)",
+        "tensor(int8)",
```
xadupre, line 621: What does it mean to have a SVD with integers? Eigen value will be likely be real even if inputs are integers tensors.

onnx/defs/math/defs.cc
williamberman: My intuition was having two separate operators with known outputs instead of one with optional outputs would be easier for consumers and runtimes but happy to change it to one operator with optional outputs! 

As an additional datapoint, pytorch has two separate operators. Tensorflow and numpy have a single operator with optional output.

onnx/defs/schema.h
williamberman: Ah ok yes makes sense re: integers. Looks like the pytorch docs say the singular values will always be real. 

Is this the appropriate type signature then?

`T -> (T, S, T)  where T: tensor(float16) | tensor(float) | tensor(double) | tensor(bfloat16) | tensor(complex64) | tensor(complex128), S: tensor(float16) | tensor(float) | tensor(double) | tensor(bfloat16)`

xadupre: It works for me.

onnx/defs/math/defs.cc
xadupre: A couple of onnx  operators already have optional outputs. I don't know what the first authors of onnx imagined for this kind of functions but i did not find any other case in onnx with two operators doing the same thing but returning different outputs. Maybe another opinion would be better.

onnx/defs/schema.h
williamberman: done!

onnx/defs/math/defs.cc
williamberman: Yep if preferring optional outputs is more standard today in onnx, I would prefer to keep it consistent. Would be rather strightforward to combine them in the current PR

I'm not sure who to ping for the other opinion?

xadupre: @gramalingam may be able to help about this topic.

williamberman: @gramalingam any thoughts here? :)

onnx/defs/math/defs.cc
gramalingam: Yes, that's right, the above definition seems fine.

gramalingam:
I realize SVD is quite well-known, but it would be useful to mention example models that use SVD, if you are aware of any, as motivation.

onnx/defs/math/defs.cc
williamberman: Ok great, will update!

williamberman:
> I realize SVD is quite well-known, but it would be useful to mention example models that use SVD, if you are aware of any, as motivation.

I'm not aware of example models off the top of my head as I just picked it up by looking through the open issues, but I will do some digging :)

cc @coltonpeltierSE since you had the original open issue, do you have any pointers on example models we could include in the PR description?

onnx/defs/math/defs.cc
williamberman: ok done, combined them :)

p-wysocki:
I found some models using SVD:
- [SVD-Embedded Deep Autoencoder for MIMO Communications](https://arxiv.org/abs/2111.02359) mentions models with SVD
- [Cloud K-SVD for Image Denoising](https://www.researchgate.net/publication/358344505_Cloud_K-SVD_for_Image_Denoising) no much info and the paper is not available yet, but it looks like it uses SVD
- [Deep residual-SVD network for brain image registration](https://www.researchgate.net/publication/361399871_Deep_residual-SVD_network_for_brain_image_registration) the paper is available on request, the model described uses SVD as a denoising measure
- [Image Fusion-Based Watermarking in IWT-SVD Domain](https://www.researchgate.net/publication/361549225_Image_Fusion-Based_Watermarking_in_IWT-SVD_Domain) watermarking algorithm using SVD
- [A Simple Yet Effective SVD-GCN for Directed Graphs](https://www.researchgate.net/publication/360724525_A_Simple_Yet_Effective_SVD-GCN_for_Directed_Graphs) Graph Neural Network using SVD

onnx/defs/math/defs.cc
wschin: I'd say let's make it differentiable following Pytorch.
```python
import torch
A = torch.ones(3, 3).requires_grad_()
U, S, V = torch.linalg.svd(A, full_matrices=False)
print(U.requires_grad) # True, so differentiable.
print(S.requires_grad) # True, so differentiable.
print(V.requires_grad) # True, so differentiable.
(U+S+V).relu().sum().backward()
print(A.grad) # A is differentiable because A.grad is
                     # tensor([[ 1.4211e+00,  1.9990e+07, -1.9990e+07],
                     #         [-1.9990e+07, -5.9821e-01,  1.9990e+07],
                     #         [ 1.9990e+07, -1.9990e+07,  7.3364e-01]])
```

williamberman:
> I found some models using SVD:
> 
> * [SVD-Embedded Deep Autoencoder for MIMO Communications](https://arxiv.org/abs/2111.02359) mentions models with SVD
> * [Cloud K-SVD for Image Denoising](https://www.researchgate.net/publication/358344505_Cloud_K-SVD_for_Image_Denoising) no much info and the paper is not available yet, but it looks like it uses SVD
> * [Deep residual-SVD network for brain image registration](https://www.researchgate.net/publication/361399871_Deep_residual-SVD_network_for_brain_image_registration) the paper is available on request, the model described uses SVD as a denoising measure
> * [Image Fusion-Based Watermarking in IWT-SVD Domain](https://www.researchgate.net/publication/361549225_Image_Fusion-Based_Watermarking_in_IWT-SVD_Domain) watermarking algorithm using SVD
> * [A Simple Yet Effective SVD-GCN for Directed Graphs](https://www.researchgate.net/publication/360724525_A_Simple_Yet_Effective_SVD-GCN_for_Directed_Graphs) Graph Neural Network using SVD

Thank you!

onnx/defs/math/defs.cc
williamberman: so to clarify, the issues with numerical stability and other sidecases with computing the gradient should not impact if the inputs and outputs are marked as differentiable in ONNX? 

i.e. example of 0 singular value for rectangular A

```python
A = torch.ones(4, 3).requires_grad_()
U, S, V = torch.linalg.svd(A, full_matrices=False)
print(S)
print(U.requires_grad)
print(S.requires_grad)
print(V.requires_grad)
(U + torch.cat((S + V, torch.zeros(1, 3)))).relu().sum().backward()
print(A.grad)

# tensor([3.4641e+00, 1.6859e-07, 0.0000e+00], grad_fn=<LinalgSvdBackward0>)
# True
# True
# True
# tensor([[nan, nan, nan],
        [nan, nan, nan],
        [nan, nan, nan],
        [nan, nan, nan]])
```

Sorry to be pedantic but I'm not entirely clear from the onnx docs if this passes the criteria 

```
Differentiable inputs, which can be referenced in Gradient's xs attribute.
Differentiable outputs, which can be referenced in Gradient's y attribute.
```

williamberman: Re: how to compute SVD's backward, what level of detail are you looking for in the PR? SVD's backward is pretty gnarly and I'm probably not qualified to go too far in depth on it. pytorch has some docs on it here https://github.com/pytorch/pytorch/blob/master/torch/csrc/autograd/FunctionsManual.cpp#L3236-L3348 and uses https://j-towns.github.io/papers/svd-derivative.pdf jax's docs reference https://people.maths.ox.ac.uk/gilesm/files/NA-08-01.pdf for the full matrix case

wschin: Let's consider a simple function `relu`. It's not differentiable at `0` point, right? We just return a reasonable sub-differentiable value for `0`. We don't mark `relu` non-differentiable even `0` happens quite often. Similarly, marking `svd` as differentiable is reasonable to me. If a function is differentiable at most areas, I'd mark it as differentiable.

`Gradient` op in ONNX computes `[dy/dx0, dy/dx1, ..., dy/dxn]`. When Pytorch call `.backward()`, it computes `dout/din` for each op, so `svd`'s output can be referenced by `y` attribute and `svd`'s inputs can be referenced by `xs` attribute.

williamberman: Ah ok thank you so much, makes a ton of sense

williamberman: Actually I'm going to do a bit more reading on how the backward pass is calculated so I can add documentation about the numerical stability to the op doc. Will see if I pick up enough to add to PR

williamberman:
> @williamberman - Hi, I'm actually not at Schneider Electric anymore (hence the "SE" in my old name), I've switched to Databricks 🥳 .
> 
> So I didn't see you had tagged me in this. I can see @p-wysocki found some models which utilize SVD (thank you!).
> 
> I'm not aware of any other public models which utilize the SVD off the top of my head but internally to SE we were looking at using the SVD as part of the model to perform some de-noising on signals before classification.

Thank you for the context @coltonpeltier-db! Makes a ton of sense

onnx/defs/math/defs.cc
williamberman: @wschin I put together some documentation and example implementations of SVDs derivative in this notebook https://github.com/williamberman/svd-derivative/blob/main/svd-derivative.ipynb . Will add to PR description

onnx/backend/test/case/node/svd.py
```diff
@@ -0,0 +1,54 @@
...
+                for (dims, dims_name) in dimensions:
+                    A = np.random.randn(*dims)
+
+                    node_args: Any = {"inputs": ["A"]}
+                    numpy_args = {}
+                    name = f"test_svd_{dims_name}"
+
+                    if full_matrices is not None:
+                        node_args["full_matrices"] = full_matrices
+                        numpy_args["full_matrices"] = bool(full_matrices)
```
xadupre, line 29: Does it work without bool? In that case, the same dictionary could be used to hold parameters for the node and the numpy function.

onnx/defs/math/defs.cc
```diff
@@ -3474,4 +3474,205 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+The derivative of the full factorization is dependent on the reciprocal of each singular value and the
+reciprocal of the differences of the squares each pair of singular values.
+
+This causes numerical instability when,
+1. Any singular value is 0.
+2. Any two singular values are the same.
+3. Any two singular values are sufficiently close.
+)DOC";
+
+void SVDInferenceFunction(InferenceContext& ctx) {
```
xadupre, line 3526: I wonder if this function works if one dimension is null or specified with a letter.

onnx/backend/test/case/node/svd.py
williamberman: Good catch! Added

onnx/defs/math/defs.cc
williamberman: Thank you for the catch! I missed this was a possibility in the docs. 

I added checks for when the shape is not present and for computing the number of singular values when one or both of the non-batched input dimensions are not present and/or symbolic. 

I added tests here https://github.com/onnx/onnx/pull/4416/commits/bbed9b2389ea8f589aa579969c0bcc45399f80bd#diff-1109925d18b450cb60dcd1438f23475daad61de5eca71acdcc112b3aa85a7c0aR8612-R8794

xadupre (approved):

enuk1dze:
@williamberman Thank you for your work! I'm having trouble with exporting my model because of SVD right now. I hope this PR will be done.

One comment on PR though. Shouldn't  binary `.pb` files be in `git-lfs` of something?

For example, like that one `onnx/backend/test/data/node/test_svd_3d_partial_compute_uv_manually_set/test_data_set_0/output_1.pb`

williamberman:
> @williamberman Thank you for your work! I'm having trouble with exporting my model because of SVD right now. I hope this PR will be done.
> 
> One comment on PR though. Shouldn't binary `.pb` files be in `git-lfs` of something?
> 
> For example, like that one `onnx/backend/test/data/node/test_svd_3d_partial_compute_uv_manually_set/test_data_set_0/output_1.pb`

of course @enuk1dze , hope to have it merged soon! 

re: git lfs, since the generated protobufs are pretty small, I would say it's ok to commit them into vanilla git as I think they already are. However, that's probably up to the onnx maintainers

yuanyao-nv:
There are different flavors of numerical methods used in practice to compute SVD (direct vs iterative, deterministic vs stochastic) suited to matrices with different sizes and properties. Do we want to provide more info in the ONNX spec with respect to what method is used (and also the accuracy/tolerance in the case of iterative methods)?

gramalingam:
> There are different flavors of numerical methods used in practice to compute SVD (direct vs iterative, deterministic vs stochastic) suited to matrices with different sizes and properties. Do we want to provide more info in the ONNX spec with respect to what method is used (and also the accuracy/tolerance in the case of iterative methods)?

If they are going to produce different results, then the op should clarify which one is expected. What's the standard/default? I would assume that's what we want. If there is a demand for more than one of these, we may need an attribute to distinguish between them. But all of these should be driven by the use-cases (motivating examples/models).

yuanyao-nv:
I think a good analogy here is with matrix diagonalization, which also has many different methods suited to different scenarios. We don't yet have a diagonalization operator in ONNX yet.

I agree with @gramalingam that it should be driven by use cases. Unfortunately. the pytorch and tensorflow and doc pages also have no mention of what method they use, but the fact that they produce all singular values suggest a direct method is used. A quick search in the literature seems to suggest the industry standard for small SVD problems is a two-phase method: first reduce to bidiagonal form, then to diagonal form, with small variations for both phases.

So far, I don't think ONNX has expanded into linear algebra computations - this would be a first. So I think it'd be worthwhile thinking more carefully about how (perhaps also if) we should handle such operations. And if SVD is included, it would seem incomplete not to include other ops, such as diagonalization, and various matrix decompositions.

williamberman:
> Unfortunately. the pytorch and tensorflow and doc pages also have no mention of what method they use, but the fact that they produce all singular values suggest a direct method is used. A quick search in the literature seems to suggest the industry standard for small SVD problems is a two-phase method: first reduce to bidiagonal form, then to diagonal form, with small variations for both phases.

Mainly restating what I pulled out of the source and docs, but happy to do more digging to go into more detail if it's helpful :)

##### Tensorflow - CPU

TF CPU uses eigen and calls [bidiagonal divide and conquer](https://eigen.tuxfamily.org/dox/classEigen_1_1BDCSVD.html) which internally falls back to [jacobi method](https://eigen.tuxfamily.org/dox/classEigen_1_1JacobiSVD.html) for matrices with less than 16 cols.  

##### Tensorflow - GPU

TF GPU uses cuSOLVER and calls [gesvdj ](https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-gesvdj)(jacobi method) for batches of smaller than 32x32 matrices. Additionally, the matrices must be either square or be computing the full factorization for the jacobi method. See [source ](https://github.com/tensorflow/tensorflow/blob/16d4106ffaa71b02ba5d5b4961eef07e46ea4f0b/tensorflow/core/kernels/linalg/svd_op_gpu.cu.cc#L122) for the full condition.

Otherwise, TF GPU uses [gesvd](https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-gesvd) which uses QR algorithm.

##### Pytorch - CPU

Pytorch CPU uses lapack and calls [gesdd](https://www.smcm.iqfr.csic.es/docs/intel/mkl/mkl_manual/lse/functn_gesdd.htm) which uses divide and conquer.

##### Pytorch - GPU - MAGMA

When using magma, pytorch calls [gesdd](https://icl.utk.edu/projectsfiles/magma/doxygen/group__magma__gesdd.html) which uses divide and conquer.

##### Pytorch - GPU - cuSOLVER

When using cuSOLVER, pytorch tip actually lets you pass an option (`driver`) that lets you choose between gesvd (QR), gesvdj (jacobi), or [gesvda](https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-gesvda) (approximates decompositions of skinny matrices). If gesvdj or gesvda are chosen, the result is checked for 
convergence and it falls back to gesvd. The default behavior is to use gesvdj with the gesvd fallback. 

See the doc string in [source](https://github.com/pytorch/pytorch/blob/614d6f19e3d30cac0d77059e738d1f25d75eb408/torch/linalg/__init__.py#L1655-L1665).

The behavior in the [current stable release](https://github.com/pytorch/pytorch/blob/235f8db3d896e1c4810a2adc3c8a0996292793db/aten/src/ATen/native/cuda/linalg/BatchLinearAlgebraLib.cpp#L805) is to use gesvdj and fallback to gesvd without the driver option.


> I agree with @gramalingam that it should be driven by use cases. 

Unfortunately none of the example models that I found gave specifics on hard requirements they needed out of the SVD implementation they used. Deep rotation estimation released their model in both pytorch and tensorflow if that's a relevant datapoint.

williamberman:
Closing this PR as it seems svd and potentially related op types are outside of the current scope of the onnx spec :)

justinchuby:
Is this a good time to reopen this?

xadupre:
Sure.

PR #4421, opened 2022-08-08
to extend OptionalHasElement and OptionalGetElement to accept tensor and sequence types 

liqunfu:
**Description**
to extend OptionalHasElement and OptionalGetElement to accept tensor and sequence types 

**Motivation and Context**
Before the 2 optional ops only accept optional types. This limits its use in cases where input can be both optional and other types. This PR generalizes the 2 ops so it become more flexible.


onnx/backend/test/case/node/optionalhaselement.py
```diff
@@ -46,3 +46,18 @@ def export_empty() -> None:
...
                name='test_optional_has_element_empty')
+
+    @staticmethod
+    def export_tensor() -> None:
+        tensor = np.array([1, 2, 3, 4]).astype(np.float32)
+        tensor_type_proto = onnx.helper.make_tensor_type_proto(elem_type=onnx.TensorProto.FLOAT, shape=[4, ])
+        input_type_proto = onnx.helper.make_optional_type_proto(tensor_type_proto)
+        node = onnx.helper.make_node(
+            'OptionalHasElement',
+            inputs=['optional_input'],
```
gramalingam, line 57: Another test-case where `inputs=[]` would be useful.

How about test-cases for optional_get_element?

onnx/defs/optional/defs.cc
```diff
@@ -62,15 +70,17 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 )DOC";
 
 ONNX_OPERATOR_SET_SCHEMA(
     OptionalHasElement,
-    15,
+    18,
     OpSchema()
-        .SetDoc(OptionalHasElement_ver1_doc)
+        .SetDoc(OptionalHasElement_ver18_doc)
         .Input(0, "input", "The optional input.", "O")
```
gramalingam, line 84: The input should be made into an optional input.

onnx/backend/test/cmd_tools.py
```diff
@@ -48,6 +48,8 @@ def prepare_dir(path: str) -> None:
                     output_dir, f'test_data_set_{i}')
                 prepare_dir(data_set_dir)
                 for j, input in enumerate(inputs):
+                    if input is None:
```
gramalingam, line 51: Trying to understand this: this file is about *model* inputs/outputs, right, and not about *node/op* inputs/outputs. For model inputs/outputs there is no possibility of optional-input/output of kind 1 (static), right? I don't understand why this is being triggered. (May be some fix in the model-generator is required if it is adding an unnecessary model-input?)

gramalingam: I am worried things will break if a model has an input X, and there is no corresponding input-file in the test-folder.

onnx/backend/test/case/node/optionalhaselement.py
```diff
@@ -21,43 +21,45 @@ class OptionalHasElement(Base):
...
-            outputs=['output']
-        )
-        output = optional_has_element_reference_implementation(tensor)
-        expect(node, inputs=[tensor], outputs=[output],
-               input_type_protos=[input_type_proto],
-               name='test_optional_has_element_tensor')
+        # OptionalHasElement takes a tensor or optional as input
+        for input_type_proto in [tensor_type_proto, optional_type_proto]:
+            input_name_options = {
+                'empty': 'optional_input',
```
gramalingam, line 50: I think the combination of `input_type_proto == tensor_type_proto` and above (`'empty': 'optional_input'`) is not a valid combination to test? That might also be triggering the change to cmd_tools.py below?

onnx/numpy_helper.py
```diff
@@ -97,51 +97,53 @@ def from_array(arr: np.ndarray, name: Optional[str] = None) -> TensorProto:
...
-    try:
-        dtype = mapping.NP_TYPE_TO_TENSOR_TYPE[arr.dtype]
-    except KeyError:
-        raise RuntimeError(f"Numpy data type not understood yet: {str(arr.dtype)}")
-    tensor.data_type = dtype
-    tensor.raw_data = arr.tobytes()  # note: tobytes() is only after 1.9.
-    if sys.byteorder == "big":
-        # Convert endian from big to little
-        convert_endian(tensor)
+    if arr is not None:
```
gramalingam, line 103: Is this change still required (after eliminating that test-case)? Wondering why this change is needed.

gramalingam (approved):
LGTM, thanks!

Issue #4278, opened 2022-06-15
[engineering] Sort import with isort


justinchuby:
We should sort imports with isort to keep the import section clean

PR #4427, opened 2022-08-10
Format all python code with black and isort - take 2

justinchuby:
**Description**

- Format all python code with black and isort
- Add format checks in CI
- Skip additional generated files in isort

**Motivation and Context**

Standardize formatting

Fixes https://github.com/onnx/onnx/issues/4278


justinchuby:
@jcwchen I created this pr to set up the config and CI for the formatters. Please let me know when the approved changes are merged and I can run the actual formatting process

onnx/__init__.py
```diff
@@ -150,12 +167,12 @@ def load_model_from_string(s: bytes, format: Optional[Any] = None) -> ModelProto
...
     Returns:
         Loaded in-memory ModelProto
-    '''
+    """
     return _deserialize(s, ModelProto())
 
 
 def load_tensor_from_string(s: bytes, format: Optional[Any] = None) -> TensorProto:
-    '''
+    """
```
jcwchen, line 175: One thing irrelevant to this PR: is it possible to silent this kind of warning (Undefined variables from ONNX) from pylint? If not, shall we build ONNX first in this CI to let pylint understand ONNX? Building ONNX might be helpful for the next mypy check as well.

jcwchen (commented):
output_X.pb shouldn't been updated, right? The are updated because different platforms might produce different result from NumPy. To keep this PR simple, I would suggest not updating them. Thanks!

justinchuby:
Do we check for output_X.pb in the CI? I ran tools/update_doc.sh and got those. I will revert the changes

onnx/__init__.py
justinchuby: #4428

justinchuby:
Done

jcwchen (approved):
LGTM. Thanks for the quick PR: https://github.com/onnx/onnx/pull/4428. It indeed resolves a lot of unused variable warnings from onnx package.

justinchuby:
@gramalingam ready for your approval

gramalingam (approved):

PR #4428, opened 2022-08-10
[ci] Install ONNX before linting

justinchuby:
**Description**

Install ONNX before linting

**Motivation and Context**

#4427


jcwchen (approved):
LGTM. Let's forward this first and see whether it resolves the pylint issue in https://github.com/onnx/onnx/pull/4427.

PR #4448, opened 2022-08-17
fix resize shape inference

liqunfu:

### Description
scale can be nullptr but hasScaleInput is still true


onnx/defs/tensor/utils.cc
```diff
@@ -92,23 +92,16 @@ void resizeShapeInference(InferenceContext& ctx) {
...
 
   // If scales or sizes are an empty constant, assume it's not provided
-  if (scales && ParseData<float>(scales).empty()) {
-    hasScalesInput = false;
+  if (scales && ParseData<float>(scales).empty())
     scales = nullptr;
-  }
-  if (sizes && ParseData<int64_t>(sizes).empty()) {
-    hasSizesInput = false;
+  if (sizes && ParseData<float>(sizes).empty())
```
daquexian, line 101: Is it changed by mistake?

onnx/defs/tensor/utils.cc
```diff
@@ -92,23 +92,16 @@ void resizeShapeInference(InferenceContext& ctx) {
...
-  bool hasScalesInput = (2 < ctx.getNumInputs()) && (ctx.getInputType(2) != nullptr);
-  bool hasSizesInput = (3 < ctx.getNumInputs()) && (ctx.getInputType(3) != nullptr);
-
   const TensorProto* scales = 2 < ctx.getNumInputs() ? ctx.getInputData(2) : nullptr;
   const TensorProto* sizes = 3 < ctx.getNumInputs() ? ctx.getInputData(3) : nullptr;
 
   // If scales or sizes are an empty constant, assume it's not provided
-  if (scales && ParseData<float>(scales).empty()) {
-    hasScalesInput = false;
+  if (scales && ParseData<float>(scales).empty())
```
daquexian, line 99: I think it's better to always write {} to avoid unexpected results.

daquexian:
> scale can be nullptr but hasScaleInput is still true

Could you please elaborate a bit on it? Thanks!

liqunfu:
>

gramalingam:
hi @liqunfu : I don't understand the problem. A quick question first: is [this line](https://github.com/microsoft/onnxruntime/blob/ce01ed02da48cf25317eb2d4dfa4841012bf122c/onnxruntime/test/providers/cpu/tensor/resize_op_test.cc#L878) supposed to be `{}` instead of `{0}` ?

gramalingam:
In general, I am a bit concerned about the previous fix in https://github.com/onnx/onnx/pull/4388 also: I think it is important to keep the spec (the description of the op) and the code in sync. If such fixes are required, we should update the op's spec to clarify it too.

liqunfu:
> hi @liqunfu : I don't understand the problem. A quick question first: is [this line](https://github.com/microsoft/onnxruntime/blob/ce01ed02da48cf25317eb2d4dfa4841012bf122c/onnxruntime/test/providers/cpu/tensor/resize_op_test.cc#L878) supposed to be `{}` instead of `{0}` ?

It is not this. I added one test case which may explain the issue better.

onnx/defs/tensor/utils.cc
```diff
@@ -98,8 +109,11 @@ void resizeShapeInference(InferenceContext& ctx) {
   const TensorProto* scales = 2 < ctx.getNumInputs() ? ctx.getInputData(2) : nullptr;
   const TensorProto* sizes = 3 < ctx.getNumInputs() ? ctx.getInputData(3) : nullptr;
 
+  // specific to opset11 where scales input is not optional but can be a empty tensor
```
gramalingam, line 112: May be we should add an extra `int opset` parameter to the method, so that we can customize the behavior for each opset as needed.

liqunfu:
> In general, I am a bit concerned about the previous fix in #4388 also: I think it is important to keep the spec (the description of the op) and the code in sync. If such fixes are required, we should update the op's spec to clarify it too.

The fix #4388 did fix an issue in ModelZoo. The issue this PR is to fix is specific to opset11 where we cannot use empty string for scales because it is not optional.

onnx/defs/tensor/utils.cc
gramalingam: Is the empty tensor input supported only for opset11? Or, also later?

liqunfu: empty string is only supported since opset13. with opset11, 'scales' is not optional and thus cannot use empty string.
According to text description of scales in opset13+, I think empty tensor cannot be used to express "no scales input", only empty sting can be used for the purpose. 
Given that, I will use 'int opset' parameter for the method.

onnx/defs/tensor/utils.cc
```diff
@@ -84,7 +84,18 @@ void resizeShapeInferenceHelper(
   }
 }
 
-void resizeShapeInference(InferenceContext& ctx) {
+bool isScalesInputAnEmptyTensor(InferenceContext& ctx) {
```
gramalingam, line 87: This can be deleted now, right?

onnx/defs/tensor/utils.cc
```diff
@@ -236,6 +240,14 @@ void resizeShapeInference(InferenceContext& ctx) {
   } // nullptr != scales
 }
 
+void resizeShapeInference_opset13_to_18(InferenceContext& ctx) {
+  resizeShapeInferenceVersioned(ctx, 18);
+}
+
+void resizeShapeInference_opset11_to_12(InferenceContext& ctx) {
+  resizeShapeInferenceVersioned(ctx, 12);
```
gramalingam, line 248: nit: I think it is better to pass in the value 11, because the op is defined in opset version 11.

onnx/defs/tensor/utils.cc
```diff
@@ -236,6 +240,14 @@ void resizeShapeInference(InferenceContext& ctx) {
   } // nullptr != scales
 }
 
+void resizeShapeInference_opset13_to_18(InferenceContext& ctx) {
+  resizeShapeInferenceVersioned(ctx, 18);
```
gramalingam, line 244: nit: may be better to pass in value 13 instead of 18. 

Also the naming is a bit challenging for the last version, because it may apply to future versions also (like version 19). So, I wonder if it is better to just call them `resizeShapeInference_opset13` and `resizeShapeInference_opset11` ? The upper end is implicit and omitted from name.

gramalingam (approved):

Issue #4447, opened 2022-08-16
Checking if input present in InferenceContext


gramalingam:
ONNX allows for optional inputs. Inference methods of ONNX ops may need to check if an optional input is provided. The class `InferenceContext` does not provide a method to check for this. The method `getNumInputs` provides only a partial answer because ONNX allows for non-trailing optional inputs to be omitted: eg., `foo('x', '', 'y')` is a call where the first and third inputs are provided, but the second input is omitted.

I suggest that we add a new virtual method `hasInput(i)` to this interface with a default-implementation (for backward-compatibility), with the default-implementation defined as:
```cc
   return (getNumInputs() > i && getInputType(i) != nullptr);
```
which is the workaround currently used in various ops.

Eg., see [this example](https://github.com/onnx/onnx/blob/a5e72dc835a2d67b675fcb635f55222dd7078bbc/onnx/defs/math/defs.cc#L2818)

gramalingam:
Other places where it is used:

https://github.com/onnx/onnx/blob/a5e72dc835a2d67b675fcb635f55222dd7078bbc/onnx/defs/math/defs.cc#L3267

https://github.com/onnx/onnx/blob/a5e72dc835a2d67b675fcb635f55222dd7078bbc/onnx/defs/quantization/defs.cc#L49

https://github.com/onnx/onnx/blob/a5e72dc835a2d67b675fcb635f55222dd7078bbc/onnx/defs/tensor/utils.cc#L95

https://github.com/onnx/onnx/blob/a5e72dc835a2d67b675fcb635f55222dd7078bbc/onnx/defs/tensor/utils.cc#L96

and the ongoing PR: https://github.com/onnx/onnx/pull/4389

etc.

p-wysocki:
It's a great idea, the previous method for checking for inputs was inconvenient. I'll apply it to #4389 as well.

PR #4451, opened 2022-08-17
Add hasInput method to InferenceContext

gramalingam:
See https://github.com/onnx/onnx/issues/4447 for a discussion.

### DIFF
--- a/onnx/defs/math/defs.cc
+++ b/onnx/defs/math/defs.cc
@@ -2815,7 +2815,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           // Set the output dimension to match the dft_length on the axis.
           // If onesided this will be adjusted later on...
           const TensorProto* dft_length = nullptr;
-          if (ctx.getNumInputs() >= 2 && ctx.getInputType(1) != nullptr) {
+          if (ctx.hasInput(1)) {
             dft_length = ctx.getInputData(1);
             if (dft_length == nullptr) {
               // If we cannot read the dft_length, we cannot infer shape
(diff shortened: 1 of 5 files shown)


p-wysocki (approved):
LGTM, I think it's a great addition.

onnx/defs/math/defs.cc
```diff
@@ -2815,7 +2815,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           // Set the output dimension to match the dft_length on the axis.
           // If onesided this will be adjusted later on...
           const TensorProto* dft_length = nullptr;
-          if (ctx.getNumInputs() >= 2 && ctx.getInputType(1) != nullptr) {
```
liqunfu, line 2818: How to specify a no optional input with empty tensor so that the input is treated as no-input? e.g. scales input of Resize opset11.

gramalingam:
@daquexian can you please take a look? This requires the approval from someone in operator SIG. Thanks!

jcwchen (commented):
Great improvement! Shall we also apply `hasInput` in
1. https://github.com/onnx/onnx/blob/main/onnx/defs/tensor/utils.cc#L95-L99
2. https://github.com/onnx/onnx/blob/main/onnx/defs/quantization/old.cc#L41

Besides, do you think it is also useful if we can have another common function for `ctx.getNumInputs() > x && hasInputShape(ctx, x)`? For example: https://github.com/onnx/onnx/blob/main/onnx/defs/nn/old.cc#L70-L77 It can be done by another PR if needed.

gramalingam:
> Great improvement! Shall we also apply `hasInput` in
> 
> 1. https://github.com/onnx/onnx/blob/main/onnx/defs/tensor/utils.cc#L95-L99
> 2. https://github.com/onnx/onnx/blob/main/onnx/defs/quantization/old.cc#L41
> 
> Besides, do you think it is also useful if we can have another common function for `ctx.getNumInputs() > x && hasInputShape(ctx, x)`? For example: https://github.com/onnx/onnx/blob/main/onnx/defs/nn/old.cc#L70-L77 It can be done by another PR if needed.

Added the first two, thanks! As for the next suggestion, `hasInputShape` already works this way, so we don't need a new function I think. (The extra condition `ctx.getNumInputs() > x` is not needed.)

jcwchen (approved):

linkerzhang (approved):
Good improvement!

daquexian (approved):
LGTM :)

Issue #4422, opened 2022-08-09
Make checker with full_check consistent for giving ModelProto or model path


jcwchen:
# Feature Request

### System information
ONNX version (you are using): latest main branch

### What is the problem that this feature solves?
Making the checker behavior consistent can prevent confusion. And checker should not modify the model in place.

### Describe the alternatives you have considered
Keep the same behavior as before.

### Describe the feature

Current behavior:
check_model+full_check with model path (Python and C++ API): the original ONNX model will be overrode by the inferred model through model path
check_model+full_check with ModelProto (Python API): the original model won't be updated
check_model+full_check with ModelProto (C++ API): the original ModelProto will be overrode by the inferred model

IMO, we should make checker with full_check consistent for giving ModelProto or model path: do not update the given models. Checker should just check the model without changing it. Related discussion: https://github.com/onnx/onnx/pull/4386#discussion_r937159690.

### Will this influence the current api?
Yes. checker won't update the given model anyway.

### Feature Area
Which area in ONNX does this impact?  (*e.g. model usage, backend, best practices, converters, shape_inference, version_converter, training, test, operators*): checker

### Are you willing to contribute it (Y/N):
Y

PR #4456, opened 2022-08-18
Make check_model with full_check always not modify the model in place

jcwchen:
### Description

- To prevent confusion, make check_model always not modify the model.
- Use `LoadProtoFromPath` to simplify duplicate code.

After this PR, all check_model with full_check Python/C++ APIs won't update the model in place. 

### Motivation and Context

Fixes https://github.com/onnx/onnx/issues/4422. Follow-up work for https://github.com/onnx/onnx/pull/4386.


daquexian (approved):
LGTM

PR #4458, opened 2022-08-19
primary ops to function milestone 1

liqunfu:
### Description
change primary ops to function milestone 1

### Motivation and Context
to reduce number of primary ops


onnx/defs/nn/defs.cc
```diff
@@ -1829,6 +1829,33 @@ bias. The formula of this operator is: If x < -lambd, y = x + bias;
...
 )DOC";
 
+static float shrink_default_lambd = 0.5f;
+static float shrink_default_bias = 0.0f;
+
+bool BuildContextDependentFunctionBodyShrink(
+    const FunctionBodyBuildContext& ctx,
+    const OpSchema& schema,
+    FunctionProto& functionProto) {
+  float lambd = ctx.getAttribute("lambd") != nullptr ? ctx.getAttribute("lambd")->f() : shrink_default_lambd;
```
gramalingam, line 1839: One general question / comment: I think, in the long term, it would be better to move towards simple function (instead of context-dependent function). I think we could do that by using attribute-references for such attributes.

The main issue, I think, is the use of default-values for these attributes. I think it may be a good idea to make it the responsibility of the function-inliner to use the default-values when inlining a call where the attribute is not explicitly provided.

Does that make sense? I realize that the existing implementation of inliner will need to be updated to do this, but just want to discuss the idea first.

onnx/defs/nn/defs.cc
```diff
@@ -1829,6 +1829,33 @@ bias. The formula of this operator is: If x < -lambd, y = x + bias;
...
+  float bias = ctx.getAttribute("bias") != nullptr ? ctx.getAttribute("bias")->f() : shrink_default_bias;
+  FunctionBuilder builder(functionProto);
+  builder.Const("lambd", std::vector<float>{lambd}).Const("bias", std::vector<float>{bias}).Add(R"(
+    LambdCastInput = CastLike(lambd, input)
+    NegLsmbda = Neg (LambdCastInput)
```
gramalingam, line 1844: nit: I guess NegLsmbda => NegLambda (typo from original probably).

onnx/defs/math/defs.cc
```diff
@@ -725,7 +822,16 @@ ONNX_OPERATOR_SET_SCHEMA(
...
              "tensor(int64)"},
             "Constrain input and output types to float/int tensors.")
-        .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput));
+        .TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput)
+        .FunctionBody(R"ONNX(
+        {
+          Zero = Constant <value = float {0.0}>()
+          ZeroCast = CastLike(Zero, X)    
+          XLessThan = Less (X, ZeroCast)
+          SlopMulX = Mul (slop, X)
```
gramalingam, line 831: `slop` => `slope` ?

onnx/defs/nn/defs.cc
linkerzhang: +1 on moving towards to simple function.

onnx/defs/math/defs.cc
```diff
@@ -879,6 +967,33 @@ specified by the inputs 'min' and 'max'. They default to
 numeric_limits::lowest() and numeric_limits::max(), respectively.
 )DOC";
 
+bool BuildContextDependentFunctionBodyClip(
+    const FunctionBodyBuildContext& ctx,
+    const OpSchema& schema,
+    FunctionProto& functionProto) {
+  bool has_min = ctx.getInputType(1);
```
gramalingam, line 974: I suggest using `ctx.hasInput(1)` instead. Same for line below.

onnx/defs/nn/defs.cc
```diff
@@ -2529,8 +2546,7 @@ ONNX_OPERATOR_SET_SCHEMA(
           // Output Y has same shape as X
           // Outputs Mean and InvStdDev have shape: [d[0], ..., d[axis-1], 1, ..., 1]
           FunctionBuilder builder(functionProto);
-          builder.AddOpset("", 16)
-              .Const("FloatEpsilon", ToTensor<float>(epsilon))
+          builder.Const("FloatEpsilon", ToTensor<float>(epsilon))
```
gramalingam, line 2549: Why change this? Seems unrelated to the main PR. Anyway: we can decide on this after the decision about opsets in function-definitions.

gramalingam:
Looks good to me. The only open question is whether we need to bump the opset-version number for some of these ops, or can we avoid that?

onnx/defs/nn/defs.cc
liqunfu: I discussed with the original author, this AddOpset("", 16) was added due to a typo. It was detected by the change that use the function's opset version if specified in test case generation.

onnx/backend/test/case/node/__init__.py
```diff
@@ -302,6 +305,22 @@ def merge(
             outputs=outputs_vi,
         )
         kwargs["producer_name"] = "backend-test"
+
+        # replace opset versions with what are specified in funtion proto
```
justinchuby, line 309:
```suggestion
        # replace opset versions with what are specified in function proto
```

onnx/defs/math/defs.cc
```diff
@@ -381,7 +390,18 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .FunctionBody(R"ONNX(
+          {
+            Alpha = Constant <value_float: float = @alpha>()
+            AlphaCast = CastLike (Alpha, X)
+            Zero = Constant <value = float {0.0}>()
+            ZeroCast = CastLike(Zero, X)
+            XLessThanZero = Less(X, ZeroCast)
+            AlphaMulX = Mul (AlphaCast, X)
+            Y = Where (XLessThanZero, AlphaMulX, X)
+		      }
```
gramalingam, line 403: Niot: indentation seems off in above line, and similarly in other defs below.

gramalingam:
LGTM, just had a nit-comment about indentation.

liqunfu:
> LGTM, just had a nit-comment about indentation.

thanks @gramalingam, I just made the correction.

onnx/backend/test/case/node/__init__.py
```diff
@@ -301,6 +304,22 @@ def merge(
             outputs=outputs_vi,
         )
         kwargs["producer_name"] = "backend-test"
+
+        # replace opset versions with what are specified in function proto
+        for opset_import in func_opset_import:
+            if "opset_imports" in kwargs:
```
gramalingam, line 310: Wouldn't it be simpler to have the if-condition outside the for-loop and change the else-statement to update `kwargs["opset_imports"] = func_opset_import` in 1 step? It looks a bit confusing, even though I guess it works.

liqunfu: The purpose is to append if "opset_imports" does not exist, otherwise update. So it cannot be outside of the loop.

liqunfu: took a second look, you are right. I made the correction. Thanks!

gramalingam (approved):

PR #4470, opened 2022-08-25
Use filesystem to load filename to prevent encoding issues on Windows

jcwchen:
### Description
- On Windows, use filesystem to load filename to prevent encoding issues (e.g., Chinese locale).
- Extend existing path functions to accept std::wstring as well.
- Add comments for requiring C++17 or higher for Windows.

### Motivation and Context
Follow-up work of https://github.com/onnx/onnx/pull/4400. Here is another security issue when loading external tensors on Windows:
On Linux, a path string is either encoded by Unicode(UTF-16), or a multiple byte encoding like GB18030/ISO-8859-1/…, and usually it is not UTF-8.  As there are so many encodings, usually we need to convert the string to UTF-16 first and process it with wchar_t. Otherwise, when you search the char ‘\’, you may accidently misclassify a half of a Chinese character to ‘\’. (It can happen in other encodings too). On Linux, the same thing could happen too, but we could claim on Linux we don’t support locales other than xxx.UTF-8. But for Windows, we need to handle other encodings. Using filesystem could be a solution.


jcwchen:
cc @onnx/sig-archinfra-approvers, @snnn: this PR is ready for review. Thanks!

onnx/checker.cc
```diff
@@ -127,6 +140,31 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
     for (const StringStringEntryProto& entry : tensor.external_data()) {
       if (entry.has_key() && entry.has_value() && entry.key() == "location") {
         has_location = true;
+#ifdef _WIN32
+        const fs::path windows_path(entry.value());
```
snnn, line 144: This would not work. 
Please manually use https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-multibytetowidechar function to convert entry.value() to a wide char string, and during the conversion you need to set code page to CP_UTF8. Because the ONNX standard expects all strings are UTF-8.

snnn: You could try if it works for paths with Chinese characters.

onnx/checker.cc
```diff
@@ -127,6 +140,31 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
...
+        std::wstring relative_path = clean_relative_path(windows_path.wstring());
+        // Check that normalized relative path starts with "../" or "..\" on windows.
+        if (relative_path.rfind(L".." + w_k_preferred_path_separator, 0) == 0) {
```
snnn, line 147: Why do you use rfind? Why search from end to start while the comment says you want to check if the string starts with something? 

And please replace == 0 with == std::wstring::npos.

onnx/common/path.cc
```diff
@@ -9,99 +9,85 @@
...
-    // Do nothing on linux.
-    return;
-  }
-
-  for (size_t i = 0; i < path.size(); i++) {
-    if (is_path_separator(path[i]) && path[i] != preferred_sep) {
-      path[i] = preferred_sep;
-    }
+#define CLEAN_RELATIVE_PATH(string_type, char_type, separator, dot)                                           \
+  template <>                                                                                                 \
```
snnn, line 22: Would you please add a comment of what does this function do?

onnx/common/path.h
```diff
@@ -8,17 +8,58 @@
...
+
+  for (size_t i = 0; i < path.size(); i++) {
+    if (is_path_separator(path[i]) && path[i] != preferred_sep) {
+      path[i] = preferred_sep;
+    }
+  }
+}
+template <typename STRING>
+STRING clean_relative_path(const STRING& path);
+std::string clean_relative_path(const char* path);
```
snnn, line 53: This part looks very confusing. Initially I thought the second one was specialization of the first one, Then I realized it missed "template" keyword. 
https://docs.microsoft.com/en-us/cpp/cpp/explicit-specialization-of-function-templates?view=msvc-170 

So, why for std::string you use template but for C style strings you use overloading? Why not using the same approach for both?

onnx/common/path.h
```diff
@@ -8,17 +8,58 @@
...
+
+#ifdef _WIN32
+std::wstring clean_relative_path(const wchar_t* path);
+inline std::wstring utf8str_to_wstring(const std::string& utf8str) {
+  int size_required = MultiByteToWideChar(CP_UTF8, 0, utf8str.c_str(), (int)utf8str.size(), NULL, 0);
+  std::wstring ws_str(size_required, 0);
```
snnn, line 59: Please use L'\0'

snnn: utf8str.size() doesn't include the nul-terminator of the string. 
So does `size_required`.
When you construct ws_str with length=size_required, will the underlaying implementation automatically add a '\0' at the end? I'm not sure.

onnx/checker.cc
jcwchen: Good catch. I should further use `utf8str_to_wstring` (including `MultiByteToWideChar`) to convert the std::string into std::wstring first.

onnx/checker.cc
jcwchen: I followed the [original implementation](https://github.com/onnx/onnx/blob/8133ea176ca6aeb28a26a6f3e8e0687ed3c99a8e/onnx/checker.cc#L132), but I agreed it looks confusing. To prevent confusion, the code should reflect what the comment said. Thus, I changed it into:
```
if (relative_path.find(L".." + w_k_preferred_path_separator, 0) != std::string::npos) {
...
```

onnx/common/path.cc
jcwchen: Sounds good. Just added in path.h.

onnx/common/path.h
jcwchen: Thanks for the pointer. I added it because we need to explicitly define a function for char[] input (directly giving raw string in double quote)

To prevent confusion, I also templatized this function and added comments. Now I added it as:
```
// std::string in template cannot be recognized as char[]
// Therefore explicitly define char[] to handle char[] input
template <typename CHAR, typename std::size_t N>
std::basic_string<CHAR> clean_relative_path(const CHAR (&path)[N]) {
  return clean_relative_path(std::basic_string<CHAR>(path));
}
```

onnx/common/path.h
jcwchen: Good point. I checked the document and it does not explicitly say we need to handle end of string on our own so I tried to rum some experiments as you suggested. I initialize some Chinese+English characters as std::wstring and then use `std::codecvt_utf8` to convert them from std::wstring into std::string. Next, I used the `utf8str_to_wstring` here to convert these std::string back to std::wstring. In result, the twice converted std::wstring is identical as the one before conversion. Therefore, I think the original method should be good. Please let me know if you still have concern. Thanks!

liqunfu (approved):

onnx/common/path.h
```diff
@@ -8,17 +8,67 @@
...
+  if (k_preferred_path_separator == "\\") {
+    return c == '\\' || c == '/';
+  }
+
+  return c == k_preferred_path_separator[0];
+}
+
+// To always use backward slash as path separator on Windows
+template <typename STRING>
+void normalize_separator(STRING& path) {
```
snnn, line 39: On Windows, please use 
```
path canonical(const path& pval, const path& base = current_path());
path canonical(const path& pval, error_code& ec);
path canonical(const path& pval, const path& base, error_code& ec);
```
from std::filesystem

onnx/common/path.h
```diff
@@ -8,17 +8,67 @@
...
+
+// Clean up relative path when there is ".." in the path, e.g.: a/b/../c -> a/c
+// std::string or std::wstring
+template <typename STRING>
+STRING clean_relative_path(const STRING& path);
+
+// std::string in template cannot be recognized as char[]
+// Therefore explicitly define char[] to handle char[] input
+template <typename CHAR, typename std::size_t N>
+std::basic_string<CHAR> clean_relative_path(const CHAR (&path)[N]) {
```
snnn, line 61: What is `const CHAR (&path)[N]` ? I'm not familiar with this grammar .

CMakeLists.txt
```diff
@@ -58,7 +58,8 @@ else()
   set(Protobuf_USE_STATIC_LIBS ON)
 endif()
 
-# Set C++11 as standard for the whole project
+# Required to use /std:c++14 or higher on Windows
```
snnn, line 61: I suggest using C++17 for Windows builds. Then you can use std::string_view in your following code. Every VC++ compiler on Windows supports C++17.

onnx/common/path.cc
```diff
@@ -9,99 +9,79 @@
...
 
-  // Use 1 separator in path.
-  normalize_separator(out);
+PATH_JOIN(std::string, k_preferred_path_separator);
+CLEAN_RELATIVE_PATH(std::string, char, k_preferred_path_separator, ".");
 
-  return out;
-}
+#ifdef _WIN32
+PATH_JOIN(std::wstring, w_k_preferred_path_separator);
```
snnn, line 83: Please try to not implement such things by yourself. Use the functions in std::filesystem instead. Because it is very very difficult to handle all the cases.  Path names on Windows have many different styles.

onnx/common/path.h
```diff
@@ -8,17 +8,67 @@
...
 #include <string>
+#ifdef _WIN32
+#include <windows.h>
+#endif
 
 namespace ONNX_NAMESPACE {
 
 #ifdef _WIN32
 const std::string k_preferred_path_separator = "\\";
+const std::wstring w_k_preferred_path_separator = L"\\";
```
snnn, line 19: Please use "constexpr const char*" or "constexpr const wchar_t*" so that they can be initialized at compile time. We should avoid having dynamic initializers whenever possible.

snnn: And on Windows you can use "path::make_preferred" function

onnx/common/path.h
```diff
@@ -8,17 +8,67 @@
...
 #else // POSIX
 const std::string k_preferred_path_separator = "/";
 #endif
 
-std::string path_join(const std::string& origin, const std::string& append);
-void normalize_separator(std::string& path);
-std::string clean_relative_path(const std::string& path);
+template <typename STRING>
+STRING path_join(const STRING& origin, const STRING& append);
```
snnn, line 25: On Windows you can use `path::operator/=`

onnx/common/path.h
```diff
@@ -8,17 +8,67 @@
...
+
+template <typename CHAR>
+bool is_path_separator(CHAR c) {
```
snnn, line 28: You should not need to have this function for Windows. Instead, you should iterate `path` as a container. 

Like this:
```C++
// filesystem_path_example.cpp
// compile by using: /EHsc /W4 /permissive- /std:c++17 (or later)
#include <string>
#include <iostream>
#include <sstream>
#include <filesystem>

using namespace std;
using namespace std::filesystem;

wstring DisplayPathInfo()
{
    // This path may or may not refer to an existing file. We are
    // examining this path string, not file system objects.
    path pathToDisplay(L"C:/FileSystemTest/SubDir3/SubDirLevel2/File2.txt ");

    wostringstream wos;
    int i = 0;
    wos << L"Displaying path info for: " << pathToDisplay << endl;
    for (path::iterator itr = pathToDisplay.begin(); itr != pathToDisplay.end(); ++itr)
    {
        wos << L"path part: " << i++ << L" = " << *itr << endl;
    }

    wos << L"root_name() = " << pathToDisplay.root_name() << endl
        << L"root_path() = " << pathToDisplay.root_path() << endl
        << L"relative_path() = " << pathToDisplay.relative_path() << endl
        << L"parent_path() = " << pathToDisplay.parent_path() << endl
        << L"filename() = " << pathToDisplay.filename() << endl
        << L"stem() = " << pathToDisplay.stem() << endl
        << L"extension() = " << pathToDisplay.extension() << endl;

    return wos.str();
}

int main()
{
    wcout << DisplayPathInfo() << endl;
    // wcout << ComparePaths() << endl; // see following example
    wcout << endl << L"Press Enter to exit" << endl;
    wstring input;
    getline(wcin, input);
}
```

onnx/common/path.h
snnn: LGTM.

onnx/common/path.h
```diff
@@ -8,17 +8,67 @@
...
+// std::string in template cannot be recognized as char[]
+// Therefore explicitly define char[] to handle char[] input
+template <typename CHAR, typename std::size_t N>
+std::basic_string<CHAR> clean_relative_path(const CHAR (&path)[N]) {
+  return clean_relative_path(std::basic_string<CHAR>(path));
+}
+
+#ifdef _WIN32
+inline std::wstring utf8str_to_wstring(const std::string& utf8str) {
+  int size_required = MultiByteToWideChar(CP_UTF8, 0, utf8str.c_str(), (int)utf8str.size(), NULL, 0);
```
snnn, line 67: Please check the range of utf8str.size() before forcing converting it to int. size_t could be bigger than int.

onnx/common/path.h
jcwchen: This is the way to create template for array of T I believe: https://stackoverflow.com/questions/33234979/how-to-write-a-template-function-that-takes-an-array-and-an-int-specifying-array

onnx/common/path.h
jcwchen: To confirm, are you suggesting that for Windows we should directly use `canonical` instead of using `clean_relative_path` to clean relative path on other platforms?

jcwchen: I think existing lexically_normal+make_prefered from filesystem should cover `clean_relative_path` and `normalize_separator` for Windows. I will update this PR soon (remove self implementation and utilize existing functions whenever possible) and let you know. Thanks for the pointer!

snnn: > To confirm, are you suggesting that for Windows we should directly use `canonical` instead of using `clean_relative_path` to clean relative path on other platforms?

Yes.

onnx/common/path.h
```diff
@@ -8,17 +8,43 @@
...
+#include <windows.h>
+#include <filesystem>
+#include "onnx/checker.h"
+#endif
 
 namespace ONNX_NAMESPACE {
 
 #ifdef _WIN32
-const std::string k_preferred_path_separator = "\\";
+constexpr char* k_preferred_path_separator = "\\";
```
snnn, line 20: You missed a const. 
constexpr const char*

snnn: And  why it is a string instead of a single char?

onnx/common/path.h
jcwchen: Previously I tried C-string pointer and it didn't work. It seems required to be defined like it for using template with char[] string (i.e., "this is a string"). Still, now Windows and POSIX have different implementation so this PR doesn't need template anymore.

onnx/common/path.h
jcwchen: Thanks for catching this unintended change. Just updated.

jcwchen:
@snnn Thank you for the reviews. I think I should solve all of them. Please take another round of review. Meanwhile, I will try to validate this PR and https://github.com/onnx/onnx/pull/4400 in ORT and see whether there is any issue in advance.

cc @postrational IIRC, you mentioned you are interested in this topic in our last operator-sig meeting. Feel free to review. Thank you!

onnx/checker.cc
```diff
@@ -127,9 +129,33 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
     for (const StringStringEntryProto& entry : tensor.external_data()) {
       if (entry.has_key() && entry.has_value() && entry.key() == "location") {
         has_location = true;
+#ifdef _WIN32
+        std::wstring relative_path = clean_relative_path(utf8str_to_wstring(entry.value()));
+        // Check that normalized relative path starts with ".." on Windows.
```
snnn, line 134: The comment doesn't match the code. You are checking if the string contains "..", not "starts"

onnx/test/cpp/common_path_test.cc
```diff
@@ -8,63 +8,59 @@
...
+  EXPECT_EQ(clean_relative_path("abc/def"), "abc/def");
+  EXPECT_EQ(clean_relative_path("a/b/c"), "a/b/c");
+  EXPECT_EQ(clean_relative_path("."), ".");
+  EXPECT_EQ(clean_relative_path(".."), "..");
+  EXPECT_EQ(clean_relative_path("../.."), "../..");
+  EXPECT_EQ(clean_relative_path("../../abc"), "../../abc");
   // Remove leading slash
-  EXPECT_EQ(clean_relative_path("/abc"), fix_sep("abc"));
-  EXPECT_EQ(clean_relative_path("/"), fix_sep("."));
+  EXPECT_EQ(clean_relative_path("/abc"), "abc");
```
snnn, line 29: I don't understand this test case. It seems that we expect this function would force converting every abs path to relative path? But why "/abc" becomes "abc"? If the current working directory is "/", it makes sense. But, what if we are not in the root dir? 

And on Windows abs path **may** start with things like "C:" , not "/". Then in that case what this function `clean_relative_path` would do?

onnx/checker.cc
jcwchen: Updated. Thanks!

onnx/test/cpp/common_path_test.cc
jcwchen: Good question... I also found it is confusing when involving absolute path. Actually ONNX IR only allows using relative path to represent the location for external tensors: https://github.com/onnx/onnx/blob/main/docs/IR.md#external-tensor-data.

Therefore, I further add absolute path check in checker, add comment for clean_relative_path that it cannot work with absolute path, and remove absolute path tests in `[common_path_test.cc](https://github.com/onnx/onnx/pull/4470/files/903b9edb3b6b1a93abf3acebc4b1d1497980c3bf#diff-9a1899e6bc8ebf64e337ebc384ed395441ea42ba716a906906cb9a2150bb2732)` to prevent confusions.

snnn (approved):

snnn (approved):

jcwchen (commented):
ORT is good with this update so I will forward this PR soon. Thanks for the reviews!

Issue #4472, opened 2022-08-25
Install from pip fails due to spaces in path


Vargol:
# Bug Report
### Describe the bug
I'm trying t run pip install onnx
and I get the following error

```
      [  9%] Built target onnxifi_wrapper
      /private/var/folders/mq/w85wcgsx58g9xjz_66mj8bk40000gn/T/pip-install-19y8i46a/onnx_08b9cb443db945b9b10e5844dcd22bbb/.setuptools-cmake-build/tools/protoc-gen-mypy.sh: line 4: /Volumes/Sabrent: No such file or directory
      --mypy_out: protoc-gen-mypy: Plugin failed with status code 127.
      gmake[2]: *** [CMakeFiles/gen_onnx_proto.dir/build.make:74: onnx/onnx-ml.pb.cc] Error 1
      gmake[1]: *** [CMakeFiles/Makefile2:113: CMakeFiles/gen_onnx_proto.dir/all] Error 2
      gmake: *** [Makefile:136: all] Error 2

The full path to my python venv is
/Volumes/Sabrent Media/Documents/Source/Python/pytorch_env/
```

The full error message is...

```
Collecting onnx
  Using cached onnx-1.12.0.tar.gz (10.1 MB)
  Preparing metadata (setup.py) ... done
Requirement already satisfied: numpy>=1.16.6 in ./lib/python3.10/site-packages (from onnx) (1.23.1)
Requirement already satisfied: protobuf<=3.20.1,>=3.12.2 in ./lib/python3.10/site-packages (from onnx) (3.19.4)
Requirement already satisfied: typing-extensions>=3.6.2.1 in ./lib/python3.10/site-packages (from onnx) (4.3.0)
Building wheels for collected packages: onnx
  Building wheel for onnx (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [124 lines of output]
      fatal: not a git repository (or any of the parent directories): .git
      running bdist_wheel
      running build
      running build_py
      running create_version
      running cmake_build
[...]

PR #4473, opened 2022-08-26
Make ONNX installable even with spaces in path

jcwchen:
### Description
Make ONNX installable even with spaces in path. Put double quotes around several installed paths.

### Motivation and Context
Fixes https://github.com/onnx/onnx/issues/4472. Currently ONNX is not installed on Mac when the installed path including spaces.


CMakeLists.txt
```diff
@@ -470,7 +470,7 @@ if(BUILD_ONNX_PYTHON)
     set_target_properties(onnx_cpp2py_export
                           PROPERTIES LINK_FLAGS "-undefined dynamic_lookup")
     target_link_libraries(onnx_cpp2py_export
-                          PRIVATE -Wl,-force_load,$<TARGET_FILE:onnx>)
+                          PRIVATE -Wl,-force_load,"$<TARGET_FILE:onnx>")
```
jcwchen, line 473: Note: I only put double quotes around $<TARGET_FILE:onnx> for Mac. Other platforms like Windows and Ubuntu originally work fine from my experiments and further put double quotes will bring other issues. Therefore I didn't put double quotes for other platforms.

daquexian: LGTM. Could you please also write the note above this line in CMakeLists.txt so that future readers can know the context?

daquexian (approved):

CMakeLists.txt
jcwchen: Good idea. Just added. Thanks!

Issue #4438, opened 2022-08-12
Support `Split` even when dimension is not evenly splittable


xuzijian629:
Currently, [Split](https://github.com/onnx/onnx/blob/v1.12.0/docs/Operators.md#split) operator splits the tensor to **equal sized** tensors if `split` input is not given.

I think this restriction is not very important. ~For example, pytorch's split also supports the case when dim cannot be evenly split. https://pytorch.org/docs/1.12/generated/torch.split.html?highlight=split#torch.split~

UPD: Torch's corresponding function was https://pytorch.org/docs/stable/generated/torch.tensor_split.html

Can we update the operator definition to support this case?

gramalingam:
Sounds reasonable to me.

xuzijian629:
Should I implement this as Split-18 at the same time with https://github.com/onnx/onnx/pull/4389, or is it better to implement this as Split-19?

gramalingam:
Adding @p-wysocki, who is working on Split-18 ... I think if everyone agrees on the extension, it would be better to have it together in Split-18 itself (if possible; otherwise, we can always do Split-19 as a fallback).

p-wysocki:
Sounds good, I'll add it to Split-18.

p-wysocki:
PR: https://github.com/onnx/onnx/pull/4481

p-wysocki:
The PR has been merged. Please reopen the issue if needed.

PR #4481, opened 2022-09-01
Add Split-18 - num_outputs attribute, support for uneven tensor splitting

p-wysocki:
### Description
- new `num_outputs` attribute
- support for splitting tensors into uneven chunks added

### Motivation and Context
- solves https://github.com/onnx/onnx/issues/4188
- solves https://github.com/onnx/onnx/issues/4438

Previously discussed in https://github.com/onnx/onnx/pull/4389


onnx/backend/test/case/node/split.py
```diff
@@ -113,7 +117,165 @@ def export_default_values() -> None:
...
+            node,
+            inputs=[input, split],
+            outputs=[y for y in expected_outputs],
+            name="test_split_variable_parts_default_axis_opset13",
+            opset_imports=[onnx.helper.make_opsetid("", 13)],
+        )
+
+    @staticmethod
+    def export_zero_size_splits_opset13() -> None:
+        input = np.array([]).astype(np.float32)
```
gramalingam, line 143: Can you clarify the input/output shapes explicitly to help the reader? I assume that the shapes are [0] (a 1-dimensional tensor with dimension-size=0).

gramalingam:
LGTM, thanks! Just one minor comment above to clarify shapes for the zero-size tensor.

jcwchen (commented):
Perhaps you can try to cherry-pick this PR: https://github.com/onnx/onnx/pull/4202 and see whether it helps to show the exact errors without seg fault.

If that PR is indeed helpful, I would say we should forward that PR for better debugging.

jcwchen (commented):
Sorry, I think my PR https://github.com/onnx/onnx/pull/4202 now needs a few updates for being compatible with https://github.com/onnx/onnx/pull/4470. We can try another way: to faster test the unknown failure in your local, perhaps you can try:
```
python setup.py install  # build ONNX with your PR
python onnx/backend/test/cmd_tools.py generate-data
git diff --exit-code --diff-filter=ADR -- . :!onnx/onnx-data.proto 
```
I suspect the number of input_*.pb or output_*.pb might be incorrect in your PR. If that is true, the commands above should show some files. If not, there might be other issues I don't know in input_*.pb or output_*.pb.

p-wysocki:
@jcwchen I ran the commands, they did not show any files. However, after removing `test_driver.cc` changes the CI started passing. I have no idea why it's passing now, maybe the issue was sporadic or has been fixed on master in the meantime. Thanks for the help!

jcwchen (changes requested):
I might figure out why previous CI has failed: some added test model files are not needed. I don't see where they are generated. For instance, `test_uneven_split` and `test_uneven_split_opset18`. I would suggest you clean all of added test files and then use `python onnx/backend/test/cmd_tools.py generate-data` to generated them again. Since related test was just removed due to recent ONNXIFI deprecation, I will add a similar check in CI to prevent future issues like it.

Another suggestion would be making test files naming convention consistent. For names of Split related tests, please use `test_split_XXX_XXX`. IMO, it will be helpful for users to search related test files.

p-wysocki:
@jcwchen I applied your suggestions, thanks! For some reason I assumed that script that generates data also removes the old files. I also approved your PR introducing a CI check for that, it sounds helpful. Test names also have been aligned.

jcwchen (commented):
Thank you for the updates! Once you have updated `TestCoverage.md` by the updated test scripts , I think this PR is ready to go.

jcwchen (approved):

gramalingam:
> I might figure out why previous CI has failed: some added test model files are not needed. I don't see where they are generated. For instance, `test_uneven_split` and `test_uneven_split_opset18`. I would suggest you clean all of added test files and then use `python onnx/backend/test/cmd_tools.py generate-data` to generated them again. Since related test was just removed due to recent ONNXIFI deprecation, I will add a similar check in CI to prevent future issues like it.
> 
> Another suggestion would be making test files naming convention consistent. For names of Split related tests, please use `test_split_XXX_XXX`. IMO, it will be helpful for users to search related test files.

I think it would be helpful to update the `AddNewOp` document (may be as a helpful Troubleshooting tip at the end) to remind people to delete the old folder before generating test-cases. (If the CI can catch it, that would be great too.)

onnx/defs/tensor/defs.cc
```diff
@@ -603,15 +603,16 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         }));
 
-static const char* Split_ver13_doc =
-    R"DOC(Split a tensor into a list of tensors, along the specified
-'axis'. Lengths of the parts can be specified using input 'split'.
-Otherwise, the tensor is split to equal sized parts.
+static const char* Split_ver18_doc =
+    R"DOC(Split a tensor into a list of tensors, along the specified 'axis'.
+Either input 'split' or the attribute 'num_outputs' should be specified, but not both.
+If the attribute 'num_outputs' is specified, then the tensor is split into equal sized parts.
```
gramalingam, line 609: Need to update above line by adding the line from below ``If the tensor is not evenly splittable, the last chunk will be smaller.``

jcwchen:
> I think it would be helpful to update the AddNewOp document (may be as a helpful Troubleshooting tip at the end) to remind people to delete the old folder before generating test-cases. (If the CI can catch it, that would be great too.)

Sounds good. I will add sentence about it in AddNewOp.md in this PR as well: https://github.com/onnx/onnx/pull/4514. Also note that that PR will let CI to catch this kind of issue. Perhaps we can forward that PR first and have that verify this PR.

> Or, may be the test-case generator should delete the folder? (Of course, it could be a bit risky in case the user has some other file there ... may be warn the user if there is some extra files there perhaps.)

Yes, but it seems possible that users would like to keep existing ones for some testing purposes? Therefore, to keep behavior consistent, I introduced --clean flag for generate_data for CI use (by default it is False so it won't clean original node directory automatically).

onnx/defs/tensor/defs.cc
p-wysocki: Good catch, done.

gramalingam (approved):

jcwchen (approved):
It passed the enhanced CI 💯  Thank you @p-wysocki for this PR and waiting! I will merge it soon.

Issue #4432, opened 2022-08-10
Reference implementation of ops


gramalingam:
It would be useful to have a reference implementation of the ONNX ops, to complement the existing spec.

The existing [guidelines for adding a new op to the spec](https://github.com/onnx/onnx/blob/main/docs/AddNewOp.md#step-2-submit-pr-) already recommends a reference implementation of the op in Python, and this has been followed for a number of recent ops. However, there are still two limitations that would be worth addressing.

* Many of the ops added to the spec earlier do not have reference implementations.
* The reference implementations are not available in a well-organized way that makes it easy for users to call them.

The proposal is to enable users to write code as below:
```py
import onnx.opset15 as op

def my_function(x, y, bias):
   t1 = op.MatMul(x, y)  # calls operation MatMul (of opset 15) with inputs x and y
   t2 = op.Add(t1, bias)
   t3 = op.Elu(t2, alpha=0.4)  # calls operation Elu with input t2 and value 0.4 for attribute alpha.
   return t3
```

Specifically, this requires ensuring that all reference implementations use a consistent API, and type-representation for values (eg., numpy arrays for tensors), and that they be organized into a module/class like object that allows us to invoke the ops belonging to a specific opset version.

xadupre:
About that, I started a while a python/numpy/C++ backend to easily debug onnx graph and I implemented many operators in python/numpy/C++ which you can find here: https://github.com/sdpython/mlprodict/tree/master/mlprodict/onnxrt/ops_cpu. I keep track of the [coverage](http://www.xavierdupre.fr/app/mlprodict/helpsphinx/backends/backend_python.html). Currently, 88% of the backend tests are ok. You can see on the page which ones are passing and failing.

gramalingam:
Closing as we have one in place.

PR #4483, opened 2022-09-02
[WIP] Reference implementation for onnx operators (python runtime for onnx)

xadupre:
### Description and Motivation

This PR addresses issue #4432. It proposes a reference implementation for onnx operators. This runtime can also be used to test function body. Currently, there is no way to check the body of a function validates the backend tests. This can also be used to validate other runtime or add more backend tests.

This work comes from a python runtime I wrote to easily check the onnx model built with a converter. onnxruntime is a fast runtime but it cannot be easily use to debug a model, like display intermediate results. This is another goal of this runtime.
When this PR is complete the runtime should validate 93% of the backend test.

### DIFF
--- a/onnx/runtime/aionnx/_op.py
+++ b/onnx/runtime/aionnx/_op.py
@@ -1,0 +1,476 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import pprint
+import numpy as np
+from onnx import TensorProto, GraphProto
+from onnx.defs import get_all_schemas_with_history
+
+
+def _build_schemas():
+    res = {}
+    for schema in get_all_schemas_with_history():
+        # Multiple version can coexist. The last one is kept.
+        if schema.name in res:
+            if schema.since_version > res[schema.name].since_version:
+                # We keep the most recent one.
+                res[schema.name] = schema
+        else:
+            res[schema.name] = schema
+        res[schema.name + '_' + str(schema.since_version)] = schema
+    return res
+
+
+_schemas = _build_schemas()
+
+
+class RuntimeTypeError(RuntimeError):
+    """
+    Raised when a type of a variable is unexpected.
+    """
+    pass
+
+
+class DefaultNone:
+    """
+    Default value for parameters when the parameter is not set
+    but the operator has a default behaviour for it.
+    """
+    pass
+
+
+class RefAttrName:
+    """
+    Implements a link between a parameter of a function
+    and an attribute in node.
+
+    :param name: name of the input
+    """
+
+    def __init__(self, name):
+        self.name = name
+
+    def __repr__(self):
+        "usual"
+        return f"{self.__class__.__name__}({self.name!r})"
+
+
+class OpRun:
+    """
+    Ancestor to all operators in this subfolder.
+    The runtime for every node can checked into
+    `ONNX unit tests
+    <https://github.com/onnx/onnx/tree/master/onnx/backend/test/case/node>`_.
+
+    :param onnx_node: :epkg:`onnx` node
+    :param log_function: function used to log information while
+        executing the onnx graph
+    """
+    _attribute_conversion_functions = {
+        TensorProto.FLOAT: lambda att: np.float32(att.f),
+        TensorProto.INT64: lambda att: np.float32(att.i),
+    }
+
+    def __init__(self, onnx_node, log_function):
+        self.onnx_node = onnx_node
+        self.log_function = log_function
+        self._schema = _schemas[onnx_node.op_type]
+        self._load_attributes()
+
+    def _extract_attribute_value(self, att):
+        """
+        Converts an attribute value into a python value.
+        """
+        if att.type in OpRun._attribute_conversion_functions:
+            return OpRun._attribute_conversion_functions[att.type](att)
+        raise NotImplementedError(
+            f"Unable to convert attribute {att.name!r} type {att.type!r} "
+            f"from node type {self.onnx_node.op_type!r}, "
+            f"domain {self.onnx_node.domain!r}.")
+
+    def _load_attributes(self):
+        "Checks and loads attributes."
+        for att in self.onnx_node.attribute:
+            name = att.name
+            value = self._extract_attribute_value(att)
+            setattr(self, name, value)
+
+        if self.onnx_node.op_type not in {'Constant'}:
+            for k, v in self._schema.attributes.items():
+                if not hasattr(self, k) and getattr(v, 'required', True):
+                    raise RuntimeError(
+                        "Attribute '{}' is expected based on ONNX specifications "
+                        "for node '{}' and options {}.".format(
+                            k, onnx_node.op_type, pprint.pformat(options)))
+
+    @staticmethod
+    def local_inputs(graph):
+        """
+        Returns all varibles not registered as inputs and not produced by
+        an node inside the graph. This inputs are part of the context
+        existing in the graph calling this one.
+        """
+        if not isinstance(graph, GraphProto):
+            raise TypeError(
+                f"Unexpected type {type(graph)!r}.")
+        local = set()
+        known = set()
+        for init in graph.initializer:
+            known.add(init.name)
+        for init in graph.input:
+            known.add(init.name)
+        for node in graph.node:
+            for o in node.output:
+                known.add(o)
+            for i in node.input:
+                if i not in known:
+                    local.add(i)
+        return list(local)
+
+    @property
+    def input(self):
+        "Returns node attribute `input`."
+        return self.onnx_node.input
+
+    @property
+    def output(self):
+        "Returns node attribute `output`."
+        return self.onnx_node.output
+
+    @property
+    def op_type(self):
+        "Returns node attribute `op_type`."
+        return self.onnx_node.op_type
+
+    @property
+    def domain(self):
+        "Returns node attribute `domain`."
+        return self.onnx_node.domain
+
+    def need_context(self):
+        """
+        Tells the runtime if this node needs the context
+        (all the results produced so far) as it may silently access
+        one of them (operator Scan, If, Loop).
+        The default answer is `False`.
+        """
+        return False
+
+    def __str__(self):
+        atts = [self.__class__.__name__ + '(',
+                f"    op_type={self.onnx_node.op_type}"]
+        for k, v in sorted(self.__dict__.items()):
+            if k in {'desc', 'onnx_node'}:
+                continue
+            if 'a' <= k[0] <= 'z' and k[-1] != '_':
+                atts.append(f'    {k}={v},')
+        atts.append(')')
+        return "\n".join(atts)
+
+    def _run(self, *args, attributes=None, **kwargs):
+        """
+        Should be overwritten.
+        Parameter *attributes* is used by functions.
+        """
+        raise NotImplementedError(
+            "Method '_run' or 'to_python' should be overwritten for operator %s."
+            "" % self.__class__.__name__)
+
+    def run(self, *args, **kwargs):  # pylint: disable=E0202
+        """
+        Calls method ``_run``, catches exceptions,
+        displays a longer error message.
+        """
+        try:
+            res = self._run(*args, **kwargs)
+        except TypeError as e:
+            raise TypeError(
+                "Issues with types {} (operator {}).".format(
+                    ", ".join(str(type(_)) for _ in args),
+                    self.__class__.__name__)) from e
+        except AttributeError as e:
+            raise AttributeError(
+                "Issues with types {} (operator {}).".format(
+                    ", ".join(str(type(_)) for _ in args),
+                    self.__class__.__name__)) from e
+        return res
+
+    @property
+    def args_default(self):
+        """
+        Returns the list of arguments as well as
+        the list of parameters with the default values
+        (close to the signature).
+        """
+        inps = []
+        if hasattr(self, 'atts'):
+            for k, v in self.atts.items():  # pylint: disable=E1101
+                if isinstance(v, (list, tuple, dict)) and len(v) == 0:
+                    v = None
+                inps.append(f'{k}={v!r}')
+        return inps
+
+    @property
+    def args_default_modified(self):
+        """
+        Returns the list of modified parameters.
+        """
+        if not hasattr(self, 'atts'):
+            return None
+
+        inps = []
+        for k, v in self.atts.items():  # pylint: disable=E1101
+            val = getattr(self, k, None)
+            if isinstance(val, np.ndarray) and isinstance(v, list):
+                val = list(val)
+            try:
+                if val != v:
+                    inps.append(f'{k}={val!r}')
+            except ValueError as e:
+                raise ValueError(
+                    f"Unexpected value for v={v!r} and val={val!r}.") from e
+        return inps
+
+    @property
+    def args_optional(self):
+        """
+        Returns the list of optional arguments.
+        """
+        inps = []
+        if hasattr(self, 'optional_inputs'):
+            for k, v in self.optional_inputs.items():  # pylint: disable=E1101
+                inps.append(f'{k}={v!r}')
+        return inps
+
+    @property
+    def args_mandatory(self):
+        """
+        Returns the list of optional arguments.
+        """
+        if hasattr(self, 'mandatory_inputs'):
+            return self.mandatory_inputs  # pylint:  disable=E1101
+        return None
+
+    @property
+    def atts_value(self):
+        "Returns all parameters in a dictionary."
+        if hasattr(self, 'atts'):
+            return {k: getattr(self, k)
+                    for k in self.atts}  # pylint: disable=E1101
+        return None
+
+
+class OpRunUnary(OpRun):
+    """
+    Ancestor to all unary operators in this subfolder.
+    Checks that inputs type are the same.
+    """
+
+    def __init__(self, onnx_node, logging_function):
+        OpRun.__init__(self, onnx_node, logging_function)
+
+    def run(self, x, attributes=None):  # pylint: disable=E0202,W0221
+        """
+        Calls method ``_run``, catches exceptions,
+        displays a longer error message.
+        Supports only unary operators.
+        """
+        try:
+            res = self._run(x, attributes=attributes)
+        except TypeError as e:
+            raise TypeError(
+                "Issues with types {} (binary operator {}).".format(
+                    ", ".join(str(type(_)) for _ in [x]),
+                    self.__class__.__name__)) from e
+        return res
+
+
+class OpRunArg(OpRunUnary):
+    """
+    Ancestor to all unary operators in this subfolder
+    and which produces position of extremas (ArgMax, ...).
+    Checks that inputs type are the same.
+    The class must have attributes *axis*, *keepdim*.
+    """
+
+    def __init__(self, onnx_node, logging_function):
+        OpRunUnary.__init__(self, onnx_node, logging_function)
+        if not hasattr(self, 'keepdims'):
+            raise AttributeError(
+                "Attribute 'keepdims' is missing.")
+        if not hasattr(self, 'axis'):
+            raise AttributeError(
+                "Attribute 'axis' is missing.")
+
+    def run(self, x, attributes=None):  # pylint: disable=E0202
+        """
+        Calls method ``OpRunUnary.run``, catches exceptions,
+        displays a longer error message.
+        """
+        res = OpRunUnary.run(self, x, attributes=attributes)
+        if res[0].dtype != np.int64:
+            raise RuntimeTypeError(
+                "Output type mismatch: should be '{}' != output '{}' "
+                "(operator '{}')".format(
+                    np.int64, res[0].dtype, self.__class__.__name__))
+        return res
+
+    def _run_no_checks_(self, x, attributes=None):  # pylint: disable=W0221
+        return OpRunUnary.run(self, x, attributes=attributes)
+
+
+class OpRunUnaryNum(OpRunUnary):
+    """
+    Ancestor to all unary and numerical operators
+    in this subfolder. Checks that inputs type
+    are the same.
+    """
+
+    def __init__(self, onnx_node, logging_function):
+        OpRunUnary.__init__(self, onnx_node, logging_function)
+
+    def run(self, x, attributes=None):  # pylint: disable=E0202
+        """
+        Calls method ``OpRunUnary.run``, catches exceptions,
+        displays a longer error message.
+        Checks that the result is not empty.
+        """
+        res = OpRunUnary.run(self, x, attributes=attributes)
+        if len(res) == 0 or res[0] is None:
+            return res
+        if not isinstance(res[0], list) and res[0].dtype != x.dtype:
+            raise RuntimeTypeError(
+                "Output type mismatch: input '{}' != output '{}' "
+                "(operator '{}')".format(
+                    x.dtype, res[0].dtype, self.__class__.__name__))
+        return res
+
+    def _run_no_checks_(self, x, attributes=None):  # pylint: disable=W0221
+        return OpRunUnary.run(self, x, attributes=attributes)
+
+
+class OpRunBinary(OpRun):
+    """
+    Ancestor to all binary operators in this subfolder.
+    Checks that inputs type are the same.
+    """
+
+    def __init__(self, onnx_node, logging_function):
+        OpRun.__init__(self, onnx_node, logging_function)
+
+    def run(self, x, y, attributes=None):  # pylint: disable=E0202,W0221
+        """
+        Calls method ``_run``, catches exceptions,
+        displays a longer error message.
+        Supports only binary operators.
+        """
+        if x is None or y is None:
+            raise RuntimeError(
+                f"x and y have different dtype: {type(x)} != {type(y)} ({type(self)})")
+        if x.dtype != y.dtype:
+            raise RuntimeTypeError(
+                "Input type mismatch: {} != {} (operator '{}', shapes {}, {})".format(
+                    x.dtype, y.dtype, self.__class__.__name__,
+                    x.shape, y.shape))
+        try:
+            res = self._run(x, y, attributes=attributes)
+        except (TypeError, ValueError) as e:
+            raise TypeError(
+                "Issues with types {} (binary operator {}).".format(
+                    ", ".join(str(type(_)) for _ in [x, y]),
+                    self.__class__.__name__)) from e
+        return res
+
+    def _run_no_checks_(self, x, y, attributes=None):  # pylint: disable=W0221
+        """
+        Calls method ``_run``.
+        """
+        try:
+            res = self._run(x, y, attributes=attributes)
+        except TypeError as e:
+            raise TypeError(
+                "Issues with types {} (binary operator {}).".format(
+                    ", ".join(str(type(_)) for _ in [x, y]),
+                    self.__class__.__name__)) from e
+        return res
+
+
+class OpRunBinaryComparison(OpRunBinary):
+    """
+    Ancestor to all binary operators in this subfolder
+    comparing tensors.
+    """
+
+    def __init__(self, onnx_node, logging_function):
+        OpRunBinary.__init__(self, onnx_node, logging_function)
+
+
+class OpRunBinaryNum(OpRunBinary):
+    """
+    Ancestor to all binary operators in this subfolder.
+    Checks that inputs type are the same.
+    """
+
+    def __init__(self, onnx_node, logging_function):
+        OpRunBinary.__init__(self, onnx_node, logging_function)
+
+    def run(self, x, y, attributes=None):  # pylint: disable=E0202
+        """
+        Calls method ``OpRunBinary.run``, catches exceptions,
+        displays a longer error message.
+        """
+        res = OpRunBinary.run(self, x, y, attributes=attributes)
+        if res[0].dtype != x.dtype:
+            raise RuntimeTypeError(
+                "Output type mismatch: {} != {} or {} (operator '{}')"
+                " type(x)={} type(y)={}".format(
+                    x.dtype, res[0].dtype, y.dtype,
+                    self.__class__.__name__, type(x), type(y)))
+        return res
+
+    def _run_no_checks_(self, x, y, attributes=None):  # pylint: disable=W0221
+        return OpRunBinary._run_no_checks_(self, x, y, attributes=attributes)
+
+
+class OpRunBinaryNumpy(OpRunBinaryNum):
+    """
+    *numpy_fct* is a binary numpy function which
+    takes two matrices.
+    """
+
+    def __init__(self, numpy_fct, onnx_node, logging_function):
+        OpRunBinaryNum.__init__(self, onnx_node, logging_function)
+        self.numpy_fct = numpy_fct
+
+    def _run(self, a, b, attributes=None):  # pylint: disable=W0221
+        return (self.numpy_fct(a, b), )
+
+
+class OpRunReduceNumpy(OpRunUnaryNum):
+    """
+    Implements the reduce logic.
+    It must have a parameter *axes*.
+    """
+
+    def __init__(self, onnx_node, logging_function):
+        if ('noop_with_empty_axes' not in expected_attributes and
+                'axes' not in expected_attributes):
+            raise RuntimeError(
+                "Parameter 'axes' is expected but not found in {} "
+                "from class {}".format(expected_attributes, type(self)))
+        if (expected_attributes.get('noop_with_empty_axes', 0) and
+                (expected_attributes['axes'] is None or
+                    len(expected_attributes['axes']) == 0)):
+            raise RuntimeError(
+                "Parameter 'axes' cannot be empty as {} (noop_with_empty_axes=1) "
+                "from class {}".format(expected_attributes, type(self)))
+        OpRunUnaryNum.__init__(self, onnx_node, logging_function)
+        if isinstance(self.axes, np.ndarray):  # pylint: disable=E0203
+            if (len(self.axes.shape) == 0 or  # pylint: disable=E0203,E1101
+                    self.axes.shape[0] == 0):  # pylint: disable=E0203,E1101
+                self.axes = None
+            else:
+                self.axes = tuple(self.axes)
+        elif self.axes in [[], tuple()]:  # pylint: disable=E0203
+            self.axes = None
+        elif isinstance(self.axes, list):  # pylint: disable=E0203
+            self.axes = tuple(self.axes)
--- a/onnx/runtime/aionnx/_op.py
+++ b/onnx/runtime/aionnx/_op.py
@@ -73,7 +73,10 @@ class OpRun:
     def __init__(self, onnx_node, log_function):
         self.onnx_node = onnx_node
         self.log_function = log_function
-        self._schema = _schemas[onnx_node.op_type]
+        if onnx_node.op_type in _schemas:
+            self._schema = _schemas[onnx_node.op_type]
+        else:
+            self._schema = None
         self._load_attributes()
 
     def _extract_attribute_value(self, att):
@@ -94,7 +97,7 @@ class OpRun:
             value = self._extract_attribute_value(att)
             setattr(self, name, value)
 
-        if self.onnx_node.op_type not in {'Constant'}:
+        if self._schema and self.onnx_node.op_type not in {'Constant'}:
             for k, v in self._schema.attributes.items():
                 if not hasattr(self, k) and getattr(v, 'required', True):
                     raise RuntimeError(
@@ -257,6 +260,33 @@ class OpRun:
             return {k: getattr(self, k)
                     for k in self.atts}  # pylint: disable=E1101
         return None
+
+class OpFunction(OpRun):
+    """
+    Runs a custom function.
+    """
+    def __init__(self, onnx_node, log_function, impl=None):
+        if impl is None:
+            raise RuntimeError(
+                f"impl cannot be None for node type {onnx_node.op_type!r} "
+                f"from domain {onnx_node.domain!r}.")
+        OpRun.__init__(self, onnx_node, log_function)
+        self.impl_ = impl
+
+    def _run(self, *inputs):
+        if len(self.impl_.input_names) != len(inputs):
+            raise RuntimeError(
+                f"Mismatch lengths between the number of inputs {len(inputs)} "
+                f"and the expected number of inputs {len(self.impl_.inputs)} "
+                f"for node {self.op_type!r} from domain {self.domain!r}.")
+        feeds = {name: value for name, value in zip(self.impl_.input_names, inputs)}
+        results = self.impl_.run(None, feeds)
+        if len(self.impl_.output_names) != len(results):
+            raise RuntimeError(
+                f"Mismatch lengths between the number of outputs {len(results)} "
+                f"and the expected number of outputs {len(self.impl_.output_names)} "
+                f"for node {self.op_type!r} from domain {self.domain!r}.")
+        return tuple(results)
 
 
 class OpRunUnary(OpRun):
--- a/onnx/test/runtime_test.py
+++ b/onnx/test/runtime_test.py
@@ -1,0 +1,212 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import unittest
+from typing import Callable, List, Optional, Tuple
+import numpy as np  # type: ignore
+from numpy.testing import assert_almost_equal
+from onnx import parser, checker, ModelProto, TensorProto
+from onnx.numpy_helper import from_array
+from onnx.helper import (
+    make_model, make_node, make_graph,
+    make_tensor_value_info, make_opsetid)
+from onnx.checker import check_model
+
+try:
+    import onnx.runtime as rt
+except ImportError:
+    # will disappear in the final version of the PR.
+    import os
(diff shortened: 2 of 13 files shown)


onnx/runtime/aionnx/_op.py
```diff
@@ -0,0 +1,198 @@
...
+import numpy as np  # type: ignore
+
+from ...onnx_pb import NodeProto
+from ..op_run import OpRun, RuntimeTypeError
+
+
+class OpRunUnary(OpRun):  # pylint: disable=W0223
+    """
+    Ancestor to all unary operators in this subfolder.
+    Checks that inputs type are the same.
```
gramalingam, line 14: Don't understand this. Unary operators have only one input, right?

gramalingam:
A couple of questions.

(a) Can we enable users to do something like below:
```py
from ... import Celu
x = np.array(...)
y = Celu.eval(x, alpha=0.5)
```

The idea is to enable users to execute/evaluate single-op on inputs+attributes, by exposing a simple python-function interface not involving any "proto" stuff. I think it can be done without much effort. I use the name `eval` since various names like `run` or `_run` are already in use in the Op interfaces.

(b) I suspect that a simple implementation of the above will construct a NodeProto / AttributeProto to construct an instance of `Celu` and run it. Ideally, however, it should be the other way around: namely, the `run` methods should call the `eval` method. But I don't know if this will be much of an effort to do. But I can see that some ops already have such a structure in place, for example `_ArgMax._run` calls `_argmax` ... this seems the preferable form. Here, `_argmax` exposes a simple python-function interface (without any protos involved).

onnx/runtime/aionnx/_op.py
xadupre: The comment was wrong. I updated it. It checks that input and output types are the same.

xadupre:
(a) Method eval was not implemented but it was easy to do. It is now done. See below.
(b) In most cases, a method `run` which calls `eval` or a method `eval` which creates an onnx node and then method `run` can be implemented. With the current design, second option is a lot easier to implement. I chose this way because some operators are more efficient if some precomputation is done such as operator using subgraphs or operator defined as function.

```python
import numpy as np
from onnx.runtime.aionnx._op_list import Celu

x = np.array([[0, 1], [-1, 2]], dtype=np.float32)
y = Celu.eval(x, alpha=0.5)
```

Or:

```python
import numpy as np
from onnx.runtime.aionnx import load_op

Celu = load_op("", "Celu")  # domain is ""
x = np.array([[0, 1], [-1, 2]], dtype=np.float32)
y = Celu.eval(x, alpha=0.5)
```

And:

```python
import numpy as np
from onnx.runtime.aionnx._op_list import Celu

onnx_node = Celu.make_node(alpha=0.5)
```

onnx/runtime/op_run.py
```diff
@@ -0,0 +1,463 @@
...
+        node = cls.make_node(n_inputs, n_outputs, **kwargs)
+        run_params = dict(verbose=verbose, log=log_function)
+        cl = cls(node, run_params)
+        return cl
+
+    @classmethod
+    def eval(
+        cls,
+        *args: List[Any],
+        n_outputs: Optional[int] = None,
```
gramalingam, line 409: nit: may be useful to give it a name like `__n_outputs` to avoid conflicting with any ONNX op attribute-name. Same for `verbose`.

onnx/test/runtime_test.py
```diff
@@ -0,0 +1,1174 @@
...
+        )
+        onnx_model = make_model(
+            graph, opset_imports=[make_opsetid("", 16), make_opsetid("experimental", 1)]
+        )
+        graph_conv = make_graph([node], "g", [X, W], [Y1])
+        onnx_model_conv = make_model(graph_conv, opset_imports=[make_opsetid("", 16)])
+        sess = rt.Inference(onnx_model)
+
+        try:
+            import onnxruntime as ort
```
gramalingam, line 1099: I am a bit unclear why we have tests of ort here?

gramalingam:
A generic comment: I just realized that `onnx.runtime` is going to cause confusion vs. `onnxruntime`. I think we may be better off calling this a `reference implementation` ... of course, we may need a suitable abbreviation for that. Perhaps `ri` ? `refimpl` ? May be we can discuss a name over tomorrow's call.

onnx/test/runtime_backend_test.py
```diff
@@ -0,0 +1,609 @@
+# SPDX-License-Identifier: Apache-2.0
+# type: ignore
+# pylint: disable=C0415,R0912,R0913,R0914,R0915,W0640,W0703
+
```
gramalingam, line 4: A few lines documenting the scope of this file would help. My understanding is that this is for the backend\test\case\node tests? Does it also include backend\test\case\model tests?

onnx/runtime/aionnx/op_average_pool.py
```diff
@@ -0,0 +1,252 @@
+# SPDX-License-Identifier: Apache-2.0
+# pylint: disable=W0221,R0913,R0914
+
+import itertools
+from typing import Optional, Tuple
+
+import numpy  # type: ignore
```
justinchuby, line 7: `import numpy as np` to follow its convention?

onnx/runtime/aionnx/op_sequence_insert.py
```diff
@@ -0,0 +1,43 @@
...
+
+from typing import Any, List
+
+import numpy as np  # type: ignore
+
+from ..op_run import OpRun
+
+
+def sequence_insert_reference_implementation(
+    sequence: List[Any], tensor: np.ndarray, position: np.ndarray = None
```
justinchuby, line 12:
```suggestion
    sequence: Union[np.ndarray, Sequence], tensor: np.ndarray, position: np.ndarray = None
```

may be this since it can be np.ndarray?

onnx/runtime/op_run.py
justinchuby: Is it possible to pass in a dict for attributes instead?

onnx/test/runtime_backend_test.py
```diff
@@ -0,0 +1,609 @@
...
+import os
+import unittest
+
+import numpy as np
+from numpy import object_ as dtype_object
+from numpy.testing import assert_almost_equal  # type: ignore
```
justinchuby, line 10: Looks like `assert_allclose` is recommended: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html#numpy.testing.assert_almost_equal.

I remember these functions do broadcasting. Do we want to wrap them and check shape and dtype as well?

onnx/test/runtime_backend_test.py
```diff
@@ -0,0 +1,609 @@
...
+    @property
+    def name(self):
+        "Returns the test name."
+        return os.path.split(self.folder)[-1]
+
+    def __len__(self):
+        "Returns the number of tests."
+        return len(self.tests)
+
+    def _compare_results(self, index, i, e, o, decimal=None):
```
justinchuby, line 192: Use longer names for parameters?

onnx/test/runtime_test.py
```diff
@@ -0,0 +1,1174 @@
...
+            a = np.array([1, 1], dtype=np.float32)
+            b = np.array([11], dtype=np.float32)
+            expected = f(x, a, b)
+            sess = rt.Inference(lr)
+            last_node = sess.rt_nodes_[-1]
+            self.assertEqual(last_node.__class__.__name__, "Clip_11")
+            got = sess.run(None, {"X": a, "A": a, "B": b})[0]
+            assert_almost_equal(expected, got)
+
+    def test_inference_lr_clip_6(self):
```
justinchuby, line 278: Doesn't have to be this PR - Looks like much of these tests can be parameterized: https://pypi.org/project/parameterized/

onnx/test/runtime_test.py
```diff
@@ -0,0 +1,1174 @@
...
+                W = np.zeros((1, 1, 3, 3), dtype=np.float32)
+                W[0, 0, :, :] = np.minimum(2 ** np.arange(9).reshape((3, -1)), 256)
+
+                B = np.array([[[[0]]]], dtype=np.float32)
+                expected = sess1.run(None, {"X": X, "W": W, "B": B})[0]
+                got = sess2.run(None, {"X": X, "W": W, "B": B})[0]
+                assert_almost_equal(expected, got)
+
+    def test_qlinearconv(self):
+        try:
```
justinchuby, line 929: We can skip this test if ort is not installed. Example https://github.com/pytorch/pytorch/blob/c7c2578f93fbfad5f769543848642a16b6071756/torch/testing/_internal/common_utils.py#L1204-L1211

justinchuby (commented):
LOVE that we will have a reference runtime! ❤️

onnx/test/runtime_test.py
xadupre: They can be removed. I'm just checking that the runtime I write and onnxruntime have the same results. Backend tests are not enough to understand what an implementation does wrong.

onnx/test/runtime_backend_test.py
xadupre: Yes, it runs all the backend test with the python runtime and checks that it produces the expected outputs.

onnx/runtime/aionnx/op_average_pool.py
xadupre: I'll do that at the end. I used many lines of codes I wrote a while ago. I usually use numpy for my teachings because it is less confusing for students.

onnx/test/runtime_backend_test.py
xadupre: Done

onnx/runtime/op_run.py
xadupre: If possible, I'd like to keep verbose as it is. I doubt an operator would use it as this parameter means the same thing accross all packages. Same goes for n_outputs. The meaning of the parameter is quite clear. If an operator uses it in its definition, it should have the same meaning. I don't mind using an attribute instead of `**kwargs`. Any particular reason? `make_node` is also using `**kwargs`.

onnx/runtime/aionnx/op_sequence_insert.py
xadupre: done

onnx/test/runtime_test.py
xadupre: parametized is not a dependency. I suggest another PR.

onnx/test/runtime_test.py
xadupre: done

onnx/runtime/op_run.py
justinchuby: > I don't mind using an attribute instead of `**kwargs`. Any particular reason? `make_node` is also using `**kwargs`.

Just to avoid naming conflicts if there is a need.

onnx/test/runtime_backend_test.py
```diff
@@ -0,0 +1,632 @@
...
+                if e.shape != o.shape:
+                    raise AssertionError(
+                        f"Output {i} of test {index} in folder {self.folder!r} failed "
+                        f"(shape mismatch {e.shape} != {o.shape})."
+                    )
+            else:
+                self._compare_results(index, i, e, o, atol=atol, rtol=rtol)
+
+
+def enumerate_onnx_tests(series, fct_filter=None):
```
liqunfu, line 318: not to local from file system, utilize onnx.backend.test.case.node.collect_testcases()

onnx/test/runtime_backend_test.py
```diff
@@ -0,0 +1,632 @@
...
+        done = 0
+        for te in enumerate_onnx_tests("node", lambda folder: folder == "test_abs"):
+            self.assertIn(te.name, repr(te))
+            self.assertGreater(len(te), 0)
+            te.run(TestOnnxBackEnd.load_fct, TestOnnxBackEnd.run_fct)
+            done += 1
+        self.assertEqual(done, 1)
+
+    def common_test_enumerate_onnx_tests_run(
+        self, valid, verbose=0, rtol=None, atol=None, check_ort_mismath=False
```
liqunfu, line 393: verbose use an enum type

onnx/test/runtime_test.py
```diff
@@ -0,0 +1,1493 @@
...
+    return wrapper
+
+
+def skip_if_no_torch(fn):
+    @wraps(fn)
+    def wrapper(*args, **kwargs):
+        try:
+            import torch  # pylint: disable=W0611
+        except ImportError:
+            raise unittest.SkipTest("onnxruntime not installed")
```
jcwchen, line 61:
```suggestion
            raise unittest.SkipTest("torch not installed")
```

onnx/test/runtime_test.py
```diff
@@ -0,0 +1,1493 @@
...
+    @skip_if_no_torch
+    def test_col2im_2x3_pads(self):
+        self.common_test_col2im(
+            28, (6, 4), (2, 3), pads=[1, 1, 1, 1], strides=[1, 1], dilations=[1, 1]
+        )
+
+    def test_col2im_2d(self):
+
+        data, image_shape, kernel_shape, dilations, pads, stride = (
+            np.array(
```
jcwchen, line 1292: nit: perhaps use:
```
data = np.zeros([6, 28])
data[0][0] = 1.
```
or wrap these long lines somehow.

jcwchen (commented):
Thank you @xadupre for the huge effort! Some preliminary thoughts (since this PR is still WIP, please ignore my point if you are already aware of them):
1. ~Add onnxruntime and torch into https://github.com/onnx/onnx/blob/main/requirements-dev.txt and test those related tests in at least one CI.~
2. Use new helper functions instead of old variables from mapping.py
3. Have a document under docs/ to explain what these files are for and how to use them.

All above points can be done by another PR if they make sense.

gramalingam:
> 1. Add onnxruntime and torch into https://github.com/onnx/onnx/blob/main/requirements-dev.txt and test those related tests in at least one CI.

I don't think that's a good idea. We don't want to create cyclic dependencies. It's preferable if onnx doesn't depend on those repos. I had a separate question elsewhere in this PR why we have test-cases that invoke onnxruntime. This PR should have nothing to do with onnxruntime, IMHO.

onnx/test/runtime_test.py
xadupre: done

onnx/test/runtime_test.py
xadupre: done

xadupre:
> Thank you @xadupre for the huge effort! Some preliminary thoughts (since this PR is still WIP, please ignore my point if you are already aware of them):
> 
> 1. Add onnxruntime and torch into https://github.com/onnx/onnx/blob/main/requirements-dev.txt and test those related tests in at least one CI.
> 2. Use new helper functions instead of old variables from mapping.py
> 3. Have a document under docs/ to explain what these files are for and how to use them.
> 
> All above points can be done by another PR if they make sense.

1. I would prefer to avoid adding more dependencies. I left the unit test using onnxruntime or torch to debug future implementation. It does not reduce the code coverage if thet are removed. 

2. I'll do it. I started this PR before this change was merged.

3. The documentation is written in class ProtoRun with a couple of examples. It can be extended.

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+        self.name = name
+
+    def __repr__(self) -> str:
+        "usual"
+        return f"{self.__class__.__name__}({self.name!r})"
+
+
+def _build_schemas() -> Dict[str, type]:
+    res: Dict[str, type] = {}
+    for schema in get_all_schemas_with_history():
```
gramalingam, line 46: Seems to ignore schema.domain? Should either use (domain, opname) or filter out only ops belonging to standard domain. Not sure which is intended.

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+    def __init__(self, g: GraphProto) -> None:
+        self.g = g
+
+
+class OpRun(ABC):
+    """
+    Ancestor to all operators in this subfolder.
+    The runtime for every node can checked into
+    `ONNX unit tests
+    <https://github.com/onnx/onnx/tree/master/onnx/backend/test/case/node>`_.
```
gramalingam, line 107: Not sure I understand what the above comment means.

gramalingam: I see some duplication of code across the above folder and the op implementations in this PR, which I guess is related. But not sure what the plan/suggestion above is.

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+                            # default value is undefined, it depends on the inputs
+                            value = None  # type: ignore
+                        else:
+                            value = self._extract_attribute_value(v.default_value, v)
+                        setattr(self, k, value)
+                        added_attributes.append(k)
+        self.attributes_names_ = set(added_attributes)
+
+    @staticmethod
+    def local_inputs(graph: GraphProto) -> List[str]:
```
gramalingam, line 225: Minor nit: may be `implicit_inputs` would be a better name? I find `local` confusing since these are *not* local (but belong to outer scope).

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+        for init in graph.initializer:
+            known.add(init.name)
+        for sparse_init in graph.sparse_initializer:
+            known.add(sparse_init.name)  # type: ignore
+        for inp in graph.input:
+            known.add(inp.name)
+        for node in graph.node:
+            for o in node.output:
+                known.add(o)
+            for i in node.input:
```
gramalingam, line 244: Currently, this does not handle nested graph-attributes, which could also contain references to outer-outer-scope.

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+        values = []
+        for name in names:
+            value = getattr(self, name)
+            if isinstance(value, RefAttrName):
+                if overridden_attributes is None:
+                    raise AttributeError(
+                        f"Attribute {name!r} of operator {type(self)} is linked but "
+                        f"'overridden_attributes' has no value for it."
+                    )
+                values.append(overridden_attributes[name])
```
gramalingam, line 275: I think `name` should be replaced by `value.name` here, if my understanding is correct.

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+            if att in overridden_attributes or att in kwargs:
+                continue
+            if not hasattr(self, att):
+                raise NameError(
+                    f"Attribute {att!r} is missing in operator {self.__class__.__name__!r}."
+                )
+            kwargs[att] = getattr(self, att)
+        try:
+            if len(overridden_attributes) > 0:
+                res = self._run(*args, **overridden_attributes, **kwargs)
```
gramalingam, line 364: Bit unclear what happens if a name appears in both overridden_attributes and kwargs. May be easier to understand it if it is made explicit in line 355-356 by choosing the correct one there.

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+                f"Method '_run' of class {self.__class__.__name__!r} does not return a tuple but {type(res)}."
+            )
+        if len(res) == 0:
+            raise ValueError(
+                f"Method '_run' of class {self.__class__.__name__!r} does not return any result."
+            )
+        if any(map(lambda t: isinstance(t, tuple), res)):
+            dtypes = [type(t) for t in res]
+            raise TypeError(
+                f"Method '_run' of class {self.__class__.__name__!r} must not return a tuple: {dtypes}."
```
gramalingam, line 385: May be this should read `must not return nested-tuple`? Otherwise, looks odd in comparison to line 376.

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+        n_inputs: Optional[int] = None,
+        n_outputs: Optional[int] = None,
+        **kwargs: Any,
+    ) -> NodeProto:  # type: ignore
+        """
+        Creates an ONNX node for this class based on the given information.
+
+        :param n_inputs: number of inputs (default is defined by the operator schema)
+        :param n_outputs: number of outputs (default is defined by the operator schema)
+        :param verbose: verbosity
```
gramalingam, line 401: Seems to be missing from parameter list above. Is it supposed to be part of kwargs?

gramalingam: I guess a copy-paste error?

onnx/funconnx/proto_run.py
```diff
@@ -0,0 +1,368 @@
...
+    Method :meth:`eval <onnx.ProtoRun.op_run.eval>` creates an onnx node
+    returned by method :meth:`make_node <onnx.funconnx.op_run.make_node>`.
+
+    .. exec_code::
+
+        import numpy as np
+        from onnx.funconnx.aionnx._op_list import Celu
+
+        onnx_node = Celu.make_node(alpha=0.5)
+        print(onnx_node)
```
gramalingam, line 122: nit: The above few lines (113-122) seem irrelevant here and can be omitted. In particular, it is an implementation detail of `eval`, and I think we can change `eval` to work without constructing a node, so not worth describing those details here.

onnx/funconnx/proto_run.py
```diff
@@ -0,0 +1,368 @@
...
+                self.onnx_graph_.sparse_initializer  # type: ignore
+            )
+            self.nodes_ = self.onnx_graph_.node
+        else:
+            self.input_names_ = list(proto.input)
+            self.output_names_ = list(proto.output)
+            self.inits_ = []
+            self.nodes_ = proto.node
+        if "" not in self.opsets:
+            self.opsets[""] = onnx_opset_version()
```
gramalingam, line 179: For a reference-implementation, it is better not to add "relaxations" such as the above ... it is an error if opset_imports does not include all used opsets.

gramalingam: Specifically, it is important to fail with an error in any situation that the standard/spec says is an error.

gramalingam: Suggest omitting lines 178-179

onnx/funconnx/proto_run.py
```diff
@@ -0,0 +1,368 @@
...
+
+        onnx_node = Celu.make_node(alpha=0.5)
+        print(onnx_node)
+    """
+
+    def __init__(  # type: ignore
+        self,
+        proto: Any,
+        opsets: Union[None, Dict[str, int]] = None,
+        functions=None,
```
gramalingam, line 129: What's the type supposed to be for functions? I am a bit unclear about the call to this in line 185.

onnx/funconnx/proto_run.py
```diff
@@ -0,0 +1,368 @@
...
+            self.input_names_ = list(proto.input)
+            self.output_names_ = list(proto.output)
+            self.inits_ = []
+            self.nodes_ = proto.node
+        if "" not in self.opsets:
+            self.opsets[""] = onnx_opset_version()
+        if functions is not None:
+            for f in functions:
+                if isinstance(f, FunctionProto):
+                    existing_functions = list(self.functions_.values())
```
gramalingam, line 183: Don't understand this line. Do the `values` also encode the domain/function-name already? Unclear to me why passing the value, without the name is sufficient for executing f.

gramalingam: Ok, never mind. I guess the `proto_` field includes this info.

onnx/funconnx/experimental/_op_list.py
```diff
@@ -0,0 +1,111 @@
...
+
+import textwrap
+from typing import Any, Union
+
+from ..op_run import OpFunction
+from ._op_run_experimental import OpRunExperimental
+from .op_im2col import Im2Col
+
+
+def _split_class_namme(name):  # type: ignore
```
gramalingam, line 12: nit: "namme" to "name"

onnx/funconnx/aionnx/_op.py
```diff
@@ -0,0 +1,198 @@
...
+        Calls method ``_run``, catches exceptions,
+        displays a longer error message.
+        Supports only binary operators.
+        """
+        self._log("-- begin %s.run(2 inputs)", self.__class__.__name__)
+        if x is None or y is None:
+            raise RuntimeError(
+                f"x and y have different dtype: {type(x)} != {type(y)} ({type(self)})"
+            )
+        if x.dtype != y.dtype:
```
gramalingam, line 85: Do all op implementations do this kind of type-checking? Or, is it based on case-by-case basis?

onnx/funconnx/aionnx/op_argmax.py
```diff
@@ -0,0 +1,56 @@
...
+    result = np.argmax(data, axis=axis)
+    result = data.shape[axis] - result - 1
+    if keepdims:
+        result = np.expand_dims(result, axis)
+    return result.astype(np.int64)
+
+
+class _ArgMax(OpRunArg):
+    def _run(self, data, axis=None, keepdims=None):  # type: ignore
+        axis = axis or self.axis  # type: ignore
```
gramalingam, line 28: Trying to understand this: is the `or self.axis` part necessary? At least, the calls to _run I saw in some part of the code I saw (in op_run.py or proto_run.py) already passed in self.axis.

onnx/funconnx/op_run.py
xadupre: This no collision of operator name accross domains but I should fix it.

onnx/funconnx/op_run.py
xadupre: I updated the comment. I forgot to change it.

onnx/funconnx/op_run.py
xadupre: renamed

onnx/funconnx/op_run.py
xadupre: Yes. I changed the error message.

onnx/funconnx/op_run.py
xadupre: The operator name and domain is already defined by class cls (the runtime). It can only execute a specific node. So it is not needed.

onnx/funconnx/proto_run.py
xadupre: moved into the docstring of make_node.

onnx/funconnx/proto_run.py
xadupre: added one annotation

onnx/funconnx/proto_run.py
xadupre: yes

onnx/funconnx/op_run.py
```diff
@@ -0,0 +1,508 @@
...
+        :param kwargs: optional inputs
+        :return: outputs
+        """
+        raise NotImplementedError(
+            f"Method '_run' must be overwritten for operator {self.__class__.__name__!r}."
+        )
+
+    def run(self, *args, linked_attributes=None, **kwargs):  # type: ignore
+        """
+        Calls method ``_run``, catches exceptions,
```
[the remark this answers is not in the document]
xadupre: I clarified. kwargs only contain argument `context`, a dictionary with the local result when running a node with a subgraph.

onnx/funconnx/op_run.py
xadupre: I clarified.

onnx/funconnx/aionnx/_op.py
xadupre: case by case, I assume the model was verified with onnx.checker.

onnx/funconnx/aionnx/op_argmax.py
xadupre: Remove, it is not necessary. This code was moved in OpRun.

onnx/funconnx/op_run.py
gramalingam: I don't understand. I meant that the description about the parameter `verbose` seems like it is here by mistake?

onnx/funconnx/experimental/_op_list.py
xadupre: fixed

onnx/funconnx/aionnx/op_loop.py
```diff
@@ -0,0 +1,87 @@
...
+
+        k_carried_away = [[] for i in range(self.K)]  # type: ignore
+        it = 0
+        while cond and it < M:
+            self._log("  -- loop> {%r}", context)
+            if len(body.input_names) > 0 and body.input_names[0] is not None:
+                inputs[body.input_names[0]] = np.array(it, dtype=M.dtype)  # type: ignore
+            if len(body.input_names) > 1 and body.input_names[1] is not None:
+                inputs[body.input_names[1]] = cond
+            outputs = self._run_body(inputs)  # type: ignore
```
gramalingam, line 60: Where is `_run_body` defined? I couldn't find it in my search in my visual-code. Thanks!

xadupre: In OpRun._load_attributes (https://github.com/onnx/onnx/pull/4483/files#diff-ea8c1377a6b06240f3181211d2d4bfecdf64493cd6cf602f1245016d7a9496c3R215). It is done automatically for every graph attribute.

onnx/funconnx/aionnx/op_loop.py
```diff
@@ -0,0 +1,87 @@
...
+            outputs = []
+            for i in body.input_names[2:]:
+                outputs.append(inputs[i])
+        else:
+            outputs = outputs[1 : 1 + self.N]
+        outputs.extend(k_carried_away)
+        while len(outputs) < len(self.onnx_node.output):
+            outputs.append(np.empty(shape=tuple()))
+        res = tuple(outputs)
+        if any(r is None for r in res):
```
gramalingam, line 85: Wonder if this should be removed now? How are optional values (like optional tensors) represented in this implementation? ONNX Loops were extended to support optional-types also at some point.

onnx/funconnx/aionnx/op_loop.py
```diff
@@ -0,0 +1,87 @@
...
+            it += 1
+            self._log("  -- loop<")
+
+        if it == 0:
+            outputs = []
+            for i in body.input_names[2:]:
+                outputs.append(inputs[i])
+        else:
+            outputs = outputs[1 : 1 + self.N]
+        outputs.extend(k_carried_away)
```
gramalingam, line 81: I guess these are scan outputs? But they should be tensors. It looks like the values are sequences in this implementation? If so, they need to be concatenated together into a tensor. Not sure if I misunderstood.

gramalingam:
It would be a good idea to update the AddNewOp.md document, especially [Step 2.2](https://github.com/onnx/onnx/blob/main/docs/AddNewOp.md#step-2-submit-pr-) so that future new ops update the op implementation here.

gramalingam:
I think that adding a `readme.md` file in the root directory `funconnx` would be helpful. In particular, to start with, we should add a disclaimer in the readme file that this is a new addition to the existing ONNX spec, and that there is potential for mismatch between the official spec and the implementation here, and that in the case of such a mismatch, the official spec overrides this implementation.

I think this is a practical approach to adding such a large PR, since I don't think we can get a reasonable review of all op implementations in a short-time, so we can continue reviewing and correcting the op implementations gradually.

onnx/helper.py
```diff
@@ -332,6 +333,20 @@ def float32_to_bfloat16(fval: float, truncate: bool = False) -> int:
...
+        return np.float32(np.nan)
+
+    expo = ival >> 7
+    prec = ival - (expo << 7)
+    sign = expo & 256
+    powe = expo & 255
+    fval = float(prec * 2.0 ** (-7) + 1) * 2.0 ** (powe - 127)
+    if sign:
+        fval = -fval
+    return np.float32(fval)
```
jcwchen, line 347: https://github.com/onnx/onnx/blob/8669fad0247799f4d8683550eec749974b4f5338/onnx/numpy_helper.py#L16
Are they different?

onnx/funconnx/aionnx/_op.py
```diff
@@ -0,0 +1,169 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Dict
+
+import numpy as np  # type: ignore
```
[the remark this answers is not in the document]
xadupre: done

onnx/funconnx/aionnx/op_loop.py
xadupre: I removed this test. It is never triggered.

onnx/funconnx/aionnx/op_loop.py
xadupre: You are right. There is no backend test covering this case.

onnx/funconnx/aionnx/op_stft.py
```diff
@@ -45,6 +39,7 @@ def _stft(x, fft_length, hop_length, n_frames, window, onesided=False):  # type:
         new_dim = sliced_x.shape[-2:-1]
         missing = (window_size - new_dim[0],)
         new_shape = sliced_x.shape[:-2] + missing + sliced_x.shape[-1:]
+        print(sliced_x.shape, begin, end, new_shape)
```
justinchuby, line 42: 👀 a print

gramalingam:
> Should the `Runtime` type itself be called `ReferenceRuntime` to be even more explicit?

Sounds good to me.

onnx/reference/ops/aionnx_preview_training/_op_run_training.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # pylint: disable=R0913,W0221
 
-from ..op_run import OpRun
+from ...op_run import OpRun
```
justinchuby, line 4: Possible to use absolute module paths?

onnx/funconnx/aionnx/op_stft.py
xadupre: thanks

onnx/reference/ops/aionnx_preview_training/_op_run_training.py
justinchuby: [PEP8](https://peps.python.org/pep-0008/#imports) recommends: 

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):

```
import mypkg.sibling
from mypkg import sibling
from mypkg.sibling import example
```

imo it improves readability (tells reader where the module comes from without having to go back to the dir structure) and makes the code easier to maintain when we move files around. Things start to get complicated when we have lines like `...foo.bar` or `....foo` and we may quickly lose count of the dots.

xadupre: I did not know. I usually use relative imports because when I work on a package and the package is also installed in site-package, absolute imports are ambiguous. I was searching for speeds but did not find anything about importing time. I assumed that relative import are faster since there is only one possible location instead of having mulitple (sys.path). But maybe I'm wrong. onnx.reference.ops._op_list imports more than 100 files, maybe that's worth measuring before changing everything.

justinchuby: Without having done the measurement myself, imported modules are cached so I think any speed difference would only matter the first time (and the difference should be small anyways)?

For site-package ambiguity, does developing in venv help?

xadupre: No worries, it is just a question of habits. I've been using relative imports since I discovered them. onnx is using absolute import in all other places, so I should use absolute too.

xadupre: done

onnx/reference/ops/op_conv.py
```diff
@@ -1,11 +1,178 @@
...
+            else:
+                res[t_res + t_kernel] = data[tuple(t_img)]
+    return res
+
+
+def im2col(
+    img: np.ndarray,
+    kernel_shape: Tuple[int, ...],
+    dilations: List[int],
+    pads: List[int],
```
justinchuby, line 81: Use Sequence when possible to allow more valid inputs?

onnx/helper.py
xadupre: I revmoved the one I added and modified the existing one by adding a default value to dims.

onnx/test/reference_evaluator_backend_test.py
```diff
@@ -583,11 +583,15 @@ def common_test_onnx_test_run(
             if verbose > 7:
                 print("  ", e, type(e))
             missed.append((te, e))
+            with open(f"missed_{te.name}.onnx", "wb") as f:
```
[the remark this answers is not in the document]
xadupre: done

onnx/reference/ops/op_conv.py
xadupre: done

onnx/numpy_helper.py
```diff
@@ -14,10 +15,14 @@ def combine_pairs_to_complex(fa: Sequence[int]) -> List[complex]:
...
 
 def bfloat16_to_float32(
-    data: np.ndarray, dims: Union[int, Sequence[int]]
+    data: np.ndarray, dims: Optional[Union[int, Sequence[int]]] = None
 ) -> np.ndarray:
     """Converts ndarray of bf16 (as uint32) to f32 (as uint32)."""
     shift = lambda x: x << 16  # noqa: E731
+    if dims is None:
+        if len(data.shape) == 0:
+            return shift(np.array([data]).astype(np.int32)).view(np.float32)[0]
```
gramalingam, line 24: This is a bit confusing. Is `data` not an `ndarray` in this case (as the type annotation above indicates)? Is it meant to be a scalar? (I notice that the scalar-transformation function has disappeared from the helper.py.) Clarify in the type-annotation or comments as appropriate.

xadupre: I removed the version I added in helper.py to handle scalar. It seemed confusing to have two versions. I extended this one. I fixed the annotation.

onnx/reference/README.md
```diff
@@ -0,0 +1,38 @@
+# ReferenceEvaluator
+
+This is a first attempt to provide an implementation for all operators
+defined by onnx. This is a pure python implementation.
+Mismatches may remain between the official specifications and the implementation here.
```
gramalingam, line 5: nit: "specifications" => "specification"

xadupre: done

jcwchen:
It seems that some tests in reference_evaluator_backend_test.py and reference_evaluator_test.py failed (see [CI](https://github.com/onnx/onnx/actions/runs/3444981516/jobs/5748174031)). One failed Windows CI uses x86 machine and another CI uses old NumPy version 1.16.6. Since other CIs are fine, is it because NumPy behaves slightly differently in Windows x86 machine and old NumPy version 1.16.6?

xadupre:
> It seems that some tests in reference_evaluator_backend_test.py and reference_evaluator_test.py failed (see [CI](https://github.com/onnx/onnx/actions/runs/3439012930/jobs/5735799774)). One failed Windows CI uses x86 machine and another CI uses old NumPy version 1.16.6. Since other CIs are fine, is it because NumPy behaves slightly differently in Windows x86 machine and old NumPy version 1.16.6?

I can either skip the test if numpy is too old and add in the documentation that the reference implementation only supports numpy >= 1.21.5 or fix the test one by one for numpy == 1.16.6. What do you prefer?

jcwchen:
Although ideally ONNX should always support the minimum required NumPy version (for now it's 1.16.6), I believe sometimes NumPy behaves quite differently in different environments (hardware, version). It might take a lot of time to fix NumPy behavior gap one by one and it's pretty close to the code freeze date (11/18). Thus, skipping them for now is fine to me.

BTW, besides old NumPy version issue (1.16.6), NumPy package for Windows x86 seems to behave differently as well so the test failed in release _win CI for x86 machines.

jcwchen:
Just speak aloud: to solve this kind of failure, is it possible that we can have loose tolerance when comparing output result from onnx reference and expected?

xadupre:
See #4657, I disabled only three tests related to cast into/from bfloat16.

belm0:
I tried using this, and get an error about ai.onnx.ml having no known functions.  I tried compiling onnx with ONNX_ML=1, but it was the same.  What am I missing?

```
>>> import onnx
>>> model = onnx.load_model('foo.onnx')
>>> from onnx.reference import ReferenceEvaluator
>>> sess = ReferenceEvaluator(model)
  File "<stdin>", line 1, in <module>
  File "/.../site-packages/onnx/reference/reference_evaluator.py", line 227, in __init__
    self._init()
  File "/.../site-packages/onnx/reference/reference_evaluator.py", line 312, in _init
    cl = self._load_impl(node)
  File "/.../site-packages/onnx/reference/reference_evaluator.py", line 379, in _load_impl
    raise NotImplementedError(
NotImplementedError: Node type 'Scaler' from domain 'ai.onnx.ml' is unknown, known functions: [].
```

gramalingam:
@xadupre

xadupre:
Is it possible to share the file foo.onnx you used?

belm0:
> Is it possible to share the file foo.onnx you used?

[foo.onnx.zip](https://github.com/onnx/onnx/files/11083787/foo.onnx.zip)

I didn't expect that the model file was relevant, since the runtime is reporting `known functions: []`.

xadupre:
I just tried with the main branch on github and it works. Which version of onnx are you using? The implementation for ml operators were introduced after the last published release on pypi.

PR #4487, opened 2022-09-05
Fix default value of RoiAlign's attribute coordinate_transformation_mode.

i4oolish:
The default semenatics of RoiAlign's attrtibute coordinate_transformation_mode in opset16 is not same with opset10.

For opset10, other framework like pytorch will manual insert an operation to process half_pixel, so it treat coordinate_transformation_mode  as output_half_pixel, while in opset16, the default value for RoiAlign's  attrtibute coordinate_transformation_mode in Operators.md is half_pixel. And at the same time, onnxruntime treat this attribute as no half_pixel in its cpu providers, below is the link.

https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/cpu/object_detection/roialign.h

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -22339,7 +22339,7 @@ Other versions of this operator: <a href="Changelog.md#RoiAlign-10">10</a>
 #### Attributes
 
 <dl>
-<dt><tt>coordinate_transformation_mode</tt> : string (default is half_pixel)</dt>
+<dt><tt>coordinate_transformation_mode</tt> : string (default is output_half_pixel)</dt>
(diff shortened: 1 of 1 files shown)


gramalingam:
Hi @i4oolish I have the same question as @daquexian. It is unclear what the issue is. Is there a bug in the pytorch converter? Or, a bug in the onnxruntime kernel? Please note that the version-converter will do the right thing for this change in spec (see [this](https://github.com/onnx/onnx/blob/f2f140805c109490f4a5d09381ab131178e54023/onnx/version_converter/convert.h#L506))

justinchuby:
Closed as the default was by design

Issue #4387, opened 2022-07-27
[Tracking] Verify Python 3.11 support 


justinchuby:
It is going to be shipped in two months.

PR #4490, opened 2022-09-06
Add Python 3.11 support

jcwchen:
### Description
Add Python 3.11 support in the future. For now, verify Python 3.11 support first.

### Motivation and Context
Closes https://github.com/onnx/onnx/issues/4387.


yan12125:
Building on 3.11 fails with: (from https://github.com/onnx/onnx/actions/runs/3002313403/jobs/4819386269)
```
/Users/runner/work/onnx/onnx/third_party/pybind11/include/pybind11/detail/type_caster_base.h:482:26: error: member access into incomplete type 'PyFrameObject' (aka '_frame')
            frame = frame->f_back;
                         ^
```
Looks like the issue is in third_party/pybind11, and it is fixed with pybind11 2.10.0: https://github.com/pybind/pybind11/commit/42a8e3125348d4e706e46eb410befc1422d42b3f.

yan12125:
Thanks for the update! Builds and tests now pass on Mac. The only remaining issue on Mac is missing 3.11 support for onnxruntime (https://github.com/microsoft/onnxruntime/issues/13482)

(From https://github.com/onnx/onnx/actions/runs/3415387401/jobs/5684459897)
> Run python -m pip install -q onnxruntime
> ERROR: Could not find a version that satisfies the requirement onnxruntime (from versions: none)
> ERROR: No matching distribution found for onnxruntime

---

On the other hand, builds on Linux fails as the pip path is incorrect

(From https://github.com/onnx/onnx/actions/runs/3415387398/jobs/5684459657)
> /opt/python//bin/pip install --no-cache-dir -q --upgrade pip
> .github/workflows/manylinux/entrypoint.sh: line 26: /opt/python//bin/pip: No such file or directory

The cause seems a mismatch between the Python version ([3.11-dev](https://github.com/onnx/onnx/pull/4490/files#diff-ecc29c1ce677ebf633712a033d0e1989082afd3a054d304aec3fbe3f6252cd38R19)) and the shell variable `python_map`, which uses [3.11](https://github.com/onnx/onnx/pull/4490/files#diff-31293c508f924625ccd034c86cad91bffd8457c6faad46b165c46475ae95c2daR18).

jcwchen:
Hi @yan12125, thank you for the pointers! I think now all Python 3.11 related CIs have passed. That is to say, at least now the latest main branch of ONNX should work with Python 3.11.

cjvolzka:
Two questions:

1. I see Python 3.7 references removed in the PR. Is support for it being dropped?
2. What onnx is this targeted for? I'm guessing it's too late for 1.13 (let me know if I'm wrong) so is this targeting 1.14?

jcwchen:
> I see Python 3.7 references removed in the PR. Is support for it being dropped?

Python 3.7's EOL is (27 Jun 2023) so ONNX won't deprecate 3.7 in the near future (at least upcoming 1.13 release will still support it).

> What onnx is this targeted for? I'm guessing it's too late for 1.13 (let me know if I'm wrong) so is this targeting 1.14?

This PR is only for testing Python 3.11 with the latest main branch and I am not planning to merge it soon (so it's a draft). The bottleneck here is the current used NumPy version 1.21.5 in ONNX does not work well with Python 3.11.... A newer NumPy version 1.23.2 works well with Python 3.11, but it does not work well with Python 3.7. Definitely this PR won't be included in the upcoming 1.13 release, but we will need to make the decision that whether upgrading NumPy version to support Python 3.11 wheel in the next 1.14 release.

daquexian:
> The bottleneck here is the current used NumPy version 1.21.5 in ONNX does not work well with Python 3.11.... A newer NumPy version 1.23.2 works well with Python 3.11, but it does not work well with Python 3.7.

FYI, it is possible to depend on different NumPy version on Python3.7 and 3.11: In requirements.txt, we can write

```
numpy==1.23.2; python_version >= "3.11"
numpy==1.21.5; python_version < "3.11"
```

which is called "requirement specifiers"
https://pip.pypa.io/en/stable/reference/requirements-file-format/ 
https://pip.pypa.io/en/stable/reference/requirement-specifiers/#requirement-specifiers

jcwchen:
Thank you @daquexian for letting me know. Previously I was worried that using inconsistent dependency might be a potential issue, but in fact ONNX itself does not have strong NumPy version requirement. In addition, we do see there are quite a few issues related to Python 3.11 support. Adding Python 3.11 support early can prevent a lot of similar issues coming up in the future.

After discussion with the ONNX infra team, we think we can add Python 3.11 support in the upcoming ONNX 1.13 release. 

cc @cjvolzka @yan12125 for awareness.

xadupre (approved):

gramalingam (approved):

PR #4495, opened 2022-09-08
Reduces the number of ignored mypy errors

xadupre:
### Description

Reduces the number of mypy errors. It works with the latest version of mypy. It adds the flag --ignore-missing-package to avoid adding `type: ignore` every time `import numpy` is used. The PR removes unnecessary `# type: ignore` every time an external package is imported.

### Motivation and Context

tools/style.sh generates many errors not related to the changes. This PR make it easier to read what is related to the user changes.


jcwchen (commented):
I believe in the current codebase a lot of code cannot pass mypy 0.971 check. Currently two CIs are still running two different versions of mypy, which is quite confusing:
1. mypy 0.782 in Linux-CI
2. mypy 0.971 in Python Lint CI

It seems there is no way to specify mypy version from the latest action-mypy: https://github.com/tsuyoshicho/action-mypy/issues/48. We probably can use an older version of tsuyoshicho/action-mypy, [v2.0.1](https://github.com/tsuyoshicho/action-mypy/releases/tag/v2.0.1), which is using 0.790. If there are still a lot of existing errors with 0.790, I would suggest either we keep one version of mypy or upgrade mypy to the latest one (0.971). Upgrade mypy to the latest one will need a lot of fixes and perhaps the best we can do for now is only keeping version 0.782. @@justinchuby What's your thoughts on this? Thanks!

onnx/__init__.py
```diff
@@ -77,7 +77,7 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
                     "Please use save_as_external_data to save tensors separately from the model file."
                 ) from e
             raise
-        return result
+        return result  # type: ignore
```
justinchuby, line 80: We should avoid ignores without error nums (too board)

onnx/backend/test/runner/__init__.py
```diff
@@ -165,7 +165,7 @@ def test_suite(self) -> unittest.TestSuite:
             unittest.TextTestRunner().run(BackendTest(backend).test_suite)
         """
         suite = unittest.TestSuite()
-        for case in sorted(self.test_cases.values()):
+        for case in sorted(self.test_cases.values()):  # type: ignore
```
jcwchen, line 168: May I understand which kind of error does it ignore?

onnx/backend/test/runner/__init__.py
```diff
@@ -178,7 +178,7 @@ def tests(self) -> Type[unittest.TestCase]:
             onnx_backend_tests = BackendTest(backend).tests
         """
         tests = self._get_test_case("OnnxBackendTest")
-        for items_map in sorted(self._filtered_test_items.values()):
+        for items_map in sorted(self._filtered_test_items.values()):  # type: ignore
```
jcwchen, line 181: May I understand which kind of error does it ignore?

onnx/test/shape_inference_test.py
```diff
@@ -100,7 +100,7 @@ def _assert_inferred(
         inferred_model = self._inferred(graph, **kwargs)
         inferred_vis = list(inferred_model.graph.value_info)
         vis = list(sorted(vis, key=lambda x: x.name))
-        inferred_vis = list(sorted(inferred_vis, key=lambda x: x.name))
+        inferred_vis = list(sorted(inferred_vis, key=lambda x: x.name))  # type: ignore
```
jcwchen, line 103: May I understand which kind of error does it ignore?

onnx/test/symbolic_shape_test.py
```diff
@@ -60,9 +60,9 @@ def _get_shape_from_name(
         for v in inputs + outputs + valueinfos:
             if v.name == name:
                 if v.type.HasField("tensor_type"):
-                    return v.type.tensor_type.shape
-                elif v.type.HasField("sparse_tensor_type"):
-                    return v.type.sparse_tensor_type.shape
+                    return v.type.tensor_type.shape  # type: ignore
+                if v.type.HasField("sparse_tensor_type"):
+                    return v.type.sparse_tensor_type.shape  # type: ignore
```
jcwchen, line 65: May I understand which kind of error does it ignore?

jcwchen (commented):
The part for adding `ignore-missing-imports` makes sense to me. Thank you for updating so many places!

IMO we should be careful about adding more ignore conditions because it's hard to resolve every of them in the future when feasible. I saw a few places were added `# type: ignore` so I would like to understand more details and whether we can resolve them properly.

onnx/__init__.py
xadupre: mypy returns `Returning Any from function declared to return "bytes"` in this case. I assume SerializeToString() is annotated as Any when this function returns bytes. mypy detects the inconsistency. I don't think we can easliy remove it. I don't know if onnx is the source of the wrong annotation or if it is protobuf.

workflow_scripts/test_model_zoo.py
```diff
@@ -119,7 +119,7 @@ def main():
         print(
             f"In all {len(model_list)} models, {len(failed_models)} models failed, {len(skip_models)} models were skipped"
         )
-        for model, error in failed_messages:
+        for model, error in failed_messages:  # type: ignore
```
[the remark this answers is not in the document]
xadupre: Because it was not detected before, I assume a couple of annotations are inconsistent. It is a unit test so I did not look into it.

onnx/test/test_external_data.py
```diff
@@ -7,7 +7,7 @@
 import uuid
 from typing import Any, List, Tuple
 
-import numpy as np  # type: ignore
+import numpy as np
 import pytest  # type: ignore
```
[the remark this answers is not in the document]
xadupre: thanks, I missed it.

onnx/__init__.py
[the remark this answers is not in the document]
xadupre: mypy returns `Returning Any from function declared to return "bytes"` in this case. So function returns a less generic type than the underlying function. I assume SerializeToString() is annotated as `Any` and this function returns `bytes`. mypy detects the inconsistency. I don't think we can easliy remove it. I don't know if onnx is the source of the wrong annotation or if it is protobuf.

onnx/backend/test/runner/__init__.py
xadupre: I assume TestCase does not really support sort. I'll check.

onnx/test/symbolic_shape_test.py
xadupre: I don't think mypy is seeing `v.type.sparse_tensor_type.shape` as a `TensorShapeProto`. I did not check why since it is coming from protobuf. I suggest fixing it in another PR. Right now, it explicitely ignores what was implicitely ignored.

tools/style.sh
```diff
@@ -25,7 +25,7 @@ black . --color --diff --check
 echo -e "::endgroup::"
 
 echo -e "\n::group:: ===> check mypy"
-mypy . --no-site-packages
+mypy . --no-site-packages --ignore-missing-imports
```
justinchuby, line 28: I don’t think we needed this change because it should read the mypy config

onnx/test/shape_inference_test.py
justinchuby: I would still suggest adding the error code in square brackets to narrow the ignore for clarity.

tools/style.sh
xadupre: I was not sure. I removed this flag since I also added it in setup.cfg.

workflow_scripts/test_model_zoo.py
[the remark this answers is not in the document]
jcwchen:
```suggestion
        for model_name, error in failed_messages:
```
This should fix it. Thanks for catching.

onnx/backend/test/runner/__init__.py
xadupre: Changed.

workflow_scripts/test_model_zoo.py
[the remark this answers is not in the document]
jcwchen:
```suggestion
        for model_name, error in failed_messages:
            print(f"{model_name} failed because: {error}")
```
This hasn't been resolved yet?

requirements-dev.txt
```diff
@@ -10,9 +10,9 @@ pyzmq
 setuptools
 twine
 # Dependencies for linting. Versions match those in setup.py.
-flake8==5.0.2
+flake8>=5.0.2
 clang-format==13.0.0
-mypy==0.782
+mypy>=0.782
```
jcwchen, line 15: Shall we bump the used mypy version in [setup.py](https://github.com/onnx/onnx/blob/5fbb7b507e8c908d210179181c9f22ffc01e80fc/setup.py#L333) since this PR enable mypy check with mypy 0.971?

And I think this PR can close this issue: https://github.com/onnx/onnx/issues/3964.

xadupre: done

workflow_scripts/test_model_zoo.py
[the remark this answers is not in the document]
xadupre: done

jcwchen (approved):
LGTM. Thank you for enabling the latest mypy!

gramalingam (approved):

gramalingam:
Thanks for fixing these!

Issue #4343, opened 2022-07-04
Support for `BitwiseXor`, `BitwiseOr`, `Bitcast`, and `BitwiseAnd` ops


josephrocca:
# New Operator

### Describe the operator
The ` BitwiseXor`, `BitwiseOr`, `Bitcast`, and `BitwiseAnd` ops:

 1. https://www.tensorflow.org/api_docs/python/tf/raw_ops/Bitcast
 2. https://www.tensorflow.org/api_docs/python/tf/raw_ops/BitwiseAnd
 3. https://www.tensorflow.org/api_docs/python/tf/raw_ops/BitwiseOr
 4. https://www.tensorflow.org/api_docs/python/tf/raw_ops/BitwiseXor

### Can this operator be constructed using existing onnx operators?
[Apparently it is not possible](https://github.com/onnx/tensorflow-onnx/issues/1985#issuecomment-1173456022):

> But after I took a quick check, there is no easy way to do it yet.

### Is this operator used by any model currently? Which one?
Yes, the Flax BART model used by DALL-E Mini, for example, requires this op when trying to convert to ONNX format (via TF Saved Model format).

### Are you willing to contribute it? (Y/N)
Yes, but I would probably require guidance since I haven't done anything like this before.

### Notes
Please see this issue for context: https://github.com/onnx/tensorflow-onnx/issues/1985

postrational:
Do we know other models which require these operations?

AlexandreEichenberger:
In PyTorch: https://pytorch.org/docs/stable/torch.html

```
bitwise_not | Computes the bitwise NOT of the given input tensor.
-- | --
bitwise_and | Computes the bitwise AND of input and other.
bitwise_or | Computes the bitwise OR of input and other.
bitwise_xor | Computes the bitwise XOR of input and other.
bitwise_left_shift | Computes the left arithmetic shift of input by other bits.
bitwise_right_shift | Computes the right arithmetic shift of input by other bits.
[bitwise_not](https://pytorch.org/docs/stable/generated/torch.bitwise_not.html#torch.bitwise_not)
Computes the bitwise NOT of the given input tensor.
[bitwise_and](https://pytorch.org/docs/stable/generated/torch.bitwise_and.html#torch.bitwise_and)
Computes the bitwise AND of input and other.
[bitwise_or](https://pytorch.org/docs/stable/generated/torch.bitwise_or.html#torch.bitwise_or)
Computes the bitwise OR of input and other.
[...]

PR #4496, opened 2022-09-08
BitwiseAnd, BitwiseOr, BitwiseXor support

philass:
### Description

This PR adds BitwiseAnd, BitwiseOr, and BitwiseXor

### Motivation and Context

There is no way to sanely do bitwise operations with the current ONNX opset. Relates to https://github.com/onnx/onnx/issues/4343

### Additional comments

Each Bitwise op has 4 tests.

Also I was in two minds if I should include `BitwiseNot` in this PR, or in another PR. If you guys feel it should be included in this one please let me know. The PR for `BitwiseNot` is here https://github.com/onnx/onnx/pull/4497


onnx/defs/logical/defs.cc
```diff
@@ -219,4 +219,99 @@ ONNX_OPERATOR_SET_SCHEMA(
         }
         )ONNX"));
 
+std::function<void(OpSchema&)> BinaryBitwiseDocGenerator(const char* name) {
+  return [=](OpSchema& schema) {
+    std::string doc;
+    POPULATE_OP_DOC_STR(doc = R"DOC(
+Returns the tensor resulted from performing the `{name}` bitwise operation
```
gramalingam, line 226: Nit:
(i) change `resulted` to `resulting`. 
(ii) exchange the order of `{name}` and `bitwise`.

gramalingam:
LGTM, thanks!

onnx/defs/logical/defs.cc
philass: Good catch! That is definitely better wording

I've made the changes to the PR

p-wysocki (approved):
Would testing 0d tensor inputs (scalars, essentially) bring any benefits? Overall LGTM, that is very clean code.

philass:
@p-wysocki thanks for the review!

I checked other unary and binary ops. And it seems that most times 0d tensors aren't checked.

If you feel it would add more coverage, then I am happy to add tests.

gramalingam (approved):

gramalingam:
I think some minor conflict-resolution is required since the other PR (NOT op) touches the same lines in the opset file.

philass:
> I think some minor conflict-resolution is required since the other PR (NOT op) touches the same lines in the opset file.

Thanks for merging in `BitwiseNot`. I have fixed the merge conflicts. And these ops should be ready to merge.

PR #4497, opened 2022-09-08
BitwiseNot implementation

philass:
Signed-off-by: Philip Lassen <philiplassen+git@gmail.com>

### Description

This PR implements the BitwiseNot operator.  I've implemented the Binary bitwise operators in https://github.com/onnx/onnx/pull/4496. 

### Motivation and Context

With this PR and https://github.com/onnx/onnx/pull/4496 we have implementations for the same set of bitwise operators numpy has support for. 
- `np.bitwise_and`
- `np.bitwise_or`
- `np.bitwise_xor`
- `np.bitwise_not`

The are also the same set of Ops which already have a ONNX logical equivalent. `And`, `Or`, `Xor`, and `Not`. In ONNX these take booleans and return booleans. This also touches on @AlexandreEichenberger's comments in https://github.com/onnx/onnx/issues/4343


gramalingam:
LGTM, thanks! The only question is whether some of the ops should be defined as functions in terms of others.

philass:
Thanks for the review!

> LGTM, thanks! The only question is whether some of the ops should be defined as functions in terms of others.

I think bitwise ops are sufficiently primitive where it may be overkill to make some of these functions. From my understanding most chips and hardware accelerators have first class support for these bitwise operations.

However if you feel that this is best done with functions, I can try and figure out how to do that.

gramalingam:
> Thanks for the review!
> 
> > LGTM, thanks! The only question is whether some of the ops should be defined as functions in terms of others.
> 
> I think bitwise ops are sufficiently primitive where it may be overkill to make some of these functions. From my understanding most chips and hardware accelerators have first class support for these bitwise operations.
> 
> However if you feel that this is best done with functions, I can try and figure out how to do that.

Not necessary. I think this is fine. I agree that in practice, it won't make a big difference.

p-wysocki (approved):
Would testing 0d tensor inputs (scalars, essentially) bring any benefits? LGTM.

philass:
@p-wysocki thanks for the review!

I checked other unary and binary ops. And it seems that most times 0d tensors aren't checked.

If you feel it would add more coverage, then I am happy to add tests.

gramalingam (approved):

PR #4498, opened 2022-09-08
Install specified mypy version first in Python Lint CI to make mypy version consistent

jcwchen:
### Description
Install specified mypy version first in Python Lint CI to make mypy version consistent

### Motivation and Context
Currently action-mypy is complaining many mypy errors with newer mypy version 0.971. See related discussion: https://github.com/onnx/onnx/pull/4495#issuecomment-1240905666.

cc @justinchuby Thanks for the solution!


.github/workflows/lint.yaml
```diff
@@ -67,6 +67,9 @@ jobs:
           level: warning
           filter_mode: diff_context
           glob_pattern: "**/*.py"
+      - name: Install specified mypy version
+        run: |
+          python -m pip install -q -r requirements-dev.txt
```
jcwchen, line 72: I chose requirements-dev.txt instead of `pip install mypy==0.782` so we don't have to update a lot of hardcoded versions when upgrading in the future.

liqunfu (approved):

PR #4505, opened 2022-09-10
To make QuantizeLinear and DquantizeLinear function ops

liqunfu:
### Description
To make QuantizeLinear and DquantizeLinear function ops

### Motivation and Context
As part of effort to reduce primary operator count.

PS: was planning to present a general way for QDQ implementation. After offline discussion, I feel it is better to:
- for integer quantize data types, to extend QuantizeLinear and DquantizeLinear to support it.
- for float point (fp8, fp16), Cast op can be used to support it.


xadupre (approved):

onnx/defs/quantization/defs.cc
```diff
@@ -67,6 +130,45 @@ for per-tensor / per layer quantization, or a 1-D tensor for per-axis quantizati
...
+    const FunctionBodyBuildContext& ctx,
+    const OpSchema& schema,
+    FunctionProto& functionProto) {
+  FunctionBuilder builder(functionProto);
+
+  auto* input_type = ctx.getInputType(0);
+  if (!input_type->has_tensor_type()) {
+    fail_schema("Invalid x input type. Expected a tensor type");
+  }
+  auto output_elt_type = TensorProto::FLOAT;
```
justinchuby, line 143: elt: is it and abbreviation?

readability: Avoid abbreviations - they add cognitive load to readers.

> Do not use abbreviations that are ambiguous or unfamiliar to readers outside the project, and do not abbreviate by deleting letters within a word.

https://google.github.io/styleguide/pyguide.html#316-naming
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

onnx/defs/quantization/defs.cc
```diff
@@ -15,6 +15,68 @@ For saturation, it saturates to [0, 255] if it's uint8, or [-128, 127] if it's i
...
+    } else if (zero_point_type->tensor_type().elem_type() == TensorProto::INT8) {
+      quantized_min = -128;
+      quantized_max = 127;
+      output_elt_type = TensorProto::INT8;
+    } else {
+      fail_schema("Invalid zero_point_type. Expected TensorProto::UINT8 or TensorProto::INT8");
+    }
+  }
+
+  bool is_per_axis = ctx.getInputType(1)->tensor_type().shape().dim().size() == 1 &&
```
gramalingam, line 51: I don't think we can rely on the static-type to figure out if this is per-axis or not: in general, the shape-information may not be available statically. Same problem for Dequantize function too.

onnx/defs/quantization/defs.cc
```diff
@@ -15,6 +15,68 @@ For saturation, it saturates to [0, 255] if it's uint8, or [-128, 127] if it's i
...
+      ctx.getInputType(1)->tensor_type().shape().dim(0).dim_value() > 1;
+
+  builder.Const("quantized_min", ToTensor(quantized_min));
+  builder.Const("quantized_max", ToTensor(quantized_max));
+  if (is_per_axis) {
+    auto* axis_attribute = ctx.getAttribute("axis");
+    int axis = axis_attribute ? ctx.getAttribute("axis")->i() : 1;
+    int x_rank = ctx.getInputType(0)->tensor_type().shape().dim().size();
```
gramalingam, line 59: As above. All computation that is dependent on shape/rank needs to be done using ONNX ops, as it is available only at run-time.

onnx/defs/quantization/defs.cc
```diff
@@ -15,6 +15,68 @@ For saturation, it saturates to [0, 255] if it's uint8, or [-128, 127] if it's i
...
+    std::vector<int64_t> y_scale_shape(x_rank - axis, 1);
+    y_scale_shape[0] = ctx.getInputType(1)->tensor_type().shape().dim(0).dim_value();
+    builder.Const("y_scale_shape", y_scale_shape);
+    builder.Add("y_scale_reshape = Reshape (y_scale, y_scale_shape)");
+    builder.Add("x_s = Div (x, y_scale_reshape)");
+    builder.Add("y_zero_point_reshape = Reshape (y_zero_point, y_scale_shape)");
+    builder.Add("y_zero_point_cast = Cast(y_zero_point_reshape)", "to", (int64_t)(input_elt_type));
+  } else {
+    builder.Add("x_s = Div (x, y_scale)");
+    builder.Add("y_zero_point_cast = Cast(y_zero_point)", "to", (int64_t)(input_elt_type));
```
gramalingam, line 69: One trouble with Quantization is that it mentions the use of saturated arithmetic. Will cast do the right thing?

liqunfu: cast from quantized (narrow) to dequantized (wide) is ok. To cast from dequantized to quantized, we need to clip first.

onnx/defs/quantization/defs.cc
liqunfu: I changed to elem to match the function name. hope this works.

onnx/defs/quantization/defs.cc
liqunfu: Here with c++ if it will generate a function graph without If node. The function graph will be composed with a block of either if or else branch. The function graph is smaller than that if we use If op.
In addition, as we discussed offline, ONNX If op with shape dependent condition is having issues in shape inference.

onnx/defs/quantization/defs.cc
liqunfu: Both quantized element type and y_scale shape depends on the context. We could manage not to use y_scale shape (for example, to force broadcasting regardless, which will be inefficient in cases where broadcast is not needed). We can only get quantized element type from the context. Thus, the op function has to be context dependent.

yufenglee:
What's the benefit to do this? Q/DQ is very general functionality. Supposedly, each framework that depends on ONNX has an implementation of them.

onnx/defs/quantization/defs.cc
```diff
@@ -15,6 +15,67 @@ For saturation, it saturates to [0, 255] if it's uint8, or [-128, 127] if it's i
...
+
+  auto output_elem_type = TensorProto::UINT8;
+  float quantized_min = 0, quantized_max = 255;
+  if (ctx.hasInput(2)) {
+    auto* zero_point_type = ctx.getInputType(2);
+    if (!zero_point_type->has_tensor_type()) {
+      fail_schema("Invalid y_zero_point input type. Expected a tensor type");
+    }
+    if (zero_point_type->tensor_type().elem_type() == TensorProto::UINT8) {
+      quantized_min = 0;
```
yufenglee, line 38: redundant code

onnx/defs/quantization/defs.cc
yufenglee: Agree. we can not fully rely on static shape inference

onnx/defs/quantization/defs.cc
```diff
@@ -15,6 +15,67 @@ For saturation, it saturates to [0, 255] if it's uint8, or [-128, 127] if it's i
...
+    builder.Add("y_zero_point_reshape = Reshape (y_zero_point, y_scale_shape)");
+    builder.Add("y_zero_point_cast = Cast(y_zero_point_reshape)", "to", (int64_t)(input_elem_type));
+  } else {
+    builder.Add("x_s = Div (x, y_scale)");
+    builder.Add("y_zero_point_cast = Cast(y_zero_point)", "to", (int64_t)(input_elem_type));
+  }
+
+  builder.Add("x_s_z = Add (x_s, y_zero_point_cast)");
+  builder.Add("x_s_z_clipped = Clip (x_s_z, quantized_min, quantized_max)");
+  builder.Add("y = Cast (x_s_z_clipped)", "to", (int64_t)(output_elem_type));
```
yufenglee, line 73: Cast is not right. It should be a rounding to closest and tie to even.

Issue #3420, opened 2021-04-13
[Operator] Why Cumsum axis is not an attribute?


masahi:
Hi, I'm wondering why [Cumsum op](https://github.com/onnx/onnx/blob/master/docs/Operators.md#CumSum), introduced in https://github.com/onnx/onnx/pull/2030, requires `axis` to be passed as input, rather than being an attribute? Other ops that take an axis, such as Reduce*, Scatter, Gather, Split, Concat etc all seem to take axis as an attribute. Also in practice, I cannot imagine a use case where we want to have a variable axis for cumsum. 

Unless there is a good reason to have a variable axis for Cumsum, I believe this is a flaw in the spec that needs to be corrected. 

cc @jcwchen @postrational @wschin

fdwr:
Older operators tend to take such parameters as constant attributes, but newer operators have been moving toward dynamic parameters, such as `TopK` moving `k` from constant to dynamic. I don't have the background for why cumulative summation's `axis` was set to a dynamic parameter, but I am curious (and I also wonder if this general trend of moving constant attributes to dynamic inputs will eventually make the distinction between "attributes" vs "inputs" rather obsolete, such that we *just* have inputs, the only distinction being whether they came from a `Constant` or from some other graph edge).

masahi:
`k` parameter of `TopK` being dynamic does make sense, although it is constant in majority of cases I believe, it is not unreasonable to want `k` param being a computed expression. 

But I'd say `axis`  parameter is very different, this is not something you pick arbitrary value for at runtime. Given that other ops that needs `axis` param take it as an attribute, I don't see why `Cumsum` should be any different.

masahi:
I see arguments vs attributes distinction doesn't really matter for typical DL frameworks or ONNXRuntime where everything is interpreted at runtime. But for DL compilers like [TVM](https://github.com/apache/tvm), whether some input is static vs dynamic has a huge difference.

fdwr:
>whether some input is static vs dynamic has a huge difference

Indeed, the same concern applies within ONNX Runtime to the DirectML execution provider (which has a constant `DML_OPERATOR_CUMULATIVE_SUMMATION::Axis` field), but thanks to constant folding and the ability to query whether a tensor input is constant or not, the distinction between parameter types from a constant "attribute" vs a const graph initializer vs a `Const` operator all becomes moot, and we can create the GPU shader ahead of time either way. The real deciding factor is *constness*, not *attributeness*.

[...]

PR #4512, opened 2022-09-15
make primary ops function step 2

liqunfu:
### Description
update reduce ops so that axes is the second optional input
change reduce ops to function op

### Motivation and Context
to reduce amount of primary ops


xadupre:
This PR will have a significant impact on converting librairies.

onnx/test/shape_inference_test.py
```diff
@@ -2311,19 +2311,21 @@ def test_gemm_no_bias(self) -> None:
 
     def test_reduce_op_shape_2_axis(self) -> None:
         graph = self._make_graph(
-            [("x", TensorProto.FLOAT, (24, 4, 11))],
-            [make_node("ReduceL1", "x", "y", axes=(1, 2), keepdims=0)],
+            [("x", TensorProto.FLOAT, (24, 4, 11)), ("axes", TensorProto.INT64, (2,))],
```
gramalingam, line 2314: Should we be retaining the old test-cases for the old version, and add copies of the test-cases for the new version?

onnx/shape_inference/implementation.cc
```diff
@@ -388,6 +393,8 @@ class ShapeInferenceImplBase {
           updateType(n.output(i), ctx.getOutputType(i));
       }
 
+      preprocess(n);
```
gramalingam, line 396: IIUC, `preprocess` is being called twice now, and we should drop one of them.

I suggest dropping the loop in line 474-476 and in 540-542, since you have it above as part of `process(node)`.

But I would move the above line to the beginning of this function, before type-and-shape-inference is called.

gramalingam: Or, drop the above line, and retain the loops in 474 and 540.

liqunfu:
> This PR will have a significant impact on converting librairies.

Yes it does. For updated ops, I have added appropriate adapters for model converter. Anything else I need to do?

gramalingam:
@BowenBao : for your awareness: this PR proposes to promote the axes attribute to input for various Reduce* ops. (This was already done for ReduceSum previously, but not for other ops.) I hope the impact on the converter won't be much? It would be good to know, thanks.

onnx/backend/test/case/node/reduce_log_sum.py
```diff
@@ -11,54 +11,67 @@
...
 
     @staticmethod
     def export_keepdims() -> None:
         node = onnx.helper.make_node(
-            "ReduceLogSum", inputs=["data"], outputs=["reduced"]
+            "ReduceLogSum", inputs=["data", "axes"], outputs=["reduced"]
         )
         data = np.random.ranf([3, 4, 5]).astype(np.float32)
         reduced = np.log(np.sum(data, keepdims=True))
+        axes = np.array([], dtype=np.int64)
```
xadupre, line 55: All Reduce operator now declare axes as optional. I don't think there is any test checking the one input is supported. Should we add one?

liqunfu: I am planning to add this type of tests in a later PR. The reason is that ORT has a bug not allowing reduce ops with only one input. I tried with reducesum with ORT and it failed. Because this PR is already too big, I try not to grow it further and to introduce dependencies on ORT.

onnx/defs/reduction/defs.cc
```diff
@@ -4,160 +4,86 @@
...
+  }
+  )ONNX";
+ONNX_OPERATOR_SET_SCHEMA(
+    ReduceLogSum,
+    18,
+    OpSchema().FillUsing(ReduceDocGeneratorWithFunctionBody("log sum", reduce_log_sum_func_body)));
+
+const char* reduce_log_sum_exp_func_body = R"ONNX(
+  {
+    data_exp = Exp (data)
```
xadupre, line 53: This implementation may cause some overflow. Should we take the maximum first?

BowenBao:
> @BowenBao : for your awareness: this PR proposes to promote the axes attribute to input for various Reduce* ops. (This was already done for ReduceSum previously, but not for other ops.) I hope the impact on the converter won't be much? It would be good to know, thanks.

Thanks for heads up. Wwe will need to update related code when supporting opset 18. I'm a little concerned over the impact on shape inference, more details in inline comments. 

BTW, what's the process of deciding which attribute to promote to input? Just wondering as it seems ops like `softmax` still has its `axes`  as attribute.

liqunfu:
> 
Shape inference caused by axes becoming an input is handled by taking Constant's value_ints as tensor during shapeinferencing. 
RE the process to decide which op qualified for such change, I think it is dependent on the op. In the case of this PR, we noticed that reducesum is the only one has axes as input. This inconsistency prevents us from making other reduce ops a function op. It is also very confusing that reduce ops have different signatures. ArgMax and ArgMin would be next target for changing but it is not urgent. The main goal is to make ops function op. So if softmax category ops can be changed to function with axes being input, we likely will do it.

onnx/defs/reduction/utils.cc
```diff
@@ -0,0 +1,148 @@
...
+      std::vector<int64_t> axes;
+      size_t num_inputs = ctx.getNumInputs();
+      if ((num_inputs == 2) && ctx.getInputType(1)) { // axes is input
+        if (ctx.getAttribute("axes")) {
+          fail_shape_inference("axes as an input and attribute cannot be specified at the same time.");
+        }
+
+        const TensorProto* axesInitializer = ctx.getInputData(1);
+        if (axesInitializer == nullptr) {
+          // skip if axes is not an initializer
```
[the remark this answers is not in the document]
liqunfu: it makes sense to check the rank of the axes input which shall be 1. I will update this.
Constants is converted to initializer by the runtime. Here I should use axes_constant as the name of the variable. I will update this PR for this.

ONNX shape inference for existing cases (where axes is just constant numbers) is handled in this PR. In future cases where constant folding can help, I believe ONNX shape inference shall be enhanced to do that.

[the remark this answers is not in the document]
jcwchen: > Dynamic axes input potentially introduces dynamic output rank for reduce ops. However it is likely the axes values in most "new" cases can be either constant folded computed or symbolically inferred. @jcwchen any thoughts?

Good catch. Yes, if the "axes" becomes input, it's possible that in the future such a shape information needs to be dynamic computed. Currently reduce-family operators does not support symbolic shape inference yet. I think for now it should be fine because this PR promotes axes from attribute to input for generic reason, instead of having a real-world model requires the axes to be dynamic computed. After promoting axes to input, if the shape information comes from either a Constant op or initializer, I think current shape inference can still handle that.

Still, I agreed that supporting dynamic input for reduce-family operator is definitely good-to-have. A question for @liqunfu: in the future will reduction operators involve dynamic computing for axes when using it in a function op?

gramalingam:
For reference: here is a previous discussion on this topic: https://github.com/onnx/onnx/issues/3420

onnx/defs/reduction/utils.cc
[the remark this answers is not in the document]
liqunfu: It is hard to say for sure. It can only be done if the input come from a constant op. In other cases, axes cannot be calculated.  However in such cases the existing reduce ops cannot work either.

onnx/defs/schema.h
```diff
@@ -837,12 +848,16 @@ class OpSchema final {
   const FunctionProto* GetFunction() const;
 
   bool HasContextDependentFunction() const {
-    return functionBuilder_ != nullptr;
+    return opset_version_to_function_builder_.find(since_version_) != opset_version_to_function_builder_.end();
+  }
+
+  bool HasContextDependentFunctionWithOpsetVersion(int opset_version) const {
```
gramalingam, line 854: I think we can use the same name `HasContextDependentFunction` as an overload (or use  single function with a default value for `opset_version` just like in the functions below.)

onnx/defs/schema.cc
```diff
@@ -665,16 +674,23 @@ void OpSchema::ParseAndSetTypes(
   }
 }
 
-OpSchema& OpSchema::SetContextDependentFunctionBodyBuilder(ContextDependentFunctionBodyBuilder functionBuilder) {
-  functionBuilder_ = std::move(functionBuilder);
+OpSchema& OpSchema::SetContextDependentFunctionBodyBuilder(ContextDependentFunctionBodyBuilder functionBuilder, int opset_version) {
+  opset_version_to_function_builder_[opset_version] = std::move(functionBuilder);
```
gramalingam, line 678: What happens if opset_version is -1 ?

onnx/defs/math/defs.cc
```diff
@@ -1070,7 +1008,21 @@ ONNX_OPERATOR_SET_SCHEMA(
 
               schema.BuildFunction(functionProto);
               return true;
-            }));
+            })
+        .SetContextDependentFunctionBodyBuilder(
+            [](const FunctionBodyBuildContext& ctx, const OpSchema& schema, FunctionProto& functionProto) -> bool {
```
gramalingam, line 1013: Nit: May be adding a comment "Function for opset 18" would be helpful. I realize the number "18" is down below, but it took me some time to realize it.

onnx/defs/math/defs.cc
```diff
@@ -1070,7 +1008,21 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+              int64_t axis = ctx.getAttribute("axis") != nullptr ? ctx.getAttribute("axis")->i() : -1;
+              FunctionBuilder builder(functionProto);
+              builder.Const1D("axes", axis).Add("X_ReduceMax = ReduceMax <keepdims = 1> (input, axes)").Add(R"(
```
gramalingam, line 1016: Nit: Is this the auto-formatter doing this? It's easier to read if we have one (ONNX op) per line.

onnx/defs/schema.h
xadupre: Do you assume a function will only use the opset the operator is defined into? I'm think8ng about ai.onnx.ml. A couple of operators can be defined as function of existing operators in ai.onnx.

gramalingam: > Do you assume a function will only use the opset the operator is defined into? I'm thinking about ai.onnx.ml. A couple of operators can be defined as function of existing operators in ai.onnx.

Good point. I think eventually we should support an interface where `opset_version` parameter is a map from domains to versions (not necessarily in this PR). Xavier's example can be handled right now by treating opset_version as the version of ai.onnx (not the domain the function belongs to). But there are examples in onnxruntime where we would want to mix a custom domain and ai.onnx

onnx/test/shape_inference_test.py
liqunfu: brought back the tests for opset 13. added tests for opset 18

onnx/defs/reduction/defs.cc
liqunfu: used case to double to solve type restriction of Exp op. It also avoids possible numerical issues

onnx/defs/schema.cc
liqunfu: It will still be inserted. Eventually the functionBuilder for opset_version of -1 will be moved to since_version_ when since_version_ is updated in SinceVersion()
I am using -1 (static constexpr int kUninitializedSinceVersion = -1;) for uninitialized since_version_ and default opset_version value.

onnx/defs/math/defs.cc
liqunfu: I agree. Formatting rules enforced is less then great. Many times I find it making code ugly. It refuses my manual correction.

onnx/defs/schema.h
liqunfu: Currently it assumes ops used to define a function op are from the same opset version of the same opset. With this change, it allows to specific the opset version. Using ops from different opsets is allowed at the ONNX IR level. However I do not see the related support with OpSchema. I agree with Rama that we need a separate PR for this.

gramalingam: Do we need `WithOpsetVersion` in the function name? Why not drop it? Overloading will resolve it, right?

onnx/defs/schema.h
```diff
@@ -822,27 +823,58 @@ class OpSchema final {
...
+      const std::vector<NodeProto>& func_nodes,
+      const std::vector<OperatorSetIdProto>& opsets,
+      int opset_version = kUninitializedSinceVersion);
 
-  OpSchema& FunctionAddOpset(const char* domain, int version);
+  OpSchema& FunctionBody(const char* func_body, int opset_version = kUninitializedSinceVersion);
 
   const FunctionProto* GetFunction() const;
 
+  const FunctionProto* GetFunctionWithOpsetVersion(int opset_version) const;
```
gramalingam, line 850: Documentation would be useful for this function. In particular, I notice that it returns something based on lower_bound. I am not sure about that for 2 reasons. I think the right thing to do is to additionally check that the candidate (returned by lower_bound) has the same opset-version for every op called within the function-body as the requested opset-version. This requires constructing the function-body first. Maybe easier to discuss this in person, over a call.

onnx/defs/schema.cc
```diff
@@ -822,18 +829,67 @@ const FunctionProto* OpSchema::GetFunction() const {
...
+    bool fail_on_invalid_op) {
+  if (requested_opset_version == function_since_version) {
+    // we validate that OpSchemas optained with requested_opset_version and function_since_version
+    // are the same for ops used in a function definition.
+    return;
+  }
+  std::stringstream err;
+  bool has_invalid_op = false;
+  for (int node_index = 0; node_index < function->node_size(); ++node_index) {
+    const NodeProto& node = function->node(node_index);
```
gramalingam, line 851: `for (auto& node : function->node())` is simpler.

onnx/defs/schema.cc
```diff
@@ -822,18 +829,67 @@ const FunctionProto* OpSchema::GetFunction() const {
...
-FunctionProto* OpSchema::GetFunctionWithOpsetInternal(int opset_version) {
-  std::map<int, std::shared_ptr<FunctionProto>>::iterator it =
-      opset_version_to_function_body_.lower_bound(opset_version);
-  if (it != opset_version_to_function_body_.cend())
-    return it->second.get();
+void OpSchema::ValidateReferencedOpsInFunciton(
+    const FunctionProto* function,
+    int requested_opset_version,
+    int function_since_version,
+    const std::string& domain,
```
gramalingam, line 841: I suggest dropping this `domain` parameter. (Or, the longer term solution is to replace the `int opset_version` parameter by a map from domain to version.)

onnx/defs/schema.cc
```diff
@@ -822,18 +829,67 @@ const FunctionProto* OpSchema::GetFunction() const {
...
+  if (requested_opset_version == function_since_version) {
+    // we validate that OpSchemas optained with requested_opset_version and function_since_version
+    // are the same for ops used in a function definition.
+    return;
+  }
+  std::stringstream err;
+  bool has_invalid_op = false;
+  for (int node_index = 0; node_index < function->node_size(); ++node_index) {
+    const NodeProto& node = function->node(node_index);
+    const OpSchema* op1 = OpSchemaRegistry::Instance()->GetSchema(node.op_type(), requested_opset_version, domain);
```
gramalingam, line 852: We must use `node.domain()` for `domain` here.

onnx/defs/schema.cc
```diff
@@ -822,18 +829,67 @@ const FunctionProto* OpSchema::GetFunction() const {
...
+    const OpSchema* op2 = OpSchemaRegistry::Instance()->GetSchema(node.op_type(), function_since_version, domain);
+    if (op1 != op2) {
+      if (!has_invalid_op) {
+        err << "A function op (" << function->name() << ") with opset version (" << requested_opset_version
+            << ") is requested.\n";
+        err << "A function definition of opset since_version (" << function_since_version << ") shall be used.\n";
+        err << "However, there is/are conflict(s) with operators used to define the function: \n";
+      }
+      err << "Operator (" << node.op_type() << ") of version " << requested_opset_version
```
gramalingam, line 861: I think we must leave it to the caller to report the correct error message. I think it may be better to return just a status result (success/failure, with a list of opnames that cause an issue (as a string or list).

onnx/defs/schema.cc
```diff
@@ -822,18 +829,67 @@ const FunctionProto* OpSchema::GetFunction() const {
...
+    }
+  }
 }
+
+FunctionProto*
+OpSchema::GetFunctionWithOpsetInternal(int opset_version, const std::string& domain, bool fail_on_invalid_op) {
+  std::map<int, std::shared_ptr<FunctionProto>>::iterator it =
+      opset_version_to_function_body_.upper_bound(opset_version);
+  if (opset_version_to_function_body_.empty() || it == opset_version_to_function_body_.begin()) {
+    ONNX_THROW_EX(std::out_of_range(
```
gramalingam, line 881: I don't think we need to use exceptions. Just return a nullptr. If desired, we should return a status/error message. I think better to handle this failure same as failure of validation down below. (And the caller decides what to do.)

onnx/defs/schema.cc
```diff
@@ -822,18 +829,67 @@ const FunctionProto* OpSchema::GetFunction() const {
...
+        std::string("Cannot find a function body that satisfies the requested opset version: op_type = ") +
+        this->name_ + ", opset_version = " + std::to_string(opset_version) + "."));
+  } else {
+    --it;
+    int function_since_version = it->first;
+    FunctionProto* function = it->second.get();
+    ValidateReferencedOpsInFunciton(function, opset_version, function_since_version, domain, fail_on_invalid_op);
```
gramalingam, line 888: If validation fails, we should not return function, we should return a nullptr (plus extra status/error message if desired).

onnx/defs/schema.cc
```diff
@@ -822,18 +829,67 @@ const FunctionProto* OpSchema::GetFunction() const {
...
+OpSchema::GetFunctionWithOpsetVersion(int opset_version, const std::string& domain, bool fail_on_invalid_op) const {
+  return const_cast<OpSchema*>(this)->GetFunctionWithOpsetInternal(opset_version, domain, fail_on_invalid_op);
 }
 
-FunctionProto* OpSchema::GetFunctionWithOpsetInternal(int opset_version) {
-  std::map<int, std::shared_ptr<FunctionProto>>::iterator it =
-      opset_version_to_function_body_.lower_bound(opset_version);
-  if (it != opset_version_to_function_body_.cend())
-    return it->second.get();
+void OpSchema::ValidateReferencedOpsInFunciton(
```
gramalingam, line 837: nit: `Funciton` to `Function`

onnx/defs/schema.cc
```diff
@@ -735,26 +737,31 @@ void OpSchema::UpdateFunctionProtoOpsetImportVersion(FunctionProto& function_pro
...
   if (opset_version == OpSchema::kUninitializedSinceVersion)
     opset_version = since_version_;
 
   std::map<int, ContextDependentFunctionBodyBuilder>::const_iterator it =
-      opset_version_to_function_builder_.find(opset_version);
-
-  if (it != opset_version_to_function_builder_.end()) {
+      opset_version_to_function_builder_.upper_bound(opset_version);
+  if (opset_version_to_function_builder_.empty() || it == opset_version_to_function_builder_.begin()) {
+    ONNX_THROW_EX(std::out_of_range(
```
gramalingam, line 749: Let's avoid the exception. Return false if unable to build a function. If we want to return extra information, we should return an extra error message.

onnx/defs/schema.cc
```diff
@@ -735,26 +737,31 @@ void OpSchema::UpdateFunctionProtoOpsetImportVersion(FunctionProto& function_pro
...
     if (!body_builder(ctx, *this, function_proto)) {
       return false;
     }
 
-    // default opset import may have been added to function_proto by OpSchema::BuildFunction
-    // we need to update its version with the specified opset_version
+    //// default opset import may have been added to function_proto by OpSchema::BuildFunction
+    //// we need to update its version with the specified opset_version
     UpdateFunctionProtoOpsetImportVersion(function_proto, opset_version);
+    ValidateReferencedOpsInFunciton(&function_proto, opset_version, it->first, domain, fail_on_invalid_op);
```
gramalingam, line 762: Return false if validation fails

onnx/test/cpp/function_context_test.cc
```diff
@@ -149,7 +149,265 @@ void RegisterCustomFunctionSchema() {
   (void)unused;
 }
 
-// Test for Context dependant function with type context
+void RegisterVersionedLogSoftmaxFunctionSchema(const std::string& op_type, bool with_missing_version) {
+  // SinceVersion of a function operator tells the last opset version where its semantic is defined.
+  // Function bodies and context dependent function builders also have versions.
+  // A function's version specifies its valid range since its since_version till the since_version
+  // of the next function of the same funtion op.
+  std::cout << "StepA\n";
```
gramalingam, line 157: Delete?

liqunfu: Sorry this is temporary for a mysterious CI failure.

onnx/test/cpp/function_context_test.cc
```diff
@@ -149,7 +149,265 @@ void RegisterCustomFunctionSchema() {
   (void)unused;
 }
 
-// Test for Context dependant function with type context
+void RegisterVersionedLogSoftmaxFunctionSchema(const std::string& op_type, bool with_missing_version) {
```
gramalingam, line 152: Is this just testing the ability to register two function builders? May be adding a comment to indicate what is being tested will be helpful.

gramalingam: Ok, never mind. I guess it is just a utility.

onnx/test/cpp/function_context_test.cc
```diff
@@ -149,7 +149,265 @@ void RegisterCustomFunctionSchema() {
...
+  NodeProto nodeProto;
+  nodeProto.set_op_type(op_type);
+  nodeProto.add_input("input");
+  nodeProto.add_output("output");
+
+  TypeProto floatTypeProto;
+  floatTypeProto.mutable_tensor_type()->set_elem_type(TensorProto_DataType::TensorProto_DataType_FLOAT);
+
+  FunctionBodyBuildContextImpl ctx(nodeProto, {floatTypeProto});
+  schema.BuildContextDependentFunctionWithOpsetVersion(ctx, fnProto, requested_opset_version, "", true);
```
gramalingam, line 221: As discussed elsewhere, better if the build function returns true/false/status to indicate success or failure.

onnx/test/cpp/function_context_test.cc
```diff
@@ -149,7 +149,265 @@ void RegisterCustomFunctionSchema() {
...
+  std::cout << "Step1\n"; 
+  // VersionedLogSoftMax has 1 function body for opset version 13.
+  // It is ok to request a function body for a model with opset import version 13.
+  FunctionProto function_proto13;
+  BuildLogSoftmaxFunction(*schema13, op_type, function_proto13, 13);
+  ASSERT_TRUE(function_proto13.name() == op_type);
+  std::cout << "Step2\n"; 
+
+  ONNX_TRY {
+    // It may not be ok to request a function body for a model with opset import version 14, 15, 16, and 17.
```
gramalingam, line 247: nit: suggest changing the wording to `A request for a function body .... 14, 15, 16, and 17 will fail`

onnx/test/cpp/function_context_test.cc
```diff
@@ -149,7 +149,265 @@ void RegisterCustomFunctionSchema() {
...
+    std::cout << err.what();
+    EXPECT_TRUE(
+        std::string(err.what())
+            .find("Operator (Sub) of version 17 is used but the op has been updated at least once since version 13.") !=
+        std::string::npos);
+  }
+
+  std::cout << "Step5\n"; 
+  ONNX_TRY {
+    // It may not be ok to request a function body for a model with opset import version 18.
```
gramalingam, line 265: As above. Avoid "may not", as it does not clearly indicate what will happen.

onnx/defs/schema.h
```diff
@@ -1007,18 +1036,19 @@ class OpSchemaRegistry final : public ISchemaRegistry {
    public:
     OpSchemaRegisterOnce(OpSchema& op_schema, int opset_version_to_load = 0) {
       ONNX_TRY {
+        std::cout << "StepF ";
```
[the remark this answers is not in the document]
liqunfu: for debugging purposes. removed

onnx/test/cpp/function_context_test.cc
liqunfu: avoided exception.

onnx/defs/schema.cc
liqunfu: updated the function to return boolean

onnx/defs/schema.cc
liqunfu: changed to return nullptr;

onnx/reference/ops/op_reduce_l1.py
```diff
@@ -3,11 +3,32 @@
 
 import numpy as np
 
+from onnx.defs import onnx_opset_version
+
 from ._op import OpRunReduceNumpy
 
 
-class ReduceL1(OpRunReduceNumpy):
+class ReduceL1_1_11_13(OpRunReduceNumpy):
```
xadupre, line 11: I should highlights the current convention, it shoud be <op name>_<op_version>. op_version is the opset it is defined. It is then valid from op_version to until there is a nore recent one. So in your case, it should be `ReduceL1_1`.

onnx/reference/ops/_op.py
```diff
@@ -167,3 +167,16 @@ def __init__(self, onnx_node: NodeProto, run_params: Dict[str, Any]):
                 self.axes = None
             elif isinstance(self.axes, list):
                 self.axes = tuple(self.axes)
+
+    def IsAxesEmpty(self, axes):
```
xadupre, line 171: Python usually uses `is_axes_empty`.

onnx/defs/schema.cc
```diff
@@ -829,63 +831,52 @@ const FunctionProto* OpSchema::GetFunction() const {
...
+    return has_no_invalid_op;
   }
-  std::stringstream err;
-  bool has_invalid_op = false;
-  for (int node_index = 0; node_index < function->node_size(); ++node_index) {
-    const NodeProto& node = function->node(node_index);
-    const OpSchema* op1 = OpSchemaRegistry::Instance()->GetSchema(node.op_type(), requested_opset_version, domain);
-    const OpSchema* op2 = OpSchemaRegistry::Instance()->GetSchema(node.op_type(), function_since_version, domain);
+  for (auto& node : function->node()) {
+    const OpSchema* op1 =
```
gramalingam, line 854: I think we should do the check below only if `node.domain()` is the standard domain? Or, should we assert that?

onnx/defs/schema.cc
```diff
@@ -829,63 +831,52 @@ const FunctionProto* OpSchema::GetFunction() const {
...
+// are not updated between function_since_version and requested_opset_version (include requested_opset_version).
+// this call only validate ops in the default domain.
+// TODO: validate ops in other domains.
+bool OpSchema::ValidateReferencedOpsInFuncton(
     const FunctionProto* function,
     int requested_opset_version,
     int function_since_version,
-    const std::string& domain,
-    bool fail_on_invalid_op) {
+    std::set<std::string>* updated_ops) const {
```
gramalingam, line 848: Should we use a reference to indicate that this input should be non-null?

onnx/reference/ops/op_reduce_l1.py
gramalingam: Agree about factorizing it so we can define `ReduceL1_1`, `ReduceL1_11` etc. using a couple of lines.

onnx/defs/schema.cc
liqunfu: However this validation works for ops in any domain, as long as the op is in the registry.

gramalingam: But the parameter requested_opset_version specified is only for the standard onnx domain. It should not be used with a different domain

onnx/defs/schema.cc
liqunfu: It does allow nullptr in which case the caller is not interested in the specific ops that are updated after since_version.

onnx/defs/schema.cc
liqunfu: True, let me add the condition.

onnx/backend/test/case/node/__init__.py
```diff
@@ -128,31 +128,45 @@ def rename_helper(internal_name: str) -> Any:
...
 
-    for attr in schema.attributes:
-        if attr in [a.name for a in node.attribute]:
-            continue
-        if schema.attributes[attr].default_value:
-            node.attribute.extend([schema.attributes[attr].default_value])
+        # function_proto.attributes
+        node_list = function_expand_helper(node, function_proto, op_prefix)
+        node_lists.append(node_list)
+        function_proto_opset_imports.append(function_proto.opset_import)
```
gramalingam, line 167: minor nit: it may be cleaner to do:
```py
   expanded_tests.append(node_list, function_proto.opset_import)
```
That is, construct and return `zip(node_lists, function_proto_opset_imports)` directly.

onnx/defs/reduction/utils.cc
```diff
@@ -0,0 +1,148 @@
...
+    schema.Output(0, "reduced", "Reduced output tensor.", "T", OpSchema::Single, true, 1, OpSchema::Differentiable);
+    schema.TypeConstraint(
+        "T",
+        GetSupportedDataTypesForReductionOps(supports_8bit_datatypes),
+        supports_8bit_datatypes ? "Constrain input and output types to high-precision and 8 bit numeric tensors."
+                                : "Constrain input and output types to high-precision numeric tensors.");
+    if (func_body) {
+      schema.FunctionBody(func_body);
+    } else if (function_builder) {
+      std::cout << "schema.SetContextDependentFunctionBodyBuilder(function_builder);" << std::endl;
```
gramalingam, line 81: Delete debugging message

onnx/defs/reduction/utils.cc
```diff
@@ -0,0 +1,148 @@
...
+      if (attr_proto) {
+        keep_dims = attr_proto->i();
+      }
+      auto noop_attr_proto = ctx.getAttribute("noop_with_empty_axes");
+      if (noop_attr_proto) {
+        noop_with_empty_axes = noop_attr_proto->i();
+      }
+      std::vector<int64_t> axes;
+      size_t num_inputs = ctx.getNumInputs();
+      if ((num_inputs == 2) && ctx.getInputType(1)) { // axes is input
```
gramalingam, line 101: nit: can replace above by `if (ctx.hasInput(1))`

onnx/reference/ops/_op.py
```diff
@@ -167,3 +167,18 @@ def __init__(self, onnx_node: NodeProto, run_params: Dict[str, Any]):
                 self.axes = None
             elif isinstance(self.axes, list):
                 self.axes = tuple(self.axes)
+
+    def is_axes_empty(self, axes):
+        return axes is None or len(axes.shape) == 0 or axes.shape[0] == 0
+
+    def HandleAxes(self, axes):
```
gramalingam, line 174: minor nit: I think the naming convention is `handle_axes` (same comment as Xavier made earlier for above method).

onnx/reference/ops/op_reduce_l1.py
```diff
@@ -3,11 +3,46 @@
...
+class ReduceL1_11(ReduceL1_1):
+    pass
+
+
+class ReduceL1_13(ReduceL1_1):
+    pass
+
+
+class ReduceL1_18(OpRunReduceNumpy):
+    def run(self, data, axes=None):  # type: ignore
```
gramalingam, line 27: I suggest adding `noop_with_empty_axes=False` as a parameter, and passing it to `_run`.

onnx/reference/ops/op_reduce_l1.py
```diff
@@ -3,11 +3,46 @@
...
+        return self._run(data, axes)
+
+    def _run(self, data, axes):  # type: ignore
+        if self.is_axes_empty(axes) and self.noop_with_empty_axes:  # type: ignore
```
gramalingam, line 31: I suggest replacing `self.noop_with_empty_axes` by `noop_with_empty_axes` (a parameter, added to `_run` as discussed above.)

onnx/cpp2py_export.cc
```diff
@@ -114,6 +114,22 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+          "all_function_opset_versions",
+          [](OpSchema* op) -> std::vector<int> {
+            std::vector<int> all_function_opset_versions = op->function_opset_versions();
+            std::vector<int> context_dependent_function_opset_versions =
+                op->context_dependent_function_opset_versions();
+            all_function_opset_versions.insert(
+                all_function_opset_versions.end(),
+                context_dependent_function_opset_versions.begin(),
+                context_dependent_function_opset_versions.end());
+            std::sort(all_function_opset_versions.begin(), all_function_opset_versions.end());
```
xadupre, line 130: Is there a possibility the result contains duplicates?

onnx/reference/ops/op_reduce_l1.py
```diff
@@ -3,11 +3,48 @@
...
+    pass
+
+
+class ReduceL1_18(OpRunReduceNumpy):
+    def run(self, data, axes=None, keepdims=None, noop_with_empty_axes=None):  # type: ignore
+        keepdims = keepdims or self.keepdims  # type: ignore
+        noop_with_empty_axes = noop_with_empty_axes or self.noop_with_empty_axes  # type: ignore
+        return self._run(data, axes, keepdims, noop_with_empty_axes)
+
+    def _run(self, data, axes, keepdims=1, noop_with_empty_axes=0):  # type: ignore
```
xadupre, line 32: You may remove `def run` and define the default values for `def _run` as None. If None, the main class automatically looks into the node definition to replace them or the schema if they are not specified.

onnx/shape_inference/implementation.cc
```diff
@@ -294,6 +295,10 @@ class ShapeInferenceImplBase {
           } else if (attr.type() == AttributeProto::SPARSE_TENSOR && attr.has_sparse_tensor()) {
             input_sparse_data_by_name[n.output(0)] = &attr.sparse_tensor();
           }
+        } else if (attr.name() == "value_ints" && attr.type() == AttributeProto::INTS) {
```
xadupre, line 298: I would put `attr.type() == AttributeProto::INTS` first. Both conditions should be equivalent but checking an int is faster than a string.

onnx/test/reference_evaluator_backend_test.py
```diff
@@ -60,6 +60,26 @@
     # not implemented
     "test__simple_gradient_of_add",  # gradient not implemented
     "test__simple_gradient_of_add_and_mul",  # gradient not implemented
+    "test_layer_normalization_2d_axis1_expanded",  # https://github.com/onnx/onnx/issues/4653
```
xadupre, line 63: See https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/split.py#L13, I followed the convention defined in this file. You can specify an opset with suffix `_opset<opset>`.

onnx/test/reference_evaluator_backend_test.py
```diff
@@ -461,7 +481,7 @@ class TestOnnxBackEndWithReferenceEvaluator(unittest.TestCase):
 
     @classmethod
     def add_test_methods(cls):
-        for folder in ["node", "pytorch-converted", "pytorch-operator", "simple"]:
+        for folder in ["node"]:
```
xadupre, line 484: Why this change?

onnx/test/reference_evaluator_backend_test.py
```diff
@@ -491,12 +511,12 @@ def _test_(
 
                 setattr(TestOnnxBackEndWithReferenceEvaluator, te.fname, _test_)
 
-    def test_onnx_backend_test_abs(self):
-        name = "test_abs"
-        code = []
-        for te in enumerate_onnx_tests("node", lambda folder: folder == name):
-            code.append(te)
-        self.assertEqual(len(code), 1)
+    # def test_onnx_backend_test_abs(self):
```
xadupre, line 514: Why disabling the test?

onnx/reference/ops/_op.py
```diff
@@ -167,3 +167,18 @@ def __init__(self, onnx_node: NodeProto, run_params: Dict[str, Any]):
...
+        if (
+            axes is not None and len(axes.shape) > 0 and axes.shape[0] > 0
```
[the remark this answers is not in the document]
gramalingam: After separate discussion, decided this is a separate issue and can be handled separately.

onnx/reference/ops/op_reduce_l1.py
gramalingam: But is this wrong? I find this current form useful since a reader can understand what happens (what is the default value) without having to look elsewhere.

onnx/defs/schema.cc
```diff
@@ -667,55 +700,186 @@ void OpSchema::ParseAndSetTypes(
...
   return *this;
 }
 
-bool OpSchema::BuildContextDependentFunction(const FunctionBodyBuildContext& ctx, FunctionProto& functionProto) const {
-  if (functionBuilder_)
-    return functionBuilder_(ctx, *this, functionProto);
-  else
-    return false;
+bool OpSchema::BuildContextDependentFunction(const FunctionBodyBuildContext& ctx, FunctionProto& function_proto) const {
+  return BuildContextDependentFunctionWithOpsetVersion(ctx, function_proto, since_version_);
```
gramalingam, line 715: Nit: Do we need these two different methods? Or, can we just use `BuildContextDependentFunction` with a default value for parameter requested_opset_version?

onnx/defs/schema.cc
```diff
@@ -667,55 +700,186 @@ void OpSchema::ParseAndSetTypes(
...
   return *this;
 }
 
 const FunctionProto* OpSchema::GetFunction() const {
-  return function_body_.node_size() > 0 ? &function_body_ : nullptr;
+  return GetFunctionWithOpsetVersion(since_version_);
+}
+
+const FunctionProto* OpSchema::GetFunctionWithOpsetVersion(int requested_opset_version) const {
+  return const_cast<OpSchema*>(this)->GetFunctionWithOpsetInternal(requested_opset_version);
```
gramalingam, line 835: Why do we need this `const_cast` ? May be outdated. I don't see any UpdateFunctionProtoOpsetImportVersion any more. May be simpler to inline the internal function here.

onnx/shape_inference/implementation.cc
```diff
@@ -529,6 +536,11 @@ class ShapeInferenceImplBase {
         attr_map[attr] = ctx.getAttribute(attr);
       }
     }
+
+    for (auto& n : func_proto.node()) {
```
gramalingam, line 540: Why do we need this?

onnx/shape_inference/implementation.cc
```diff
@@ -463,10 +470,10 @@ class ShapeInferenceImplBase {
       }
       processInitializer(tp.values().name(), tp, initializer_type, input_sparse_data_by_name);
     }
-    // Collect data from constant nodes and check if any experimental ops exist
-    for (auto& n : *graph.mutable_node()) {
-      preprocess(n); // process constant node
-    }
+    // // Collect data from constant nodes and check if any experimental ops exist
+    // for (const auto& n : graph.node()) {
```
gramalingam, line 474: Delete this?

onnx/reference/ops/op_reduce_l1.py
liqunfu: @xadupre , I agree with @gramalingam  here. We need at least on call (_run here) to follow precisely the op schema signature.

liqunfu: In addition, I think the suggested _run signature does not work with nest calls - when the op is part of other function ops. There for now I like to keep the current code.

onnx/cpp2py_export.cc
liqunfu: good point, added code to remove duplications. op authors shall not define both type of functions with the same version. all_function_opset_versions is used to check whether there is any function so we generate doc properly.

onnx/test/reference_evaluator_backend_test.py
liqunfu: you are right that every test ending with _ver18 shall pass. I enabled those tests.
regarding naming of expended tests, they are auto generated. I used _ver# for opset import for the function body.

onnx/test/reference_evaluator_backend_test.py
liqunfu: My mistake. I commented for debugging. Forgot to bring it back.

onnx/test/reference_evaluator_backend_test.py
liqunfu: forgot bring it back after debugging

jcwchen (commented):
Some nit comments. I saw some test_data_set_0 only has input_1.pb without input_0.pb. Is it expected? Since this PR really introduces a lot of test files (model.onnx, input/output pb files), it would be super great if we can verify them in ORT in advance.

docs/Operators-ml.md
```diff
@@ -9,8 +9,8 @@ For an operator input/output's differentiability, it can be differentiable,
             is not specified, that variable has undefined differentiability.
 
 ### ai.onnx.ml
-|**Operator**|**Since version**|
-|-|-|
+|**Operator**|**Since version**||
+|-|-|-|
```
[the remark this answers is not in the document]
liqunfu: This is for function ops which has its own version. We need to tell the opset import for the function body.

onnx/defs/schema.cc
```diff
@@ -667,55 +700,176 @@ void OpSchema::ParseAndSetTypes(
...
+    requested_opset_version = since_version_;
+  std::map<int, std::shared_ptr<FunctionProto>>::const_iterator it =
+      opset_version_to_function_body_.upper_bound(requested_opset_version);
+  if (opset_version_to_function_body_.empty() || it == opset_version_to_function_body_.begin()) {
+    return nullptr;
+  } else {
+    --it;
+    int function_since_version = it->first;
+    const FunctionProto* function = it->second.get();
+    ValidateReferencedOpsInFuncton(function, requested_opset_version, function_since_version);
```
gramalingam, line 837: Change to return nullptr if validation fails.

docs/Operators-ml.md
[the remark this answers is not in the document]
liqunfu: > test_reduce_op_shape_2_axis_opset_version_13

good point! I did some verification in ort. Some test requires changes in ORT, e.g. new kernels. For these tests, I will do soon after the PR is merged. This is due to bandwidth and not being able to access the branch from my repo by ORT.

liqunfu:
> Some nit comments. I saw some test_data_set_0 only has input_1.pb without input_0.pb. Is it expected? Since this PR really introduces a lot of test files (model.onnx, input/output pb files), it would be super great if we can verify them in ORT in advance.

That is because due to the test change, only input1 is changed. input0 stays unchanged so it is not in the PR.

gramalingam (approved):
LGTM. Only nit comment left is whether `HasContextDependentFunctionWithOpsetVersion` can be renamed `HasContextDependentFunction`.

PR #4514, opened 2022-09-15
Validate the generated number of input.pb and output.pb in CIs

jcwchen:
### Description
Validate the generated number of input.pb and output.pb in CIs

### Motivation and Context
Some validation will be missing after https://github.com/onnx/onnx/pull/4431. We should additionally handle the validation of generated number of input.pb and output.pb

### DIFF
--- a/.azure-pipelines/Linux-CI.yml
+++ b/.azure-pipelines/Linux-CI.yml
@@ -127,5 +127,10 @@ jobs:
         echo "git diff for test generation returned failures. Please check updated node test files"
         exit 1
       fi
+      git diff --exit-code --name-only -- . ':!onnx/onnx-data.proto' ':!onnx/onnx-data.proto3'
+      if [ $? -ne 0 ]; then
+        echo "Test generation returned failures. Please check the number of node test files (input_*.pb or output_*.pb)"
+        exit 1
+      fi
 
     displayName: Test backend test data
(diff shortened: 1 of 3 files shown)


p-wysocki (approved):
I think it's a great addition and this check can save a lot of development time, especially for new contributors. I had issues myself at https://github.com/onnx/onnx/pull/4481, which would be easily detected by this check.

jcwchen (commented):
.

jcwchen:
Not sure whether we can directly add --clean in the script: https://github.com/onnx/onnx/blob/de2459972d3f834dba24a586982d1ffce5532e0b/tools/update_doc.sh#L25
Or even set --clean=True by default.

jcwchen:
> Would it be worthwhile for the doc generator to track files it generated and finally issue a warning message if there were old files (not generated in this run)?

Makes sense to me. I have added such a check in cmd_tools and if there are additional files, it will throw a warning to let users understand. Please review it. Thanks!

liqunfu (approved):

jcwchen:
@gramalingam if you have time, could you please review this PR? (It needs an approval from SIG-operators-approvers as well). Thank you!

gramalingam:
So, all the deleted *.pb files were spurious (out-dated) files created by this issue? Good to catch them and remove them!

jcwchen:
> So, all the deleted *.pb files were spurious (out-dated) files created by this issue? Good to catch them and remove them!

Yes, they were caught by the enhanced CI check (with --clean) because they cannot be regenerated by any existing test script. Therefore I removed them.

gramalingam (approved):

PR #4518, opened 2022-09-16
[document] Add vcpkg instruction step

JackBoosY:
`onnx` is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `onnx` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `onnx`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/onnx/portfile.cmake). We try to keep the library maintained as close as possible to the original library. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,14 @@ pip install onnx
 
 [Weekly packages](https://test.pypi.org/project/onnx-weekly/) are published in test pypi to enable experimentation and early testing.
 
+## vcpkg packages
+onnx is in the maintenance list of [vcpkg](https://github.com/microsoft/vcpkg), you can easily use vcpkg to build and install it.
+```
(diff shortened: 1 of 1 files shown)


jcwchen (approved):
LGTM. Please follow [the steps](https://github.com/onnx/onnx/blob/main/CONTRIBUTING.md#dco) to signoff DCO. Thanks!

JackBoosY:
I will open a new PR to sign DCO. Sorry.

Issue #4515, opened 2022-09-15
Explicit error message when converting version for opsets from non-official domain


jcwchen:
### System information

The latest main branch: a252e651fda8e73935827505a148146459c2e06d.

### What is the problem that this feature solves?

Currently the version_converter only says: `unordered_map::at: key not found` when converting a model which consists of non-official opset. For instance, every quantized model in [ONNX Model Zoo](https://github.com/onnx/models) (*-int8.onnx). The error message should be more explicit.

### Describe the feature

Make the error message more explicit.

### Will this influence the current api (Y/N)?

N

### Feature Area

version_converter

### Are you willing to contribute it (Y/N)

Yes

PR #4521, opened 2022-09-18
Skip version conversion if non-ONNX domain and throw explicit error

jcwchen:
### Description
Skip version conversion if non-ONNX domain and throw explicit error.

### Motivation and Context
Fixes https://github.com/onnx/onnx/issues/4515.


onnx/version_converter/convert.cc
```diff
@@ -76,9 +76,9 @@ void DefaultVersionConverter::convert_graph(
                "Please be advised the converted model may not be working properly if target runtime does not support this "
                "experimental op."
             << std::endl;
-        continue;
-      }
-      if (op_name != "Undefined" && op_name != "Captured") {
+      } else if (cur_op->domain() != "" && cur_op->domain() != "ai.onnx") {
```
xadupre, line 79: What if a node is Undefined from a domain != from "" or "ai.onnx"?

jcwchen: Good question. Did you mean an operator named "Undefined" but from domain "" or "ai.onnx"? If yes, current behavior is do nothing without any error message.

liqunfu (approved):

Issue #3921, opened 2022-01-06
AttributeHasValue and StaticOptionalHasValue


gramalingam:
This is a proposal to add a couple of primitive operators, whose primary purpose is to allow us to define other ops as functions. It is intended to specifically help address scenarios where the op/function has optional attributes or optional inputs, and we need to check if the optional attribute/input was supplied or not.

# New Operator

(a) AttributeHasValue() : an operator that returns true if it has an attribute specified.
(b) StaticOptionalHasValue(X): an operator that returns true if an optional input is specified.

### Describe the operator
As mentioned above, this is intended to enable function-definition for ops that have optional
attributes or inputs which determine the behavior of the function.

### Can this operator be constructed using existing onnx operators?
No.

### Is this operator used by any model currently? Which one?
There are examples of ops: eg., CumSum has two optional attributes. Squeeze is an example of an op with an optional input. 

### Are you willing to contribute it? (Y/N)
Y

### Notes
The existing mechanism of context-dependent functions is a workaround for the above issue. However, context-dependent functions have their own disadvantages. Eg., it is not easy to embed their function expansion as a model-local function. The above primitives will allow us to express such functions as context-free functions, which are easy to export as a model-local function.

gramalingam:
Some more info about part (b) described above. There are two kinds of optional inputs in ONNX. (Eg: see https://github.com/onnx/onnx/issues/4301 ). The first kind originally used in ONNX is not part of the *type* of the input (eg., input `ratio` of `Dropout`), while the second kind incorporates `optional` as part of the type (see ops `Optional`, `OptionalHasElement` and `OptionalGetElement`).

The second kind (and associated ops) are meant to serve a similar purpose as described in this issue (to write conditional logic dependent on whether an optional input is specified or not)). However, these ops cannot be used with optional inputs of the first kind because of their current typing.

Proposal: generalize the ops `OptionalHasElement` and `OptionalGetElement` to allow optional inputs of the first kind as well. This will help unify the two kinds of optional inputs.

gramalingam:
[PR 4421](https://github.com/onnx/onnx/pull/4421) completes part (b).

As for part (a), the conclusion for now is that it can be managed using suitable default-values for attributes, and, hence, this extension is not required.

gramalingam:
Reopening this for part (a).

[...]

PR #4525, opened 2022-09-20
Add AttributeHasValue op

liqunfu:

### Description
Add AttributeHasValue op

### Motivation and Context
This PR is to solve part of the issue (AttributeHasValue and StaticOptionalHasValue): #https://github.com/onnx/onnx/issues/3921

This is an op to help convert primary ops with optional attribute to simply function ops.
This op is a function op thus not a burden for runtimes. It is not simple function ops though.


onnx/defs/tensor/defs.cc
```diff
@@ -3720,4 +3720,65 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  builder.Add("output = Cast (output0)", "to", (int64_t)(TensorProto_DataType_BOOL));
+  schema.BuildFunction(functionProto);
+  return true;
+}
+
+ONNX_OPERATOR_SET_SCHEMA(
+    AttributeHasValue,
+    18,
+    OpSchema()
+        .SetDoc(R"DOC(Returns which elements of the input are NaN.)DOC")
```
gramalingam, line 3752: Change to "Returns true iff at least one of the attribute-value is specified."

onnx/defs/tensor/defs.cc
```diff
@@ -3720,4 +3720,65 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .SetContextDependentFunctionBodyBuilder(BuildContextDependentFunctionBodyAttributeHasValue)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          const size_t numInputs = ctx.getNumInputs();
+          if (numInputs != 0) {
+            fail_type_inference("AttributeHasValue does not expected any input.");
+          }
+          const size_t numOutputs = ctx.getNumOutputs();
+          if (numOutputs != 1) {
+            fail_type_inference("AttributeHasValue is expected to have 1 output.");
+          }
```
gramalingam, line 3778: The checks above are unnecessary and can be omitted. The default-checker will already check this from the opschema information.

onnx/defs/tensor/defs.cc
```diff
@@ -3720,4 +3720,65 @@ ONNX_OPERATOR_SET_SCHEMA(
...
 
+bool BuildContextDependentFunctionBodyAttributeHasValue(
+    const FunctionBodyBuildContext& ctx,
+    const OpSchema& schema,
+    FunctionProto& functionProto) {
+  FunctionBuilder builder(functionProto);
+  auto ints_attr_proto = ctx.getAttribute("ints");
+  bool has_ints = (nullptr != ints_attr_proto) && ints_attr_proto->ints_size() > 0;
+  auto int_attr_proto = ctx.getAttribute("int");
+  bool has_int = (nullptr != int_attr_proto) && int_attr_proto->has_i();
```
gramalingam, line 3731: I guess you will add a check for all other attributes as well?

onnx/defs/tensor/defs.cc
```diff
@@ -3720,4 +3720,65 @@ ONNX_OPERATOR_SET_SCHEMA(
           return true;
         }));
 
+bool BuildContextDependentFunctionBodyAttributeHasValue(
+    const FunctionBodyBuildContext& ctx,
+    const OpSchema& schema,
+    FunctionProto& functionProto) {
+  FunctionBuilder builder(functionProto);
+  auto ints_attr_proto = ctx.getAttribute("ints");
+  bool has_ints = (nullptr != ints_attr_proto) && ints_attr_proto->ints_size() > 0;
```
gramalingam, line 3729: I think we should omit the second check on ints_size?

onnx/defs/tensor/defs.cc
gramalingam: May be something like:
```cpp
   for (const char *name : { "int", "ints", ... })
      if ctx.getAttribute(name) != nullptr {
         has_attr = true;
         break;
      }
```

onnx/defs/tensor/defs.cc
```diff
@@ -3720,4 +3720,65 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  auto int_attr_proto = ctx.getAttribute("int");
+  bool has_int = (nullptr != int_attr_proto) && int_attr_proto->has_i();
+
+  // ToTensor does not work with boolean element type. Need to create a Constant int
+  // and then Cast to tensor of boolean.
+  if (has_ints) {
+    builder.Const("output0", 1);
```
gramalingam, line 3736: I think something like
```cpp
   builder.Add("w = Constant < value = bool {1} > ()");
```
may work?

onnx/backend/test/case/node/attributehasvalue.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+
+            node = onnx.helper.make_node(
+                "AttributeHasValue",
+                inputs=[],
+                outputs=["output"],
+                **kwargs,
+            )
+
+            output = np.array(True)
```
gramalingam, line 36: May be replace `True` by `bool(kwargs)` ... then you can use the same test-case with no-attributes as well as one-or-more attributes.

onnx/defs/tensor/defs.cc
liqunfu: Yes I will find a way to iterate all types defined in proto. For now, I just want to covert ints so that I can validate it in converting reduce ops to function ops.

onnx/defs/tensor/defs.cc
liqunfu: I think I added the check due to test failure. Let me double check.

onnx/defs/tensor/defs.cc
liqunfu: will try

onnx/defs/tensor/defs.cc
```diff
@@ -3720,4 +3720,65 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+  } else if (has_int) {
+    builder.Const("output0", 1);
```
jcwchen, line 3738: Merge this condition (has_int) with the condition of has_ints?

onnx/defs/tensor/defs.cc
liqunfu: Sorry I misunderstood your comment. I will remove the second check.

onnx/backend/test/case/node/attributehasvalue.py
```diff
@@ -23,7 +26,7 @@ def test_one_attribute(name, **kwargs):
                 node,
                 inputs=[],
                 outputs=[output],
-                name="test_attribute_has_value_{name}_false".format(name=name),
+                name=f"test_attribute_has_value_{name}_false",
```
gramalingam, line 29: May be change to `test_attribute_has_{name}_false` ? Or, omit the `value_` part in the call-site. Otherwise, we will get `value_value_int`.

gramalingam (approved):
LGTM (except for the nit about repeated value_value in the name), thanks!

gramalingam (approved):

Issue #4499, opened 2022-09-08
[Feature request] Reduce potential clutter in "Examples" section in operator docs


p-wysocki:
### System information

N/A

### What is the problem that this feature solves?

Contributors are currently recommended to keep tests for previous opsets when adding a new operator. This comes down to refactoring tests into a following fashion:

`def example_test_opset13()`
`def example_test_opset18()`

A more accurate example has been given in https://github.com/onnx/onnx/pull/4481/files#diff-104e42616962149b1dd1280f0464bc1c1a45c197934e9f21c26ba4900cfbb0eb (file `split.py`).

It results in a cluttered `Examples` section in the `Operators.md` file, example of which can be seen here: https://github.com/p-wysocki/onnx/blob/pwysocki/new_split18/docs/Operators.md#examples-142. In case the link is no longer available I'm also attaching a screenshot:
![image](https://user-images.githubusercontent.com/55858107/189184949-ce4079c1-c538-4ee8-a0c3-71fd4722f4c4.png)

Only the newest version is included in the `Operators.md`, so it may not be the place for older opset tests. It clutters the view and forces the user to visually make sure that he/she is looking at the correct opset. The pictured section may seem small at first glance, but it will build up more and more as time goes by, for each operator.

### Alternatives considered

An easy solution would be to drop the idea of keeping older opset tests, but I do not think it's appropriate.

### Describe the feature

The feature could involve, for example:

- parsing the test name to see which opset it's related to and putting into docs only if it tests the newest opset,
- checking the `opset_imports` value for a test (https://github.com/onnx/onnx/pull/4481/files#diff-104e42616962149b1dd1280f0464bc1c1a45c197934e9f21c26ba4900cfbb0ebR76),
- adding a different directory/file to store "legacy" tests, such as it's done with `old.cc`.

### Will this influence the current api (Y/N)?

N

### Feature Area

Documentation

### Are you willing to contribute it (Y/N)

Yes

Issue #2290, opened 2019-09-06
Inconsistencies between MaxPool, AveragePool, and LpPool specs


JamesAllingham:
The `MaxPool` op has a few attributes that neither `AveragePool` nor `LpPool` have (`dilations` and `storage_order`). It isn't clear to me why these attributes should be exclusive to `MaxPool`. Though, to be honest, I'm not really sure what `storage_order` is for at all :P.

Similarly, `LpPool` only has `auto_pad`, `kernel_size`, and `strides` in common with `AveragePool` and `MaxPool`.  It is not clear to me why `ceil_mode`, and `pads` should not apply here too.

(On the other hand, I do see why the `count_include_pad` attribute of  `AveragePool` and the `p` attribute of  `LpPool` are specific to their ops.)

Can anyone provide some insight here?

ebarsoum:
storage_order specify if the tensor format is row major or column major, this shouldn't be part of the spec.
Regarding dilation, because if you compute average or norm, should you count the skipped cells or not? We can add an attribute similar to count_include_pad for that, but I am not aware of any model that use dilation with average and LpPool.

kraiskil:
I wonder if the `storage_order` could be removed? Or at least deprecated? It feels like an unnecessary pain to implement in a backend.

jcwchen:
Reopen it since this issue still exists.

p-wysocki:
I'm working on it.

p-wysocki:
Related PRs: https://github.com/onnx/onnx/pull/4534 and https://github.com/onnx/onnx/pull/4533.

gramalingam:
As a clarification: the `storage_order` is used in MaxPool to convert the n-tuple of indices (i0, i1, i2, ...) into a single integer for its second output. Since the other pooling ops don't have such an output, it is not relevant for them. It could have been avoided if the output returned the n-tuple of indices (which is what an op like `NonZero` does). But it's not worth making a compatibility-breaking change to the op at this point.

p-wysocki:
Most likely final PR (WIP at the time of writing): https://github.com/onnx/onnx/pull/4790

Alignment after it gets merged:
![image](https://user-images.githubusercontent.com/55858107/213475201-968cb51b-008a-4f2a-9985-4c64b5f9c03e.png)

p-wysocki:
The last PR has been merged. The current alignment:
[...]

PR #4533, opened 2022-09-22
Add MaxPool-18 - remove `storage_order` attribute

p-wysocki:

### Description
Add `MaxPool-18`, which no longer has `storage_order` attribute.

### Motivation and Context
- Partially solves https://github.com/onnx/onnx/issues/2290, but more discussion may be needed.

### Notes
- Only a single test at `onnx/backend/test/case/node/maxpool.py` has been moved as an explicit MaxPool-12 test. I believe the way we should go about testing previous versions of operators needs to be clarified (related to https://github.com/onnx/onnx/issues/4499).
image
A related PR concerns adding `ceil_mode` attribute to the `LpPool` operator: https://github.com/onnx/onnx/pull/4534.


docs/Changelog.md
```diff
@@ -14466,7 +14466,7 @@ This version of the operator has been available since version 12 of the default
 
 <dl>
 <dt><tt>auto_pad</tt> : string (default is NOTSET)</dt>
-<dd>auto_pad must be either NOTSET, SAME_UPPER, SAME_LOWER or VALID. Where default value is NOTSET, which means explicit padding is used. SAME_UPPER or SAME_LOWER mean pad the input so that `output_shape[i] = ceil(input_shape[i] / strides[i])` for each axis `i`. The padding is split between the two sides equally or almost equally (depending on whether it is even or odd). In case the padding is an odd number, the extra padding is added at the end for SAME_UPPER and at the beginning for SAME_LOWER.</dd>
+<dd>auto_pad must be either NOTSET, SAME_UPPER, SAME_LOWER or VALID. Where default value is NOTSET, which means explicit padding is used. SAME_UPPER or SAME_LOWER mean pad the input so that the output spatial size match the input.In case of odd number add the extra padding at the end for SAME_UPPER and at the beginning for SAME_LOWER. VALID mean no padding.</dd>
```
kraiskil, line 14469: Reading `the output spatial size match the input` (in my broken English) sounds a bit like "input size is equal to output size". Is the original: `output_shape[i] = ceil(input_shape[i] / strides[i])` for each axis `i` no longer valid?

p-wysocki: I do not actually remember changing that at all, I am not sure why it appeared in the changelog. I'll look into it.

onnx/version_converter/convert.h
```diff
@@ -523,6 +524,7 @@ class DefaultVersionConverter : public BaseVersionConverter {
     registerAdapter(make_unique<CompatibleAdapter>("OptionalHasElement", OpSetID(17), OpSetID(18)));
     registerAdapter(make_unique<CompatibleAdapter>("ScatterND", OpSetID(17), OpSetID(18)));
     registerAdapter(make_unique<CompatibleAdapter>("ScatterElements", OpSetID(17), OpSetID(18)));
+    registerAdapter(make_unique<MaxPool_17_18>());
```
gramalingam, line 527: You can used the predefined parametrized adapter `RemoveAttribute` ... no need to define a new adapter class.
```cpp
   registerAdapter("MaxPool", 17, 18, RemoveAttribute(kstorage_order));
```

onnx/version_converter/adapters/maxpool_17_18.h
```diff
@@ -0,0 +1,30 @@
...
+// Adapter for MaxPool in default domain from version 17 to 18
+
+#pragma once
+
+#include "onnx/version_converter/adapters/adapter.h"
+
+namespace ONNX_NAMESPACE {
+namespace version_conversion {
+
+class MaxPool_17_18 final : public Adapter {
```
gramalingam, line 14: No need for a specialized adapter. Please see comment below.

onnx/backend/test/case/node/maxpool.py
```diff
@@ -186,7 +186,7 @@ def export_maxpool_2d_precomputed_strides() -> None:
         )
 
     @staticmethod
-    def export_maxpool_with_argmax_2d_precomputed_strides() -> None:
+    def export_maxpool_with_argmax_2d_precomputed_strides_opset12() -> None:
```
gramalingam, line 189: Does the `storage_order` attribute affect the output-value? I assume it has no effect on the output, is that correct? Just trying to make sure I understand. If it has an effect, then the adapter needs to be different.

kraiskil: I don't see how the storage order should affect on the python code level. Also looking at the `.pb` tensors files generated out of this test (both input and output), the data seems to be plain-old-row-major. (which makes sense - grepping for `storage_order` in the onnx sources shows it is nowhere implemented).

But indeed, the output *is* affected - the optional output-indices (tensor z) calculated in this test seem to have indices calculated in a column-major fashion, and `z` needs transposing for the test to be correct.

onnx/version_converter/convert.h
p-wysocki: Good point, done.

onnx/version_converter/adapters/maxpool_17_18.h
p-wysocki: Done.

docs/Changelog.md
p-wysocki: Fixed, there was a wrong reference to a docstring in `old.cc`. Thanks for letting me know!

onnx/backend/test/case/node/maxpool.py
p-wysocki: Should `z` be corrected if this test is for `opset12` only? The operator behavior has not been changed, so as long as `opset_imports=[onnx.helper.make_opsetid("", 12)],` is present, the test should pass just like it did before. I did not reintroduce this test to `opset18` because `storage_order` is being removed.

kraiskil: No I don't think `z` is wrong in the old opset tests where `storage_order` is set.
Only when storage_order is removed does `z` become incorrect - I just didn't notice the test was being removed. Removing it is a good fix for this :)

gramalingam: My question was motivated by the adapter, rather than this test-case. If I understand correctly, simply omitting the storage_order attribute is not correct for converting a model with the old op to the new op ... do we need to insert some transpose to get the desired behavior of the original op (when the attribute is used)?

gramalingam: In other words: it looks like the storage_order is a meaningful attribute ... it is used to convert the n-tuple of indices into a single index. It makes sense for MaxPool, since the value comes from a single index (ignoring duplicate max values), unlike the other pooling operations, and it is not relevant for other pooling ops, since they don't have such an op. So, I wonder whether we should just the op be as is, instead of removing the attribute?

p-wysocki:
After an internal discussion it has been decided that `storage_order` should be kept in order to maintain backwards compatibility.

PR #4534, opened 2022-09-22
Add LpPool-18 - add `ceil_mode` and `dilations` attributes

p-wysocki:

### Description
Add `LpPool-18`, which now has `ceil_mode` and `dilations` attributes. Deprecate `auto_pad`, like it is done in the case of [AveragePool](https://github.com/onnx/onnx/blob/main/docs/Operators.md#averagepool) and [MaxPool](https://github.com/onnx/onnx/blob/main/docs/Operators.md#maxpool).

### Motivation and Context
- Partially solves https://github.com/onnx/onnx/issues/2290, but more discussion may be needed.

### Notes
- [LpPoolOpSchemaGenerator](https://github.com/onnx/onnx/blob/main/onnx/defs/nn/defs.cc#L541) uses [convPoolShapeInference](https://github.com/onnx/onnx/blob/main/onnx/defs/nn/defs.cc#L36), which [already accounts](https://github.com/onnx/onnx/blob/main/onnx/defs/nn/defs.cc#L176-L180) for `ceil_mode` attribute. As far as I understand, it makes adding the attribute to `ceil_mode` trivial, as it is done in this PR. The same goes for the `dilations` attribute.
- The tests for `LpPool` are non-existent (there is not such a file as `onnx/onnx/backend/test/case/node/lppool.py`). Is it intentional? Pooling operators seem to have a lot of common code, so maybe `LpPool` is considered to be covered by tests for `MaxPool` and `AveragePool`. I am not sure if it is intended or an overlook.
image
A related PR concerns removing `storage_order` attribute from the MaxPool operator: https://github.com/onnx/onnx/pull/4533.


kraiskil:
I don't see why explicit `LpPool` tests would be left out intentionally. After all, the differnt pools calculate different things.

In my experience, when writing a backend, such coverage tests are very valuable, partly as a safety net but also as additional documentation / clarification. I can not see any big cost in adding such tests, except of course for the initial effort of writing them.

onnx/version_converter/adapters/lppool_17_18.h
```diff
@@ -0,0 +1,31 @@
...
+#include "onnx/version_converter/adapters/transformers.h"
+
+namespace ONNX_NAMESPACE {
+namespace version_conversion {
+
+class LpPool_17_18 final : public Adapter {
+ public:
+  explicit LpPool_17_18() : Adapter("LpPool", OpSetID(17), OpSetID(18)) {}
+
+  void adapt_lppool_17_18(std::shared_ptr<Graph>, Node* node) const {
```
gramalingam, line 19: I guess this file is not needed? `convert.h` seems to use a CompatibleAdapter, which seems correct.

gramalingam:
Agree with @kraiskil ... adding the test-cases would be useful.

gramalingam:
Why not add the `dilations` attribute as well?

onnx/version_converter/adapters/lppool_17_18.h
p-wysocki: Good catch, I deleted the file and used `registerAdapter("LpPool", 17, 18, SetAttribute(kceil_mode, 0));` instead.

p-wysocki:
@kraiskil @gramalingam I'm working on both adding tests and adding `dilations` attribute.

onnx/backend/base.py
```diff
@@ -60,7 +60,7 @@ class BackendRep:
     BackendRep to retrieve the corresponding results.
     """
 
-    def run(self, inputs: Any, **kwargs: Any) -> Tuple[Any, ...]:
+    def run(self, inputs: Any, **kwargs: Any) -> Tuple[Any, ...]:  # type: ignore
```
p-wysocki, line 63: I had to ignore this line in `mypy` to pass linter check. It was throwing
`onnx/backend/base.py:63:5: error: Missing return statement  [empty-body]`
Is this a known issue? I did not make changes to this file. If you want me to fix it please let me know.

gramalingam: I see Jacky has a PR: https://github.com/onnx/onnx/pull/4643

jcwchen: Just merged. Updating the latest main branch in this PR should fix this issue. Thanks!

gramalingam (approved):

Issue #4532, opened 2022-09-22
Relax the protobuf version requirement.


xkszltl:
# Bug Report

### Describe the bug

Protobuf now has 3.20.2, so recommend to write <4 instead:
- https://github.com/onnx/onnx/blob/895593af6ac17d4fe2749091659083d3b1246cae/requirements.txt#L2
- https://pypi.org/project/protobuf/3.20.2/

jcwchen:
Hi @xkszltl,
Thank you for catching this! Fresh #4535 should fix this.

NikolaBorisov:
Why <4? I need to use the latest version of protobuf and onnx.

jcwchen:
There is a breaking change in Protobuf 4.21: https://github.com/protocolbuffers/protobuf/issues/10051. For now many Protobuf related project such as ONNX, PyTorch, TensorFlow still use Protobuf 3 to build wheels and it will bump into issues with these wheels with Protobuf 4 due to breaking change. (For example, https://github.com/onnx/onnx/issues/4222).

To workaround, a possible way might be building ONNX from source with Protobuf 4 (protoc>=21. and Python Protobuf >=4.21.) and then the installed onnx should be compatible with the latest Protobuf.

snnn:
>  building ONNX from source with Protobuf 4

Then the user also needs to patch requirements.txt

Zahlii:
With tensorflow 2.12 going to release soon (and allowing proto > 4), is there a way to relax this dependency?
As I am on a M2 chip, all older grpcio/protobuf versions cannot be installed directly, making it impossible for me to run latest tensorflow release candidate and train/export to ONNX.

PR #4535, opened 2022-09-22
Relax protobuf version requirement for version below 4

jcwchen:
### Description
Protobuf now has 3.20.2. Using Protobuf<4 should be better.

### Motivation and Context
Fixes https://github.com/onnx/onnx/issues/4532.


liqunfu (approved):

snnn:
But I found an interesting thing: ORT's Windows CI build pipelines build ONNX python package from source with protobuf 3.21.8. Then when ORT tries to run its python tests(which depends on protobuf and onnx python packages), then if the protobuf version is 3.18.3, the tests fail. However, if we uninstall protobuf and install onnx again, because onnx says the max supported protobuf version is 3.20.1, then pip will get that version, and it works!

jcwchen:
Thanks for bringing this up. I bumped into similar issue before... I was considering to aggressively bump Protobuf version as 3.20 for building ONNX, but there were some test failures if the built ONNX played with Python Protobuf<=3.19. Therefore, if ONNX decides to bump its Protobuf version for building ONNX as 3.20, probably ONNX needs to upgrade its required Python Protobuf as `protobuf >= 3.20, < 4` in https://github.com/onnx/onnx/blob/main/requirements.txt. The compatibility is kind of limited in that case.

Thus, I took a step back in next release: I decided to bump Protobuf version as 3.18.3 for building ONNX instead and so ONNX can preserve much broader compatibility for Python Protobuf version: https://github.com/onnx/onnx/pull/4544. Still, in the future while ONNX needs higher Protobuf version, this issue will hit us again anyway. We need to keep in mind about the limited compatibility issue for Protobuf>=3.20 or Protobuf>=3.21.

snnn:
But they say 3.18.x releases are end of support? Should you use a newer one?

jcwchen:
Good question. Protobuf 3.19.5 or 3.19.6 has another issue about fixed build_type in their build config. So the options we have here are 3.18.3 or 3.20.2+. I slightly prefer 3.18.3, because other ONNX related tools might have Protobuf version requirement and using Protobuf 3.20.2+ for building ONNX might break something, as your example above, if they really need lower Python Protobuf version (<3.20) somehow. In addition, we can add an announcement in next release and say in the following release ONNX will upgrade its Protobuf version to 3.20 for build and the minimum required Python Protobuf version will be 3.20, to let users be aware of it in advance.

IIRC, ONNX Runtime is using Protobuf 3.18.3 for build also. @snnn Are you planning to upgrade ORT's Protobuf version to 3.20+ soon? If yes, definitely we should re-evaluate if ONNX should upgrade its Protobuf version to 3.20, to prevent some conflict issue in the near future.

jcwchen:
> But they say 3.18.x releases are end of support?

Forgot to answer your question -- yes Protobuf people said 3.18.x are out of their support window, but 3.18.3 is a quite fresh patch they have released last month.

snnn:
> Are you planning to upgrade ORT's Protobuf version to 3.20+ soon?

yes. For this issue: https://github.com/microsoft/onnxruntime/issues/13335

But the change won't be as trivial as before. It needs some code changes.

Issue #4549, opened 2022-09-27
printable_graph error and INVALID_GRAPH (ONNXRuntime) when using bf16 datatype


brevity2021:
# Bug Report

### Is the issue related to model conversion?
No, ONNX checker does not report issues

### Describe the bug
`onnx.helper.printable_graph` throws key error when loading a model with data type bfloat16. But `onnx.checker.check_model` does not give any error. Also ONNXRuntime throws invalid graph error when loading the model.

The errors are as follows (I also include these in the `Reproduction instructions`):
```
File /usr/local/lib/python3.8/dist-packages/onnx/helper.py:817, in printable_attribute(attr, subgraphs)
 814         content.append("<Tensor>")
    815     else:
    816         # special case to print scalars
--> 817         field = STORAGE_TENSOR_TYPE_TO_FIELD[attr.t.data_type]
    818         content.append(f'<Scalar Tensor {str(getattr(attr.t, field))}>')
    819 elif attr.HasField("g"):

KeyError: 16
```

This may result in the INVALID_GRAPH error when using the ONNXRuntime to create an InferenceSession from it.
```
File /usr/local/lib/python3.8/dist-packages/onnxruntime/capi/onnxruntime_inference_collection.py:384, in InferenceSession._create_inference_session(self, providers, provider_options, disabled_optimizers)
    382 session_options = self._sess_options if self._sess_options else C.get_default_session_options()
    383 if self._model_path:
--> 384     sess = C.InferenceSession(session_options, self._model_path, True, self._read_config_from_model)
    385 else:
    386     sess = C.InferenceSession(session_options, self._model_bytes, False, self._read_config_from_model)

InvalidGraph: [ONNXRuntimeError] : 10 : INVALID_GRAPH : Load model from ./pegasus_cnn/encoder.onnx failed:This is an invalid model. Type Error: Type 'tensor(bfloat16)' of input parameter (onnx::Pow_284) of operator (Pow) in node (Pow_349) is invalid.
```

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): Linux Ubuntu 20.04
- ONNX version (*e.g., 1.7*):  1.12.0 (installed from pip)
- Python version: 3.8.10
- ONNX Runtime version (*e.g. 1.7*):  1.12.0

### Reproduction instructions
```
from onnxruntime import InferenceSession
import onnx
import torch

from transformers import AutoTokenizer,PegasusForConditionalGeneration

[...]

PR #4551, opened 2022-09-28
Fix several issues regarding recent mapping update

jcwchen:
### Description
- Add missing case for SequenceProto.MAP in numpy_helper.to_list.
- Fix issue onnx/issues/4552, make `tensor_dtype_to_field` not throw deprecation warning by introducing a identical dictionary `_STORAGE_TENSOR_TYPE_TO_FIELD`. Same case for `NP_TYPE_TO_TENSOR_TYPE`.
- Partially fix https://github.com/onnx/onnx/issues/4549: enable `printable_attribute` to handle bfloat16.

### Motivation and Context
Recent https://github.com/onnx/onnx/pull/4270 refactored existing code, but I forgot to add the case for SequenceProto.MAP in to_list.

cc @gramalingam Please review this PR. Sorry that I should catch this in advance before my PR got merged. Thank you!

### DIFF
--- a/onnx/onnx.proto
+++ b/onnx/onnx.proto
@@ -542,7 +542,7 @@ message TensorProto {
   // float16 values must be bit-wise converted to an uint16_t prior
   // to writing to the buffer.
   // When this field is present, the data_type field MUST be
-  // INT32, INT16, INT8, UINT16, UINT8, BOOL, or FLOAT16
+  // INT32, INT16, INT8, UINT16, UINT8, BOOL, FLOAT16 or BFLOAT16
   repeated int32 int32_data = 5 [packed = true];
 
   // For strings.
--- a/onnx/test/helper_test.py
+++ b/onnx/test/helper_test.py
@@ -743,6 +743,17 @@ class TestHelperMappingFunctions(unittest.TestCase):
     def test_np_dtype_to_tensor_dtype_not_throw_warning(self) -> None:
         _ = helper.np_dtype_to_tensor_dtype(np.dtype("float32"))
 
+    def test_tensor_dtype_to_np_dtype_bfloat16(self) -> None:
+        self.assertEqual(
+            helper.tensor_dtype_to_np_dtype(TensorProto.BFLOAT16), np.dtype("float32")
+        )
+
+    def test_tensor_dtype_to_storage_tensor_dtype_bfloat16(self) -> None:
+        self.assertEqual(
+            helper.tensor_dtype_to_storage_tensor_dtype(TensorProto.BFLOAT16),
+            TensorProto.UINT16,
+        )
+
     # BFloat16 tensor uses TensorProto.UINT16 as storage type;
     # And the field name for TensorProto.UINT16 is int32_data
     def test_tensor_dtype_to_field_bfloat16(self) -> None:
--- a/onnx/test/helper_test.py
+++ b/onnx/test/helper_test.py
@@ -724,6 +724,26 @@ def test_make_tensor_raw(tensor_dtype: int) -> None:
     np.testing.assert_equal(np_array, numpy_helper.to_array(tensor))
 
 
+# TODO (#4554): remove this test after the deprecation period
+# Test these new functions should not raise any depreaction warnings
+class TestHelperMappingFunctions(unittest.TestCase):
+    @pytest.mark.filterwarnings("error::DeprecationWarning")
+    def test_tensor_dtype_to_np_dtype_not_throw_warning(self) -> None:
+        _ = helper.tensor_dtype_to_np_dtype(TensorProto.FLOAT)
+
+    @pytest.mark.filterwarnings("error::DeprecationWarning")
+    def test_tensor_dtype_to_storage_tensor_dtype_not_throw_warning(self) -> None:
+        _ = helper.tensor_dtype_to_storage_tensor_dtype(TensorProto.FLOAT)
+
+    @pytest.mark.filterwarnings("error::DeprecationWarning")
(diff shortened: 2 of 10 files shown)


gramalingam (approved):

linkerzhang (approved):
LGTM.

gramalingam (approved):

xadupre:
It is possible to add a unit test to check the warning is raised when using the dictionary and not raised when using the function?

jcwchen:
> It is possible to add a unit test to check the warning is raised when using the dictionary and not raised when using the function?

Good idea. I just added tests by `TestHelperMappingFunctions` and actually `np_dtype_to_tensor_dtype` needs to be fixed as well. With the added tests, now I am confident that these new introduced helpers should not unexpectedly throw a deprecation warnings. Please review it. Thank you!

onnx/test/helper_test.py
```diff
@@ -724,6 +724,26 @@ def test_make_tensor_raw(tensor_dtype: int) -> None:
     np.testing.assert_equal(np_array, numpy_helper.to_array(tensor))
 
 
+# TODO (#4554): remove this test after the deprecation period
+# Test these new functions should not raise any depreaction warnings
```
xadupre, line 728: Misspelling deprecation

onnx/test/helper_test.py
```diff
@@ -724,6 +724,26 @@ def test_make_tensor_raw(tensor_dtype: int) -> None:
...
+class TestHelperMappingFunctions(unittest.TestCase):
+    @pytest.mark.filterwarnings("error::DeprecationWarning")
+    def test_tensor_dtype_to_np_dtype_not_throw_warning(self) -> None:
+        _ = helper.tensor_dtype_to_np_dtype(TensorProto.FLOAT)
+
+    @pytest.mark.filterwarnings("error::DeprecationWarning")
```
xadupre, line 734: I wonder if it is worth checking what happens in case of bfloat16.

onnx/test/helper_test.py
jcwchen: Corrected. Thanks!

onnx/test/helper_test.py
jcwchen: Tests for bfloat16 looks useful since the mapping for bfloat16 is quite confusing. Adding these tests can prevent future regression. Just added them. PTAL. Thanks!

onnx/onnx.proto
```diff
@@ -542,7 +542,7 @@ message TensorProto {
   // float16 values must be bit-wise converted to an uint16_t prior
   // to writing to the buffer.
   // When this field is present, the data_type field MUST be
-  // INT32, INT16, INT8, UINT16, UINT8, BOOL, or FLOAT16
+  // INT32, INT16, INT8, UINT16, UINT8, BOOL, FLOAT16 or BFLOAT16
```
jcwchen, line 545: Please note that I think BFLOAT16 is missing here.

onnx/test/helper_test.py
```diff
@@ -724,6 +724,44 @@ def test_make_tensor_raw(tensor_dtype: int) -> None:
...
+        )
+
+    # BFloat16 tensor uses TensorProto.UINT16 as storage type;
+    # And the field name for TensorProto.UINT16 is int32_data
+    def test_tensor_dtype_to_field_bfloat16(self) -> None:
+        self.assertEqual(
+            helper.tensor_dtype_to_field(TensorProto.BFLOAT16), "int32_data"
+        )
+
+
```
xadupre, line 764: I can be done in another PR but I would check all the types:

```python
    def test_numeric_types(self):  # type: ignore
        dtypes = [
            np.float16,
            np.float32,
            np.float64,
            np.int8,
            np.int16,
            np.int32,
            np.int64,
            np.uint8,
            np.uint16,
            np.uint32,
            np.uint64,
            np.complex64,
            np.complex128,
        ]
        for dt in dtypes:
            with self.subTest(dtype=dt):
                t = np.array([0, 1, 2], dtype=dt)
                ot = from_array(t)
                u = to_array(ot)
                self.assertEqual(t.dtype, u.dtype)
                assert_almost_equal(t, u)

    def test_make_tensor(self):  # type: ignore
        for pt, dt in TENSOR_TYPE_TO_NP_TYPE.items():
            if pt == TensorProto.BFLOAT16:
                continue
            with self.subTest(dt=dt, pt=pt, raw=False):
                if pt == TensorProto.STRING:
                    t = np.array([["i0", "i1", "i2"], ["i6", "i7", "i8"]], dtype=dt)
                else:
                    t = np.array([[0, 1, 2], [6, 7, 8]], dtype=dt)
                ot = make_tensor("test", pt, t.shape, t, raw=False)
                self.assertFalse(ot is None)
                u = to_array(ot)
                self.assertEqual(t.dtype, u.dtype)
                self.assertEqual(t.tolist(), u.tolist())
            with self.subTest(dt=dt, pt=pt, raw=True):
                t = np.array([[0, 1, 2], [6, 7, 8]], dtype=dt)
                if pt == TensorProto.STRING:
                    with self.assertRaises(TypeError):
                        make_tensor("test", pt, t.shape, t.tobytes(), raw=True)
                else:
                    ot = make_tensor("test", pt, t.shape, t.tobytes(), raw=True)
                    self.assertFalse(ot is None)
                    u = to_array(ot)
                    self.assertEqual(t.dtype, u.dtype)
                    assert_almost_equal(t, u)
```

jcwchen: Thank you @xadupre for the suggestion. It looks good to me, but I am inclined to have it in another PR (this PR more focuses on bug fixes) Feel free to let me know if you still have other concern.

PR #4555, opened 2022-09-29
update req-dev

AllenTiTaiWang:
fixes https://github.com/microsoft/onnx-script/pull/143#issuecomment-1261584469


requirements-dev.txt
```diff
@@ -15,4 +15,5 @@ clang-format==13.0.0
 mypy>=0.971
 types-protobuf==3.18.4
 black>=22.3
-isort[colors]>=5.10
+isort>=5.10.1
+colorama==0.4.5
```
jcwchen, line 19: Is it OK to use `>=` instead? 
```suggestion
colorama>=0.4.5
```

AllenTiTaiWang: Yes, >= makes more sense here. @justinchuby I see you are setting up fixed version for lint in onnx-script. Is there any concerns that we use >= ?

justinchuby: Not really. whatever works

setup.py
```diff
@@ -346,7 +346,8 @@ def run(self):
     "mypy>=0.971",
     "types-protobuf==3.18.4",
     "black>=22.3",
-    "isort[colors]>=5.10",
+    "isort>=5.10.1",
+    "colorama==0.4.5",
```
jcwchen, line 350: Ditto.
```suggestion
    "colorama>=0.4.5",
```

AllenTiTaiWang:
> Do you know why it didn't fail previously in ONNX CIs? Is it simply because colorama might be missing in some environments?

I think it gives the warning, but not sure why it didn't make tests flaky. @justinchuby Do you know what happened?

AllenTiTaiWang:
BTW, I don't have write access on this repo ...

justinchuby:
Not completely sure. Could it be a glitch in pip? It only happens for py37 from what I have seen

requirements-dev.txt
```diff
@@ -15,4 +15,5 @@ clang-format==13.0.0
 mypy>=0.971
 types-protobuf==3.18.4
 black>=22.3
-isort[colors]>=5.10
+isort>=5.10.1
+colorama>=0.4.5
```
jcwchen, line 19: To make `requirements-dev.txt`compact, if it only fails with Python 3.7 somehow while using isort, could we add a TODO here saying colorama can be removed after Python 3.7 is deprecated?

It seems to me at least the latest isort does set colorama as a required dependency: https://github.com/PyCQA/isort/blob/12cc5fbd67eebf92eb2213b03c07b138ae1fb448/pyproject.toml#L47.

justinchuby: looking at the pyproject, colorama is set as an optional dependency. usually installing `isort[colors]` will work (which downloads colorama). In this case something's wrong with the version resolution.

Do we actually see prolonged test runs in onnx for py37? It could also be that onnx-script didn't have `wheels` in the environment before.

AllenTiTaiWang: Even though we don't really see the prolonged test runs for py37, it's potentially a risk proved in another repo, and now that we already have this PR, I think it's no harm to merge? But I am fine either way.

jcwchen:
For simplicity, I am inclined to take this PR only if there is an actual issue happening from onnx side. Now I will close this one. Thank you @AllenTiTaiWang for proposing this PR and it will still be helpful if we do see similar issues in the future.

PR #4556, opened 2022-09-29
Move check_is_experimental_op from check_node to check_graph

jcwchen:
### Description
- Move check_is_experimental_op from check_node to check_graph and check_function.
- Use std::cout instead of std::cerr since it's a warning. (Not sure if it is OK)

### Motivation and Context
Compress a lot of similar warnings from check_node into single warning from check_graph.


jcwchen:
cc @askhade Thank you for bringing this up to make checker more succinct.

liqunfu (approved):

onnx/checker.cc
```diff
@@ -595,19 +595,25 @@ void check_attribute(const AttributeProto& attr, const CheckerContext& ctx, cons
   }
 }
 
+void print_warning_if_has_experimental(const std::unordered_set<std::string>& existed_experimental_ops) {
```
gramalingam, line 598: Minor nit: `used_experimental_ops` would be a slightly better name than `existed_experimental_ops`.

onnx/checker.cc
```diff
@@ -733,6 +744,10 @@ void check_graph(const GraphProto& graph, const CheckerContext& ctx, const Lexic
       }
     }
 
+    if (check_is_experimental_op(node.op_type())) {
```
gramalingam, line 747: I think we should check that the domain is the standard domain also. If it is some other domain, we should not treat it as experimental. (I realize that this is a pre-existing issue, but would be good to fix it.)

onnx/checker.cc
jcwchen: Updated.

onnx/checker.cc
jcwchen: Good point. I have further considered domain. Please note that I also changed the public function `check_is_experimental_op`'s interface from
`bool check_is_experimental_op(std::string node_op_type)`
to
`bool check_is_experimental_op(const NodeProto& node)`. I think right now check_is_experimental_op is only used internally in onnx repo (shape_inference's implementation.cc) and passing the node itself is simpler. Let me know if you have concern with it. Thanks!

onnx/checker.cc
```diff
@@ -1022,8 +1041,9 @@ std::set<std::string> experimental_ops = {
     "Scale",
     "ScaledTanh"};
 
-bool check_is_experimental_op(std::string node_op_type) {
-  return (experimental_ops.count(node_op_type)) ? true : false;
+bool check_is_experimental_op(const NodeProto& node) {
+  return (node.domain() == ONNX_DOMAIN || node.domain() == "ai.onnx") && experimental_ops.count(node.op_type()) ? true
```
gramalingam, line 1045: Looks good, thanks!

Another nit (about the old code): I think the ` ? true : false` at the end is unnecessary and can be omitted?

gramalingam (approved):

onnx/checker.cc
jcwchen: Good catch.. Will update it soon. Thanks!

gramalingam (approved):

PR #4564, opened 2022-09-30
Fix documentation generation

xadupre:
### Description

Current pipeline to generate the documentation run sphinx based on the version released on pypi. This change installs the development version before running the documentation. It also runs the documentation to check a check does not break it.


.azure-pipelines/Linux-CI.yml
```diff
@@ -123,3 +128,10 @@ jobs:
       fi
 
     displayName: Test backend test data
+
+  - script: |
+      if [ '$(documentation)' == '1' ]; then
+        pip install -r docs/docsgen/source/requirements.txt
+        cd docs/docsgen && make html
+      fi
+    displayName: Builds documentation
```
jcwchen, line 137: So is this builds doc in this CI more like for check/test purpose? If yes, shall we use a more precise displayName?

xadupre: Test documentation?

jcwchen: Sounds good!

jcwchen (approved):

PR #4568, opened 2022-10-03
Fix shebang for protoc-gen-mypy.sh script

GaetanLepage:
### Description
Changed the shebang in `tools/protoc-gen-mypy.sh.in` from `#!/bin/bash` to `#!/usr/bin/env bash`.

### Motivation and Context
The `tools/protoc-gen-mypy.sh.in` file which generates the `protoc-gen-mypy.sh` script employs a shebang which assumes `bash` is in `/bin`.
However, this is not the case on all unix systems.
I propose to use the `#!/usr/bin/env bash` shebang instead.


jcwchen (approved):
LGTM. Thanks for catching this!

PR #4570, opened 2022-10-04
Build onnx in page creating CI

jcwchen:
### Description
- Build onnx in page creating CI with the full prerequisite such as upgrading pip, installing shared protoc, setting config.
- All CIs use the same version of actions/checkout@v3

### Motivation and Context
The CI of Deploy static content to Pages is still failing because ONNX was not built successfully.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -39,7 +39,6 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v3
-
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -11,9 +11,9 @@ jobs:
     name: Lint Python
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Setup Python
-        uses: actions/setup-python@v4.1.0
+        uses: actions/setup-python@v4
         with:
           python-version: "3.10"
       # Install onnx so that it is found by the linters
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -30,17 +30,27 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+      - name: Setup Python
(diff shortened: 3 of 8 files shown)


xadupre (approved):

jcwchen:
@onnx/sig-archinfra-approvers please review/approve this PR and let's make this CI green first.

liqunfu (approved):

jcwchen:
Now the [pages.yml](https://github.com/onnx/onnx/blob/main/.github/workflows/pages.yml) can build ONNX successfully, but it still fails later while building docs: `cd docs/docsgen && make html -j auto`. I saw a similar failure in AzurePipelines (it shows warning instead of error [there](https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=31544&view=logs&j=983f5ddd-a6d2-5453-804f-c4202009b4f1&t=23721ae6-d1a8-5bda-b271-f882ff462c68)). It is a warning so the AzurePipeline CI doesn't catch this failure. @xadupre do you know what's the exact failure is? Did you bump into similar issue from your local?

xadupre:
This is expected. As you can see, the build shows build succeeded. The errors tells that sphinx cannot compile some part of the documentation and if you look at the errors, it comes from the documentation for every operator written in c++ files. It does not stop the html pages to be generated. Azure devops probably interprets the errord on stderr as failures. The regular ci is set to continue even there failure. For the job generating the pages, maybe we set ask azure devops to look for string "build succeeded" and declare the task as a success if it is found. What do you think? I'll work later to improve and the rendering and fix the errors. Right now, the rendering is on par with thr markdown pages or better even with errors.

jcwchen:
Thanks for the details. It's a weird behavior that it says build succeed, but it also throws stderr... If the original API for doc generation cannot silent those warnings somehow, yes probably we need to handle that from our own for now. 

> maybe we set ask azure devops to look for string "build succeeded" and declare the task as a success if it is found. What do you think?

Sounds like a plan. Or, I was thinking whether we can directly add `always()` (see [reference](https://docs.github.com/en/actions/learn-github-actions/expressions#always)) in the section for building doc to make it always continue to run, and then further check the generated doc before deploying. The problem is it seems hard to know whether the build is successful without the output message from the make command. I think yours is more feasible.

PR #4572, opened 2022-10-05
Implement type and shape inference for ArrayFeatureExtractor

cbourjau:
### Description
This PR implements type and shape inference for the `ArrayFeatureExtractor` operator. This functionality was previously missing. 

### Motivation and Context
Having shape/type inference is better than not having it :) I did not find a related issue. Is there a tracking issue for missing or incomplete type/shape inference functions?


gramalingam:
Thanks for the contribution!

onnx/defs/traditionalml/defs.cc
```diff
@@ -19,6 +19,22 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+          const auto& input_shape = ctx.getInputType(0)->tensor_type().shape();
+          const auto input_ndim = input_shape.dim_size();
+
+          auto output_shape = ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+          // This operator only applies to the last dimension; thus -1
+          for (int i = 0; i < input_ndim - 1; ++i) {
+            *output_shape->add_dim() = input_shape.dim(i);
+          }
+          // The length of the last dimension is not statically known
+          output_shape->add_dim();
```
gramalingam, line 36: This looks correct. I think we could make it more precise if the second tensor's dimension-size is statically known, I suppose.

gramalingam: Specifically, I think we could do something like
```cpp
   if (hasInputShape(ctx, 1)) {
      const auto& indices_shape = getInputShape(ctx, 1);
      if (indices_shape.dim_size() > 0) {
         *output_shape->add_dim() = indices_shape.dim(0)
      }
   }
```

gramalingam: with the existing code for the else-cases

gramalingam: However, I notice that the op's description doesn't mention the order of axes for the output. This looks like a missing piece of information.

gramalingam: onnxruntime, however, seems to be consistent with the above ordering.

cbourjau: While missing in the specification, this appears to me to be the behavior people would expect. I added your suggestion to the shape inference. Thanks!

cbourjau:
Thank you very much for your feedback @gramalingam . I incorporated it and expanded the tests accordingly.

gramalingam (approved):
LGTM, thanks!

Issue #4548, opened 2022-09-27
Question on ScatterND's support on negative index and duplicate entries in indices


fengyuentau:
# Ask a Question

### Question

Q1: Does ScatterND support negative index? There is a clear specification for Scatter saying it should support negative index, but not for ScatterND. And there is no example with negative index for ScatterND as well.

Q2: Could anyone provide an example on duplicate entries of indices not being supported for ScatterND? There is a specification for ScatterND saying:

> Note that indices should not have duplicate entries. That is, two or more updates for the same index-location is not supported.

, but some of the ScatterND examples (scatternd_add/max/min/multiply) provide a `indices = np.array([[0], [0]], dtype=np.int64)`, which looks it has duplicate indices to me.

### Further information
- Relevant Area:  operators

- Is this issue related to a specific model?  N/A

### Notes

N/A

gramalingam:
Regarding the second question: that's a good point. The original version of ScatterND did not support duplicate indices, but it was  updated (I think may be in opset 16) to allow duplicate entries, but only if the reduction attribute is specified to indicate what to do with the duplicate entries. Please see the later part of the operator documentation on reduction. The specific line you refer to is  confusing, I agree, and is copied over from the previous version.

gramalingam:
Also, regarding the first question: IMHO, negative index values are permitted, just like in the other Scatter/Gather ops. I agree it will be good to clarify this in the spec, will update it, thanks!

PR #4575, opened 2022-10-05
Update ScatterND documentation

gramalingam:
Clarify a couple of issues in ScatterND documentation (see https://github.com/onnx/onnx/issues/4548).

* Negative index values are allowed.
* Delete outdated comment about duplicate indices.


onnx/defs/tensor/defs.cc
```diff
@@ -1247,14 +1247,14 @@ ScatterND takes three inputs `data` tensor of rank r >= 1, `indices` tensor of r
 and `updates` tensor of rank q + r - indices.shape[-1] - 1. The output of the operation
 is produced by creating a copy of the input `data`, and then updating its value to values
 specified by `updates` at specific index positions specified by `indices`. Its output shape
-is the same as the shape of `data`. Note that `indices` should not have duplicate entries.
```
wschin, line 1250: Why is it removed? It looks like a behavior change and needs a new op version.

wschin (approved):
LGTM. Thank you!

Issue #4580, opened 2022-10-10
Shapes of intermediate tensors


justinchuby:
# Ask a Question

### Question
In a onnx graph, I can see the tensor shapes for the inputs and outputs. Is there a way to know what shapes the intermediate tensors are? I consulted the shape inferencing document but it looks like there is only a c++ api.

### Further information
- Relevant Area: shape_inference

### Notes

Model exported from pytorch

```python
import torch
from torch import nn

class Upsampling(nn.Module):
    def __init__(self):
        super().__init__()
        self.upsample = nn.Upsample(size=[2,2], mode='nearest')

    def forward(self, input, weight):
        weight =  self.upsample(weight)
        return torch.nn.functional.conv2d(input, weight)

torch.onnx.export(Upsampling(), (torch.tensor([[[[1,2],[1,2]]]], dtype=torch.float32), torch.tensor([[[[1]]]], dtype=torch.float32)), "test.onnx")
```

jcwchen:
Not sure I understand this question correctly -- If you are looking for shape information of intermediate tensors after onnx.shape_inference, you could find them in `onnx_model.graph.value_info`.

justinchuby:
Thanks! That's very helpful. For some reason the pytorch exported model does not include shape information. I needed to run `onnx.infer_shape` to obtain a model with non empty `value_info`

justinchuby:
For this model

<img src="https://user-images.githubusercontent.com/11205048/193432597-4d9367ed-6719-4e96-b25a-3da31e8a23e4.png" width="200">

I am looking for the shape of `/upsample/Resize_output_0`.


I got 

```
[name: "/upsample/Shape_output_0"
type {
  tensor_type {
    elem_type: 7
    shape {
      dim {
        dim_value: 4
      }
    }
  }
}
, name: "/upsample/Constant_output_0"
type {
  tensor_type {
    elem_type: 7
    shape {
      dim {
        dim_value: 1
      }
    }
  }
}
, name: "/upsample/Constant_1_output_0"
type {
  tensor_type {
    elem_type: 7
    shape {
      dim {
        dim_value: 1
      }
    }
  }
}
, name: "/upsample/Constant_2_output_0"
type {
  tensor_type {
    elem_type: 7
    shape {
      dim {
        dim_value: 1
      }
    }
  }
}
, name: "/upsample/Slice_output_0"
type {
  tensor_type {
    elem_type: 7
    shape {
      dim {
        dim_value: 2
      }
    }
[...]

PR #4582, opened 2022-10-11
Enable partial data propagation for Resize's sizes

jcwchen:
### Description
Enable partial data propagation for Resize's sizes

### Motivation and Context
Current Resize's size cannot get symbolic shape from ONNX's data propagation. Closes https://github.com/onnx/onnx/issues/4580.


jcwchen:
cc @justinchuby I have tested this PR with your provided model and it can infer Resize's shape as [1, 1, 2, 2].

justinchuby:
Thank you!

gramalingam:
Thanks Jacky for this. I wonder, however, if we can use [getInputShape](https://github.com/onnx/onnx/blob/a525f98d5ed31660b629bab90c680a39658a5c08/onnx/defs/shape_inference.cc#L428) ... I think it was abstracted as a function for this purpose.

jcwchen:
> Thanks Jacky for this. I wonder, however, if we can use [getShapeInput](https://github.com/onnx/onnx/blob/a525f98d5ed31660b629bab90c680a39658a5c08/onnx/defs/shape_inference.cc#L428) ... I think it was abstracted as a function for this purpose.

Good catch. Yes, I should use getShapeInput instead. Updated. Please review it. Thanks!

onnx/defs/tensor/utils.cc
```diff
@@ -157,54 +165,53 @@ void resizeShapeInferenceVersioned(InferenceContext& ctx, int opset_version) {
...
+      }
+    } else {
+      // sizes_data contains scales for all axes
+      if (sizes_data.size() != rank_x) {
+        fail_shape_inference(
+            "Number of elements of input 'sizes' (",
+            sizes_data.size(),
+            ") must be same as rank of input 'X' (",
+            rank_x,
+            ").");
```
jcwchen, line 194: Actually I was confused about this check. I saw the document also mentioned this, but existing test models like `test_resize_tf_crop_and_resize_axes_2_3` and `test_resize_tf_crop_and_resize_axes_3_2` do not follow this. Take `test_resize_tf_crop_and_resize_axes_2_3` as an example, the model is (input rank 4 != sizes 2):

image

@gramalingam do you think they are invalid ONNX models?

gramalingam: Hi, I think this may be a bug introduced when the `axes` attribute was added to Resize. If the `axes` attribute is specified, we need to account for it (and the PR that added the attribute may have failed to account for it both in code and the documentation).

gramalingam: This PR: https://github.com/onnx/onnx/pull/4126

gramalingam: Good catch!

onnx/defs/tensor/utils.cc
```diff
@@ -96,20 +96,28 @@ void resizeShapeInferenceVersioned(InferenceContext& ctx, int opset_version) {
   bool hasSizesInput = ctx.hasInput(3);
 
   const TensorProto* scales = 2 < ctx.getNumInputs() ? ctx.getInputData(2) : nullptr;
-  const TensorProto* sizes = 3 < ctx.getNumInputs() ? ctx.getInputData(3) : nullptr;
+  std::vector<int64_t> sizes_data;
+  if (3 < ctx.getNumInputs()) {
+    bool found_sizes = false;
+    const auto sizes_shape = getShapeInput(ctx, 3, found_sizes);
+    // If sizes is an empty shape, assume it's not provided
```
gramalingam, line 103: Can you clarify why? Specifically: is my understanding correct that this is an error-situation? Or, is this supposed to be a valid situation? (I guess it could be valid if `axes` is empty ... except for that edge-case.)

gramalingam: Oh ... I remember, is this the old test-case fix where empty-tensor was used (before input was made optional)?

jcwchen: https://github.com/onnx/onnx/pull/4388 I think it just provides some flexibility to let ONNX handle old models.

>  I remember, is this the old test-case fix where empty-tensor was used (before input was made optional)?

Yes! Some existing models in ONNX Model Zoo have this kind of usage.

onnx/defs/tensor/utils.cc
jcwchen: Thank you for the helpful context! Sorry that I think I provided the wrong model above. Here is the right model for `test_resize_tf_crop_and_resize_axes_2_3`.

image


Based on the naming of this test model, does it mean that this model should have axis attribute?

gramalingam: So, it is difficult to see what was in the old-version vs. new-version here ... the check is fine if axes attribute wasn't specified, right? The new-version seems to have this check only if axes attribute is not specified?

jcwchen: IIUC, the only difference about this issue is the update from this PR enables "sizes" input to have the shape information even if the input tensor is unknown. Previously the tensor (sizes) will be nullptr anyway so it did not check the rank for input x and input sizes... IMO, a possible fix might be making these 2 models have attribute axes (2, 3) and axes (3, 2) and then they should pass the existing check. Does it look good to you?

gramalingam: I agree the test-cases seem to be missing the axes attribute. We should add them. I guess you are saying that the checker wasn't catching the error previously, and we are able to catch it now (because of the improvement due to data-propagation)?

jcwchen: > I guess you are saying that the checker wasn't catching the error previously, and we are able to catch it now (because of the improvement due to data-propagation)?

Yes. data-propagation will at least keep the shape information (the shape of "sizes" in this case) so it can do more check. I will update these models then. Thanks!

jcwchen: I have updated these 2 models with attribute axes. Now both of them have passed onnx.shape_inference. PTAL.

gramalingam (approved):
LGTM, thanks!

wine99:
If I understand correctly, all the nodes in this graph support data propagation. However, running

```
inferred_model = shape_inference.infer_shapes(model, strict_mode=True, data_prop=True)
```
still leaves me with four unknown dimensions. Could you help me identify the problem here? I am using the python package onnx 1.17.0

image

PR #4593, opened 2022-10-12
Use github-pr-review to enable manual resolution

jcwchen:
### Description

Use github-pr-review instead of github-pr-check to enable manual resolution.

### Motivation and Context
We cannot manually resolve a warning from reviewdog with current github-pr-check. See discussion: https://github.com/reviewdog/reviewdog/issues/568. The drawback is old warnings won't be automatically resolved with new commits for correction so you will need to resolve each of them manually. It's a trade-off.


jcwchen:
After internal discussion, close it since it is not really needed for now.

PR #4598, opened 2022-10-13
fix spelling new python doc

andife:
Signed-off-by: andife <fehlner@arcor.de>


jcwchen (approved):
LGTM. Thank you for catching these! I am surprise that reviewdog/action-misspell
https://github.com/onnx/onnx/blob/4637098e881d0d2f63caa5e945a9fdc620e11e06/.github/workflows/lint.yaml#L48 did not catch these in advance..

andife:
> LGTM. Thank you for catching these! I am surprise that reviewdog/action-misspell
> 
> https://github.com/onnx/onnx/blob/4637098e881d0d2f63caa5e945a9fdc620e11e06/.github/workflows/lint.yaml#L48
> did not catch these in advance..

Maybe we should check it with different mispelt words? ...

gramalingam (approved):

PR #4600, opened 2022-10-13
Add type inference for CategoryMapper & TreeEnsembleRegressor/Classifier

jbachurski:
### Description

Adds/extends type & shape inference for operators in `ai.onnx.ml`:
- `CategoryMapper`
- `TreeEnsembleRegressor`
- `TreeEnsembleClassifier`

### Motivation and Context

Traditional ML operators under `ai.onnx.ml` tend to be lacking in type and shape inference.

There is still a need for improving documentation for these operators and extending inference for other operators. They also generally lack sanity-checking whether they are properly specified - for example for the tree ensemble family, where there's a lot of attributes and almost no checking is done.

As a general note, adding new inference/verify functions is a bit of a pain. There's a lot of somewhat generic patterns (attributes of equal length, only one of attribute group is set, make dimensions equal to given shape list) that seem to be done really inefficiently, involving a lot of mutation. 

Adding inference functions is currently not an open issue but there probably should be one opened. Existence of operators like these causes total failure of the shape inference routines (as all succeeding operators fail their inference).

Related to #4572, which makes a similar improvement by implementing another inference function.

### DIFF
--- a/onnx/defs/traditionalml/defs.cc
+++ b/onnx/defs/traditionalml/defs.cc
@@ -2,8 +2,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include "onnx/defs/schema.h"
 #include <stdio.h>
+#include "onnx/defs/schema.h"
 
 #ifdef ONNX_ML
 namespace ONNX_NAMESPACE {
@@ -854,7 +854,7 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
           // Second axis of Z - E (classes)
           std::vector<int64_t> class_ids;
           auto has_ids = getRepeatedAttribute(ctx, "class_ids", class_ids);
-          if(has_ids) {
+          if (has_ids) {
             snd_dim_z->set_dim_value(class_ids.size());
           }
         }));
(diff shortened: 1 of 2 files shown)


onnx/defs/traditionalml/defs.cc
```diff
@@ -831,6 +836,26 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
             fail_shape_inference(
                 "Only one of the attributes 'base_values', 'base_values_as_tensor' should be specified.");
           }
+
+          // First axis of Y & Z - N (inputs)
```
gramalingam, line 840: IMO, the code can be rewritten as:
```cc
checkInputRank(ctx, 0, 2);
Dim N, E;
unifyInputDim(ctx, 0, 0, N);
std::vector<int64_t> class_ids;
auto has_ids = getRepeatedAttribute(ctx, "class_ids", class_ids);
if (has_ids) {
   unifyDim(E, class_ids.size()); // or, just: E.set_dim_value(class_ids.size());
}
updateOutputShape(ctx, 0, {N});
updateOutputShape(ctx, 1, {N, E});
```

jbachurski: Great suggestion! I wasn't aware of these higher-level functions. I applied it & refactored in the cases of the tree ensemble operators.

gramalingam (approved):

PR #4607, opened 2022-10-19
Fix LayerNormalization equation

longlee0622:
### Description

Variable `Diff` is not being defined. It should be `D`


gramalingam:
Hi, Thanks for finding the typo, and the PR. Unfortunately, operators.md is generated documentation. So, you will need to fix the source code [here](https://github.com/onnx/onnx/blob/1010de3a3bfd77d753860f9981f07cb3424ca202/onnx/defs/nn/defs.cc#L2406) and generate the documentation (after installing the changes) by running `python onnx\defs\gen_doc.py` ... I realize that's a lot for a single line fix :(

gramalingam:
As the CI indicates, there are other places that need to be fixed as well: `docs/Changelog.md`. (Running the document-generator generates all these automatically.)

gramalingam (approved):

PR #4612, opened 2022-10-21
Update maxpool doc

gramalingam:
Clarify the documentation of MaxPool operator to explain why the storage-order attribute is used. (See https://github.com/onnx/onnx/issues/2290 ).


wschin (approved):
:shipit:

PR #4621, opened 2022-10-25
add GroupNormalization-18

yuanyao-nv:

### Description
Add new function op `GroupNormalization` to opset 18.

### Motivation and Context
Group norm is a commonly used normalization op, similar to some of the existing ONNX ops such as `BatchNormalization`, `InstanceNormalization`, and `LayerNormalization`.

Original paper that introduced the op: https://arxiv.org/abs/1803.08494

Related issue: https://github.com/onnx/onnx/issues/2020

Implementations in other frameworks:
https://www.tensorflow.org/addons/api_docs/python/tfa/layers/GroupNormalization
https://pytorch.org/docs/stable/generated/torch.nn.GroupNorm.html


onnx/defs/nn/defs.cc
```diff
@@ -2595,4 +2595,151 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                  .Add("GroupSize = Div (C, NumGroups)")
+                  .Const1D("Zero1D", int64_t(0))
+                  .Const1D("IntMax1D", std::numeric_limits<int64_t>::max())
+                  .Add("N = Slice (XShape, Zero1D, One1D)") // batch size
+                  .Add("HW = Slice (XShape, Two1D, IntMax1D)") // data instance shape
+
+                  // NewShape = [N, num_groups, group_size, H, W]
+                  .Add("NewShape = Concat <axis = 0> (N, NumGroups, GroupSize, HW)")
+                  .Add("XReshaped = Reshape (X, NewShape)")
+                  .Add("RankPlusOne = Add (Rank, One1D)")
```
gramalingam, line 2716: Is this `RankPlusOne` used? I can't see it, may be I missed it.

onnx/defs/nn/defs.cc
```diff
@@ -2595,4 +2595,151 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                return tp;
+              };
+              FunctionBuilder builder(functionProto);
+              builder.Const1D("FloatEpsilon", epsilon)
+                  .Add("Epsilon = Cast (FloatEpsilon)", "to", T)
+                  .Add("XShape = Shape (X)") // shape of input tensor: 1D tensor
+                  .Add("Rank = Size (XShape)") // rank of input tensor: scalar
+                  .Const1D("One1D", int64_t(1))
+                  .Const1D("Two1D", int64_t(2))
+                  .Add("C = Slice (XShape, One1D, Two1D)") // number of channels
```
gramalingam, line 2705: We could replace these multiple statements with just `C = Shape <start = 1, end = 2> (X)`.

onnx/defs/nn/defs.cc
```diff
@@ -2595,4 +2595,151 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                  .Const1D("NumGroups", num_groups)
+                  .Add("GroupSize = Div (C, NumGroups)")
+                  .Const1D("Zero1D", int64_t(0))
+                  .Const1D("IntMax1D", std::numeric_limits<int64_t>::max())
+                  .Add("N = Slice (XShape, Zero1D, One1D)") // batch size
```
gramalingam, line 2710: Similarly, `N = Shape <start=0, end=1> (X)` should work here.

onnx/defs/nn/defs.cc
```diff
@@ -2595,4 +2595,151 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                  .Add("HW = Slice (XShape, Two1D, IntMax1D)") // data instance shape
```
gramalingam, line 2711: And `HW = Shape <start=2>(X)` should work here. Also renaming `HW` to `Rest` will probably more accurately describe the non-image cases.

onnx/defs/nn/defs.cc
```diff
@@ -2595,4 +2595,151 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+
+              auto* epsilon_attr = ctx.getAttribute("epsilon");
+              float epsilon = (epsilon_attr != nullptr) ? epsilon_attr->f() : 1e-5f;
+              auto* num_groups_attr = ctx.getAttribute("num_groups");
+              if (num_groups_attr == nullptr)
+                return false;
+              int64_t num_groups = num_groups_attr->i();
+
+              auto& input_shape = ctx.getInputType(0)->tensor_type().shape();
+              int64_t input_ndim = input_shape.dim_size();
```
gramalingam, line 2687: Ok, this could be a problem. For better or worse, ONNX tries to allow dynamic shapes and tensors of statically unknown rank (eg., arising out of imprecision in shape-inference). We can't assume that the rank is known above.

I think the function body would be simplified if we flatten the input tensor into a 3D tensor first, eg. via
```
   Shape3D = Constant <value_ints = [0, 0, -1]>()
   X3D = Reshape (X, Shape3D)
```

gramalingam: Because the function doesn't really care about the distinctions between 3rd dimension onwards.

onnx/backend/test/case/node/groupnorm.py
```diff
@@ -0,0 +1,64 @@
...
+import onnx
+
+from ..base import Base
+from . import expect
+
+
+class GroupNormalization(Base):
+    @staticmethod
+    def export() -> None:
+        def _groupnorm_test_mode(x, num_groups, scale, bias, epsilon=1e-5):
```
gramalingam, line 14: Nit: Suggest making this a top-level function (move it outside the class), as a "reference implementation". Also, suggest dropping the suffix `_test_mode` (feels like a copy-paste artifact not relevant to this op, since it doesn't have a separate test-mode).

onnx/backend/test/case/node/groupnorm.py
```diff
@@ -0,0 +1,64 @@
...
+            dim_ones = (1,) * (len(new_shape) - 2)
+            scale = scale.reshape(-1, *dim_ones)
+            bias = bias.reshape(-1, *dim_ones)
+            res = scale * (x_reshaped - mean) / np.sqrt(var + epsilon) + bias
+            return res.reshape(x.shape)
+
+        x = np.random.randn(3, 4, 2, 2).astype(np.float32)
+        num_groups = 2
+        scale = np.random.randn(num_groups)
+        bias = np.random.randn(num_groups)
```
gramalingam, line 33: May be add `.astype(np.float32)` to bias and scale as well? Not sure if that is the default, but I notice x has it above.

gramalingam:
LGTM. I think there may be a minor interaction between this PR and https://github.com/onnx/onnx/pull/4512 (which is rationalizing all Reduce* functions to take axes as an input instead of attribute) which will need to be resolved when both get merged.

yuanyao-nv:
@gramalingam Thanks for the suggestions. I've made the updates. Please review again. Thanks!

gramalingam:
LGTM, thanks! I was wondering whether you had a chance to test the correctness of the function-body definition (eg., by running onnxruntime, or any other backend, on the generated expanded test-cases)?

yuanyao-nv:
> LGTM, thanks! I was wondering whether you had a chance to test the correctness of the function-body definition (eg., by running onnxruntime, or any other backend, on the generated expanded test-cases)? You may need to manually patch the opset of the generated test-cases to be the previous opset, and then it should be a valid test-case for the previous opset, I think, runnable by onnxruntime etc.

Thanks for pointing this out. I just tested and actually found a small error in the function body. Now fixed.

gramalingam (approved):

PR #4623, opened 2022-10-26
Rename onnx_docs_folder to operator in docs

natke:


.github/workflows/pages.yml
```diff
@@ -32,39 +31,22 @@ jobs:
...
-      - name: Check whether previous build docs succeeded
-        run: |
-          cat docs/docsgen/stdout_stderror.log
-          if grep -F "build succeeded" docs/docsgen/stdout_stderror.log
-          then
-            echo "Build succeeded."
-          else
-            echo "Build failed."
-            exit 1
-          fi
```
jcwchen, line 66: More context about why manually check the error log: https://github.com/onnx/onnx/pull/4570#issuecomment-1268809300. To confirm: have those build errors resolved by this PR? If not, probably we still need to keep this code.

natke: It only fails when you use `make -j auto`, which I removed.

When I run the original workflow the docs were not generated but the build still passed.

Hence to make this workflow more reliable, I removed the hack and now the build takes a little longer (~5mins). I considered this a worthwhile tradeoff

jcwchen (approved):
LGTM. Thank you for the improvement! Please signoff DCO: https://github.com/onnx/onnx/blob/main/CONTRIBUTING.md#dco.

natke:
> LGTM. Thank you for the improvement! Please signoff DCO: https://github.com/onnx/onnx/blob/main/CONTRIBUTING.md#dco.

Done

PR #4629, opened 2022-10-31
Upgrade to newer Protobuf 3.20.2

jcwchen:
### Description
Upgrade to newer Protobuf 3.20.2.

### Motivation and Context
- 3.18.3 is out of maintenance windows
- https://github.com/protocolbuffers/protobuf/issues/10691 Python Protobuf Mac wheel has unknown issue
- ONNX Runtime will upgrade its Protobuf to 3.20+ soon: https://github.com/onnx/onnx/pull/4535#issuecomment-1296014764. Making ONNX and ONNX Runtime use same Protobuf version can prevent potential conflict issue.

Disadvantage: If ONNX decides to upgrade to newer Protobuf 3.20.2, the minimum required Python Protobuf version will be upgraded to 3.20.2 as well (from 3.12.2).


jcwchen:
After internal discussion today, we decided to forward this upgrade before next ONNX 1.13 release. @snnn may I understand when (exact time) will you upgrade ONNX Runtime's Protobuf?

snnn:
It can be at any time.

snnn:
@djee-ms suggests using 3.21.x. See #13335 .

jcwchen:
I strongly recommend not upgrading to Protobuf 3.21+ at this moment, because there are breaking changes between 3.21 and 3.20: https://github.com/protocolbuffers/protobuf/issues/10051. Currently PyTorch and TensorFlow are still using older Protobuf version (3.13.1 and 3.11.4) and I don't think they will upgrade them in the near future. Therefore, it should take more time to let the whole ONNX ecosystem start to use Protobuf 3.21+. At least for this release, I think upgrading Protobuf from 3.16.0 to 3.20.2 is aggressive enough.

gramalingam (approved):

PR #4632, opened 2022-11-02
Consider GraphInferenceContext in inference functions: InferenceContext

jcwchen:
### Description

- Consider GraphInferenceContext in inference functions (InferenceContextImpl) to cover more use cases.
- Add a test to test node with subgraph.
- TODO in another PR: expose functionProto as well.

### Motivation and Context

Follow up for https://github.com/onnx/onnx/pull/4409. Further consider GraphInferenceContext for inference function


onnx/cpp2py_export.cc
```diff
@@ -58,7 +58,8 @@ std::unordered_map<std::string, py::bytes> CallNodeInferenceFunction(
     const py::bytes& nodeBytes,
     std::unordered_map<std::string, py::bytes> valueTypesByNameBytes,
     std::unordered_map<std::string, py::bytes> inputDataByNameBytes,
-    std::unordered_map<std::string, py::bytes> inputSparseDataByNameBytes) {
+    std::unordered_map<std::string, py::bytes> inputSparseDataByNameBytes,
+    std::unordered_map<std::string, py::bytes> outerScopeValueTypesByNameBytes) {
```
gramalingam, line 62: Do we need this extra-parameter? I think the parameter `valueTypesByName` can serve both purposes. I realize that the underlying APIs distinguish them, but it seems simpler to manage with a single map for the python API, unless there is a good reason to make the distinction. I think there is no need since the same set of names are visible and can be used as node-inputs or inside graph-attributes.

jcwchen: Thanks for pointing out! I was also confused here, because I thought there might be some use cases for different maps. I have updated it with single map. PTAL.

onnx/cpp2py_export.cc
```diff
@@ -69,8 +69,13 @@ std::unordered_map<std::string, py::bytes> CallNodeInferenceFunction(
   const auto& inputData = ParseProtoFromBytesMap<const TensorProto>(inputDataByNameBytes);
   const auto& inputSparseData = ParseProtoFromBytesMap<const SparseTensorProto>(inputSparseDataByNameBytes);
 
+  std::unordered_map<std::string, int> opset_imports;
+  opset_imports[schema->domain()] = schema->SinceVersion();
+  shape_inference::GraphInferenceContext graphInferenceContext(valueTypes.second, opset_imports);
```
jcwchen, line 74:
```
  GraphInferenceContext(
      const std::unordered_map<std::string, TypeProto*>& outer_scope_value_types_by_name_in,
      const std::unordered_map<std::string, int> opset_imports_in,
      SymbolTable* symbol_table_in = nullptr,
      const ModelLocalFunctionsMap& model_local_functions_in = {},
      const ISchemaRegistry* schema_registry_in = OpSchemaRegistry::Instance(),
      std::unordered_map<std::string, TensorShapeProto>* generated_shape_data_by_name_in = nullptr,
      const int ir_version_in = IR_VERSION)
```
For now I only passed `outer_scope_value_types_by_name` and `opset_imports`. Not sure whether they are sufficient though. Feel free to let me know if you think other parameters are useful.

onnx/cpp2py_export.cc
```diff
@@ -58,8 +58,7 @@ std::unordered_map<std::string, py::bytes> CallNodeInferenceFunction(
     const py::bytes& nodeBytes,
     std::unordered_map<std::string, py::bytes> valueTypesByNameBytes,
     std::unordered_map<std::string, py::bytes> inputDataByNameBytes,
-    std::unordered_map<std::string, py::bytes> inputSparseDataByNameBytes,
-    std::unordered_map<std::string, py::bytes> outerScopeValueTypesByNameBytes) {
+    std::unordered_map<std::string, py::bytes> inputSparseDataByNameBytes) {
```
gramalingam, line 61: Should add `opset_imports` as another input

gramalingam:
Adding a test-case would be useful

onnx/shape_inference.py
```diff
@@ -88,16 +88,13 @@ def infer_node_outputs(
...
         return {}
     if input_data is None:
         input_data = {}
     if input_sparse_data is None:
         input_sparse_data = {}
-    if outer_scope_value_types is None:
-        outer_scope_value_types = {}
 
     # To avoid copying on C++ side, pass only what is needed for this inference call
     passed_input_types = {
```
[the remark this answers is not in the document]
gramalingam: Currently, the filtering in the following two lines is too aggressive (since we are using this for outer-scope values as well). So, the filtering in the lines below has to be removed.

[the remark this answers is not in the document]
gramalingam: So, I suggest `passed_input_types = input_types`

onnx/test/inference_function_test.py
```diff
@@ -145,6 +152,47 @@ def test_reshape_inference(self) -> None:
...
+
+        assert onnx.shape_inference.infer_node_outputs(
+            onnx.defs.get_schema("Scan", 9),
+            make_node(
+                "Scan",
+                ["loop_state_orig", "scan_input"],
+                ["loop_state_final", "scan_output"],
+                num_scan_inputs=1,
+                body=subgraph,
+            ),
```
jcwchen, line 187: FYI: I didn't use parser for creating NodeProto here, because it seems that parse_node is missing.

onnx/test/inference_function_test.py
```diff
@@ -145,6 +152,54 @@ def test_reshape_inference(self) -> None:
...
+        ]
+        output_value_infos = [
+            make_tensor_value_info("loop_state_out", TensorProto.UNDEFINED, None),
+            make_tensor_value_info("output", TensorProto.UNDEFINED, None),
+        ]
+
+        subgraph = make_graph(
+            [
+                make_node("Identity", ["loop_state_in"], ["loop_state_out"]),
+                make_node("Identity", ["input"], ["output"]),
```
gramalingam, line 172: Can we change this to include an outer-scope variable reference? Like
```py
   make_node("Add", ["input", "outer"], ["output"]),
```

onnx/test/inference_function_test.py
```diff
@@ -145,6 +152,54 @@ def test_reshape_inference(self) -> None:
...
+                "Scan",
+                ["loop_state_orig", "scan_input"],
+                ["loop_state_final", "scan_output"],
+                num_scan_inputs=1,
+                body=subgraph,
+            ),
+            _to_tensor_types(
+                {
+                    "loop_state_orig": (TensorProto.FLOAT, (loop_state_size,)),
+                    "scan_input": (TensorProto.FLOAT, (seq_len, input_size)),
```
gramalingam, line 191: And add the type for `outer` here

jcwchen: Added outer with an Add node. PTAL. Thanks!

onnx/shape_inference.py
```diff
@@ -88,18 +88,30 @@ def infer_node_outputs(
     input_types: Dict[str, onnx.TypeProto],
     input_data: Optional[Dict[str, onnx.TensorProto]] = None,
     input_sparse_data: Optional[Dict[str, onnx.SparseTensorProto]] = None,
+    subgraph_opset_imports: Optional[List[onnx.OperatorSetIdProto]] = None,
```
gramalingam, line 91: minor nit: I think `opset_imports` and `ir_version` might be sufficient as the parameter names. I realize they are used only for subgraphs, but it is unlikely to cause any confusion or ambiguity to use the shorter name. It will be easier for callers, when they use keyword-arguments.

gramalingam (approved):
LGTM, thanks! Just have one minor comment about naming of parameters.

onnx/shape_inference.py
jcwchen: I was in the debate of `opset_imports` and `subgraph_opset_imports`, but yes I agreed using shorter name here is cleaner. Just updated. Will merge this PR soon. Thank you for bringing this up!

gramalingam (approved):

PR #4638, opened 2022-11-07
Replace hard-coded constants with Protobuf enums

paxbun:
### Description
Changed hard-coded type constants in `PrimitiveTypeNameMap` or `AttributeTypeNameMap` into enums generated by Protobuf, (e.g. `1` -> `TensorProto_DataType_FLOAT`)

### Motivation and Context
I was experimenting with ONNX, and found that changing enum values defined in `.in.proto` file makes a runtime error like the following:
```
  ...
  File ".../tf2onnx/__init__.py", line 11, in <module>
    from tf2onnx import tfonnx, utils, graph, graph_builder, graph_matcher, shape_inference, schemas, convert  # pylint: disable=wrong-import-order
  File ".../tf2onnx/tfonnx.py", line 19, in <module>
    from tf2onnx.graph import Graph
  File ".../tf2onnx/graph.py", line 18, in <module>
    from tf2onnx.schemas import get_schema, infer_onnx_shape_dtype
  File ".../tf2onnx/schemas.py", line 96, in <module>
    _schemas = _register_all_schemas_with_history()
  File ".../tf2onnx/schemas.py", line 65, in _register_all_schemas_with_history
    onnx_schemas = defs.get_all_schemas_with_history()
RuntimeError: Error parsing function body:[ParseError at position (line: 3 column: 44)]
Error context:            Q_Min = Constant<value = float {0.0}>()
```
and I found that this is because of the mismatch between the C++ code and the Protobuf code that occurred by changing the `*.in.proto` files. This PR just replaces the hard-coded type constants with the Protobuf-generated enums.


onnx/defs/parser.h
```diff
@@ -72,22 +72,22 @@ class StringIntMap {
...
+    map_["int64"] = TensorProto_DataType_INT64;
+    map_["string"] = TensorProto_DataType_STRING;
+    map_["bool"] = TensorProto_DataType_BOOL;
+    map_["float16"] = TensorProto_DataType_FLOAT16;
+    map_["double"] = TensorProto_DataType_DOUBLE;
+    map_["uint32"] = TensorProto_DataType_UINT32;
+    map_["uint64"] = TensorProto_DataType_UINT64;
+    map_["complex64"] = TensorProto_DataType_COMPLEX64;
+    map_["complex128"] = TensorProto_DataType_COMPLEX128;
+    map_["bfloat16"] = TensorProto_DataType_BFLOAT16;
```
daquexian, line 90: Is it possible to reuse [TypeWrapper](https://github.com/onnx/onnx/blob/main/onnx/defs/data_type_utils.cc#L413) here to avoid duplication? cc @gramalingam

daquexian (commented):
Thanks for your contribution!

onnx/defs/parser.h
paxbun: Seems like `PrimitiveTypeNameMap` has the same feature as `DataTypeUtils` does. I think it would be better to remove `PrimitiveTypeNameMap` and `AttributeTypeNameMap`, merge `PrimitiveTypeNameMap` to `DataTypeUtils`, and create a similar class named `AttributeTypeUtils` that succeeds `AttributeTypeNameMap`.

paxbun: I'm working on other projects now, I will make a fix for this ASAP.

gramalingam: Thanks for the fix. LGTM. Please note that code-freeze for the next release is in a few days. Please let us know whether we should merge this PR as-is (and do the extension discussed above as a separate PR) or wait for the extension. Thanks!

gramalingam (approved):

PR #4640, opened 2022-11-08
Add Map ops to opset18

shubhambhokare1:
### Description
Adding operators for map type to opset 18:
- MapConstruct
- MapKeys
- MapValues
- MapInsertPair
- MapDeletePair
- MapHasKey
- MapHasValue

### Motivation and Context
These ops are necessary for enabling export of pytorch models with map type.


onnx/defs/map/defs.cc
```diff
@@ -0,0 +1,402 @@
...
+
+          const auto& input_shape = ctx.getInputType(0)->map_type().value_type().sequence_type().elem_type().tensor_type().shape();
+          auto final_output_shape = ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
+          final_output_shape->add_dim()->set_dim_value(input_shape.dim_size());
+        }));
+
+
+static const char* MapValues_ver18_doc = R"DOC(
+Outputs a tensor that consists of all the 'values' present
+in the map structure.
```
gramalingam, line 162: Is the output a Tensor? Or a Sequence (as described in line 171 below)?

onnx/defs/map/defs.cc
```diff
@@ -0,0 +1,402 @@
...
+            OpSchema::all_tensor_sequence_types(),
+            "Constrain output to a sequence of any tensor type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          const auto input_type = ctx.getInputType(0);
+          if (nullptr == input_type) {
+            fail_type_inference("Input map is expected to have type info. Current type is null.");
+          }
+          ctx.getOutputType(0)->CopyFrom(input_type->map_type().value_type());
+
+          const size_t numOutputs = ctx.getNumOutputs();
```
gramalingam, line 187: These checks are unnecessary, and will be done automatically (given the information in the schema).

onnx/defs/map/defs.cc
```diff
@@ -0,0 +1,402 @@
...
+        .TypeConstraint(
+            "S",
+            OpSchema::all_tensor_sequence_types(),
+            "Constrain output to a sequence of any tensor type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          const auto input_type = ctx.getInputType(0);
+          if (nullptr == input_type) {
+            fail_type_inference("Input map is expected to have type info. Current type is null.");
+          }
+          ctx.getOutputType(0)->CopyFrom(input_type->map_type().value_type());
```
gramalingam, line 185: Doesn't look right. Is it supposed be represented as a sequence?

onnx/defs/map/defs.cc
```diff
@@ -0,0 +1,402 @@
...
+    }
+    return all_map_types;
+}
+
+static const char* MapConstruct_ver18_doc = R"DOC(
+Constructs either an empty map of a certain type specified by the key_type and value_type attributes,
+or a map structure with specified 'keys' and 'values'
+'keys' and 'values' must have the same number of elements.
+All 'keys' must have the same data type.
+All 'values' must all be of the same type (tensor, sequence or map) and have the same data type.
```
gramalingam, line 30: It's a bit unclear what's supported as values (seeing description below as well).

onnx/defs/map/defs.cc
```diff
@@ -0,0 +1,402 @@
...
+        .Input(0, "keys", "Keys.", "T", OpSchema::Optional)
+        .Input(1, "values", "Sequence enclosing the values.", "S", OpSchema::Optional)
+        .Output(0, "map", "Map.", "M")
+        .TypeConstraint(
+            "T",
+            OpSchema::all_map_key_tensor_types(),
+            "Constrain input keys to integral and string tensor types.")
+        .TypeConstraint(
+            "S",
+            OpSchema::all_tensor_sequence_types(),
```
gramalingam, line 57: So, the map's value type is Tensor?

onnx/defs/map/defs.cc
```diff
@@ -0,0 +1,402 @@
...
+            "Constrain input values to a sequence of any tensor type.")
+        .TypeConstraint(
+            "M",
+            all_map_types(),
+            "Constrain output to a map type.")
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+
+          const size_t numInputs = ctx.getNumInputs();
+          const auto* key_proto = ctx.getAttribute("key_type");
+          const auto* value_proto = ctx.getAttribute("value_type");
```
gramalingam, line 67: It's safer to use something like `bool key_input_present = ctx.hasInput(0);` for checks below.

abock:
@shubhambhokare1 can you point to some of the PyTorch models that require this map support?

yuslepukhin:
Please, [see ](https://github.com/onnx/onnx/issues/5072)this as well.

rosaliaschneider:
Supporting Dict would be super useful. I am trying to use torch.onnx.export on https://github.com/NVIDIA/DeepLearningExamples/tree/master/DGLPyTorch/DrugDiscovery/SE3Transformer and hitting issues with the Dictionaries

thiagocrepaldi:
@gramalingam @shubhambhokare1 should we complete or close this PR?

gramalingam:
I would say close it. Is there any use-case for it in the pytorch exporter? These maps (like sequences) are just convenient intermediate structures, which should eventually disappear when we get to backends/runtimes (which are unlikely to support such data-structures).

shubhambhokare1:
Closing for now, as no longer needed.

AnnaTrainingG:
栓Q

Issue #4436, opened 2022-08-10
Add Apple Silicon support (M1/M2 machines): provide universal2 wheel


jcwchen:
# Feature Request

### System information
ONNX version (you are using): the latest main branch

### What is the problem that this feature solves?
Currently there is no prebuilt wheel for Apple Silicon machines like M1/M2 on PyPI. Users need to build ONNX from source on their own.

### Describe the alternatives you have considered
Directly provide xxx_macosx_11_0_arm64.whl instead, but currently GitHub Action has not provided Apple Silicon machines yet. (All of ONNX released wheels are using GitHub Action).

### Describe the feature
To Add Apple Silicon support (M1/M2 machines): provide universal2 wheel. Universal2 wheels are fat binaries with support for both 64-bit Intel and Apple processors. Although GitHub Action currently lacks Apple Silicon machines, ONNX can provide universal wheel for now, because it is at least testable in CIs. In the future, ONNX will further provide direct support like `xxx_macosx_11_0_arm64.whl` after GitHub Action has provided Apple Silicon machines. More related discussion: https://github.com/onnx/onnx/issues/4317#issuecomment-1194376761.

### Will this influence the current api?
No

### Feature Area
Which area in ONNX does this impact?  (*e.g. model usage, backend, best practices, converters, shape_inference, version_converter, training, test, operators*): 

cc @daquexian

daquexian:
BTW, [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) is a great tool to generate Python packages with c++ library because it handles different python versions and architectures and is easy to use. I have successfully used it to package onnxsim and onnx optimizer. Can we also use it for onnx?

jcwchen:
Hi @daquexian,
Good question. In the beginning of developing release CIs in ONNX, I was also considering to use cibuildwheel or other third-party tools. I thought ideally it's safer that we do not rely on other tools too much (in case they don't update someday) so I developed the release pipelines with dockers from scratch. 

Still, if it is useful for adding new support and saves a lot of engineering time, then yes I think it's good to have it. cibuildwheel looks promising and well-maintained. Feel free to apply it and simplify our current release CIs. Please let me know if you need any help. Thank you!

glenn-jocher:
@daquexian this would be a big help for YOLOv5!

joeyballentine:
Any update on this? Is m1 support planned?

jcwchen:
[...]

PR #4642, opened 2022-11-09
provide macos universal2 wheel

daquexian:
### Description
Fixes https://github.com/onnx/onnx/issues/4436

It is easier than I thought to build macos universal2 binaries. A proper `CMAKE_OSX_ARCHITECTURES` is all we need 😂

I have tested the universal2 wheel on my m1 machine.


jcwchen:
Thank you for the contribution! Please allow me to reopen this PR to trigger release CIs

.github/workflows/release_mac.yml
```diff
@@ -52,14 +53,19 @@ jobs:
...
         source workflow_scripts/protobuf/build_protobuf_unix.sh $NUM_CORES $(pwd)/protobuf/protobuf_install
         export CMAKE_ARGS="-DONNX_USE_LITE_PROTO=ON"
         # Currently GitHub Action agent is using macos-11, use -p to force change the final MACOSX_DEPLOYMENT_TARGET
         # Change -p if MACOSX_DEPLOYMENT_TARGET is different
         if [ '${{ github.event_name }}' == 'schedule' ]; then
-          python setup.py bdist_wheel -p macosx_10_12_x86_64 --weekly_build
+          python setup.py bdist_wheel -p macosx_10_12_${{ matrix.target-architecture }} --weekly_build
         else
-          python setup.py bdist_wheel -p macosx_10_12_x86_64
+          python setup.py bdist_wheel -p macosx_10_12_${{ matrix.target-architecture }}
```
jcwchen, line 68: nit: not sure whether we want to use `macosx_10_12_` for universal2 as well. Probably we can assume only 10.15+ users will try universal2 wheel.

jcwchen (approved):
Overall looks good to me. I also tested this universal2 wheel on my Mac (non M1/M2) and it worked well.

.github/workflows/release_mac.yml
daquexian: Thanks for your review! It is what I ignored.

What's the reason for choosing 10.15? The minimum macos version that can run on apple silicon chips seems 11. Maybe we can use `macosx_11_0_`.

jcwchen: Good point. `macosx_11_0_` looks better. Thanks!

daquexian: Updated. Waiting for CI :)

daquexian: @jcwchen I found that python3.7 is incompatible with `macosx_11_0_`.

```bash
$ python3.7 -m pip install onnx-1.12.0-cp37-cp37m-macosx_11_0_x86_64.whl
ERROR: onnx-1.12.0-cp37-cp37m-macosx_11_0_x86_64.whl is not a supported wheel on this platform.
$ python3.7 -m pip install onnx-1.12.0-cp37-cp37m-macosx_10_15_x86_64.whl
Processing ...
$ python3.8 -m pip install onnx-1.12.0-cp38-cp38-macosx_11_0_x86_64.whl
Processing ...
```

So I prefer keeping the versions of x86_64 and universal2 wheels the same. It is also what [numpy](https://pypi.org/project/numpy/1.21.0/#files) does:

image

Issue #3958, opened 2022-01-21
[proposal] ONNX Model metadata for provenance


rgesteve:
# Feature Request: ONNX Model metadata for provenance (proposal)

### System information
(future) ONNX version

### What is the problem that this feature solves?
Please detail the discrepancy with our current functionality.

### Describe the alternatives you have considered
Raw text, tribuo provenance support (described in notes)

### Describe the feature
Enrich and structure metadata (which is to be machine-readable) to aid model provenance tracking, discoverability of models on hubs, annotation of fairness/Ethical AI considerations and other concerns (in an extensible manner).  

### Will this influence the current api?
No

### Feature Area
model usage, backend, best practices, converters

### Are you willing to contribute it (Y/N):
Y

#### Goals

* Model details, Domain, Datasets, Applicable areas, Fairness, Transparency, Human centered approach, Trusted, Secure, Privacy protections, ..
* Facilitate metadata checks to be implemented in runtime(s) as part of Preprocessing element of ONNX Graph
* Leverage Semantic Web infrastructure (RDF) for defining the structured metadata to keep it extensible, machine readable, Embeddable (compatible with existing ONNX metadata fields). Rely on industry wide initiatives for introducing Controlled vocabularies for Provenance, Explainable and Ethical ML are already being developed (eg [IEEE P7003](https://standards.ieee.org/project/7003.html))
* Enables queries to be performed on model zoo to filter for relevant use cases

Simplify the user interactions with multiple variants hosted in ONNX model zoo and allow the ONNX community to simplify sharing from various industry contributors.

#### ONNX Model provenance metadata

Expand on the fields implemented for model hub implementation. The metadata fields hold optional information not required for the basic level of functionality

This should be considered as an initial proposal for metadata fields to capture model provenance &amp; mixed precision representation. We encourage a broader discussion around what metadata should be included in the official model zoo when they get published.

#### Example Metadata for model provenance and mixed precision

Initial proposal (noncomprehensive) list of metadata attached &amp; associated with each model that gets published in the zoo:

| **Category** | **Name** | **Type** | **Note** |
| --- | --- | --- | --- |
| Model details | Name | String |
|| --- | --- | --- | --- |
| Domain | String | Vision, Text, NLP, Graph NN etc.. |
| Objective | String | Classification, Regression, Forecast, Cluster, Security etc… |
| Precisions | Double | FP32, FP16, Int8, FP32+Int8, FP16 + Int8, etc… |
[...]

PR #4644, opened 2022-11-09
Machine-readable metadata workflow

rgesteve:
### Description
Writeup for "Machine-readable metadata" proposal (originally described in issue #3958, and discussed and presented for feedback in several fora since).

### Motivation and Context
As the number of available pre-trained onnx models increases, finding an appropriate one for a given application becomes a more pressing question.   Similarly, the provenance,  original purpose and other characteristics are becoming increasingly relevant to deploy a model.   We propose a workflow that builds on the metadata facilities already present in the standard to address the above concerns.


andife:
Thank you for this proposal and the effort put into it, @rgesteve!

After reviewing this PR in 2026, we're closing it as the approach has become outdated. Since this was filed in 2022, the ML ecosystem has converged on simpler, YAML-based model card formats (most notably the [Hugging Face model card metadata spec](https://huggingface.co/docs/hub/model-cards-guide)) rather than RDF/SPARQL-based solutions. The referenced FAIRNet ontology and IEEE P7003 have not gained broad adoption as tooling in practice.

The underlying need — structured, machine-readable model provenance metadata — remains valid. If there is renewed interest in a formal ONNX metadata workflow, we'd encourage a fresh proposal aligned with current ecosystem conventions, brought to the relevant SIG for discussion.

PR #4648, opened 2022-11-11
Fix _parse_repo_info

mszhanyi:
### Motivation and Context

For example, if the repo info is `jcwchen/models:jcw/vnni-result`,
the return value of current function will be ` jcwchen, models, jcw`
So, the _get_base_url would be incorrect.

In fact, the` jcwchen, models, jcw/vnni `is the correct return value


jcwchen (approved):
My branch name in ONNX Model Zoo causes this. Thanks for the improvement!

jcwchen (commented):
Please signoff DCO. Thanks!

PR #4655, opened 2022-11-14
Correct types in GridSample def.cc

liqunfu:
### Description
type specified for GridSample is incorrect

### Motivation and Context
fix the error


p-wysocki:
Hi! Should I add this PR to ONNX 1.13 milestone? The CF is on 18th November.

onnx/defs/tensor/defs.cc
```diff
@@ -2395,25 +2399,26 @@ ONNX_OPERATOR_SET_SCHEMA(
...
             true,
             1,
             OpSchema::Differentiable)
-        .TypeConstraint("T1", OpSchema::all_tensor_types(), "Constrain input types to all tensor types.")
+        .TypeConstraint("T1", OpSchema::all_numeric_types(), "Constrain input to only numeric types.")
         .TypeConstraint(
             "T2",
             {"tensor(float16)", "tensor(float)", "tensor(double)"},
-            "Constrain output types to float tensors.")
+            "Constrain grid and output types to float tensors.")
```
gramalingam, line 2421: Only "grid"? Not "output"?

docs/Operators.md
```diff
@@ -9325,6 +9325,10 @@ Other versions of this operator: <a href="Changelog.md#GreaterOrEqual-12">12</a>
   Given an `input` and a flow-field `grid`, computes the `output` using `input` values and pixel locations from `grid`.
   Currently, only spatial (4-D) inputs are supported. For `input` with shape (N, C, H, W) and `grid` with shape (N, H_out, W_out, 2),
   the `output` will have shape (N, C, H_out, W_out).
+
+  Input are pixels in 2-dimensional space with each pixel as a square whose value is at the square center.
```
gramalingam, line 9329: I think this should be explained better. I personally like what PR 4089 says. But as a compromise and a short summary for now, how about:
```
The tensor `Input` describes the value of a function (over a 2-dimensional space), or pixel, at integer coordinates.
The tensor `grid` describes the grid-positions where the function-value is to be computed using a specified
interpolation method (the mode) and a padding mode (for grid positions falling outside the dimensions of `input`)
```

gramalingam: May be `contains` would be better than `describes` in above.

gramalingam: So, something like
```
The tensor `Input` contains the pixel-values at integer-coordinates. The tensor `grid` describes the grid-positions
(using floating-values) where the pixel-value is to be computed using ... etc. (as above).
```

onnx/defs/tensor/defs.cc
```diff
@@ -2340,15 +2340,16 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-Input are pixels in 2-dimensional space with each pixel as a square whose value is at the square center.
-Output are samples from the 2-dimensional space whose values are obtained using a specified interpolation method (the mode), and a padding mode,
-if a sample position is out of the input image.
-For each output location `output[N, C, H_out, W_out]`, the size-2 vector `grid[N, H_out, W_out]` specifies `input` pixel locations `x` and `y`,
-which are used to interpolate the output value `output[N, C, H_out, W_out]`.
+The tensor `X` contains values at centers of square pixels in a H by W 2-dimensional image.
+The tensor `grid` describes normalized positions where the output `Y` is to be computed
+using a specified interpolation method (the mode) and a padding mode (for grid positions falling outside the 2-dimensional image).
+
+Elements in `grid[N, H_out, W_out]` are size-2 vectors specifying positions in the 2-dimentional space of `X`.
```
gramalingam, line 2351: nit: dimentional => dimensional

gramalingam (approved):

PR #4658, opened 2022-11-15
Fix backend test for ReferenceEvaluator with numpy 1.16.6

xadupre:
Replaces #4657.

### DIFF
--- a/onnx/reference/ops/_op_common_pool.py
+++ b/onnx/reference/ops/_op_common_pool.py
@@ -181,7 +181,12 @@ def _pool(
             no_nan = window_vals[np.where(~np.isnan(window_vals))]
             y[shape] = fpool(no_nan)
             if indices:
-                window_vals_min = np.nan_to_num(window_vals, nan=no_nan.min())
+                try:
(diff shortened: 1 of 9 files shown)


jcwchen:
Thank you for fixing this. Please allow me to reopen this PR to trigger release CIs.

jcwchen (commented):
Tests with NumPy 1.16.6 now looks good, but tests in Windows x86 are still failing. (See https://github.com/onnx/onnx/actions/runs/3471448294/jobs/5800951391)

xadupre:
Sorry for that, building onnx on Windows is always difficult. I'm working on that.

xadupre:
The issues come from the use of int64 for integers. A couple of functions in x64 distribution don't support that.

xadupre:
I added some code to catch the exceptions raised by numpy on python x86 to convert int64 into python int. It was working locally for me. I added an explicit comment that it was needed for x86 in the few places it was needed.

jcwchen (approved):
Overall looks good to me. Thank you for fixing them!

PR #4659, opened 2022-11-15
Corrected a typo in concept.rst

AnandKri:
In line 54, it should be y = onnx.Add(r, c) instead of y = onnx.Add(ax, c)

### Description

In line 54,
Changed `y = onnx.Add(ax, c)` to `y = onnx.Add(r, c)`

### Motivation and Context

Further description of the above code in Documentation was not matching the previous code. Below is the snippet.
`This code implements a function f(x, a, c) -> y = a @ x + c. And x, a, c are the inputs, y is the output. r is an intermediate result.`


jcwchen (approved):
Thanks for catching this! Please signoff DCO: https://github.com/onnx/onnx/blob/main/CONTRIBUTING.md#dco

PR #4670, opened 2022-11-21
fix DFT operator documantation

kuguma:
### Description
Corrected the description about the output shape of the DFT operator.

>axis : int (default is 1)
The axis on which to perform the DFT. By default this value is set to 1, which corresponds to the first dimension after the batch index.

This fix follows the onnx 1.12.0 implementation.

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -6052,7 +6052,7 @@ This version of the operator has been available since version 17 of the default 
 
 <dl>
 <dt><tt>output</tt> : T1</dt>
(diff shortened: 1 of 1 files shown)


kuguma:
image

gramalingam:
@kuguma : thanks for the fix! Unfortunately, the documentation files are auto-generated. The change needs to be done in the source (here: https://github.com/onnx/onnx/blob/main/onnx/defs/math/defs.cc#L2912 ) and then the documentation generated (see: https://github.com/onnx/onnx/blob/8fb26ede15adcda983ed126b2d6dfba52af4e748/tools/update_doc.sh#L31 )

docs/Operators.md
```diff
@@ -6052,7 +6052,7 @@ This version of the operator has been available since version 17 of the default
...
+<dd>The Fourier Transform of the input vector.If onesided is 0, the following shape is expected: [batch_idx][signal_dim1][signal_dim2]...[signal_dimN][2]. If axis=1 and onesided is 1, the following shape is expected: [batch_idx][floor(signal_dim1/2)+1][signal_dim2]...[signal_dimN][2]. If axis=2 and onesided is 1, the following shape is expected: [batch_idx][signal_dim1][floor(signal_dim2/2)+1]...[signal_dimN][2]. If axis=N and onesided is 1, the following shape is expected: [batch_idx][signal_dim1][signal_dim2]...[floor(signal_dimN/2)+1][2]. The signal_dim at the specified axis is equal to the dft_length.</dd>
```
xadupre, line 6055: The same message must be modified in c++ definition as well.

gramalingam:
Created https://github.com/onnx/onnx/pull/4807 to redo this PR since this seems to have stalled. Closing this one.

PR #4681, opened 2022-11-28
misc fixes for issues found in ort integration

liqunfu:


gramalingam (approved):

onnx/test/shape_inference_test.py
```diff
@@ -8567,7 +8567,7 @@ def test_tree_ensemble_classifier(self) -> None:
                 "TreeEnsembleClassifier",
                 ["x"],
                 ["y", "z"],
-                class_ids=[0, 1, 2, 3, 4],
+                classlabels_int64s=[0, 1, 2, 3, 4],
```
gramalingam, line 8570: Curious what changed here? Is this just an old bug?

gramalingam: Oh, I remember ... I guess the shape-inference was added recently to these onnxml operators.

onnx/defs/traditionalml/defs.cc
```diff
@@ -864,10 +864,16 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
           unifyInputDim(ctx, 0, 0, N);
-          std::vector<int64_t> class_ids;
-          auto has_ids = getRepeatedAttribute(ctx, "class_ids", class_ids);
-          if (has_ids) {
-            unifyDim(E, class_ids.size());
+
+          if (using_strings) {
+            unifyDim(E, classlabels_strings.size());
+          } else {
+            std::vector<int64_t> classlabels_int64s;
```
gramalingam, line 871: Nit: we could also check if both attributes are given and produce an error message in this case.

liqunfu: Agree. 
I also think the op's documentation can be improved.

onnx/test/shape_inference_test.py
liqunfu: class_ids' size is not the number of classes. Instead, either classlabels_strings or classlabels_int64s has a size of the output class numbers.

liqunfu: Yes it is an old bug.

p-wysocki:
Is this related to https://github.com/microsoft/onnxruntime/issues/13747? I'm wondering if it should be backported to the release branch.

liqunfu:
> Is this related to [microsoft/onnxruntime#13747](https://github.com/microsoft/onnxruntime/issues/13747)? I'm wondering if it should be backported to the release branch.

Good point. Yes, it is related to the issue. I am now validating ONNX 1.13 rc with ORT. Will update once I have completed my validation. Thanks

onnx/defs/traditionalml/defs.cc
```diff
@@ -28,21 +28,43 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
           const auto input_ndim = input_shape.dim_size();
 
           auto output_shape = ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
-          // This operator only applies to the last dimension; thus -1
-          for (int i = 0; i < input_ndim - 1; ++i) {
-            *output_shape->add_dim() = input_shape.dim(i);
+          if (input_ndim == 1) {
+            output_shape->add_dim()->set_dim_value(1);
```
gramalingam, line 32: I don't understand why this is needed. Is the onnxruntime implementation adding an extra dimension in this case?

onnx/defs/traditionalml/defs.cc
```diff
@@ -28,21 +28,43 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
...
+          auto last_dim = output_shape->add_dim();
           if (hasInputShape(ctx, 1)) {
             const auto& indices_shape = getInputShape(ctx, 1);
             if (indices_shape.dim_size() > 0) {
-              auto dim = indices_shape.dim(0);
-              *output_shape->add_dim() = dim;
-              return;
+              int64_t num_indices = 1;
+              std::string single_symbolic_dim;
+              for (int i = 0; i < indices_shape.dim_size(); i++) {
```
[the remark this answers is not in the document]
gramalingam: If so, it would be worth explaining this also in the operator documentation. The current documentation is so sketchy.

onnx/test/shape_inference_test.py
```diff
@@ -8620,3 +8620,4 @@ def test_array_feature_extractor(self) -> None:
 
 if __name__ == "__main__":
     unittest.main()
+    unittest.main()
```
xadupre, line 8623: duplicated line

gramalingam (approved):

PR #4684, opened 2022-12-01
Fix input_1 data to has got same shape as in model

praasz:
### Description
The test models in backend tests:
 - `test_castlike_BFLOAT16_to_FLOAT` 
 - `test_castlike_FLOAT_to_BFFLOAT16`
 have changed between versions 1.11 and 1.12. The input 'Like' has got shape `3x4` instead `1` but data for this input is still shape of `1`. 

This change add consistency between model and data.

### Motivation and Context
The [OpenVINO](https://github.com/openvinotoolkit/openvino) project use this tests. After upgrade from `1.11` -> `1.12` these test start failing as there is no consistency between input data and inputs shapes.


daquexian:
Thanks for your contribution!

However, it seems that the shape of `like` should be [1] instead of [3, 4]

https://github.com/onnx/onnx/blob/791ab90a320dd9a5948c951146f28aa808a1e840/onnx/backend/test/case/node/castlike.py#L123

And I think the problem is that we specified a wrong type proto here:

https://github.com/onnx/onnx/blob/791ab90a320dd9a5948c951146f28aa808a1e840/onnx/backend/test/case/node/castlike.py#L135 

`output_type_proto` (whose shape is [3,4]) is used as the type proto of `like` (whose shape is [1])

praasz:
> Thanks for your contribution!
> 
> However, it seems that the shape of `like` should be exactly [1] instead of [3, 4]
> 
> https://github.com/onnx/onnx/blob/791ab90a320dd9a5948c951146f28aa808a1e840/onnx/backend/test/case/node/castlike.py#L123
> 
> And I think the mismatch you encountered comes from that we specified a wrong type proto here:
> 
> https://github.com/onnx/onnx/blob/791ab90a320dd9a5948c951146f28aa808a1e840/onnx/backend/test/case/node/castlike.py#L135
> 
> `output_type_proto` (whose shape is [3,4]) is incorrectly used as the type proto of `like` (whose shape is [1])

You've got right about proto. I've created like proto which use correct shape and generated test model will have correct inputs aligned with test data.

p-wysocki (approved):

gramalingam (approved):

p-wysocki:
@gramalingam @jcwchen could you please approve necessary CI workflows?

PR #4692, opened 2022-12-06
update test_dft pb files

ZonePG:
### Description

on my m1 mac, after cloning and build from source code, run `pytest`

I found 3 test files (`test_dft_*`) failed maybe because the expected output data was incorrect.


gramalingam:
The binary data for the test-cases are generated. If there is a problem, it should be fixed in the source code.

The source code for this test-case is here: https://github.com/onnx/onnx/blob/37ba77e1360d3691fede22d9dfe3f24e7fce8901/onnx/backend/test/case/node/dft.py#L21

I guess there is some platform-specific issue here? But the test-case is binary, so it seems unlikely that the problem could be with the test-case generation.

@smk2007, any ideas/thoughts? Thanks.

ZonePG:
> The binary data for the test-cases are generated. If there is a problem, it should be fixed in the source code.
> 
> The source code for this test-case is here:
> 
> https://github.com/onnx/onnx/blob/37ba77e1360d3691fede22d9dfe3f24e7fce8901/onnx/backend/test/case/node/dft.py#L21
> 
> I guess there is some platform-specific issue here? But the test-case is binary, so it seems unlikely that the problem could be with the test-case generation.
> 
> @smk2007, any ideas/thoughts? Thanks.

I think It is indeed a platform-specific problem. After I run `onnx/backend/test/cmd_tools` to regenerate the test data, there is no problem. I will close this pr. Thanks!

PR #4683, opened 2022-11-30
Update `OnnxReleases.md`

p-wysocki:
Signed-off-by: p-wysocki <przemyslaw.wysocki@intel.com>

### Description
- add Python 3.11 to validation table
- remove link to `keras2onnx`, which is archived and frozen to `onnx==1.10`: https://github.com/onnx/keras-onnx

PR #4701, opened 2022-12-12
Bump current opset to 19

p-wysocki:

### Description
Release branch has been cut, new opset can be introduced. Based on https://github.com/onnx/onnx/pull/4134.

### DIFF
--- a/onnx/defs/operator_sets.h
+++ b/onnx/defs/operator_sets.h
@@ -1058,6 +1058,15 @@ class OpSet_Onnx_ver18 {
   }
 };
 
+// Forward declarations for ai.onnx version 19
+
+// Iterate over schema from ai.onnx version 19
+class OpSet_Onnx_ver19 {
+ public:
+  static void ForEachSchema(std::function<void(OpSchema&&)> fn) {
+  }
+};
+
 inline void RegisterOnnxOperatorSetSchema() {
   RegisterOpSetSchema<OpSet_Onnx_ver1>();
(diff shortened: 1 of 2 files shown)


gramalingam:
> Looks like the compiler throws an error due to an unused variable `fn`. Should I close the PR and leave adding this to the first new operator in opset19 or should I add a dummy placeholder?

I think you should be able to just have a dummy use, like `(void) fn;`  (if that's what you meant).

jcwchen (approved):
LGTM. I think this PR will be a better demo PR than mine (https://github.com/onnx/onnx/pull/4134), because it handles `(void)fn` :) Probably we can update  the demo PR in OnnxReleases.md.

p-wysocki:
@jcwchen Done in https://github.com/onnx/onnx/pull/4683.

onnx/defs/operator_sets.h
```diff
@@ -1058,6 +1058,16 @@ class OpSet_Onnx_ver18 {
   }
 };
 
+// Forward declarations for ai.onnx version 19
+
+// Iterate over schema from ai.onnx version 19
+class OpSet_Onnx_ver19 {
+ public:
+  static void ForEachSchema(std::function<void(OpSchema&&)> fn) {
+    (void)fn;
```
jcwchen, line 1067: tiny nit: perhaps add a TODO saying this line should be removed after introducing an actual new schema?

p-wysocki: Done.

Issue #4706, opened 2022-12-15
DeprecatedWarningDict not printing the replacement


zhongsanming:
# Bug Report

### Describe the bug
https://github.com/onnx/onnx/blob/45f508bb7c3206cf504b923d54076e42d1ed591c/onnx/mapping.py#L86-L105

in line 100, the string should be a f-string

jcwchen:
Hi @zhongsanming,
Thank you for catching this! I will submit a PR to fix this.

PR #4707, opened 2022-12-15
Add missing f-string for DeprecatedWarningDict in mapping.py

jcwchen:
### Description

Add missing f-string for DeprecatedWarningDict in mapping.py to correctly print replacement.

### Motivation and Context
Fixes https://github.com/onnx/onnx/issues/4706.


onnx/mapping.py
```diff
@@ -97,7 +97,7 @@ def __getitem__(self, key: Union[int, str, np.dtype]) -> Any:
             warnings.warn(
                 str(
```
justinchuby, line 98: Remove str and do

```python
    f"`mapping.{self._origin_function}` is now deprecated and will be removed in the next release or so."
    f"To silence this warning, please use `helper.{self._future_function}` instead."
```

?

optional, nit:

"in the next release or so" -> "in a future release"?

onnx/mapping.py
```diff
@@ -96,8 +96,8 @@ def __getitem__(self, key: Union[int, str, np.dtype]) -> Any:
         else:
             warnings.warn(
                 str(
-                    f"`mapping.{self._origin_function}` is now deprecated and will be removed in the next release or so."
-                    + "To silence this warning, please use `helper.{self._future_function}` instead."
+                    f"`mapping.{self._origin_function}` is now deprecated and will be removed in a future release."
+                    + f"To silence this warning, please use `helper.{self._future_function}` instead."
```
xadupre, line 100: `+` is not needed here. Python interpreter automatically merges strings when surrounded by `()`.

jcwchen: Good catch! Updated. Thanks!

PR #4714, opened 2022-12-19
Add reference implementation for domain ai.onnx.ml

xadupre:

### Description
Domain ai.onnx.ml is missing from the reference implementation. This PR adds it. It implements all operators except ZipMap and checks the output is close to onnxruntime outputs. Parameter `post_transform` is  implemented for regressor. The specifications are too vague. ZipMap is left out. Users should not use it. It slows down models and its outputs is never consumed by any other operator.

### Motivation and Context
Complete implementation of `ReferenceEvaluator` including domain `ai.onnx.ml`.


onnx/test/reference_evaluator_ml_test.py
```diff
@@ -0,0 +1,1414 @@
...
+                onx = make_model(graph, opset_imports=OPSETS)
+                check_model(onx)
+                # onnxruntime answer seems odd.
+                # self._check_ort(onx, {"X": x}, rev=True)
+                sess = ReferenceEvaluator(onx)
+                got = sess.run(None, {"X": x})
+                assert_allclose(expected[1], got[1], atol=1e-6)
+                assert_allclose(expected[0], got[0])
+
+    @unittest.skipIf(not ONNX_ML, reason="onnx not compiled with ai.onnx.ml")
```
justinchuby, line 647: I wonder if the whole class can be skipped conditionally to make things simpler

xadupre: I did not write a test checking the runtime does not work in case the flag is not present. But this way of doing makes it still possible.

justinchuby:
I think it’s a safe choice to disable too-many-returns under https://github.com/onnx/onnx/blob/main/pyproject.toml#L63

onnx/reference/ops/aionnxml/op_linear_classifier.py
```diff
@@ -51,7 +51,7 @@ def _run(  # type: ignore
         elif post_transform == "LOGISTIC":
             scores = expit(scores)
         elif post_transform == "SOFTMAX":
-            np.subtract(scores, scores.max(axis=1, keepdims=1), out=scores)
+            np.subtract(scores, scores.max(axis=1, keepdims=1), out=scores)  # pylint: disable=E1123
```
justinchuby:
```suggestion
            np.subtract(scores, scores.max(axis=1, keepdims=1), out=scores)  # pylint: disable=unexpected-keyword-arg
```

for readability?

gramalingam (approved):
CodeQL/linter etc. seem to have a number of suggestions, which may be worth looking into.

Issue #4687, opened 2022-12-03
onnx.compose.add_prefix_graph won't rename subgraph


Yosshi999:
# Bug Report

### Describe the bug
onnx.compose.add_prefix_graph won't rename subgraph, such as If branch. (also Loop block?)
This means it may break the onnx graph which contains If / Loop nodes.

### System information
Google Colab CPU

### Reproduction instructions
https://gist.github.com/Yosshi999/e6fda24ddc0e26687a2540c5db32d658

### Expected behavior

onnx.checker.check_graph passes.

### Notes
This function has to call the renaming routine recursively when it meets If / Loop nodes.

jcwchen:
Hi @Yosshi999,
Thanks for catching this issue. I can repro it. I have submitted a PR to fix it: https://github.com/onnx/onnx/pull/4718. Feel free to review. Thanks!

PR #4718, opened 2022-12-22
Rename attribute's subgraph in onnx.compose.add_prefix_graph

jcwchen:
### Description

Rename attribute's subgraph in onnx.compose.add_prefix_graph

### Motivation and Context

Fixes https://github.com/onnx/onnx/issues/4687.


onnx/compose.py
```diff
@@ -420,7 +420,9 @@ def add_prefix_graph(
     rename_outputs: Optional[bool] = True,
     rename_initializers: Optional[bool] = True,
     rename_value_infos: Optional[bool] = True,
+    rename_attribute_graph: Optional[bool] = True,
```
gramalingam, line 423: Can we drop this flag? I am not sure if it makes sense to set some other rename flag to true, and make this false. We may as well recursively process subgraphs always.

jcwchen: Good point. Just removed. Thanks!

gramalingam (approved):

onnx/compose.py
```diff
@@ -421,6 +421,7 @@ def add_prefix_graph(
     rename_initializers: Optional[bool] = True,
```
justinchuby, line 421: nit: Do these flags ever take None? If not I would type them as `bool` instead of `Optional[bool]`

Issue #4700, opened 2022-12-09
[Feature request] ONNX checker checks function proto validity 


justinchuby:
### What is the problem that this feature solves?

ONNX checker currently checks model proto validity with check_model_proto. I would be very helpful to have a similar API that checks for function proto validity.

This greatly helps testing function libraries built with ONNX script.

### Alternatives considered

Make a model. Not all functions can be directly converted into a Model. 

### Feature Area

checker

### Are you willing to contribute it (Y/N)

None

jcwchen:
Hi @justinchuby,
Thanks for proposing this. Currently ONNX only provides [check_function](https://github.com/onnx/onnx/blob/main/onnx/checker.cc#L867-L956) in C++. Does the current check meet your requirement? If it is useful, we can also expose a Python API for it.

justinchuby:
A Python api would be great/necessary for unit testing the function library.

PR #4720, opened 2022-12-23
Expose check_function in Python API

jcwchen:
### Description

- Expose check_function in Python API.
- Add a test
- Add more functions in API doc on Web.

### Motivation and Context

Closes https://github.com/onnx/onnx/issues/4700.


onnx/cpp2py_export.cc
```diff
@@ -372,6 +372,12 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
     checker::check_node(proto, ctx, lex_ctx);
   });
 
+  checker.def("check_function", [](const py::bytes& bytes, const checker::CheckerContext& ctx) -> void {
```
xadupre, line 375: Is it possible to add a link in the documentation (here: https://github.com/onnx/onnx/blob/main/docs/docsgen/source/api/checker.rst) so that this function appears on the documentation?

jcwchen: Good idea! I will try to add it in this PR.

onnx/cpp2py_export.cc
```diff
@@ -372,6 +372,12 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+    FunctionProto proto{};
+    ParseProtoFromPyBytes(&proto, bytes);
+    checker::check_function(proto, ctx, checker::LexicalScopeContext());
```
gramalingam, line 378: I wonder why the last parameter (lexical scope context) is not exposed to the Python interface? I realize that the same problem exists in the other existing methods, but maybe we should consider exposing it (while maintaining backward compatibility).

docs/docsgen/source/api/checker.rst
```diff
@@ -5,6 +5,36 @@ onnx.checker
 .. contents::
     :local:
 
+check_value_info
++++++++++++
```
xadupre, line 9: The underline should have the same length.

docs/docsgen/source/api/checker.rst
```diff
@@ -5,6 +5,36 @@ onnx.checker
 .. contents::
     :local:
 
+check_value_info
```
xadupre, line 8: `C.CheckerContext`  should be exposed as well.

justinchuby:
Is this ready? 👀

docs/docsgen/source/api/checker.rst
jcwchen: Updated. Thanks!

docs/docsgen/source/api/checker.rst
jcwchen: Just exposes it. Not sure whether the usage is correct though. PTAL. Thanks!

jcwchen:
> Is this ready? 👀

I think this PR will be very helpful for debugging FunctionProto with onnx-script. Rama has a [concern](https://github.com/onnx/onnx/pull/4720#discussion_r1068360499) that ideally we should expose lexical scope context as well. To faster provide this functionality, @gramalingam is it OK for us to merge this first? I will try to further expose lexical scope context in another PR.

gramalingam:
> To faster provide this functionality, @gramalingam is it OK for us to merge this first? I will try to further expose lexical scope context in another PR.

Sure, that's fine

gramalingam (approved):

PR #3306, opened 2021-03-02
Remove usage of numpy aliases for builtins (#3305)

SnoopJeDi:
As described in #3305, beginning in [NumPy 1.20](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations), the use of type aliases for builtin types (`np.object, np.bool, np.int`, etc.) is deprecated. This raises a `DeprecationWarning` when used in `onnx`.

This PR replaces deprecated usage with NumPy's official recommendations.

PR #4721, opened 2022-12-23
Fix types deprecated in `numpy==1.24`

p-wysocki:

### Description
- Fix types deprecated in `numpy==1.24`


onnx/backend/test/case/node/castlike.py
```diff
@@ -87,14 +87,14 @@ def export() -> None:
...
-                    # Converting input to str, then give it np.object dtype for generating script
+                    # Converting input to str, then give it np.object_ dtype for generating script
                     ss = []
                     for i in input.flatten():
                         s = str(i).encode("utf-8")
                         su = s.decode("utf-8")
                         ss.append(su)
 
-                    output = np.array(ss).astype(np.object).reshape([3, 4])
+                    output = np.array(ss).astype(np.object_).reshape([3, 4])
```
daquexian, line 97: Is `object` better than `np.object_` here?

> To give a clear guideline for the vast majority of cases, for the types bool, object, str (and unicode) using the plain version is shorter and clear, and generally a good replacement.

https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

jcwchen: +1 I think object is better, which is just like what this PR: https://github.com/onnx/onnx/pull/3306 did.

onnx/backend/test/case/node/bernoulli.py
```diff
@@ -25,8 +25,8 @@ def export_bernoulli_without_dtype() -> None:
             outputs=["y"],
         )
 
-        x = np.random.uniform(0.0, 1.0, 10).astype(np.float)
-        y = bernoulli_reference_implementation(x, np.float)
+        x = np.random.uniform(0.0, 1.0, 10).astype(np.float32)
+        y = bernoulli_reference_implementation(x, np.float32)
```
jcwchen, line 29:
```suggestion
        y = bernoulli_reference_implementation(x, float)
```
Actually np.float was np.float64 instead of np.float32. Still, directly using float seems more clear here.

jcwchen (commented):
Few comments. Thanks for catching this!

onnx/backend/test/case/node/bernoulli.py
p-wysocki: I changed the reference implementation values dtype to `float`s.

onnx/backend/test/case/node/bernoulli.py
```diff
@@ -25,8 +25,8 @@ def export_bernoulli_without_dtype() -> None:
             outputs=["y"],
         )
 
-        x = np.random.uniform(0.0, 1.0, 10).astype(np.float)
-        y = bernoulli_reference_implementation(x, np.float)
+        x = np.random.uniform(0.0, 1.0, 10).astype(np.float32)
```
jcwchen, line 28:
```suggestion
        x = np.random.uniform(0.0, 1.0, 10).astype(float)
```

docs/Operators.md
```diff
@@ -2385,8 +2385,8 @@ node = onnx.helper.make_node(
     outputs=["y"],
 )
 
-x = np.random.uniform(0.0, 1.0, 10).astype(np.float)
-y = bernoulli_reference_implementation(x, np.float)
+x = np.random.uniform(0.0, 1.0, 10).astype(np.float32)
```
jcwchen, line 2388:
```suggestion
x = np.random.uniform(0.0, 1.0, 10).astype(float)
```

docs/TestCoverage.md
```diff
@@ -1706,8 +1706,8 @@ node = onnx.helper.make_node(
     outputs=["y"],
 )
 
-x = np.random.uniform(0.0, 1.0, 10).astype(np.float)
-y = bernoulli_reference_implementation(x, np.float)
+x = np.random.uniform(0.0, 1.0, 10).astype(np.float32)
```
jcwchen, line 1709:
```suggestion
x = np.random.uniform(0.0, 1.0, 10).astype(float)
```

jcwchen (commented):
Fixing these should help to resolve the test failures.

p-wysocki:
Thank you @jcwchen, the tests are now passing locally.

jcwchen (approved):
Thank you for the fix!

gramalingam (approved):

PR #4736, opened 2023-01-03
Replace np.float by np.float32, np.float was removed in numpy==1.24

xadupre:

### Description
A couple of backend example fail due to the use of deprecated type `np.float` and removed in numpy==1.24. This PR replaces it by `np.float32`.

### Motivation and Context
Fix a bug happening with numpy==1.24.


daquexian:
Thanks for the contribution! However, I'm afraid there is already a PR doing similar things https://github.com/onnx/onnx/pull/4721

xadupre:
I missed it. I'll close it then.

PR #4743, opened 2023-01-05
Add missing class Split_18 in reference implementation

xadupre:

### Description
Split implementation in reference implementation was missing. This PR adds it.


onnx/test/reference_evaluator_test.py
```diff
@@ -1965,6 +1965,96 @@ def test_stft_with_window(self):
         got1 = ref1.run(None, feeds)
         assert_allclose(expected, got1[0])
 
+    def test_split(self):
```
gramalingam, line 1968: I wonder if we should just rely on ONNX node test cases? I believe the reference implementation is tested against ONNX node test cases right? Why duplicate test-cases?

xadupre: The test cases do not cover all the possible cases, negative values for axis, all possible options for a parameter... There is no test case for ai.onnx.ml.

gramalingam (approved):

PR #4748, opened 2023-01-05
Add attribute type in documentation

### LINKED ISSUE #4724: Attribute types and default values missing in docs at onnx.ai
# Bug Report

### Describe the bug

The HTML version of the ONNX operator specs at https://onnx.ai/onnx/operators/index.html are missing types and default values for attributes. Compare the [HTML version of Resize](https://onnx.ai/onnx/operators/onnx__Resize.html#resize-18) against the [markdown version](https://github.com/onnx/onnx/blob/main/docs/Operators.md#resize). This is annoying as the HTML version loads faster and is more convenient to browse otherwise.

### Expected behavior

The HTML version of the operator documentation should include types and default values for attributes.

xadupre:

### Description
Attribute type and default value are easier to find in the documentation.
It fixes the configuration of the documentation for sphinx 6.0.

Rendering:

image

Fixes #4724.


jcwchen (approved):
Thank you for the improvement!

Not relevant to this PR and just think out loud: is it possible that in the PR we can have certain CI pipeline generate updated web files from the modified files? It would be helpful for the reviewers to see the previews.

docs/docsgen/source/conf.py
```diff
@@ -12,7 +12,7 @@
 # -- Project information -----------------------------------------------------
 
 author = "ONNX"
-copyright = "2022"
+copyright = "2023"
```
gramalingam, line 15: nit: can the year be auto-generated?

gramalingam: Not sure what it should be, however. Based on date of original or when documentation-generation runs.

gramalingam:
A related question that came up previously was when we should run/update the published documentation. For PRs like this, we do want to update the published documentation, but we don't want to publish any op changes/additions since last release.

We could satisfy both by updating the documentation-generator to consider only opschemas belonging to last released version. @jcwchen, what do you think?

gramalingam (approved):

jcwchen:
Previously I was thinking having a special branch which only has updates about every web content update except new operators. Then we can either automatically deploy the content or manually.

> We could satisfy both by updating the documentation-generator to consider only opschemas belonging to last released version.

Your method looks much easier, as long as it's doable. Perhaps we can directly modify the web content generator logic for operator document with a specified released opset version? Then, we can eventually make the main branch automatically generate updated web content without introducing developing operators.

xadupre:
This function (https://github.com/onnx/onnx/blob/main/docs/docsgen/source/onnx_sphinx.py#L265) gathers all schemas to document. We could add a parameter in `conf.py` to specify up to which opset to document. I can prepare a separate PR for that.

Another thing we can do is to keep the documentation for every release and show something like: 
image
.

PR #4753, opened 2023-01-08
Remove operators under development from the documentation

xadupre:

### Description
The documentation includes all available operators by default including those not published yet. This change adds a parameter to documentation configuration to keep or remove those.


docs/docsgen/source/conf.py
```diff
@@ -16,6 +16,7 @@
 project = "ONNX"
 release = onnx.__version__
 version = onnx.__version__
+max_opsets = {"": 18, "ai.onnx.ml": 3}
```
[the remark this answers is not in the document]
gramalingam: Can we just get it programmatically from this table: https://github.com/onnx/onnx/blob/19f700e01b05a263d121d6820d35df7dbf1e7918/onnx/helper.py#L51

gramalingam (approved):
LGTM, thanks. May be worth addressing the comment about max_opsets above.

docs/docsgen/source/conf.py
```diff
@@ -17,6 +17,18 @@
 release = onnx.__version__
 version = onnx.__version__
 
+# define the latest opset to document,
+# this is meant to avoid documenting opset not released yet
+max_opset = 18
```
gramalingam, line 22: Why not
```py
max_opset = onnx.helper.VERSION_TABLE[-1][2]
```
That way, it will automatically do the right thing after a new release. Or, do we want to manually control this?

gramalingam: That's what the above code will do. The table is supposed to be updated to include 19 only at the time of next release. What am I missing?

xadupre: I'll do that. I was not sure how this table was updated.

xadupre: done

gramalingam (approved):

PR #4761, opened 2023-01-10
add lppool test cases

liqunfu:

### Description
lpPool op has been added. Test cases are added in this PR.

### Motivation and Context
add test cases for backend and runtime validation


onnx/backend/test/case/node/pool_op_common.py
```diff
@@ -95,12 +106,14 @@ def pool(
...
             f = np.max
+        elif pooling_type == "LPPOOL":
+            f = lp_pool_p
         else:
             raise NotImplementedError(
                 f"Pooling type {pooling_type} does not support. Should be AVG, MAX"
             )
 
-        if count_include_pad == 1 and pooling_type == "AVG":
+        if count_include_pad == 1 and (pooling_type == "AVG" or pooling_type == "LPPOOL"):
```
gramalingam, line 116: I am curious about the existing else branch below. Does the ONNX spec specify what happens with NaN values for the pooling ops? (Not specific to current PR, but something worth checking.)

onnx/backend/test/case/node/lppool.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+class LpPool(Base):
+    @staticmethod
+    def export_lppool_1d_default() -> None:
+        """
+        input_shape: [1, 3, 32]
+        output_shape: [1, 3, 31]
+        """
+        p = 3
```
gramalingam, line 19: I see 3 and 4 used a lot for p. May be worth changing some of them to p=1.

gramalingam (commented):
LGTM. Suggest changing some of the p values to 1, which is a useful special case.

gramalingam (approved):

p-wysocki (approved):

onnx/backend/test/case/node/pool_op_common.py
liqunfu: if count_include_pad is set to 1, the caller needs to specify a value for padded pixels. In this case, there is not NaN in window_vals. If count_include_pad is not set to 0 that means the caller does not have a value for the padded pixel. In such case, padded contains NaN and so does window_vals. In this case, NaN shall be excluded from computation.
BTW compare any number with NaN is false so there is no need for MaxPool.

onnx/backend/test/case/node/lppool.py
liqunfu: p=1 will be proportionally AvgPool. I am flexible there though.

gramalingam: Just to clarify: I think p=1 differs from AveragePool in two ways: first, it uses the absolute value and second, it computes a sum, not average.

liqunfu: the absolute value is what we shall discuss - I do not see pytorch's implementation does that. there is an issue which we shall solve in ONNX and then in Ort. https://github.com/microsoft/onnxruntime/issues/6302. With that issue soloved, p=1 is close to AvgPool as Pytorch document stated:
At p = 
∞
∞, one gets Max Pooling

At p = 1, one gets Sum Pooling (which is proportional to average pooling)

PR #4654, opened 2022-11-13
Adopt lintrunner for linting

justinchuby:
### Description

- Adopt lintrunner for linting
- Update CI and readme
- Run `lintrunner -a` on all files
- Explicitly import all proto classes in the onnx namespace to (1) make mypy happy and (2) avoid import *
- Add `.editorconfig`

### Motivation and Context

Lint everything with a single command and config

Now we don't need the custom grep script to check namespaces:

(demo example)

image

PR #4763, opened 2023-01-11
Enable pylint for onnx

justinchuby:
### Description

- Enable pylint for onnx via lintrunner
- Update pylint config to a separate file to be more relaxed in CI.
- Remove `tools/mypy-onnx.py`

### Motivation and Context

Code quality

cc @jcwchen @gramalingam


onnx/backend/sample/ops/__init__.py
```diff
@@ -9,17 +9,17 @@
 
 
 def collect_sample_implementations() -> Dict[str, str]:
-    dict: Dict[str, str] = {}
-    _recursive_scan(sys.modules[__name__], dict)
-    return dict
+    dict_: Dict[str, str] = {}
```
justinchuby, line 12: Avoid redefining builtin `dict`

tools/mypy-onnx.py
```diff
@@ -1,25 +0,0 @@
...
-        subprocess.check_call(["mypy", ".", "--no-site-packages"])
-
-        exit(0)
-    except subprocess.CalledProcessError:
-        # Catch this exception because we don't want it to output a backtrace that would clutter the mypy output
-        exit(1)
-
-
-if __name__ == "__main__":
-    main()
```
jcwchen, line 25: Is this file no longer needed because of this PR https://github.com/onnx/onnx/pull/4654?

.github/workflows/lint.yaml
```diff
@@ -51,14 +51,6 @@ jobs:
...
           level: info
           filter_mode: diff_context
-      - name: pylint
-        uses: dciborow/action-pylint@0.0.7
-        with:
-          github_token: ${{ secrets.github_token }}
-          reporter: github-pr-check
-          level: warning
-          filter_mode: diff_context
-          glob_pattern: "**/*.py"
```
jcwchen, line 61: To confirm, so go forward pylint check will be included in the new .lintrunner. Am I correct?

onnx/compose.py
```diff
@@ -76,7 +78,7 @@ def _edge_names(graph: GraphProto, exclude: Optional[Set[str]] = None) -> List[s
     return result
 
 
-def merge_graphs(
+def merge_graphs(  # pylint: disable=too-many-branches,too-many-statements
```
jcwchen, line 81: Do you think it is OK for us to disable too-many-branches and too-many-statements globally?

jcwchen (commented):
Thank you for fixing many pylint errors in this PR! Just have few questions.

tools/mypy-onnx.py
justinchuby: I think so.

.github/workflows/lint.yaml
justinchuby: That's right

onnx/compose.py
justinchuby: I think it is still a good reminder to write small(er) functions. We use this to exempt existing functions, but future functions, if having many branches or statements, should still declare their conscious intent by adding this disable inline.

gramalingam: What's the cutoff size for triggering this?

justinchuby: I think it is 50 for statements and 12 for branches. It can be adjusted

jcwchen (approved):
LGTM. Thanks for answering!

gramalingam (approved):

justinchuby:
@jcwchen ready to merge I think!

PR #4769, opened 2023-01-12
fix col2im test data

liqunfu:

### Description
fix a data error in col2im test case

### Motivation and Context
bug fix


jcwchen (commented):
Is this issue caught when verifying ONNX in ONNX Runtime?

gramalingam (approved):

PR #4770, opened 2023-01-12
Add one backend test mode==max of RoiAlign + ReferenceEvaluator

xadupre:

### Description
Backend tests do not cover `mode=='max'` for RoiAlign. The test checks that onnxruntime and onnx reference implementation produces the same results.


gramalingam (approved):
Would be great to update backend tests to add such a test.

gramalingam:
@xadupre : in case you overlooked this, it looks like there's a conflict to be resolved. Otherwise, it's good to go

xadupre:
done

PR #4775, opened 2023-01-13
Add reference implementation for LpPool

xadupre:

### Description
Add reference implementation for LpPool.

### DIFF
--- a/onnx/reference/ops/op_lp_pool.py
+++ b/onnx/reference/ops/op_lp_pool.py
@@ -16,7 +16,6 @@ class LpPool(CommonPool):
         kernel_shape=None,
         p=None,
         pads=None,
-        storage_order=None,
         strides=None,
     ):
         if (
@@ -33,7 +32,6 @@ class LpPool(CommonPool):
                 kernel_shape=kernel_shape,
                 p=p,
                 pads=pads,
-                storage_order=storage_order,
                 strides=strides,
             )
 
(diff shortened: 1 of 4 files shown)


onnx/test/reference_evaluator_test.py
```diff
@@ -2611,6 +2611,104 @@ def test_mvn(self):
                 self.assertEqual(expected.shape, got.shape)
                 assert_allclose(expected, got)
 
+    def test_lp_pool_1d_1(self):
```
gramalingam, line 2614: Please take a look at https://github.com/onnx/onnx/pull/4761 ... is the test-coverage there adequate? In the long run, it would help to ensure that can serve as a complete test-suite for all onnx backends.

onnx/reference/ops/op_lp_pool.py
```diff
@@ -0,0 +1,360 @@
...
+                auto_pad=auto_pad,
+                ceil_mode=ceil_mode,
+                dilations=dilations,
+                kernel_shape=kernel_shape,
+                p=p,
+                pads=pads,
+                strides=strides,
+            )
+
+        return CommonPool._run(
```
gramalingam, line 38: So, does this not handle dilations/stride? Trying to understand the special case above.

onnx/test/reference_evaluator_test.py
xadupre: I created the unit tests because there was no backend test to use. But I'll change that when PR #4761 is merged.

xadupre:
Done in another PR.

PR #4776, opened 2023-01-13
Fix missing page for ai.onnx.ml domain

xadupre:

### Description
ai.onnx.ml are missing from page: https://onnx.ai/onnx/operators/. It was broken by PR: https://github.com/onnx/onnx/pull/4753.


gramalingam (approved):
Hmmm, it looks like all the new code is just extra-validation. I don't see what was causing the missing page problem or its fix.

xadupre:
I used ops as a loop variable but it was also used as a function argument. I should probably remove ops since it is not used.

Issue #2665, opened 2020-03-17
DCN is supported in frameworks time to add it to ONNX ?


YumainOB:
As posted by @vlordier in the too quickly closed issue #2395 (IMHO), DCN is now a well supported feature in various frameworks. Maybe time has come to add this feature to ONNX? Is there a plan to have it in future releases?

> But to generate it in ONNX it need to be converted from someplace. If this cause a massive accuracy gain, why no framework implemented it naively? Especially, this was published 2 years ago. To add an OP in ONNX, it needs to be available in one framework at least.

@ebarsoum Deformable Convolutions are also used to process equirectangular projections of omnidirectional images by parametrising the convolution kernel

![74554848-5fb79200-4f52-11ea-9862-8f86c4a097fb](https://user-images.githubusercontent.com/5443125/75030134-62a70b00-549b-11ea-8ee0-15f9f69cdc39.png)

More recent work [Kernel Transformer Networks for Compact Spherical Convolution](https://arxiv.org/pdf/1812.03115.pdf) 

[Distortion-Aware Convolutional Filters for Dense
Prediction in Panoramic Images](http://campar.in.tum.de/pub/tateno2018eccv/tateno2018eccv.pdf
)

or [Corners for Layout: End-to-End Layout Recovery from 360 Images](https://arxiv.org/pdf/1903.08094.pdf)

_Originally posted by @vlordier in https://github.com/onnx/onnx/issues/2395#issuecomment-589613934_

linkerzhang:
are you interested in proposing op definitions PRs needed for it?

YumainOB:
Thanks for your reply @linkerzhang 
As I am not an expert in ONNX  (just a basic user in facts), it maybe quite long if I do this proposal and may results in inaccurate and inefficient solution. Maybe someone with better knowledge of the ONNX project would be better suited for this proposal?

Zha0q1:
I am also interested in this op. Is anyone moving this forward at the moment?

hyperfraise:
Also very interested

marcown:
Any news on DCN for onnx?

daixiangzi:
I find baidu:https://github.com/PaddlePaddle/Paddle2ONNX/blob/develop/docs/zh/op_list.md support DCN to onnx

Zha0q1:
> I find baidu:https://github.com/PaddlePaddle/Paddle2ONNX/blob/develop/docs/zh/op_list.md support DCN to onnx

but how 🤔

[...]

Issue #2395, opened 2019-10-11
Will onnx support dcn future?


jinfagang:
Is that deformable convolution network will be added into onnx?

ebarsoum:
Is DCN implemented in any framework natively?

jinfagang:
I think not. However, this is a good feature since DCN makes convolution networks gain a massive accuracy. I'll go to framework community to ask this feature

ebarsoum:
But to generate it in ONNX it need to be converted from someplace. If this cause a massive accuracy gain, why no framework implemented it naively? Especially, this was published 2 years ago. To add an OP in ONNX, it needs to be available in one framework at least.

zldrobit:
@ebarsoum MXNet has deform conv implementation (https://mxnet.incubator.apache.org/api/python/docs/api/gluon/contrib/index.html).
Moreover, PyTorch member is considering adding deform conv (https://github.com/pytorch/pytorch/issues/28486)

vlordier:
@ebarsoum 
Deformable Convolutions are now supported indeed in those 3 frameworks

 - [TensorFlow](https://github.com/tensorflow/addons/issues/179) https://github.com/tensorflow/addons/issues/179
- [PyTorch](https://github.com/pytorch/vision/pull/1586) https://github.com/pytorch/vision/pull/1586
- [MXNet](https://mxnet.incubator.apache.org/api/python/docs/api/gluon/contrib/index.html#mxnet.gluon.contrib.cnn.DeformableConvolution)

vlordier:
> But to generate it in ONNX it need to be converted from someplace. If this cause a massive accuracy gain, why no framework implemented it naively? Especially, this was published 2 years ago. To add an OP in ONNX, it needs to be available in one framework at least.

@ebarsoum Deformable Convolutions are also used to process equirectangular projections of omnidirectional images by parametrising the convolution kernel

![74554848-5fb79200-4f52-11ea-9862-8f86c4a097fb](https://user-images.githubusercontent.com/5443125/75030134-62a70b00-549b-11ea-8ee0-15f9f69cdc39.png)

More recent work [Kernel Transformer Networks for Compact Spherical Convolution](https://arxiv.org/pdf/1812.03115.pdf) 

[Distortion-Aware Convolutional Filters for Dense
Prediction in Panoramic Images](http://campar.in.tum.de/pub/tateno2018eccv/tateno2018eccv.pdf
)

[...]

PR #4783, opened 2023-01-18
add DeformConv-19

yuanyao-nv:
### Description

Add new op `DeformConv` to opset 19.

### Motivation and Context
Deformable convolution (introduced in https://arxiv.org/abs/1703.06211 and https://arxiv.org/abs/1811.11168) is a popular variant of the convolutional neural network used in situations where scale, pose, viewpoints, etc. deform parts of the image. Instead of sampling using an evenly spaced kernel, it adds offsets to the sampling locations.

Related issues: https://github.com/onnx/onnx/issues/2395, https://github.com/onnx/onnx/issues/2665

Torchvision implementation: https://pytorch.org/vision/stable/_modules/torchvision/ops/deform_conv.html


onnx/backend/test/case/node/deformconv.py
```diff
@@ -0,0 +1,129 @@
...
+
+import numpy as np
+
+import onnx
+
+from ..base import Base
+from . import expect
+
+
+class DeformConv(Base):
```
p-wysocki, line 11: Could you please provide a better coverage of operator attributes? For example, `bias` has no test associated with it.

onnx/defs/nn/defs.cc
```diff
@@ -1363,6 +1363,93 @@ output_shape can also be explicitly specified in which case pads values are auto
...
+            "pads",
+            "Padding for the beginning and end along each spatial axis, it can take any nonnegative value. "
+            "The value represents the number of pixels added to the beginning and end of the corresponding "
+            "axis. The format should be as follow [x1_begin, x2_begin, ..., x1_end, x2_end, ...], where xi_begin "
+            "is the number of pixels added at the beginning of axis `i` and xi_end, the number of pixels "
+            "added at the end of axis `i`. Default is 0 along each axis.",
+            AttributeProto::INTS,
+            OPTIONAL_VALUE)
+        .Attr(
+            "group",
```
p-wysocki, line 1444: Is the `group` attribute necessary? Torchvision implementation seems to deal with it "under the hood". https://pytorch.org/vision/main/generated/torchvision.ops.deform_conv2d.html

p-wysocki (commented):
I believe new operators now require a reference implementation as well. Reference evaluator can be found at https://github.com/onnx/onnx/tree/main/onnx/reference.

yuanyao-nv:
Hi @xadupre , can I ask whether there's a way to call the python implementation of an existing op when writing it for a new op in `onnx/reference/ops`? For example in this case, I would like to make use of the existing implementation for `GridSample` and `Conv`. Thanks in advance!

xadupre:
Page https://onnx.ai/onnx/api/reference.html shows an example which uses an implementation of new custom operator. The same mechanism can be used to replace an existing op. Is that what you are looking for?

yuanyao-nv:
I see. Thanks!
________________________________

onnx/defs/nn/defs.cc
```diff
@@ -1363,6 +1363,93 @@ output_shape can also be explicitly specified in which case pads values are auto
...
+    DeformConv,
+    19,
+    OpSchema()
+        .SetDoc(DeformConv_ver19_doc)
+        .Input(
+            0,
+            "X",
+            "Input data tensor. For 2D image data, it has shape (N, iC, H, W) where N is the batch size, "
+            "iC is the number of input channels, and H and W are the height and width. "
+            "In general, the shape is (N, iC, D1, D2, ... , Dn) for n-dimensional data, where "
```
gramalingam, line 1380: Is the general n-dimensional version used? Or, would it be enough to have just the 2D version or 2D/3D version?

onnx/defs/nn/defs.cc
yuanyao-nv: It's true it can be inferred from the dimensions of `X` and `W`. I see `Conv` is also listing it as an explicit attribute so maybe we should keep it this way for consistency? `kernel_shape` is another attribute that can be inferred from input shape.

onnx/defs/nn/defs.cc
yuanyao-nv: All applications in the original paper are 2D. I see `Conv` includes support for arbitrary dimensions. Should we be consistent?

onnx/defs/nn/defs.cc
```diff
@@ -1363,6 +1363,93 @@ output_shape can also be explicitly specified in which case pads values are auto
...
+
+ONNX_OPERATOR_SET_SCHEMA(
+    DeformConv,
+    19,
+    OpSchema()
+        .SetDoc(DeformConv_ver19_doc)
+        .Input(
+            0,
+            "X",
+            "Input data tensor. For 2D image data, it has shape (N, iC, H, W) where N is the batch size, "
```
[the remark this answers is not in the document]
yuanyao-nv: Changed input channels to `C` for consistency.

onnx/backend/test/case/node/deformconv.py
yuanyao-nv: Added some more cases.

onnx/defs/nn/defs.cc
gramalingam: While generalizing the spec of an op for higher dimensions is relatively easy, it can be non-trivial to generalize an implementation (I am commenting generally based on experience with other ops, I don't know enough about this specific op). I often see specialized implementations for 2 or 3 D, with a separate generic implementation for the generic N-D case.

In such cases, it seems to me that it is justifiable to define the 2/3 D version if that is the only one used in practice.

yuanyao-nv: @gramalingam I added a sentence to the op description and the input description saying the common use cases are 2D and 3D, after consulting internal stakeholders. Let me know if you're ok with that. Thanks!

onnx/reference/ops/op_deform_conv.py
```diff
@@ -0,0 +1,176 @@
+# SPDX-License-Identifier: Apache-2.0
+# pylint: disable=R0912,R0913,R0914,R0915,R1702,W0221
+
+import numpy as np
+
+from onnx.reference.op_run import OpRun
+
+
+def _conv_implementation(  # type: ignore
```
gramalingam, line 9: Naive question: is this an implementation of deform_conv (if so, better to rename it accordingly), or is it a common implementation (that subsumes conv also)?

yuanyao-nv: Thanks for catching. I renamed it to `_deform_conv_implementation`.

yuanyao-nv:
Hi @gramalingam @p-wysocki , do you have any more comments or questions about this PR?

gramalingam (approved):

liqunfu:
Thanks for this PR @yuanyao-nv ! feel free merging it anytime. I have 2 follow up questions on implementation of this op in ORT:
(1) I see you added 2 references to DeformConv models. I wonder is there any new models use the op after these 2 publications. I searched for any uses of pytorch DeformConv2d (https://pytorch.org/vision/main/generated/torchvision.ops.DeformConv2d.html) but did not find any use of it on internet. 
(2) Now we have many ops extended to 3D and ND, any examples of using DeformDonv for 3D data is very helpful.
There is a plan to add DeformConv CPU kernels to ORT. It will be very helpful if you have any pointers to above 2 questions. thanks!

liqunfu:
> Hi @liqunfu , to answer your questions:
> 
> 1. Besides the original papers, there are some other models mentioned in the issues I linked as well, such as Kernel Transformer Network https://arxiv.org/abs/1812.03115.
> 2. Examples of 3D case: https://arxiv.org/abs/2004.02803 https://arxiv.org/abs/1911.01898

Thank you very much @yuanyao-nv !

PR #4790, opened 2023-01-19
Add AveragePool-19 - add `dilations` attribute

p-wysocki:
Signed-off-by: p-wysocki [przemyslaw.wysocki@intel.com](mailto:przemyslaw.wysocki@intel.com)

### Description
Add AveragePool-19, which now has `dilations` attribute.

### Motivation and Context
- Partially solves https://github.com/onnx/onnx/issues/2290
image
- Related PRs: https://github.com/onnx/onnx/pull/4533 and https://github.com/onnx/onnx/pull/4534


onnx/reference/ops/op_average_pool.py
```diff
@@ -27,3 +28,89 @@ def _run(  # type: ignore
...
+            auto_pad=auto_pad,
+            ceil_mode=ceil_mode,
+            dilations=dilations,
+            kernel_shape=kernel_shape,
+            pads=pads,
+            strides=strides,
+        )
+
+
+if onnx_opset_version() > 11:
```
xadupre, line 109: It should be `>= 19`. AveragePool - 11 is used from opset 11 to 18.

xadupre:
You may remove line https://github.com/onnx/onnx/blob/main/onnx/reference/ops/_op_common_pool.py#L216 (216 to 220). They are no longer necessary. That's what causing the issue.

onnx/reference/ops/op_average_pool.py
gramalingam: Similar points for the checks below, I think?

xadupre: Yes. I removed these lines in #4800 to make it simpler.

onnx/backend/test/case/node/averagepool.py
```diff
@@ -455,3 +455,34 @@ def export_averagepool_2d_ceil() -> None:
...
+        node = onnx.helper.make_node(
+            "AveragePool",
+            inputs=["x"],
+            outputs=["y"],
+            kernel_shape=[3, 3],
+            strides=[1, 1],
+            dilations=[2, 2],
+            ceil_mode=True,
+        )
+        x = np.array(
```
xadupre, line 474: I don't know if it is better to let it that way or to replace it by `(np.arange(16) + 1).astype(np.int64).reshape((1, 1, 4, 4))`.

xadupre (approved):

onnx/backend/test/case/node/averagepool.py
p-wysocki: Good point, done.

xadupre (approved):

gramalingam (approved):

PR #4792, opened 2023-01-19
Fix attribute substitution within subgraphs during function type/shape inference

gramalingam:
### Description
Fix bug in shape inference.

### Motivation and Context
See https://github.com/microsoft/onnx-script/issues/318.

Attribute parameter substitution, during inference for a function body, does not recursively do the substitution within subgraphs of control-flow constructs.


onnx/shape_inference/implementation.cc
```diff
@@ -486,21 +486,43 @@ class ShapeInferenceImplBase {
...
-        if (attr_map.count(attr.ref_attr_name())) {
-          auto copy_attr = *attr_map[attr.ref_attr_name()];
-          copy_attr.set_name(attr.name());
-          copy_n.add_attribute()->CopyFrom(copy_attr);
+  void replaceAttrParams(NodeProto& n, std::unordered_map<std::string, const AttributeProto*> attr_map) {
+    auto& attributes = *n.mutable_attribute();
+    for (auto attr_iter = attributes.begin(); attr_iter != attributes.end();) {
+      auto& attr = *attr_iter;
+      if (!attr.ref_attr_name().empty()) {
+        // Attribute-references must be replaced by the corresponding attribute-value in the call-node
```
wschin, line 494: What is "call-node"?

onnx/shape_inference/implementation.cc
```diff
@@ -486,21 +486,43 @@ class ShapeInferenceImplBase {
...
+        // if the call-node contains the attribute. Otherwise, this attribute must be removed.
+        auto entry = attr_map.find(attr.ref_attr_name());
+        if (entry != attr_map.cend()) {
+          // Copy value of attribute, but retain original name:
+          std::string name = attr.name();
+          attr = *(entry->second);
+          attr.set_name(name);
```
wschin, line 501: Is the resulted graph still valid if the original attribute name is reused? This sounds a name conflict.

onnx/shape_inference/implementation.cc
```diff
@@ -486,21 +486,43 @@ class ShapeInferenceImplBase {
...
-    NodeProto copy_n(n);
-    // Add attribute information into the temporary node
-    copy_n.clear_attribute();
-    for (const auto& attr : n.attribute()) {
-      if (attr.has_ref_attr_name()) {
-        if (attr_map.count(attr.ref_attr_name())) {
-          auto copy_attr = *attr_map[attr.ref_attr_name()];
-          copy_attr.set_name(attr.name());
-          copy_n.add_attribute()->CopyFrom(copy_attr);
+  void replaceAttrParams(NodeProto& n, std::unordered_map<std::string, const AttributeProto*> attr_map) {
```
wschin, line 489: Why is this called "replaceAttrParams"?
What does "Param" stand for? Is it an attribute in NodeProto?

onnx/test/model_inference_test.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_mi_function_subgraph_attr(self):
+        '''Test use of functions.'''
+        model = '''
+            <
+                ir_version: 7,
+                opset_import: [ "" : 17, "local" : 1]
+            >
+            agraph (float[N] x, bool flag) => (y)
+            {
+                y = local.cast<target=6>(x, flag)
```
xadupre, line 87: It would be great to add a unit test with a graph which calls the same function twice with a different value for the parameter.

onnx/shape_inference/implementation.cc
gramalingam: Consider the line `y = Cast<to:int = @target> (x)` from the test-case in this PR. Here, `target` is an attribute-parameter of the containing function. After we do substitution, we should create an attribute with name `to` and a value equal to the value of `target`. That's what we are doing.

onnx/shape_inference/implementation.cc
gramalingam: It is a node representing a call to this function. For example, in our test-case, it is the node in the main graph representing `y = local.cast<target=6>(x)`. It specifies the attribute-value for `target`, which is used in the replacement.

onnx/shape_inference/implementation.cc
jcwchen: Also confused about Params here. How about calling `replaceAttrValues`?

onnx/shape_inference/implementation.cc
```diff
@@ -486,21 +486,43 @@ class ShapeInferenceImplBase {
...
         }
-      } else {
-        copy_n.add_attribute()->CopyFrom(attr);
       }
+      // Subgraphs must be recursively processed.
+      if (attr.has_g()) {
+        replaceAttrParams(*attr.mutable_g(), attr_map);
+      }
+      for (auto& graph : *attr.mutable_graphs())
+        replaceAttrParams(graph, attr_map);
```
jcwchen, line 512: nit:
```suggestion
    for (auto& graph : *attr.mutable_graphs()) {
        replaceAttrParams(graph, attr_map);
    }
```

gramalingam: re. the name Params: I find `replaceAttrValues` confusing. I could change to `replaceAttrRefs` I suppose. The point is that an Attribute inside a function-body could contain an attribute-value or it could be an attribute-parameter of the function itself. We want to replace the references to attribute-parameters by the values.

onnx/shape_inference/implementation.cc
gramalingam: An ONNX function has parameters of two kinds (just like an ONNX ops): the regular parameters (tensors, etc., whose values will be known only at runtime) and attribute-parameters ... "Param" here is shorthand for "Parameter". As discussed elsewhere, I changed the name to "AttrRefs", though I think both names should be fine.

jcwchen (approved):
LGTM. Thanks for fixing this!

PR #4793, opened 2023-01-19
Pad-19: Add 'wrap' mode

yuanyao-nv:

### Description
Add one more option, "wrap", to the "mode" attribute of Pad.

### Motivation and Context
The convolution operator in PyTorch has circular padding mode. Adding this to ONNX would allow us to more easily export from PyTorch, and also bring ONNX into better alignment with `numpy.pad` and `torch.nn.functional.pad`.
https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html
https://numpy.org/doc/stable/reference/generated/numpy.pad.html
https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html


daquexian:
Not sure if this change is necessary. It seems we can always add a Pad op before the Conv op to handle the `padding_mode`. (Currently Pad op does not support 'circular' mode but we can easily add it)

yuanyao-nv:
> Not sure if this change is necessary. It seems we can add a Pad op before the Conv op to handle non-default PyTorch `padding_mode`. (Currently Pad op does not support 'circular' mode but we can easily add it)

Good point. I changed this PR to modify the Pad operator instead.

gramalingam:
Looks good to me, thanks!

One minor suggestion: can we create a file, say utils.h and/or utils.cc, for common code used by both old.cc and defs.cc and avoid duplicating it in both? This may help future PRs (eg., fixing bugs in shape-inference or improving the documentation). It seems to me much of the code is common for the two opschemas, except for the documentation/attribute. May be a function with two parameters for this could do it. Eg., something like https://github.com/onnx/onnx/blob/main/onnx/defs/math/utils.cc#L9

yuanyao-nv:
> Looks good to me, thanks!
> 
> One minor suggestion: can we create a file, say utils.h and/or utils.cc, for common code used by both old.cc and defs.cc and avoid duplicating it in both? This may help future PRs (eg., fixing bugs in shape-inference or improving the documentation). It seems to me much of the code is common for the two opschemas, except for the documentation/attribute. May be a function with two parameters for this could do it. Eg., something like https://github.com/onnx/onnx/blob/main/onnx/defs/math/utils.cc#L9

Hi Rama, what exactly are you suggesting we factor out? Are you suggesting that the different opsets of Pad share a common DocGenerator? It seems like while opsets 18 and 19 share quite a bit in common, the opsets prior 18 differ in several places in the doc/attribute/input. I'm not sure if it's easy to refactor all of them.

gramalingam:
Hi @yuanyao-nv : I mean just the opsets 18 and 19, not any of the older ones.

yuanyao-nv:
> Hi @yuanyao-nv : I mean just the opsets 18 and 19, not any of the older ones.

@gramalingam thanks for clarifying. I've updated it.

gramalingam (approved):
LGTM, thanks!

One question to think about is whether we want to also update the Conv op, keeping both ops consistent in terms of supported padding modes. Any thoughts about it? Those changes can always be done separately in a later PR, just raising the question to get people's thoughts about it.

yuanyao-nv:
> LGTM, thanks!
> 
> One question to think about is whether we want to also update the Conv op, keeping both ops consistent in terms of supported padding modes. Any thoughts about it? Those changes can always be done separately in a later PR, just raising the question to get people's thoughts about it.

The original motivation was to allow for exportation from PyTorch's Conv. This change should sufficiently allow for that in a clean way.

My take on adding it to `Conv` directly is that one might then feel compelled to add a padding mode to other Conv-related ops such as `ConvInteger` and `ConvTranspose`, leading to a proliferation of attributes.

gramalingam:
> The original motivation was to allow for exportation from PyTorch's Conv. This change should sufficiently allow for that in a clean way.
> 
> My take on adding it to `Conv` directly is that one might then feel compelled to add a padding mode to other Conv-related ops such as `ConvInteger` and `ConvTranspose`, leading to a proliferation of attributes.

Sounds good to me, thanks!

PR #4795, opened 2023-01-20
fix version_conversion error of Pad-2 to Pad-11 caused by missing value attribu…

Tee0125:
…te which is not mandatory

### Description

not to fail version_conversion, create `value` attribute with default value 0.f when `value` attribute is not present for `Pad-2` operator.

### Motivation and Context
`value` of `Pad-2` is optional, so missing `value` attribute is not a problem but if `value` attribute is not present, version conversion from 2 to 10 is failing.

test code is here:

```
import onnx

from onnx import version_converter
from onnx.helper import make_tensor, make_tensor_value_info
from onnx.helper import make_model, make_graph, make_node


input_shape = (1, 3, 2, 4)
output_shape = (1, 3, 5, 7)

input_type = onnx.TensorProto.FLOAT
output_type = onnx.TensorProto.FLOAT

input_tensors = [make_tensor_value_info('A', input_type, input_shape)]
output_tensors = [make_tensor_value_info('B', output_type, output_shape)]

node = make_node('Pad',  inputs=['A'], outputs=['B'], pads=(0, 0, 2, 1, 0, 0, 1, 2))

graph = make_graph([node], 'graph', input_tensors, output_tensors)

model = make_model(graph)
model.opset_import[0].version = 2

new_model = convert_version(model, 11)
onnx.checker.check_model(new_model)
```

### DIFF
--- a/onnx/version_converter/adapters/pad_10_11.h
+++ b/onnx/version_converter/adapters/pad_10_11.h
@@ -27,6 +27,8 @@ class Pad_10_11 final : public Adapter {
     node->removeAttribute(kpads);
     // Turn value attribute into input
     if (!node->hasAttribute(kmode) || node->s(kmode) == "constant") {
+      if (!node->hasAttribute(kvalue))
+        node->f_(kvalue, 0.);
       Tensor t_value;
       t_value.elem_type() = TensorProto_DataType_FLOAT;
       auto& data_value = t_value.floats();
--- a/onnx/test/version_converter_test.py
+++ b/onnx/test/version_converter_test.py
@@ -1835,6 +1835,22 @@ class TestVersionConverter(unittest.TestCase):
         assert converted_model.opset_import[0].version == to_opset
         assert len(converted_model.graph.node[0].attribute) == 1
 
+    # Test Pad Adapter: 10 -> 11
+    def test_pad_10_11(self) -> None:
+        pads = (0, 1, 2, 0, 2, 1)
+        nodes = [helper.make_node("Pad", ["X"], ["Y"], pads=pads)]
+        graph = helper.make_graph(
+            nodes,
+            "test",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (1, 2, 2))],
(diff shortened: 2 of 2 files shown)


xadupre:
Should not we add a unit test to cover the case?

Tee0125:
Unittest is added. :)

daquexian (approved):
LGTM. Could you please add another test with pad value specified?

Tee0125:
value specified Pad test is also added. 8)

onnx/version_converter/adapters/pad_10_11.h
```diff
@@ -27,6 +27,8 @@ class Pad_10_11 final : public Adapter {
     node->removeAttribute(kpads);
     // Turn value attribute into input
     if (!node->hasAttribute(kmode) || node->s(kmode) == "constant") {
+      if (!node->hasAttribute(kvalue))
```
gramalingam, line 30: nit: I think it would be simpler to change line 35 to
```cpp
data_value.emplace_back(node->hasAttribute(kvalue) ? node->f(kvalue) : 0.);
```
instead of inserting the attribute and then deleting it again in line 40.

Tee0125: If apply your suggestion, `removeAttribute(kvalue)` makes error because there's no `value` attribute. So line 40 should also be changed like in below:

```
if (node->hasAttribute(kvalue)
    node->removeAttribute(kvalue);
```

Which do you prefer? inserting and deleting value attribute or checking `hasAttribute(kvalue)` twice?

gramalingam: Makes sense. Let's go ahead merge this as-is, the difference is not important enough to iterate more.

jcwchen:
Please allow me to reopen this PR for rebasing.

jcwchen (approved):
It still shows `This branch is out-of-date with the base branch` so it cannot be merged... @Tee0125 could you please try to manually merge with the main branch from your end? Thanks!

PR #4805, opened 2023-01-25
Introduce float 8 types, FLOATE4M3, FLOATE5M2

xadupre:
### Description

Introduces four new types for quantization / computation to speed up deep learning models on GPU.

First pair for NVIDIA:

* FLOATE8E4M3FN: float 8 bits with 4 bits for the exponent and 3 for the mantissa, usually used for the coefficients, support nan values
* FLOATE8E5M2: float 8 bits with 5 bits for the exponent and 2 for the mantissa, usually used for the gradients, supports nan values and infinities

Second pair for GraphCore:

* FLOATE8E4M3FNUZ: float 8 bits with 4 bits for the exponent and 3 for the mantissa, usually used for the coefficients, only (no infinities)
* FLOATE8E5M2FNUZ: float 8 bits with 5 bits for the exponent and 2 for the mantissa, usually used for the gradients, supports nan values only (no infinities)

Suffix FN means no infinities, UZ means the negative zero is used to represent nan values.

The PR modifies operator Cast, CastLike, QuantizeLinear, DequantierLinear to make them support the four new types. It adds function to cast from/to float32.

### Motivation and Context

Latest NVIDIA, Arm, Inter, GraphCore introduces float 8 for faster computation.

* [FP8 FORMATS FOR DEEP LEARNING](https://arxiv.org/pdf/2209.05433.pdf): specifications for Nvidia, Arm, Intel
* [8-bit Numerical Formats For Deep Neural Networks](https://arxiv.org/pdf/2206.02915.pdf) (GraphCore)

Other related papers:

* [FP8 Formats for Deep Learning](https://arxiv.org/abs/2209.05433)
* [RFC: FP8 in XLA](https://github.com/openxla/xla/discussions/22)


linkerzhang (commented):
Are these "new" types common enough to be added here? or they mey be added internally in ORT?

docs/IR.md
```diff
@@ -392,7 +392,7 @@ It is common to represent a tensor as a nested list. This generally works fine,
 
 |Group|Types|Description|
 |---|---|---|
-Floating Point Types|float16, float32, float64|Values adhering to the IEEE 754-2008 standard representation of floating-point data.
+Floating Point Types|float16, float32, float64, bfloat16, floate4m3, floate5m2|Values adhering to the IEEE 754-2008 standard representation of floating-point data or defined in paper [FP8 Formats for Deep Learning](https://arxiv.org/abs/2209.05433)
```
askhade, line 395: Reference the GraphCore paper as well? Since the handling of spcial values will be treated as an implementation detail.

xadupre: You mean this paper: https://arxiv.org/pdf/2206.02915.pdf. The float8 number implemented are slightly different. There is no (0 and -0) as opposed to the solution chosen by NVIDIA. I was afraid that quoting two papers may confuse readers.

onnx/defs/quantization/defs.cc
```diff
@@ -93,9 +97,9 @@ ONNX_OPERATOR_SET_SCHEMA(
             static_cast<int64_t>(1))
         .TypeConstraint(
             "T",
-            {"tensor(int8)", "tensor(uint8)", "tensor(int32)"},
+            {"tensor(int8)", "tensor(floate4m3)", "tensor(floate5m2)", "tensor(uint8)", "tensor(int32)"},
             "Constrain 'x_zero_point' and 'x' to 8-bit/32-bit integer tensor.")
-        .SetDoc(DequantizeLinear_ver13_doc)
+        .SetDoc(DequantizeLinear_ver19_doc)
         .TypeAndShapeInferenceFunction([](ONNX_NAMESPACE::InferenceContext& ctx) {
           auto y_type = ctx.getOutputType(0);
```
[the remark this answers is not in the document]
xadupre: It should. I looked into CUDA API and it does not support the conversion from float to float8, only float16 to float8.

onnx/defs/quantization/defs.cc
```diff
@@ -42,9 +44,9 @@ ONNX_OPERATOR_SET_SCHEMA(
         .TypeConstraint("T1", {"tensor(float)", "tensor(int32)"}, "Constrain 'x' to float or int32 tensor.")
         .TypeConstraint(
             "T2",
-            {"tensor(int8)", "tensor(uint8)"},
+            {"tensor(int8)", "tensor(floate4m3)", "tensor(floate5m2)", "tensor(uint8)"},
```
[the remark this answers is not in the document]
xadupre: done

onnx/defs/quantization/defs.cc
```diff
@@ -7,17 +7,19 @@
 
 namespace ONNX_NAMESPACE {
 
-static const char* QuantizeLinear_ver13_doc = R"DOC(
+static const char* QuantizeLinear_ver19_doc = R"DOC(
 The linear quantization operator. It consumes a high precision tensor, a scale, and a zero point to compute the low precision / quantized tensor.
 The scale factor and zero point must have same shape, and can be either a scalar for per-tensor / per layer quantization, or a 1-D tensor for per-axis quantization.
 The quantization formula is y = saturate ((x / y_scale) + y_zero_point).
 For saturation, it saturates to [0, 255] if it's uint8, or [-128, 127] if it's int8.
```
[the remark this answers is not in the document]
xadupre: done

onnx/defs/quantization/defs.cc
```diff
@@ -59,17 +61,19 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         }));
 
-static const char* DequantizeLinear_ver13_doc = R"DOC(
+static const char* DequantizeLinear_ver19_doc = R"DOC(
 The linear dequantization operator. It consumes a quantized tensor, a scale, and a zero point to compute the full precision tensor.
 The dequantization formula is `y = (x - x_zero_point) * x_scale`. `x_scale` and `x_zero_point` must have same shape, and can be either a scalar
 for per-tensor / per layer quantization, or a 1-D tensor for per-axis quantization.
 `x_zero_point` and `x` must have same type. `x` and `y` must have same shape. In the case of dequantizing int32,
 there's no zero point (zero point is supposed to be 0).
+`y_zero_point` is not used for quantization to floate4m3 or floate5m2, only the scale but
```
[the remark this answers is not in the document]
xadupre: done

onnx/helper.py
```diff
@@ -334,6 +334,99 @@ def float32_to_bfloat16(fval: float, truncate: bool = False) -> int:
     return (ival + rounded) >> 16
 
 
+def float32_to_floate4m3(fval: float, scale: float = 1.0) -> int:
+    """
+    Convert a float32 value to a floate4m3 (as int).
+
+    :param fval: float to convert
+    :param scale: scale
```
askhade, line 342: Add some description for scale and also explaination for default value 1.0 basically when not specified fp32 will be truncated to fp8.

onnx/defs/quantization/defs.cc
```diff
@@ -39,12 +43,15 @@ ONNX_OPERATOR_SET_SCHEMA(
             "(Optional) The axis of the quantization dimension of the input tensor. Ignored for per-tensor quantization. Negative value means counting dimensions from the back. Accepted range is [-r, r-1] where r = rank(input).",
             AttributeProto::INT,
             static_cast<int64_t>(1))
-        .TypeConstraint("T1", {"tensor(float)", "tensor(int32)"}, "Constrain 'x' to float or int32 tensor.")
+        .TypeConstraint(
+            "T1",
+            {"tensor(float)", "tensor(float16)", "tensor(bfloat16)", "tensor(int32)"},
+            "Constrain 'x' to float or int32 tensor.")
```
gramalingam, line 49: Nit: comment doesn't match updated type constraint.

onnx/defs/quantization/defs.cc
```diff
@@ -39,12 +43,15 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         .TypeConstraint(
             "T2",
-            {"tensor(int8)", "tensor(uint8)"},
+            {"tensor(int8)", "tensor(uint8)", "tensor(floate4m3)", "tensor(floate5m2)"},
```
gramalingam, line 52: Message/comment below needs to be updated too `8-bit integer/float`

onnx/defs/quantization/defs.cc
```diff
@@ -59,43 +66,49 @@ ONNX_OPERATOR_SET_SCHEMA(
...
         }));
 
-static const char* DequantizeLinear_ver13_doc = R"DOC(
+static const char* DequantizeLinear_ver19_doc = R"DOC(
 The linear dequantization operator. It consumes a quantized tensor, a scale, and a zero point to compute the full precision tensor.
 The dequantization formula is `y = (x - x_zero_point) * x_scale`. `x_scale` and `x_zero_point` must have same shape, and can be either a scalar
 for per-tensor / per layer quantization, or a 1-D tensor for per-axis quantization.
 `x_zero_point` and `x` must have same type. `x` and `y` must have same shape. In the case of dequantizing int32,
 there's no zero point (zero point is supposed to be 0).
+`y_zero_point` is not used for quantization to floate4m3 or floate5m2,
```
gramalingam, line 75: May be better to say "Since zero-point is not used in the case of floate4m3 or floate5m2 quantization"

docs/docsgen/source/technical/float8.rst
```diff
@@ -0,0 +1,99 @@
...
+The float precision is much lower but the training precision
+does not suffer too much.
+
+`FP8 Formats for Deep Learning <https://arxiv.org/abs/2209.05433>`_
+from NVIDIA introduces two types following
+`IEEE specifciations <https://en.wikipedia.org/wiki/IEEE_754>`_.
+First one is E4M3, 1 bit for the sign, 4 bits for the exponents and 3
+bits for the mantissa. Second one is E5M2, 1 bit for the sign,
+3 bits for the exponents and 2 for the mantissa. The first types
+is mostly used for the coefficients, the second one for the gradient.
```
gramalingam, line 25: Is "coefficient" the right word? Is it supposed to be for weights?

xadupre: yes, in French, they are synonyms. I'll change that.

onnx/defs/quantization/defs.cc
xadupre: done

onnx/helper.py
xadupre: The function documentation now refers to a page in the main documentation.

docs/docsgen/source/technical/float8.rst
xadupre: done

prasanthpul:
for better readability, instead of FLOATE4M3 and FLOATE5M2, I suggest naming them FLOAT8_E4M3 and FLOAT8_E5M2 (or FLOAT8E4M3 and FLOAT8E5M2 if we don't want underscore). Otherwise the fact that it is FP8 is lost in the naming. HLO, for example, calls their types f8E5M2 and f8E4M3

xadupre:
Which name do you prefer FLOAT8_E4M3 or F8E4M3?

gramalingam:
> Which name do you prefer FLOAT8_E4M3 or F8E4M3?

I personally prefer the FLOAT8_Exxx variants, as they are more in line with FLOAT16. However, I suggest you wait a bit in case Prasanth or anyone else has any other suggestions.

I notice that [Protobuf enums](https://protobuf.dev/programming-guides/proto/#enum) allow "aliases". We could potentially rationalize the naming further by adding aliases, and we can make these aliases following a uniform naming. Eg., creating aliases like `FLOAT64` or `FP64` for `DOUBLE` and so on.

docs/Changelog.md
```diff
@@ -22622,6 +22622,208 @@ This version of the operator has been available since version 19 of the default
...
+
+  Conversion from a numerical type to any numerical type is always allowed.
+  User must be aware of precision loss and value change caused by range difference between two types.
+  For example, a 64-bit float 3.1415926459 may be round to a 32-bit float 3.141592. Similarly, converting
+  an integer 36 to Boolean may produce 1 because we truncate bits which can't be stored in the targeted type.
+
+  In more detail, the conversion among numerical types should follow these rules:
+
+  * Casting from floating point to:
+    * floating point: +/- infinity if OOR (out of range).
```
georgen117, line 22650: What happens when casting FP32->Floatm4e3 and the value is OOR?

Floatm4e3 does not have infinities. Does it saturate?

docs/docsgen/source/technical/float8.rst
```diff
@@ -0,0 +1,105 @@
...
+Papers
+======
+
+Two papers have been published in 2022 to introduce floats
+stored on a byte as opposed to float 32 stored on 4 bytes.
+The float precision is much lower but the training precision
+does not suffer too much.
+
+`FP8 Formats for Deep Learning <https://arxiv.org/abs/2209.05433>`_
+from NVIDIA introduces two types following
```
georgen117, line 20: It looks like the paper was jointly authored by NVIDIA, Intel, and ARM. 
see: https://developer.nvidia.com/blog/nvidia-arm-and-intel-publish-fp8-specification-for-standardization-as-an-interchange-format-for-ai/
and: https://www.intel.com/content/www/us/en/newsroom/news/new-hardware-specs-speed-ai-software-development.html

docs/Changelog.md
georgen117: Reading through the "FP9 Formats For Deep Learning" paper is states. 

> For conversions to E4M3 this means that both infinities and NaNs in the wider type (for example, single
precision) turn into NaNs in FP8.

That seems to indicate that OOR values would be converted to NaNs for FP8_E4M3.

awf:
Would it be good to follow the way FP8 has been introduced to LLVM/MLIR and hence XLA.  (See e.g. [https://reviews.llvm.org/D143744](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Freviews.llvm.org%2FD143744&data=05%7C01%7Cxadupre%40microsoft.com%7Cf4db8307ed884af91e5e08db1b2e1a4c%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638133658534701238%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000%7C%7C%7C&sdata=QafIxBBxIFVDyVc39USsTpX6Z%2FlGH06j0xUdz%2FMiPYw%3D&reserved=0)).  The approach there is to have four FP8 datatypes, named
```
Float8E5M2FNUZ 
Float8E4M3FNUZ
Float8E5M2
Float8E4M3FN 
``` 
With semantics defined e.g. at [https://llvm.org/doxygen/APFloat_8cpp_source.html#l00130](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fllvm.org%2Fdoxygen%2FAPFloat_8cpp_source.html%23l00130&data=05%7C01%7Cxadupre%40microsoft.com%7Cf4db8307ed884af91e5e08db1b2e1a4c%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638133658534701238%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000%7C%7C%7C&sdata=5UsTyVx2cb9DrZ0pqEvf%2FOeEY6reGyhuTS7iNwDeqSU%3D&reserved=0)

xadupre:
I followed your suggestion. I still need to complete the conversion from float to the two new types.

docs/docsgen/source/technical/float8.rst
xadupre: updated

onnx/defs/quantization/defs.cc
xadupre: done

docs/Changelog.md
awf: In general, I suggest that such semantics are drawn from the LLVM implementation by default. Even if those disagree with the papers, they are at least a consistent set, which we are reasonably confident match the various hardware available today
https://llvm.org/doxygen/APFloat_8cpp_source.html#l00130

georgen117:
> > Which name do you prefer FLOAT8_E4M3 or F8E4M3?
> 
> I personally prefer the FLOAT8_Exxx variants, as they are more in line with FLOAT16. However, I suggest you wait a bit in case Prasanth or anyone else has any other suggestions.
> 
> I notice that [Protobuf enums](https://protobuf.dev/programming-guides/proto/#enum) allow "aliases". We could potentially rationalize the naming further by adding aliases, and we can make these aliases following a uniform naming. Eg., creating aliases like `FLOAT64` or `FP64` for `DOUBLE` and so on.

I agree FLOAT8_Exxx variants fit inline with the other data types the best and is most readable. I would also be good with FP8_Exxx but this does not fit in with the other naming conventions so I think it is best avoided.

justinchuby:
nit: docs infinitie -> infinite?

onnx/backend/test/cmd_tools.py
```diff
@@ -86,11 +86,14 @@ def prepare_dir(path: str) -> None:
                             assert case.model.graph.input[j].type.HasField(
                                 "tensor_type"
                             )
-                            f.write(
-                                numpy_helper.from_array(
-                                    input, case.model.graph.input[j].name
-                                ).SerializeToString()
-                            )
+                            if isinstance(input, TensorProto):
```
gramalingam, line 89: I guess this is because numpy doesn't natively support 8-bit floats? I wonder if we need something similar to what is done for bfloat16 (use some 8-bit view type from numpy) for inter-operation with numpy?

onnx/defs/schema.h
```diff
@@ -686,6 +707,16 @@ class OpSchema final {
     return all_tensor_types_with_bfloat;
   }
 
+  static const std::vector<std::string>& all_tensor_types_with_bfloat_float8() {
```
gramalingam, line 710: I wonder if the sequence types need to be updated too.

onnx/onnx-ml.proto
```diff
@@ -547,7 +559,7 @@ message TensorProto {
   // float16 values must be bit-wise converted to an uint16_t prior
```
gramalingam, line 559: Add `float8` to previous line? And add a similar line like this for `float8` ?

onnx/backend/test/cmd_tools.py
xadupre: True. Numpy does not support blfoat16 or float8. You can see how i modified the reference implementation to use custom dtype. I tried to  make it simple and intuitive. Feel free to tell it is not. Finding the most simple way to implement things is not always easy.

onnx/defs/schema.h
xadupre: If we agree float8 are only limited to a small subset of operators, i assumed sequence would not be part of it.

rajeevsrao:
The [FP8 Formats for Deep Learning](https://arxiv.org/abs/2209.05433) paper does not clearly outline the behavior for handling overflows in FP8 conversion, and is inconsistent in a few places:
- “Converting a special value from a wider precision to FP8 results in the corresponding special value in FP8. For conversions to E4M3 this means that both infinities and NaNs in the wider type (for example, single precision) turn into NaNs in FP8”.
- “Values that overflow are then saturated to the maximum representable value.”

We sought clarification from the authors of the paper, and the following is a summary.

When converting higher-precision types to FP8, first rounding is applied to reduce the mantissa bit count to the destination format, after that if the rounded magnitude:
- is above the maximum destination magnitude:
  - if using saturating conversion mode: generate the max normal FP8 magnitude
  - if using non-saturating mode:
    - E5M2 destination: generate an infinity
    - E4M3 destination: generate a NaN (since there is no encoding for infinity)
- is in the representable range of FP8: corresponding FP8 exponent bits are generated

Recommendation for ONNX:
**Saturation on overflows is always preferred for model weights, and for Cast/QuantizeLinear in inference.**

We propose that ONNX specification support saturating overflow at the least (as default behavior). i.e.,
- fp32/fp16 NaN -> NaN
- fp32/fp16 inf -> max_norm
- (fp32/fp16 value -> max_norm) -> max_norm

justinchuby:
Saw: https://github.com/jax-ml/ml_dtypes

onnx/onnx-ml.proto
xadupre: done

xadupre:
done

rajeevsrao:
> I modified the PR to cast inf into max_value for E4M3FN, E4M3FNUZ, E5M2FNUZ. The behaviour is different from https://github.com/jax-ml/ml_dtypes which converts this value into nan. I'll check what NVIDIA does.

@xadupre NVidia recommendation is to also saturate (inf->max_norm) for E5M2. Saturation to fp8_max is not limited to conversion of infinity - we saturate whenever we overflow.

Should we add fp8 conversion tests for all cases (you added tests for infs `test_float8_inf`)?
- fp32/fp16 NaN -> NaN
- fp32/fp16 inf -> max_norm
- (fp32/fp16 value > max_norm) -> max_norm

Thanks.

georgen117:
> We propose that ONNX specification support saturating overflow at the least (as default behavior). i.e.,
> 
> * fp32/fp16 NaN -> NaN
> * fp32/fp16 inf -> max_norm
> * (fp32/fp16 value -> max_norm) -> max_norm

I missed that this was proposing Inf -> max_norm.  What is the use case of converting Inf -> max_norm?

Shouldn't it be:
Using saturation: (proposed default behavior)
fp32/fp16 NaN -> NaN
fp32/fp16 inf -> inf (5m2e) or NaN for (FN variants)
fp32/fp16 Out of Bound Value -> max_norm

non-saturating:
fp32/fp16 NaN -> NaN
fp32/fp16 inf -> inf (5m2e) or NaN for (FN variants)
fp32/fp16 Out of Bound Value -> inf (5m2e) or NaN for (FN variants)

Please don't change the PR based on this comment I am trying to understand the logic and use case.

xadupre:
I did not know how intel would implement the conversion of float32(inf) into e4m3fn. It makes sense to choose nan otherwise onnx would define something no hardware does. I'll revert the saturation i did today. I suggest I update the quantize operator to enable the use of zero point for float8. It seems more intuitive to follow the same logic whatever the quantized type is.

rajeevsrao:
> I did not know how intel would implement the conversion of float32(inf) into e4m3fn. It makes sense to choose nan otherwise onnx would define something no hardware does. I'll revert the saturation i did today. I suggest I update the quantize operator to enable the use of zero point for float8. It seems more intuitive to follow the same logic whatever the quantized type is.

@xadupre saturation by definition means we want to clip the overflow values during conversion and max_norm is the natural choice. Casting them to NaNs defeats the purpose (as the NaNs propagate across the network). Mapping inf -> inf can still produce NaN in the network - ex: softmax(inf) = NaN which is why we want to clip all values (except NaN) to max_norm in saturation mode.

rajeevsrao:
If non-saturation mode is required for inference (not the case for NVidia HW), we should introduce that as a separate mode (Cast/Q/DQ attribute), but default to saturation for inference.

xadupre:
The PR is casting overflow to max_value but I think we should cast inf into NaN as specified in this paper [FP8 FORMATS FOR DEEP LEARNING](https://arxiv.org/pdf/2209.05433.pdf): **Values that overflow are then saturated to the maximum representable value. [...] For conversions to E4M3 this means that both infinities and NaNs in the wider type (for example, single precision) turn into NaNs in FP8.** That's the choice by NVidia, Arm, Intel. Runtime based on their API but to implement this behaviour. I think onnx should follow the same specifications.

rajeevsrao:
> The PR is casting overflow to max_value but I think we should cast inf into NaN as specified in this paper [FP8 FORMATS FOR DEEP LEARNING](https://arxiv.org/pdf/2209.05433.pdf): **Values that overflow are then saturated to the maximum representable value. [...] For conversions to E4M3 this means that both infinities and NaNs in the wider type (for example, single precision) turn into NaNs in FP8.** That's the choice by NVidia, Arm, Intel. Runtime based on their API but to implement this behaviour. I think onnx should follow the same specifications.

@xadupre can we please discuss this in the SIG meeting? I am representing NVidia's official position and we can have the authors of that paper and experts to clarify some of the inconsistencies in the paper.

gramalingam:
> 
> @xadupre can we please discuss this on a call (since the SIG meeting was cancelled)? I am representing NVidia's official position and we can have the authors of that paper and some experts to clarify some of the inconsistencies in the paper.

Yes, the operator sig meeting next week has been cancelled. I think a separate meeting would be better so that this PR can be merged in for the upcoming release.

xadupre:
> @xadupre can we please discuss this on a call (since the SIG meeting was cancelled)? I am representing NVidia's official position and we can have the authors of that paper and some experts to clarify some of the inconsistencies in the paper.

Sure.

rajeevsrao:
> > @xadupre can we please discuss this on a call (since the SIG meeting was cancelled)? I am representing NVidia's official position and we can have the authors of that paper and some experts to clarify some of the inconsistencies in the paper.
> 
> Sure.

Thanks - @yuanyao-nv will help coordinate. If anyone else would like to be added please let us know.

onnx/defs/tensor/defs.cc
```diff
@@ -37,11 +37,11 @@ an integer 36 to Boolean may produce 1 because we truncate bits which can't be s
 In more detail, the conversion among numerical types should follow these rules:
 
 * Casting from floating point to:
-  * floating point: +/- infinity if OOR (out of range).
+  * floating point: +/- highest value if OOR (out of range, that includes float8 infinities).
   * fixed point: undefined if OOR.
   * bool: +/- 0.0 to False; all else to True.
 * Casting from fixed point to:
-  * floating point: +/- infinity if OOR. (+ infinity in the case of uint)
+  * floating point: +/- highest value if OOR (out of range, float8 infinities and infinity for uint8)
```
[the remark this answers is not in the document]
xadupre: See last comment.

xadupre:
Following table describes the rules for the conversion from float to float8. A new parameter was added to Cast, CastLike, QuantizeLinear to saturate (inf -> MAX VALUE) or not (inf -> nan).

=========== ========= ========== ========= ===========
Src value   E4M3FN    E4M3FNUZ   E5M2      E5M2FNUZ
=========== ========= ========== ========= ===========
0           0         0          0         0
NaN         NaN       NaN        NaN       NaN
Inf         FLT_MAX   NaN        FLT_MAX   NaN
> FLT_MAX   FLT_MAX   FLT_MAX    FLT_MAX   FLT_MAX
< FLT_MIN   0         0          0         0
else        RNE       RNE        RNE       RNE
=========== ========= ========== ========= ===========

The behavior changes if the parameter 'saturate' is set to False.
The rules then become:

=========== ======== ========== ====== ===========
Src Value   E4M3FN   E4M3FNUZ   E5M2   E5M2FNUZ
=========== ======== ========== ====== ===========
0           0        0          0      0
NaN         NaN      NaN        NaN    NaN
Inf         NaN      NaN        Inf    NaN
> FLT_MAX   NaN      NaN        Inf    NaN
< FLT_MIN   0        0          0      0
else        RNE      RNE        RNE    RNE
=========== ======== ========== ====== ===========

docs/docsgen/source/technical/float8.rst
```diff
@@ -0,0 +1,168 @@
...
+
+.. list-table:: Float8 types values
+   :widths: 10 10 10
+   :header-rows: 1
+
+   * - 
+     - E4M3FN
+     - E5M2
+   * - exponent :math:`\neq` 0
+     - :math:`(-1)^S 2^{\sum_{i=3}^6 b_i 2^{i-3} - 7} \left( 1 + \sum_{i=0}^2 b_i 2^{i-3} \right)`
```
yuanyao-nv, line 98: I don't know if it's just me, but these Latex equations don't seem to be rendered properly in the generated file.

onnx/defs/tensor/defs.cc
```diff
@@ -34,33 +34,78 @@ User must be aware of precision loss and value change caused by range difference
...
 * Casting from bool to:
   * floating point: `{1.0, 0.0}`.
   * fixed point: `{1, 0}`.
   * bool: no change.
+
+Float 8 type were introduced to speed up the training of
+deep models. By default the conversion of a float *x* obeys
+to the following rules. `[x]` means the rounded value.
+
+=============== ========= ========== ========= ===========
```
yuanyao-nv, line 60: I think we need to create tables in md files like this: https://www.markdownguide.org/extended-syntax/

docs/docsgen/source/technical/float8.rst
xadupre: The syntax is RST, it will properly render here: https://onnx.ai/onnx/index.html.

awf (commented):
Some text/typo suggestions.

onnx/defs/tensor/defs.cc
```diff
@@ -34,33 +34,78 @@ User must be aware of precision loss and value change caused by range difference
...
     signed types). For example, 200 (int16) -> -56 (int8).
   * bool: zero to False; nonzero to True.
 * Casting from bool to:
   * floating point: `{1.0, 0.0}`.
   * fixed point: `{1, 0}`.
   * bool: no change.
+
+Float 8 type were introduced to speed up the training of
+deep models. By default the conversion of a float *x* obeys
+to the following rules. `[x]` means the rounded value.
```
awf, line 58: We may need to expand this a little, e.g. "the value rounded to the target mantissa width"

onnx/checker.cc
```diff
@@ -967,7 +971,7 @@ void check_model(const ModelProto& model, CheckerContext& ctx) {
     fail_check("The model does not have an ir_version set properly.");
   }
   if (model.ir_version() > IR_VERSION) {
-    fail_check("Your model ir_version is higher than the checker's.");
+    fail_check("Your model ir_version ", model.ir_version(), " is higher than the checker's (", IR_VERSION, ").");
```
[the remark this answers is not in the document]
xadupre: Thanks.

onnx/defs/tensor/defs.cc
xadupre: done

docs/docsgen/source/technical/float8.rst
georgen117: Unfortanatly github lacks support for RST :math:.

For markdown files math latex syntax with $ will create render the latex. I think this is also works for RST files in github. 

example $S.b_6 b_5 b_4 b_3 b_2 b_1 b_0$

onnx/defs/quantization/defs.cc
```diff
@@ -7,25 +7,29 @@
...
 
-static const char* QuantizeLinear_ver13_doc = R"DOC(
+static const char* QuantizeLinear_ver19_doc = R"DOC(
 The linear quantization operator. It consumes a high precision tensor, a scale, and a zero point to compute the low precision / quantized tensor.
 The scale factor and zero point must have same shape, and can be either a scalar for per-tensor / per layer quantization, or a 1-D tensor for per-axis quantization.
-The quantization formula is y = saturate ((x / y_scale) + y_zero_point).
+The quantization formula is `y = saturate ((x / y_scale) + y_zero_point)`.
 For saturation, it saturates to [0, 255] if it's uint8, or [-128, 127] if it's int8.
-For (x / y_scale), it's rounding to nearest ties to even. Refer to https://en.wikipedia.org/wiki/Rounding for details. 'y_zero_point' and 'y' must have same type.
+For (x / y_scale), it's rounding to nearest ties to even. Refer to https://en.wikipedia.org/wiki/Rounding for details.
```
yuanyao-nv, line 15: nit: I think it should read like this:
```suggestion
For (x / y_scale), it's rounding ties to the nearest even. Refer to https://en.wikipedia.org/wiki/Rounding for details.
```

onnx/backend/test/case/node/cast.py
```diff
@@ -82,6 +104,101 @@ def export() -> None:
...
+                        vect_float32_to_float8e5m2(np_fp32, fn=True, uz=True),
+                        fn=True,
+                        uz=True,
+                    )
+                    input = make_tensor(
+                        "x", TensorProto.FLOAT8E5M2FNUZ, [3, 4], input_values.tolist()
+                    )
+                else:
+                    raise ValueError(
+                        "Conversion from {from_type} to {to_type} is not tests."
```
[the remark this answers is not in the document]
xadupre: done

onnx/defs/tensor/defs.cc
xadupre: done

onnx/defs/tensor/defs.cc
```diff
@@ -34,33 +34,74 @@ User must be aware of precision loss and value change caused by range difference
...
 
 * Casting from floating point to:
-  * floating point: +/- infinity if OOR (out of range).
+  * floating point: +/- highest value if OOR (out of range),
+    infinities are converted to NaN if infinity is not available
   * fixed point: undefined if OOR.
   * bool: +/- 0.0 to False; all else to True.
 * Casting from fixed point to:
-  * floating point: +/- infinity if OOR. (+ infinity in the case of uint)
+  * floating point: +/- highest value if OOR (out of range),
```
yuanyao-nv, line 46: Now that we have separated the fp8 rules into a separate paragraph down below, is there any reason to modify this section? Is infinity not available in any other data type? Also I think "+/- infinity" would be clearer than "+/- highest value" since "highest value" sounds a bit like max norm.

nit: I think we can just use OOR here without the full name because it's already defined previously.

onnx/defs/quantization/defs.cc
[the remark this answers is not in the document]
xadupre: done

onnx/defs/tensor/defs.cc
xadupre: true, I'll revert the changes for this section.

xadupre: done

yuanyao-nv (approved):

liqunfu (approved):

lutzroeder:
@xadupre @yuanyao-nv @liqunfu @guoyu-wang @edgchen1 @linkerzhang will the `DataType` changes get ported to [ort.fbs](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/flatbuffers/schema/ort.fbs) to keep the two in sync?

edgchen1:
> @xadupre @yuanyao-nv @liqunfu @guoyu-wang @edgchen1 @linkerzhang will the `DataType` changes get ported to [ort.fbs](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/flatbuffers/schema/ort.fbs) to keep the two in sync?

I think so, once we update ONNX Runtime to consume a version of ONNX with this change. Thanks for the heads-up.
FYI @skottmckay

PR #4814, opened 2023-01-26
Fix cast to string in reference implementation

jbachurski:
### Description
Currently, the reference implementation of `Cast` handles strings improperly. This is because ONNX relates the `numpy.object_` dtype to `TensorProto.STRING`, rather than `numpy.str_`.
This behavior is somewhat often a source of bugs or questionable behavior, maybe someone could shed some light why was this choice made?

### Motivation and Context
Fixes string casts to actually convert to string representation, rather than converting array elements to Python objects. This is important for functionality of the reference implementation in other libraries.


onnx/reference/ops/op_cast.py
```diff
@@ -29,6 +29,8 @@ def cast_to(x, to):
             el = float32_to_bfloat16(xf[i], truncate=True)  # type: ignore[assignment]
             y[i] = el
         return y.reshape(x.shape)
+    if to == TensorProto.STRING:
```
xadupre, line 32: I would prefer a unit test which convert to string and then back to float to make sure it works both ways. Did you try to modify the type here: https://github.com/onnx/onnx/blob/main/onnx/mapping.py#L61?

jbachurski: I don't think this PR should modify the `STRING <-> np.object_` design choice, as I am afraid it could break downstream code. I would leave this decision for another PR by a collaborator.
I can try to add a unit test for this cast.

jbachurski: I added two unit tests (for basic float-to-string, I think one already existed in backend tests too) and a float-to-string-to-float test.

xadupre (approved):

jbachurski:
Thanks, ready to merge.

xadupre (approved):

PR #4815, opened 2023-01-27
Reference implementation inconsistently handles optionals

jbachurski:
### Description

Attempt to fix the `OptionalHasElement` reference implementation. 

Currently, any call to it returns a `list` - `[array([False])]`, as if it was a single-element vector wrapped in an `optional(tensor(bool))`, rather than a scalar `tensor(bool)` - `array(False)` - as would be expected. 

This also remains somewhat inconsistent with how ONNX Runtime handles optionals (it doesn't wrap them, a missing optional is just returned to Python as `None`). Maybe changing the representation would be useful to make this more clear?

@xadupre What was your idea for representing optional? From what I gathered I understand you wanted to use singleton lists to indicate optionals, i.e. `[None]` is a `Nothing`, and otherwise `[x]` is `Some x` (where `x` is a tensor)? Doesn't this intersect with `[x]` being a single-element sequence of `x`?
I'm also not sure what the meaning of an empty-list result `[]` in e.g. `OptionalGetElement` and some branches of `OptionalHasElement`?

### Motivation and Context

`OptionalHasElement` currently is not standard-consistent, the `output` in the docs [for optionals] is described as `A scalar boolean tensor.`, while an optional-wrapped single-element vector is returned instead. This is because the reference implementation doesn't differentiate sequences and optionals.

### DIFF
--- a/onnx/reference/ops/op_optional_has_element.py
+++ b/onnx/reference/ops/op_optional_has_element.py
@@ -11,9 +11,8 @@ class OptionalHasElement(OpRun):
         if x is None:
             return ([],)
         if isinstance(x, list):
-            if len(x) > 0:
-                return (np.array([e is not None for e in x]),)
+            (e,) = x
+            return (np.array(e is not None),)
         elif isinstance(x, np.ndarray):
-            if len(x.shape) > 0 and x.shape[0] > 0:
-                return (np.array([e is not None for e in x]),)
+            return (np.array(True),)
         return ([],)


jbachurski:
I didn't notice the operator was updated for 18 to allow sequences to be passed too, as it doesn't specify explicitly in that case the result should be a vector. 
As I see it we can't detect when the input is actually an optional and a scalar should be returned. I think the representation of optionals should be changed to allow this. A basic idea would be how ONNX Runtime does it (don't wrap in a list, just use `None` for a missing value). Do you have thoughts on this @xadupre ?

jbachurski:
I see. What about using an internal sentinel object, for instance `_NOTHING = object()`, to internally represent empty optionals (and full optionals can have no 'tag' on them, it's just an array)? 
Then we could replace `_NOTHING` with `None` - like ORT - once it's returned from `run`.
I think it's useful to have the Optional type for more general-purpose ONNX applications, even though it might not be present in deep learning models right now.

xadupre:
I'm still unsure about introducing a new type. A value is either undefined (None) or defined (!= None). Optional types can be used to check whether or not the value can be None. I don't think we need to introduce a new type when this information is already available from every node `OpRun`. It can be access the node type, its attributes to detect optional input. This is not checked by the reference implementation. I created PR #4835 to fix OptionalGetElement and OptionalHasElement. They were working due to another bug reference_evaluator_backend_test.py. More details are in the PR.

gramalingam:
I agree that `None` should suffice to represent an Optional value (that is absent), along with other legal values.

It would be helpful to pin down the representation. My initial guess would be:
* A tensor is represented by an np.array
* A sequence by an python list
* An optional by either None or any of the other valid values
* A map by a python dictionary?

@xadupre can you clarify? I assume that you are returning a tuple of values only because an Op may have multiple output values?

xadupre:
Everything you said is true. None for missing optional value, np.array for tensor, list for sequence, dictionary for map but this option never happens right now. All op return a tuple even if there is only one result. I wanted to avoid checking if it is a tuple or another type in many places.

jbachurski:
Solved by #4835

PR #4823, opened 2023-01-30
[CI] Compile with UBSan in CI and fix ParseData nullptr issue

### LINKED ISSUE #4820: [Feature request] CI: compile with UBSan and ASan
### System information

_No response_

UBSan finds undefined behavior in https://github.com/pytorch/pytorch/issues/93174. We can use it in our CI (maybe)? https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

### Alternatives considered

_No response_

### Describe the feature

_No response_

_No response_

### Feature Area

_No response_

None

### Notes

_No response_

jcwchen:
### Description

- At least running [UndefinedBehaviorSanitizer.html](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) check in certain CI: Compile with UBSan in Linux-CI.
- Fix ParseData nullptr issue while loading an empty tensor with empty raw_data.

### Motivation and Context

Fixes #4820


liqunfu (approved):

.azure-pipelines/MacOS-CI.yml
```diff
@@ -58,11 +58,14 @@ jobs:
       fi
       export ONNX_BUILD_TESTS=1
       export CMAKE_ARGS="-DONNX_WERROR=ON"
+      # enable UndefinedBehaviorSanitizer
+      export CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_C_FLAGS='-fsanitize=undefined -fno-sanitize-recover=undefined '"
```
justinchuby, line 62: Also -fsanitize=memory -fPIE -pie
-fsanitize=address

?

https://github.com/google/sanitizers/wiki/AddressSanitizer

https://github.com/google/sanitizers/wiki/MemorySanitizer

justinchuby: I discovered gcc has these options too https://developers.redhat.com/blog/2014/10/16/gcc-undefined-behavior-sanitizer-ubsan

justinchuby:
Looks like -fsanitize=fuzzer is also useful : https://learn.microsoft.com/en-us/cpp/build/reference/fsanitize?view=msvc-170

justinchuby:
Edit: We would need code change if we did want to use a fuzzer

.azure-pipelines/MacOS-CI.yml
```diff
@@ -58,11 +58,13 @@ jobs:
       fi
       export ONNX_BUILD_TESTS=1
       export CMAKE_ARGS="-DONNX_WERROR=ON"
+      # enable UndefinedBehaviorSanitizer
+      export CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_CXX_FLAGS='-fno-sanitize-recover=undefined -fsanitize=memory -fPIE -pie '"
```
justinchuby, line 62: Need -fsanitize=undefined to enable it or no?

.azure-pipelines/Linux-CI.yml
```diff
@@ -61,8 +61,8 @@ jobs:
       fi
       export CMAKE_ARGS="-DONNX_WERROR=ON -DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
       export ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI
-      # enable UndefinedBehaviorSanitizer
-      export CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_CXX_FLAGS='-fsanitize=undefined,address -fno-sanitize-recover=undefined '"
+      # enable more sanitizer
+      export CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_CXX_FLAGS='-fsanitize=undefined,address,null,alignment -fno-sanitize-recover=undefined '"
```
justinchuby, line 65: I think null is part of undefined?

jcwchen: > I think null is part of undefined?

Thanks. I think you are right.

BTW, currently the added sanitizer checks look fine without error when building ONNX. Is there something missing from the config? If not, is it possible that actually https://github.com/onnx/onnx/blob/10d126ef0754e32aeba93d714444fd0ab72d1e7f/onnx/defs/tensor_proto_util.cc#L129 is OK from ONNX perspective?

Either way, adding more sanitizer checks should be good to have.

justinchuby: The code will need to be executed I think. Was the line covered?

justinchuby: This needs to be `-fno-sanitize-recover=all ` it seems: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#:~:text=%2Dfsanitize%2Drecover%5B%3Dopts%5D

justinchuby:
We can prob disregard the leaks and hunt the UBs first. 

So -fsanitize=undefined only

justinchuby:
May need pytest -sv

justinchuby:
@jcwchen was this ready?

jcwchen:
> @jcwchen was this ready?

It is ready. Could @onnx/sig-operators-approvers please review/approve it? Thanks!

> I think it should be part of 1.13.1 and PyTorch 2.0 as we see errors from PyTorch users

I thought such an error comes from PyTorch's onnx submodule? If yes, perhaps we can just let PyTorch repo use newer ONNX commit? This seems like an old issue (not a regression) that ONNX (ParseData from C++ code) cannot handle an empty tensor well.

justinchuby:
> I thought such an error comes from PyTorch's onnx submodule? If yes, perhaps we can just let PyTorch repo use newer ONNX commit? This seems like an old issue (not a regression) that ONNX (ParseData from C++ code) cannot handle an empty tensor well.

Sounds good!

gramalingam (approved):

PR #4824, opened 2023-01-31
Handle variants of constant op in shape inference

gramalingam:
Fixes issue https://github.com/onnx/onnx/issues/3985


titaiwangms:
cc @jcwchen @liqunfu

jcwchen (approved):
LGTM. Might be worthy to add some tests if it is not difficult.

gramalingam:
> LGTM. Might be worthy to add some tests if it is not difficult.

Yes, I was thinking about this too. I couldn't think of a direct test, but I think I can do one using data-propagation, so will try it.

onnx/test/model_inference_test.py
```diff
@@ -132,6 +133,53 @@ def test_mi_function_multiple_calls(self):
...
 
+    def _check_shape(self, model_text: str, *expected: typing.Sequence[int]):
+        """Check that the model inference infers the expected shapes for outputs.
+        Restricted to the simple case of tensor type outputs with completely
+        known shapes.
+        """
+        model = onnx.parser.parse_model(model_text)
+        inferred = onnx.shape_inference.infer_shapes(model, True, True, True)
+        outputs = inferred.graph.output
+        for (output, expected_shape) in zip(outputs, expected):
```
justinchuby, line 144: Assert equal length and non empty?

gramalingam: Not necessary, I feel. That wouldn't be testing the inference-code.

justinchuby: Just wanted to make sure the loop is run at least once. NBD

Issue #4809, opened 2023-01-26
Gather-1/Gather-11 example in documentation is not matched with onnxruntime's result


Tee0125:
# Bug Report

### Is the issue related to model conversion?
No

### Describe the bug

Documented Gather operator's behavior is not matched w/ onnxruntime's result.

Gather-1 / Gather-11 example descriptibed in Changelog.md is as follow:

```
data = [
      [1.0, 1.2, 1.9],
      [2.3, 3.4, 3.9],
      [4.5, 5.7, 5.9],
  ]
  indices = [
      [0, 2],
  ]
  axis = 1,
  output = [
      [
          [1.0, 1.9],
          [2.3, 3.9],
          [4.5, 5.9],
      ],
  ]
```

however, onnxruntime's results is as in below:

```
output = [
    [
        [1. , 1.9]
    ],
    [
        [2.3, 3.9]
    ],
    [
        [4.5, 5.9]
    ]
]
```

shape of documentation's result is (1, 3, 2) and shape of onnx runtime's result is (3, 1, 2).

### System information

- OS X
- ONNX version: 1.12.0
- Python version: 3.8

### Reproduction instructions

```
import onnx
import onnxruntime

import numpy as np

from onnx.helper import make_node, make_graph, make_model, make_tensor_value_info, make_tensor


data = [
      [1.0, 1.2, 1.9],
      [2.3, 3.4, 3.9],
      [4.5, 5.7, 5.9],
]

indices = [
      [0, 2],
]

FLOAT = onnx.TensorProto.FLOAT
INT32 = onnx.TensorProto.INT32

input_shape = (3, 3)
output_shape = (3, 1, 2)

input_tensors = [make_tensor_value_info('input', onnx.TensorProto.FLOAT, input_shape)]
output_tensors = [make_tensor_value_info('output', onnx.TensorProto.FLOAT, output_shape)]

node = make_node('Gather',
                 inputs=['input', 'indices'],
                 outputs=['output'],
                 axis=1)

tensors = [make_tensor('indices', onnx.TensorProto.INT32, (1, 2), np.array(indices).astype(np.int32).flatten())]

graph = make_graph([node],
                   'graph',
                   input_tensors,
                   output_tensors,
                   tensors)

model = make_model(graph)
model.opset_import[0].version = 11

inputs = {'input': np.array(data).astype(np.float32)}
sess = onnxruntime.InferenceSession(model.SerializeToString())
outputs = sess.run(None, inputs)

[...]

PR #4827, opened 2023-01-31
fix wrong example of Gather-11/Gather-1 documentation

Tee0125:
Signed-off-by: Tee Jung <tee.ty.jung@openedges.com>

### Description

Gather-1/Gather-11 example in documentation is not matched with onnxruntime's result

Fixing wrong documentation of Gather-1/Gather-11's behavior when axis=1
 
### Motivation and Context

onnxruntime's result of Gather-13 is matched with document, but results of Gather-1/Gather-11 are not matched with document so I fixed it

https://github.com/onnx/onnx/issues/4809


daquexian:
Thanks for your contribution!

docs/Changelog.md is auto-generated. Please update the content of `Gather_ver11_doc` and `Gather_ver1_doc` in onnx/defs/tensor/old.cc instead.

jcwchen (approved):
LGTM. Thank you for fixing!

gramalingam:
After changing the source, need to run `python onnx\defs\gen_doc.py` to generate the documentation and include those changes in the PR

Tee0125:
ok, I pushed fresh new docs

onnx/defs/tensor/old.cc
```diff
@@ -1631,11 +1631,9 @@ output[i_{0}, ..., i_{q-1}, j_{0}, ..., j_{r-2}] = input[j_{0}, k, j_{1}, ..., j
   ]
   axis = 1,
   output = [
-      [
-          [1.0, 1.9],
-          [2.3, 3.9],
-          [4.5, 5.9],
-      ],
+      [[1.0, 1.9]],
```
gramalingam, line 1634: Thanks for the fix. I think the line 1621 also needs to be updated (as in Gather-13)? May be same for Gather-1 also down below?

Tee0125: line 1621 is updated as Gather-13 and Gather-1's doc has only input/output definition in example so, I think, no update is needed.

gramalingam (approved):

gramalingam:
Great, thanks very much for the PR! For some reason, it doesn't allow me to update to current main branch, you may need to do it.

Issue #1937, opened 2019-04-15
Add Tensor(String) Input support for Equal ops


waterknows:
We would like to have the ONNX Equal operator to support tensor(string) input type, similar to what Equal operator in Tensorflow is doing: https://www.tensorflow.org/api_docs/python/tf/math/equal

linkerzhang:
do you want to send a PR to propose the update?

yuxiaosun:
Thanks! I am drafting the PR. This is more of a data type change to exsiting operator - should I modify the exisiting operator? or what would be the best way to start?

PR #4828, opened 2023-01-31
Extend Equal operator to support string type

gramalingam:
### Description
Enable equality comparison of string tensors.

### Motivation and Context

Fixes https://github.com/onnx/onnx/issues/1937


wschin:
String is always a difficult format to standardize. I am not sure if this is the right way to go. There are several types we can use to store string; for example, "abc" can be stored to int8 array [97, 98, 99] and then we can treat string comparisons as number comparisons.

wschin (approved):

xadupre:
Should we add a test in the backend tests?

gramalingam:
> Should we add a test in the backend tests?

Added, thanks!

liqunfu (approved):

PR #4831, opened 2023-01-31
Fix SplitToSequence reference implementation

xadupre:

### Description
SplitToSequence current implementation does not support scaler values for input *split*. This PR fixes it.


onnx/backend/test/case/node/splittosequence.py
```diff
@@ -0,0 +1,78 @@
...
+        data = np.arange(18).reshape((3, 6)).astype(np.float32)
+        split = np.array(2, dtype=np.int64)
+
+        node = onnx.helper.make_node(
+            "SplitToSequence", ["data", "split"], ["seq"], axis=1
+        )
+
+        expected_outputs = [
+            [
+                np.array([[0.0, 1.0], [6.0, 7.0], [12.0, 13.0]], dtype=np.float32),
```
justinchuby, line 23: So I think one difference with the backend test and torch's OpInfo is the effort required to create one. OpInfo is essentially a dataclass with information about specific ops, including example inputs. This makes it very easy to add test inputs, typically as a matter of 2 lines.

justinchuby: As an example, the input samples for split: https://github.com/pytorch/pytorch/blob/23d58fedb107181b90725a72b97c5ea184252c93/torch/testing/_internal/common_methods_invocations.py#L162-L183

Negative tests: https://github.com/pytorch/pytorch/blob/23d58fedb107181b90725a72b97c5ea184252c93/torch/testing/_internal/common_methods_invocations.py#L221-L236

And the OpInfo: https://github.com/pytorch/pytorch/blob/23d58fedb107181b90725a72b97c5ea184252c93/torch/testing/_internal/common_methods_invocations.py#L13640-L13648

xadupre: I think onnx is close to that. The only thing I find annoying is the the generation of the documentation, test case and stat coverage before adding the changes. That may be the topic of another PR.

justinchuby:
cc @gramalingam @jcwchen

onnx/backend/test/case/node/splittosequence.py
gramalingam: Yes, it would be great to have automatic generation (of docs etc.) happen fully automatically in the backend, but that's a different topic.

onnx/test/reference_evaluator_test.py
```diff
@@ -2914,6 +2914,61 @@ def test_cast_float_to_string_and_back(self):
         got = ref.run(None, {"X": data})[0]
         assert_allclose(got, np.array([1.152512, -0.152612, 0.0, np.nan]))
 
+    def test_split_to_sequence(self):
+        X = make_tensor_value_info("X", TensorProto.FLOAT, [None])
```
gramalingam, line 2918: I think you mean `None` instead of `[None]`. `None` means a tensor of unknown rank, while `[None]` means a 1-dimensional tensor of unknown dimension.

onnx/reference/ops/op_split_to_sequence.py
```diff
@@ -1,19 +1,41 @@
...
+                split.append(left)
+
+        sli = [slice(0, s) for s in mat.shape]
+        res = []
+        pos = 0
+        for spl in split:
+            sli[axis] = slice(pos, pos + spl)  # type: ignore
+            pos += spl
+            res.append(mat[tuple(sli)])
+        return tuple(res)
```
gramalingam, line 28: Why `tuple`? IIUC, it should be a list.

onnx/reference/ops/op_split_to_sequence.py
```diff
@@ -1,19 +1,41 @@
...
+            length = None
         else:
             if len(split.shape) == 0:
-                num_outputs = split
+                length = split
             else:
-                num_outputs = split.shape[0]
-        res = list(self.common_run(mat, split, axis=axis, num_outputs=num_outputs))
+                length = split.shape[0]
+        res = list(self.common_run(mat, split, axis=axis, length=length))
```
gramalingam, line 38: The line below looks incorrect. May be it is supposed to be `if keepdims == 0 and split is None` ?

onnx/reference/ops/op_split_to_sequence.py
```diff
@@ -1,19 +1,41 @@
...
+from onnx.reference.op_run import OpRun
 
 
-class SplitToSequence(CommonSplit):
+class SplitToSequence(OpRun):
+    def common_run(self, mat, split, axis, length):  # type: ignore
+        if split is None:
+            split = [1 for _ in range(mat.shape[axis])]
+        elif len(split.shape) == 0:
+            # A scaler
```
gramalingam, line 12: nit: `scaler` => `scalar`

onnx/reference/ops/op_split_to_sequence.py
```diff
@@ -1,19 +1,41 @@
 # SPDX-License-Identifier: Apache-2.0
 # pylint: disable=C0200,W0221
 
-from .op_split import CommonSplit
+from onnx.reference.op_run import OpRun
 
 
-class SplitToSequence(CommonSplit):
+class SplitToSequence(OpRun):
+    def common_run(self, mat, split, axis, length):  # type: ignore
```
gramalingam, line 8: IIUC, the input parameter `length` is not used here and can be omitted.

gramalingam:
While reviewing this PR, I found the ONNX documentation of the op incomplete/unclear. I updated it here: https://github.com/onnx/onnx/pull/4847 ... please take a look if that looks okay, thanks!

onnx/reference/ops/op_split_to_sequence.py
xadupre: It is used in line 15.

gramalingam: But that value is defined in line 14, it's not the passed in parameter value

xadupre: Sorry, i did not see I was computing length in both methods. I removed the unnecessary code.

onnx/reference/ops/op_split_to_sequence.py
gramalingam: Thanks for addressing the other issues ... I think this one is still an issue?

xadupre: > The line below looks incorrect. May be it is supposed to be `if keepdims == 0 and split is None` ?

This test is checking that case: https://github.com/onnx/onnx/pull/4831/files#diff-e88ac7f22fe3c2c4e48cc6e90a49600312dde1f3e7757e0427426b0543ebfb87R2959.

gramalingam: But why is the condition `axis is None` ? It doesn't depend on the axis attribute. It depends on whether the optional input split is specified, right?

xadupre: True, axis cannot be None aynway. I remove this condition.

onnx/reference/ops/op_split_to_sequence.py
```diff
@@ -37,7 +37,7 @@ def common_run(
 
     def _run(self, mat, split=None, axis=None, keepdims=None):  # type: ignore
```
justinchuby, line 38: Possible to annotate the params?

onnx/reference/ops/op_split_to_sequence.py
```diff
@@ -37,7 +37,7 @@ def common_run(
...
         res = self.common_run(mat, split, axis=axis)
-        if keepdims == 0 and axis is not None:
+        if keepdims == 0:
             for i in range(len(res)):
```
justinchuby, line 41: Consider enumerate?

xadupre: I chose this syntax because I modify the list while walking through it.

onnx/reference/ops/op_split_to_sequence.py
gramalingam: Thanks. But I think we do need the condition `and split is None`. Consider [this line of the spec](https://github.com/onnx/onnx/blob/main/onnx/defs/sequence/defs.cc#L299) ... in general, removing the dim does not make sense unless that dim has a value of 1 in each tensor in the sequence. So, the spec kind of makes sense, and it is better if we align with this exactly.

xadupre: Sorry, I missed it. Argument `keepdims` is pretty common. I probably never read this line. I did not expect the value to be ignored in some cases.

onnx/reference/ops/op_split_to_sequence.py
xadupre: done

onnx/reference/ops/op_split_to_sequence.py
```diff
@@ -1,42 +1,49 @@
...
-        if keepdims == 0 and axis is not None:
+    def _run(
+        self,
+        mat: np.ndarray,
+        split: Optional[np.ndarray] = None,
+        axis: int = 0,
+        keepdims: int = 1,
+    ) -> Tuple[np.ndarray]:
+        res = self.common_run(mat, split, axis=axis)
+        if split is None or not keepdims:
```
gramalingam, line 46: Sorry, I think the `or` should be `and`

xadupre: Sorry. I changed it.

PR #4835, opened 2023-01-31
Fix OptionalHasElement in reference implementation

xadupre:

### Description
Fix implementation of OptionalHasElement, OptiionalGetElement (replaces #4815).

There were two issues. One was in OptionalHasElement, an undefined result is None and cannot be an empty list. The second one was in the backend test. Input proto were not correctly loaded. An OptionalProto could loaded as an SequenceProto with the code (SequenceProto().ParseFromString()). This code was expected to fail but does not. That explains why OptionalHasElement was confusing None and empty lists.


onnx/reference/ops/op_optional_get_element.py
```diff
@@ -9,9 +9,7 @@
 class OptionalGetElement(OpRun):
     def _run(self, x):  # type: ignore
         if x is None:
-            return ([],)
-        if isinstance(x, list):
+            return (None,)
```
gramalingam, line 12: I think the best implementation would be to fail (raise an exception) if `x is None` and `return x` otherwise.

Of course, this is dependent on the overall conventions used in the reference implementation. I agree with your comments in https://github.com/onnx/onnx/pull/4815 that we don't use any wrappers. A value is either None or one of the other valid values, such as a tensor or list of tensors.

onnx/reference/ops/op_optional_has_element.py
```diff
@@ -9,11 +9,7 @@
...
-            if len(x) > 0:
-                return (np.array([e is not None for e in x]),)
-        elif isinstance(x, np.ndarray):
-            if len(x.shape) > 0 and x.shape[0] > 0:
-                return (np.array([e is not None for e in x]),)
-        return ([],)
+            return (np.array(False),)
+        if isinstance(x, (np.ndarray, list)):
+            return (np.array(True),)
+        return (np.array(False),)
```
gramalingam, line 15: When is line 15 expected to be used? I would expect `return (np.array(x is None),)` to be sufficient ... unless you want to check whether an invalid input is supplied and raise an exception for that case, but that's a different point anyway.

gramalingam: Specifically, I think raising error is better in line 15.

xadupre: done

onnx/reference/ops/op_optional_get_element.py
xadupre: done

gramalingam (approved):

PR #4848, opened 2023-02-03
Update docs to reflect CI and lint changes

justinchuby:
### Description

Update docs to reflect CI and lint changes

- Remove section on mypy because it is covered by `lintrunner`
- Format the markdown file


setup.py
```diff
@@ -274,27 +274,12 @@ def build_extensions(self):
             self.copy_file(src, dst)
 
 
-class mypy_type_check(ONNXCommand):
```
justinchuby, line 277: We should eventually deprecate all setup.py commands

jcwchen (commented):
Thanks for updating!

It could be done by another PR: Because recently you helped to update a few CI pipelines, at some point it seems worthy to update https://github.com/onnx/onnx/blob/main/docs/CIPipelines.md accordingly as well.

justinchuby:
Can merge

PR #4851, opened 2023-02-04
Use ruff to replace flake8 and plugins

justinchuby:
### Description

Use ruff https://github.com/charliermarsh/ruff to lint files, replacing flake8 and its plugins. It is faster and has auto-fix.

- Also enforce PEP8-naming
- Simplified the lint CI

### Motivation and Context

Code quality


jcwchen (approved):
Please fix DCO. To confirm, so current codebase does follow pep8-naming?

justinchuby:
> Please fix DCO. To confirm, so current codebase does follow pep8-naming?

I think so! Although I am going to replace this with ruff

justinchuby:
@jcwchen it good now

pyproject.toml
```diff
@@ -70,3 +70,44 @@ disable = [
     "no-name-in-module",
     "useless-return",
 ]
+
+[tool.ruff]
+target-version = "py37"
```
jcwchen, line 75: Why only set the lowest supported Python version here?

onnx/tools/net_drawer.py
```diff
@@ -61,18 +61,20 @@ def ReallyGetOpNode(op: NodeProto, op_id: int) -> pydot.Node:
             node.set_URL(url)
         return node
 
-    return ReallyGetOpNode
+    return really_get_op_node
 
 
-def GetPydotGraph(
+def get_pydot_graph(
```
jcwchen, line 67: Not sure whether it is safe to change the public API name here. Possibly some users might rely on these APIs already?

pyproject.toml
```diff
@@ -70,3 +70,44 @@ disable = [
...
+]
+ignore = [
+    "B9", # Opinionated bugbear rules
+    "E402", # warns on imports coming after statements. There are important use cases
+            # like demandimport (https://fburl.com/demandimport) that require statements
+    "E501",
+    "F405", # FIXME: is hard to silence since we indeed do star import
+    "N803", # Argument casing
+    "N812",
+    "N806",
```
jcwchen, line 93: Could you please also add comments for these 2 if possible? Thanks!

.github/workflows/lint.yaml
```diff
@@ -6,34 +6,16 @@ on:
...
-          sudo apt-get install -qq -o=Dpkg::Use-Pty=0 -y --no-install-recommends dos2unix
-          git submodule update --init --recursive
-          export ONNX_BUILD_TESTS=0
-          export ONNX_ML=1
-          export CMAKE_ARGS="-DONNXIFI_DUMMY_BACKEND=ON -DONNX_WERROR=ON"
-          # ONNX lite
-          export CMAKE_ARGS="${CMAKE_ARGS} -DONNX_USE_LITE_PROTO=ON"
-          export ONNX_NAMESPACE=ONNX_NAMESPACE_FOO_BAR_FOR_CI
-
-          python setup.py install
```
jcwchen, line 36: To confirm, we don't need to build ONNX from source in the Lint CI anymore?

pyproject.toml
justinchuby: This is so ruff doesn’t auto fix with styles Python 3.7 doesn’t support.

.github/workflows/lint.yaml
justinchuby: Yes for the optional lints. Lintrunner still needs built onnx

onnx/tools/net_drawer.py
justinchuby: If you like I can revert this change ?

.github/workflows/lint.yaml
justinchuby: https://github.com/charliermarsh/ruff#target-version

onnx/tools/net_drawer.py
jcwchen: Yes, I think it's safer. And please add a TODO comment to change it in next major version bump. Thank you!

pyproject.toml
justinchuby: Updated and refined

onnx/tools/net_drawer.py
justinchuby: Done

justinchuby: Actually, this is a commandline tool. I don't think anyone will use the functions themselves? If they do personally I think it's still a minor break?

justinchuby: (also, tools/executables (anything with a `main()`) shouldn't be imported so implementation details are expected to change imo)

justinchuby:
Looks like coverage changed after the model updates in ci: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=37519&view=logs&j=2105a913-e5a3-5f74-1c30-4f8b7a56f2db&t=c2afa4d8-3a4c-5e2d-1542-4328675067bd&l=85

jcwchen:
> Looks like coverage changed after the model updates in ci: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=37519&view=logs&j=2105a913-e5a3-5f74-1c30-4f8b7a56f2db&t=c2afa4d8-3a4c-5e2d-1542-4328675067bd&l=85

It failed, because some of real models was not downloaded from download.onnxruntime.ai. The new URLs seem not reliable enough... I have rerun the pipelines and now they are all green.

onnx/tools/net_drawer.py
justinchuby: I see ok! Thanks for doing the search. I will revert in this case. I don't think it's worth changing this just for the naming. Although we should not have exposed functions in a file with `__main__`

jcwchen (approved):
LGTM. Thank you for introducing ruff!

justinchuby:
> @justinchuby a question: are the changes from an automated tool, or do they include some manual changes? Asking since manual changes probably benefit from a careful review, unlike automated changes.

only naming changes are manual (pep8 naming)

gramalingam (approved):
LGTM, thanks!

justinchuby:
@jcwchen something's wrong with the queue?

justinchuby:
Looks like the workflows need more triggers

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group

https://github.com/community/community/discussions/39933

justinchuby:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions

jcwchen:
Here is the status:
image

It's a little tricky -- for required pipeline running by GitHub action, adding `merge_group` should help. For other pipeline providers, for instance Azurepipelines, using some branch detection should help. However, for required DCO, I have no idea what we can do to make it pass... I will try to ask it in https://github.com/orgs/community/discussions/46757. It's still in beta so it might take some time to change and improve. For now, I am inclined to disable it, if there is no way to make DCO run in the queue.

Issue #4850, opened 2023-02-03
[Reference runtime] ReduceSum an integer is required (got type RefAttrName)


justinchuby:
While running the gpt2 model:

```
ReduceSum(self_5, dim_13) -> result
F

================================ FAILURES =================================
_______________ TestFxToOnnxWithOnnxRuntime.test_gpt2_tiny ________________
Traceback (most recent call last):
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/onnx/reference/ops/op_reduce_sum.py", line 43, in _run
    res = np.sum(x, axis=axes, keepdims=keepdims, dtype=x.dtype)
  File "<__array_function__ internals>", line 180, in sum
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 2298, in sum
    return _wrapreduction(a, np.add, 'sum', axis, dtype, out, keepdims=keepdims,
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 86, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
TypeError: an integer is required (got type RefAttrName)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/onnx/reference/op_run.py", line 421, in run
    res = self._run(*args, **kwargs)
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/onnx/reference/op_run.py", line 574, in _run
    results = self.impl_.run(None, feeds, attributes=attributes)
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/onnx/reference/reference_evaluator.py", line 457, in run
    outputs = node.run(*inputs, **linked_attributes)
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/onnx/reference/ops/op_reduce_sum.py", line 20, in run
    res = self._run(x, axes=axes, keepdims=keepdims)
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/onnx/reference/ops/op_reduce_sum.py", line 46, in _run
    raise TypeError(
[...]

PR #4856, opened 2023-02-06
Fix linked_attributes in reference implementation of ReduceSum

xadupre:

### Description
Fixes issue #4850.

### DIFF
--- a/onnx/reference/ops/op_reduce_sum.py
+++ b/onnx/reference/ops/op_reduce_sum.py
@@ -15,20 +15,10 @@ class ReduceSum_1(OpRunReduceNumpy):
 
 
 class ReduceSum_13(OpRunReduceNumpy):
-    def run(self, x, axes=None, keepdims=None):  # type: ignore
-        keepdims = keepdims or self.keepdims  # type: ignore
-        res = self._run(x, axes=axes, keepdims=keepdims)
-        if res[0].dtype != x.dtype:
-            raise RuntimeTypeError(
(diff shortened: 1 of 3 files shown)


justinchuby (approved):
Thank you!

onnx/reference/ops/op_reduce_sum.py
```diff
@@ -15,20 +13,10 @@ def _run(self, x, axes=None, keepdims=None):  # type: ignore # pylint: disable=W
...
-                f"(operator {self.__class__.__name__!r})."
-            )
-        return res
-
-    def _run(self, x, axes=None, keepdims=None):  # type: ignore
+    def _run(self, x, axes=None, keepdims=None, noop_with_empty_axes=None):  # type: ignore
         if (
             axes is None or len(axes.shape) == 0 or axes.shape[0] == 0
-        ) and self.noop_with_empty_axes:  # type: ignore
+        ) and noop_with_empty_axes:  # type: ignore
```
gramalingam, line 19: Wondering why there is a condition `len(axes.shape) == 0` in previous line? This may have nothing to do with the issue this PR is fixing, but just trying to understand. I believe len(axes.shape) == 0 means a zero-dimensional tensor (which is a scalar with one value?

xadupre: One unit test in the backend test defines the axes as `axes = np.array([], dtype=np.int64)`. shape is `(0,)`. I'll improve the code.

gramalingam (approved):

Issue #4857, opened 2023-02-06
s3.amazonaws.com/download.onnx/models are forbidden for download


jcwchen:
# Bug Report

### Is the issue related to model conversion?

No

### Describe the bug

s3.amazonaws.com/download.onnx/models are forbidden for download. People will see test failure if testing ONNX (pytest under onnx directory) in a fresh envionrment (including current CI pipelines), because the model download link is broken. e.g., https://s3.amazonaws.com/download.onnx/models/opset_10/resnet50.tar.gz.

### Reproduction instructions

```
cd onnx
pytest
```

### Expected behavior

pytest should always succeed.

philass:
Currently `onnx-mlir` pipelines are broken because we pull in these tests

jcwchen:
@philass Unfortunately, these models were setup long time ago and it's hard to find the original owner of s3 instance to fix them. In addition, currently in onnx repo there is a weekly pipeline to test the latest onnx with all ONNX model Zoo models, which should cover the same tests by these models.

Since this issue is blocking pipelines in ONNX as well, I have created a PR to remove them: https://github.com/onnx/onnx/pull/4859. Going forward I would suggest using [onnx.hub](https://github.com/onnx/onnx/blob/main/docs/Hub.md) to download these real ONNX models (you should be able to find corresponding models in https://github.com/onnx/models). Please note that git-lfs quota for onnx/models is quite limited so I would also suggest to cache them somehow in the pipelines or running it less frequently.

jcwchen:
Please note that we have transferred the model from AWS to download.onnxruntime.ai by this PR https://github.com/onnx/onnx/pull/4865. We will have a patch release 1.13.1 for this update soon. Reopen this PR to engage more awareness for a while.

philass:
@jcwchen thanks for explaining the situation and the suggestions for workarounds. It is much appreciated

jcwchen:
The fresh ONNX 1.13.1 patch should resolve this issue by https://github.com/onnx/onnx/pull/4865. Using the latest ONNX should not bump into this issue. I will close this issue here. Please note that going forward we would like to remove these real model tests in next ONNX release. See detail here: https://github.com/onnx/onnx/issues/4885.

PR #4859, opened 2023-02-06
Temporarily disable testing real models from AWS

jcwchen:
### Description

These models were moved to git-lfs in https://github.com/onnx/models. Currently all ONNX Model Zoo models will be tested weekly by the weekly CI pipleine: https://github.com/onnx/onnx/blob/main/.github/workflows/weekly_mac_ci.yml. ~Removing testing real models from AWS should be fine.~ To faster unblock global pipeline faillure, temporarily disable testing real models from AWS. We will add them back in the near future.

### Motivation and Context

See https://github.com/onnx/onnx/issues/4857. The models from AWS are broken.


liqunfu (approved):
THanks for the fix!  I agree to remove the failing tests for now. Shall we replace them later?

wschin:
It's better to explain how future developers will run equivalent tests on their local machines in this PR. Can you add a document with real code to explain how to use ONNX hub for test a model with example data? I saw [limited doc](https://github.com/onnx/onnx/blob/main/docs/Hub.md) for loading a model but doesn't find instructions to really execute the model with example data.

yuslepukhin:
This is massif. Can we try this PR on ORT branch to make sure this fixes our pipelines?

jcwchen:
After discussion with Liqun, we decide to temporarily disable testing real models from AWS in onnx repo to faster unblock pipeline failures. Liqun will try to move them elsewhere like Azure. Then, we will add these tests back with updated URL instead of the current AWS URL.

> It's better to explain how future developers will run equivalent tests on their local machines in this PR. Can you add a document with real code to explain how to use ONNX hub for test a model with example data? I saw [limited doc](https://github.com/onnx/onnx/blob/main/docs/Hub.md) for loading a model but doesn't find instructions to really execute the model with example data.

Good point. IIUC, onnx repo only tests the model itself: https://github.com/onnx/onnx/blob/main/onnx/test/test_backend_test.py#L38-L42 since previously there is no runtime implementation here. So the corresponding tests will be (Take AlexNet as an example)

```
import itertools
import onnx
model = onnx.hub.load('alexnet', opset=9)
onnx.checker.check_model(model)
model = onnx.shape_inference.infer_shapes(model, check_type=True, strict_mode=True)
value_infos = {
    vi.name: vi
    for vi in itertools.chain(model.graph.value_info, model.graph.output)
}
for node in model.graph.node:
    for i, output in enumerate(node.output):
        if node.op_type == "Dropout" and i != 0:
            continue
        assert output in value_infos
        tt = value_infos[output].type.tensor_type
        assert tt.elem_type != onnx.TensorProto.UNDEFINED
        for dim in tt.shape.dim:
            assert dim.WhichOneof("value") == "dim_value"
```

I will provide the full equivalent testing method if we eventually decide to remove them for good.

jcwchen:
> This is massif. Can we try this PR on ORT branch to make sure this fixes our pipelines?

ORT pipelines will be OK for now by manually disabling these real model tests (from onnx repo) there: https://github.com/microsoft/onnxruntime/pull/14606. In the near future, after we have transferred these models to a new place, we will update the URL in ONNX repo, have a patch release commit with that update and then consume that ONNX commit in ORT to recover these tests. Ideally only URLs will be updated and all other stuff in ONNX will be consistent.

xadupre:
Instead of copying the models somewhere else, would it be possible to remove the initializers and replace them by a node `ConstantOfShape`. The models should be a lot smaller and could be part of the repository without disabling the tests.

jcwchen:
Thank you all for the input. This PR is superseded by https://github.com/onnx/onnx/pull/4865 Instead of disabling them, we have replaced the broken URL with the new URL under onnxruntime.ai. Current ONNX 1.13 still has the old broken URLs and so we will have a a patch 1.13.1 release which includes #4865.

PR #4861, opened 2023-02-07
Use light local onnx file instead of heavy remote files

xadupre:
### Description
One test is download heavy onnx files. This change use use local files instead. Every heavy initializer in the original model was replaced by a node ConstantOfShape. Every name was shortened.

### Motivation and Context
Reduce network load, reduce CI breaks.


xadupre:
A unit test checking the backend based on the reference implementation should be added.

gramalingam:
Thanks, this looks great!

Can you add a couple of lines `readme.md` to the `light` and `real` folders to explain what they contain? Eg., for the light models, something like the text in the PR description above would help (explaining how these models were generated from the real model). Actually, if you used a script, may be adding that script to the folder may help anyone who wants to add another model in the future. As for `real`, I am unclear what it is.

gramalingam:
I guess as part of onnx package is better (though I will leave it to you), but that may also mean adding a couple of lines in the `docs\PythonAPIOverview.md` to help users discover and use it.

onnx/backend/test/data/light/README.md
```diff
@@ -0,0 +1,7 @@
+# Light models
+
+The models in this folder were creared by replacing
```
gramalingam, line 3: `creared` to `created`

onnx/backend/test/runner/__init__.py
```diff
@@ -298,17 +299,64 @@ def device_test_func(*args: Any, **kwargs: Any) -> Any:
...
+        """
+        Generates a random tensor based on the input definition.
+        """
+        if not x.type.tensor_type:
+            raise NotImplementedError(
+                f"Unable to generate random data for model {name!r} and input {x}."
+            )
+        if x.type.tensor_type.elem_type != 1:
+            raise NotImplementedError(
+                f"Unable to generate random data for model {name!r} and input {x}."
```
gramalingam, line 315: May be useful to add reason: e.g., "Currently limited to float tensors".

onnx/backend/test/runner/__init__.py
```diff
@@ -298,17 +299,64 @@ def device_test_func(*args: Any, **kwargs: Any) -> Any:
...
+    @staticmethod
+    def generate_random_data(
+        x: ValueInfoProto, seed: int = 0, name: str = ""
+    ) -> np.ndarray:
+        """
+        Generates a random tensor based on the input definition.
+        """
+        if not x.type.tensor_type:
+            raise NotImplementedError(
+                f"Unable to generate random data for model {name!r} and input {x}."
```
gramalingam, line 311: May be useful to add reason: e.g., "Input expected to have tensor type."

onnx/backend/test/runner/__init__.py
```diff
@@ -298,17 +299,64 @@ def device_test_func(*args: Any, **kwargs: Any) -> Any:
...
+            )
+        if x.type.tensor_type.elem_type != 1:
+            raise NotImplementedError(
+                f"Unable to generate random data for model {name!r} and input {x}."
+            )
+        shape = tuple(d.dim_value for d in x.type.tensor_type.shape.dim)
```
gramalingam, line 317: Assumes `d.HasField("dim_value")` ... may be useful to add a check

onnx/backend/test/runner/__init__.py
```diff
@@ -317,18 +365,62 @@ def run(test_self: Any, device: str) -> None:
...
+                    inputs.append(name)
+                    n_input += 1
+                    x = onx.graph.input[i]
+                    value = self.generate_random_data(
+                        x, seed=0, name=model_test.model_name
+                    )
+                    feeds[x.name] = value
+                    with open(name, "wb") as f:
+                        f.write(onnx.numpy_helper.from_array(value).SerializeToString())
+                # Reference implementation of operator Conv is too slow.
```
gramalingam, line 398: Don't understand above comment. Is that some copy-paste error?

onnx/tools/replace_constants.py
```diff
@@ -0,0 +1,236 @@
...
+    make_tensor_value_info,
+    set_model_props,
+    tensor_dtype_to_np_dtype,
+)
+from onnx.numpy_helper import from_array
+
+
+def _replace_constant(
+    node: NodeProto, threshold: int
+) -> Tuple[Optional[NodeProto], NodeProto]:
```
gramalingam, line 31: Nit: may be simpler and more general to return a `List[NodeProto]`.

onnx/backend/test/data/light/README.md
xadupre: done

onnx/backend/test/runner/__init__.py
xadupre: done

onnx/backend/test/runner/__init__.py
xadupre: done

onnx/backend/test/runner/__init__.py
xadupre: done

onnx/backend/test/runner/__init__.py
xadupre: It is not from a copy paste. ReferenceEvaluator is used to generate the expected output of the light models. It is slow if one of them is including a Conv operator which is usually the case. I extended the comment and also create a new test file checking the backend API with the ReferenceEvaluator.

onnx/tools/replace_constants.py
xadupre: done

onnx/tools/replace_constants.py
```diff
@@ -0,0 +1,232 @@
...
+    make_tensor_value_info,
+    set_model_props,
+    tensor_dtype_to_np_dtype,
+)
+from onnx.numpy_helper import from_array
+
+
+def _replace_constant(node: NodeProto, threshold: int) -> List[NodeProto]:
+    """
+    Replace a node *Constant* two nodes, one *Constant* for the shape,
```
gramalingam, line 31: Nit: Suggested replacement:
```
Replaces a Constant node with a large tensor (with more than threshold elements)
by a sequence of nodes that produces a dummy constant of same shape as original tensor.
```

gramalingam (approved):

onnx/tools/replace_constants.py
xadupre: done

justinchuby:
Some windows tests failing https://github.com/onnx/onnx/actions/runs/4265586790/jobs/7425083584#step:14:823

jcwchen:
> Some windows tests failing https://github.com/onnx/onnx/actions/runs/4265586790/jobs/7425083584#step:14:823

At first glance I didn't find test_backend_reference.py was introduced in this PR. Since the tests for test_cast_FLOAT_to_BFLOAT16 only fails on Windows's x86 machine, I guess NumPy on on Windows's x86 machine might behave differently that others and causes this issue?

justinchuby:
np on windows defaults int to INT32 than INT64 is one difference I knowLe 27 févr. 2023 à 08:52, Chun-Wei Chen ***@***.***> a écrit :﻿

Some windows tests failing https://github.com/onnx/onnx/actions/runs/4265586790/jobs/7425083584#step:14:823

At first glance I didn't find test_backend_reference.py was introduced in this PR. Since the tests for test_cast_FLOAT_to_BFLOAT16 only fails on Windows's x86 machine, I guess NumPy on on Windows's x86 machine might behave differently that others and causes this issue?

xadupre:
I added `test_backend_reference.py` to test the code in the backend running the light models (this section https://github.com/onnx/onnx/blob/main/onnx/backend/test/runner/__init__.py#L383). numpy does not know about bfloat16 and the backend end uses `to_array` to convert the values into numpy arrays and compare the expected values. I guess the bug could be partially resolved if this comparison would check the raw onnx values and not the numpy values. The reference implementation still misses a way to distinguish between bfloat16 and other types. The python API of onnxruntime as well unless `OrtValues` are used. I tried to adress that in the reference implementation if the PR related to F8 (https://github.com/onnx/onnx/pull/4805/files#diff-f6942c248259f3f996f990449ad45f656baabe1ad8bd047845a56137ca4d11b3R127, https://github.com/onnx/onnx/pull/4805/files#diff-80d356b8b3f7abd1c103d3fb1b8e8febf1286dff744bbdca1f7cc47fe21ed460R5). But I did not modify functions `to_array` and `from_array`. That would break backward compatibility even though I don't think many people are using these functions on bfloat16.

jcwchen:
Thanks @xadupre for the insight! To faster resolve the test failures on Windows x86 machine, shall we have another PR to disable those tests first if https://github.com/onnx/onnx/pull/4805 needs more time for review?

xadupre:
I think it would be better. #4805 is completly unrelated to that issue.

onnx/test/test_backend_reference.py
```diff
@@ -0,0 +1,152 @@
...
+    "|test_range_float_type_positive_delta_expanded"
+    "|test_range_int32_type_negative_delta_expanded"
+    "|test_scan_sum)"
+)
+
+# The following tests are about deprecated operators.
+backend_test.exclude("(test_scatter_with_axis|test_scatter_without)")
+
+
+# The following tests are too slow with the reference implementation (Conv).
```
yuanyao-nv, line 105: Is the recommendation now to exclude these models from routine backend testing? They do seem to be very slow:
```
9335.38s call     test_backend.py::OnnxBackendRealModelTest::test_vgg19_cuda
8523.93s call     test_backend.py::OnnxBackendRealModelTest::test_resnet50_cuda
6091.74s call     test_backend.py::OnnxBackendRealModelTest::test_densenet121_cuda
2532.54s call     test_backend.py::OnnxBackendRealModelTest::test_inception_v2_cuda
1631.90s call     test_backend.py::OnnxBackendRealModelTest::test_inception_v1_cuda
595.61s call     test_backend.py::OnnxBackendRealModelTest::test_squeezenet_cuda
530.00s call     test_backend.py::OnnxBackendRealModelTest::test_zfnet512_cuda
428.94s call     test_backend.py::OnnxBackendRealModelTest::test_shufflenet_cuda
200.11s call     test_backend.py::OnnxBackendRealModelTest::test_bvlc_alexnet_cuda
```

xadupre: I made a PR to fix that issue: https://github.com/onnx/onnx/pull/4960. Not merged yet. They are slow because the reference implementation of operator Conv is slow.

onnx/backend/test/runner/__init__.py
```diff
@@ -316,18 +369,66 @@ def run(test_self: Any, device: str) -> None:
...
+                ref = onnx.reference.ReferenceEvaluator(onx)
+                feeds = {}
+                inits = set(i.name for i in onx.graph.initializer)
+                n_input = 0
+                inputs = []
+                for i in range(len(onx.graph.input)):
+                    if onx.graph.input[i].name in inits:
+                        continue
+                    name = os.path.join(test_data_set, f"input_{n_input}.pb")
+                    inputs.append(name)
```
yuanyao-nv, line 397: Small fix to `input` here:https://github.com/onnx/onnx/pull/5026. Let me know if it looks good.

PR #4852, opened 2023-02-06
Update Linux-CI.yml. Ubuntu18.04 deprecated (unsupported by 2023/04/01)

andife:
"The Ubuntu 18.04 Actions runner image will begin deprecation on 2022/08/08 and will be fully unsupported by 2023/04/01"
https://github.com/actions/runner-images/issues/6002

Does python 3.7 work with ubuntu 20.04 ? should not make a difference?

Signed-off-by: Andreas Fehlner <fehlner@arcor.de>

PR #4867, opened 2023-02-08
Update description regarding Ubuntu in README.md

jcwchen:
### Description

Update description regarding Ubuntu in README.md since recently ONNX has upgraded its Ubuntu version in Linux Pipleine. In the future ONNX cannot promise it can work with Ubuntu 18.04.

### Motivation and Context

Follow up for https://github.com/onnx/onnx/pull/4852.


andife:
But if we really want and need this (18.04) we might as well run it in Docker containers? There we could continue to run 18.04? (good would be maybe a bit of an overhead?)

jcwchen:
> But if we really want and need this (18.04) we might as well run it in Docker containers? There we could continue to run 18.04? (good would be maybe a bit of an overhead?)

I think it should be fine without having that, because 18.04 will be globally EOL soon.

gramalingam (approved):

PR #4873, opened 2023-02-09
Use mypy 1.0.0

justinchuby:
### Description

Use mypy 1.0.0

### Motivation and Context

Faster lint


jcwchen (approved):
Is it OK to always use the latest mypy version?

justinchuby:
I think it is unless there are too many breaking changes we need to fix

.lintrunner.toml
```diff
@@ -95,7 +95,7 @@ init_command = [
     'run',
     'pip_init',
     '--dry-run={{DRYRUN}}',
-    'mypy==0.991',
+    'mypy==1.0.0',
```
jcwchen, line 98:
```suggestion
    'mypy',
```
> I think it is unless there are too many breaking changes we need to fix

Agreed. I was thinking directly use mypy here to always get the latest mypy. I thought mypy should be stable without introducing breaking changes and then we don't need to manually bump it every time.  What do you think?

justinchuby: I recommend that we pin a version so every run is repeatable. For example, we can check out any commit, run lintrunner and it should pass. The runner pip initializer (inherited from pytorch) enforces that we provide a version: https://github.com/justinchuby/lintrunner-adapters/blob/f67d4bb4a3061d64bfd73eef493b54f8d9e8169b/lintrunner_adapters/adapters/pip_init.py#L71-L75

PR #4875, opened 2023-02-09
Introduce scorecard.yaml (OpenSSF Scorecard - Security health metrics for Open Source)

andife:
Signed-off-by: Andreas Fehlner <fehlner@arcor.de>

### Description
Add the scorecard github action to monitor / improve security

### Motivation and Context
Currently we use or pass the "Best Practices" according to the Open Source Security Foundation (OpenSSF) by self-declaration. 

The project Insights of the Linux-Foundation https://insights-v2.lfx.linuxfoundation.org/lfai-onnx/overview shows points we can improve.

One point is "Security" and at the OpenSSF there is another project that deals with security that I would like to introduce. (https://github.com/ossf/scorecard)

I prefer the projects of OpenSSF compared to others because we already "use" OpenSSF and it is also a project of the Linux Foundation. 

When it is integrated, I will make other changes to make it better for example in our readme (to show the badge...) and to reduce the number of detections.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,72 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '24 10 * * 6'
+  push:
+    branches: [ "main" ]
+
+# Declare default permissions as read only.
+permissions: read-all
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
```
prasanthpul, line 35: why does it need the specific commit instead of latest version?

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@99c53751e09b9529366343771cc321ec74e9bd3d # v2.0.6
```
prasanthpul, line 40: same question as above about commit id

prasanthpul (changes requested):

PR #4876, opened 2023-02-09
Mark `VERSION_NUMBER` as final for 1.13.1 release

p-wysocki:
### Description
Mark the `VERSION_NUMBER` as the final 1.13.1 release. Please do not merge until agreed that we're ready to push to PyPI.


jcwchen (commented):
Please fix the lint error here:
```
===> check flake8...
  onnx/reference/ops/op_string_normalizer.py:89:17: B028 No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
```

It was not caught previously, because the lint CI has been updated recently. Thanks!

jcwchen (approved):
I think we can finalize this release since there isn't any regression.

PR #4880, opened 2023-02-09
Fix function shape inference bug

gramalingam:
Shape inference makes use of known constant values. The propagation of constant-values reuses tensors stored as attributes in Constant op nodes. This works fine for inference over the main graph. However, this is problematic for functions, since that creates copies of nodes (to do attribute-value substitution). Hence, these nodes are temporary nodes that do not persist for the lifetime of inference.

The fix is to make a copy of such tensors. We restrict this to functions to avoid unnecessary copies for the top-level graph (the most common use-case today). 

TODO: Even in the case of functions, we should ideally restrict copying of these tensors to small-sized tensors (typically shapes represented as tensors) using some cutoff size-limit.


onnx/shape_inference/implementation.cc
```diff
@@ -294,33 +294,41 @@ class ShapeInferenceImplBase {
...
             case AttributeProto::FLOATS: {
               std::vector<float> floats{attr.floats().begin(), attr.floats().end()};
-              addTemporaryConstant(n.output(0), floats);
+              addTemporaryConstant(output_name, floats);
               break;
             }
             case AttributeProto::FLOAT: {
               std::vector<float> floats({attr.f()});
-              addTemporaryConstant(n.output(0), floats);
+              addTemporaryConstant(output_name, floats);
```
jcwchen, line 331: Just out of curiosity: I saw you have handled the case of tensor and sparse_tensor with `reuse_constant_tensors` above(L301-310). Why don't we need to handle similar things in this else condition?

jcwchen (commented):
Is this PR for memory/performance improvement? Or it would fix a wrong behavior?

onnx/shape_inference/implementation.cc
gramalingam: They already do it. The main fix is to create a temporary-constant (of the TensorProto, which is stored in `input_data_by_name_holder`), which already happens in `addTemporaryConstant`.

This is to fix wrong behavior. The problem is that we are storing the address of `attr.t()`, but this is stack-allocated temporary (see [this line](https://github.com/onnx/onnx/blob/main/onnx/shape_inference/implementation.cc#L552)) which will disappear once the Constant node is processed.

gramalingam: The if-condition is for performance. We don't need the fix when doing inference for a Graph, only when we do inference for a Function, which is when we create the temporary NodeProto on stack.

jcwchen (approved):
LGTM. Thank you for fixing!

PR #4881, opened 2023-02-09
add sample and test case for binarizer ml op

liqunfu:

### Description
sample and test case for binarizer ml op are missing. add them.

### Motivation and Context
shows how to implement the op and help runtime to validate its implementation.


.azure-pipelines/Linux-CI.yml
```diff
@@ -108,14 +108,21 @@ jobs:
...
+      if [ '$(onnx_ml)' == '1' ]; then
+        export exclude_onn_ml_filter=''
+      else
+        export exclude_onn_ml_filter=':!*/data/node/test_ai_onnx_ml_*'
+      fi
+
       # Skip *output_*.pb because NumPy functions might behave differently on different platforms
-      # Skip test_log's input.pb because it uses np.random, which might behave differently on different platforms
-      git diff --exit-code -- . ':!onnx/onnx-data.proto' ':!onnx/onnx-data.proto3' ':!*output_*.pb' ':!*input_*.pb'
+      # Skip test_log's input.pb because it uses np.random, which might behave differently on different platforms
```
xadupre, line 118: extra spaces to remove

onnx/reference/ops/aionnxml/op_binarizer.py
```diff
@@ -1,16 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 # pylint: disable=R0913,R0914,W0221
 
-import numpy as np
-
+from ....backend.test.case.node.ai_onnx_ml.binarizer import compute_binarizer
```
xadupre, line 4: We should define which subfolder must import the other one to avoid circular import. I chose to avoid importing anything from the test case because the implementation was not always complete or even available. That's why I kept subfolders reference and backend completely separated.

.azure-pipelines/Linux-CI.yml
```diff
@@ -108,14 +108,21 @@ jobs:
       source venv/bin/activate
       python onnx/backend/test/cmd_tools.py generate-data --clean
       git status
+      if [ '$(onnx_ml)' == '1' ]; then
+        export exclude_onn_ml_filter=''
+      else
+        export exclude_onn_ml_filter=':!*/data/node/test_ai_onnx_ml_*'
```
jcwchen, line 114: May I understand why we would like to ignore the added test directory (`/data/node/test_ai_onnx_ml_*'`) entirely? Typically we only exclude output_*.pb, because np.random will possibly produce different result from different platforms. I thought the model itself should always be consistent?

jcwchen (commented):
May I understand why this test needs to be

.azure-pipelines/Linux-CI.yml
liqunfu: test named with test_ai_onnx_ml_*  (will put into a folder as @xadupre suggested one this code passes) can only get generated when ONNX_ML is set. So we need to exclude git diff test on these files is ONNX_ML is not set because they will be shown as deleted instead of unchanged.

jcwchen: Thank you for the context. If that is the case, I am inclined to change the logic in the generate-data tool instead. For instance, do not delete those directories with "test_ai_onnx_ml_*" prefix when ONNX_ML=0. It is one behavior change in tool vs a few modifications in relevant pipelines.

jcwchen (commented):
Another question: could we use a shorter prefix name rather than `test_ai_onnx_ml_*`? Long filename sometimes is headache on old Windows: https://github.com/onnx/onnx/issues/2821.

onnx/backend/test/case/node/ai_onnx_ml/binarizer.py
```diff
@@ -0,0 +1,34 @@
...
+
+import numpy as np
+
+import onnx
+
+from ...base import Base
+from .. import expect
+
+
+def compute_binarizer(x, threshold=None):
```
gramalingam, line 11: Is it possible to call the reference-implementation (in `onnx/reference/ops/aionnxml`)?

onnx/backend/test/case/node/ai_onnx_ml/binarizer.py
```diff
@@ -0,0 +1,34 @@
...
+    y = x.copy()
+    cond = y > threshold
+    not_cond = np.logical_not(cond)
+    y[cond] = 1
+    y[not_cond] = 0
+    return (y,)
```
gramalingam, line 17: Would `return np.where(x > threshold, 0, 1)` work? Or, `return (x <= threshold).astype(int)`, using whatever desired target type is (int32 or int64).

onnx/defs/traditionalml/defs.cc
```diff
@@ -85,7 +85,11 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
             "T",
             {"tensor(float)", "tensor(double)", "tensor(int64)", "tensor(int32)"},
             "The input must be a tensor of a numeric type. The output will be of the same tensor type.")
-        .Attr("threshold", "Values greater than this are mapped to 1, others to 0.", AttributeProto::FLOAT, 0.f));
+        .Attr("threshold", "Values greater than this are mapped to 1, others to 0.", AttributeProto::FLOAT, 0.f)
+        .TypeAndShapeInferenceFunction([](InferenceContext& ctx) {
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          *ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape() = getInputShape(ctx, 0);
```
gramalingam, line 91: I think `propagateShapeAndTypeFromFirstInput(ctx)` should work? It will also check if input shape is known before copying it.

onnx/backend/test/case/node/ai_onnx_ml/binarizer.py
gramalingam: Ok, I see the output type needs to be same as input type. So, perhaps `return (x < threshold).astype(x.dtype)` or something like that? I see this comes from the reference implementation, so anyway, this is a suggestion for that. As mentioned above, we can directly reference that from here.

onnx/reference/ops/aionnxml/op_binarizer.py
gramalingam: Can we have the test-cases use the reference-implementation?

onnx/backend/test/case/node/ai_onnx_ml/binarizer.py
liqunfu: I was thinking that it is better to put reference implementation as close to the sample as possible. Now that noth @xadupre and @gramalingam commented the same. I must be wrong for this. I will move the implementation back to where it is - the reference implementation location.

.azure-pipelines/Linux-CI.yml
liqunfu: good point! @jcwchen it makes this PR much cleaner.

onnx/backend/test/case/node/ai_onnx_ml/binarizer.py
liqunfu: code updated according to @gramalingam 's recommendation

onnx/defs/traditionalml/defs.cc
liqunfu: code updated according to @gramalingam 's recommendation

xadupre (approved):

yuanyao-nv:
Cross linking the issue I posted on the backend-scoreboard repo: https://github.com/onnx/backend-scoreboard/issues/55
Currently, the changes in this MR cause these tests to be automatically picked up for any scoreboard pipeline. That's probably not what we want as most backends do not support the ML ops. Should we prevent the backend tests from picking up the tests for ML ops?

PR #4882, opened 2023-02-10
Add one backend test for ArrayFeatureExtractor

xadupre:
### Description
Add one backend test for ArrayFeatureExtractor.


liqunfu (commented):
it requires update of test tool and CI in order to add test cases from domain other then onnx. There is another PR trying the same. May be we shall let one proceed first in order to avoid duplicated work?
https://github.com/onnx/onnx/pull/4881

xadupre:
I'll wait for your PR to be merged first then. Your changes related to CI seems to be more advanced than mine. We did not work on the same operator so this should be fine. I added a couple of comments on #4881.

gramalingam (approved):

gramalingam:
The other PR has been merged. So, I suppose we should be able to go ahead with this?

xadupre:
Yes, I'll submit other PR for other ml operators.

PR #4887, opened 2023-02-13
Fix squeeze opset 11 reference runtime bug

HarmonyHu:
squeeze in opset 11 should getattr(self, "axes"), or shape would be wrong. For example, [1, 5, 1, 5] with axes[2], will get wrong output shape.

Signed-off-by: pengchao.hu <pengchao.hu@sophgo.com>

### Description
For squeeze in opset 11, get axes by getattr.

### Motivation and Context
Many Onnx model use opset 11, and if it contains squeeze, shape inference is not correct as no axes.

### DIFF
--- a/onnx/reference/ops/op_squeeze.py
+++ b/onnx/reference/ops/op_squeeze.py
@@ -8,6 +8,8 @@ from onnx.reference.op_run import OpRun
 
 class Squeeze_1(OpRun):
     def _run(self, data, axes=None):  # type: ignore
+        if axes is None:
+            axes = getattr(self, "axes", axes)
         if isinstance(axes, np.ndarray):
             axes = tuple(axes)
         elif axes in [[], tuple()]:


daquexian:
Thanks for your contribution! Could you please describe the case that the main branch fails to handle? 

IMO the value of attribute 'axes' is passed as the argument of `_run` method, so the original code should be correct.

HarmonyHu:
> Thanks for your contribution! Could you please describe the case that the main branch fails to handle?
> 
> IMO the value of attribute 'axes' is passed as the argument of `_run` method, so the original code should be correct.

[squeeze_test.onnx.zip](https://github.com/onnx/onnx/files/10718664/squeeze_test.onnx.zip)
test as below by attached onnx model:
``` python
import onnx
import numpy as np
from onnx.reference.reference_evaluator import ReferenceEvaluator
x = np.random.rand(1, 3, 100, 1, 100).astype(np.float32)
y = np.random.rand(1, 3, 100, 1, 100).astype(np.float32)
sess = ReferenceEvaluator("squeeze_test.onnx", verbose=1)
results = sess.run(None, {"input0":x,"input1":y})
print(results)
```

xadupre:
Your example with the current version in main branch. `axes` is automatically replaces by the value stored in the onnx node. Your change should not be needed.

HarmonyHu:
no need

PR #4888, opened 2023-02-13
[Fix typo] Update Operators.md

jiahy0825:
Signed-off-by: HongyuJia <jiahongyu@baidu.com>

### Description
Fix typo: `negatve->negative`.

### Motivation and Context
Solve typo errors.

### DIFF
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -25925,9 +25925,9 @@ expect(node, inputs=[x], outputs=[y], name="test_size")
   If `steps` are omitted, they are set to `[1, ..., 1]` of length `len(starts)`
 
   The effective values are initialized as `start[i] = 0`, `end[i] = dims[i]` where
-  `dims` are the dimensions of `input` and `step[i] = `1.
+  `dims` are the dimensions of `input` and `step[i] = 1`.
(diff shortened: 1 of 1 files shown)


docs/Operators.md
```diff
@@ -25925,9 +25925,9 @@ expect(node, inputs=[x], outputs=[y], name="test_size")
   If `steps` are omitted, they are set to `[1, ..., 1]` of length `len(starts)`
 
   The effective values are initialized as `start[i] = 0`, `end[i] = dims[i]` where
-  `dims` are the dimensions of `input` and `step[i] = `1.
+  `dims` are the dimensions of `input` and `step[i] = 1`.
```
xadupre, line 25928: You need to apply the same modification here: https://github.com/onnx/onnx/blob/main/onnx/defs/tensor/defs.cc#L762 (or recompile onnx and run *onnx/defs/gen_doc.py* with flag `ONNX_ML=0`).

gramalingam: It is best to change the source and run gen_doc.py (as mentioned by Xavier above). That will ensure all files are updated.

PR #4893, opened 2023-02-14
Dependabot version updates

prasanthpul:
Using dependabot to handle version updates is a recommended best practice


andife:
We already have dependabot enabled, but in terms of transparency, I also prefer the configuration through a yml file. I just wonder how does the file play together with our current configuration 

at https://github.com/onnx/onnx/security Dependabot alerts - Enabled
Get notified when one of your dependencies has a vulnerability

jcwchen (approved):

andife (approved):

PR #4894, opened 2023-02-14
Pin actions to a full length commit SHA

andife:
### Description
Pin actions to https://github.com/step-security/secure-repo#2-pin-actions-to-a-full-length-commit-sha 

### Motivation and Context

- Get a higher score at scorecard at https://securityscorecards.dev/ (https://api.securityscorecards.dev/projects/github.com/onnx/onnx) 
- Not fixed at the release pipeline so far... (would update also this after possible discussion)


andife:
For newer version of github actions I suggest (https://github.com/onnx/onnx/pull/4901).

jcwchen (commented):
If we want to do so, just do it globally (including release CIs)? I am fine with this update. Thanks!

andife:
> If we want to do so, just do it globally (including release CIs)? I am fine with this update. Thanks!

I updated also the release CIs.

PR #4901, opened 2023-02-15
Get Update if new github actions are available (Update dependabot.yml)

andife:
- Disable version updates for python-deps (numpy, protobuf... we only want security updates there?)
- Show version updates for github actions (documentation here... https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#configuration-options-for-the-dependabotyml-file)

=> Does also relate to https://github.com/onnx/onnx/pull/4894 (fixating the used there.... and get's update if there is a new version available)


jcwchen (commented):
In principle, I don't think we need dependabot to remind us there is a new version, but if there are some security issues in certain old version, I hope dependabot can warn us and recommend the target updated version. To confirm, is it what this PR wants to achieve?

andife:
> In principle, I don't think we need dependabot to remind us there is a new version, but if there are some security issues in certain old version, I hope dependabot can warn us and recommend the target updated version. To confirm, is it what this PR wants to achieve?

**For the pip setting "open-pull-requests-limit: 0" means that we get only security updates
We have only specific versions for numpy and protobuf and we want to handle / update it according to our criteria.
Therefore we don't need notifications if there are new versions in general. 

As we would pin our github actions to a "full length commit SHA", we would not get any updates (unlike when we specify a version number) automatically. 
But since we run the pipelines very often I would like to benefit from the improvements and in case of doubt decide for myself whether we take the update or not. 

We would keep control over the updates and not get new versions in the background without noticing it. Merging the proposed updates, then only leads to the original behavior, only that we have more control over it. I would like to try it out and evaluate it later.  **

jcwchen (approved):

PR #4904, opened 2023-02-15
Update action workflows to set default permission to read

prasanthpul:
set top-level default permissions as security best practice.

Per https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions, by setting permission for any context, all other unspecific contexts default to None


andife:
Is the green tick enough to see that the run was successful? I would already think that this fits.

But how do we treat the skipped CIs? Can we somehow run them manually on the PR? Possibly something like issue_comment (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only) would be interesting?

jcwchen (approved):
> But how do we treat the skipped CIs? Can we somehow run them manually on the PR? Possibly something like issue_comment (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only) would be interesting?

The current method is manually adding labels as run release CIs or test ONNX Model Zoo to trigger those skipped CIs. I have added run release CIs and they seem fine.

andife (approved):

PR #4906, opened 2023-02-15
[Non functional changes] Added links to the SIG and Working group pages to reduce redundancy

AlexandreEichenberger:
Added reference to new compiler SIG, and added a direct link from the list of sigs in this repo to the equivalent SIG directory in the SIG repo, for better flow of info.

If you don't like this second change, I will remove these new links.


prasanthpul:
There is similar info at https://github.com/onnx/sigs. To avoid duplications, should we turn this file in onnx/onnx into just a link to the sigs/readme?

andife:
> There is similar info at https://github.com/onnx/sigs. To avoid maintaining in 2 places, should we turn this file in onnx/onnx into just a link to the sigs/readme?

We also had another place: https://github.com/onnx/onnx/tree/main/community#sig---special-interest-groups => maybe we should only list the name of the sigs there and remove the description there (and add a link to the place of the description...)

Am unsure, I probably would have left it the description in onnx/onnx in the "community/governance" directory since it is some sort of approval of the terms of reference by the steering committee.... then in onnx/sig i would have rather seen the minutes and work products.

gramalingam:
I like the idea of consolidating the content to one place and having a link in other places: wherever we decide to place the content.

AlexandreEichenberger:
FYI, the comments overwhelmingly recommend to reduce redundancy. I figured that the SIGS and the Working groups would be best placed to know their missions and/or which one is active or not, so it made sense to me to have the onnx repo have links to these specialized repos on SIGs and Working groups.

One more link to follow... indirections is the solution to all CS problems :-)

andife (approved):

PR #4907, opened 2023-02-16
Support use of `Path` objects in `_get_file_path`

JCRPaquin:
### Description

- Fix a minor issue with how `_get_file_path`, used for loading external data, handles `Path` objects.

### Motivation and Context

While debugging an issue with fastT5 I noticed that onnx was improperly handling an absolute path stored in a `PosixPath`. Under the current implementation, when you pass an absolute `PosixPath` into `_get_file_path` it will produce a new absolute path relative to the current working directory. For example, "/home/ubuntu/models/ul2-encoder.onnx" would become "/home/ubuntu/ul2-encoder.onnx" under the current implementation. This causes issues downstream when loading external data: because external data is relative to the model directory, at least for onnxruntime(?), the data won't be found.


onnx/__init__.py
```diff
@@ -87,7 +88,7 @@ def _save_bytes(content: bytes, f: Union[IO[bytes], str]) -> None:
 
 
 def _get_file_path(f: Union[IO[bytes], str]) -> Optional[str]:
```
daquexian, line 90: Please also update the type hint of f

daquexian (commented):
thanks for your contribution!

onnx/__init__.py
```diff
@@ -86,8 +87,8 @@ def _save_bytes(content: bytes, f: Union[IO[bytes], str]) -> None:
             writable.write(content)
 
 
-def _get_file_path(f: Union[IO[bytes], str]) -> Optional[str]:
-    if isinstance(f, str):
+def _get_file_path(f: Union[IO[bytes], str, Path]) -> Optional[str]:
+    if isinstance(f, str) or isinstance(f, Path):
```
daquexian, line 91: nit:
```suggestion
    if isinstance(f, (str, Path)):
```

daquexian (approved):

onnx/__init__.py
JCRPaquin: Fair point. I'll add this in after merging again.

xadupre:
A unit test should be added to check the changes.

JCRPaquin:
> A unit test should be added to check the changes.

Upon further reflection I am worried that folks may have programmed around this issue; is there a way to verify this doesn't break dependent packages? Like a step in CI to pull large dependent packages and run their tests?

I'll add a regression test with a model + external data later today or tomorrow.

onnx/test/helper.py
```diff
@@ -0,0 +1,23 @@
+import unittest
```
JCRPaquin, line 1: Wasn't sure where to put this; I'm open to moving/renaming the file.

onnx/test/test_external_data.py
```diff
@@ -672,5 +674,28 @@ def test_check_model_absolute(self) -> None:
...
+
+    @skip_if_no_onnxruntime
+    def test_file_path_is_correct_during_quantization(self) -> None:
+        from onnxruntime.quantization import quantize_dynamic, QuantType
+        
+        prev_dir = os.getcwd()
+        os.chdir(os.path.dirname(self.temp_dir))
+
+        try:
+            quantize_dynamic(
```
JCRPaquin, line 690: Using `quantize_dynamic` here because that's where the initial issue occurred. `onnxruntime` has since patched this on their end by converting all `Path` objects to strings before making calls to `onnx`.

onnx/test/test_external_data.py
```diff
@@ -672,5 +674,28 @@ def test_check_model_absolute(self) -> None:
...
+    def test_load_from_path_object(self) -> None:
+        model_path = Path(self.model_filename)
+        model = onnx.load_model(model_path)
+
+    @skip_if_no_onnxruntime
+    def test_file_path_is_correct_during_quantization(self) -> None:
+        from onnxruntime.quantization import quantize_dynamic, QuantType
+        
+        prev_dir = os.getcwd()
+        os.chdir(os.path.dirname(self.temp_dir))
```
JCRPaquin, line 687: Changing directories to make sure the loader isn't treating absolute paths as relative.

onnx/shape_inference.py
```diff
@@ -74,6 +74,7 @@ def infer_shapes_path(
         if output_path == "":
             output_path = model_path
         C.infer_shapes_path(model_path, output_path, check_type, strict_mode, data_prop)
+        return
```
JCRPaquin, line 77: Included in this PR because the test with `quantize_dynamic` failed on this code.

JCRPaquin:
I can't reproduce the build failures locally in a container; the `onnx.test.helper` module resolves when testing locally. I'll try with an ADO container in a bit to see what I might be doing wrong. I was testing with an ubuntu:22.04 container with Python 3.10.

onnx/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 # isort:skip_file
 import os
```
justinchuby, line 4: Prefer importing modules only
——
I would avoid importing a class or a function from a module.

Import only the module instead to keep the name space clean and provide readers with more context on where its members come from. For example, readers don’t need to backtrack to see `sleep` is a function from `time`, as opposed to a function defined in the file. https://google.github.io/styleguide/pyguide.html#22-imports

onnx/test/test_external_data.py
```diff
@@ -672,5 +674,28 @@ def test_check_model_absolute(self) -> None:
...
+
+        try:
+            quantize_dynamic(
+                model_input=self.model_filename,
+                model_output=os.path.join(self.temp_dir, f"quantized-{str(uuid.uuid4())}.onnx"),
+                weight_type=QuantType.QInt8,
+                use_external_data_format=True,
+            )
+        finally:
+            os.chdir(prev_dir)
```
justinchuby, line 697: Can this be done with a context manager ?

JCRPaquin: There's no built-in one for this; should I add one or maybe there's one that I couldn't find?

justinchuby:
I would just create oneLe 26 févr. 2023 à 04:05, Jean-Claude Paquin ***@***.***> a écrit :﻿
@JCRPaquin commented on this pull request.



In onnx/test/test_external_data.py:
> +    @skip_if_no_onnxruntime
+    def test_file_path_is_correct_during_quantization(self) -> None:
+        from onnxruntime.quantization import quantize_dynamic, QuantType
+        
+        prev_dir = os.getcwd()
+        os.chdir(os.path.dirname(self.temp_dir))
+
+        try:
+            quantize_dynamic(
+                model_input=self.model_filename,
+                model_output=os.path.join(self.temp_dir, f"quantized-{str(uuid.uuid4())}.onnx"),
+                weight_type=QuantType.QInt8,
+                use_external_data_format=True,
+            )
+        finally:
+            os.chdir(prev_dir)

There's no built-in one for this; should I add one or maybe there's one that I couldn't find?

onnx/__init__.py
```diff
@@ -86,8 +87,8 @@ def _save_bytes(content: bytes, f: Union[IO[bytes], str]) -> None:
             writable.write(content)
 
 
-def _get_file_path(f: Union[IO[bytes], str]) -> Optional[str]:
-    if isinstance(f, str):
+def _get_file_path(f: Union[IO[bytes], str, Path]) -> Optional[str]:
```
justinchuby, line 90: Consider using os.PathLike?

justinchuby:
Already supported. Thanks!

PR #4911, opened 2023-02-16
proto to support attributes with default values

liqunfu:
### Description
attributes may have default values, current proto does not serialize default values

### Motivation and Context
add support of default values in proto.


justinchuby:
Excited for this

docs/Syntax.md
```diff
@@ -69,7 +69,7 @@ The grammar below describes the syntax:
    graph ::= id value-info-list '=>' value-info-list node-list
    other-data ::= id ':' value
    other-data-list ::= '<' other-data (',' other-data)* '>'
-   fun-attr-list ::= '<' id-list '>'
+   fun-attr-list ::= '<' id | attr (, id | attr) '>'
```
xadupre, line 72: What about str for attributes without values?

onnx/defs/parser.cc
```diff
@@ -106,6 +106,32 @@ Status OnnxParser::Parse(char open, IdList& idlist, char close) {
...
 
+Status OnnxParser::Parse(IdList& idlist, AttrList& attrlist) {
+  idlist.Clear();
+  attrlist.Clear();
+  do {
+    std::string id;
+    ParseIdentifier(id);
+    auto next = NextChar();
+    if (next == ':' || next == '=')
+      Parse(*attrlist.Add(), id);
```
xadupre, line 117: Should we forbid the use of attribute without default values after one attribute with a default value?

onnx/test/parser_test.py
```diff
@@ -72,6 +72,49 @@ def test_parse_model_error(self) -> None:
...
+            ir_version: 9,
+            opset_import: [ "" : 15, "custom_domain" : 1],
+            producer_name: "FunctionProtoTest",
+            producer_version: "1.0",
+            model_version: 1,
+            doc_string: "A test model for model local functions."
+          >
+         agraph (float[N] x) => (float[N] out)
+         {
+            out = custom_domain.Selu<gamma=2.0, gamma=3.0>(x)
```
xadupre, line 87: gamma or alpha? Maybe this should raise an error.

xadupre:
make_function should be updated as well.

onnx/test/cpp/parser_test.cc
```diff
@@ -488,6 +488,34 @@ square (x) => (y) {
...
   CheckModel(code);
+
+  const char* code_function_with_attributes = R"ONNX(
+<
+  ir_version: 9,
+  opset_import: [ "" : 15, "custom_domain" : 1]
+>
+agraph (float[N] x) => (float[N] out)
+{
+  out = custom_domain.foo<alpha=2.0, gamma=3.0>(x)
```
xadupre, line 499: We should also have an example with any value specified for alpha.

onnx/defs/parser.cc
justinchuby: Does order matter for attributes?

docs/Syntax.md
gramalingam: A * is missing after the )

onnx/defs/parser.cc
gramalingam: No, I think it is better to leave the order be immaterial. So, okay to be more lenient than strict here, IMO.

gramalingam:
I think it would be useful to update `IR.md` document as well, and explain updates for this IR version. That is, I think we should explain the new changes, and we should also clarify what IR version supports it.

onnx/test/parser_test.py
gramalingam: I guess it was meant to be alpha. I think/believe the `check_model` should catch the error. If not, we need to fix it.

onnx/test/cpp/parser_test.cc
gramalingam: You mean an example with no value specified for alpha? Yes, agree. (But the important step is to extend the checker and inference methods to support this. It will mostly work fine for parsing, but not for checking/inference.)

gramalingam:
Are the test-data checked in because these models use the updated IR version? Not sure we want to do that, right?

gramalingam:
It seems to me that we may need to update the checker/inference code to use these new fields for model-local functions?

docs/Syntax.md
liqunfu: is that this line for?
repeated string attribute = 6;
and in such case, it will be:
fun-attr-list ::= '<' id (, id)* '>'
which is a subset of the one defined at line 72.

onnx/defs/parser.cc
liqunfu: > Does order matter for attributes?

not in the proto. So w.r.t. @xadupre 's comment, I agree with @gramalingam  that we do not (or cannot) check for this rule.

docs/Syntax.md
gramalingam: It used to be for "repeated string attribute". Now, it is for both that line and the "repeated AttributeProto" you added.

onnx/test/parser_test.py
liqunfu: checker does not check for schema error if it is a custom op. I added a check for this type of errors when check a node. I will create another pr to verify consistency between customer ops and local functions in general.

liqunfu:
> Are the test-data checked in because these models use the updated IR version? Not sure we want to do that, right?

The test cases are updated because IR version bump. I would wait for https://github.com/onnx/onnx/pull/4805 in order to reduce merge conflicts. However, if this PR need to be merged first, I will work on this test cases.

onnx/checker.cc
```diff
@@ -622,7 +622,14 @@ void check_node(const NodeProto& node, const CheckerContext& ctx, const LexicalS
   }
   auto domain_version = dit->second;
 
+  // for ops referencing local functions, there is no schame to verify it.
```
rajeevsrao, line 625: typo: s/schame/schema

onnx/test/cpp/parser_test.cc
liqunfu: added test for combination of with or without values.

onnx/test/parser_test.py
```diff
@@ -116,5 +116,74 @@ def test_parse_function_with_attributes(self) -> None:
...
+         }
+        """,
+        ]
+        # graph_text = """
+        # agraph (float[N] x) => (float[N] out)
+        #  {
+        #     out = custom_domain.Selu(x)
+        #  }
+        # """
+        for graph_text in graph_texts:
```
justinchuby, line 180: Can this be parameterized? https://pypi.org/project/parameterized/

yuanyao-nv:
I'm curious how default values will be set for attributes that reply on input shapes. For example in `Conv`, the `pads` and `strides` attributes are lists whose lengths depend on the input dimensions. Will we always know the input dimensionality when serializing the proto?
cc: @rajeevsrao

docs/Syntax.md
abock: The `,` should be quoted (`','`) for the 2+ case:

```
fun-attr-list ::= '<' id | attr (',' id | attr) '>'
```

onnx/test/parser_test.py
```diff
@@ -116,5 +116,74 @@ def test_parse_function_with_attributes(self) -> None:
...
+         }
+        """,
+        """
+        agraph (float[N] x) => (float[N] out)
+         {
+            out = custom_domain.Selu<alpha=2.0, gamma=3.0>(x)
+         }
+        """,
+        ]
+        # graph_text = """
```
gramalingam, line 174: Delete these commented lines?

onnx/test/parser_test.py
```diff
@@ -116,5 +116,74 @@ def test_parse_function_with_attributes(self) -> None:
...
+        #  {
+        #     out = custom_domain.Selu(x)
+        #  }
+        # """
+        for graph_text in graph_texts:
+            graph = onnx.parser.parse_graph(graph_text)
+            opset_imports = [OperatorSetIdProto(domain = "", version = 15), OperatorSetIdProto(domain = "custom_domain", version = 1)]
+
+            model = onnx.helper.make_model(graph, functions= functions, opset_imports=opset_imports)
+            checker.check_model(model)
```
gramalingam, line 185: Nit: I am not sure this check belongs to `parser_test.py`. It probably belongs to `checker_test.py` or `model_inference_test.py`, depending on what we want to check.

onnx/test/parser_test.py
```diff
@@ -116,5 +116,74 @@ def test_parse_function_with_attributes(self) -> None:
...
+               neg = Mul(gamma_x, alpha_x_exp_x_)
+               pos = Mul(gamma_x, X)
+               _zero = Constant<value_float=0.0>()
+               zero = CastLike(_zero, X)
+               less_eq = LessOrEqual(X, zero)
+               C = Where(less_eq, neg, pos)
+           }
+        """
+
+        functions = [onnx.parser.parse_function(function_text)]
```
gramalingam, line 146: May be useful to check/assert that the returned FunctionProto has the expected attributes.

gramalingam: At least check `len(f.attribute_protos)`.

gramalingam:
> I'm curious how default values will be set for attributes that reply on input shapes. For example in `Conv`, the `pads` and `strides` attributes are lists whose lengths depend on the input dimensions.

@yuanyao-nv That's a good question. The purpose of this extension is to allow users to define their own functions, which are stored in a model. We cannot encode all possible usage scenarios using this feature. But it works as long as we can find a default-value that  is not a legitimate attribute-value, since we can encode the desired behavior by checking if the attribute-value equals special-value.

Please take a look at [this PR](https://github.com/onnx/onnx/pull/4525) that was proposed as one general mechanism. Unfortunately, it has some complications, because of the way the ONNX checker works. It is hard to write and if-then-else in ONNX where one of the branches (that will not execute) violates some ONNX check/constraint. So, this PR was reverted.

If you have any alternative suggestions, that would be great and welcome. Thanks!

onnx/test/parser_test.py
liqunfu: https://github.com/onnx/onnx/issues/4957

liqunfu: I will create a pr to fix the issue

onnx/test/parser_test.py
liqunfu: With more verifications add for output form the parser, I think it is now more a parser test and can stay in this file.

onnx/test/parser_test.py
liqunfu: good point! code updated

onnx/test/parser_test.py
liqunfu: added verification for parser outcomes.

gramalingam: I don't see it. May not be checked in yet?

onnx/test/parser_test.py
```diff
@@ -115,18 +117,63 @@ def test_parse_function_with_attributes(self) -> None:
...
          domain: "custom_domain",
          opset_import: [ "" : 15],
          doc_string: "Test function proto"
          >
            Selu
-           <alpha: float=1.67326319217681884765625, gamma: float=1.05070102214813232421875>
+           <alpha: float={default_alpha}, gamma: float={default_gamma}>
            (X) => (C)
-           {
+           {{
```
gramalingam, line 176: Why 2 braces? I must be missing something ...

gramalingam: Oh, because of the formatting `f` ...

gramalingam (approved):

Issue #4872, opened 2023-02-09
`check_function` does not raise when branches of the if statement return a different type


justinchuby:
# Bug Report

### Is the issue related to model conversion?

No

### Describe the bug

ONNX checker does not raise when branches of the if statement return a different type.

https://github.com/microsoft/onnx-script/pull/421#discussion_r1100820582

### System information

ONNX 1.14

### Reproduction instructions

```python
@onnxscript.script()
def aten_arange_start(
    start, end, dtype: int = -1
) -> TensorType:
    # arange.start(Scalar start, Scalar end, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor

    # Cast input to double if dtype is specified, because the input dtype may be e.g. bool
    # which Range does not support. The output type is ensured because the output
    # is casted to the specified dtype.
    if dtype != -1:
        start = op.Cast(start, to=DOUBLE.dtype)
        end = op.Cast(end, to=DOUBLE.dtype)

    result = op.Range(start, end, 1)
    if dtype != -1:
        result = op.Cast(result, to=dtype)

    return result

onnx.checker.check_function(aten_arange_start.to_function_proto())
```

### Expected behavior

`onnx.checker.check_function()` should raise

### Notes

cc @gramalingam @jcwchen

gramalingam:
Yes, not surprising. ONNX has two parts that check validity: the checker and the type/shape inference. The checker itself does only basic sanity checks (ensuring protobuf objects have expected fields defined, etc.) The type/shape inference based checker is required to catch type-based errors.

I have been working on exposing the second functionality to enable checking functions for type correctness. I will create a PR for this soon.

PR #4912, opened 2023-02-16
Add method to type-check a function

gramalingam:
### Description
Add method to type-check a function, given types of its inputs and its attribute values.

### Motivation and Context

Enables checking a function-body in a specific context for type-validity.

Eg., see: https://github.com/onnx/onnx/issues/4872


onnx/shape_inference/implementation.cc
```diff
@@ -649,6 +649,10 @@ class ShapeInferenceImplBase {
     }
   }
 
+  const std::vector<std::string>& Errors() const {
```
jcwchen, line 652: Not sure the naming convention for getting reference here -- shall we use getErrors() instead?

jcwchen (commented):
It would be great to expose this API on the website: https://github.com/onnx/onnx/blob/main/docs/docsgen/source/api/shape_inference.rst

onnx/shape_inference.py
```diff
@@ -131,4 +131,23 @@ def infer_node_outputs(
     return {key: onnx.TypeProto.FromString(out) for key, out in outputs.items()}
 
 
+def infer_function_output_types(
```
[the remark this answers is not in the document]
gramalingam: The name is longer, but I prefer that since it is more precise. `infer_function_type`, in my mind, sounds different. To me it suggests inferring a type-signature for the function. The key difference is that we are doing the inference for a specific calling-context (as opposed to inferring a type that is valid for all calling-contexts). Hard to convey the distinction, but the current name is the best I can come up with.

onnx/cpp2py_export.cc
```diff
@@ -448,6 +448,38 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
 
+  shape_inference.def(
+      "infer_function_output_types",
+      [](const py::bytes& function_proto_bytes,
+         const std::vector<py::bytes> input_types_bytes,
+         const std::vector<py::bytes> attributes_bytes) -> std::vector<py::bytes> {
+        FunctionProto proto{};
+        ParseProtoFromPyBytes(&proto, function_proto_bytes);
+
+        std::vector<TypeProto> input_types;
```
xadupre, line 459: use reserve

onnx/cpp2py_export.cc
```diff
@@ -448,6 +448,38 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+        for (const py::bytes& bytes : input_types_bytes) {
+          TypeProto type;
+          ParseProtoFromPyBytes(&type, bytes);
+          input_types.push_back(type);
+        }
+
+        std::vector<AttributeProto> attributes;
```
xadupre, line 466: reserve

onnx/cpp2py_export.cc
```diff
@@ -448,6 +448,38 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+        for (const py::bytes& bytes : attributes_bytes) {
+          AttributeProto attr;
+          ParseProtoFromPyBytes(&attr, bytes);
+          attributes.push_back(attr);
+        }
+
+        std::vector<TypeProto> output_types = shape_inference::InferFunctionOutputTypes(proto, input_types, attributes);
+        std::vector<py::bytes> result;
+        for (auto& type_proto : output_types) {
```
xadupre, line 475: reserve

onnx/shape_inference/implementation.cc
```diff
@@ -823,6 +827,104 @@ void InferShapeForFunctionNode(
...
+  const TypeProto* getInputType(size_t index) const override {
+    return (index < input_types_.size()) ? &input_types_[index] : nullptr;
+  }
+
+  TypeProto* getOutputType(size_t index) override {
+    return (index < output_types_.size()) ? &output_types_[index] : nullptr;
+  }
+
+  GraphInferencer* getGraphAttributeInferencer(const std::string& attribute_name) override {
+    (void)attribute_name; // This method is unused for function-type-inference.
```
xadupre, line 870: What's the purpose of this line?

gramalingam: It fails the lint checker because of unused parameter. If you have a better suggestion, please let me know.

gramalingam: Replaced with macro used elsewhere.

onnx/shape_inference/implementation.cc
gramalingam: done

docs/PythonAPIOverview.md
```diff
@@ -260,6 +260,35 @@ onnx.shape_inference.infer_shapes_path("path/to/the/model.onnx", "output/inferre
...
 ```
 
+## Running Type Inference on an ONNX Function
+
+```python
+import onnx
+import onnx.helper
+import onnx.parser
+import onnx.shape_inference
+from onnx import AttributeProto, TypeProto
```
jcwchen, line 270:
```suggestion
```
These two seem unused.

jcwchen (approved):

docs/PythonAPIOverview.md
gramalingam: Done

PR #4920, opened 2023-02-17
Fix STFT output shape

jstoecker:
### Description
Fixes STFT shape inference logic to compute the number of DFTs using the frame/window size instead of the number of per-frame DFT output elements. The updated logic matches the description of the operator.

### Motivation and Context
See https://github.com/microsoft/onnxruntime/issues/14316


jstoecker:
@pranavsharma : @martinb35  suggested I bring this to your attention. Thanks.

onnx/defs/math/defs.cc
```diff
@@ -3464,17 +3464,15 @@ ONNX_OPERATOR_SET_SCHEMA(
           }
 
           bool is_onesided = static_cast<bool>(getAttribute(ctx, "onesided", 0));
-          if (is_onesided) {
-            dft_size = is_onesided ? ((dft_size >> 1) + 1) : dft_size;
-          }
+          int64_t dft_unique_bins = is_onesided ? ((dft_size >> 1) + 1) : dft_size;
```
gramalingam, line 3467: The new code looks better since it eliminates some redundancy. But the PR description sounds like the behavior is changed, and it seems to me there would be no change in the behavior of the code here? Am I missing something? Thanks!

jstoecker: The behavior does change for the calculation of `n_dfts`, which is supposed to be using the frame/window size (stored in dft_size variable) and not the length of each output DFT.

An alternative (buy functionally equivalent) change would have been the following:

```diff
+ auto n_dfts = static_cast<int64_t>((signal_size - dft_size) / static_cast<float>(frame_step_value)) + 1;
bool is_onesided = static_cast<bool>(getAttribute(ctx, "onesided", 0));
if (is_onesided) {
  dft_size = is_onesided ? ((dft_size >> 1) + 1) : dft_size;
}

- auto n_dfts = static_cast<int64_t>((signal_size - dft_size) / static_cast<float>(frame_step_value)) + 1;
```

gramalingam: Oh, ok, thanks for the explanation. I didn't see the use of dft_size in the computation of n_dfts.

gramalingam:
@smk2007 can you please take a look, thanks!

gramalingam (approved):

postrational (approved):

xadupre:
Is is possible to fix the DCO?

PR #4924, opened 2023-02-19
Disallow and fix relative imports

justinchuby:
Signed-off-by: Justin Chu <justinchu@microsoft.com>

### Description

- Enable ruff rule to disallow relative imports
- Upgrade ruff to 0.0.248 to enable auto-fixing imports

### Motivation and Context

https://google.github.io/styleguide/pyguide.html#22-imports


onnx/test/test_external_data.py
```diff
@@ -206,8 +206,8 @@ def create_data_tensors(
     def create_test_model_proto(self) -> ModelProto:
         tensors = self.create_data_tensors(
             [
-                (self.attribute_value, "attribute_value"),
-                (self.initializer_value, "input_value"),
+                (self.attribute_value, "attribute_value"),  # type: ignore[list-item]
+                (self.initializer_value, "input_value"),  # type: ignore[list-item]
```
jcwchen, line 210: Just curious: is it some new check brought by the newer ruff or black? (was wondering why the ignore isn't needed previously)

jcwchen (approved):
Thanks for the effort!

onnx/test/test_external_data.py
justinchuby: mypy may be picking them up now that it knows the path I suppose

justinchuby:
@jcwchen should we push this forward to allow further conflicts?

jcwchen:
> @jcwchen should we push this forward to avoid further conflicts?

Yes, I was thinking the same thing, but it still needs an approval from @onnx/sig-operators-approvers. Could anyone please review/approve this PR? Thanks!

gramalingam (approved):
Thanks!

PR #4928, opened 2023-02-21
Set default CheckerContext for check_function properly

jcwchen:
### Description

Set default CheckerContext for check_function properly: inherit opset_imports from given FunctionProto

### Motivation and Context

Closes https://github.com/microsoft/onnx-script/issues/455.


onnx/checker.py
```diff
@@ -84,11 +84,15 @@ def check_node(node: NodeProto, ctx: C.CheckerContext = DEFAULT_CONTEXT) -> None
     pass
 
 
-@_create_checker(FunctionProto)
 def check_function(
-    graph: FunctionProto, ctx: C.CheckerContext = DEFAULT_CONTEXT
+    function: FunctionProto, ctx: C.CheckerContext = DEFAULT_CONTEXT
```
gramalingam, line 88: I think it is better to use `ctx: Optional[C.CheckerContext] = None`.

onnx/checker.py
```diff
@@ -84,11 +84,15 @@ def check_node(node: NodeProto, ctx: C.CheckerContext = DEFAULT_CONTEXT) -> None
...
 ) -> None:
-    pass
+    if ctx == DEFAULT_CONTEXT:
```
gramalingam, line 90: And change this to `if cts is None`, and create a new context

jcwchen: Updated. PTAL thanks!

onnx/checker.py
```diff
@@ -84,11 +84,17 @@ def check_node(node: NodeProto, ctx: C.CheckerContext = DEFAULT_CONTEXT) -> None
...
 
-@_create_checker(FunctionProto)
 def check_function(
-    graph: FunctionProto, ctx: C.CheckerContext = DEFAULT_CONTEXT
+    function: FunctionProto, ctx: Optional[C.CheckerContext] = None
 ) -> None:
-    pass
+    if ctx is None:
+        ctx = C.CheckerContext()
+        ctx.ir_version = IR_VERSION
```
jcwchen, line 92: Not sure whether we can just use the latest ir version here

gramalingam: Another alternative is to add another optional parameter `ir_version`, and require user to specify exactly one of `ctx` or `ir_version`. What do you think?

gramalingam: My thought is to avoid situations where the test-cases/behaviors change when we have a new release, where IR_VERSION gets bumped. For example, we have [this PR](https://github.com/onnx/onnx/pull/4911) where the test-case data has changed. I don't understand why (some test have changed, but not others). I suspect it is because of the bump in IR_VERSION. I think it is better if we make these choices consciously, instead of depending on a default-value.

jcwchen: Good point... I was thinking using [helper.find_min_ir_version_for](https://github.com/onnx/onnx/blob/7089e41df17292729854eaabfe5f53b0e135309a/onnx/helper.py#L96) to always get the fixed ir_version according to the given opset_imports. For those who are interested in using customized opset_imports and ir_version for check_function, they can create their own `onnx.onnx_cpp2py_export.checker.CheckerContext` with specified opset_imports and ir_version (We might need to shorten this module for easier usage). Does it make sense to you?

gramalingam: Sure, that sounds good. It will produce fixed text-cases.

gramalingam (approved):

onnx/checker.py
jcwchen: [helper.find_min_ir_version_for](https://github.com/onnx/onnx/blob/7089e41df17292729854eaabfe5f53b0e135309a/onnx/helper.py#L96) would raise issue when encountering unknown domain. Thus, I introduced ignore unknown to let check_function ignore unknown domain. @gramalingam PTAL. Thanks!

gramalingam (approved):
LGTM, thanks!

PR #4940, opened 2023-02-25
fix tensor.h bug and enlarge ParseData dtype

HSQ79815:
### Description

1. In onnx/common/tensor.h, there have some bugs
    - add `elem_num()` function to count the number of elements.  
       - older function is `size_from_dim()` , but it will assert fail when tensor is a scalar.
   - remove build-in unary/binaray op
       ```
       // For int32, uint8, int8, uint16, int16, bool, and float16 values
       // to writing to the buffer.
      // When this field is present, the data_type field MUST be
      // INT32, INT16, INT8, UINT16, UINT8, BOOL, FLOAT16 or BFLOAT16
      repeated int32 int32_data = 5 [packed = true];

     // Serializations can either use one of the fields above, or use this
     // raw bytes field. The only exception is the string case, where one is
     // required to store the content in the repeated bytes string_data field.

     // When this raw_data field is used to store tensor value, elements MUST
     // be stored in as fixed-width, little-endian order.
     // Floating-point data types MUST be stored in IEEE 754 format.
     // Complex64 elements must be written as two consecutive FLOAT values, real component first.
     // Complex128 elements must be written as two consecutive DOUBLE values, real component first.
     // Boolean type MUST be written one byte per tensor element (00000001 for true, 00000000 for false).
     //
      // Note: the advantage of specific field rather than the raw_data field is
      // that in some cases (e.g. int data), protobuf does a better packing via
      // variable length storage, and may lead to smaller binary footprint.
      // When this field is present, the data_type field MUST NOT be STRING or UNDEFINED
     bytes raw_data = 9;

      ```
     According to TensorProto definition, when `raw_data` is used,   elements must be stored in as `fixed-width, little-endian order`. in tensor.h , for example, `binary op` , when `elem_type_`is `bool int8 uint8 int16 uint16 int32`,  use`data<int32_t>()` to get a `int32` value pointer,  that will be raise memory segmentation  when value is stored in `raw_data` [tensor binary op code](https://github.com/onnx/onnx/blob/main/onnx/common/tensor.h#L343)
    
      Actually, the implement of unary/binary op  is incomplete, it will  just impact  the fuse_bn_into_conv pass of onnxoptimizer after removing it. I have  refactored  the fuse_bn_into_conv pass https://github.com/onnx/optimizer/pull/126

2. ~~add some  specialization  implement of `ParseData()`  in onnx/def/tensor_util.cc~~

### Motivation and Context
I am a maintainer of [onnx optimizer](https://github.com/onnx/optimizer). I am going to update onnx optimizer, this pr  make it more convenient and solid.


onnx/defs/tensor_util.cc
```diff
@@ -54,5 +57,29 @@ DEFINE_PARSE_DATA(int64_t, int64s)
...
+DEFINE_PARSE_DATA(uint32_t, uint64s)
+
+#undef DEFINE_PARSE_DATA
+
+template <>
+const std::vector<bool> ParseData<bool>(const Tensor* tensor) {
+  std::vector<bool> res;
+  if (!tensor->is_raw_data()) {
+    std::transform(tensor->int32s().cbegin(), tensor->int32s().cend(), std::back_inserter(res), [](int32_t d) -> bool {
+      return !!d;
```
daquexian, line 73:
```suggestion
      return static_cast<bool>(d);
```

onnx/common/tensor.h
```diff
@@ -222,49 +217,25 @@ struct Tensor final {
...
-  // FLOAT, DOUBLE,
-  // TODO: Support for FLOAT16
-  void sqrt();
-
-  // Element wise scaling of tensor s
-  // s is one dimensional, has size M, where M is size of first dimension of tensor
-  // s must have has data type corresponding to this
-  // Supported for
-  // FLOAT16, FLOAT, DOUBLE
-  void scale_by_first_dim(const Tensor& s);
```
daquexian, line 265: I personally highly agree to remove these functions. They were written because of some specific passes in onnx optimizer, so only a confusing subset of operations (like `scale_by_first_dim`) was implemented. Now onnx optimizer decides to not use these operation implementations any more (for more context please see https://github.com/onnx/optimizer/issues/9). However, I don't know whether there are libraries other than onnx-optimizer using these functions.

Could you please share your opinions? @askhade @gramalingam @jcwchen

onnx/defs/tensor_util.cc
HSQ79815: You are right, I will change it

onnx/common/tensor.h
jcwchen: Thanks for bringing this up. I am fine with removing them and I am not aware any other libraries using these functions.

onnx/defs/tensor_util.cc
```diff
@@ -54,5 +57,29 @@ DEFINE_PARSE_DATA(int64_t, int64s)
 DEFINE_PARSE_DATA(float, floats)
 DEFINE_PARSE_DATA(double, doubles)
 DEFINE_PARSE_DATA(uint64_t, uint64s)
+DEFINE_PARSE_DATA(uint8_t, int32s)
+DEFINE_PARSE_DATA(int8_t, int32s)
+DEFINE_PARSE_DATA(uint16_t, int32s)
+DEFINE_PARSE_DATA(int16_t, int32s)
+DEFINE_PARSE_DATA(uint32_t, uint64s)
```
jcwchen, line 64: Now these ParseData for these new data types are not used in onnx repo. Will they all be covered in onnx optimizer?

onnx/common/tensor.h
```diff
@@ -108,7 +99,11 @@ struct Tensor final {
   std::vector<int64_t>& sizes() {
     return sizes_;
   }
-
+  /// if tensor is a scaler, the sizes is empty, but the element number is actually 1.
```
gramalingam, line 102: typo: "scaler" => "scalar"

HSQ79815: I will correct this typo

onnx/defs/tensor_util.cc
HSQ79815: Good idea, what is your opinion ? @daquexian

daquexian: Maybe we can add these new template specifications in onnx-optimizer repo so that no unused functions are introduced in onnx repo.

HSQ79815: yeah, it is wise to implement function for these new data types in onnx-optimizer. I will undo this change in onnx repo. cc @daquexian @jcwchen

onnx/common/tensor.h
gramalingam: Sounds fine to me too, thanks!

onnx/common/tensor.h
```diff
@@ -108,7 +99,11 @@ struct Tensor final {
   std::vector<int64_t>& sizes() {
     return sizes_;
   }
-
+  /// if tensor is a scalar, the sizes is empty, but the element number is actually 1.
+  /// size_from_dim() can not handle that tensor is a scalar, while elem_num() can correctly
```
daquexian, line 103:
```suggestion
  /// size_from_dim() cannot handle this case, while elem_num() handles it correctly
```

daquexian (approved):

PR #4943, opened 2023-02-26
replace MIT with Apache-2.0

andife:
### Description
* correct license information in built proto files (original onnx.in.proto already had the correct license information)
* Add spdx license to cc / h files

### Motivation and Context
These files where also highlighted at the fossology scan from the linux foundation. (mail could be found at https://lists.lfaidata.foundation/g/onnx-technical-discuss/topic/96811486)


jcwchen (approved):
Thanks for catching these!

PR #4952, opened 2023-03-01
C implementation for slow python implementation (Conv)

xadupre:
### Description
Python implementation `Conv` is very slow in python. It is difficult to compute the output of a model of several Mb. This speeds up the inference by add a C implementation of operator `Conv` (python is still available).

### Motivation and Context
Performance.

### DIFF
--- a/onnx/cpp2py_export.cc
+++ b/onnx/cpp2py_export.cc
@@ -15,6 +15,7 @@
 #include "onnx/defs/printer.h"
 #include "onnx/defs/schema.h"
 #include "onnx/py_utils.h"
+#include "onnx/reference/c_ops/c_op_conv.h"
 #include "onnx/shape_inference/implementation.h"
 #include "onnx/version_converter/convert.h"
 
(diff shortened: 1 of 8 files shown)


gramalingam:
Where is the C implementation? I don't see it in the modified files. Also, I do think that the main point of the reference implementation is to clarify the spec, not perf. It is more important to have a declarative, easy-to-understand, less-buggy implementation than to have better performance.

xadupre:
The C code is here: https://github.com/onnx/onnx/pull/4952/files#diff-c9427752cb3471a694d21c36299a208a5f5977135ed2e3d98e1bcafe41c26ee3R1. I agree with you about the reference implementation. This is a proposal to speed up the reference implementation of Conv. The current one is really too slow to be used to run the deep learning model (densenet121, ...). This is one option to speed it up. I did it because I knew I could do it in a couple of hours. The other option to speed up Conv is to rewrite it and use im2col. But I think I still need to have a fast implementation of im2col.

jbachurski:
I agree with @gramalingam that avoiding a C implementation would be best. The reference implementation should be convenient to debug and its correctness should be easy to ensure, rather than causing another liability that has to be built and may possibly cause fatal errors (segfaults).

onnx/cpp2py_export.cc
```diff
@@ -498,6 +499,32 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
   printer.def("model_to_text", ProtoBytesToText<ModelProto>);
   printer.def("function_to_text", ProtoBytesToText<FunctionProto>);
   printer.def("graph_to_text", ProtoBytesToText<GraphProto>);
+
+  // Submodule `schema`
```
justinchuby, line 503: Possible to break out ref implementation to a separate file?

onnx/reference/c_ops/__init__.py
```diff
@@ -0,0 +1 @@
+# SPDX-License-Identifier: Apache-2.0
```
justinchuby, line 1: Should we add text to caution that we should avoid reference implementation in C unless it’s absolutely need like in this case?

onnx/reference/c_ops/c_op_common.h
```diff
@@ -0,0 +1,165 @@
+#if !defined(_CRT_SECURE_NO_WARNINGS)
+#define _CRT_SECURE_NO_WARNINGS
+#endif
+
+#include <pybind11/numpy.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+namespace ONNX_NAMESPACE {
```
justinchuby, line 9: Should we hide it from the ONNX_NAMESPACE?

onnx/reference/c_ops/c_op_common.h
```diff
@@ -0,0 +1,165 @@
...
+template <class NTYPE>
+NTYPE flattened_dimension(const std::vector<NTYPE>& values) {
+  NTYPE r = 1;
+  for (auto it = values.begin(); it != values.end(); ++it)
+    r *= *it;
+  return r;
+}
+
+template <class NTYPE>
+NTYPE flattened_dimension(const std::vector<NTYPE>& values, int64_t first) {
```
justinchuby, line 50: Nit: https://google.github.io/styleguide/cppguide.html#Function_Names seems to be consistent with the rest in this repo

justinchuby:
Yeah I personally think the comments above are reasonable

xadupre:
I'll probably close this PR unless somebody tells me otherwise.

daquexian:
Even though I agree that c implementation will bring an extra maintenance burden, the current implementation is really slow. Is it possible to optimize the implementation with pure python/numpy? Conv is a computation-intensive op and should have a decent speed if we leverage the vectorization operations in numpy.

xadupre:
I think it should be doable by using im2col + gemm. I started to do something in that way but did not finish. I was wondering if we should add im2col in onnx specifications (there is only col2im right now).

xadupre:
I'll submit another then.

gramalingam:
> I think it should be doable by using im2col + gemm. I started to do something in that way but did not finish. I was wondering if we should add im2col in onnx specifications (there is only col2im right now).

Good point. I think there was a proposal to define Conv as a function in terms of other functions (I thought that was one of the motivations for col2im).

xadupre:
Col2im is for convtranspose. Im2col is for conv.

PR #4242, opened 2022-06-01
[WIP] Upgrade Protobuf to v21.1 to accept Python Protobuf 4.21

jcwchen:
**Description**
Upgrade Protobuf to v21.1 to accept Python Protobuf 4.21

After upgrading to Protobuf 21. for building ONNX, ONNX with older Python Protobuf (<=3.20) will behave differently from ONNX with Python Protobuf (4.21): the output order for ONNX ModelProto will be different. To keep consistent behavior, the minimum supported Protobuf versiob would be 4.21 after this upgrade.

**Motivation and Context**
Closes https://github.com/onnx/onnx/issues/4239.

PR #4956, opened 2023-03-02
Upgrade Protobuf to v21.12 to accept Python Protobuf 4.21

jcwchen:
### Description

Upgrade Protobuf to v21.12 to accept Python Protobuf 4.21

### Motivation and Context

Supersede https://github.com/onnx/onnx/pull/4242


requirements.txt
```diff
@@ -1,3 +1,3 @@
 numpy >= 1.16.6 # TODO: update once TensorFlow 2.6 is end of life
-protobuf >= 3.20.2, < 4
+protobuf
```
xadupre, line 2: Should we let the lower constraint `>=3.20.2`?

gramalingam (approved):

setup.py
```diff
@@ -26,7 +26,7 @@
 SRC_DIR = os.path.join(TOP_DIR, "onnx")
 TP_DIR = os.path.join(TOP_DIR, "third_party")
 CMAKE_BUILD_DIR = os.path.join(TOP_DIR, ".setuptools-cmake-build")
-PACKAGE_NAME = "onnx"
+PACKAGE_NAME = "onnx-protobufv21"
```
justinchuby, line 29: What does this do?

jcwchen: Good catch. This was for testing the built wheel with PyTorch... Here is the PR to fix this: https://github.com/onnx/onnx/pull/5089.

PR #4960, opened 2023-03-03
Add expected outputs for light models to make the backend test faster

xadupre:
### Description
Current backend test use ReferenceEvaluator to compute the expected output for random input. However, implementation for operator Conv is far too slow to make them usable. This PR creates explicit inputs for those models and add the expected outputs to the reposity to avoid relying on ReferenceEvaluator to compute them.

### Motivation and Context
Agility.


jcwchen (commented):
Thanks for also adding tests for reference/runtime. It seems useful to have a CI running test_backend_onnxruntime.py with ONNX Runtime.

xadupre:
The test are running if onnxruntime is installed on CI. It is not the case right now. This file makes it easier for developer to check onnxruntime on existing tests.

onnx/backend/test/data/light/light_bvlc_alexnet_output_0.pb
```diff
@@ -0,0 +1 @@
...o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:o�:
```
yuanyao-nv, line 1: Should we organize these files in such a way that for the scoreboard pipelines running these light models no special logic would be needed?
For the node tests, the structure is:
test_op
|--model.onnx
|--test_data_set_0
    |--input_0.pb
    |--input_1.pb
I'm actually not familiar with where the input/output for the original real models were placed. But it would be convenient to the backend/scoreboard maintainers to mimic the original layout if possible, so that no/minimal changes to the backend scoreboard pipelines would be necessary.

xadupre: For those models, the input is not stored but automatically created. That's why I did not follow the same architecture. Only the expected is stored to avoid running the reference implementation, too slow to be used in CI. I can change it if you think it is better. I can also add a file like `input_0.pb.md` to tell how the input must be created.

yuanyao-nv: I think instead of generating the input as you are doing in this PR, ie `np.arange(n)`, we can probably upload the randomized input pb files alongside the output pb files? There are two benefits for doing this:

First, we can then unify the the data storage for light model and node model directories so that there is no need for having separate data loading logic for these tests, ie:
```
test_light_model_x/
|--model.onnx
|--test_data_set_0/
    |--input_0.pb
    |--input_1.pb
    |--output_0.pb
```
(Also, I might not be aware of all the details but can we also just put these light models and I/O files in the real model directories instead of having the real model directories point back to the light models?)

Second, generating the data like `np.arange(n)` can result in some very large values when the input tensor size is large. This might or might not give numerical issues later on for output comparison purposes. If we upload pre-generated input tensors then we can make sure the input values stay in some desirable range.

xadupre: About `arange`, the formula is `(np.arange(n).reshape(shape) / n).astype(np.float32)`. So the input is in `[0, 1]`. It should be ok. The input size is 600kb for each model (an image= 3x224x224*sizeof(float)). Three 8 models, so it would add 5Mb. I'd rather avoid it if possible.

gramalingam (approved):

PR #4963, opened 2023-03-06
Clarify reduction op inputs can be rank-0 tensors (scalars)

gramalingam:
### Description
Clarify that the input of a reduction op can be a rank-0 tensor.

### Motivation and Context
There seems to be some confusion whether this is valid or not. The interpretation and return-value are straightforward with no ambiguity. So, clarify this explicitly.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -17212,9 +17212,10 @@ This version of the operator has been available since version 13 of the default 
 
 ### <a name="ReduceL1-13"></a>**ReduceL1-13**</a>
 
-  Computes the L1 norm of the input tensor's element along the provided axes. The resulting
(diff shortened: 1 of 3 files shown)


jcwchen (approved):

justinchuby:
@xiaowuhu @fatcat-z this means some functions we created don’t need to explicitly account for scalars

daquexian (approved):

justinchuby:
Do we also have backend tests for this?

xadupre:
Probably not otherwise the reference implementation would probably have failed those tests.

justinchuby:
Should we clarify max and min as well? I found they also support scalars.

PR #4966, opened 2023-03-07
Add ONNX Runtime tests in CI pipelines with installed ONNX Runtime

jcwchen:
### Description

Add ONNX Runtime tests in CI pipelines with installed ONNX Runtime.

### Motivation and Context

Currently these are some new tests which need installed onnxruntime. Add a section in release CI pipelines to run these tests with installed onnxruntime.


onnx/test/reference_evaluator_ml_test.py
```diff
@@ -12,6 +12,7 @@
 from onnx.checker import check_model
 from onnx.defs import onnx_opset_version
 from onnx.helper import (
+    ORT_MAX_IR_SUPPORTED_VERSION,
```
xadupre, line 15: I think we should avoid adding constants related to onnxruntime in the code. What about the other runtime? This information should be placed in onnxruntime.

onnx/test/reference_evaluator_backend_test.py
```diff
@@ -615,6 +616,13 @@ def common_test_onnx_test_run(
                 print("CHECK RUNTIME onnxruntime")
                 from onnxruntime import InferenceSession
 
+                # The new IR version is not supported by onnxruntime yet
```
xadupre, line 619: There should be issue with the opset as well. I used onnxruntime to check the reference implementation was giving the same answer. It was not meant to be run on CI but only locally. I'm wondering if we should move those tests inside another subfolder, not included in the onnx package.

onnx/test/reference_evaluator_ml_test.py
jcwchen: Same thought. I am not sure whether current ONNX Runtime's API provides that and so I created an issue to track this: https://github.com/microsoft/onnxruntime/issues/14932.

onnx/test/reference_evaluator_backend_test.py
jcwchen: > There should be issue with the opset as well. 


Thanks for the reminder. I added check for opset as well.

> It was not meant to be run on CI but only locally. I'm wondering if we should move those tests inside another subfolder, not included in the onnx package.

IIUC, python files under onnx/test are not packaged in the onnx Python wheel. If that is the case, it seems fine to have them here? Still, putting them in another directory might be more organized (remind users these relevant tests need onnxruntime).

onnx/helper.py
```diff
@@ -73,6 +73,14 @@
 
 VersionMapType = Dict[Tuple[str, int], int]
 
+# TODO (https://github.com/microsoft/onnxruntime/issues/14932): Get max supported version from onnxruntime directly
+ORT_MAX_ONNX_OPSET_SUPPORTED_VERSION = 18
```
xadupre, line 77: I would probably make it as environment variable in the CI job rather than mentioning ORT in the repository.

onnx/test/test_with_ort.py
```diff
@@ -2,36 +2,44 @@
...
+        # test ONNX functions
+        model = load(example1)
+        checker.check_model(model)
+        checker.check_model(model, True)
+        inferred_model = shape_inference.infer_shapes(model, True)
+        temp_filename = "temp.onnx"
+        save(inferred_model, temp_filename)
+
+        # test ONNX Runtime functions
+        sess = InferenceSession(temp_filename)
```
xadupre, line 35: I would write `InferenceSession(inferred_model.SerializeToString())` to avoid writing a file on disk.

onnx/helper.py
jcwchen: Although we will need to manually bump the supported versions as environment variables in 4 CIs, I cannot think of other better thought. I have added them as environment variables.

onnx/test/test_with_ort.py
jcwchen: Good point. Updated. PTAL. Thanks!

xadupre (approved):

Issue #4964, opened 2023-03-06
[CI] The pipeline to generate and publish ONNX docs failed after IR bumped


jcwchen:
# Bug Report

### Describe the bug

The pipeline to generate and publish ONNX docs [failed](https://github.com/onnx/onnx/actions/runs/4320587058/jobs/7540968637) after [IR bumped](https://github.com/onnx/onnx/pull/4911).

### System information
- OS Platform and Distribution (*e.g. Linux Ubuntu 20.04*): ubuntu-latest
- ONNX version (*e.g. 1.13*): latest main branch
- Python version: 3.10

### Expected behavior

The CI should be green and produce correct web content.

### Notes

cc @liqunfu @xadupre

PR #4969, opened 2023-03-07
Fix web content CI: freeze ir_version as a relesed one

jcwchen:
### Description

Fix web content CI: freeze ir_version as a released one. Then the created models can be run by ONNX Runtime. Not sure whether python.rst is the only failed file after IR bumped.

### Motivation and Context

Fixes https://github.com/onnx/onnx/issues/4964.


gramalingam (approved):

docs/docsgen/source/intro/python.rst
```diff
@@ -1504,7 +1505,7 @@ This case is still puzzling.
 
         graph = make_graph([node1, node2], 'example', [X, A], [Y])
 
-        onnx_model = make_model(graph, opset_imports=opset_imports)
+        onnx_model = make_model_gen_version(graph, opset_imports=opset_imports)
```
xadupre, line 1508: There are many `make_model` in the tutorial. Should we fix all of them? The function is not imported so the whole example will fail. I think we should be more careful with this change if it is only introduced for onnxruntime.

jcwchen: I think as long as the created ONNX models were not run by `onnxruntime.InferenceSession` (onnxruntime cannot handle unreleased opset and ir_version), we can still use make_model. I have went through all the .rst files for doc generation: only `docs/docsgen/source/expect_onnxruntime.rst` (the expect function seems not called though) and `docs/docsgen/source/intro/python.rst` will run onnxruntime.InferenceSession with created ONNX models. With this PR, this issue should be resolved. Please let me know if I miss other places. Thanks!

justinchuby: A comment on why would be great I think

xadupre: It may happen in other places in the future. I think using a function different from `make_model` is confusing for the users, specially in a tutorial. I usually freeze the ir_version and the opset when it is the case. Plus the function `make_model_gen_version` is not imported in the example. The documentation should still break. The function `make_model_gen_version` does not appear in the documentation so users need to look into the code to understand what it does.

jcwchen: Good point... I was thinking make_model is more proper for a tutorial as well. As you suggested, I guess for this one we can instead using make_model with setting a fixed ir_version (+ comments why we set ir_version manually). Will update it.

gramalingam: But it may be worth considering what users should be encouraged to use: `make_model` or `make_model_gen_version` ?

gramalingam: It can help if the tutorial explains the differences between the two and why/when one should be preferred.

xadupre: I think we should keep it simple, the tutorial should use only one. If it is make_model_gen_version, then it must be added to the documentation and be documented (it is not right now). We can also change the default behaviour of make_model. Is it really good to to ir_version=9 with opset 17 ?

PR #4970, opened 2023-03-07
Update documentation of optional arguments

gramalingam:
Update IR.md to clarify the distinction between the two kinds of optional arguments in ONNX.

### DIFF
--- a/docs/IR.md
+++ b/docs/IR.md
@@ -343,6 +343,8 @@ For each variadic operator input, N or more node inputs must be specified where 
 
 #### Optional Inputs and Outputs
 
+##### Pre IR-8 
+
 Some operators have inputs that are marked as optional, which means that a referring node MAY forgo providing values for such inputs.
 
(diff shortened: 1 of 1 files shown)


docs/IR.md
```diff
@@ -351,6 +353,18 @@ There are two ways to leave an optional input or output unspecified: the first,
 
 Each node referring to an operator with optional outputs MUST provide a name for each output that is computed and MUST NOT provide names for outputs that are not computed.
 
+Optional inputs and outputs of the above kind are referred to as _static-optional_.
+
+##### IR-8 Version and Later
```
jcwchen, line 358: I like the sub sections here, which are very clear for different IR versions.

nit: I feel `#####` is a little small? (Even smaller then the content below?) Will simply using bold text better?
```suggestion
**IR-8 Version and Later**
```

gramalingam: Done, thanks!

jcwchen (approved):

Issue #4489, opened 2022-09-05
convert_version for MaskRCNN and FasterRCNN causes segmentation fault


kshpv:
# Bug Report

### Is the issue related to model conversion?
No

### Describe the bug
When I try to `convert_version()` function for Mask RCNN or Faster RCNN models from [ONNX Model Zoo ](https://github.com/onnx/models/tree/main/vision/object_detection_segmentation), the segmentation fault occured.

### System information

- OS Platform and Distribution (*e.g. Linux Ubuntu 16.04*): Windows 10
- ONNX version (*e.g. 1.7*):  1.11
- Python version: 3.9

### Reproduction instructions
- Describe the code to reproduce the behavior.
```
import onnx
from onnx.version_converter import convert_version
model = onnx.load('Faster-RCNN.onnx')
converted_model = convert_version(model, 13)
```
Models:
https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/faster-rcnn/model/FasterRCNN-12.onnx
https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/mask-rcnn/model/MaskRCNN-12.onnx

### Expected behavior
To obtain opset 13 version models

jcwchen:
Hi @kshpv,
I can repro this seg fault issue with ONNX 1.11.0, but it seems resolved in the latest ONNX 1.12.0. I guess this PR helps: https://github.com/onnx/onnx/pull/4079. Note that although version_converter from ONNX 1.12.0 won't cause seg fault with MaskRCNN-12.onnx, it still cannot convert MaskRCNN-12.onnx because of this: https://github.com/onnx/onnx/issues/4101.

Issue #4101, opened 2022-03-30
version_converter bug


kshpv:
# Bug Report

### Describe the bug
For all ssd models from https://github.com/onnx/models/tree/main/vision/object_detection_segmentation/ssd/model when I try to convert them from their opset to 13 opset I got the following error:

`Assertion target_shape.size() != 0 failed: Version conversion for Softmax failed because input shape is unknown.`

### System information
I am using ONNX 1.11

### Reproduction instructions
- Describe the code to reproduce the behavior.
```
def add_input_from_initializer(model: onnx.ModelProto) -> None:
          """
          Currently onnx.shape_inference doesn't use the shape of initializers, so add
          that info explicitly as ValueInfoProtos.
          Mutates the model.
          Args:
              model: The ModelProto to update.
          """
          # All (top-level) constants will have ValueInfos before IRv4 as they are all inputs
          if model.ir_version < 4:
              nncf_logger.info('Could not process model, as it has {} < 4'.format(model.ir_version))
              return model

          def add_const_value_infos_to_graph(graph: onnx.GraphProto):
              inputs = {i.name for i in graph.input}
              existing_info = {vi.name: vi for vi in graph.input}
              for init in graph.initializer:
                  # Check it really is a constant, not an input
                  if init.name in inputs:
                      continue

                  # The details we want to add
                  elem_type = init.data_type
                  shape = init.dims

                  # Get existing or create new value info for this constant
                  vi = existing_info.get(init.name)
                  if vi is None:
                      vi = graph.input.add()
                      vi.name = init.name

                  # Even though it would be weird, we will not overwrite info even if it doesn't match
                  tt = vi.type.tensor_type
                  if tt.elem_type == onnx.TensorProto.UNDEFINED:
                      tt.elem_type = elem_type
                  if not tt.HasField("shape"):
                      # Ensure we set an empty list if the const is scalar (zero dims)
                      tt.shape.dim.extend([])
                      for dim in shape:
                          tt.shape.dim.add().dim_value = dim

              # Handle subgraphs
              for node in graph.node:
                  for attr in node.attribute:
                      # Ref attrs refer to other attrs, so we don't need to do anything
                      if attr.ref_attr_name != "":
                          continue

                      if attr.type == onnx.AttributeProto.GRAPH:
                          add_const_value_infos_to_graph(attr.g)
                      if attr.type == onnx.AttributeProto.GRAPHS:
                          for g in attr.graphs:
                              add_const_value_infos_to_graph(g)

[...]

Issue #4529, opened 2022-09-21
RuntimeError: /github/workspace/onnx/version_converter/adapters/softmax_12_13.h:53: adapt_softmax_12_13: Assertion `target_shape.size() != 0` failed: Version conversion for Softmax failed because input shape is unknown.


luchangli03:
when I convert onnx model by
onnx_model = version_convert(onnx_model, tgt_opset=15)
this error happens for some model
actually there are input and output shape for this node
I don't know how to solve this problem by modify the model.

luchangli03:
Is there a way to disable conversion of specific type of node?

jcwchen:
Hi @luchangli03,

It's a known limitation: https://github.com/onnx/onnx/issues/4101. However, if the input shape for that Softmax node does present and you still saw this error, there might be a bug. If possible, could you please share the model you have? I will take a deeper look at it.

> Is there a way to disable conversion of specific type of node?

Currently there is no way to do it.

luchangli03:
> Hi @luchangli03,
> 
> It's a known limitation: #4101. However, if the input shape for that Softmax node does present and you still see this error, there might be a bug. If possible, could you please share the model you have? I will take a deeper look at it.
> 
> > Is there a way to disable conversion of specific type of node?
> 
> Currently there is no way to do it. A workaround I can think of would be: if you are comfortable to build ONNX from source, you can modify the source code by adding "Softmax" in this line:
> 
> https://github.com/onnx/onnx/blob/895593af6ac17d4fe2749091659083d3b1246cae/onnx/version_converter/convert.cc#L81
> 
> to skip the version conversion for it.

I'm sorry I can't upload the model since it's too large.

jcwchen:
It should be fixed by https://github.com/onnx/onnx/pull/4974. Please let us if you still bump into issues when converting Softmax-12 to Softmax-13. Thanks!

PR #4974, opened 2023-03-09
Convert softmax op without knowing the shape

chenchongsong:
### Description
This PR enables version conversion for Softmax Op when the shape is unknown.

### Motivation and Context
The limitation is described in PR https://github.com/onnx/onnx/pull/3861. This PR solves the limitation.

Related issue: https://github.com/onnx/onnx/issues/4101


onnx/version_converter/adapters/softmax_12_13.h
```diff
@@ -44,38 +54,19 @@ class Softmax_12_13 final : public Adapter {
       node->output()->setUniqueName(original_output_name + "_intermediate");
```
gramalingam, line 54: Thanks for the PR. I wonder why we have the old lines 35-39? Specifically, `flatten` produces a 2D output, and so does it matter if the axis is 1 or -1 ? Don't they mean the same for a 2D tensor?

Just wondering, I realize it has nothing to do with the scope of the current PR.

gramalingam:
Is there a test-case to validate this change?

onnx/version_converter/adapters/softmax_12_13.h
chenchongsong: I think axis 1 and -1 mean the same. Let me simplify these lines in the next commit

xadupre (approved):

jcwchen (commented):
Thank you for proposing this PR! It can close https://github.com/onnx/onnx/issues/4489 https://github.com/onnx/onnx/issues/4529 as well.

Previously [some of ONNX Model Zoo models](https://github.com/onnx/onnx/blob/main/workflow_scripts/config.py#L13-L19) cannot use version_converter because of this limitation. Currently I am testing this PR with these failed models. Overall looks good. The only issue I have encountered is the converted model (which has Softmax-12 to Softmax-13) might have an empty shape and fail strict shape_inference, but such a converted model is still runnable by ONNX Runtime. For instance, [ssd-12.onnx](https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/ssd/model/ssd-12.onnx)

gramalingam:
> converted model (which has Softmax-12 to Softmax-13) might have an empty shape and fail strict shape_inference
What's the error message? Trying to understand what goes wrong.

jcwchen:
> What's the error message? Trying to understand what goes wrong.

Converted [ssd-12.onnx](https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/ssd/model/ssd-12.onnx) 's shape inference error: `
InferenceError: [ShapeInferenceError] Shape inference error(s): (op_type:Transpose, node name: Transpose_661): [TypeInferenceError] Invalid attribute perm {0, 2, 1}, input shape = {`
Not sure whether we can make Transpose's shape inference allow empty input shape here.


More version_converter experiments with this PR:
- [MaskRCNN-12.onnx](https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/mask-rcnn/model/MaskRCNN-12.onnx) can pass onnx.shape_inference, but fail with ONNX Runtime:`
InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : indices element out of data bounds, idx=1 must be within the inclusive range [0,-1]`
It seems somehow produce a 0 shape before a Gather node.

- [gpt2-lm-head-bs-12.onnx](https://github.com/onnx/models/blob/main/text/machine_comprehension/gpt2-bs/model/gpt2-lm-head-bs-12.onnx) can pass onnx.shape_inference, but fail with ONNX Runtime: `
Fail: [ONNXRuntimeError] : 1 : FAIL : Node (loop) Op (Loop) [TypeInferenceError] Graph attribute inferencing failed: Node (MatMul_221) Op (MatMul) [ShapeInferenceError] Input tensors of wrong rank (0).`

- Otherwise, the converted t5-encoder-12.onnx, t5-decoder-with-lm-head-12.onnx, bertsquad-12.onnx are good with onnx.shape_inference and ONNX Runtime

gramalingam:
@jcwchen : so looks like there is a version converter issue. But are these models that we could version-convert previously or not? Trying to understand if the error is in the conversion of this op, or if it is an error in the conversion of some other op that we hadn't caught before.

gramalingam:
The only issue I see with this (softmax) converter is that it requires at least the rank of the input tensor. (This can be improved if we omit the optimization at the beginning for a special case.) Not sure what will happen if the rank is not known. But real models probably have the rank known for all tensors typically, so doubt if this could be the cause.

onnx/version_converter/adapters/softmax_12_13.h
```diff
@@ -25,57 +25,45 @@ class Softmax_12_13 final : public Adapter {
...
-      }
-      Node* constant = graph->create(kConstant);
-      constant->insertBefore(node);
-      constant->t_(kvalue, t);
-      reshape->addInput(constant->output());
-
       // Fix outputs & wiring
-      node->output()->wipeSizes();
-      reshape->output()->setSizes(target_shape);
+      reshape->output()->setSizes(node->output()->sizes());
```
jcwchen, line 59:
```suggestion
      if (node->output()->sizes().size() != 0) {
        reshape->output()->setSizes(node->output()->sizes());
      }
```
Potentially it will make the output rank of Reshape as zero if the original Softmax's output is unknown. Would it better not to assign it if the original rank is zero?

jcwchen (commented):
> so looks like there is a version converter issue. But are these models that we could version-convert previously or not? Trying to understand if the error is in the conversion of this op, or if it is an error in the conversion of some other op that we hadn't caught before.

Previously they cannot be version_converted due to missing input shape for Softmax-12 to Softmax-13. As you mentioned, it's possible that there are other version_conversion issues, but for these errors they happened right after the converted Softmax+Reshape.

> The only issue I see with this (softmax) converter is that it requires at least the rank of the input tensor. (This can be improved if we omit the optimization at the beginning for a special case.) Not sure what will happen if the rank is not known. But real models probably have the rank known for all tensors typically, so doubt if this could be the cause.

Good point. I guess that's the issue -- if onnx.shape_inference cannot infer the rank for original Softmax, the new converted Reshape will inherit that rank 0 and cause the shape inference error from the following node due to rank 0.

onnx/version_converter/adapters/softmax_12_13.h
jcwchen: I tested this condition with the failed ONNX Model models (ssd-12.onnx, MaskRCNN-12.onnx, gpt2-lm-head-bs-12.onnx) and they can pass both onnx.checker with full_check and onnxruntime.InferenceSession.

chenchongsong: Yes, this looks better. Thanks for the suggestion!

chenchongsong:
@gramalingam @jcwchen @xadupre All conversations are currently resolved. Could anyone help to merge this PR? Thanks

jcwchen (approved):

PR #4975, opened 2023-03-09
[WIP] Numpy API for ONNX (don't review yet)

xadupre:
### Description, Motivation, Context

Many users have difficulties to write onnx graphs. Many packages tries to symplify it either by implementing their own api very close to onnx operators (sklearn-onnx, tf2onnx, spox, onnx-script). This contribution tries a different approach by implementing a numpy API for ONNX. It does not cover everything numpy or ONNX can do but it can easily be used to define loss functions for example without knowing too much about ONNX.

Every user training machine learned models knows about numpy and its functions. This PR proposes a way to automatically convert a code using numpy into ONNX. This is the first step, it does not support control flow yet (test and loops). There is no easy syntax for that yet and the main challenge is to deal with local context.

Since the onnx package now supports a reference implementation, it is now possible to implement and test such development. The code written with this API can be executed with the reference implementation as a standalone graph. Eager mode is also available. The runtime can easily be replaced with onnxruntime if the user wants a faster execution or even try on other providers such as CUDA.


cbourjau:
Could you elaborate on why such a library should be part of `onnx/onnx` itself? It would seem to me perfectly possible and probably preferable to implement this proposal as an independent project.

pyproject.toml
```diff
@@ -109,11 +123,18 @@ ban-relative-imports = "all"
 "onnx/reference/ops/**" = ["N801"] # Class casing
 "onnx/reference/ops/_op_list.py" = ["F401"]
 "onnx/__init__.py" = ["F401"]
+"onnx/npx/__init__.py" = ["F401", "F403"]
+"onnx/npx/npx_functions.py" = ["F821"]
+"onnx/npx/npx_functions_test.py" = ["F821"]
+"onnx/npx/npx_graph_builder.py" = ["N811", "N801"]
```
justinchuby, line 129: Possible to add a comment on why they need to be ignored? I see codes that aren’t in the list. If possible I would just noqa in-line so they are not ignored for the whole file. I also hope to see we reduce this ignore list instead of growing it.

pyproject.toml
```diff
@@ -109,11 +123,18 @@ ban-relative-imports = "all"
...
+"onnx/npx/npx_types.py" = ["B904"]
+"onnx/npx/npx_var.py" = ["N801", "N802", "B904"]
 "onnx/reference/__init__.py" = ["F401"]
 "onnx/reference/ops/__init__.py" = ["F401"]
 "onnx/reference/ops/aionnxml/_op_list.py" = ["F401"]
 "onnx/reference/ops/aionnxml/__init__.py" = ["F401"]
 "onnx/reference/ops/aionnx_preview_training/__init__.py" = ["F401"]
 "onnx/reference/ops/aionnx_preview_training/_op_list.py" = ["F401"]
 "onnx/reference/ops/experimental/__init__.py" = ["F401"]
+"onnx/test/npx_test.py" = ["B904", "F821"]
```
justinchuby, line 139: Maybe fix instead?

pyproject.toml
```diff
@@ -46,6 +46,20 @@ module = [
...
+    'onnx.npx.npx_graph_builder',
+    'onnx.npx.npx_functions',
+    'onnx.npx.npx_functions_test',
+    'onnx.npx.npx_jit_eager',
+    'onnx.npx.npx_tensors',
+    'onnx.npx.npx_types',
+    'onnx.npx.npx_var',
+    'onnx.test.npx_test',
+]
+ignore_errors = true
```
justinchuby, line 61: I would use ignores inside files. It also seems dangerous to ignore all errors (which turns off mypy completely)?

justinchuby:
I like that users do not need to know onnx to create models. I wonder (without carefully considering) if this can be expressed and provided as a set of apis in onnx-script?

xadupre:
> Could you elaborate on why such a library should be part of `onnx/onnx` itself? It would seem to me perfectly possible and probably preferable to implement this proposal as an independent project.

**Where this comes from?**

This work comes from discussion I had with many developers using onnx. They were complaining about the lack of documentation and the difficulty to write onnx graphs. I wrote most of the documentation added last year. The reference implementation is a tool to help testing onnx graphs without the need of a new dependency. I implemented it to help me debugging the converters I was implementing. Making it part of onnx repository was a way the make sure that every new operator comes from a reference implementation. This is also a way to unit test the backend API provided in onnx package.

The numpy API addresses the last feedback I often received: onnx is difficult to use. There is almost no contribution to converting libraries except from the authors of them. Almost every library with onnx implemented its own API to create onnx graph and make it more simple. But these API are based on onnx operators list. It is still difficult to understand what a function does. Everybody knows numpy. With such a tool, everybody can create a short onnx graph. This can be used to implement simple functions (loss, gradient update, ...) or converters for machine learned models. People have pieces of code written with numpy they would like to migrate to onnx. There is other way than to rewrite them right now. This API is meant for them.

**No control flow?**

There is no control flow in sklearn-onnx. Most of the time, there is no control flow. It is not that this API will not support that but as @jbachurski mentioned, it is already big. And I did not find an intuitive syntax yet.

**Documentation**

It is very short. I agree. I initiated the first onnx documentation. I wrote the tutorial. I'll continue.

**Why here?**

It can be placed in an independant project. I put it here to get feedback. However, I think it could be nice if people see the numpy API the same time they discover onnx. It would gives them an easy to play with onnx without any new dependency. I added many examples in the [tutorial](https://onnx.ai/onnx/intro/python.html) but it is still difficult to test functions different from the ones provided there. With this API, users can easily use their own functions written with numpy, use onnxruntime, benchmark it...

pyproject.toml
xadupre: mypy does not work well with dynamic types (see [_make_type](https://github.com/onnx/onnx/pull/4975/files#diff-4172614a6538ec08dffe5b2799444c67fb64835cfbac8e3d82082f4472b838baR566). lintrunner gives different results on the machine I use comare to CI. I'll do better if the PR reaches a consensus.

jbachurski:
> The reference implementation is a tool to help testing onnx graphs without the need of a new dependency. I implemented it to help me debugging the converters I was implementing. Making it part of onnx repository was a way the make sure that every new operator comes from a reference implementation. This is also a way to unit test the backend API provided in onnx package.

I think that's a really good goal to strive for. The example I was making was only about the code volume and relating to [this comment](https://github.com/onnx/onnx/pull/4483#issuecomment-1283079050) on that PR - it would be better not to create very large changes before they are battle-tested and preferably can be reviewed by more people. For instance, features can be added one-by-one instead of all at the same time. I might not be aware of the exact development process behind this.

> People have pieces of code written with numpy they would like to migrate to onnx. There is other way than to rewrite them right now. This API is meant for them.

I see. That's an interesting goal, but it sounds like it's not a core ONNX functionality and should remain in a downstream package in that case. There's also many possible ways of approaching it that should be compared first before choosing  one implementation.

> It can be placed in an independant project. I put it here to get feedback.

The main piece of feedback I can give is that tracebacks are very important for efficient development. Currently, changing `float32` to `float64` raises a `RuntimeTypeError` at runtime after the 'JIT' is performed - the traceback to the operation that is failing is lost.
This is going beside the point that type promotion is also something that's deeply expected by numpy users. It is good that constants seem to work, though.

onnx/test/npx_test.py
```diff
@@ -1247,6 +1246,7 @@ def impl(A):
         self.assertEqual("A\nB\nC\n", text)
 
     @unittest.skipIf(InferenceSession is None, reason="onnxruntime is needed.")
+    @unittest.skipIf(True, reason="eager mode not ready yet")
```
justinchuby, line 1248:
```suggestion
    @unittest.skip(reason="eager mode not ready yet")
```


nit

cbourjau:
Thanks for your reply, @xadupre. After reading it I still strongly suggest that a (necessarily opinionated) library like this should not be part of `onnx/onnx` itself. Besides that, I'd have some remarks about the fundamental design choices. Given that it is already largely implemented I'm not sure if they are applicable at this point. What kind of feedback are you looking for?

xadupre:
Thanks for the feedback. I think this feature would help many more datascientists to use onnx to create onnx graphs. Beside the fact the amount of code is big or the internal design, what do you think of the feature itself? Should it be part of the onnx account as a standalone project?

This PR comes from a previous work doing the same thing. I used it to convert tensorflow RFFT op before DFT operator was brought to onnx (see [onnx_rfft_2d_any_test](https://github.com/onnx/tensorflow-onnx/blob/main/tf2onnx/onnx_opset/signal.py#L441). Notation such as `[:, 0]` are two painful to write with onnx.

xadupre:
By large graphs, I meant graphs with more than 2Gb which is the protobuf limit. I don't know if that works since I did not test it. My first attempt to build a numpy API was on the top on another API but I found difficult to maintain two APIs instead of directly using the make function provided by onnx. Control flow is not supported right now because I did not find an intuitive syntax about it. I saw the one proposed in Spox [If](https://spox.readthedocs.io/en/latest/guides/advanced.html#Conditional---If). This some directions I followed but I find it difficult to track which local variables a subgraph can use. My goal was more to implement an Array API (see https://data-apis.org/array-api/latest/API_specification/index.html). The users should not need to know the implementation details with possibly an application to this PR (https://github.com/scikit-learn/scikit-learn/pull/22554, but SVD is not part of onnx operators right now). [onnx-script](https://github.com/microsoft/onnx-script) is more simple in that regard but it is onnx operators schemas so it is difficult to use any code written with numpy as is which is my goal.

jbachurski:
> By large graphs, I meant graphs with more than 2Gb which is the protobuf limit.

I also misunderstood that statement before - but in my understanding this limitation is basically unavoidable anyway? It's inherent in the format.

> My first attempt to build a numpy API was on the top on another API but I found difficult to maintain two APIs instead of directly using the make function provided by onnx.

Using the proto functions directly requires e.g. resolving naming (hindering composition and abstraction) and powering through lack of any type hints or doc strings. This is what actually makes development extremely hard for new users (and developers maintaining the API). IMHO the ONNX repository already has a lot of low-level code that is hard to read and fix.

> My goal was more to implement an Array API (see https://data-apis.org/array-api/latest/API_specification/index.html).

I think that's absolutely the right goal, the point being that it should be built on top of the right abstractions to ensure it can be maintained easily (which is impossible when it's placed directly in the ONNX repo).

---

I am not sure how much control flow is relevant to the numpy debate. I can however give some clarifications:

> Control flow is not supported right now because I did not find an intuitive syntax about it. I saw the one proposed in Spox [If](https://spox.readthedocs.io/en/latest/guides/advanced.html#Conditional---If). This some directions I followed but I find it difficult to track which local variables a subgraph can use.

The build algorithm I designed for Spox allows usage of any and all variables that you can access in Python. If this makes it easier to follow if-branches, subgraphs can be functions instead of `lambda`s. 
Because `Var` objects are immutable, it's impossible to create a dependency cycle, and the nodes are placed in subgraphs accordingly to where they are required. If more advanced usage examples are necessary do let me know and I'll add them to the documentation. 
Spox does not interfere with the function signatures or the AST, so almost everything should work - as long as internal Spox state is left unmodified.

> [onnx-script](https://github.com/microsoft/onnx-script) is more simple in that regard

If you're referring to specifically control flow & usability in Spox vs onnx-script, then my personal opinion is that it may seem easier, but due to the inherent AST manipulation many constructs legal in Python (and Spox) will likely never be allowed in onnx-script. If it proves a large usability goal I might consider implementing a Spox addon (in a separate library) for control flow syntax (like onnx-script), but initial experiments only showed me how many pit falls there are in this approach.

xadupre:
I'll close the PR. It was an interesting discussion at least for me. spox and onnx-script chose two different paths to build a more user friendly API in order to create onnx graphs. In my case, I need a numpy friendly one. The onnx API (make_node, make_graph, ...) is usually used on small graphs. When many more graphs are built and bigger ones, a new API is usually developped to reduce redundancy in the code and be more agile. So far, I only saw new APIs every time it was needed.

Every time a new version of onnx is released, we need to make sure that the converting libraries support the new opset but we usually need onnxruntime to be released first. It takes week. Adding a new dependency based on onnx would extend that delay as opposed to have a more user friendly API added to the onnx package.

adityagoel4512:
Many people on this thread will be familiar with this by now but we've actually implemented the an ONNX-backed array library called [ndonnx](https://github.com/Quantco/ndonnx) that implements the Array API standard. It actually builds on top of Spox and provides a convenient abstraction for building ONNX graphs (as well as having the code reusability benefits of the Array API). Feel free to check it out: https://github.com/Quantco/ndonnx.

PR #4980, opened 2023-03-11
Fix UnionTypeInfo bug

gramalingam:
### Description
Fix bug in `UnionTypeInfo` utility. When taking the union of two types, if either type lacks a shape, then the union should also lack a shape. (Unfortunately, protobuf implicitly creates an empty shape-field if the shape-field is accessed when it does not exist, and all accesses must be protected by a check if the field has a shape.)


justinchuby:
Maybe this is the same fix for

```
E   onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Node (_0xb7ddbb0_n5) Op (If) [TypeInferenceError] Graph attribute inferencing failed: Node (_0xb7ddbb0_n0) Op (Transpose) [TypeInferenceError] Invalid attribute perm {1, 0}, input shape = {2}

The above exception was the direct cause of the following exception:
onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py:956: in test_output_match
    function_output = self.function_executor(flattened_torch_outputs)(
onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py:847: in _capture_graph_and_evaluate_torch_script_evaluator
    raise AssertionError(
E   AssertionError: ONNX Runtime failed to evaluate:
E   Inputs:
E   {'input_0': array([-2.1084523,  8.267502 ], dtype=float32)}
E   Model:
E   <
E      ir_version: 8,
E      opset_import: ["" : 18, "onnxscript.atenlib" : 1],
E      producer_name: "pytorch",
E      producer_version: "2.1.0"
E   >
E   torch_jit (float[2] input_0) => (float[2] 1) {
E      1 = onnxscript.atenlib.aten_t (input_0)
E   }
E   <
E     domain: "onnxscript.atenlib",
E     opset_import: ["" : 18]
E   >
E   aten_t (self) => (result_2)
E   {
E      tmp = Shape (self)
E      rank = Size (tmp)
E      tmp_0 = Constant <value = int64 tmp_0 {2}> ()
E      tmp_0_cast = CastLike (tmp_0, rank)
E      cond = Equal (rank, tmp_0_cast)
E      result_2 = If (cond) <then_branch = thenGraph_6 () => ( result) {
E         result = Transpose <perm = [1, 0]> (self)
E      }, else_branch = elseGraph_6 () => ( result_1) {
E         result_1 = Identity (self)
E      }>
E   }
```

?

onnx/defs/shape_inference.cc
```diff
@@ -249,16 +249,23 @@ void UnionTypeInfo(const TypeProto& source_type, TypeProto& target_type) {
...
       fail_type_inference(
           "Mismatched sparse tensor element type:", " source=", source_elem_type, " target=", target_elem_type);
     }
-
-    UnionShapeInfoForTensor(source_type.sparse_tensor_type().shape(), *target_type.mutable_sparse_tensor_type());
+    if (source_type.sparse_tensor_type().has_shape()) {
+      UnionShapeInfoForTensor(source_type.sparse_tensor_type().shape(), *target_type.mutable_sparse_tensor_type());
+    } else {
+      target_type.mutable_sparse_tensor_type()->clear_shape();
+    }
```
jcwchen, line 268: If UnionShapeInfoForTensor is not a public function, is it possible for us to change its interface to make it take source_type as an input instead of shape and then handle the has_shape() stuff inside the function?

gramalingam: That sounds good. But, this function is exposed in the shape_inference.h file. We probably should not expose these functions there, but it is a bit too late ... we could deprecate it and remove it at some point, I suppose.

gramalingam: I refactored the code as suggested, creating new overloads for `UnionShapeInfo`.

jcwchen (approved):
LGTM. Please fix Lint. Thanks!

onnx/defs/shape_inference.cc
```diff
@@ -230,6 +230,22 @@ void UnionShapeInfo(const TensorShapeProto& source_shape, TypeProto_Tensor& targ
   UnionShapeInfoForTensor(source_shape, target_type);
 }
 
+void UnionShapeInfo(const TypeProto_Tensor& source_type, TypeProto_Tensor& target_type) {
+  if (source_type.has_shape()) {
```
wschin, line 234: Please explain the reason behind the code. E.g.,
```
We merge shapes from source and target if at least one of them exists. Otherwise, union with empty shape should remain empty shape.

**Review by wschin — APPROVED** [review]
Looks good overall. My only comment is to document properly (e.g., describing the design concepts involved in a function so that people don't need to reverse-engineering when reading the code).

File: onnx/defs/shape_inference.cc [hunk 3]
```diff
@@ -230,6 +230,22 @@ void UnionShapeInfo(const TensorShapeProto& source_shape, TypeProto_Tensor& targ
   UnionShapeInfoForTensor(source_shape, target_type);
 }
 
+void UnionShapeInfo(const TypeProto_Tensor& source_type, TypeProto_Tensor& target_type) {
```
wschin, line 233: Not sure if this should also call
```
source_type.clear_shape()
```
when target_type.has_shape() == false.

onnx/defs/shape_inference.cc
gramalingam: Done, thanks!

PR #5000, opened 2023-03-15
Write backend test models as text proto

### LINKED ISSUE #4995: [Feature request] Support text proto
### System information

_No response_

Models exported can be stored as text proto for readability. 

cc @jcwchen 

### Alternatives considered

_No response_

### Describe the feature

_No response_

_No response_

### Feature Area

_No response_

Yes

### Notes

_No response_

justinchuby:
### Description

Models exported can be stored as text proto for readability.

This is especially true for the reference test models in this repo. Otherwise they are not verifiable in code reviews.

### Motivation and Context

Fixes #4995


xadupre:
Two concerns:
* It is going to increase the size of onnx repository, onnx package. That would be good to quantify it.
* This has a significant impact on all runtimes using onnx backend to validate their runtime. That should not change the libraries using the python backend API but that will impact all other languages reading the existing onnx files.

The model in the backend are usually graph with one single node built from the test cases. The differences observed in the models are the same as the ones observed in the tests. netron can be used too. Is there any other way to solve that issue rather than changing the format used to stored the onnx models?

PR #5001, opened 2023-03-15
Create testing infrastructure for type-checking function definitions

gramalingam:
### Description
The existing testing infrastructure does not catch certain type-errors in function-definitions for ops in the ONNX standard. (E.g., the [function-definition](https://github.com/onnx/onnx/pull/4979) for `Softplus` was missing a `CastLike`.)

This PR introduces a test that type-checks function-definitions with respect to all allowed types for its inputs.

(The test reveals a few issues with existing function-definitions that remains to be resolved.)


onnx/test/cpp/function_verify_test.cc
```diff
@@ -102,6 +103,175 @@ void VerifyTypeConstraint(const OpSchema& function_op, const FunctionProto* func
   ++counter;
 }
 
+// Testing the function-definitions provided for function-ops in ONNX schema registry.
```
justinchuby, line 106: I was thinking to do similar things in the atenlib. I wonder if this can be exposed from onnx or adapted for onnxscript somehow?

gramalingam: Something like that would be helpful, yes. But, for now, since the functions are available in ONNX's opschema, this was simpler, while in onnxscript we would want to do it for custom functions. We should add some abstraction to support both. I didn't want to round-trip the FunctionProto from C++ to Python and back to C++ for the tests below.

liqunfu (approved):

onnx/test/cpp/function_verify_test.cc
```diff
@@ -102,6 +103,175 @@ void VerifyTypeConstraint(const OpSchema& function_op, const FunctionProto* func
...
+          :
+          // Select the type bound to the type-var in current instantiation
+          typeVarBindings[input.GetTypeStr()];
+      input_types.push_back(Utils::DataTypeUtils::ToTypeProto(datatype));
+    }
+
+    for (auto& attribute_vals : *attribute_cases) {
+      ONNX_TRY {
+        auto output_types = shape_inference::InferFunctionOutputTypes(function_proto, input_types, attribute_vals);
+        // recordSuccess(attribute_vals);
```
daquexian, line 257: Should we uncomment this line?

daquexian (approved):

onnx/test/cpp/function_verify_test.cc
gramalingam: Done, thanks!

PR #5002, opened 2023-03-16
Update TypeAnnotations.md

justinchuby:
### Description

Remove outdated writing in TypeAnnotations.md


jcwchen (approved):
It would be great if we can have a new example for our current type annotations, but anyway we should remove the old way here. Thanks for catching these.

docs/TypeAnnotations.md
```diff
@@ -16,50 +17,7 @@ Mypy is an opt-in type checker. Annotations are optional and, if present, will b
...
-
-    $ pip3 install mypy
-
-Running then works as it does above for python 3.
-
-    ~/onnx $ python setup.py typecheck
-
-Since you've installed mypy manually, you won't get updates for it when we change the version of the mypy dependency in setup.py, so if you're seeing a CI error and can't reproduce it locally, check your mypy version.
 
 ### What should I annotate?
```
xadupre, line 21: This page has one title, one subtitle, one subsubtitle. Maybe too much?

justinchuby: Yeah maybe I can removed completely ? I just remove outdated info without touching the rest

justinchuby:
As type annotations are standard nowadays I think we should either remove this file or just strip the legacy notes

docs/TypeAnnotations.md
xadupre: Maybe remove it completely is not a bad idea.

jcwchen (approved):
Please fix DCO and I will merge it. Thanks!

PR #5011, opened 2023-03-17
Add proto support for overloaded functions

gramalingam:
A minimal and simple extension to the proto definitions to allow multiple (overloaded) functions in its model-local function list with same (domain, name).

Model-local functions in ONNX provide a default-implementation for certain ops used in the model (expressed in terms of other primitive ops). This extension will allow us to provide different default-implementations for different calls to the same op (which might differ in the types of input-arguments or in the values of attribute-parameters).

* Update proto
* Update helper.py constructors
* Update parser
* Update shape inference implementation
* Update in-memory IR
* Add test-case
* Minimal extensions in inliner to support this


jbachurski:
Just wanted to say I think this is a really good idea. Functions are currently way too limited to be useful for a runtime/tooling IMO. Adding overloads can be a good first step towards actually expressing a wider part of the standard with them. If I understand correctly, they would be parametrised by input types and attribute values which seems like exactly the way to go.

BowenBao:
Thank you @gramalingam ! This will enable clean and unified function names for dynamo onnx exported specialized nn.Module functions, as well as potentially the currently inlined 'trace_only' functions in onnxscript torchlib.

One thing to point out though this does appear to vary slightly with traditional pl function overload concept, whereas here it is possible to have two function overloads with exact same input and attribute types. I think following the pl convention would restrict the usefulness in above mentioned use cases, but I'd love to learn about concerns if any.

docs/IR.md
```diff
@@ -190,6 +190,7 @@ A serialized function (a _FunctionProto_) has the following properties:
 |---|---|---|
 name|string|The name of the function
 domain|string|The domain to which this function belongs
+overload|string|Part of unique id of function (added in IR version 9)
```
BowenBao, line 193: is it correct to understand overload is an optional string that maps to "" when unprovided?

You probably also need to update `Nodes` section with the new field.

BowenBao:
cc @justinchuby, @pranavsharma

gramalingam:
Also: @yuslepukhin for comments/feedback if any

docs/IR.md
gramalingam: Correct re. first point. Good catch re. second point, will add it.

xadupre:
The new string would be consumed by a NodeProto to avoid any ambiguity or can it be left empty or is it just an extension of the function name? I assume this PR goes with #5903.

BowenBao (commented):
Is `onnx.printer.to_text` updated to reflect this change?

justinchuby (approved):

gramalingam:
> The new string would be consumed by a NodeProto to avoid any ambiguity or can it be left empty or is it just an extension of the function name? I assume this PR goes with #5903.

Both FunctionProto and NodeProto have this new field. But it is an optional string field. It can be left empty. As I understand Protobuf, older versions (IR version 9 and below), when read in (with new proto), it will have an empty string value.

When both FunctionProto and NodeProto have this empty, the behavior is same as before. However, if this is non-empty, then it is used to match the names from NodeProto to FunctionProto.

gramalingam:
> Is `onnx.printer.to_text` updated to reflect this change?

Good catch. I updated it, thanks.

gramalingam:
@xadupre : one thing pending is updating the reference implementation. Do you know where it looks up a model-local function to execute a node?

justinchuby:
> A question on the syntax (used in onnxtext format): I use "opname:overload" to indicate the overload. Please let me know if you have any suggestions: eg., if you think "opname::overload" would be better. A single colon seems sufficient and minimal. Double colon has the advantage of being aligned with C++ notation.

Does `.` work like what PyTorch has?

justinchuby:
Hmm maybe not since domain can be dot separated already. `:` LGTM

justinchuby:
Should op schemas be extended to support function overloads?

PR #5020, opened 2023-03-21
Allow new `OpSchema`s to be initialized in Python

### LINKED ISSUE #5019: [Feature request] Make OpSchema writable from Python

Not being able to auto create OpSchema for functions in Python; not being able to do shape inference for custom ops.

### Alternatives considered

Create a Python class that mirrors OpSchema. We can do that and expose a similar interface, but for onnx shape inference to work, we need to be able to create an OpSchema, register it and write shape inference functions in Python.

### Describe the feature

Extend the OpSchema to be writable, and support registering this schema with ONNX.

- [x] #5020 
- [x] https://github.com/onnx/onnx/issues/5292
- [x] repr methods for the classes #5117 
- [x] OpSchema registration https://github.com/onnx/onnx/pull/5906
- [x] Shape inference binding: create inference code with Python

Y. It will be an extension of the current OpSchema api.

### Feature Area

API

Yes

### Notes

assigned to @justinchuby

justinchuby:
### Description

**This PR needs to be in the 1.14 release**

The change

- Creates Python constructors for `OpSchema`, `Attribute`, `FormalParameter `and `TypeConstraintParam`.
- Adds overloads for `Schema::Input` and `Schema::Output` that takes a `FormalParameter`
- Updates the pyi file to reflect the change.
- Hides the camel cased fields in `FormalParameter` for API consistency. snake_case versions are created and exposed, but this will not break existing usage as the camelCase versions are still retained, just not annotated in the pyi file. So new users will only discover the snake_case versions.

Code change

- [x] pybind update
- [x] Tests
- [x] Allow specifying default values in Attribute

Next PRs

- Documentation
- `repr` methods for the classes
- OpSchema registration
- Shape inference binding: create inference code with Python

### Motivation and Context

 #5019


justinchuby:
cc @jcwchen @gramalingam

onnx/test/schema_test.py
```diff
@@ -20,5 +20,38 @@ def test_function_body(self) -> None:
...
+        self.assertEqual(schema.doc, "Test doc")
+
+    def test_domain_can_be_set(self) -> None:
+        schema = defs.OpSchema()
+        schema.domain = "TestDomain"
+        self.assertEqual(schema.domain, "TestDomain")
+
+    def test_add_attribute_adds_one_attribute(self) -> None:
+        schema = defs.OpSchema()
+        schema.add_attribute("TestAttr", "TestAttr doc", defs.OpSchema.AttrType.INT, required=True)
```
justinchuby, line 44: Make description optional.

justinchuby: Done

justinchuby:
@jcwchen could you help add this to the milestone?

jcwchen:
Added. Please note that the code freeze date is 4/7.

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,187 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
   py::register_exception<SchemaError>(defs, "SchemaError");
 
   py::class_<OpSchema> op_schema(defs, "OpSchema", "Schema of an operator.");
-  op_schema.def_property_readonly("file", &OpSchema::file)
+  op_schema.def(py::init<>());
+
+  // Define the class enums first because they are used as default values in function definitions
+  py::enum_<OpSchema::FormalParameterOption>(op_schema, "FormalParameterOption")
```
justinchuby, line 121: Just moved up from below so that pybind11 sees them when they are used in the opschema definition

onnx/onnx_cpp2py_export/defs.pyi
```diff
@@ -73,20 +100,32 @@ class OpSchema:
...
+            differentiation_category: OpSchema.DifferentiationCategory = OpSchema.DifferentiationCategory.Unknown,  # noqa: F821
+        ) -> None: ...
         @property
         def name(self) -> str: ...
         @property
-        def types(self) -> Set[str]: ...
+        def types(self) -> set[str]: ...
         @property
-        def typeStr(self) -> str: ...
+        def type_str(self) -> str: ...
```
xadupre, line 118: The name follows python convention but i wonder if we should keep the old name for a couple of releases with a warning saying it is going to be deprecated. Some tools may break if they use these names to automate documentation...

liqunfu (commented):
LGTM

justinchuby:
MacOS

```
In file included from /Users/justinc/Documents/GitHub/onnx/onnx/cpp2py_export.cc:6:
In file included from /opt/homebrew/include/pybind11/stl.h:17:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/map:726:17: error: no viable overloaded '='
        __ref() = __v.__get_value();
        ~~~~~~~ ^ ~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__tree:1662:39: note: in instantiation of member function 'std::__value_type<std::string, onnx::OpSchema::Attribute>::operator=' requested here
            __cache.__get()->__value_ = *__first;
                                      ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__tree:1619:9: note: in instantiation of function template specialization 'std::__tree<std::__value_type<std::string, onnx::OpSchema::Attribute>, std::__map_value_compare<std::string, std::__value_type<std::string, onnx::OpSchema::Attribute>, std::less<std::string>, true>, std::allocator<std::__value_type<std::string, onnx::OpSchema::Attribute>>>::__assign_multi<std::__tree_const_iterator<std::__value_type<std::string, onnx::OpSchema::Attribute>, std::__tree_node<std::__value_type<std::string, onnx::OpSchema::Attribute>, void *> *, long>>' requested here
        __assign_multi(__t.begin(), __t.end());
        ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/map:1041:21: note: in instantiation of member function 'std::__tree<std::__value_type<std::string, onnx::OpSchema::Attribute>, std::__map_value_compare<std::string, std::__value_type<std::string, onnx::OpSchema::Attribute>, std::less<std::string>, true>, std::allocator<std::__value_type<std::string, onnx::OpSchema::Attribute>>>::operator=' requested here
            __tree_ = __m.__tree_;
                    ^
/Users/justinc/Documents/GitHub/onnx/onnx/defs/schema.h:146:7: note: in instantiation of member function 'std::map<std::string, onnx::OpSchema::Attribute>::operator=' requested here
class OpSchema final {
      ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__utility/pair.h:257:11: note: candidate function not viable: no known conversion from 'const std::__value_type<std::string, onnx::OpSchema::Attribute>::value_type' (aka 'const pair<const std::string, onnx::OpSchema::Attribute>') to 'const typename conditional<is_copy_assignable<first_type>::value && is_copy_assignable<second_type>::value, pair<string &, Attribute &>, __nat>::type' (aka 'const std::__nat') for 1st argument
    pair& operator=(typename conditional<
          ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__utility/pair.h:270:11: note: candidate function not viable: no known conversion from 'const std::__value_type<std::string, onnx::OpSchema::Attribute>::value_type' (aka 'const pair<const std::string, onnx::OpSchema::Attribute>') to 'typename conditional<is_move_assignable<first_type>::value && is_move_assignable<second_type>::value, pair<string &, Attribute &>, __nat>::type' (aka 'std::__nat') for 1st argument
    pair& operator=(typename conditional<
          ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__utility/pair.h:79:62: note: candidate template ignored: disabled by 'enable_if' [with _Tuple = const std::pair<const std::string, onnx::OpSchema::Attribute> &]
    using _EnableB _LIBCPP_NODEBUG_TYPE = typename enable_if<_Val, bool>::type;
                                                             ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__utility/pair.h:42:29: note: candidate function (the implicit copy assignment operator) not viable: no known conversion from 'const pair<const std::__value_type<std::string, onnx::OpSchema::Attribute>::key_type, std::__value_type<std::string, onnx::OpSchema::Attribute>::mapped_type>' to 'const pair<std::string &, onnx::OpSchema::Attribute &>' for 1st argument
struct _LIBCPP_TEMPLATE_VIS pair
                            ^
1 error generated.
```

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,187 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+          "__init__",
+          [](OpSchema::FormalParameter& self,
+             std::string name,
+             std::string type_str,
+             const std::string& description,
+             OpSchema::FormalParameterOption param_option,
+             bool is_homogeneous,
+             int min_arity,
+             OpSchema::DifferentiationCategory differentiation_category) {
+            // Use a lambda to swap the order of the arguments to match the Python API
```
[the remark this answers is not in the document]
justinchuby: Yes. This function was doing the swap?

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,187 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+      .def_property_readonly("type_str", &OpSchema::FormalParameter::GetTypeStr)
+      .def_property_readonly("description", &OpSchema::FormalParameter::GetDescription)
+      .def_property_readonly("option", &OpSchema::FormalParameter::GetOption)
+      .def_property_readonly("is_homogeneous", &OpSchema::FormalParameter::GetIsHomogeneous)
+      .def_property_readonly("min_arity", &OpSchema::FormalParameter::GetMinArity)
+      .def_property_readonly("differentiation_category", &OpSchema::FormalParameter::GetDifferentiationCategory)
+      // Legacy camel cased names. We retain them for backward compatibility.
+      .def_property_readonly("typeStr", &OpSchema::FormalParameter::GetTypeStr)
+      .def_property_readonly("isHomogeneous", &OpSchema::FormalParameter::GetIsHomogeneous)
+      .def_property_readonly("differentiationCategory", &OpSchema::FormalParameter::GetDifferentiationCategory);
```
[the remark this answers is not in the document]
justinchuby: > Potentially it may bring a lot of deprecation message.

That's what I worry, but I also think that's OK. I will add a message if there's no objection.

onnx/onnx_cpp2py_export/defs.pyi
justinchuby: I did keep the old names. I just removed them from the pyi file so people will not discover them easily.

onnx/cpp2py_export.cc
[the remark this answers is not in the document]
justinchuby: Done

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,187 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+
+      .def_property_readonly("name", &OpSchema::FormalParameter::GetName)
+      .def_property_readonly("types", &OpSchema::FormalParameter::GetTypes)
+      .def_property_readonly("type_str", &OpSchema::FormalParameter::GetTypeStr)
+      .def_property_readonly("description", &OpSchema::FormalParameter::GetDescription)
+      .def_property_readonly("option", &OpSchema::FormalParameter::GetOption)
+      .def_property_readonly("is_homogeneous", &OpSchema::FormalParameter::GetIsHomogeneous)
+      .def_property_readonly("min_arity", &OpSchema::FormalParameter::GetMinArity)
+      .def_property_readonly("differentiation_category", &OpSchema::FormalParameter::GetDifferentiationCategory)
+      // Legacy camel cased names. We retain them for backward compatibility.
```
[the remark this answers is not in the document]
justinchuby: Done

BowenBao:
> MacOS

Chatted offline to try g++/clang++ instead of AppleClang++ on mac. See how it goes.

onnx/cpp2py_export.cc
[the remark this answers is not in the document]
andife: I wonder what does deprecation mean in that case? Is removed in e.g. 2 releases? or is there again discussed? or the other variant we recommended? Am just curious ;-)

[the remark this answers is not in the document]
justinchuby: I am proposing to remove on 1.16. Tracked by #5074. The recommended variants are the snake_case ones

justinchuby:
Tested with g++ and clang++. Got same errors

justinchuby:
I think this may be a clue, but I can't parse it

```
onnx/onnx/defs/schema.h:146:7: note: in instantiation of member function 'std::map<std::string, onnx::OpSchema::Attribute>::operator=' requested here
class OpSchema final {
      ^
```

justinchuby:
Isolated the error from `OpSchema.__init__`

justinchuby:
Error in inputs, outputs, type_constraints and/or attributes. Can't seem to isolate further

justinchuby:
Should I make something move_assignable

justinchuby:
Is it this thing

image

justinchuby:
Or copy assignable

justinchuby:
Could it be this

```
OpSchema& OpSchema::Attr(Attribute attr) {
  auto name = attr.name; // copy name so we can move attr in the next line
  attributes_.insert(std::make_pair(std::move(name), std::move(attr)));
  return *this;
}
```

onnx/defs/schema.h
```diff
@@ -438,8 +438,8 @@ class OpSchema final {
           required(false),
           default_value(std::move(default_value_)) {}
 
-    const std::string name;
-    const std::string description;
+    std::string name;
+    std::string description;
```
justinchuby, line 442: But why?

justinchuby: ChatGPT says:

-------

Yes, using placement new to reinitialize an object at the same memory location can be a good option in some cases. However, you need to be careful when using placement new with const fields.

If your OpSchema class has const fields, you cannot reassign their values directly. Instead, you need to use a constructor that initializes those fields during object construction. One way to achieve this is to use an initialization list in the constructor.

Here's an example that demonstrates how to use placement new to reinitialize an object of type OpSchema at the same memory location:

```c++
void resetOpSchema(OpSchema& self, const OpSchema& schema) {
    // Call destructor to destroy the existing object
    self.~OpSchema();

    // Use placement new to construct a new object at the same memory location
    new (&self) OpSchema(schema);
}
```

In this example, the resetOpSchema function takes a reference to an existing OpSchema object (self) and a reference to a const OpSchema object (schema). It first calls the destructor of OpSchema to destroy the existing object at the memory location pointed to by self.

Next, it uses placement new to construct a new OpSchema object at the same memory location. The new operator calls the constructor of OpSchema and passes schema as its argument. The constructor initializes all non-const fields of the OpSchema object, including any const fields that are initialized in the constructor's initialization list.

Note that this technique assumes that the existing object at the memory location pointed to by self was previously constructed using placement new. If the object was constructed using other means, you need to make sure that its memory allocation and alignment match the requirements of the OpSchema class.

justinchuby: I asked: Should I just remove the const fields instead?

justinchuby: It replied

Whether or not to remove const fields from the OpSchema class depends on the requirements of your application. Removing const fields can make it easier to reinitialize an object using placement new, but it may also affect the semantics of the class.

If the const fields are there to enforce invariants or prevent unintended modifications to the object, removing them may compromise the correctness of your code. On the other hand, if the const fields are not critical to the behavior of the class and removing them does not affect the correctness of your code, then it may be a reasonable option.

If you need to reinitialize an object of a class that has const fields, you can consider using a copy constructor or a move constructor to create a new object and then swap it with the existing object. This approach avoids the need to reinitialize the existing object and can be more efficient than copying the entire object using placement new.

Here's an example that demonstrates how to use a copy constructor to create a new OpSchema object and swap it with the existing object:

```c++
void resetOpSchema(OpSchema& self, const OpSchema& schema) {
    // Create a temporary object using the copy constructor
    OpSchema temp(schema);

    // Swap the temporary object with the existing object
    std::swap(self, temp);
}
```

In this example, the resetOpSchema function creates a temporary OpSchema object temp using the copy constructor, which initializes all fields of the object. It then uses std::swap to exchange the contents of self and temp. The effect is to reinitialize self with the contents of temp, while leaving temp uninitialized. This approach can be more efficient than copying the entire object using placement new, especially if the class has many fields or if the const fields are expensive to copy.

xadupre:
Is it possible to add an example in the documentation to show how to create a new operator schema?

onnx/defs/schema.h
xadupre: Pybind11 should try to map std::map to a dictionary. Maybe you can manually create a py::dict to export the attributes. That should bypass some of the pybind11 templates.

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,207 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+             std::string doc,
+             std::vector<OpSchema::FormalParameter> inputs,
+             std::vector<OpSchema::FormalParameter> outputs,
+             std::vector<std::tuple<std::string, std::vector<std::string>, std::string>> type_constraints,
+             std::vector<OpSchema::Attribute> attributes) {
+            auto schema = OpSchema().SetName(name).SetDomain(domain).SinceVersion(since_version).SetDoc(doc);
+            // Add inputs and outputs
+            for (auto i = 0; i < inputs.size(); ++i) {
+              const auto& input = inputs[i];
+              schema = schema.Input(
```
taoyouh, line 260: You are using `OpSchema::operator=` here, while OpSchema::Attribute had a const field and was not assignable.

justinchuby:
> Is it possible to add an example in the documentation to show how to create a new operator schema?

Which page should I add it to? @xadupre

justinchuby:
Mystery solved. Thanks @taoyouh for helping me analyze. 

Initially I assigned the result of `schema.Input` to `schema` (`schema = schema.Input(...)`) in the pybind constructor. This triggers a move assignment of `OpSchema`. In `Schema` there is a `std::map<std::string, Attribute>` in which `Attribute` has two `const` fields `name` and `description`. When the move assignment happens on `OpSchema`, due to Apple's STL implementation, `std::map`'s move constructor will try to move the `const` fields `name` and `description` which are not movable. The error message was definitely not helpful when it comes to locating the error.

Instead of removing the `const` of `name` and `description` to make Attribute movable, I opted to initialize in the beginning of the pybind constructor and use its methods to modify its fields inplace. So there is no more moving that needs to happen.

onnx/test/schema_test.py
```diff
@@ -15,11 +19,234 @@ def test_typecheck(self) -> None:
...
+                defs.OpSchema.FormalParameter("input2", "T"),
+            ],
+            outputs=[
+                defs.OpSchema.FormalParameter("output1", "T"),
+                defs.OpSchema.FormalParameter("output2", "T"),
+            ],
+            type_constraints=[("T", ["tensor(int64)"], "")],
+            attributes=[
+                defs.OpSchema.Attribute(
+                    "attr1", defs.OpSchema.AttrType.INTS, "attr1 description"
```
justinchuby, line 68: Should we use AttributeProto here?

justinchuby:
I can't reproduce the notebook test errors

justinchuby:
Fixed

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,215 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+            // Use a lambda to swap the order of the arguments to match the Python API
+            return OpSchema::Attribute(name, description, type, required);
+          }),
+          py::arg("name"),
+          py::arg("type"),
+          py::arg("description") = "",
+          py::kw_only(),
+          py::arg("required") = true)
+      .def(
+          py::init([](std::string name, const py::object& default_value, std::string description) {
```
justinchuby, line 163: @jcwchen for some reason I cannot take `AttributeProto` as the parameter. If so pybind will reject `onnx.AttributeProto`.

onnx/defs/schema.h
justinchuby: Thanks for the suggestion. I think I found a way

onnx/test/schema_test.py
```diff
@@ -15,11 +19,244 @@ def test_typecheck(self) -> None:
...
+    def test_init_without_optional_arguments(self) -> None:
+        op_schema = defs.OpSchema("test_op", "test_domain", 1)
+        self.assertEqual(op_schema.name, "test_op")
+        self.assertEqual(op_schema.domain, "test_domain")
+        self.assertEqual(op_schema.since_version, 1)
+        self.assertEqual(len(op_schema.inputs), 0)
+        self.assertEqual(len(op_schema.outputs), 0)
+        self.assertEqual(len(op_schema.type_constraints), 0)
+
+    # ChatGPT generated tests
```
xadupre, line 108: Should we keep that comment?

justinchuby: I can remove

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,215 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+                  output.GetDifferentiationCategory());
+            }
+            // Add type constraints
+            for (const auto& type_constraint : type_constraints) {
+              std::string type_str;
+              std::vector<std::string> constraints;
+              std::string description;
+              tie(type_str, constraints, description) = type_constraint;
+              self.TypeConstraint(type_str, constraints, description);
+            }
```
edgchen1, line 296: can use structured bindings and move from `type_constraints`
```suggestion
            for (auto& [type_str, constraints, description] : type_constraints) {
              self.TypeConstraint(std::move(type_str), std::move(constraints), std::move(description));
            }
```

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,215 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+                      std::vector<OpSchema::FormalParameter> outputs,
+                      std::vector<std::tuple<std::string, std::vector<std::string>, std::string>> type_constraints,
+                      std::vector<OpSchema::Attribute> attributes) {
+            auto self = OpSchema();
+
+            self.SetName(name).SetDomain(domain).SinceVersion(since_version).SetDoc(doc);
+            // Add inputs and outputs
+            for (auto i = 0; i < inputs.size(); ++i) {
+              const OpSchema::FormalParameter& input = inputs[i];
+              self.Input(
```
edgchen1, line 267: would it be worth adding OpSchema::Input/Output overloads that take an OpSchema::FormalParameter directly?

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,215 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+              std::string type_str;
+              std::vector<std::string> constraints;
+              std::string description;
+              tie(type_str, constraints, description) = type_constraint;
+              self.TypeConstraint(type_str, constraints, description);
+            }
+            // Add attributes
+            for (const auto& attribute : attributes) {
+              self.Attr(attribute);
+            }
```
edgchen1, line 300:
```suggestion
            for (auto& attribute : attributes) {
              self.Attr(std::move(attribute));
            }
```

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,215 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+                      std::string domain,
+                      int since_version,
+                      std::string doc,
+                      std::vector<OpSchema::FormalParameter> inputs,
+                      std::vector<OpSchema::FormalParameter> outputs,
+                      std::vector<std::tuple<std::string, std::vector<std::string>, std::string>> type_constraints,
+                      std::vector<OpSchema::Attribute> attributes) {
+            auto self = OpSchema();
+
+            self.SetName(name).SetDomain(domain).SinceVersion(since_version).SetDoc(doc);
```
edgchen1, line 263:
```suggestion
            self.SetName(std::move(name)).SetDomain(std::move(domain)).SinceVersion(since_version).SetDoc(doc);
```

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,215 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+          py::kw_only(),
+          py::arg("required") = true)
+      .def(
+          py::init([](std::string name, const py::object& default_value, std::string description) {
+            // Construct an attribute with a default value.
+            // Attributes with default values are not required
+            auto bytes = default_value.attr("SerializeToString")().cast<py::bytes>();
+            AttributeProto proto{};
+            ParseProtoFromPyBytes(&proto, bytes);
+            return OpSchema::Attribute(name, description, proto);
```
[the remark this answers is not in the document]
justinchuby: I remember reading that the compiler will know to move automatically in some cases. Is there a good rule of thumb on when that will happen?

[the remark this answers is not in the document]
edgchen1: we can start with this advice from [here](https://herbsutter.com/2020/02/17/move-simply/):

>To pass a named object `a` as an argument to a `&&` “move” parameter (rvalue reference parameter), write `std::move(a)`. That’s pretty much the only time you should write std::move, because C++ already uses move automatically when copying from an object it knows will never be used again, such as a temporary object or a local variable being returned or thrown from a function.

a named object is an lvalue, so we would need to cast it to an rvalue reference which is just what std::move does.

in this code, we also deal with functions which take a string (or some other type we might not want to copy) by value, for example, `OpSchema::Attribute::Attribute()`. this way, we can avoid writing two versions of this constructor, one taking a `const string&` (where we copy the parameter) and the other taking a `string&&` (where we move from the parameter). so we can use `std::move(variable_name)` here, in the spirit of the advice above, to move into the by-value parameter instead of copy into it.

some more details on this pass-by-value pattern [here](https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference).

as another example, the compiler will move without us writing std::move if we pass an rvalue as the parameter:
`Attribute attr(GetName(), ...);`

onnx/cpp2py_export.cc
justinchuby: This looks great! I did a quick search and it looks like structured bindings are since c++17? @jcwchen is 17 the lowest version we support?

justinchuby:
I have to collapse all review comments but I think this is great for future reference. So I am reproducing Ed's response here:

https://github.com/onnx/onnx/pull/5020#discussion_r1156643396

### When do we need to explicitly move

> we can start with this advice from [here](https://herbsutter.com/2020/02/17/move-simply/):
> 
> > To pass a named object `a` as an argument to a `&&` “move” parameter (rvalue reference parameter), write `std::move(a)`. That’s pretty much the only time you should write std::move, because C++ already uses move automatically when copying from an object it knows will never be used again, such as a temporary object or a local variable being returned or thrown from a function.
> 
> a named object is an lvalue, so we would need to cast it to an rvalue reference which is just what std::move does.
> 
> in this code, we also deal with functions which take a string (or some other type we might not want to copy) by value, for example, `OpSchema::Attribute::Attribute()`. this way, we can avoid writing two versions of this constructor, one taking a `const string&` (where we copy the parameter) and the other taking a `string&&` (where we move from the parameter). so we can use `std::move(variable_name)` here, in the spirit of the advice above, to move into the by-value parameter instead of copy into it.
> 
> some more details on this pass-by-value pattern [here](https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference).
> 
> as another example, the compiler will move without us writing std::move if we pass an rvalue as the parameter: `Attribute attr(GetName(), ...);`

onnx/cpp2py_export.cc
justinchuby: I think so! I will add

justinchuby:
Thanks @edgchen1 for kindly reviewing this. I will make changes following the suggestions

justinchuby:
@edgchen1 When we move the strings from the pybind apis, will python lose access to them? Should I not use a value after it was moved?

onnx/cpp2py_export.cc
jcwchen: https://github.com/onnx/onnx/blob/515361a12ec6dace2e99f761490cb55fc2696168/CMakeLists.txt#L60-L64

No. Only Windows build requires C++17 or higher. We didn't set the corresponding required version for other platforms, because some of partners might still use lower C++ versions. Please let me know if you really want C++17 features and we can review the upgrade, although I think it might be too close to the upcoming release and can only happen to the one after that.

justinchuby: SG

onnx/test/schema_test.py
justinchuby: Done

justinchuby:
I will follow up with documentation updates to keep this PR manageable. Also so that we can merge faster and iterate on wording in the doc

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,203 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+      .def_property_readonly("is_homogeneous", &OpSchema::FormalParameter::GetIsHomogeneous)
+      .def_property_readonly("min_arity", &OpSchema::FormalParameter::GetMinArity)
+      .def_property_readonly("differentiation_category", &OpSchema::FormalParameter::GetDifferentiationCategory)
+      // Legacy camel cased names. We retain them for backward compatibility.
+      // TODO(#5074): Remove these before the 1.16 release.
+      .def_property_readonly(
+          "typeStr",
+          [](const OpSchema::FormalParameter& self) {
+            auto warnings = py::module::import("warnings");
+            warnings.attr("warn")(
```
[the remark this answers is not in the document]
justinchuby: Is there a deprecation policy in onnx? Does 1.16 sound reasonable?

edgchen1:
> When we move the strings from the pybind apis, will python lose access to them?
> Looks like they will be copied when the functions are being called?

Since the string is passed by value, a string instance will be constructed (could be from a copy or a move from rvalue reference) for the argument. On the other hand, if it was passed by non-const reference, the caller might have to worry about it getting modified (e.g., moved from) even when the caller doesn't use `std::move(x)`. I'm not sure what the pybind calling code looks like, possibly the string is copied there if Python still needs it.

> Should I not use a value after it was moved?

Right. Generally, for types that support moving, moved from objects are in a valid but unspecified state. You could do limited things with such an object, e.g., assign to it or destroy it.

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,203 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+            for (auto i = 0; i < outputs.size(); ++i) {
+              self.Output(i, outputs[i]);
+            }
+            // Add type constraints
+            for (auto& type_constraint : type_constraints) {
+              std::string type_str;
+              std::vector<std::string> constraints;
+              std::string description;
+              tie(type_str, constraints, description) = type_constraint;
+              self.TypeConstraint(type_str, constraints, description);
```
edgchen1, line 283: ah, too bad we can't use C++17 yet. but this is a way to avoid copying here.
```suggestion
              tie(type_str, constraints, description) = std::move(type_constraint);
              self.TypeConstraint(std::move(type_str), std::move(constraints), std::move(description));
```

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,203 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+                      std::vector<OpSchema::FormalParameter> inputs,
+                      std::vector<OpSchema::FormalParameter> outputs,
+                      std::vector<std::tuple<std::string, std::vector<std::string>, std::string>> type_constraints,
+                      std::vector<OpSchema::Attribute> attributes) {
+            auto self = OpSchema();
+
+            self.SetName(std::move(name)).SetDomain(std::move(domain)).SinceVersion(since_version).SetDoc(doc);
+            // Add inputs and outputs
+            for (auto i = 0; i < inputs.size(); ++i) {
+              self.Input(i, inputs[i]);
```
edgchen1, line 272:
```suggestion
              self.Input(i, std::move(inputs[i]));
```

onnx/cpp2py_export.cc
```diff
@@ -115,13 +115,203 @@ PYBIND11_MODULE(onnx_cpp2py_export, onnx_cpp2py_export) {
...
+            }
+            for (auto i = 0; i < outputs.size(); ++i) {
+              self.Output(i, outputs[i]);
```
edgchen1, line 275:
```suggestion
              self.Output(i, std::move(outputs[i]));
```

onnx/defs/schema.cc
```diff
@@ -553,6 +553,14 @@ OpSchema& OpSchema::AllowUncheckedAttributes() {
   return *this;
 }
 
+OpSchema& OpSchema::Input(int n, FormalParameter formal_parameter) {
+  if (inputs_.size() <= static_cast<size_t>(n)) {
+    inputs_.resize(n + 1);
+  }
+  inputs_[n] = formal_parameter;
```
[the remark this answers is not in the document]
justinchuby: Done

onnx/defs/schema.cc
```diff
@@ -604,6 +610,14 @@ OpSchema& OpSchema::Input(
       differentiation_category);
 }
 
+OpSchema& OpSchema::Output(int n, FormalParameter formal_parameter) {
+  if (outputs_.size() <= static_cast<size_t>(n)) {
+    outputs_.resize(n + 1);
+  }
+  outputs_[n] = formal_parameter;
```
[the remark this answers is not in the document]
justinchuby: Done

onnx/cpp2py_export.cc
justinchuby: Done

onnx/cpp2py_export.cc
justinchuby: Done

onnx/cpp2py_export.cc
justinchuby: Done

onnx/cpp2py_export.cc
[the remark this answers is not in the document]
justinchuby: @jcwchen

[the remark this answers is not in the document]
jcwchen: Strictly ONNX should only have breaking changes when major version bumps: https://github.com/onnx/onnx/blob/a220317da13d5d00870f5b4a8a45f3304a6097d2/docs/Versioning.md#semver-files-and-consumers, but here is mainly for consistency of naming convention, two release after sounds reasonable.

jcwchen (approved):
Basically LGTM. Since it's a large important feature, please get approvals from more reviewers before merge. Thanks for the effort!

Note: next items after this PR -- add corresponding API documentation in .md files and on web ([onnx.ai](https://onnx.ai/onnx/api/)).

onnx/cpp2py_export.cc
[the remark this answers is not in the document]
justinchuby: done

onnx/cpp2py_export.cc
justinchuby: Thanks! Resolved

onnx/defs/schema.cc
```diff
@@ -613,21 +627,20 @@ OpSchema& OpSchema::Output(
...
-      param_option,
-      is_homogeneous,
-      min_arity,
-      differentiation_category);
+          std::move(type_str),
+          param_option,
+          is_homogeneous,
+          min_arity,
+          differentiation_category));
   return *this;
```
edgchen1, line 644:
```suggestion
```

edgchen1 (approved):
looks good!

onnx/defs/schema.cc
justinchuby: Thanks for catching

gramalingam (approved):

onnx/defs/schema.h
```diff
@@ -494,6 +494,8 @@ class OpSchema final {
     std::string description;
   };
 
+  OpSchema& Input(int n, FormalParameter formal_parameter);
```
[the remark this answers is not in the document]
justinchuby: Done

PR #5022, opened 2023-03-21
[WIP] Extend OptionalGetElement

gramalingam:
### Description
Enable writing ONNX code equivalent to the following:
```
if OptionalHasElement(Bias):
   Y = X + OptionalGetElement(Bias)
else:
   Y = X
```
which is useful in authoring functions with optional inputs. Enable the idiom to handle both static-optional and dynamic-optional inputs uniformly.


liqunfu (commented):
do we need a test case for OptionalGetElement where input is not provided? I understand that onnxtestrunner will fail for this test case.

liqunfu:
> do we need a test case for OptionalGetElement where input is not provided? I understand that onnxtestrunner will fail for this test case.

Probably the test case serves nothing.

liqunfu (approved):

PR #5025, opened 2023-03-21
add missing copyright to various files

andife:
### Description

add missing copyright to various files

used command:
reuse annotate --no-replace --copyright="Copyright (c) ONNX Project Contributors" 

#### Motivation and Context
Purpose: Make onnx https://reuse.software/ compliant

### DIFF
--- a/onnx/reference/ops/__init__.py
+++ b/onnx/reference/ops/__init__.py
@@ -1,4 +1,6 @@
 # Copyright (c) ONNX Project Contributors
+
+# Copyright (c) ONNX Project Contributors
 #
 # SPDX-License-Identifier: Apache-2.0
 
--- a/onnx/backend/test/case/model/__init__.py
+++ b/onnx/backend/test/case/model/__init__.py
@@ -1,3 +1,5 @@
+# Copyright (c) ONNX Project Contributors
+
 # SPDX-License-Identifier: Apache-2.0
 
 import sys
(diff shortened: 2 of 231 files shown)


gramalingam (approved):

jcwchen (approved):

justinchuby:
Do we need to create a lintrunner lint to ensure new files have this? Something similar to https://github.com/onnx/onnx/blob/main/.lintrunner.toml#L164-L184

onnx/reference/ops/__init__.py
```diff
@@ -1,3 +1,5 @@
+# Copyright (c) ONNX Project Contributors
+
```
justinchuby, line 2: Redundant?

andife: yes you are right. I will change that.

andife:
> Do we need to create a lintrunner lint to ensure new files have this? Something similar to https://github.com/onnx/onnx/blob/main/.lintrunner.toml#L164-L184

Yes something like that was my intention, but first I would like to have the repo adapted or cleared if necessary. Then  would like to use a tool that was developed for that purpose. 

Then there is the question whether fossology in a CI (I think that is still too complicated from my point of view), so I would like to use the Github plugin from https://github.com/marketplace/actions/reuse-compliance-check (but I would discuss that first).

PR #5026, opened 2023-03-22
fix light model input in backend test

yuanyao-nv:
### Description

In backend test for light models, the input list to backend should not data paths - change it to actual data arrays.

### Motivation and Context

PR that introduced this issue: https://github.com/onnx/onnx/pull/4861
This should fix the current scoreboard pipeline failure caused by incorrect input type.


jcwchen:
I guess it has been covered by https://github.com/onnx/onnx/pull/4960. Shall we close this now? Thanks.

yuanyao-nv:
> I guess it has been covered by #4960. Shall we close this now? Thanks.

yes let me close it.

PR #5031, opened 2023-03-22
add copyright and licence information

andife:
### Description
next files with added copyright and licences


docs/CONTRIBUTING.md
```diff
@@ -1,3 +1,7 @@
+<!--
+Copyright (c) ONNX Project Contributors
+-->
+
```
jcwchen, line 4: Duplicate `<!--Copyright (c) ONNX Project Contributors-->`?

docs/docsgen/source/requirements.txt
```diff
@@ -1,3 +1,6 @@
+# Copyright (c) ONNX Project Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
```
jcwchen, line 4: Probably fine that not to add license in requirements.txt?

docs/CONTRIBUTING.md
andife: Thank you! 

Yes, I removed it ["reuse lint" did not realize the information was there, only combined with copyright and license]. I will check that, how to handle that, if we want to check such cases automatically by a tool.

docs/docsgen/source/requirements.txt
andife: In general, every file should have these two information, but I can also add that to the .reuse/dep5 file. 

I'm not quite clear yet what best practices would be desired or fit for us?

What exactly don't you like about the addition? 

For other points I'm also not sure yet how to deal with it, e.g. if documentation is automatically generated from it. 

I'll remove that change from the commit.

jcwchen (approved):

docs/docsgen/source/requirements.txt
andife: I think I would add it everywhere where the license is clear, so that it is from us..., and does not collide with any functionality. For non-trivial files like __init__.py you could of course also choose cc0? Generated files where the "reuse checker" does not work. I would also exclude and list in the dep5 file. I would have given the header to the whole pipelines... But of course you can leave it aside in the first round and discuss it later.  Why would you exclude CI related files? You mean excluding from putting a header / comment to the top? or mark with the licence and copyright in dep5?

jcwchen: > Why would you exclude CI related files? You mean excluding from putting a header / comment to the top? or mark with the licence and copyright in dep5?

I just feel strictly CI related files are not part of source code. At least exclude them from putting a header / comment to the top, I think. I am fine with having them in dep5.

andife: ok, we can start this way. I added some directories to the dep5 file.

.reuse/dep5
```diff
@@ -9,6 +9,18 @@ Source: onnx.ai
...
+Files: .azure-pipelines/*.yml
+COPYRIGHT: Copyright (c) ONNX Project Contributors
+License: Apache-2.0
+
+
+Files: .github/**/*.md
+COPYRIGHT: Copyright (c) ONNX Project Contributors
+License: Apache-2.0
+
+Files: .github/**/*.yml
```
jcwchen, line 21: BTW, some of CI related files have .yaml as extension instead... I will have a PR to update them for consistency.

yuanyao-nv (approved):

Issue #4459, opened 2022-08-20
Add .clang-tidy check in CI pipelines


jcwchen:
# Feature Request

### System information

Latest main branch

### What is the problem that this feature solves?

Follow-up work for https://github.com/onnx/onnx/pull/4391.

### Describe the feature

ONNX just introduced .clang-tidy file for developers. It would be better if ONNX can also have a CI running it to make sure the codebase follows it.

### Will this influence the current api (Y/N)?

N

### Feature Area

CI pipelines

PR #5032, opened 2023-03-23
Minor fix from clang-tidy

justinchuby:
### Description

Minor fix from clang-tidy
1. Make `type_str` not const because `clang-tidy` says this prevents it from being moved.
2. Change `GetErrorWithNodeInfo` to take a const reference of `std::runtime_error`

### Motivation and Context

Minor adjustment to tidy code


jcwchen (approved):
Thank you for catching these. Just wondering are these the only ones that clang-tidy disallows in the whole codebase?

https://github.com/onnx/onnx/issues/4459 It would be great if we can have a CI running it to catch it early.

onnx/shape_inference/implementation.cc
```diff
@@ -1007,7 +1007,7 @@ std::vector<const TypeProto*> GraphInferencerImpl::doInferencing(
   return graph_output_types;
 }
 
-std::string GetErrorWithNodeInfo(const NodeProto& n, std::runtime_error err) {
+std::string GetErrorWithNodeInfo(const NodeProto& n, const std::runtime_error& err) {
```
jcwchen, line 1010: onnx/shape_inference/implementation.h needs to be updated as well

jcwchen (commented):
Thanks for the update. I think we can merge this PR: https://github.com/onnx/onnx/pull/5041 first and see whether the introduced clang-tidy can catch the warnings in this PR?

jcwchen (approved):

jcwchen:
https://github.com/onnx/onnx/pull/5041 The new added clang-tidy does not work normally... Investigating.

gramalingam (approved):

jcwchen:
I cannot solve the clang-tidy CI issue shortly and thus I will merge this PR first. Thanks for waiting!

PR #5053, opened 2023-03-28
Enable address sanitizer in CI

justinchuby:


justinchuby:
==3231==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.

jcwchen (commented):
> @jcwchen I remember you were able to link asan successfully? I am having a hard time doing that

I have tried ASAN for a while, but never succeeded... That's why I only enabled UBSan in my PR to faster unblock old issue.

andife:
@cyyever @justinchuby should we give that a new try with our current pipeline, or could it closed?

cyyever:
I want to try it

PR #5061, opened 2023-03-28
Clarify spec for Round, Ceil, Floor

yuanyao-nv:
### Description
Add this sentence to the spec for Round, Ceil, and Floor to be more precise about the behavior of these ops on infinity, NaN, +/- 0.0:
```
If input x is integral, +0, -0, NaN,  or infinite, x itself is returned.
```

### Motivation and Context
This is also how these ops are defined on the Linux man page, see https://linux.die.net/man/3/round, https://linux.die.net/man/3/ceil, https://linux.die.net/man/3/floor.

Numpy doesn't explicitly say this in their doc, but their behavior conforms with this.


gramalingam (commented):
Thanks!

Is this the same behavior at the hardware level (for whatever one you do know the answer for!) or, say IEEE floating-point spec, etc.? Just wondering, in case you know.

ArchRobison:
Yes, IEEE 754-2019 codifies the behavior. Section "5.9 Details of operations to round a floating-point datum to integral value". It includes an operation:
> roundToIntegralTiesToEven(x) rounds x to the nearest integral value, with halfway cases rounding to even.

and has a sentence covering all the float-to-float rounding operations:
> These operations convert zero operands to zero results of the same sign, and infinite operands to infinite results of the same sign.

Clause 6.2 covers says that quiet NaNs propagate unless otherwise specified.

gramalingam (approved):

PR #5063, opened 2023-03-28
Improve performance of make_attribute on iterable

jbachurski:
### Description

This PR updates the implementation of `make_attribute` to handle the case of iterables much more efficiently, by re-implementing how the `str | bytes` case gets handled. 

### Motivation and Context

Previously, `onnx.make_attribute` was really slow and had a major overhead in comparison to the pure protobuf interface. This PR addresses this by improving the performance on long integer lists twice-fold (as run on my somewhat old laptop, Intel i7-4700MQ):

```sh
$ python3 -m timeit -s "import onnx" -- "onnx.helper.make_attribute('name', list(range(10**6)))"
1 loop, best of 5: 1.64 sec per loop
$ python3 -m timeit -s "import onnx" -- "onnx.helper.make_attribute('name', list(range(10**6)))"
1 loop, best of 5: 732 msec per loop
```

The previous implementation always (even in the case the input was just integers) used an expensive `try..except` check to determine whether a value was `str | bytes`, in the `_to_bytes_or_false` function.


xadupre (approved):

jbachurski:
@xadupre @jcwchen Are we good to merge this now? Thank you!
There seem to be problems with CI freezing.

jcwchen (approved):

PR #5060, opened 2023-03-28
Parsing NodeProto and handling optional inputs of functions

gramalingam:
### Description
A few extensions/fixes:
(a) Expose node-level parsing in python, as it helps write node-level tests in python
(b) Fix an issue in function type/shape inference to support optional inputs.
(c) Add support for parsing/printing type-attributes

PR #5066, opened 2023-03-29
Parsing NodeProto and handling optional inputs of functions

gramalingam:
### Description

(Redo of [PR 5060](https://github.com/onnx/onnx/pull/5060))

A few extensions/fixes:
(a) Expose node-level parsing in python, as it helps write node-level tests in python
(b) Fix an issue in function type/shape inference to support optional inputs.
(c) Add support for parsing/printing type-attributes
(d) Add helper method to get attribute-value from a node


onnx/helper.py
```diff
@@ -658,6 +658,15 @@ def get_attribute_value(attr: AttributeProto) -> Any:
     raise ValueError(f"Unsupported ONNX attribute: {attr}")
 
 
+def get_node_attr_value(node: NodeProto, attr_name: str) -> Any:
+    matching = [x for x in node.attribute if x.name == attr_name]
+    if len(matching) > 1:
+        raise ValueError(f"Node has multiple attributes with name {attr_name}")
+    if len(matching) < 1:
```
jcwchen, line 665:
```suggestion
    if not matching:
```
tiny nit

jcwchen (approved):
LGTM. Thanks for extending parser!

onnx/helper.py
gramalingam: If this was the only condition, I would agree. But when we handle the 3 conditions (len > 1, = 1, < 1), it is better to be explicit. So, the symmetry with line 663 makes this better, IMHO.

onnx/shape_inference/implementation.cc
```diff
@@ -569,18 +569,24 @@ class ShapeInferenceImplBase {
...
+      auto& parameter_name = func_proto.input().Get(i);
+      auto* type_ptr = (i < num_actual_inputs) ? ctx.getInputType(i) : nullptr;
+      // nullptr is valid, and indicates a missing optional input
+      if (type_ptr != nullptr) {
+        // Use a temporary copy of original type.
+        // TODO: investigate whether we can eliminate use of temporary copy
+        types_cache[i] = *type_ptr;
+        value_types_by_name[parameter_name] = &types_cache[i];
+      } else
+        value_types_by_name[parameter_name] = nullptr;
```
jcwchen, line 585: https://github.com/onnx/onnx/issues/5212 onnx.shape_inference with data_prop might crash when there's a nested function, because `fail_type_inference("Input ", i, " type is missing.");` was changed to `value_types_by_name[parameter_name] = nullptr;` in this PR.

@gramalingam May I understand what the intention to store nullptr is? It seems dangerous when other wants to access value_types_by_name[name] from other code?

gramalingam: This is because we are generalizing type-inference for a function to handle functions with optional inputs. There will be no types present for a missing optional input. If it is crashing somewhere, that line needs to be updated to check if the type is present/absent and do something appropriate ... I may have overlooked some line of code that needs to be updated. Can you point me to the line that is crashing? Thanks!

BowenBao: This was retrieved from #5212
```
#0  0x00007ffecc66e86c in onnx::TypeProto::value_case (this=0x0) at /home/bowbao/onnx/.setuptools-cmake-build/onnx/onnx-ml.pb.h:11615
#1  0x00007ffecc71b18e in onnx::TypeProto::_internal_has_tensor_type (this=0x0) at /home/bowbao/onnx/.setuptools-cmake-build/onnx/onnx-ml.pb.h:11097
#2  0x00007ffecc71b1b2 in onnx::TypeProto::_internal_tensor_type (this=0x0) at /home/bowbao/onnx/.setuptools-cmake-build/onnx/onnx-ml.pb.h:11128
#3  0x00007ffecc71b1e6 in onnx::TypeProto::tensor_type (this=0x0) at /home/bowbao/onnx/.setuptools-cmake-build/onnx/onnx-ml.pb.h:11134
#4  0x00007ffecc8ce5b3 in onnx::<lambda(onnx::DataPropagationContext&)>::operator()(onnx::DataPropagationContext &) const (__closure=0x7fffffff7020, ctx=...)
    at /home/bowbao/onnx/onnx/defs/tensor/old.cc:5120
#5  0x00007ffecc8d6f3f in std::_Function_handler<void(onnx::DataPropagationContext&), onnx::GetOpSchema() [with T = onnx::Shape_Onnx_ver15]::<lambda(onnx::DataPropagationContext&)> >::_M_invoke(const std::_Any_data &, onnx::DataPropagationContext &) (__functor=..., __args#0=...) at /usr/include/c++/9/bits/std_function.h:300
#6  0x00007ffecc910695 in std::function<void (onnx::DataPropagationContext&)>::operator()(onnx::DataPropagationContext&) const (this=0x7fffffff7020, __args#0=...)
    at /usr/include/c++/9/bits/std_function.h:688
#7  0x00007ffecc90c22b in onnx::shape_inference::ShapeInferenceImplBase::process (this=0x7fffffff7300, n=...) at /home/bowbao/onnx/onnx/shape_inference/implementation.cc:447
```

BowenBao: https://github.com/onnx/onnx/blob/da892d3a061b2dea5dbb2fff74612f3af157236a/onnx/defs/tensor/old.cc#L5163-L5164

gramalingam: Ok, I suspect it has nothing to do with this PR. It must be due to a pending item: the checker and shape-inference wasn't updated to handle attribute-parameters with default-values in functions (which was added recently too). That's my guess at this point.

PR #5069, opened 2023-03-30
Faster reference implementation for operator Conv based on im2col

xadupre:
### Description
Current implementation of Operator Conv is very slow. This is a faster one based on im2col.

Measures with shape(X)=(1, 1, 64, 64), shape(kernel)=(3, 3), before time is 0.0033s, after time is 0.00075, > 4 times faster. This ratio is higher with bigger matrices.

### Motivation and Context
Performance.


onnx/reference/ops/op_conv.py
```diff
@@ -3,98 +3,48 @@
...
+        i = np.tile(kind.ravel(), n_C).reshape(-1, 1) + iind.reshape(1, -1)
+        indices.append(i)
+
+    d = np.repeat(np.arange(n_C), kernel_size).reshape(-1, 1)
+
+    nc = [(0, 0)] * 2
+    padding = [(pads[i], pads[i + n_dims]) for i in range(n_dims)]
+    X_padded = np.pad(X, tuple(nc) + tuple(padding), mode="constant")
+
+    getitem = (slice(0, m), d, *tuple(indices))
```
justinchuby, line 44:
```suggestion
    getitem = (slice(0, m), d, *indices)
```

nit

gramalingam:
A general idea (applies to both `conv` and `im2col`). It makes sense to me to support multiple implementations of an op (specifically, here, one may be slower/naive but easier to verify correct or read and understand, while another may be more efficient).

Shall we adopt some naming convention for the methods in the classes that implement the ops to support this? 

So, for example, your naive implementation of im2col can go back into the main class (instead of being in the testing code). We just use an appropriate suffix to distinguish the two implementations. Eg., may be add `_optimized` for the faster on. Or, add `_spec` to the naive implementation.

pyproject.toml
```diff
@@ -116,4 +116,5 @@ ban-relative-imports = "all"
 "onnx/reference/ops/aionnx_preview_training/__init__.py" = ["F401"]
 "onnx/reference/ops/aionnx_preview_training/_op_list.py" = ["F401"]
 "onnx/reference/ops/experimental/__init__.py" = ["F401"]
+"onnx/test/reference_evaluator_test.py": ["C408"]
```
justinchuby: I would add a comment on why we need to disable this. I think C408 can potentially be disabled all together if we see the use of dict() for creating kwargs.

onnx/reference/ops_optimized/__init__.py
```diff
@@ -7,4 +7,4 @@
 from onnx.reference.ops_optimized.op_conv_optimized import Conv
 
 
-__all__ = [conv]
+__all__ = [Conv]
```
justinchuby, line 10: It needs to be a string list

justinchuby: [“Conv”]

onnx/reference/reference_evaluator.py
```diff
@@ -11,6 +11,7 @@
 from onnx.defs import onnx_opset_version
 from onnx.onnx_pb import FunctionProto, GraphProto, ModelProto, NodeProto, TypeProto
 from onnx.reference.op_run import OpRun, RuntimeContextError
+from onnx.reference.ops_optimized import Conv as ConvOptimized
```
gramalingam, line 14: Nit: instead of hard-coding the list of optimized ops here, may be better to import a list of optimized ops from `onnx.reference.ops_optimized` ? So, users need to update only one place when adding a new op there?

onnx/reference/ops/experimental/op_im2col.py
```diff
@@ -4,7 +4,7 @@
 # pylint: disable=R0913,R0914,W0221
 
 from onnx.reference.ops.experimental._op_run_experimental import OpRunExperimental
-from onnx.reference.ops.op_conv import im2col
+from onnx.reference.ops_optimized.op_conv_optimized import im2col_fast
```
gramalingam, line 7: I am a bit confused: (i) Why is this under "experimental"? (ii) Why is this called the optimized version? Shouldn't it be in ops_optimized if it is an optimized version?

onnx/reference/reference_evaluator.py
xadupre: done

gramalingam (approved):

PR #5071, opened 2023-03-30
[WIP] add reuse github action

andife:


.github/pull_request_template.md
```diff
@@ -1,3 +1,4 @@
+
```
yuanyao-nv, line 1: Is this empty line needed?

yuanyao-nv (approved):

.github/pull_request_template.md
andife: No it isn't. Well seen. At first I wanted to add the headers there too. But then I wasn't sure if it would be sent through Github in case of doubt. Therefore I have dropped it for the time being. And apparently my change is not completely undone.

jcwchen (approved):

PR #5078, opened 2023-04-01
add local function test cases

liqunfu:
### Description
there are no test cases for runtime to test model with local functions

### Motivation and Context
add test cases for runtime to test model with local functions


onnx/backend/test/case/model/selu_function.py
```diff
@@ -0,0 +1,138 @@
...
+        # save model and test data for runtime test
+        sess = reference.ReferenceEvaluator(model)
+        x = np.random.randn(3).astype(np.float32)
+        input_dict = {"x": x}
+        y = sess.run(None, input_dict)
+
+        expect(model, inputs=[x], outputs=[y[0]], name="test_selu_function")
+
+    @staticmethod
+    def export() -> None:
```
xadupre, line 66: duplicated method naùe

xadupre:
many test case not related to Selu were updated as well. They should not be part of this PR.

PR #5090, opened 2023-04-05
security: add commit hash to docker file

andife:
### Description
best practice (security) to add commit hash


justinchuby:
Will this be auto updated by https://github.com/onnx/onnx/blob/main/.github/dependabot.yml ?

andife:
> Will this be auto updated by https://github.com/onnx/onnx/blob/main/.github/dependabot.yml ?

not yet. I have to add an entry for docker there. Will update that.

andife:
> > Will this be auto updated by https://github.com/onnx/onnx/blob/main/.github/dependabot.yml ?
> 
> not yet. I have to add an entry for docker there. Will update that.

added/updated the dependabot rules for docker

.github/workflows/release_linux_x86_64.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         python -m pip install -q -r requirements-release.txt
 
     - name: Build manylinux2014_x86_64
-      uses: docker://quay.io/pypa/manylinux2014_x86_64
+      uses: docker://quay.io/pypa/manylinux2014_x86_64:latest@sha256:085b7c2c1bcb45b936e5478fe24a5142b0519e242a0104142fd50d569d6cb922
```
jcwchen, line 46: I am inclined to always use the latest docker image. Will [dependabot](https://github.com/apps/dependabot) also help us to bump the version if there is an update?

jcwchen (commented):
https://github.com/onnx/onnx/blob/89a4eb071185244882e1bbd167ec384e9b8a07a4/.github/workflows/release_linux_aarch64.yml#L27 This line will need to be updated also?

.github/workflows/release_linux_x86_64.yml
andife: I added a rule for docker at .github/dependabot.yml ... so it should work.

jcwchen:
Shall we also update:
[[onnx/.github/workflows/release_linux_aarch64.yml](https://github.com/onnx/onnx/blob/89a4eb071185244882e1bbd167ec384e9b8a07a4/.github/workflows/release_linux_aarch64.yml#L27)](https://github.com/onnx/onnx/blob/89a4eb071185244882e1bbd167ec384e9b8a07a4/.github/workflows/release_linux_aarch64.yml#L27)
with specified commit hash?

.github/workflows/release_linux_aarch64.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     env:
       # setting up python and docker image
       py: /opt/python/${{ matrix.python-version }}/bin/python
-      img: quay.io/pypa/manylinux2014_aarch64
+      img: docker://quay.io/pypa/manylinux2014_aarch64@a2d8ba969c23fb91e383b99f62324769944fe2518404b56b186037bab7ad1117
```
jcwchen, line 27:
```suggestion
      img: docker://quay.io/pypa/manylinux2014_aarch64:latest@sha256:a2d8ba969c23fb91e383b99f62324769944fe2518404b56b186037bab7ad1117
```

andife: maybe we ignore these Docker files in environments for now. Somehow the syntax doesn't seem to fit quite right. 
For the score at scorecard they don't seem to be detected, and then dependabot should also find them as stale. If one has the correct syntax, I would perhaps test this with the second last hash if we then get an automatic update suggestion.

jcwchen (approved):

PR #5091, opened 2023-04-05
Fix `compose.add_prefix` on unused inputs

jbachurski:
### Description

`onnx.compose.add_prefix` did not update the names of unused inputs, even though they are still part of the  used names. 

### Motivation and Context

The bug was uncovered in `spox.inline`, as unused graph inputs were not getting renamed:

https://github.com/Quantco/spox/blob/0ca5281cfd3e5dc4fabc7b815cac2dff877128d0/src/spox/_inline.py#L90-L105

This required an ugly workaround by adding Identity nodes.

The bug was caused by a lazy branch, assuming that when `rename_edges`, all input/outputs names will be found. As prefixing is deterministic (as implemented by the `_prefixed` helper), this fix does not affect anything, and runtime effects are negligible.


justinchuby:
Possibly a test that can ensure the behavior?

jbachurski:
I added an unused input to the model there, but unfortunately that whole test is not really useful. Someone copy-pasted the implementation into the test, instead of writing down the expected answer. I don't really want to go through all the cases right now to fix that, but it should probably be done.

justinchuby:
Copying the logic into the test may not be the right thing to do. Potentially add a FIXME in test can create a tracking issue?

jbachurski:
See #5094 for fixing the tests. 

If anyone will be working on this, I was also considering that it would be useful if `compose` essentially had `add_prefix` and `add_prefix_graph` split into a utility like `apply_rename`, which would apply an arbitrary renaming function, and not just prefixing. This is basically already done, as first a `name_map` is constructed, and then it is applied. There could be another utility like `collect_names` for constructing the `name_map`.

PR #5097, opened 2023-04-05
Fix partially failed during doc generation in Linux CI

jcwchen:
### Description

Fix partially failed during doc generation in Linux CI:
- ~Use -Q to suppress warnings as errors.~ Use `--keep-going ` to keep going when getting warnings without failure
- Remove make -j auto, because it will lead to:
```
Sphinx error:
Builder name auto not registered or available through entry point
make: *** [Makefile:24: auto] Error 2
make: *** Waiting for unfinished jobs....

```
Without -j auto, the speed seems fine.

### Motivation and Context

Currently Linux-CI always shows partially succeed: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=42534&view=results after test generation.


docs/docsgen/Makefile
```diff
@@ -14,11 +14,11 @@ BUILDDIR      = build
 
 # Put it first so that "make" without argument is like "make help".
 help:
-	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O) -Q
```
xadupre, line 17: Option -Q disables every output. Did you try other options to disable the warnings:

```
console output options:
  -v                increase verbosity (can be repeated)
  -q                no output on stdout, just warnings on stderr
  -Q                no output at all, not even warnings
  -w FILE           write warnings (and errors) to given file
  -W                turn warnings into errors
  --keep-going      with -W, keep going when getting warnings
```

jcwchen: --keep-going seems more appropriate and it works. Thanks for the suggestion!

xadupre (approved):

PR #5103, opened 2023-04-07
Remove `stderr` prints in `version_converter`

jbachurski:
### Description

Removes `std::cerr` prints directly into STDERR whenever `version_converter` encounters ops outside `ai.onnx`. 
For now, the prints are kept behind an `if (DEBUG)` similar to some other ones in the same function. This is to make this change as smooth as possible. They can be removed entirely in the future, as the function could be refactored anyway (what are "ConstantFill", "Undefined", "Captured" ops? I've never seen them)

### Motivation and Context

Though `version_converter` might usually be used in an interactive context, Spox uses it often programmatically. The current prints pollute the standard output log and the are confusing as they are not properly labelled with traceback. 

I would appreciate merging this before the code freeze a lot!

### DIFF
--- a/onnx/version_converter/convert.cc
+++ b/onnx/version_converter/convert.cc
@@ -73,13 +73,15 @@ void DefaultVersionConverter::convert_graph(
       debug(std::string("Finding schema for ") + std::string(cur_op->kind().toString()));
       const std::string op_name = cur_op->kind().toString();
       if (op_name == "ConstantFill") {
-        std::cerr
-            << "Warning: skipping schema search for experimental op 'ConstantFill' and keeping the op as is. "
-               "Please be advised the converted model may not be working properly if target runtime does not support this "
-               "experimental op."
-            << std::endl;
+        if (DEBUG)
+          std::cerr
+              << "Warning: skipping schema search for experimental op 'ConstantFill' and keeping the op as is. "
+                 "Please be advised the converted model may not be working properly if target runtime does not support this "
+                 "experimental op."
+              << std::endl;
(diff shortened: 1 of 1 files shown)


onnx/version_converter/convert.cc
```diff
@@ -73,13 +73,15 @@ void DefaultVersionConverter::convert_graph(
       debug(std::string("Finding schema for ") + std::string(cur_op->kind().toString()));
       const std::string op_name = cur_op->kind().toString();
       if (op_name == "ConstantFill") {
-        std::cerr
-            << "Warning: skipping schema search for experimental op 'ConstantFill' and keeping the op as is. "
-               "Please be advised the converted model may not be working properly if target runtime does not support this "
-               "experimental op."
-            << std::endl;
+        if (DEBUG)
```
jcwchen, line 76: tiny nit: perhaps put cerr into brackets? It's up to you, but it will look clearer to me.

jcwchen (approved):
Thanks you for catching this. It makes sense to me. Definitely we should prevent printing cerr for warnings. An alternative would be just printing them into cout, but the log might be lengthy in some cases. Therefore, I am fine with only enabling them when DEBUG is specified.

onnx/version_converter/convert.cc
jbachurski: Agreed, it's better to have braces here. The linter didn't remind me.

justinchuby:
https://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html#capturing-standard-output-from-ostream Would this help?

PR #5105, opened 2023-04-09
Utility to inline model-local functions

gramalingam:
### Description

An initial version of utility to inline all calls to model-local functions.

Extensions to be handled in future:
* Ideally, we should allow users to specify filters to indicate which functions to inline and/or which not to inline.
* Doesn't yet handle opschema-defined functions.


jbachurski:
I would suggest implementing a utility to inline a given list of FunctionProtos, which is more general and probably already indirectly implemented here. It'll be more general and finding all model local functions is trivial.

gramalingam:
> I would propose to implement a utility to inline a given list of FunctionProtos, which is more general and already indirectly implemented here. It'll have more uses and finding all model local functions is trivial.

I think that is equivalent? To be more specific, the implementation provides a core (namely `inline_functions`) that is generic and intended to enable other scenarios as well, and one specific scenario (namely `inline_local_functions`). But only the latter is currently exposed as an external API in the include file. The reason is that I think the core API can be made even more general, and so it may be good to iterate through other scenarios as well before finalizing the signature of this core API.

For example, I think it would be useful to allow the callers to specify which functions to inline or NOT inline (via the names of functions).

Anyway: I agree that other use-cases will arise and we should support them. This is just the first step.

jbachurski:
Thanks for responding! I think we are on the same page then. I noticed it ends up implemented in C++ but I had only looked for a Python test/bindings. 
In any case, it would be nice to have an ergonomic API for this graph transformation.

onnx/inliner/inliner.cc
```diff
@@ -0,0 +1,325 @@
...
+// Function lookup function. Returns true iff lookup is successful, in which case
+// the found FunctionProto is copied into *return_value. The opset version is not
+// a parameter since it is determined by the domain for a given model.
+using FunctionResolver =
+    std::function<bool(const std::string& domain, const std::string& op, FunctionProto* return_value)>;
+
+// We use a string of the form "domain::name" as the id for a function.
+using FunctionId = std::string;
+
+FunctionId GetFunctionId(const FunctionProto& function) {
```
justinchuby, line 29: Do we want to move the functions etc into an unnamed namespace?

justinchuby: I was reading https://google.github.io/googletest/advanced.html#testing-private-code and https://google.github.io/styleguide/cppguide.html#Internal_Linkage

gramalingam: Done, thanks!

onnx/inliner/inliner.cc
```diff
@@ -0,0 +1,331 @@
...
+      for (auto& graph : *attr.mutable_graphs())
+        Transform(graph);
+      ++attr_iter;
+    }
+  }
+
+  // Process a sub-graph, contained as an attribute in a control-flow op node.
+  void Transform(GraphProto& graph) {
+    rename_scopes.emplace_back();
+    for (auto& x : *graph.mutable_input())
```
xadupre, line 228: I would add a comment on how the new name is built to help developper debug.

onnx/test/cpp/inliner_test.cc
```diff
@@ -0,0 +1,164 @@
...
+static void InlineFunctions(ModelProto& model, const char* input) {
+  OnnxParser parser(input);
+  auto status = parser.Parse(model);
+  EXPECT_TRUE(status.IsOK()) << status.ErrorMessage();
+  EXPECT_TRUE(parser.EndOfInput()) << "Extra unparsed input unexpected.";
+
+  checker::check_model(model);
+  shape_inference::InferShapes(model);
+
+  std::cout << ProtoToString(model) << "\n";
```
xadupre, line 29: Is it expected in a test to print something?

onnx/test/cpp/inliner_test.cc
```diff
@@ -0,0 +1,164 @@
...
+  // The second node's call, to bar, must be inlined.
+  auto& second_node = model.graph().node(1);
+  // Check that it is a call to Add
+  ASSERT_EQ(second_node.op_type(), "Add");
+
+  // The non-inlined foo must still be in the function list.
+  ASSERT_EQ(model.functions_size(), 1);
+  ASSERT_EQ(model.functions(0).name(), "foo");
+}
+
```
xadupre, line 162: I would add a test where the same function is added twice to check there is no duplicated name. And I would also add a test to check the inlining works well when a local function calls another local function (recursivity).

onnx/test/cpp/inliner_test.cc
gramalingam: No, that was just for my own local validation. Will remove or comment it out, thanks for catching it.

onnx/test/cpp/inliner_test.cc
gramalingam: There is already a test for one local function calling another local function.

I will add a test for same function being called twice.

gramalingam: Done

onnx/inliner/inliner.cc
gramalingam: Done. See comments added to function `MakeUnique`.

xadupre (approved):

justinchuby:
I tested the inliner with this model, but it seemed to have returned the same model:

gramalingam:
> I tested the inliner with this model, but it seemed to have returned the same model:
> 
> [gpt2-dataprop.zip](https://github.com/onnx/onnx/files/11392758/gpt2-dataprop.zip)

I tried it. Looks like the model uses onnx opset 17, while the functions use opset 18. Currently, inliner does not inline functions when there is an opset mismatch. Changed model opset to 18, and I think it did something: looks like it inlined 24 of 26 functions, the number of nodes in main graph went up from 383 to 595. Didn't check anything more.

gramalingam:
> > I tested the inliner with this model, but it seemed to have returned the same model:
> > [gpt2-dataprop.zip](https://github.com/onnx/onnx/files/11392758/gpt2-dataprop.zip)
> 
> I tried it. Looks like the model uses onnx opset 17, while the functions use opset 18. Currently, inliner does not inline functions when there is an opset mismatch. Changed model opset to 18, and I think it did something: looks like it inlined 24 of 26 functions, the number of nodes in main graph went up from 383 to 595. Didn't check anything more.

Integrating the version-converter to automatically convert functions to model's opset is work in progress. (Even that may not always succeed for for version down-conversion; version up-conversion should be more complete.)

justinchuby:
> I tried it. Looks like the model uses onnx opset 17

Ah that's right. Didn't realize that. Thanks!

PR #5112, opened 2023-04-10
Support textproto as a serialization format

justinchuby:
### Description

Support textproto as a serialization format by using the format option. Users can use

```python
model = onnx.load_model("model.onnx", format="textproto")
```

to load a textproto model.

This PR also

- Creates a generic Serializer class to handle serialization.
- Makes `load_model_from_string` accept a `str` input.

### Motivation and Context

#4995


onnx/__init__.py
```diff
@@ -121,7 +135,7 @@ def _serialize(proto: Union[bytes, google.protobuf.message.Message]) -> bytes:
 _Proto = TypeVar("_Proto", bound=google.protobuf.message.Message)
 
 
-def _deserialize(s: bytes, proto: _Proto) -> _Proto:
+def _deserialize(s: bytes, proto: _Proto, format_: _SupportedFormat) -> _Proto:
```
xadupre, line 138: I would keep `Union[Literal["protobuf"], Literal["textproto"]]`. It is easier to understand for users or it should be documented and not added as a private attribute of the module.

onnx/__init__.py
```diff
@@ -140,24 +154,28 @@ def _deserialize(s: bytes, proto: _Proto) -> _Proto:
...
 
-    decoded = typing.cast(Optional[int], proto.ParseFromString(s))
-    if decoded is not None and decoded != len(s):
-        raise google.protobuf.message.DecodeError(
-            f"Protobuf decoding consumed too few bytes: {decoded} out of {len(s)}"
-        )
+    if format_ == "textproto":
+        google.protobuf.text_format.Parse(s.decode("utf-8"), proto)
+    else:
+        assert format_ == "protobuf"
```
xadupre, line 160: I would raise an exception with a better error message. It is likely to be seen by users when they call save_model.

onnx/__init__.py
```diff
@@ -91,17 +98,24 @@ def _get_file_path(f: Union[IO[bytes], str]) -> Optional[str]:
...
 
     Returns:
         Serialized proto in bytes.
     """
     if isinstance(proto, bytes):
         return proto
+    if format_ == "textformat":
+        return google.protobuf.text_format.MessageToString(proto).encode("utf-8")
+
+    assert format_ == "protobuf"
```
xadupre, line 118: See comment for `_deserialize`.

onnx/__init__.py
```diff
@@ -91,17 +98,24 @@ def _get_file_path(f: Union[IO[bytes], str]) -> Optional[str]:
...
         proto: An in-memory proto, such as a ModelProto, TensorProto, etc
+        format: The serialization format. Can be one of :py:const:`_SupportedFormat`.
 
     Returns:
         Serialized proto in bytes.
     """
     if isinstance(proto, bytes):
         return proto
+    if format_ == "textformat":
+        return google.protobuf.text_format.MessageToString(proto).encode("utf-8")
```
xadupre, line 116: I wonder if the encoding should be exposed as a parameter of the function.

onnx/__init__.py
```diff
@@ -180,59 +197,56 @@ def load_model(
...
     s: bytes,
-    format: Optional[Any] = None,  # pylint: disable=redefined-builtin
+    format: _SupportedFormat = "protobuf",  # pylint: disable=redefined-builtin
 ) -> TensorProto:
     """Loads a binary string (bytes) that contains serialized TensorProto.
 
     Args:
         s: a string, which contains serialized TensorProto
-        format: for future use
+        format: The serialization format. Can be one of :py:const:`_SupportedFormat`. Default is "protobuf".
```
xadupre, line 238: Any attribute started with `_` is usually left out of the documentation. It should be renamed.

onnx/test/basic_test.py
```diff
@@ -36,57 +36,57 @@ def _simple_tensor(self) -> TensorProto:
         )
         return tensor
 
-    def test_save_and_load_model(self) -> None:
+    def test_save_and_load_model_when_input_is_string(self) -> None:
         proto = self._simple_model()
-        cls = ModelProto
-        proto_string = onnx._serialize(proto)
+        proto_string = onnx._serialize(proto, "protobuf")
```
xadupre, line 41: No test with textproto? Maybe I missed them.

onnx/__init__.py
justinchuby: I think we should see it expanded in the sphinx docs? I tested in vscode and intellisense was very good at telling me what are the allowed values. I put them in a single type alias considering it will be easier when we expand this set?

onnx/__init__.py
justinchuby: Ok!

onnx/__init__.py
justinchuby: I am inclined to suggest we only support one single encoding. I don’t think there is a point of supporting others?

onnx/__init__.py
justinchuby: I think it will be expanded inline? Let me test that.

onnx/test/basic_test.py
justinchuby: Good catch. Let me double check

onnx/__init__.py
justinchuby: Done

onnx/__init__.py
justinchuby: Done

onnx/test/basic_test.py
justinchuby: Created tests

onnx/__init__.py
justinchuby: image

Looks like the type annotation works. I will remove this duplicated info in the docstring.

justinchuby: Done

onnx/__init__.py
justinchuby: Done. PATL

xadupre (approved):

justinchuby:
From onnx discussion: we actually would like to drop encoding until it is needed so we don't carry apis we don't know we need. @gramalingam @abock

justinchuby:
I updated the abc to `ProtoSerializer` to make it very clear that it is only for protobuf representations. Future support for potential in memory IR can then use a more generic name that will not collide with it.

justinchuby:
@jcwchen tested with ort. It fails to load:

```
onnxruntime.capi.onnxruntime_pybind11_state.InvalidProtobuf: [ONNXRuntimeError] : 7 : INVALID_PROTOBUF : Load model from resnet50.textproto failed:Protobuf parsing failed.
```

jcwchen (commented):
Thanks for the experiment and update. If onnxruntime cannot work with textproto for now, it seems better to have another file extension for textproto format. To confirm, do we conclude that we should use ".textproto" in this case?

One more question: have you tried your serialization/deserialization for textproto format with a ONNX model with external data? Ideally it should work normally, but it seems worthy to confirm.

justinchuby:
> Thanks for the experiment and update. If onnxruntime cannot work with textproto for now, it seems better to have another file extension for textproto format. To confirm, do we conclude that we should use ".textproto" in this case?

This change doesn't control the file extension. A user can use anything as they do with the binary proto format (conventionally .onnx), but textproto is conventionally the extension for it.

> One more question: have you tried your serialization/deserialization for textproto format with a ONNX model with external data? Ideally it should work normally, but it seems worthy to confirm.

I can try that. Is there a good example I can use?

jcwchen:
> but textproto is conventionally the extension for it.

That's good to know. We should document the suggested file extension for textproto in next PR. I was just afraid users will use textproto format with a .onnx file extension and complain why such a .onnx model is not runnable by onnxruntime.

> I can try that. Is there a good example I can use?

https://github.com/onnx/onnx/blob/main/onnx/test/test_external_data.py#L103-L129

justinchuby:
Thanks. I will update the tests

onnx/__init__.py
```diff
@@ -240,63 +189,62 @@ def load_model(
...
+    s: bytes | str,
+    format: _SupportedFormat = "protobuf",  # pylint: disable=redefined-builtin
 ) -> ModelProto:
     """Loads a binary string (bytes) that contains serialized ModelProto.
 
     Args:
         s: a string, which contains serialized ModelProto
-        format: for future use
+        format: The serialization format. Default is "protobuf".
+        encoding: The encoding of the bytes when format is a text format. Default is "utf-8".
```
xadupre, line 216: encoding parameter was removed but I would still mention that utf-8 is used. Encoding is usually not used for english languages but used for every other language.

justinchuby:
Todo: rename the methods; updated supported types to include str

onnx/serialization.py
```diff
@@ -0,0 +1,125 @@
...
+import google.protobuf.text_format
+
+import onnx
+
+_Proto = TypeVar("_Proto", bound=google.protobuf.message.Message)
+# Encoding used for serializing and deserializing text files
+_ENCODING = "utf-8"
+
+
+class ProtoSerializer(abc.ABC):
```
justinchuby, line 26: Change to protocol

onnx/__init__.py
justinchuby: Done

justinchuby:
@jcwchen Tests updated and suggestions from @xadupre applied. PTAL.

onnx/serialization.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (c) ONNX Project Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+__all__ = [
+    "ProtoSerializer",
```
jcwchen, line 8: nit: Do we need to keep `ProtoSerializer` private at least for now? In the near future possibly ONNX will support more formats and the base code here might change, although it looks good now.

justinchuby: SG

justinchuby, line 8:
```suggestion
```

onnx/__init__.py
```diff
@@ -240,63 +192,65 @@ def load_model(
...
+    format: _SupportedFormat = "protobuf",  # pylint: disable=redefined-builtin
 ) -> ModelProto:
     """Loads a binary string (bytes) that contains serialized ModelProto.
 
     Args:
         s: a string, which contains serialized ModelProto
-        format: for future use
+        format: The serialization format. Default is "protobuf". The encoding is
+            assumed to be "utf-8" when the format is a text format.
+        encoding: The encoding of the bytes when format is a text format. Default is "utf-8".
```
justinchuby, line 221:
```suggestion
```

onnx/serialization.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+        return proto
+
+
+class _TextProtoSerializer(ProtoSerializer):
+    """Serialize and deserialize text proto."""
+
+    supported_formats = ("textproto",)
+
+    def serialize_proto(self, proto: _Proto) -> bytes | str:
```
justinchuby, line 109:
```suggestion
    def serialize_proto(self, proto: _Proto) -> bytes:
```

onnx/serialization.py
justinchuby: Done

jcwchen (approved):
Basically LGTM. TODO in next PRs:
- Add instruction in https://github.com/onnx/onnx/blob/main/docs/PythonAPIOverview.md and
- Expose this new support in https://github.com/onnx/onnx/blob/main/docs/docsgen/source/api/serialization.md.

gramalingam:
Maybe a test for user-registered custom format would be useful? Some dummy (simple) method, of course. We could try the printer/parser, for example.

jcwchen:
> Maybe a test for user-registered custom format would be useful? Some dummy (simple) method, of course. We could try the printer/parser, for example.

In offline decision: let's merge this PR first and have another PR to add tests for user-registered custom format.

PR #5114, opened 2023-04-11
Bump opset version to 20

yuanyao-nv:
### Description
Release branch rel-1.14.0 has been cut. Bump opset number to 20.


VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.14.0
+1.14.1
```
justinchuby, line 1: To 1.15.0?

yuanyao-nv: Updated. Do we no plan to have a minor release after this?

justinchuby: I think we can just cut rel branches and change this for minor releases. That's what pytorch does.

jcwchen (approved):
LGTM. Thanks!

PR #5117, opened 2023-04-11
Create repr functions for OpSchema

justinchuby:
### Description

Create repr functions for OpSchema

Example 
```
OpSchema(
    name='aten_abs',
    domain='onnxscript.atenlib',
    since_version=1,
    doc='abs(Tensor self) -> Tensor',
    type_constraints=[OpSchema.TypeConstraintParam(type_param_str='T', allowed_type_strs=['tensor(int64)'], description='Foo')],
    inputs=[OpSchema.FormalParameter(name='self', type_str='T', description='', param_option=<FormalParameterOption.Single: 0>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>)],
    outputs=[OpSchema.FormalParameter(name='return_val', type_str='T', description='', param_option=<FormalParameterOption.Single: 0>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>)],
    attributes={}
)
```

### Motivation and Context

Follow up of #5019


justinchuby:
@jcwchen @gramalingam

jcwchen (approved):

onnx/defs/__init__.py
```diff
@@ -47,6 +62,55 @@ def _attribute_default_value(self):  # type: ignore
...
+    inputs={self.inputs!r},
+    outputs={self.outputs!r},
+    attributes={self.attributes!r}
+)"""
+
+
+OpSchema.__repr__ = _op_schema_repr  # type: ignore
+
+
+def _op_schema_formal_parameter_repr(self) -> str:
```
xadupre, line 82: Why not implementing them in C++?

justinchuby: I think it is easier and cleaner in Python. For example, when repr() is called on OpSchema, it is automatically called on fields that contain data structures like FormalParameter etc. If we do it in c++, we need to hand format lists and call repr on nested data structure etc. I think? 

Also if anyone wants to change the code to prototype, they wouldn’t need to recompile from source.

justinchuby: Marking as resolved. LMK if you have further comments, thanks!

xadupre (approved):

gramalingam (approved):

gramalingam:
Do we need to increase the CI time-limit? We seem to have one time-out failure

Issue #4992, opened 2023-03-14
error: static assertion failed: Protobuf only supports C++14 and newer


yurivict:
# Bug Report

### Describe the bug
The build fails because protobuf header is built with -std=gnu++11 when it requires c++17.
I added -std=c++20 to CXXFLAGS but this is added earlier than -std=gnu++11 and doesn't take effect.

```
[ 10%] Building CXX object CMakeFiles/onnx_proto.dir/onnx/onnx-ml.pb.cc.o
/depot/gcc-12.2.0/bin/g++ -DONNX_API="__attribute__((__visibility__(\"default\")))" -DONNX_ML=1 -DONNX_NAMESPACE=onnx -D__STDC_FORMAT_MACROS -I/u/yuriv/nn-trainer/deps/onnx -isystem /u/yuriv/local/include -std=c++20 -Wnon-virtual-dtor -O3 -DNDEBUG -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -std=gnu++11 -MD -MT CMakeFiles/onnx_proto.dir/onnx/onnx-ml.pb.cc.o -MF CMakeFiles/onnx_proto.dir/onnx/onnx-ml.pb.cc.o.d -o CMakeFiles/onnx_proto.dir/onnx/onnx-ml.pb.cc.o -c /u/yuriv/nn-trainer/deps/onnx/onnx/onnx-ml.pb.cc
In file included from /u/yuriv/nn-trainer/deps/onnx/onnx/onnx-ml.pb.h:11,
                 from /u/yuriv/nn-trainer/deps/onnx/onnx/onnx-ml.pb.cc:4:
/u/yuriv/local/include/google/protobuf/port_def.inc:200:15: error: static assertion failed: Protobuf only supports C++14 and newer.
  200 | static_assert(PROTOBUF_CPLUSPLUS_MIN(201402L), "Protobuf only supports C++14 and newer.");
      |               ^~~~~~~~~~~~~~~~~~~~~~
/u/yuriv/local/include/google/protobuf/port_def.inc:200:15: note: the comparison reduces to ‘(201103 >= 201402)’
In file included from /u/yuriv/local/include/google/protobuf/stubs/common.h:44,
                 from /u/yuriv/local/include/google/protobuf/io/coded_stream.h:130,
                 from /u/yuriv/nn-trainer/deps/onnx/onnx/onnx-ml.pb.h:24:
```

### System information
[...]

PR #5119, opened 2023-04-12
Bump CMAKE_CXX_STANDARD version as 14

jcwchen:
### Description

Bump CMAKE_CXX_STANDARD version as 14.

### Motivation and Context

Close https://github.com/onnx/onnx/issues/4992. New Protobuf will only supports C++14 and newer. 11 is quite old and thus it seems fair for ONNX updates its required C++ version for better comparability with other packages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,9 +56,9 @@ else()
 endif()
 
 # Required to use /std:c++17 or higher on Windows
-# For other platforms, set C++11 as standard for the whole project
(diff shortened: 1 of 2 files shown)


justinchuby:
Let's go!

jcwchen:
I am inclined not to be too aggressive for upgrade at this moment and so let's just use 14 for now. We can reconsider whether to upgrade as 17 in next cycle (release).

PR #5120, opened 2023-04-13
Fix some build errors with the latest protobuf version

snnn:
### Description

onnx/defs/schema.h forgot to include <map> header file, but the file uses std::map

### Motivation and Context
Make it be compatible with the latest protobuf code.

### DIFF
--- a/onnx/defs/schema.h
+++ b/onnx/defs/schema.h
@@ -10,6 +10,7 @@
 #include <initializer_list>
 #include <iostream>
 #include <limits>
+#include <map>
 #include <memory>
 #include <ostream>
 #include <set>


jcwchen (approved):

jcwchen:
Please fix DCO also. Thanks! Let's cherry-pick this into upcoming ONNX 1.14.0. @onnx/sig-operators-approvers please help review/approve this PR.

gramalingam (approved):

snnn:
To clarify: this simple change is not enough to make ONNX be compatible with protobuf 22.x.  Previously, protobuf is just a single library.Either you use libprotobuf.lib, or you use libprotobuf-lite.lib. Once you added the file to your link command, you are done. Now protobuf has extra dependencies. protobuf depends on absl and utf8_validator. So the following code in ONNXs cmake file do not work anymore:

```cmake
  find_library(Protobuf_PROTOC_LIBRARY
                 NAMES protoc
                 PATHS ${_PROTOBUF_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}
                 NO_DEFAULT_PATH)
    if(ONNX_USE_LITE_PROTO)
      find_library(Protobuf_LITE_LIBRARY
        NAMES protobuf-lite
        PATHS ${_PROTOBUF_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}
        NO_DEFAULT_PATH)
    else()
      find_library(Protobuf_LIBRARY
        NAMES protobuf
        PATHS ${_PROTOBUF_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}
        NO_DEFAULT_PATH)
    endif(ONNX_USE_LITE_PROTO)
              NO_DEFAULT_PATH)
    find_package(Protobuf REQUIRED)
```
You cannot manually call find_library. You need to use find_package(protobuf ...) and use the cmake targets it provides. In general, when you call 
```
target_link_libraries(my_target PRIVATE lib1 lib2 ...)
```
Each lib can be either a filename, or a cmake target. cmake supports both, for historical reasons. However, when possible, you should only use targets. Otherwise you once one of your dependencies got new dependencies, you will need to update your cmake file.

jcwchen:
Thanks @snnn for the clarification and heads up. ONNX and ONNX Runtime are both using protobuf 21.12 right now. To confirm, do you think the update you mentioned for Protobuf 22.x needs to be included in the upcoming ONNX release? Or perhaps we can leave it to next release?

snnn:
I think either is ok.

jcwchen:
> I think either is ok. If the release date is coming soon, maybe it's better to not have big changes.

Same thought. Then I will prefer to do it after the release. I will merge this PR first. Thanks!

PR #5123, opened 2023-04-13
Extend float 8 support to control flow operators

xadupre:
### Description
Control flow does not support float 8 in opset 19 but it should probably make sense to have it.

It could happen in opset 19 or opset 20.

### DIFF
--- a/onnx/defs/controlflow/old.cc
+++ b/onnx/defs/controlflow/old.cc
@@ -8,6 +8,792 @@
 namespace ONNX_NAMESPACE {
 using SupportType = OpSchema::SupportType;
 
+static std::vector<std::string> control_flow_types_ir4() {
+  auto t = OpSchema::all_tensor_types_ir4();
+  auto s = OpSchema::all_tensor_sequence_types_ir4();
+  auto o = OpSchema::all_optional_types_ir4();
+  t.insert(t.end(), s.begin(), s.end());
+  t.insert(t.end(), o.begin(), o.end());
+  return t;
+}
+
+void IfInferenceFunction_16(InferenceContext& ctx) {
+  // there are no inputs so we just need to run the subgraph inferencing for
+  // then/else subgraphs and apply those to the outputs.
+  std::vector<const TypeProto*> subgraph_input_types; // none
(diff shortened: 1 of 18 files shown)


onnx/defs/controlflow/old.cc
```diff
@@ -8,6 +8,792 @@
...
+static std::vector<std::string> control_flow_types_ir4() {
+  auto t = OpSchema::all_tensor_types_ir4();
+  auto s = OpSchema::all_tensor_sequence_types_ir4();
+  auto o = OpSchema::all_optional_types_ir4();
+  t.insert(t.end(), s.begin(), s.end());
+  t.insert(t.end(), o.begin(), o.end());
+  return t;
+}
+
+void IfInferenceFunction_16(InferenceContext& ctx) {
```
gramalingam, line 20: Is there any change to the inference function? If not, it would be nice to avoid code-duplication and use a common function. The other folders use a `utils.h` and `utils.cc` for code shared by `defs.cc` and `old.cc`

onnx/defs/controlflow/utils.h
```diff
@@ -0,0 +1,21 @@
...
+
+#pragma once
+
+#include "onnx/defs/schema.h"
+
+namespace ONNX_NAMESPACE {
+
+void ClearShape(TypeProto& input_type);
+
+int handle_negative_axis_validate(const std::string& attrib, int axis, int rank);
```
justinchuby, line 13: Is this intentionally name differently in snake case?

onnx/defs/controlflow/utils.cc
```diff
@@ -0,0 +1,357 @@
...
+            subgraph_output_type->value_case());
+      }
+
+      // if there's an existing type check it matches. otherwise propagate
+      propagateElemTypeWithValidation(subgraph_output_type, loop_output_type);
+
+      if (is_loop_state_var) {
+        // shape may change across iterations so ignore.
+      } else {
+        // propogate shape
```
justinchuby, line 167: nit: from misspell: "propogate" is a misspelling of "propagate"

PR #5130, opened 2023-04-14
Add __all__ to __init__.py

justinchuby:
### Description

Add `__all__` to `__init__.py`

### Motivation and Context

API clarity and control. Intelisense will be able to tell what is exposed. Previously it often complains that "TypeProto is not exposed in onnx" etc.


justinchuby:
@jcwchen

jcwchen (approved):
It's great to have them in ```__all__``` for better explicitly. Thanks!

PR #5122, opened 2023-04-13
Tiny fix to the description of the Abs operator

ShayHacohen:
Previous description seemed like a typo; this commit is a suggested improvement to it

### Description
Just a tiny phrasing modification to the description of the Abs operator.

### Motivation and Context
The current description seems to contain a typo. Not a huge deal but I figured I might as well propose what I hope is a reasonable correction.

PR #5136, opened 2023-04-15
Slight fix to the Abs operator documentation

ShayHacohen:
including the resulting changes after running onnx/defs/gen_doc.py.

### Description
Just a tiny phrasing modification to the description of the Abs operator since the previous one seems to contain a typo.

### Motivation and Context
The current description seems to contain a typo. Not a huge deal but I figured I might as well propose what I hope is a reasonable correction.


ShayHacohen:
This is a remake of https://github.com/onnx/onnx/pull/5122 to avoid some weird shenanigans.
Thanks to @yuanyao-nv and @xadupre for the help!

Please let me know if there's any feedback.

xadupre (approved):

PR #5138, opened 2023-04-16
Enable string concatenation checks, G, C4, UP in ruff, applied autofix | chore(lint)

justinchuby:
### Description

Enable to ISC rules to check to error prone concatenations. Also     "C4", # flake8-comprehensions; "G", # flake8-logging-format and "UP" pyupgrade

All changes are autofix from ruff.

### Motivation and Context

Robustness and readability


onnx/backend/test/case/model/sequence.py
```diff
@@ -15,46 +14,46 @@
 from onnx.backend.test.case.model import expect
 
 
-def SequenceEmptyImpl() -> List[Optional[np.ndarray]]:
+def SequenceEmptyImpl() -> list[np.ndarray | None]:
```
gramalingam, line 17: This change seems orthogonal to what the PR description says. Also: is this important? I guess `List` is compatible with more python versions than `list`? What do we gain by such changes?

gramalingam:
Are all changes automated changes introduced by some tool? If so, it would be worth stating this in the PR description (along with the command used), since manual changes require more careful review.

onnx/backend/test/case/model/sequence.py
justinchuby: It's the fix from pyupgrade. Updated description. Since we have from `__future__ import annotations` it will work on python >=3.7

justinchuby: Also `List` is deprecated from 3.9 so it would be good for us to take gradual steps toward conventions. ruff ensures we support the lowest python version specified (3.7 for now). When we drop support we will be able to change the config in pyproject.toml and have ruff automatically modernize the code

gramalingam (approved):

Issue #4138, opened 2022-04-14
[Tracking] Eliminate duplicate string in operator document to prevent binary size increasing


jcwchen:
# Feature Request

### System information
ONNX version (you are using): 1.11

### What is the problem that this feature solves?
Prevent binary size increasing.

### Describe the feature
While bumping version for certain operator, currently people just created another string for the new document for the new version. Sometimes the document is very similar to previous document with few additional sentences or even is exactly the same as the previous one (for instance, add more type support does not influence the document content). It's a potential issue for binary size increasing if ONNX keeps using different string for the same/similar content. Also, using the same string is easier to maintain if we would like to update the same content for different versions. We should go through all operators under all defs.cc and old.cc to fix this. Open this issue to track this work item.

Furthermore, going forward we should try our best to share the string when bumping versions.

### Will this influence the current api?
No

### Feature Area
Package binary size.

### Are you willing to contribute it (Y/N):
Y

jcwchen:
Thanks for catching this. Yes, it is still open and anyone interested feel free to pick this up.

jcwchen:
Sure. Thanks for your interest. Typically operators in different opset_versions have duplicate string, which can be found from `onnx/defs/xxx/defs.cc` or `onnx/defs/xxx/old.cc`. Take Reshape-14 and Reshape-13 as an example:

https://github.com/onnx/onnx/blob/b8482d71eca8d35414e125a5cfbfc16797367fe4/onnx/defs/tensor/old.cc#L5002-L5016

and

https://github.com/onnx/onnx/blob/b8482d71eca8d35414e125a5cfbfc16797367fe4/onnx/defs/tensor/old.cc#L367-L374

The documents are very similar and only have few different sentences. Ideally we should make the duplicate part as a common string. You can directly modify the C++ files (defs.cc or old.cc).

PR #5146, opened 2023-04-17
Fix type constraints of Reshape(19)

xadupre:
### Description
Fix a wrong renaming introduced in PR #5123. Float 8 types were missing from Reshape allowed types.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -23461,7 +23461,7 @@ This version of the operator has been available since version 19 of the default 
 #### Type Constraints
 
 <dl>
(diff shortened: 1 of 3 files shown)


gramalingam (approved):
Minor nit: I guess the "doc" string can be shared by old and new defs since they are identical by moving it to utils.h/utils.cc. And renaming it to something appropriate like `Reshape_ver14_and_ver19_doc`. Helps when somebody is browsing the code.

jcwchen (approved):
cc release manager @yuanyao-nv

xadupre:
I suggest keeping it that way for this release and let us take time to think about a mechanism to avoid duplication in documentation.

jcwchen:
> I suggest keeping it that way for this release and let us take time to think about a mechanism to avoid duplication in documentation.

P.S. Tracked by this https://github.com/onnx/onnx/issues/4138.

PR #5160, opened 2023-04-20
Set providers argument in test_backend_onnxruntime.py

xadupre:
### Description
test_backend_onnxruntime.py does not work when onnxruntime is compiled with multiple providers. Fixes it.

### Motivation and Context
Tests are now passing on CPU and CUDA.


jcwchen (commented):
Do we need to similar update for https://github.com/onnx/onnx/blob/3ccb8eac6f8cfbeb014d60641613e0fe2e29f4e9/onnx/test/reference_evaluator_backend_test.py#L671?

jcwchen (approved):

jcwchen:
It seems that after this PR, two tests started to fail: `test_equal_string_broadcast_cpu
` and `test_equal_string_cpu`. See detail: https://github.com/onnx/onnx/actions/runs/4796246332/jobs/8531802272. It was not caught by Azure Pipelines in advance because of https://github.com/onnx/onnx/pull/5175.

Issue #3651, opened 2021-08-13
[Tracking] Spec clarifications


jcwchen:
Creating this issue to track work items for spec clarifications. There are already multiple GitHub issues regarding op spec not being clear or error in expected outputs. I will collect them in this issue. Feel free to add more work item here. Thanks!

jcwchen:
Old issues which look still valid:

### Pooling related
  - Inconsistent p range between LpNormalization, LpPool and GlobalLpPool: https://github.com/onnx/onnx/issues/1614
  - Ambiguity in Pool spec (AveragePool and MaxPool): https://github.com/onnx/onnx/issues/1922
  - Inconsistencies between MaxPool, AveragePool, and LpPool specs: https://github.com/onnx/onnx/issues/2290
  - MaxUnpool indices incompatible with PyTorch and inconsistent with MaxPool for kernels sizes which aren't a multiple of output size: https://github.com/onnx/onnx/issues/2398
  - operator MaxPool output tensor shape is ambiguously documented: https://github.com/onnx/onnx/issues/2927
  - Confused about AveragePool op ceil_mode attribute: https://github.com/onnx/onnx/issues/3314

### Conv/ConvTranspose related
  - convtranspose with dynamic output shape: https://github.com/onnx/onnx/issues/1772
  - Order of kernel application in Conv & ConvTranspose: https://github.com/onnx/onnx/issues/1886
  - Padding specification unclear for Convolution/AveragePool/MaxPool: https://github.com/onnx/onnx/issues/1651
  - Output shape of Conv not specified: https://github.com/onnx/onnx/issues/1795

### RNN/LSTM related
  - Documentation for RNN, LSTM, GRU shall specify output sequence order when direction is set to reverse: https://github.com/onnx/onnx/issues/626
  - GRU/RNN/LSTM's activation function: https://github.com/onnx/onnx/issues/1166
  - Is RNN hidden_size attribute optional or not? https://github.com/onnx/onnx/issues/1213
  - Spec bug to do with LSTM operator: https://github.com/onnx/onnx/issues/2116

### Other ops
  - [Document] The equivalent pseudo C code of loop: https://github.com/onnx/onnx/issues/3389
  - Rounding behavior of Resize when interpolating on integral types is unspecified: https://github.com/onnx/onnx/issues/3390
  - [ScatterElements] Clarify the requirement on indices uniqueness : https://github.com/onnx/onnx/issues/3234
[...]

PR #5169, opened 2023-04-21
Extend function type inference to handle missing optional parameters

gramalingam:
Extend function type inference to handle missing optional parameters.


onnx/test/function_inference_test.py
```diff
@@ -73,6 +76,31 @@ def test_fi_attribute(self):
         dtype_10 = onnx.helper.make_attribute("dtype", 10)
         self._check(code, [float_type_], [dtype_10], [float16_type_])
 
+    def test_fi_optional_input1(self):
```
justinchuby, line 79: maybe a more descriptive test name (than `input1`)?

gramalingam: The description is `optional_input`. I have two test-cases, so added suffixes 1 and 2. The comments describe the test-cases.

justinchuby: I recommend we split each group (behavior) into its own test case. For example, the first test can be named as “test_fi_omits_type_for_missing_trailing_optional_parameter” and so on. 

This way each test is focused on a single behavior, and it is super clear in the logs when any test fails.

gramalingam: Hmmm. I personally disagree. Trying to capture the full intention of a test in the function-name can't be met 100%, and I think that's ok. I feel the suggested name goes overboard. I prefer using comments for a more complete description.

justinchuby: I think it can be hard to capture 100% of the intention in a test name, and that’s ok. It is good to still do the exercise. When we can’t, it is usually a good hint that we may be able to test a smaller unit (of the logic or the behavior). 

This way of naming (`test_<method or function name>_<expected behavior>_[when_<condition>]`) encourages that we create tests that will validate a particular behavior (and in many cases that behavior only). So that we have cleanly separated, easy to understand test cases. 

In this particular case, I can see parameterizing the input creation part of the logic useful so that we don’t need to duplicate the model texts.

onnx/shape_inference/implementation.cc
```diff
@@ -588,10 +588,12 @@ class ShapeInferenceImplBase {
     // Create a temporary initializer value map
     for (int i = 0; i < num_actual_inputs && i < num_func_inputs; ++i) {
       const TypeProto* type = ctx.getInputType(i);
-      if (type->value_case() == TypeProto::kTensorType && ctx.getInputData(i) != nullptr) {
-        input_data_by_name[func_proto.input().Get(i)] = ctx.getInputData(i);
-      } else if (type->value_case() == TypeProto::kSparseTensorType && ctx.getInputSparseData(i) != nullptr) {
-        input_sparse_data_by_name[func_proto.input().Get(i)] = ctx.getInputSparseData(i);
+      if (type != nullptr) {
```
xadupre, line 591: What if the optional parameter is the first one and the second one is defined? What about the line 876 in that case?

gramalingam: It looks fine to me. I also don't see the connection between this line and line 876. Line 876 is returning the type for a particular input, and the contract is to return nullptr in the case of missing optional inputs. This line is part of the pre-processing to create known constant-values of inputs to help shape-inference. This line change is just fixing a bug that it wasn't handling optional inputs correctly.

gramalingam: Added test-case with a missing non-trailing optional parameter.

onnx/test/function_inference_test.py
gramalingam: Fair point. Let me clarify what I mean further. First, IMHO, this is a low-priority. Eg: I would love to see the items in [this issue](https://github.com/onnx/onnx/issues/3651) addressed ... they are higher priority because they impact a much wider audience (every backend that implements onnx and every frontend/exporter that generates onnx) when some aspect of an onnx operator spec is unclear. Yet, that remains mostly unaddressed, and has for a long time. Second, it is more important to ensure complete coverage of all behavior we want to test. Breaking them into units as you suggest may sometimes incur an overhead (it is even possible to end up with misleading names because precisely capturing the expected behavior and the conditions in a single identifier is often tricky). So, I will leave it as is. (I actually ended up merging the two test-cases into one, at least now we don't have artificial names with "1" and "2".)

xadupre (approved):

PR #5183, opened 2023-04-28
Create a generic Serializer class to handle serialization

justinchuby:
### Description

Create a generic Serializer class to handle serialization. Based on https://github.com/onnx/onnx/pull/5112

### Motivation and Context

- Refactor and modularize code from #5112 to handle more serialization scenarios.
- Also makes `load_model_from_string` accept a `str` input.

cc @jcwchen @gramalingam @xadupre


justinchuby:
I am going to combine this with #5112

Issue #5006, opened 2023-03-17
connect two onnx models as one


TMullerSG:
Hi, supposely i have two onnx models and i want to combine both of them to get the mean of the two output, how should i do?
 for example:
x=onnx.load("x.onnx")
y=onnx.load("y.onnx")

i want to have one onnx model which give the same result as x+y

thanks for any help

jbachurski:
There are some utilities in `onnx.compose` in this repository that could help, like `onnx.compose.merge_models` - you'd need to prepare a third model which computes a mean of its inputs and then that could work.

I can also recommend a solution using [Spox](https://github.com/Quantco/spox). First you'll have to find all the input/output types of the models. Then you can use `spox.inline` to put the two models in the same model. With `spox.argument` you can create placeholder model inputs that you can forward the composed model inputs with. Then you can compute a mean (like arithmetic mean).

```py
import numpy as np
import onnx
import onnx.compose
import spox.opset.ai.onnx.v17 as op
from spox import argument, build, inline, Tensor

# I'll assume your model inputs are some tensors a, b of these types
# You can adapt this to your case
a = argument(Tensor(int, (256, 256, 3)))
b = argument(Tensor(np.float32, ('N',)))

# Load the models
model1 = onnx.load("out/model1.onnx")
model2 = onnx.load("out/model2.onnx")

# Inline the models and forward the arguments
# Also unwrap all the results - here assume there's just one
(r1,) = inline(model1)(a, b).values()
(r2,) = inline(model2)(a, b).values()

# Take the arithmetic mean
# This assumes the outputs are float32 (as op.constant(value_float=...) is)
r = op.div(op.add(r1, r2), op.constant(value_float=2))  

# Build the composed model with given model inputs and outputs (with names)
model = build({'first': a, 'second': b}, {'result': r})
# onnx.save(model, ...)
```

I hope you can work out a solution based on this, but let me know if I can help further (or if you have more details on the models, like what input/output types they have).

praneeth5222:
Hi @jbachurski 
Can you also provide a sample example to combine two onnx models sequentially (i.e. output of the first model is passed as an input to the 2nd) which are of different domain (one is a sklearn onnx model and the other is a tensorflow one)

jbachurski:
What do you mean of different domain? Spox should attempt to do version conversion to move the models to a common opset version, you can also try to use `onnx.version_converter` yourself to do that. Is there a specific exception you are getting?

praneeth5222:
[...]

PR #5188, opened 2023-05-01
Fix shape inference errors for `map` type

jbachurski:
### Description

1. Fixes an ungracefully unhandled type case (in materialising a symbolic dimension), which always caused errors when a `map`-typed operator was used.
2. Completes ZipMap type inference to properly type the map value types as _scalar_ float tensors.
3. Adds a test for running ZipMap type inference (and hence testing that maps now are in fact supported by shape inference). 

### Motivation and Context

ONNX support for the `map` type is poor and the least documented. This is a small step towards adding more operators using `map` for more advanced pre-processing capabilities, more dynamic than what is allowed by the existing, attribute-oriented, `ai.onnx.ml` operators (like `LabelEncoder`).

Currently, any and all uses of shape inference where a `map` type is present results in the following error (e.g. for `ZipMap`):

```py
onnx.onnx_cpp2py_export.shape_inference.InferenceError: [ShapeInferenceError] (op_type:ZipMap): [ShapeInferenceError] type case unsupported for symbolic shape inference. inferred=5
```

This is due to a piece of code unrelated to a specific inference implementation, and hence map operators cannot be used at all as they will always raise this error. The shape inference test module had a similar lack of a map case. I skimmed through the code and I didn't see other such cases.

Also, related [comment](https://github.com/onnx/onnx/issues/5006#issuecomment-1530360359) where I uncovered this - `skl2onnx`-produced classifiers use `ZipMap` by default and hence can never succeed in shape inference.


onnx/test/shape_inference_test.py
```diff
@@ -8770,6 +8778,39 @@ def test_one_hot_encoder(self) -> None:
             ],
         )
 
+    @unittest.skipUnless(ONNX_ML, "ONNX_ML required to test ai.onnx.ml operators")
+    def test_zip_map(self) -> None:
```
xadupre, line 8782: Should we add a test to check the inference is correct when classlabels_strings is set up?

jbachurski: I can add it here for completeness of `test_zip_map` at the current version, though the primary goal of the PR is to have map-type inference work at all so I don't think I'll add other test cases.

xadupre (approved):

onnx/defs/traditionalml/defs.cc
```diff
@@ -1077,7 +1077,12 @@ ONNX_ML_OPERATOR_SET_SCHEMA(
           std::vector<std::string> classlabels_strings;
           bool result = getRepeatedAttribute(ctx, "classlabels_strings", classlabels_strings);
           auto output_map_type = ctx.getOutputType(0)->mutable_sequence_type()->mutable_elem_type()->mutable_map_type();
-          output_map_type->mutable_value_type()->mutable_tensor_type()->set_elem_type(TensorProto::FLOAT);
+          auto output_value_tensor_type = output_map_type->mutable_value_type()->mutable_tensor_type();
+          output_value_tensor_type->set_elem_type(TensorProto::FLOAT);
+          output_value_tensor_type->mutable_shape(); // Initialize to scalar
+          if (getInputShape(ctx, 0).dim_size() != 1 && getInputShape(ctx, 0).dim_size() != 2) {
```
gramalingam, line 1083: This should be further protected by `if (hasInputShape(ctx, 0))`

onnx/test/shape_inference_test.py
```diff
@@ -8770,6 +8778,48 @@ def test_one_hot_encoder(self) -> None:
...
+                    "ZipMap",
+                    ["input"],
+                    ["output"],
+                    **attrs,
+                    domain="ai.onnx.ml",
+                )
+            ],
+            [],
+        )
+        typ = onnx.TypeProto()
```
gramalingam, line 8805: nit: I guess the corresponding helper function is missing in helper.py. It may be worth adding it [here](https://github.com/onnx/onnx/blob/main/onnx/helper.py#L1090)

gramalingam (approved):

onnx/shape_inference/implementation.cc
```diff
@@ -229,6 +229,8 @@ void MaterializeSymbolicShape(TypeProto* inferred_type, SymbolTable& symbol_tabl
     MaterializeSymbolicShape(inferred_type->mutable_sequence_type()->mutable_elem_type(), symbol_table);
   } else if (inferred_val_case == TypeProto::kOptionalType) {
     MaterializeSymbolicShape(inferred_type->mutable_optional_type()->mutable_elem_type(), symbol_table);
+  } else if (inferred_val_case == TypeProto::kMapType) {
```
gramalingam, line 232: Thanks for this fix. This was also uncovered in an earlier PR adding more map ops, but that PR seems to be on hold. However, it does have one more line of fix which may be relevant and worth adding here: see [this PR](https://github.com/onnx/onnx/pull/4640/files#diff-df8e9d76c8ada1b583a5dd877963bd651fbf76ec0cac241d05bf97af4ef8c8f4)

gramalingam: image

gramalingam: The change to `mergeShapesAndTypes` is highlighted above

jbachurski: Thanks! I took a short look and it seems the only other general fix is for `from_dict` in `numpy_helper`, but I think that function should probably be rewritten in another PR.

jbachurski: I'm unsure why tests didn't catch this, though. I think the code isn't actually entered in our shape inference tests, as if the type was entirely undefined before, the inferred type is copied directly, so the undefined-key-type branch was never hit: 
https://github.com/onnx/onnx/blob/ebb64fcbf48d81fe82af2c551e6ca514e553071b/onnx/shape_inference/implementation.cc#L260-L274

Issue #5194, opened 2023-05-03
ModuleNotFoundError: No module named 'cmake' during pip install -e .


guoyuhong:
# Bug Report

### Is the issue related to model conversion?

No. It related to the onnx build system.

### System information

1. System: Linux or MacOs
2. ONNX version: main branch
3. GCC/Compiler: Apple clang version 13.1.6 or gcc (GCC) 8.3.1
4. miniconda, python3.10 or python 3.8
5. cmake  installed by `pip install cmake`

### Reproduction instructions

`pip install -e .`
#### MacOs
![image](https://user-images.githubusercontent.com/19584326/235886209-248e5ac2-6ead-459a-9144-63fee0038fc1.png)

#### Linux
![image](https://user-images.githubusercontent.com/19584326/235886073-dcb98f75-e0fe-4a0c-88de-85bc2d4be239.png)

#### Windows
![image](https://user-images.githubusercontent.com/19584326/236002702-d018839a-a3ca-49a2-a516-e6c19fe4ffdf.png)

### Notes

`python setup.py build` works fine.

jcwchen:
Hi @guoyuhong,
I cannot repro this issue somehow, although my `pip install -e .` also added `*-pip-build-env-*` into PYTHONPATH during build. What's your setuptools and pip version? Have you tried the latest setuptools and pip? I wonder why they add pip-build-env into PYTHONPATH and not sure whether it is intentional.

guoyuhong:
@jcwchen In my env, cmake should be installed from pip and the system cmake should not be used. If I use the system cmake, I cannot repro this issue either.  My env setup:
1. cmake version 3.26.3 (installed by pip)
2. pip 22.3.1 or 23.1.2
3. setuptools             67.7.2

guoyuhong:
> @jcwchen In my env, cmake should be installed from pip and the system cmake should not be used. If I use the system cmake, I cannot repro this issue either. My env setup:
> 
> 1. cmake version 3.26.3 (installed by pip)
> 2. pip 22.3.1 or 23.1.2
> 3. setuptools             67.7.2

@jcwchen can you repro this failure with this configuration?

jcwchen:
> @jcwchen can you repro this failure with this configuration?

Yes, thanks for providing the configuration. After removing system cmake and installing cmake from PyPI, I can repro this issue. Also confirmed your PR can solve it.

balazon:
[...]

PR #5195, opened 2023-05-03
Fix ModuleNotFoundError: No module named 'cmake'

guoyuhong:
### Description

Described by https://github.com/onnx/onnx/issues/5194

### Motivation and Context

Since `python setup.py build` works fine, I compared the `os.environ` and found that `pip install -e .` add a strange `PYTHONPATH` to `os.environ`. Deleting it make the installation working fine.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -219,17 +219,15 @@ class CmakeBuild(setuptools.Command):
                 raise RuntimeError(
                     "-DONNX_DISABLE_EXCEPTIONS=ON option is only available for c++ builds. Python binding require exceptions to be enabled."
                 )
-            if "PYTHONPATH" in os.environ and sys.platform != "win32":
-                # To solve problem that the extra PYTHONPATH env variable added by pip.
-                segs = os.environ["PYTHONPATH"].split(":")
-                new_paths = []
-                for path in segs:
-                    if path not in sys.path:
-                        new_paths.append(path)
-                if len(new_paths) == 0:
-                    del os.environ["PYTHONPATH"]
-                else:
-                    os.environ["PYTHONPATH"] = ":".join(new_paths)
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -219,6 +219,17 @@ def run(self):
                 raise RuntimeError(
                     "-DONNX_DISABLE_EXCEPTIONS=ON option is only available for c++ builds. Python binding require exceptions to be enabled."
                 )
+            if "PYTHONPATH" in os.environ:
+                # To solve problem that the extra PYTHONPATH env variable added by pip.
+                segs = os.environ["PYTHONPATH"].split(":")
```
xadupre, line 224: This won't work on Windows.

guoyuhong: @xadupre Does windows has the same problem? Is it OK to skip windows platform using `"PYTHONPATH" in os.environ and sys.platform != "win32"`

xadupre: I assume miniconda works the same way on Windows. You remove every path from PYTHONPATH already added in sys.path so it starts looking cmake in the installed packages first. Is it possible to add a longer comment then to tell this is needed for conda.

guoyuhong: @xadupre I can get the same error in Windows env. With the new fix, the problem is also fixed in Windows env.
image

With the fix:

image

justinchuby:
Is this tested in all different environments: venv, conda and w/o venv?

guoyuhong:
@justinchuby I have tested this change in following envs:
1. conda env in Linux, MacOS, Windows 
2. virtualenv in Linux and MacOS
3. w/o venv(python installed in system) in Linux and MacOS

xadupre (approved):

Issue #5162, opened 2023-04-20
[Tracking] Make ONNX be compatible with protobuf 22.x. 


jcwchen:
### System information

Latest main branch

### What is the problem that this feature solves?

Make ONNX be compatible with protobuf 22.x. 

### Describe the feature

To make ONNX be compatible with protobuf 22.x, CMake file will need to updated. Details: https://github.com/onnx/onnx/pull/5120#issuecomment-1507589785

### Will this influence the current api (Y/N)?

N

### Feature Area

build

### Are you willing to contribute it (Y/N)

None

guoyuhong:
@jcwchen I would like to take some of the tasks.
In my opinion, it would take several steps. For example, 
1. make both cmake compatible with both protobuf 21.x and 22.x.
2. change the CI system to use 22.x.
3. Bump the python dependency to 22.x.
4. etc.
I would like to take the first step.

jcwchen:
@guoyuhong Thank you for the contribution! I think 3. (Bump the python dependency to 22.x.) won't happen in the near future since ONNX has just bumped its Protobuf version to 21. in the upcoming release, but definitely we can at least have 1. and 2. done first.

PR #5196, opened 2023-05-03
Support protobuf v22.x in cmake

guoyuhong:
### Description

https://github.com/onnx/onnx/issues/5162

### Motivation and Context

1. In Cmake file, support both v22.x and v3.21 of protobuf.
2. The externally provided protobuf can still be used, which can save a lot of building time.

#### use the installed protobuf
CI case: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=45493&view=logs&j=6a6c43ac-ab1d-5bbb-9edb-01c3c074f0c7&t=4989e718-93eb-58c7-f2f7-f09ffccbc645

#### build protobuf v22.x
CI case: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=45493&view=logs&j=4361362a-6b98-56f3-7411-ef4bf23e58c7&t=623793f8-a29b-5bfc-e8b0-f715ca237fd9


guoyuhong:
@snnn Please help to take a look.

snnn:
LGTM.

snnn:
Though this change is pretty good, we should also realize that building ONNX from source will be much more difficult than before, because of the new dependencies. I would suggest considering replacing the find_package calls to FetchContent:https://cliutils.gitlab.io/modern-cmake/chapters/projects/fetch.html . The FetchContent is superior to find_package. It has all the functionality that find_package provides(since it internally falls back to find_package ), also it would not require users to install anything before running cmake.

guoyuhong:
@snnn Yes, FetchContent is a better way. If FetchContent is used, the script is not necessary. Furthermore, it can unify the Unix-like system and the Windows system. I will take a look at this change later on.

guoyuhong:
@snnn @jcwchen This PR is ready for review. `FetchContent` is used, so there is no need to add a script to build protobuf. The external protobuf build is still supported. Only when there is no external protobuf, onnx will build its own protobuf libs.

guoyuhong:
@jcwchen Would you please take a look at this change?

jcwchen:
> @jcwchen Would you please take a look at this change?

Sure. This PR is on my radar, but I will need more time to review/test it, since it's fundamental build change. Thank you for proposing this support!

CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
-# Minimum CMake required
-cmake_minimum_required(VERSION 3.1)
+# Minimum CMake required.
+# FetchContent module requires 3.11.
+cmake_minimum_required(VERSION 3.11)
 include(cmake/Utils.cmake)
+include(FetchContent)
```
jcwchen, line 5: Could we only include FetchContent when Build_Protobuf is needed?

jcwchen (commented):
Thank you for the contribution. I am very looking forward to seeing this enhancement. Couple of questions:
- Could we also add similar Protobuf build tests in Windows CI as well?
- I am thinking whether we should only keep "single" internal Protobuf version (v22) for simplicity. For users who interested in other Protobuf versions, they can just build Protobuf from their own. WDYT?
- If possible, could we also enable debug build for Protobuf at this moment? For instance, make Build_Protobuf consume CMAKE_BUILD_TYPE and build a debug build Protobuf when CMAKE_BUILD_TYPE is Debug.

guoyuhong:
@jcwchen For your comments:
1. I don't know how the Windows env has installed the external Protobuf libs, shall I change it in next PR? By the way, is protobuf installed by conda command `conda install -n py$(python.version) -y -c conda-forge numpy libprotobuf==3.21.5`?
2. Yes, only one version is simple. I have removed the old protobuf.
3. I have verified that `CMAKE_BUILD_TYPE` will affect the sub-project from `FetchContent`. Furthermore, `BUILD_SHARED_LIBS` will also affect the sub-project behavior. However, the `BUILD_SHARED_LIBS`  of sub-projects should be determined by `ONNX_USE_PROTOBUF_SHARED_LIBS`. I also fix the problem.

CMakeLists.txt
guoyuhong: Good point, this will have better compatibility. Changed.

.azure-pipelines/Linux-CI.yml
```diff
@@ -12,26 +12,31 @@ jobs:
...
         onnx_ml: 0
         build_type: 'Release'
         documentation: 0
+        protobuf_type: 'Internal'
       py39-ml:
         python.version: '3.9'
         onnx_ml: 1
         build_type: 'Release'
         documentation: 1
+        protobuf_type: 'InternalV22'
```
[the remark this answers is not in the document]
guoyuhong: Done

CMakeLists.txt
```diff
@@ -176,12 +172,121 @@ else()
     find_path(Protobuf_INCLUDE_DIR google/protobuf/service.h
               PATHS ${_PROTOBUF_INSTALL_PREFIX}/include
               NO_DEFAULT_PATH)
-    find_package(Protobuf REQUIRED)
+    find_package(Protobuf)
+    if (Protobuf_FOUND)
+      if (ONNX_USE_PROTOBUF_V22 AND "${Protobuf_VERSION}" VERSION_LESS "4.22.0")
```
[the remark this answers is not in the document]
guoyuhong: FetchContent won't set that variable, so I set one. If protobuf is an external lib, it will have the version.
For our url used as external lib, the version is "4.22.3", the old protobuf is "3.21.12".
image

CMakeLists.txt
```diff
@@ -176,12 +172,121 @@ else()
...
+      utf8_range::utf8_validity
+    )
+  else()
+    set(protobuf_ABSL_USED_TARGETS "")
   endif()
 endif()
 
-# Build the libraries with -fPIC
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+# Protobuf build would fail if ONNX_WERROR is on.
```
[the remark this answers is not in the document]
guoyuhong: Actually,  the build error comes from abseil. I have changed the comment.
image

.azure-pipelines/Windows-CI.yml
```diff
@@ -52,9 +61,13 @@ jobs:
       git submodule update --init --recursive
       set ONNX_BUILD_TESTS=1
       set ONNX_ML=$(onnx_ml)
-      set CMAKE_ARGS=-DONNX_USE_PROTOBUF_SHARED_LIBS=ON -DONNX_USE_LITE_PROTO=ON -DONNX_WERROR=ON
+      if '$(protobuf_type)' == 'External' (
+        set CMAKE_ARGS=-DONNX_USE_PROTOBUF_SHARED_LIBS=ON -DONNX_USE_LITE_PROTO=ON -DONNX_WERROR=ON
+      ) else (
+        set CMAKE_ARGS=-DONNX_USE_PROTOBUF_SHARED_LIBS=OFF -DONNX_USE_LITE_PROTO=ON -DONNX_WERROR=OFF
```
guoyuhong, line 67: @jcwchen One more thing, I found that in Windows Env, when onnx is built with internal protobuf, there will be compile warning treated as error. I use `-DONNX_WERROR=OFF` when internal protobuf is used. Is this acceptable?
image

.azure-pipelines/Windows-CI.yml
```diff
@@ -52,9 +61,13 @@ jobs:
...
+      )
 
-      pip install -e .
+      pip install -e . -v
```
jcwchen, line 70: Thank you for adding `-v` to make the log more friendly for debugging! Could you please add `-v` for `pip install -e .` in Linux and Mac CI as well?

guoyuhong: Done

CMakeLists.txt
```diff
@@ -405,13 +498,13 @@ target_compile_definitions(onnx_proto PRIVATE ${ONNX_API_DEFINE})
...
   if(TARGET protobuf::libprotobuf-lite)
-    target_link_libraries(onnx_proto PUBLIC protobuf::libprotobuf-lite)
+    target_link_libraries(onnx_proto PUBLIC protobuf::libprotobuf-lite PRIVATE ${protobuf_ABSL_USED_TARGETS})
   else()
     target_link_libraries(onnx_proto PUBLIC ${PROTOBUF_LITE_LIBRARIES})
   endif()
 else()
   if(TARGET protobuf::libprotobuf)
-    target_link_libraries(onnx_proto PUBLIC protobuf::libprotobuf)
+    target_link_libraries(onnx_proto PUBLIC protobuf::libprotobuf PRIVATE ${protobuf_ABSL_USED_TARGETS})
```
jcwchen, line 507: `protobuf_ABSL_USED_TARGETS` is only needed when `("${Protobuf_VERSION}" VERSION_GREATER_EQUAL "4.22.0")`?

jcwchen (approved):
Overall looks good to me. Thank you for resolving the comments! Previously, missing installed Protobuf is a very common error when building onnx and users need to get Protobuf installed from their own. The PR will be also helpful in that case.

cc @snnn in case you have other comments about this PR for internally installing Protobuf and support Protobuf v22. Otherwise, we are close to merge it. Thanks!

CMakeLists.txt
guoyuhong: This is a good question. I have test many cases:
1. If the protobuf is built in this project, the generated `.cmake` files will expose the dependency on ABSL libs, so it not necessary to use this `protobuf_ABSL_USED_TARGETS`, which means it's OK to be empty. 
2. `("${Protobuf_VERSION}" < "4.22.0")` there is no such dependency.
3. When protobuf is pre-installed and  when `("${Protobuf_VERSION}" >= "4.22.0")`, there could be no such generated `.cmake` files in pre-installed package and `find_package` only returns the protobuf libs only. In this case, we need to use this protobuf_ABSL_USED_TARGETS. Otherewise, there will be link errors.

Finnally, the conclusion is that `protobuf_ABSL_USED_TARGETS` is needed when protobuf is pre-installed and  when `("${Protobuf_VERSION}" VERSION_GREATER_EQUAL "4.22.0")`.

PR #5193, opened 2023-05-03
Modify OneHot operator explanation

DZADSL72-00558:
### Description

referring to https://github.com/onnx/onnx/blob/main/onnx/defs/tensor/defs.cc#L2676-L2683

Onehot operator should support scalar and tensor with rank 1 both for depth input.

PR #5191, opened 2023-05-02
Deprecate Python 3.7 due to EOL

jcwchen:
### Description

- Deprecate Python 3.7 due to EOL.
- Use `pip install -e .` instead of `python setup.py install` in CIs, because it is the suggested way to develop onnx mentioned in README.md.
- Set correct permission for clang-tidy.
- Remove dependency: typing_extensions entirely, because starting Python 3.8, native typing support Literal and Protocol.

### Motivation and Context

- Closes https://github.com/onnx/onnx/issues/5163. Python 3.7 will be EOL on 27 Jun 2023.
- Closes https://github.com/onnx/onnx/issues/3750.

PR #5197, opened 2023-05-03
Modify OneHot operator explanation

andembry:
referring to https://github.com/onnx/onnx/blob/main/onnx/defs/tensor/defs.cc#L2676-L2683

### Description
Onehot operator should support scalar and tensor with rank 1 both for depth input.

Previous PR by my teammate was [here](https://github.com/onnx/onnx/pull/5193). I made the changes mentioned by @xadupre in that PR: modified Operators.md, built the package and ran gen_doc.py, and signed the commit.


andembry:
I see there is one python3.7 test failure. I don't see how these doc changes could've caused that, but at any rate it should be resolved by the [python3.7 EOL PR](https://github.com/onnx/onnx/pull/5191).

gramalingam (approved):

jcwchen (approved):
> I see there is one python3.7 test failure. I don't see how these doc changes could've caused that, but at any rate it should be resolved by the https://github.com/onnx/onnx/pull/5191.

I rerun the CI and now that one has passed. Thanks for improving the document!

PR #5204, opened 2023-05-05
Transpose scalar shape inference

adityagoel4512:
### Description
Shape inference when transposing scalars.

### Motivation and Context
Currently when transposing a scalar, the output shape is left unset leading to Shape Inference errors. This PR makes sure this doesn't happen by ensuring the output type has a 0 dimensional shape for scalar inputs.


onnx/defs/tensor/defs.cc
```diff
@@ -1191,6 +1191,12 @@ ONNX_OPERATOR_SET_SCHEMA(
             }
           }
 
+          // If transposing a scalar, we still want to infer a 0 dimension output shape
+          if (!shape.dim_size()) {
```
gramalingam, line 1195: I was wondering why the existing code wouldn't do the job already (with the zero-dimensional case being an instance of the more general n-dimensional case). I think what is missing is the initialization of the output-shape. I think the line
```cpp
getOutputShape(ctx, 0);
```
should be sufficient to do this initialization. May be worth trying.

adityagoel4512: I agree it is the initialisation of the output shape - I wasn't quite sure how to initialise this so thanks

gramalingam (approved):

jcwchen (approved):
Thanks for fixing this!

PR #5208, opened 2023-05-08
Update the latest IR release date: IR VERSION 9 published on May 5, 2023

jcwchen:
### Description

Update the latest IR release date: IR VERSION 9 published on May 5, 2023.

### Motivation and Context

Although previously onnx only updated its TBD date when there is a newer IR bump in future release, it seem better and harmless to me that we can update the proto files right after the fresh IR bump for better documentation.

### DIFF
--- a/docs/OnnxReleases.md
+++ b/docs/OnnxReleases.md
@@ -117,5 +117,5 @@ The ONNX project, going forward, will plan to release roughly on a four month ca
 **Bump opset version for ai.onnx**
 * Bump opset version for ai.onnx domain in `onnx/defs/operator_sets.h` and `onnx/defs/schema.h` for use by future operator additions and changes. For example, this [demo PR](https://github.com/onnx/onnx/pull/4134/files).
 
-**Update IR TBD date if there is a IR bump in release**
-* Update the latest IR TBD date in https://github.com/onnx/onnx/blob/main/onnx/onnx.in.proto and regenerate correpsonding proto files in the main branch if there is an IR bump in the release.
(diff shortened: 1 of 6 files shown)


docs/OnnxReleases.md
```diff
@@ -62,13 +62,6 @@ The ONNX project, going forward, will plan to release roughly on a four month ca
 * Test the PyPI package installation with different combinations of various Python versions, Protobuf versions and platforms.
   * Python versions : Applicable python versions for the release.
   * Protobuf versions : Latest protobuf version at the time of the release + protobuf version used for previous release
-  * Utilize the following matrix to check:
```
gramalingam, line 65: Wondering what this is/was?

gramalingam (approved):

docs/OnnxReleases.md
jcwchen: Just feel no need to manually update the Python version here every time. The sentence above has expressed the same meaning.

gramalingam: I see. Ok. But isn't the testing of the PyPI package installation automated (the set of combinations that are tested)? Is this done manually?

jcwchen: Yes, you are right. That's another reason. Now release CI pipelines in onnx repo has covered every case and so the release manager doesn't need to test every of them manually. Ideally I would say the release manager can just test the wheel in their local for 1-2 environments, which should be sufficient. Thus I don't want the document here have explicit matrix for testing.

PR #5211, opened 2023-05-08
Integrate function-inlining with version-conversion

gramalingam:
This extends the function-inliner to take advantage of the version-converter to inline functions targetting ONNX opset X into a model targetting ONNX opset Y (when conversion of the function from version X to version Y is successful).

The two are coupled because the existing version-converter does not support functions. It is not easy to extend the version-converter to handle functions because some of the version-conversion rules depend on context-information (such as the types of inputs or values of attributes) that may not be available in a function body. However, in the context of inlining a function into a callsite in a model, the necessary context information can be obtained, enabling the conversion to proceed.

The functionality is complete. But there are a couple of pending improvements, which can be delayed to a later PR

(i) It may be useful to expose an option to the caller to indicate if we should attempt version-conversion or not.

(ii) Version-conversion requires us to update type information for the model after one or more inlining steps, in order to apply version-conversion to nested function-calls. The current implementation uses a somewhat naive strategy to do this. We can improve this by using an incremental type-inference implementation.


onnx/checker.h
```diff
@@ -81,6 +89,7 @@ class CheckerContext final {
   bool is_main_graph_ = true;
   const ISchemaRegistry* schema_registry_ = OpSchemaRegistry::Instance();
   std::string model_dir_;
+  bool skip_opset_compatibility_check_ = false;
```
xadupre, line 92: Why use the prefix skip?

onnx/common/constants.h
```diff
@@ -13,7 +13,18 @@ namespace ONNX_NAMESPACE {
 constexpr const char* AI_ONNX_ML_DOMAIN = "ai.onnx.ml";
 constexpr const char* AI_ONNX_TRAINING_DOMAIN = "ai.onnx.training";
 constexpr const char* AI_ONNX_PREVIEW_TRAINING_DOMAIN = "ai.onnx.preview.training";
+// The following two are equivalent in an onnx proto representation.
 constexpr const char* ONNX_DOMAIN = "";
+constexpr const char* AI_ONNX_DOMAIN = "ai.onnx";
+
+inline std::string NormalizeDomain(std::string domain) {
```
xadupre, line 20: Why std::string here and const std::string& just below?

onnx/inliner/inliner.cc
```diff
@@ -262,11 +310,120 @@ class Specializer {
...
+      return vi.type();
+  }
+  for (auto& vi : model.graph().output()) {
+    if (vi.name() == var)
+      return vi.type();
+  }
+  ONNX_ASSERTM(false, "Type unknown for ", var);
+}
+
+void ConvertVersion(ModelProto& model, const NodeProto& call_node, FunctionProto& function, int target_version) {
```
xadupre, line 367: What if there are multiple opset involved?

onnx/inliner/inliner.cc
```diff
@@ -262,11 +310,120 @@ class Specializer {
...
+  for (auto& function_node : function_nodes)
+    *nodes.Add() = function_node;
+  function_nodes.Clear();
+
+  auto converted = ONNX_NAMESPACE::version_conversion::ConvertVersion(function_as_model, target_version);
+
+  function_nodes.Swap(converted.mutable_graph()->mutable_node());
+
+  // Append new initializers to main graph initializers
+  auto& added_initializer = converted.graph().initializer();
```
xadupre, line 416: sparse initializers?

onnx/inliner/inliner.cc
```diff
@@ -298,27 +466,31 @@ void InlineFunctions(ModelProto& model, FunctionResolver resolver) {
...
+  // For every function, we check if there is a mismatch between the opset versions
+  // required for the function and the model. If there is no mismatch, we can inline
+  // this function. If there is a mismatch only for the standard ONNX domain, we
+  // can inline after version-conversion (if the version-conversion is successful).
+  // Otherwise, we cannot inline, since currently version-conversion supports only
+  // standard ONNX domain.
+
   for (auto& function : model.functions()) {
     auto mismatches = model_imports.Mismatches(function.opset_import());
+    auto iter = mismatches.find(ONNX_DOMAIN);
```
xadupre, line 482: What about other domains?

onnx/test/cpp/inliner_test.cc
```diff
@@ -174,19 +174,80 @@ bar (x) => (y) {
...
   // Check that it is a call to Add
   ASSERT_EQ(second_node.op_type(), "Add");
 
-  // The non-inlined foo must still be in the function list.
-  ASSERT_EQ(model.functions_size(), 1);
-  ASSERT_EQ(model.functions(0).name(), "foo");
+  ASSERT_EQ(model.functions_size(), 0);
+}
+
+TEST(FunctionInliner, VersionConversion) {
```
xadupre, line 190: What happens if main opset < function opset?

onnx/checker.h
gramalingam: I find it less ambiguous (than something like `opset_compatibility_check`).

onnx/common/constants.h
gramalingam: Fixed, thanks!

onnx/inliner/inliner.cc
gramalingam: Currently, the version-converter supports only version-conversion for a single domain. If and when the version-converter is extended to support other domains, we can extend this too.

onnx/inliner/inliner.cc
gramalingam: We will not attempt version conversion if there is a mismatch for any other domain, since we don't yet have a version-converter for any other domain.

onnx/test/cpp/inliner_test.cc
gramalingam: Version-conversion is on a best-effort basis. It may not succeed. Current support is better for up-conversion than down-conversion. If there is a demand, hopefully we can improve the down-conversion too.

xadupre: I was thinking about converting the main graph to the latest opset and keeping the inlined function to the same opset it already uses.

gramalingam: That's a good point. For now, users can do it themselves by invoking the version-converter directly. So, I think we can ignore it here. If we want a unified interface for that, let's do it as a separate PR ... there are other possibilities, like some functions have lower version, some functions have higher version than the model ... may be it is better to let the users specify the desired target version. instead of us making a choice automatically?

xadupre: Maybe we could have an error message which suggests that option. I don't remember what happens for the local functions when the user upgrades the opset for the pain graph.

onnx/inliner/inliner.cc
xadupre: another PR i guess

onnx/inliner/inliner.cc
gramalingam: Added. However, none of existing version-converters add a sparse initializer. Some of them do add a new initializer. I think it is a better idea to eliminate even that, and uniformly add a new Constant node (instead of an initializer), which some of the version-converter rules do. (In the long run, ideally we would not require either of these.)

gramalingam:
> Would it be possible to add a section in the python documentation as well about the inliner (https://github.com/onnx/onnx/tree/main/docs/docsgen/source/api)?

Done

liqunfu (approved):

PR #5219, opened 2023-05-11
Fix `shape` data propagation function to handle missing optional parameters

BowenBao:


gramalingam:
Thanks for the fix. I don't think it has to do with optional parameter (because Shape doesn't have an optional parameter). But this problem can show up in other cases (eg., erroneous model, etc.). The model-checker will catch cases where we have a call to Shape without a parameter, but if we try running shape-inference before running the checker (as you mentioned elsewhere), this could be a problem. It could happen in other places too, because this style of code has been used quite a bit.

gramalingam:
As I discuss elsewhere, this could be a cascading problem arising out of the fact that attribute-parameters with default-values are not yet handled by checker/inference.

gramalingam (approved):

BowenBao:
Thanks @gramalingam for analysis! Indeed there is another root problem to be fixed. I'm going to try track in the example model where it first happened. Will update on the original issue.

gramalingam:
> Thanks @gramalingam for analysis! Indeed there is another root problem to be fixed. I'm going to try track in the example model where it first happened. Will update on the original issue.

This change helps me process the short-example you had ("maf_gpt2_tiny.onnx"). May be you can try it on the bigger example. (I am sure there will be other places where change is required to handle attributes-with-default-values.)

BowenBao:
Hmm, applying that still segfaults on my end, but maybe I updated a few things in the model that is different from one that was shared. Resharing the[ latest one](https://microsoft-my.sharepoint.com/:u:/p/bowbao/Ebtt_SQsuaFDotGl5p0sIZcBAQzdDCZ4BBBI5MEFCQOm3A?e=l3Trvy) . 

I have a few observations so far and some guesses.

It looks like shape inference for `aten_view_1` in `transformers.pytorch_utils.Conv1D(L__self___h_0_attn_c_attn)` stopped in the middle. An error was captured and added to the `inference_errors` list, and shape inference continues to the next function node.
```
# printf logs
Caught InferenceError for node aten_view_1 error msg [ShapeInferenceError] (op_type:Cast, node name: n0): [ShapeInferenceError] Data for input  0 already exists.
Adding inference error for node aten_view_1
```

https://github.com/onnx/onnx/blob/065c103533c6cef70de2f23152fce67a69806600/onnx/shape_inference/implementation.cc#L413-L423

Due to this, the next op received nullptr for its input type. It propagated until reaching a `Shape` node which segfaults.

I wonder if this is the cascading result of the attribute issue or something else?

Issue #5216, opened 2023-05-10
`onnx.helper.make_attribute` could not infer type correctly from empty list


take-cheeze:
# Bug Report

### Describe the bug

Since `all([])` is always `True` passing empty list to `make_attribute` will be always list of ints
https://github.com/onnx/onnx/blob/a02166572855083b3b6f5a329ba438f8cd8635f9/onnx/helper.py#L858-L860

### Reproduction instructions

```python
onnx.helper.make_attribute("attr_name", [])  # Always .type == INTS which isn't always expected
```

### Expected behavior

For empty list `make_attribute` method should fail due to type inference error or there should be an explicit typed `make_attribute` method

jbachurski:
You're right, this behavior is a bit awkward,  though at least it has been like this for a while so it's predictable. It could be sensible to add an explicit `type: Optional[AttributeProto.DataType] = None ` keyword argument, and have an error raised when the iterable is missing and the type can't be inferred.

PR #5220, opened 2023-05-11
Improve error handling in empty list input for helper.make_attribute

take-cheeze:
### Description

Closes https://github.com/onnx/onnx/issues/5216

### Motivation and Context

Empty list input should raise type inference failure and attribute type should be able to specified explicitly


onnx/helper.py
```diff
@@ -854,34 +857,57 @@ def make_attribute(  # pylint: disable=too-many-statements
     # Iterable cases
     elif isinstance(value, collections.abc.Iterable):
         value = list(value)
-        types = {type(v) for v in value}
-        if all(issubclass(t, numbers.Integral) for t in types):
+        if len(value) == 0 and attr_type is None:
+            raise ValueError("Could not infer attribute type from empty iterator")
+        if attr_type is None:
+            types = {type(v) for v in value}
```
xadupre, line 863: After this line, you can check the length of types and raise an exception if it is >= 2 instead of check all types in this list are the same.

onnx/helper.py
```diff
@@ -854,34 +857,57 @@ def make_attribute(  # pylint: disable=too-many-statements
     # Iterable cases
     elif isinstance(value, collections.abc.Iterable):
         value = list(value)
-        types = {type(v) for v in value}
-        if all(issubclass(t, numbers.Integral) for t in types):
+        if len(value) == 0 and attr_type is None:
+            raise ValueError("Could not infer attribute type from empty iterator")
```
xadupre, line 861: It would be nice to add the parameter name in the error message. It would be really helpful when writing a converter.

take-cheeze, line 861:
```suggestion
            raise ValueError(f"Could not infer attribute `{key}` type from empty iterator")
```

onnx/helper.py
```diff
@@ -854,34 +857,57 @@ def make_attribute(  # pylint: disable=too-many-statements
     # Iterable cases
     elif isinstance(value, collections.abc.Iterable):
         value = list(value)
-        types = {type(v) for v in value}
-        if all(issubclass(t, numbers.Integral) for t in types):
+        if len(value) == 0 and attr_type is None:
+            raise ValueError(f"Could not infer attribute `{key}` type from empty iterator")
```
take-cheeze, line 861:
```suggestion
            raise ValueError(
                f"Could not infer attribute `{key}` type from empty iterator"
            )
```

gramalingam (approved):

PR #5221, opened 2023-05-11
Expand ONNX schema registration/deregistration API

q-ycong-p:
### Description

- Introduces functions to deregister all schema of a domain. This clears map<OperatorSetVersion, OpSchema>, which should free up memory allocated for the schemas.

### Motivation and Context

[Pull#3266](https://github.com/onnx/onnx/pull/3266) introduced handle for runtime to specify opset to register. Similarly, it makes sense for a runtime to "offload" the schemas when it's not needed. This aims to (1) reduce runtime memory, and (2) improve the feature completeness for a runtime to control ONNX schema registration as it needs.


onnx/defs/operator_sets.h
```diff
@@ -1164,4 +1164,10 @@ inline void RegisterOnnxOperatorSetSchema(int target_version) {
   OpSchemaRegistry::Instance()->SetLoadedSchemaVersion(target_version);
 }
 
+inline void DeregisterOnnxOperatorSetSchema() {
```
q-ycong-p, line 1167: Since we're manipulating the static map `OpName_Domain_Version_Schema_Map` at runtime, how should we best guarantee thread-safety here?

q-ycong-p:
Hi @jcwchen , I've updated the commit to:
1. Allow the registrations of multiple schema versions via the existing `RegisterOnnxOperatorSetSchema` API. The behavior is as we discussed in #3266, that the final schema follows: Registration(VA, VB) == Registration(VA) U Registration(VB).
2. Keep the de-registration API `DeregisterOnnxOperatorSetSchema` to deregister all schemas, instead of allowing a specific version. This is because the later appears tricky to do right - when deregistering a version `VX`, it's not straight forward to identify which schema version should be kept/removed at ONNX level. For example, a runtime registers opset-1, which adds `RandomNormal-1` (t0), then if it registers opset-7 and immediately de-register opset-7 (t1), we'd expect the schema state now (t1) to be equivalent to t0. However I don't see a quick way for ONNX to check whether `RandomNormal-1` was brought in with opset-7 hence should be removed, or brought in by previous registrations hence should be kept. I think the current API to de-register all is the easiest, for both ONNX to implement, and for runtime to understand. Pls let me know if I understood wrong here.
3. Add unit tests to cover several API call cases from a runtime perspective.

I have several remaining questions. Would really like to get your thoughts on!
- Since the schema map is a static object, and that a runtime may run multi-threaded, do we need mutex to guard operations on the map?
- Current schema registration handles are exposed to runtimes via C++ API. How can we expose the some handles to the Python users?

onnx/defs/schema.h
```diff
@@ -1136,74 +1136,127 @@ class OpSchemaRegistry final : public ISchemaRegistry {
...
+        // Invalidates ver-schema pairs and frees memory, leaving m[op_name][op_domain] empty
+        opset_version_schema_map.clear();
+        domain_map.erase(domain);
+      }
+    }
+  }
+
   // Return the latest schema for an operator in specified domain.
   // Domain with default value ONNX_DOMAIN means ONNX.
   static const OpSchema* Schema(const std::string& key, const std::string& domain = ONNX_DOMAIN) {
```
q-ycong-p, line 1251: The changes in two `Schema` functions below is to safe-guard against checking an empty `m[op_name][domain]` map, which was not previously accounted for - it's good to check since know we're iteratively adding more handle for runtimes to manipulate this map.

onnx/defs/schema.h
```diff
@@ -1136,74 +1136,127 @@ class OpSchemaRegistry final : public ISchemaRegistry {
...
-        m[op_name][op_domain].insert(std::pair<int, OpSchema&&>(ver, std::move(op_schema)));
+        CheckDomainAndVersionToRegister(op_schema, op_name, op_domain);
+        schema_ver_map.insert(std::pair<int, OpSchema&&>(ver, std::move(op_schema)));
       }
       ONNX_CATCH(const std::exception& e) {
         ONNX_HANDLE_EXCEPTION([&]() { std::cerr << "Schema error: " << e.what() << std::endl; });
       }
     }
+
+   private:
```
q-ycong-p, line 1185: Separated out the static utility codes to simplify the `OpSchemaRegisterOnce` logic above.

q-ycong-p:
Also, I think it makes sense to have a static bool variable, e.g. `disable_onnx_static_registration` which turns true/false based on `#ifndef __ONNX_DISABLE_STATIC_REGISTRATION` during compilation. It can be exposed to runtime by an additional func like `ONNX_NAMESPACE::IsONNXStaticRegistrationDisabled()`. The motivation is that a runtime may approach the freedom of schema registration differently based on how this compilation flag was initially set. @jcwchen , what do you think?

onnx/defs/operator_sets.h
```diff
@@ -1140,28 +1140,35 @@ inline void RegisterOnnxOperatorSetSchema() {
 inline void RegisterOnnxOperatorSetSchema(int target_version) {
```
jcwchen, line 1140: Shall we also extend an option here to further control fail_duplicate_schema in case someone is still interested in having that duplicate check?

onnx/defs/schema.h
jcwchen: Great! Thank you for refactoring.

onnx/defs/operator_sets.h
q-ycong-p: Done.

q-ycong-p:
Hi @jcwchen thanks for the review. I've addressed your suggestion to expand user option. Would appreciate an approval if codes look ready. Thanks a ton!

My two remaining questions are: (1) Current schema registration handles are exposed to runtimes via C++ API. How can we expose the some handles to the Python users?
(2) Shall we add user API to reflect the compilation flag `ONNX_DISABLE_STATIC_REGISTRATION`? See explanation [here](https://github.com/onnx/onnx/pull/5221#issuecomment-1553840751).

jcwchen:
> My two remaining questions are: (1) Current schema registration handles are exposed to runtimes via C++ API. How can we expose the some handles to the Python users?
(2) Shall we add user API to reflect the compilation flag ONNX_DISABLE_STATIC_REGISTRATION? See explanation https://github.com/onnx/onnx/pull/5221#issuecomment-1553840751.
(for these two points, I can follow up in separate PR. Just wanna take your thoughts!)

1. Technically we can directly use pybind11 to expose Python API from C++ API. Here is a simple example PR: https://github.com/onnx/onnx/pull/4720/files.
2. My first thought would be will exposing `ONNX_DISABLE_STATIC_REGISTRATION` complicate runtime too much? I can imagine much different code will be used with/without  `ONNX_DISABLE_STATIC_REGISTRATION` during runtime, but if there is really a real-world use case, I am fine to have such an API.

jcwchen (commented):
Sorry I think I miss some comments while submitting comments yesterday. Please let me know if I still miss something important. Thank you!

> Since the schema map is a static object, and that a runtime may run multi-threaded, do we need mutex to guard operations on the map?

Good point. I think it's good to have from std library.

q-ycong-p:
> > My two remaining questions are: (1) Current schema registration handles are exposed to runtimes via C++ API. How can we expose the some handles to the Python users?
> > (2) Shall we add user API to reflect the compilation flag ONNX_DISABLE_STATIC_REGISTRATION? See explanation [#5221 (comment)](https://github.com/onnx/onnx/pull/5221#issuecomment-1553840751).
> > (for these two points, I can follow up in separate PR. Just wanna take your thoughts!)
> 
> 1. Technically we can directly use pybind11 to expose Python API from C++ API. Here is a simple example PR: https://github.com/onnx/onnx/pull/4720/files.
> 2. My first thought would be will exposing `ONNX_DISABLE_STATIC_REGISTRATION` complicate runtime too much? I can imagine much different code will be used with/without  `ONNX_DISABLE_STATIC_REGISTRATION` during runtime, but if there is really a real-world use case for dynamic setting `ONNX_DISABLE_STATIC_REGISTRATION`, I am fine to have such an API.

1. Makes sense. I plan to follow up a separate PR to expose it to Python API once current feature PR is merged.
2. Makes sense. My concern originated from that the `map()` func that get called at multiple places (e.g. in the getter `Schema(...)`) declares static `SchemaRegisterer` [here](https://github.com/onnx/onnx/blob/main/onnx/defs/schema.cc#L1109) based on `__ONNX_DISABLE_STATIC_REGISTRATION`'s status. This makes me think, when this cmake args is `OFF`, a runtime should probably not mess with schema registration at all. But I experimented with a unit test, that even when this cmake arg is `OFF`, it can deregister, query, and register schema without issue. Pls correct me if I'm missing something here. I'll add this UT in PR.

Found another issue, `loaded_schema_version` [here](https://github.com/onnx/onnx/blob/main/onnx/defs/schema.cc#L25) should probably be assigned based on `__ONNX_DISABLE_STATIC_REGISTRATION`, no? Like:
```
#ifdef __ONNX_DISABLE_STATIC_REGISTRATION
  int OpSchemaRegistry::loaded_schema_version = -1;
#else
  int OpSchemaRegistry::loaded_schema_version = 0;
#endif
```

On thread-safety, I plan to add `std::unique_lock` to guard the map's read-write operation. I will fix the lint.

q-ycong-p:
> Sorry I think I miss some comments while submitting comments yesterday. Please let me know if I still miss something important. Thank you!
> 
> > Since the schema map is a static object, and that a runtime may run multi-threaded, do we need mutex to guard operations on the map?
> 
> Good point. I think it's good to have from std library.

I thought of having a `OpSchemaRegistry`-level mutex, which guards the critical sections in `OpSchemaRegisterOnce` and `OpSchemaDeregisterAll` that read/write the static OpSchema map. But `OpSchemaRegisterOnce` is a class inside `OpSchemaRegistry`. I am a bit confused what's the right mutex impl here. Could you shed some light? We can also merge the current feature PR and iterate with separate PR on the good-to-have features: (1) thread-safety (2) expose to Python API.

onnx/defs/schema.h
```diff
@@ -1136,74 +1136,128 @@ class OpSchemaRegistry final : public ISchemaRegistry {
...
+            << "in onnx/defs/schema.h)." << std::endl;
+        fail_schema(err.str());
+      }
+    }
   };
 
+  // Deregister all ONNX opset schemas from domain
+  // Domain with default value ONNX_DOMAIN means ONNX.
+  static void OpSchemaDeregisterAll(const std::string& domain = ONNX_DOMAIN) {
+    auto& m = GetMapWithoutEnsuringRegistration();
```
justinchuby, line 1236: Nit: (I know we haven’t been perfect in naming things but we can start giving them more thought with new code) readability: I would consider giving variables more meaningful names (m here and m_p in the loop below) and avoid abbreviations - they add cognitive load to readers.

> Do not use abbreviations that are ambiguous or unfamiliar to readers outside the project, and do not abbreviate by deleting letters within a word.

https://google.github.io/styleguide/pyguide.html#316-naming
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

onnx/defs/schema.cc
```diff
@@ -22,14 +22,18 @@ namespace ONNX_NAMESPACE {
 // -1 means ONNX schema hasn't been loaded yet
 // 0 means all versions of ONNX schema have been loaded
 // Other positive integer means the ONNX schemas for the specified version have been loaded
+#ifdef __ONNX_DISABLE_STATIC_REGISTRATION
```
justinchuby, line 25: Do we need documentation for this flag?

q-ycong-p: Currently it's a cmake arg [here](https://github.com/onnx/onnx/blob/main/CMakeLists.txt#L28) with a brief explanation. Where is the suitable place to add documentation?

onnx/defs/schema.h
q-ycong-p: Used `schema_map` and `schema_map_pair` respectively to be more descriptive than `m` and `m_p`. Pls suggest if there's clearer naming option tho!

onnx/defs/schema.cc
q-ycong-p: I don't find a top-level `.md` to add documentation on this cmake arg, apart from the description in `CMakeLists.txt`. Pls let me know where you'd like it added. Beside it looks like this cmake arg is not in `setup.py`, so I added it there in the new commit.

Let me know what else i can address. Ty!

justinchuby: Thanks for making the updates. I’m not sure either. Will leave it for others to comment

justinchuby: Why do we need this flag in the first place?

q-ycong-p: @justinchuby This flag was introduced in #3266, and explained there. 
By default ONNX loads opset schema of all versions. #3266 introduced this compile-time flag to not register any schema by default, and leave the runtime to register specific version of opset schema as needed.

This current PR expands the API on that, so that (1) a runtime can now de-register schema when no longer need, and (2) a runtime will not incur error when registering multiple versions (which can encounter duplicate schema request).

The motivation is to reduce MaxRSS in runtime.

onnx/defs/schema.h
```diff
@@ -1136,74 +1136,128 @@ class OpSchemaRegistry final : public ISchemaRegistry {
 
   class OpSchemaRegisterOnce final {
    public:
-    OpSchemaRegisterOnce(OpSchema& op_schema, int opset_version_to_load = 0) {
+    OpSchemaRegisterOnce(OpSchema& op_schema, int opset_version_to_load = 0, bool fail_duplicate_schema = true) {
```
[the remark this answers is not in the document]
q-ycong-p: Not allow duplicate schema was the previous behavior in `OpSchemaRegisterOnce `. I don't know why it was so - I'd appreciate ONNX dev to shed light on the reason here. Therefore, I left `fail_duplicate_schema ` default to `True` for backward-compatibility of this API.

As far as this PR attempts to do: it wants to allow a runtime to register specific versions of opset schema without running into error. Consider `RandomNormal` which only exists for version-1. A runtime may register opset-7, then opset-11 as needed. A runtime would expect the second attempt not to fail just because `RandomNormal-1` already exists. Hence from a runtime perspective, as the caller of `RegisterOnnxOperatorSetSchema`, it's natural to assume not to fail on request to register duplicate schema.

When `fail_duplicate_schema=False`, it does not really register the same schema twice. It just won't throw error, and will not register that schema.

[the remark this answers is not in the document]
q-ycong-p: So overall, this inconsistency is from the natural expectation of the intended caller, and to maintain backward-compatibility of the existing `OpSchemaRegisterOnce`'s interface.

I can change the runtime-level handle `RegisterOnnxOperatorSetSchema` to default on `fail_duplicate_schema=True` to be consistent. But to me, such a API is more confusing for a runtime to use. This is because, for a runtime who doesn't know of ONNX implementation, it may not expect registering opset-7 then opset-11 to throw duplicate schema error.

Shall we make `OpSchemaRegisterOnce` to allow duplicate schema registration request by default? To me (without ONNX expertise), it seems ok to make it a default behavior to allow such request to return **without throwing an error** - what's important is that we aren't actually loading two copies of the same schema here in either case.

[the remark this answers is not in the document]
q-ycong-p: What is the preferred defaults here? Would like to hear your thoughts @justinchuby @jcwchen. Ty!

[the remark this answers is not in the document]
jcwchen: IMO it's better to keep backward-compatibility for RegisterOnnxOperatorSetSchema as well with default fail_duplicate_schema = true, which is also consistent with all ONNX APIs to prevent confusions.

onnx/defs/schema.cc
```diff
@@ -22,14 +22,18 @@ namespace ONNX_NAMESPACE {
...
 int OpSchemaRegistry::loaded_schema_version = -1;
+#else
+int OpSchemaRegistry::loaded_schema_version = 0;
```
jcwchen, line 28: > Found another issue, loaded_schema_version [here](https://github.com/onnx/onnx/blob/main/onnx/defs/schema.cc#L25) should probably be assigned based on __ONNX_DISABLE_STATIC_REGISTRATION, no?

https://github.com/onnx/onnx/blob/9184c2989b2fb3e555680985c3c3a5aa19555454/onnx/defs/operator_sets.h#L1137 loaded_schema_version will be set after `RegisterOnnxOperatorSetSchema` and initially it will be -1, which means onnx hasn't registered any opset yet. Not sure why setting 0 here?

jcwchen (commented):
> My concern originated from that the map() func that get called at multiple places (e.g. in the getter Schema(...)) declares static SchemaRegisterer [here](https://github.com/onnx/onnx/blob/main/onnx/defs/schema.cc#L1109) based on __ONNX_DISABLE_STATIC_REGISTRATION's status. This makes me think, when this cmake args is OFF, a runtime should probably not mess with schema registration at all. But I experimented with a unit test, that even when this cmake arg is OFF, it can deregister, query, and register schema without issue. Pls correct me if I'm missing something here. I'll add this UT in PR.

Thank you for elaboration! I got your point now. I think that would be helpful for runtime and so it's good to have.

onnx/defs/schema.cc
q-ycong-p: This ties to the intention [here](https://github.com/onnx/onnx/pull/5221#pullrequestreview-1442771211), where a runtime might, at some point, use `GetLoadedSchemaVersion` to understand what ONNX schemas are loaded. In the case where all schema is registered at start (by default), the call would otherwise returns `-1` to runtime, which is a wrong information suggesting nothing is loaded.

onnx/defs/schema.h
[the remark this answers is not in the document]
q-ycong-p: Understood, will use default `fail_duplicate_schema = true` in both APIs, `OpSchemaRegisterOnce` and `RegisterOnnxOperatorSetSchema`.

q-ycong-p:
Looks like the release build's ORT tests failed (not sure why it didn't run before) on "Verify ONNX with ONNX Runtime PyPi package". The failed tests were due to ONNX schema not registered - the ONNX build must have turned on `-DONNX_DISABLE_STATIC_REGISTRATION`. I took a look but not fully understanding the CI pipeline here. Any thought?

Do we expect the release build to permute to all available cmake bool flags? Or else how did it got turned on in the release tests?

onnx/defs/schema.h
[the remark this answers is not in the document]
q-ycong-p: Done

onnx/defs/schema.cc
q-ycong-p: I didn't change this line in [641cd5a](https://github.com/onnx/onnx/pull/5221/commits/641cd5a5e8b43fd4a29f741a00a1790166ecc6ca) as explained above. Let me know what you think. If I missed something or it doesn't make sense, I'll make change.

onnx/defs/operator_sets.h
```diff
@@ -1137,31 +1137,47 @@ inline void RegisterOnnxOperatorSetSchema() {
...
   OpSchemaRegistry::Instance()->SetLoadedSchemaVersion(target_version);
 }
 
+inline void DeregisterOnnxOperatorSetSchema() {
+  OpSchemaRegistry::Instance()->OpSchemaDeregisterAll(ONNX_DOMAIN);
+  // -1 means no ONNX schema is loaded
+  OpSchemaRegistry::Instance()->SetLoadedSchemaVersion(-1);
+}
+
+inline bool IsOnnxStaticRegistrationDisabled() {
```
q-ycong-p, line 1175: @jcwchen It's seems a working solution but maybe there's more correct way to expose this flag to runtime. If so pls let me know!

onnx/defs/schema.cc
jcwchen: Not quite follow here -- let me organize the flows for these two options:
- Turn on ONNX_DISABLE_STATIC_REGISTRATION: runtime might register onnx schema for certain opset_version multiple times
  - Initially loaded_schema_version=-1; which means no schema has been registered yet
  - runtime will further call `RegisterOnnxOperatorSetSchema` with certain opset_version to register specified opset_version (and `loaded_schema_version` will be set as that specified version).
  - (Optional) runtime may call `DeregisterOnnxOperatorSetSchema` to clear all registration (and `loaded_schema_version` will be set as -1) and then call `RegisterOnnxOperatorSetSchema` with another opset_version  (and `loaded_schema_version` will be changed accordingly).
- Turn off ONNX_DISABLE_STATIC_REGISTRATION: runtime will just register onnx schema once
  - Initially loaded_schema_version=-1; which means no schema has been registered yet
  - runtime will further call `RegisterOnnxOperatorSetSchema` to register all onnx schema and then `loaded_schema_version` will be set as 0.

Thus, I think anyway OpSchemaRegistry::loaded_schema_version will initially declared as -1 by default, which lets onnxruntime know there is no schema has been registered yet. Did I missing something obvious?

jcwchen (commented):
> Looks like the release build's ORT tests failed (not sure why it didn't run before) on "Verify ONNX with ONNX Runtime PyPi package". The failed tests were due to ONNX schema not registered - the ONNX build must have turned on -DONNX_DISABLE_STATIC_REGISTRATION. I took a look but not fully understanding the CI pipeline here. Any thought?

Those release CIs will be triggered only when having run release CIs (I added it recently because this PR is important and I would like to test it in advance). IIUC, this PR didn't change the default behavior for schema registration (ONNX_DISABLE_STATIC_REGISTRATION=OFF). I guess the failure might be irrelevant to this PR. Please let me merge main branch with this PR to see whether it can help resolve those failures.

jcwchen:
> Those release CIs will be triggered only when having run release CIs (I added it recently because this PR is important and I would like to test it in advance). IIUC, this PR didn't change the default behavior for schema registration (ONNX_DISABLE_STATIC_REGISTRATION=OFF and it will register all opset schema). I guess the failure might be irrelevant to this PR. Please let me merge main branch with this PR to see whether it can help resolve those failures.

Such an error happened in other PRs as well. Probably because onnxruntime just has a fresh release (1.15.0) recently. We will have a PR to fix these failures and please just ignore those failures in release CIs for now. Thanks!

onnx/defs/schema.cc
q-ycong-p: Understood - agree. Though in the "turned off" case, runtime won't normally explicitly call `RegisterOnnxOperatorSetSchema`, right? Updated PR to bring it back to the correct `-1` value.

q-ycong-p:
> > Those release CIs will be triggered only when having run release CIs (I added it recently because this PR is important and I would like to test it in advance). IIUC, this PR didn't change the default behavior for schema registration (ONNX_DISABLE_STATIC_REGISTRATION=OFF and it will register all opset schema). I guess the failure might be irrelevant to this PR. Please let me merge main branch with this PR to see whether it can help resolve those failures.
> 
> Such an error happened in other PRs as well. Probably because onnxruntime just has a fresh release (1.15.0) recently. We will have a PR to fix these failures and please just ignore those failures in release CIs for now. Thanks!
> 
> (Update) The PR to resolve those failures is here: #5261.

Yes, the Windows-CI failures log complains on the tests that involve the missing ops #5261 adds. I am not sure whats' the common practice here. Do we wait on #5261 to merge, then update this one to run release tests before move forward with review? Otherwise I think all comments are addressed and it's in state for review/merge. Thanks!

jcwchen (approved):
Now it passed all CIs. It will still need approval from @onnx/sig-operators-approvers. It might take some time because some of them are OOF this week. Thank you for waiting.

gramalingam (approved):

gramalingam (approved):

q-ycong-p:
Hi @jcwchen , thank you very much for working with me on this. I followed up a separate PR #5291 aiming to address schema map thread-safety based on our discussion above. Would appreciate your suggestion if I miss something there - Ty!

onnx/defs/schema.h
```diff
@@ -1136,74 +1136,131 @@ class OpSchemaRegistry final : public ISchemaRegistry {
 
   class OpSchemaRegisterOnce final {
    public:
-    OpSchemaRegisterOnce(OpSchema& op_schema, int opset_version_to_load = 0) {
+    explicit OpSchemaRegisterOnce(
```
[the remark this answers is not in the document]
q-ycong-p: Hi @jcwchen sorry for late reply on this thread, I saw #5322 and I don't have any objection against removing the `explicit`.

The original reason for adding it, was that the ctor has two args `int` and `bool` with default values. I saw a style-check suggestion that in such a case, the ctor should be made `explicit` as a best practice. My understanding of the ORT [error](https://github.com/microsoft/onnxruntime/pull/16325#issuecomment-1587836118) is that [here](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/graph/contrib_ops/contrib_defs.h#L47-L50) requires `OpSchemaRegisterOnce` to allow implicit conversion?

In any case, pls feel free to remove `explicit`. Thanks for addressing it for ORT backward-comp.

setup.py
```diff
@@ -49,6 +49,9 @@
 ONNX_NAMESPACE = os.getenv("ONNX_NAMESPACE", "onnx")
 ONNX_BUILD_TESTS = bool(os.getenv("ONNX_BUILD_TESTS") == "1")
 ONNX_DISABLE_EXCEPTIONS = bool(os.getenv("ONNX_DISABLE_EXCEPTIONS") == "1")
+ONNX_DISABLE_STATIC_REGISTRATION = bool(
```
xadupre, line 52: Is there a place where these options are documented?

jcwchen: I am not aware of any, although these options are identical as the options in CMakeLists.txt: https://github.com/onnx/onnx/blob/391f5703abe1a0d5c8cbb0b38df079b0bacec079/CMakeLists.txt#L25-L28

which has a simple sentence as description.

Issue #5212, opened 2023-05-08
Segfault running shape inference with `data_prop=True` on model with nested function.


BowenBao:
# Bug Report

### Is the issue related to model conversion?

No, checker passed. inliner passed. 
Same shape inference call passed on inlined model.

### Describe the bug

As title

### System information

ONNX version: 1.15 dbeee2465b8628ddf742e8ff52144b424f3c45db

### Reproduction instructions
Model can be downloaded [here](https://microsoft-my.sharepoint.com/:u:/p/bowbao/EZVQDNP6-wpBqmls_ScqhgYB9iQJe_P0vbsZct55Gtw7MQ?e=9MBbVO).
```
import onnx
import onnx.shape_inference
import onnx.inliner

model = onnx.load("maf_mnist.onnx")

onnx.checker.check_model(model)
print("Pass checker!")

# Works fine
onnx.shape_inference.infer_shapes(model, data_prop=False)
print("Success with data_prop=False !")

inlined_model = onnx.inliner.inline_local_functions(model)
print("Success with inlining!")
onnx.shape_inference.infer_shapes(inlined_model, data_prop=True)
print("Success with data_prop=True on inlined model !")

# Segfaults
onnx.shape_inference.infer_shapes(model, data_prop=True)

```

jcwchen:
Hi @BowenBao,
Thanks for report. After several experiments, segfault starts to happen after this PR: https://github.com/onnx/onnx/pull/5066. Will investigate more.

BowenBao:
Got another repro for [gpt2 tiny](https://microsoft-my.sharepoint.com/:u:/p/bowbao/EZMzLywywzNKikgZWwJ7I24Bp5L28aDU9WHwdCKXiJ9j-Q?e=zzhbji)

This one segfaults with just shape inference
```
import onnx
import onnx.shape_inference
import onnx.inliner
import onnxscript

model = onnx.load("maf_gpt2_tiny.onnx")
onnx.shape_inference.infer_shapes(model)
print("Pass shape inference!")
```

It doesn't pass checker though, but I think it's because we weren't able to use shape_inference to fill in some missing output shapes at `onnxscript/function_libs/torch_lib/graph_building.py`.

gramalingam:
May be worth checking the model-generation part. I tried the gpt2 tiny example above. I see it has functions like "aten_rsub" which uses an attribute-parameter alpha, but the function's attribute list is empty.

gramalingam:
[...]

PR #5223, opened 2023-05-12
Fix a couple of shape inference issues

gramalingam:
### Description
Fixes a couple of issues revealed by [Issue 5212]( https://github.com/onnx/onnx/issues/5212).

(i) The data-propagation logic was not handling function-calls correctly. It was reusing the same map (for tracking values) in the caller and callee. We need to create a new map for the callee, and propagate values corresponding to the inputs (and outputs) to (and back from) the callee map.

(ii) Recently the proto definition was extended to specify default-values of attributes in FunctionProto. This extends the inference logic to account for it. (Note that the checker may require further extensions to handle this extension, that's to be done separately.)

(iii) Add test-cases to test the fixes.

### Motivation and Context
Fixes a couple of issues revealed by [Issue 5212]( https://github.com/onnx/onnx/issues/5212)


onnx/defs/generator/old.cc
```diff
@@ -12,6 +12,116 @@ This operator produces a constant tensor. Exactly one of the provided attributes
 or value_* must be specified.
 )DOC";
 
+void ConstantInference(InferenceContext& ctx) {
```
gramalingam, line 15: Note: No change to this logic in the file in this PR. It just moves the type-and-shape-inference function for the Constant op into a separately defined function. It was hard to set breakpoints inside this inference function and step through it in a debugger when it is included in a macro (as before).

onnx/shape_inference/implementation.cc
```diff
@@ -345,6 +346,56 @@ class ShapeInferenceImplBase {
...
+        auto it = caller_map.find(actual);
+        if (it != caller_map.end()) {
+          callee_map[formal] = it->second;
+        }
+      }
+    }
+  }
+
+  // Update a DataValueMap for a calling function from the DataValueMap of the callee
+  void bindValuesOnReturn(
```
xadupre, line 369: The code is very similar to the previous one, maybe that would be worth merging hem.

gramalingam: Yes, they are similar. But hard to merge them. One focuses on inputs, the other on outputs. One copies from caller to callee, the other from callee to caller. But, in either case, we need to check caller-side input/output to see if it is empty. We will need to add so many extra-parameters (to merge them) that understanding the logic will be harder. I think it is better in its current form.

xadupre: This is what I had in mind. I probably missed something.

```c++
  void bindValuesOnCallOrReturn(
      const DataValueMap& caller_map,
      const NodeProto& caller,
      bool io,
      DataValueMap& callee_map,
      const FunctionProto& callee) {
    int num_io = io 
      ? (std::min)(caller.input_size(), callee.input_size())
      : (std::min)(caller.output_size(), callee.output_size());
    for (int i = 0; i < num_io; ++i) {
      const std::string& actual = io ? caller.input(i) : caller.output(i);
      if (!actual.empty()) {
        const std::string& formal = io ? callee.input(i) : callee.output(i);
        auto it = caller_map.find(actual);
        if (it != caller_map.end()) {
          callee_map[formal] = it->second;
        }
      }
    }
  }

  // Initialize a DataValueMap for a called function from the DataValueMap of the caller
  void bindValuesOnCall(
      const DataValueMap& caller_map,
      const NodeProto& caller,
      DataValueMap& callee_map,
      const FunctionProto& callee) {
    bindValuesOnCallOrReturn(caller_map, caller, true, callee_map, callee);
  }

  // Update a DataValueMap for a calling function from the DataValueMap of the callee
  void bindValuesOnReturn(
      const DataValueMap& callee_map,
      const FunctionProto& callee,
      DataValueMap& caller_map,
      const NodeProto& caller) {
    bindValuesOnCallOrReturn(callee_map, caller, false, caller_map, callee);
  }
```

gramalingam: In the innermost 3 lines, you will also need to change `caller_map` to `io ? caller_map : callee_map`, and similarly for `callee_map`. I prefer to keep the current form, as it is more understandable and is easier to verify is correct when we examine the code.

gramalingam: And similarly we need to do the same for `actual` and `formal` in the innermost 3 lines.

xadupre (approved):

onnx/defs/generator/old.cc
jcwchen: Not relevant to this PR: might be worthy to make ConstantInference as a common function with options to handle different versions of Constant. Just like https://github.com/onnx/onnx/blob/a1e97a1ff257c7e3b1ecd62dc4872fa79059a275/onnx/defs/tensor/utils.cc#L150

jcwchen (approved):
LGTM. Thank you for fixing this! Also confirmed this PR can help to resolve segfault issue for maf_gpt2_tiny.onnx from https://github.com/onnx/onnx/issues/5212.

Issue #4983, opened 2023-03-13
Whether support will be provided in the future about warp_grid and grid_sample openrates of torch?


wangzhenlin123:
# Ask a Question
Whether support will be provided in the future about warp_grid and grid_sample openrates of torch?
### Question

RuntimeError: Exporting the operator affine_grid_generator to ONNX opset version 13 is not supported. Please feel free to request support or submit a pull request on PyTorch GitHub.

### Further information
- Relevant Area: 

- Is this issue related to a specific model?
**Model name**: 
**Model opset**:

justinchuby:
I think grid sample is supported at Opset 16 IIRC. This would be a good question in the PyTorch repo.

wangzhenlin123:
> So does that support  warp_affine operator in onnx?

liqunfu:
affine_grid_generator is missing and shall be added to onnx - to add AffineGrid operator ('aten::affine_grid_generator'):
There is a GridSample operator in onnx:
https://github.com/onnx/onnx/blob/main/docs/Operators.md#gridsample

GridSample takes 'grid' as input. 'grid' is usually generated with affine_grid:
https://pytorch.org/docs/stable/generated/torch.nn.functional.affine_grid.html
As it stated: "This function is often used in conjunction with grid_sample() to build Spatial Transformer Networks"

Without grid generator like affine_grid, GridSample is almost not useful.

PR #5225, opened 2023-05-13
add affine_grid op

liqunfu:
### Description
[](https://github.com/onnx/onnx/issues/4983)

### Motivation and Context

https://github.com/onnx/onnx/issues/4983


onnx/defs/tensor/defs.cc
```diff
@@ -2482,6 +2482,197 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+        .Attr(
+            "align_corners",
+            "if align_corners=1, consider -1 and 1 to refer to the centers of the corner pixels. "
+            "if align_corners=0, consider -1 and 1 to refer to the outer edge the corner pixels.",
+            AttributeProto::INT,
+            static_cast<int64_t>(0))
+        .Input(
+            0,
+            "theta",
+            "input batch of affine matrices with shape (N, 2, 3) for 2D or (N, 3, 4) for 3D",
```
gramalingam, line 2501: Some more comment to explain what the 3 (or 4) values refer to, in the last dimension.

onnx/test/parser_test.py
```diff
@@ -119,6 +119,205 @@ def test_parse_function_with_attributes(self) -> None:
         model = onnx.parser.parse_model(input)
         checker.check_model(model)
 
+    def test_parse_afline_grid(self) -> None:
```
gramalingam, line 122: Seems better to move this test to a different file, like `test_affine_grid_function.py`, it is not a parser test, it just uses the parser.

onnx/test/parser_test.py
```diff
@@ -119,6 +119,205 @@ def test_parse_function_with_attributes(self) -> None:
...
+                        grid_N_DHW_3 = Transpose <perm = [0, 2, 1]> (grid_N_3_DHW)
+                        N_D_H_W_3_seq = SequenceConstruct (N, D, H, W, int_three_1d)
+                        N_D_H_W_3 = ConcatFromSequence <axis: int=-1, new_axis: int=0> (N_D_H_W_3_seq)
+                        grid_3d_else = Reshape(grid_N_DHW_3, N_D_H_W_3)
+                        }
+                    >
+           }
+        """
+
+        model = onnx.parser.parse_model(input)
```
gramalingam, line 286: Instead of duplicating the function-definition, would it be better to get the function-definition from the op-schema, and test it?

onnx/defs/tensor/defs.cc
```diff
@@ -2482,6 +2482,197 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "grid",
+            "output tensor of shape (N, C, H, W, 2) of 2D sample coordinates or (N, C, D, H, W, 3) of 3D sample coordinates.",
+            "T1",
+            OpSchema::Single,
+            true,
+            1,
+            OpSchema::Differentiable)
+        .TypeConstraint(
+            "T1",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
```
gramalingam, line 2527: Why not bfloat16 also? May be better to define and use some set like `all_float_types_ir4` (like [here](https://github.com/onnx/onnx/blob/a1e97a1ff257c7e3b1ecd62dc4872fa79059a275/onnx/defs/schema.h#L579) ... that will make ops follow similar pattern.

onnx/defs/tensor/defs.cc
```diff
@@ -2482,6 +2482,197 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                      then_branch = h1 () => (float[H] grid_h_then, float[W] grid_w_then) {
+                          step_h = Div (two, H_float)
+                          step_w = Div (two, W_float)
+                          step_h_half = Div (step_h, two)
+                          start_h = Add (minus_one, step_h_half)
+                          grid_h_then = Range (start_h, one, step_h)
+
+                          step_w_half = Div (step_w, two)
+                          start_w = Add (minus_one, step_w_half)
+                          grid_w_then = Range (start_w, one, step_w)
```
gramalingam, line 2567: I am a bit concerned about the use of Range with float types. Couldn't numeric precision affect whether the number of elements generated is off-by-one ? Would it be better to use an integer-based range that is converted into a float range?

xadupre: step_w does not seem to be an integer.

onnx/test/parser_test.py
```diff
@@ -119,6 +119,205 @@ def test_parse_function_with_attributes(self) -> None:
...
+        checker.check_model(model)
+        onnx.save(model, "C:/Temp/affine_grid_test.onnx")
+
+        import torch
+        from torch.nn.functional import affine_grid
```
xadupre, line 291: It should be better to use ReferenceEvaluator and froze the expected outputs instead of using torch.

onnx/test/parser_test.py
```diff
@@ -119,6 +119,205 @@ def test_parse_function_with_attributes(self) -> None:
...
+                                W_float_nimus_one = Sub (W_float, one)
+                                step_h = Div (two, H_float_nimus_one)
+                                step_w = Div (two, W_float_nimus_one)
+                                epsilon = Constant <value = float {1e-6}>()
+                                one_plus_epsilon = Add (one, epsilon)
+                                grid_h_else = Range (minus_one, one_plus_epsilon, step_h)
+                                grid_w_else = Range (minus_one, one_plus_epsilon, step_w)
+                            }
+                        >
+                        size_ones_H_by_W_transpose = Transpose (ones_H_by_W) # (3, 2)
```
gramalingam, line 193: Doesn't this just produce `one_W_by_H` ? What does the `(3, 2)` mean?

onnx/defs/tensor/defs.cc
gramalingam: Yes, but we know that we want to create a vector of size W. It seems better to use `Range(0, W, 1)` and map each i in this range to a corresponding coordinate-value using arithmetic.

gramalingam: Basically, something like
```
   i= Range (zero_int, W, one_int)
   i_step_w = Mul (i, step_w)
   grid_w_then = Add (start_w, i_step_w)
```

gramalingam: That may help avoid all the use of `epsilon` in the else-branch. Further, the above fragment is common across the if-then-else, only start_w and step_w depend on the flag.

onnx/defs/tensor/defs.cc
liqunfu: comments with details added

onnx/test/parser_test.py
liqunfu: this test is for this PR development only - quick validation of the function body. Once all tests are passing, I will remove this test. the test body in this test will be moved to tensor/def.cc.

onnx/test/parser_test.py
liqunfu: Good point. Only one test body for AffineGrid shall exist which is in the op-schema.

onnx/test/parser_test.py
liqunfu: I will remove changes in this file. It is used for this PR's development purpose so that I can quickly validate the function body.

onnx/test/parser_test.py
liqunfu: the comment (# (3, 2)) shall be removed. ones_W_by_H is a much better name.

onnx/test/parser_test.py
```diff
@@ -119,6 +119,205 @@ def test_parse_function_with_attributes(self) -> None:
...
+                        grid_h_1 = Mul(size_ones_H_by_W_transpose, grid_h_0)
+                        grid_h = Transpose (grid_h_1)
```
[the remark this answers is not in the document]
liqunfu: You are right! add zeros is more efficient and mul by ones.

onnx/defs/tensor/defs.cc
liqunfu: Great suggestion @gramalingam !, I will do the change accordingly.

onnx/defs/tensor/defs.cc
```diff
@@ -2482,6 +2482,197 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            "T1",
+            OpSchema::Single,
+            true,
+            1,
+            OpSchema::Differentiable)
+        .TypeConstraint(
+            "T1",
+            {"tensor(float16)", "tensor(float)", "tensor(double)"},
+            "Constrain grid types to float tensors.")
+        .TypeConstraint("T2", {"tensor(int32)", "tensor(int64)"}, "Constrain indices to integer types")
```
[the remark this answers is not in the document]
liqunfu: good point. Before doing it, shall we include all ints (int64, 32, 16, 8, uint64, 32, 16, 8)? On the other hand, Shape in ONNX is always int64. I probably need to use int64 only here because input "size" is actually shape of the output.

onnx/reference/ops/_op_list.py
```diff
@@ -31,6 +31,7 @@
 from onnx.reference.ops.op_acos import Acos
 from onnx.reference.ops.op_acosh import Acosh
 from onnx.reference.ops.op_add import Add
+from onnx.reference.ops.op_affine_grid import AffineGrid
```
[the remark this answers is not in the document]
liqunfu: this is an expected warning.

onnx/defs/schema.h
```diff
@@ -739,6 +739,19 @@ class OpSchema final {
     return all_tensor_types_ir4;
   }
 
+  static const std::vector<std::string>& all_float_types_ir4() {
```
gramalingam, line 742: We don't yet have any computational op supporting float8. I think it is better if affine_grid restricts it to float16 and above.  May be we should define a set `float_16plus_types_ir4` ? Of course, we may need to check whether the function also works for all these types (in case some op doesn't support them all.)

onnx/reference/ops/op_affine_grid.py
```diff
@@ -0,0 +1,90 @@
...
+        )
+        grid_n = np.matmul(
+            theta_n, original_grid_transposed
+        )  # shape (N, dim_3d, D * H * W)
+        # transpose to (N, D * H * W, dim_3d) and then reshape to (N, D, H, W, dim_3d)
+        grid = np.reshape(np.transpose(grid_n, (0, 2, 1)), (N, D, H, W, dim_3d))
+        return grid
+
+
+class AffineGrid(OpRun):
```
gramalingam, line 84: Question: given that affine-grid is a function-op, should we just rely on that for the implementation? or, have a separate implementation? 

What are the pros/cons? Using the function, eliminates duplicate definition, and will test it better. Just thinking aloud, since we will face a similar question for other complex function-ops too ... may be good to have a general guideline.

onnx/defs/tensor/defs.cc
```diff
@@ -2470,6 +2470,302 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+            fail_shape_inference("Length of input 'size' is ", size_length, ". It must be 4 for 2D or 5 for 5D.");
+          }
+
+          auto add_and_set_dim = [](auto& dim, auto* output_shape) {
+            auto* output_dim = output_shape->add_dim();
+            if (dim.has_dim_value()) {
+              output_dim->set_dim_value(dim.dim_value());
+            } else {
+              output_dim->set_dim_param(dim.dim_param());
+            }
```
gramalingam, line 2742: I think the above can be shortened to:
```cpp
*output_shape->add_dim() = dim;
```

onnx/defs/tensor/defs.cc
```diff
@@ -2470,6 +2470,302 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          propagateElemTypeFromInputToOutput(ctx, 0, 0);
+          if (!hasNInputShapes(ctx, 1)) {
+            return;
+          }
+
+          const auto& size_shape = ctx.getInputType(1)->tensor_type().shape();
+          const auto size_rank = size_shape.dim_size();
+          if (size_rank != 1) {
+            fail_shape_inference("Rank of input 'size' is ", size_rank, ". It must be 1.");
+          }
```
gramalingam, line 2715: Can replace above 5 lines with `checkInputRank(ctx, 1, 1)`.

onnx/defs/tensor/defs.cc
```diff
@@ -2470,6 +2470,302 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+          if (size_initializer) {
+            auto sizeData = ParseData<int64_t>(size_initializer);
+            for (auto val : sizeData) {
+              size_proto.add_dim()->set_dim_value(val);
+            }
+          } else if (size_input) {
+            size_proto.CopyFrom(*size_input);
+          } else {
+            return;
+          }
```
gramalingam, line 2729: I think you may be able to reuse the function [getInputShape](https://github.com/onnx/onnx/blob/efa11e6461960825ae5fc23ce48893ac25853bad/onnx/defs/shape_inference.h#L535) for this?

gramalingam: Sorry, I attached the correct link above, but the name of the function is `getShapeInput` (which is different from `getInputShape`). This is used for any input that represents a shape. It tries to get the information from `getInputData` and `getSymbolicInput` etc. as done above). See [this implementation](https://github.com/onnx/onnx/blob/391f5703abe1a0d5c8cbb0b38df079b0bacec079/onnx/defs/shape_inference.cc#L449)

onnx/defs/tensor/defs.cc
```diff
@@ -2461,6 +2461,302 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                      then_branch = h1 () => (start_h_then, step_h_then, start_w_then, step_w_then) { # => (float, float, float, float)
+                          step_h_then = Div (two, H_float)
+                          step_w_then = Div (two, W_float)
+                          step_h_half = Div (step_h_then, two)
+                          start_h_then = Add (minus_one, step_h_half)
+                          step_w_half = Div (step_w_then, two)
+                          start_w_then = Add (minus_one, step_w_half)
+                      },
+                      else_branch = h2 () => (start_h_else, step_h_else, start_w_else, step_w_else) { # => (float, float, float, float)
+                          H_float_nimus_one = Sub (H_float, one)
```
gramalingam, line 2571: nit: spelling "nimus" => "minus"

docs/Operators.md
```diff
@@ -485,6 +486,172 @@ expect(node, inputs=[x, y], outputs=[x + y], name="test_add_uint8")
...
+  Generates a 2D or 3D flow field (sampling grid), given a batch of affine matrices theta
+  (https://pytorch.org/docs/stable/generated/torch.nn.functional.affine_grid.html).
+  An affine matrix `theta` is applied to a position tensor represented in its homogeneous expression. Here is an example in 3D:
+  ```
+  [r00, r01, r02, t0]   [x]   [x']
+  [r10, r11, r12, t1] * [y] = [y']
+  [r20, r21, r22, t2]   [z]   [z']
+  [0,   0,   0,   1 ]   [1]   [1 ]
+  ```
+  where (x, y, z) is the position in the original space, (x', y', z') is the position in the output space.
```
justinchuby, line 500:
```suggestion
  where `(x, y, z)` is the position in the original space, `(x', y', z')` is the position in the output space.
```

style all of these and below as code for readability?

docs/Operators.md
```diff
@@ -485,6 +486,172 @@ expect(node, inputs=[x, y], outputs=[x + y], name="test_add_uint8")
...
+  The last row is always [0, 0, 0, 1] and is not stored in the affine matrix. Therefore we have `theta` of shape (N, 2, 3) for 2D or (N, 3, 4) for 3D.
+
+  Input `size` is used to define grid of positions evenly spaced in the original 2D or 3D space, with dimensions ranging from -1 to 1.
+  The output `grid` contains positions in the output space.
+
+  When `align_corners`=1, consider -1 and 1 to refer to the centers of the corner pixels (mark v in illustration).
```
justinchuby, line 506:
```suggestion
  When `align_corners=1`, consider `-1` and `1` to refer to the centers of the corner pixels (mark `v` in illustration).
```

onnx/backend/test/case/node/affinegrid.py
```diff
@@ -0,0 +1,199 @@
...
+        shear_x = np.array([-0.5, 0.5])
+        shear_y = np.array([0.3, -0.3])
+        scale_x = np.array([2.2, 1.1])
+        scale_y = np.array([3.1, 0.9])
+        theta_2d = create_affine_matrix_2d(
+            angle, offset_x, offset_y, shear_x, shear_y, scale_x, scale_y
+        )
+        N, C, W, H = len(angle), 3, 5, 6
+        data_size = (W, H)
+        for align_corners in [0, 1]:
```
justinchuby, line 130:
```suggestion
        for align_corners in (0, 1):

```

nit

onnx/defs/tensor/defs.cc
```diff
@@ -2461,6 +2461,302 @@ ONNX_OPERATOR_SET_SCHEMA(
...
+                  constant_shape_HW_3 = Constant <value_ints: ints = [-1, 3]> ()
+                  original_grid_HW_3 = Reshape (original_grid, constant_shape_HW_3)
+                  original_grid_3_HW_ = Transpose (original_grid_HW_3)
+
+                  original_grid_3_HW = CastLike (original_grid_3_HW_, theta)
+                  grid_N_2_HW = MatMul (theta, original_grid_3_HW)
+                  grid_N_HW_2 = Transpose <perm = [0, 2, 1]> (grid_N_2_HW)
+                  N_H_W_2_seq = SequenceConstruct (N, H, W, int_two_1d)
+                  N_H_W_2 = ConcatFromSequence <axis: int=-1, new_axis: int=0> (N_H_W_2_seq)
+                  grid_2d_then_ = Reshape(grid_N_HW_2, N_H_W_2)
```
justinchuby, line 2607: Consistency: white space after op? Here and elsewhere

```suggestion
                  grid_2d_then_ = Reshape (grid_N_HW_2, N_H_W_2)
```

onnx/backend/test/case/node/affinegrid.py
justinchuby, line 1: Just making sure: Are we testing the function on all possible dtypes?

onnx/reference/ops/op_affine_grid.py
liqunfu: good point! I updated the code to use reference implementation which in turn take advantage of the function body of the body definition. However, I am not sure if not having numpy reference implementation is desirable. For example, people may want to see the sample implementation to understand the op.

onnx/backend/test/case/node/affinegrid.py
liqunfu: As discussed offline. It is a general question: whether we need an option to automatically generate test cases based on types supported by an op.

onnx/defs/tensor/defs.cc
liqunfu: I noticed this convention before. I wonder reason behind this and do we want to enforce it? To me without space is more nature. @gramalingam ?

gramalingam: I agree consistency is good. I personally prefer the space, especially for ONNX here (no nested expressions, only a single call, so it works well).

onnx/backend/test/case/node/affinegrid.py
gramalingam: A comment: actually, we can test the function body for all dtypes at least for type/shape inference. You can do this here: https://github.com/onnx/onnx/blob/8a980683df9acbcb82dc3385fc7eb8cce4ed840f/onnx/test/cpp/function_verify_test.cc#L137 by specifying attribute values for the checking. This helps catch errors like adding float 1.0 with float16 X.

Actually testing for the output values is different.

xadupre: Testing for multiple types is tricky. For example, CUDA implements Gemm for float 8 but it does the computation with floats. The onnx standard does not specify the computation type.

onnx/defs/schema.h
xadupre: +1

onnx/reference/ops/op_affine_grid.py
```diff
@@ -0,0 +1,90 @@
...
+
+from onnx.reference.op_run import OpRun
+
+
+def construct_original_grid(data_size, align_corners):
+    is_2d = len(data_size) == 2
+    size_zeros = np.zeros(data_size)
+    original_grid = [np.ones(data_size)]
+    for dim, dim_size in enumerate(data_size):
+        assert dim_size > 0
```
xadupre, line 16: If this test is only for debugging, it should be removed, otherwise, a better error message would be helpful.

onnx/reference/ops/op_affine_grid.py
xadupre: If this op has a custom implementation, it also means the function body is not tested against the test cases.

onnx/test/reference_evaluator_backend_test.py
```diff
@@ -1,873 +1,886 @@
-# Copyright (c) ONNX Project Contributors
```
xadupre, line 1: EOL is wrong.

onnx/test/automatic_upgrade_test.py
```diff
@@ -1,1813 +1,1821 @@
-# Copyright (c) ONNX Project Contributors
```
xadupre, line 1: EOL is wrong.

onnx/reference/ops/op_affine_grid.py
liqunfu: This is a common issue with function ops: only runtime can validate the function body. For this reason I added test cases for both cases: use or not use reference evaluator.

onnx/defs/tensor/defs.cc
liqunfu: this is very useful! I updated my change to use this function.

docs/Operators.md
liqunfu: document corrected

docs/Operators.md
liqunfu: corrected

onnx/backend/test/case/node/affinegrid.py
liqunfu: corrected

onnx/test/test_backend_onnxruntime.py
```diff
@@ -1,278 +1,279 @@
-# Copyright (c) ONNX Project Contributors
```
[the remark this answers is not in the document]
liqunfu: this is frustrating - lintrunner did this. Otherwise I need to manually edit to make lint happy. Not sure if people working on Windows bearing the same pain?

onnx/test/automatic_upgrade_test.py
liqunfu: there is an easy way in VS CODE to change EOL. Done

onnx/defs/schema.h
gramalingam: I think the name `all_float_types_ir4` is fine, but we should drop the float8 types from the set. We can extend it when all other ops are extended to support float8. (Otherwise, I suspect the type-checker will fail for the function-body.)

onnx/defs/tensor/defs.cc
liqunfu: Sure, I updated with the space convention. To summarize it: open brackets shall be preceded with a single space.

onnx/defs/schema.h
liqunfu: good point. done

onnx/test/test_backend_onnxruntime.py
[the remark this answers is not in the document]
xadupre: There should be an option in your editor for that. Check also git as it may do the replacement as well. On Windows, I use VSCode on WSL. I don't have this issue any more.

onnx/reference/ops/op_affine_grid.py
liqunfu: done

onnx/test/test_backend_onnxruntime.py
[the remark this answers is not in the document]
liqunfu: fixed - vs code allows to set EOL

onnx/reference/ops/op_affine_grid.py
liqunfu: Here is further investigation:
If we do not provide the op's reference implementation AffineGrid(OpRun) - that is: it does not exist or is not registered with ReferenceEvaluator (from onnx.reference.ops.op_affine_grid import AffineGrid in _op_list.py). ReferenceEvaluator can be used to generate expected outputs. In this case the op's function body is used in computation.

If AffineGrid(OpRun) is provided and registered, ReferenceEvaluator can still be used to generated the expected outputs. In this case AffineGrid(OpRun), instead of the op's function body, is used in computation. In this case, however, I would NOT use ReferenceEvaluator because it hides the implementation and defeat the purpose of providing example code. I would bring code from AffineGrid._run() directly into test case generating code.

Back to @gramalingam's comment, I would argue that it is always desirable to have an example/reference implementation to show the computation of an op. We shall always provide an implementation in the form of python methods and call it from test case generating code and reference implementations derived from OpRun. I am very worried on using function body to generate test cases. It will confuse people if the op's function body is incorrect.

I will reimplement the test case generating code to remove uses of ReferenceEvaluator once we conclude this discussion.

onnx/reference/ops/op_affine_grid.py
```diff
@@ -0,0 +1,89 @@
...
+        else:
+            x = np.reshape(a, (1, dim_size)) + size_zeros
+            original_grid = [x, *original_grid]
+    return np.stack(original_grid, axis=2 if is_2d else 3)
+
+
+def apply_affine_transform(theta_n, original_grid_homo):
+    # theta_n: (N, 2, 3) for 2D, (N, 3, 4) for 3D
+    # original_grid_homo: (H, W, 3) for 2D, (D, H, W, 4) for 3D
+    assert theta_n.ndim == 3
```
xadupre, line 49: It is better to have explicit error messages.

gramalingam (approved):

onnx/backend/test/case/node/affinegrid.py
liqunfu: 2 tests added to function_verify_test.cc

onnx/reference/ops/op_affine_grid.py
liqunfu: I have reimplemented the test case generating code to remove uses of ReferenceEvaluator.

PR #5227, opened 2023-05-14
Fix check_tensor to work with large models on Windows

PatriceVignola:
### Description
Fix check_tensor to work with large models

### Motivation and Context
Calling check_tensor on large 7B+ LLM models currently fails because the size doesn't fit in int32, so using `_wstat64` instead fixes the problem.


jcwchen (approved):
Thank you for enhancing checker! Please follow [the steps](https://github.com/onnx/onnx/blob/main/CONTRIBUTING.md#dco) to fix DCO issue before merge.

onnx/checker.cc
```diff
@@ -151,8 +151,8 @@ void check_tensor(const TensorProto& tensor, const CheckerContext& ctx) {
               "' points outside the directory");
         }
         std::wstring data_path = path_join(utf8str_to_wstring(ctx.get_model_dir()), relative_path);
-        struct _stat buff;
-        if (_wstat(data_path.c_str(), &buff) != 0) {
+        struct _stat64 buff;
+        if (_wstat64(data_path.c_str(), &buff) != 0) {
```
gramalingam, line 155: Thanks for the fix. I wonder if we need a similar fix for the POSIX version below?

xadupre:
You need to sign your commit (git commit -s ...) to make the DCO pass or follow the instructions after clicking on details.

onnx/checker.cc
jcwchen: +1 It seems useful and harmless that we can also change:
https://github.com/onnx/onnx/blob/30619cd473119c94088e20762089e54443cdac50/onnx/checker.cc#L187-L188
as `stat64` for non-Windows?

jcwchen (approved):

PR #5238, opened 2023-05-18
Refactor attribute binding

gramalingam:
### Description

* Refactor attribute-binding out as a separate transformation. This logic was duplicated in the inliner as well as shape-inference, and this helps eliminate the duplication.

* Improve the visitor class, and build the attribute-binder as a visitor instance.


onnx/inliner/inliner.cc
```diff
@@ -330,6 +254,9 @@ struct ComputeUsedVars : public NodeVisitorBase {
     return true; // process sub-graphs
   }
 
+ public:
+  std::vector<std::string> result;
```
xadupre, line 258: What if they are duplicated names here?

onnx/common/visitor.h
```diff
@@ -0,0 +1,129 @@
...
+  virtual bool ProcessAttribute(const AttributeProto& attr) {
+    ONNX_UNUSED_PARAMETER(attr);
+    return true;
+  }
+
+  virtual ~Visitor() {}
+};
+
+// MutableVisitor: A version of Visitor that allows mutation of the visited objects.
+struct MutableVisitor {
```
xadupre, line 74: The class is very similar to the previous one. Is there a way to factorize the code?

gramalingam: I thought of using templates to do it, but it seems a bit complicated and not worth it. First, we need to find a way to parametrize over "const" types (one works with "const Proto" and the other with "non const Proto"). Second, protobuf generated code works differently for accessing "const" and "non-const" fields, so we can't write the same code anyway. The mutable version is more general and can always be used for both purposes, but the readonly version is better documentation for readonly methods.

onnx/inliner/inliner.cc
gramalingam: Fixed

onnx/inliner/inliner.cc
```diff
@@ -246,103 +200,114 @@ class Specializer {
...
+// In the case of variables referenced in sub-graphs, only non-local variables
+// are treated as implicit inputs.
 
-struct ComputeUsedVars : public NodeVisitorBase {
-  std::vector<std::string> result;
+class ComputeInputs : private Visitor {
+ private:
+  std::vector<std::unordered_set<std::string>> namescopes;
+
+  bool InNestedScope() {
```
xadupre, line 257: const

onnx/inliner/inliner.cc
```diff
@@ -246,103 +200,114 @@ class Specializer {
...
+    return !namescopes.empty();
+  }
+
+  std::unordered_set<std::string>& CurrentScope() {
```
xadupre, line 261: const?

onnx/shape_inference/implementation.cc
```diff
@@ -404,9 +408,9 @@ class ShapeInferenceImplBase {
     // Resolve domain for node
     auto dit = opset_imports.find(n.domain());
     if (dit == opset_imports.end()) {
-      // Both "" and "ai.onnx" refer to the default ONNX domain
-      if (n.domain() == "") {
-        dit = opset_imports.find("ai.onnx");
+      // Both "" (ONNX_DOMAIN) and "ai.onnx" (AI_ONNX_DOMAIN) refer to the default ONNX domain
+      if (n.domain() == ONNX_DOMAIN) {
```
xadupre, line 412: What if there are other domains, like custom ones?

gramalingam: It is handled 2 lines above (new line number 409). This is just a fallback, because there are two ways of referring to the standard domain. (Anyway, this is just pre-existing code, not part of the current PR ... I was just cleaning up the reference to the string constants using the symbolic identifier.

onnx/inliner/inliner.cc
gramalingam: Not possible, since it returns a mutable set.

onnx/inliner/inliner.cc
```diff
@@ -246,103 +200,114 @@ class Specializer {
...
+    return namescopes.back();
+  }
+
+  bool IsLocalVar(const std::string& name) {
```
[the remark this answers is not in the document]
gramalingam: Done

onnx/inliner/inliner.cc
gramalingam: Done.

xadupre (approved):

onnx/inliner/inliner.cc
```diff
@@ -246,103 +200,114 @@ class Specializer {
...
+
+  std::unordered_set<std::string>& CurrentScope() {
+    return namescopes.back();
+  }
+
+  bool IsLocalVar(const std::string& name) const {
+    for (auto& scope : namescopes) {
+      if (scope.count(name) > 0)
+        return true;
+    }
```
justinchuby, line 269:
```suggestion
      if (scope.count(name) > 0) {
        return true;
      }
    }
```

code robustness against editing errors. @jcwchen I wonder if we should update the clangformat config?

onnx/inliner/inliner.cc
```diff
@@ -448,8 +413,11 @@ void InlineFunctions(ModelProto& model, FunctionMap map, bool convert_version) {
     if (iter != map.end()) {
       callee = *iter->second.first;
       int64_t target_version = iter->second.second;
-      // Rename and specialize called function body
-      Specializer::Specialize(node, callee, "__" + std::to_string(++inline_count), name_generator);
+      // Bind attribute parameters
+      internal::AttributeBinder::BindAttributes(node, callee);
+
+      // Rename variable names in callee
```
justinchuby, line 419: Maybe describe the renaming logic to help readers?

onnx/inliner/inliner.cc
```diff
@@ -246,103 +200,114 @@ class Specializer {
...
-      ++attr_iter;
-    }
+    return true; // Process attribute subgraphs in traversal
   }
 
   // Process a sub-graph, contained as an attribute in a control-flow op node.
-  void Transform(GraphProto& graph) {
+  // Since we need both pre-processing and post-processing in the traversal, we
+  // override the VisitGraph method.
+  void VisitGraph(GraphProto* graph) override {
```
justinchuby, line 219: Just for my learning: so graph needs to be a pointer and not a smart pointer or something?

onnx/inliner/inliner.cc
jcwchen: Personally I vote for always using bracket, but I don't have strong preference. Feel free to update the config and see if anyone has any concern.

onnx/inliner/inliner.cc
gramalingam: Yes.

onnx/inliner/inliner.cc
gramalingam: See lines 135-140.

onnx/inliner/inliner.cc
gramalingam: Done (for this instance; not to config)

onnx/inliner/inliner.cc
justinchuby: Curious what the reason was?

PR #5246, opened 2023-05-22
Fixes a missing attribute in function implementation of CastList

xadupre:
### Description
The function implementation of CastLike does not propagate the attribute *saturate*. This PR fixes it and make the changes in the reference implementation to check the function implementation is working.

### DIFF
--- a/onnx/defs/tensor/defs.cc
+++ b/onnx/defs/tensor/defs.cc
@@ -255,8 +255,9 @@ ONNX_OPERATOR_SET_SCHEMA(
                 return false;
               }
               auto target_elt_type = target_type->tensor_type().elem_type();
+              int64_t saturate = ctx.getAttribute("saturate") != nullptr ? ctx.getAttribute("saturate")->i() : 1;
               FunctionBuilder builder(functionProto);
-              builder.Add("output = Cast (input)", "to", (int64_t)(target_elt_type));
+              builder.Add(MakeString("output = Cast <to= ", (int64_t)(target_elt_type), ", saturate=", saturate, "> (input)").c_str());
               schema.BuildFunction(functionProto);
               return true;
             }));
--- a/onnx/reference/op_run.py
+++ b/onnx/reference/op_run.py
@@ -8,7 +8,7 @@ import numpy as np
 
 from onnx import TensorProto
 from onnx.defs import get_all_schemas_with_history, get_schema, onnx_opset_version
-from onnx.helper import make_node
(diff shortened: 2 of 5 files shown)


onnx/defs/tensor/defs.cc
```diff
@@ -255,8 +255,11 @@ ONNX_OPERATOR_SET_SCHEMA(
                 return false;
               }
               auto target_elt_type = target_type->tensor_type().elem_type();
+              int64_t saturate = ctx.getAttribute("saturate") != nullptr ? ctx.getAttribute("saturate")->i() : 1;
               FunctionBuilder builder(functionProto);
-              builder.Add("output = Cast (input)", "to", (int64_t)(target_elt_type));
+              builder.Add(
+                  MakeString("output = Cast <to= ", (int64_t)(target_elt_type), ", saturate=", saturate, "> (input)")
```
gramalingam, line 261: nit: we can just use "saturate: int = @saturate", without explicitly looking up the value of the attribute (in line 258 above).

gramalingam: It is meant to be lower-case "saturate" in above, but the @ symbol is reformatting this somehow. Not sure how to escape this special-character

xadupre: done

onnx/reference/op_run.py
```diff
@@ -604,24 +628,69 @@ def __init__(self, onnx_node: NodeProto, log_function: Any, impl: Any = None):
...
+        version = parent.opsets[onnx_node.domain]
+        self.schema_ = get_schema(onnx_node.op_type, version, onnx_node.domain)
+
+    def _run(self, *inputs, **kwargs):
+        # Input types are known. They are used to properly
+        # created the body for this operator.
+        types = []
+        for t in inputs:
+            try:
+                ttype = np_dtype_to_tensor_dtype(t.dtype)
```
gramalingam, line 677: Should this fix be made to the called function `np_dtype_to_tensor_dtype`?

xadupre: I chose the following convention https://github.com/onnx/onnx/blob/main/onnx/reference/custom_element_types.py#L7 to represent arrays with bfloat16 or float8. This is only available the for the reference implementation. It is not supported by function from_array, to_array (https://github.com/onnx/onnx/blob/main/onnx/numpy_helper.py#L175). I don't mind moving this code inside np_dtype_to_tensor_dtype but we modifiy the other functions too, including the convention I made to represent float8 numpy arrays. I would be more confident if it is a separate PR.

gramalingam (approved):

Issue #3314, opened 2021-03-05
Confused about  AveragePool op ceil_mode attribute


blacklong28:
# Ask a Question

### Question
When I was study the AveragePool op, I am confused about the ceil_mode attr.
What I see at  [AveragePool](https://github.com/onnx/onnx/blob/master/docs/Operators.md#AveragePool)

> ceil_mode : int (default is 0)
> Whether to use ceil or floor (default) to compute the output shape.

But I notice when auto_pad is set,
it only uses ceil to compute the output shape.
Is the ceil_mode attr  only used when auto_pad is 'NOTSET'?

### Further information

- This is the python code I find to compute the output shape, the code also is only use ceil to compute output shape.  [get_output_shape](https://github.com/onnx/onnx/blob/1cbd549193661c7e9c870c4e9595144d0c6971c5/onnx/backend/test/case/node/pool_op_common.py#L24)
- I notice pytorch's op AvgPool2d have ceil_mode( [AvgPool2d,](https://pytorch.org/docs/stable/generated/torch.nn.AvgPool2d.html) )I guess the onnx ceil_mode attr is only used when auto_pad is 'NOTSET', and when auto_pad is not 'NOTSET',the ceil_mode default value is 1 ,is true? 

Who can help me? Thanks very much.

jcwchen:
Hi @blacklong28,
I think Operator document is correct and the test script in ONNX you mentioned [get_output_shape](https://github.com/onnx/onnx/blob/1cbd549193661c7e9c870c4e9595144d0c6971c5/onnx/backend/test/case/node/pool_op_common.py#L24) is not updated while introducing ceil_mode. You can check the inference logic here in ONNXRunitme:
https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/providers/cpu/nn/pool_attributes.h
By default, ceil_mode is 0 no matter what auto_pad is. Thanks.

kraiskil:
For reference - I had a similar question about ceil mode in MaxPool #2971 
@jcwchen can you confirm that the onnx test scripts are wrong? Since they seem to be the de-facto acceptance tests for backends, at least I went with the script and not the documentation as the basis for my backend implementation.

blacklong28:
@kraiskil @jcwchen 
I have read this code: https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/providers/cpu/nn/pool_attributes.h
[...]

PR #5248, opened 2023-05-22
Clarify ceil_mode when using auto_pad for pooling

jcwchen:
### Description

- Clarify ceil_mode when using auto_pad for pooling: honor ceil_mode when using auto_pad -- when ceil_mode is enabled, ceil will be used; otherwise use floor for output_shape. The document and test code will be updated. No updates for the existing backend data in ONNX and onnxruntime's behavior for this case is consistent as well.
- Clarify the backend test code should use NOTSET for auto_pad by default.

### Motivation and Context

Closes https://github.com/onnx/onnx/issues/3314. The document is ambiguous when using ceil_mode with auto_pad.


liqunfu (approved):

onnx/defs/nn/defs.cc
```diff
@@ -222,11 +222,16 @@ std::function<void(OpSchema&)> PoolOpSchemaGenerator(
  ```
  if ceil_mode is enabled `pad_shape[i]` is the sum of pads along axis `i`.
 
- `auto_pad` is a DEPRECATED attribute. If you are using them currently, the output spatial shape will be following:
+ `auto_pad` is a DEPRECATED attribute. If you are using them currently, the output spatial shape will be following when ceil_mode is enabled:
```
liqunfu, line 225: I found ceil mode is specific to Pytorch where SAME is to TF. Maybe we shall not link equivalence them?

jcwchen: > ceil mode is specific to Pytorch where SAME is to TF.

Yes, initially there is only auto_pad in ONNX pooling family operators and later ONNX brought ceil_mode from PyTorch. When introducing ceil_mode, the combination behavior of these two attributes (auto_pad+ceil_mode) are not specified and caused confusion: https://github.com/onnx/onnx/issues/3314. Since it's an undefined behavior in ONNX-level, I am slightly inclined to make ONNX definition clear as current ONNX Runtime's behavior: honor ceil_mode when using auto_pad, to prevent runtime behavior changes, although in practice there shouldn't any real-world models using them (auto_pad+ceil_mode) simultaneously since one is from PyTorch and another is from TensorFlow. It seems fine to have this combination, as long as the document is clear.

Another option is making ONNX forbid using auto_pad and ceil_mode at the same time, but I am not 100% sure whether it would break backward compatibility in runtime.

liqunfu: Good point! Thank you for this clarification. It is definitely better to explicitly state the behavior of combination of auto_pad with ceil_mode. I am in the mean time preparing a PR to add to this mix another extra dilation input. In my case I was planning to avoid the interplay between dilation and auto_pad.

liqunfu: According to TF documentation: https://www.tensorflow.org/api_docs/python/tf/keras/layers/MaxPool1D#args
"same" means padding to make output size the same as input. "valid" means no padding. Sounds like "same" is close to ceil_mode which is opposite to the changed doc here.
Here is my experiment:
>>> x = tf.constant([1., 2., 3., 4., 0.])
>>> x = tf.reshape(x, [1, 5, 1])
>>> max_pool_1d = tf.keras.layers.MaxPooling1D(pool_size=2, strides=4, padding='same')
>>> max_pool_1d(x)
<tf.Tensor: shape=(1, 2, 1), dtype=float32, numpy=
array([[[2.],
        [0.]]], dtype=float32)>
which matches:
>>> import torch
>>> from torch import nn
>>> m = nn.MaxPool1d(2, stride=4, padding=0, dilation=1, ceil_mode=True)
>>> input = torch.from_numpy(np.array([[[1, 2, 3, 4, 0]]])).float()
>>> output = m(input)
>>> output
tensor([[[2., 0.]]])


on the other hand:
>>> x = tf.constant([1., 2., 3., 4., 0.])
>>> x = tf.reshape(x, [1, 5, 1])
>>> max_pool_1d = tf.keras.layers.MaxPooling1D(pool_size=2, strides=4, padding='same')
>>> max_pool_1d(x)
<tf.Tensor: shape=(1, 2, 1), dtype=float32, numpy=
array([[[2.],
        [0.]]], dtype=float32)>

which matches:
>>> m = nn.MaxPool1d(2, stride=4, padding=0, dilation=1, ceil_mode=True)
>>> input = torch.from_numpy(np.array([[[1, 2, 3, 4, 0]]])).float()
>>> output = m(input)
>>> output
tensor([[[2., 0.]]])

onnx/backend/test/case/node/averagepool.py
```diff
@@ -199,7 +199,7 @@ def export_averagepool_1d_default() -> None:
         x_shape = np.shape(x)
         kernel_shape = [2]
         strides = [1]
-        out_shape = get_output_shape("VALID", x_shape[2:], kernel_shape, strides)
+        out_shape = get_output_shape("NOTSET", x_shape[2:], kernel_shape, strides)
```
gramalingam, line 202: Why is this VALID changing to NOTSET? Is this another bug?

jcwchen: Good question. It's more like a clarification -- by default pooling operators will use NOTSET instead of VALID and thus the code here looks confusing. However, the result of shape are just accidentally the same here for NOTSET or VALID so no test data needs to be changed. I have mentioned this clarification in the PR description as well.

gramalingam (approved):

Issue #5161, opened 2023-04-20
It's worthy to fix reference implementation for AveragePool-19 as well?


liqunfu:
It's worthy to fix reference implementation for AveragePool-19 as well?

_Originally posted by @jcwchen in https://github.com/onnx/onnx/pull/5158#discussion_r1172743868_

liqunfu:
https://github.com/onnx/onnx/pull/5250

Issue #4849, opened 2023-02-03
(in _op_common_pool.py) CommonPool._run returns incorrect output shape when dilations is set


liqunfu:
# Bug Report

### Is the issue related to model conversion?
No

### Describe the bug
(in onnx\reference\ops\_op_common_pool.py) CommonPool._run returns incorrect output shape when dilations is set.

### System information
NP

### Reproduction instructions
https://github.com/onnx/onnx/blob/bb97271c2cd06e67fc3c5f9b7516932ec8e38c03/onnx/test/reference_evaluator_backend_test.py#L59
enable test for test_lppool_2d_dilations (comment this line), run
reference_evaluator_backend_test.py -k test_lppool_2d_dilations

this test fails because expected output shape differs from real. 

with the same data running Pytorch nn.MaxPool2d show that the computed shape is incorrect
- Attach the ONNX model to the issue (where applicable)-->

### Expected behavior
the test shall pass

Tweezky66:
@liqunfu Hi,i'm looking forward to work on this,may i?

kkawalpreet710-droid:
Hi, I've opened a PR to fix this issue: onnx/onnx#8174
The fix accounts for dilations in the output shape calculation by computing the effective kernel size.

LeonxLJX:
I'd like to take this one.

From the report, the useful first step is to pin down the exact code path involved before touching anything — these tend to have a small true cause behind a broad-looking symptom, and guessing at it usually produces a patch that masks the behavior instead of fixing it.

Plan: trace the path, reproduce it, then make a focused change with a test if this area has coverage. I'll keep the diff small so it's easy to review.

If there's an existing convention in this repo for how this kind of thing is handled, I'd rather follow it than invent my own — happy to be pointed at one.

(claiming via @LeonxLJX)

PR #5250, opened 2023-05-24
reference implementation of pool ops to support dilation

liqunfu:
### Description
With reference implementation, dilation is not supported for AvergePool and LPPool. This PR provide such support

### Motivation and Context
To improve test coverage
https://github.com/onnx/onnx/issues/4849
https://github.com/onnx/onnx/issues/5161

The next step will be to replace maxpool with the new reference implementation. I still need time to figure out how to support argmax for maxpool with the new reference implementation therefore need another PR for this work.


onnx/backend/test/case/node/pool_op_common.py
```diff
@@ -27,10 +27,52 @@ def get_pad_shape(
         pass
     return pad_shape
 
+# Use pads to calculate output shape. Use output shape in turn to calculate the actual pads
```
xadupre, line 30: I would put that information in the docstring of the function rather than in a comment.

onnx/backend/test/case/node/pool_op_common.py
```diff
@@ -27,10 +27,52 @@ def get_pad_shape(
...
+    strides_spatial: Sequence[int],
+    ceil_mode: bool,
+) -> Tuple[Sequence[int], Sequence[int]]:
+    
+    # compute output shape according to:
+    # https://pytorch.org/docs/stable/generated/torch.nn.MaxPool1d.html?highlight=max+pool#torch.nn.MaxPool1d
+    output_spatial_shape = [0] * len(input_spatial_shape)
+    dims = len(input_spatial_shape)
+    for dim in range(dims):
+        dim_size = (float)(input_spatial_shape[dim] + pads_spatial_shape[dim] + pads_spatial_shape[dims + dim] - dialations_spatial_shape[dim] * (kernel_spatial_shape[dim] - 1) - 1) / strides_spatial[dim] + 1
```
xadupre, line 46: float(...) / 1... but / is a float division anyway in python (!= from c++)

onnx/backend/test/case/node/pool_op_common.py
```diff
@@ -27,10 +27,52 @@ def get_pad_shape(
...
+        if ceil_mode:
+            output_spatial_shape[dim] = int(np.ceil(dim_size))
+        else:
+            output_spatial_shape[dim] = int(np.floor(dim_size))
+
+    # import numpy as np
```
xadupre, line 52: If it is important, you should put that in the docstring as well. It will show up in the documentation.

onnx/backend/test/case/node/averagepool.py
```diff
@@ -197,11 +198,12 @@ def export_averagepool_1d_default() -> None:
         )
         x = np.random.randn(1, 3, 32).astype(np.float32)
         x_shape = np.shape(x)
+        dilations = [1]
         kernel_shape = [2]
         strides = [1]
         out_shape = get_output_shape("VALID", x_shape[2:], kernel_shape, strides)
         padded = x
-        y = pool(padded, x_shape, kernel_shape, strides, out_shape, [0], "AVG")
+        y = pool(padded, x_shape, kernel_shape, strides, out_shape, [0, 0], "AVG", dilations=dilations)
```
gramalingam, line 206: Can we not handle this (and similar examples below) by using a default-value of None for dilations in the implementation of `pool`?

gramalingam: That would match the ONNX representation above (where `dilations` is not specified).

onnx/backend/test/case/node/pool_op_common.py
liqunfu: good point. I added comment to the function body.

onnx/backend/test/case/node/pool_op_common.py
liqunfu: code corrected

onnx/backend/test/case/node/maxpool.py
```diff
@@ -1,512 +1,707 @@
-# Copyright (c) ONNX Project Contributors
```
xadupre, line 1: EOL is wrong.

liqunfu: vs code does allow switch between EOLs easily.

onnx/backend/test/case/node/averagepool.py
liqunfu: dilations as an argument removed from the call

gramalingam (approved):

Issue #4160, opened 2022-04-25
Extend shape inference tests to cover older opset versions


jcwchen:
# Feature Request

### What is the problem that this feature solves?
Shape inference tests for older opset versions will be replaced with newer opset version if there is a opset bump. The reason is [shape_inference_test.py](https://github.com/onnx/onnx/blob/main/onnx/test/shape_inference_test.py) relies on onnx.helper and it will always use the latest opset version to create the ONNX models for testing. Therefore, ONNX should extend the test to cover shape inference test for older opset versions as well.

### Describe the alternatives you have considered
A clear and concise description of any alternative solutions or features you've considered

### Describe the feature
Why is this feature necessary? What does it accomplish?

### Will this influence the current api?
No

### Feature Area
Which area in ONNX does this impact?  (*e.g. model usage, backend, best practices, converters, shape_inference, version_converter, training, test, operators*): test for shape inference

### Are you willing to contribute it (Y/N):
Y

jcwchen:
Still valid. Comment to prevent from bot closing it.

guoyuhong:
@jcwchen I'm trying to solve the problem by introducing two functions in https://github.com/onnx/onnx/pull/5263 :
1. `_get_op_version` is used to get the OP versions:
        1. If op has only one version, [] is returned.
        2. If op has many version, [versions] is returned.
        3. If op is not in the opset, None is returned.
3. `_test_an_op` is used to run many test functions for different versions of an op. There is a function call pattern:
        1. If an op has only on version, the pattern is `_internal_${test_funtion_middle_name}`. In the PR, there is an example `_internal_test_transpose`, because Transpose has only one version.
        2. If an op has many versions, the pattern is `_internal_${test_funtion_middle_name}_v${version}`. In the PR, there are 2 exmaple functions: ``, `_internal_test_squeeze_v11` and `_internal_test_squeeze_v13`. I cannot implement `_internal_test_squeeze_v1`, because there is a Reshape create error(See the figure at the end). Therefore, I skip the test when `if int(version) <= 5`.

In this way, if Transpose has many versions later one, this design requires the developer to keep the old version test and add a new version test.

![image](https://github.com/onnx/onnx/assets/19584326/43ab2fe7-e86e-4342-a39e-d59d39d536cb)

jcwchen:
@guoyuhong Thank you for your contribution! It seems good to me. I will take a closer look at the PR and let's continue there.

PR #5263, opened 2023-05-27
Refactor shape_inference_test for old Op versions PR1

guoyuhong:
### Description

https://github.com/onnx/onnx/issues/4160

### Motivation and Context

Shape inference tests for older opset versions are not supported before this PR. This refactor would help to test old-versioned ops.
This PR introduces a function `all_versions_for` to help generate test functions for all versions of one Op. The users can use `@parameterized.expand(all_versions_for("${op_name}"))` to annotate the test function to apply the test function to all op versions. The test function should have the arguments like `def test_op(self, version_name, op_version)`. `version_name` is usually unused but it's better to keep that to make the parameterized test name more readable. For example, with this `version_name`, the test name will have a version suffix showed as follows.
image

### DIFF
--- a/onnx/test/shape_inference_test.py
+++ b/onnx/test/shape_inference_test.py
@@ -652,7 +652,9 @@ class TestShapeInference(TestShapeInferenceHelper):
     def test_concat_missing_shape(self) -> None:
         self._test_an_op("Concat", self.get_caller_function_name())
 
-    def _internal_test_concat_missing_shape_v11(self, op_name, version) -> None:
+    def _internal_test_concat_missing_shape_v11(
+        self, op_name, version  # pylint: disable=unused-argument
+    ) -> None:
         assert op_name == "Concat"
         graph = self._make_graph(
             [
@@ -1521,7 +1523,9 @@ class TestShapeInference(TestShapeInferenceHelper):
     def test_squeeze(self) -> None:
         self._test_an_op("Squeeze", self.get_caller_function_name())
 
-    def _internal_test_squeeze_v11(self, op_name, version) -> None:
+    def _internal_test_squeeze_v11(
+        self, op_name, version  # pylint: disable=unused-argument
+    ) -> None:
         assert op_name == "Squeeze"
         graph = self._make_graph(
             [("x", TensorProto.FLOAT, (1, 3, 1, 1, 2, 1))],
@@ -1534,7 +1538,9 @@ class TestShapeInference(TestShapeInferenceHelper):
             opset_imports=[helper.make_opsetid(ONNX_DOMAIN, 11)],
         )
 
-    def _internal_test_squeeze_v13(self, op_name, version) -> None:
+    def _internal_test_squeeze_v13(
+        self, op_name, version  # pylint: disable=unused-argument
+    ) -> None:
         assert op_name == "Squeeze"
         graph = self._make_graph(
             [
--- a/onnx/test/shape_inference_test.py
+++ b/onnx/test/shape_inference_test.py
@@ -34,6 +34,7 @@ from onnx.helper import (
     make_tensor_value_info,
 )
 from onnx.parser import parse_graph
+from onnx.reference.ops import _op_list as op_list
 
 
 class TestShapeInferenceHelper(unittest.TestCase):
@@ -173,6 +174,41 @@ class TestShapeInferenceHelper(unittest.TestCase):
                 "Unrecognized value info type in _compare_value_infos: ", str(vi_type)
             )
 
+    def _get_op_version(self, op_name):
+        op_names = dir(op_list)
+        if op_name in op_names:
+            return []
+        import re
+
+        ret = []
+        name_pattern = re.compile("^{}_([0-9]+)$".format(op_name))
+        for name in op_names:
+            m = name_pattern.search(name)
+            if m is not None:
+                ret.append(m.group(1))
+        if op_name in op_names or len(ret) > 0:
+            return ret
+        else:
+            return None
+
+    def _test_an_op(self, op_name, test_function_middel_name) -> None:
+        versions = self._get_op_version(op_name)
+        assert versions is not None
+        if len(versions) == 0:
+            # Only one version found.
+            function_name = f"_internal_{test_function_middel_name}"
+            test_func = self.__getattribute__(function_name)
+            assert test_func is not None, "Cannot find {function_name}"
+            test_func()
+        for version in versions:
+            if int(version) <= 5:
+                # Not sure how to refactor the Reshape in self._make_graph.
+                continue
+            function_name = f"_internal_{test_function_middel_name}_v{version}"
+            test_func = self.__getattribute__(function_name)
+            assert test_func is not None, "Cannot find {function_name}"
+            test_func()
+
 
 class TestShapeInference(TestShapeInferenceHelper):
     def test_empty_graph(self) -> None:
@@ -190,6 +226,9 @@ class TestShapeInference(TestShapeInferenceHelper):
         )
 
     def test_transpose(self) -> None:
+        self._test_an_op("Transpose", "test_transpose")
+
+    def _internal_test_transpose(self) -> None:
         graph = self._make_graph(
             [("X", TensorProto.FLOAT, (2, 3, 4))],
             [make_node("Transpose", ["X"], ["Y"], perm=[1, 0, 2])],
@@ -1170,6 +1209,24 @@ class TestShapeInference(TestShapeInferenceHelper):
         )  # type: ignore
 
     def test_squeeze(self) -> None:
+        self._test_an_op("Squeeze", "test_squeeze")
+
+    def _internal_test_squeeze_v11(self) -> None:
+        graph = self._make_graph(
+            [("x", TensorProto.FLOAT, (1, 3, 1, 1, 2, 1))],
+            [make_node("Squeeze", "x", "y", axes=[0, 2, 3, 5])],
+            [],
+        )
+        operatorsetid = OperatorSetIdProto()
+        operatorsetid.domain = ""
+        operatorsetid.version = 11
+        self._assert_inferred(
+            graph,
+            [make_tensor_value_info("y", TensorProto.FLOAT, (3, 2))],
+            opset_imports=[operatorsetid],
+        )
+
+    def _internal_test_squeeze_v13(self) -> None:
         graph = self._make_graph(
             [
                 ("x", TensorProto.FLOAT, (1, 3, 1, 1, 2, 1)),
@@ -1179,8 +1236,13 @@ class TestShapeInference(TestShapeInferenceHelper):
             [],
             initializer=[make_tensor("axes", TensorProto.INT64, (4,), (0, 2, 3, 5))],
         )
+        operatorsetid = OperatorSetIdProto()
+        operatorsetid.domain = ""
+        operatorsetid.version = 13
         self._assert_inferred(
-            graph, [make_tensor_value_info("y", TensorProto.FLOAT, (3, 2))]
+            graph,
+            [make_tensor_value_info("y", TensorProto.FLOAT, (3, 2))],
+            opset_imports=[operatorsetid],
         )
 
     def test_unsqueeze_regular(self) -> None:
--- a/onnx/test/shape_inference_test.py
+++ b/onnx/test/shape_inference_test.py
@@ -43,11 +43,11 @@ from onnx.parser import parse_graph
 
 
 class TestShapeInferenceHelper(unittest.TestCase):
-    all_op_versions = {}
+    all_op_versions: dict[str, set[int]] = {}
 
     @classmethod
-    def _get_available_versions(cls, schema):
-        versions = set()
+    def get_available_versions(cls, schema):
+        versions: set[int] = set()
         for version in range(schema.since_version, 0, -1):
             try:
                 versions.add(defs.get_schema(schema.name, version).since_version)
@@ -191,43 +191,35 @@ class TestShapeInferenceHelper(unittest.TestCase):
                 "Unrecognized value info type in _compare_value_infos: ", str(vi_type)
             )
 
-    def _test_an_op(self, op_name, test_function_name) -> None:
-        """Given an operator name and test_function_name, run tests for op versions > 5.
-        :param op_name: The operator name.
-        :param test_function_name: The test function name for the op. It's used to form test functions.
-            The caller can use self.get_caller_function_name() to get the function name.
-        :return: None
-        """
-        versions = list(self.op_versions[op_name])
-        assert versions is not None
-        if len(versions) == 0:
-            # The version starts from 0 to now.
-            function_name = f"_internal_{test_function_name}"
-            test_func = getattr(self, function_name, None)
-            if test_func is None:
-                raise NotImplementedError(f"Cannot find test function {function_name}")
-            test_func(op_name=op_name)
-            return
-        for version in versions:
-            # Many versions found for this op.
-            if int(version) <= 5:
-                # Not sure how to fix the Reshape error in self._make_graph.
-                continue
-            function_name = f"_internal_{test_function_name}_v{version}"
-            test_func = getattr(self, function_name, None)
-            if test_func is None:
-                raise NotImplementedError(f"Cannot find test function {function_name}")
-            test_func(op_name=op_name, version=version)
-
 
 class TestShapeInference(TestShapeInferenceHelper):
     @staticmethod
-    def single_op_tester(op_name):
+    def single_op_all_version_tester(op_name):
+        """This is decorator to run one test case for all versions of an Op.
+
+        However, there is a creation issue for Reshape Operator under version 5.
+        Therefore, all test cases only run when version > 5.
+
+        Args:
+            op_name: The name of the Op, which should be in `defs.get_all_schemas`.
+
+        Returns:
+            The decorated function for unittest case.
+            For example, we can use:
+            ```
+            @single_op_tester("Transpose")
+            def test_transpose(self, version):
+                ...
+            ```
+
+        Raises:
+            KeyError: If the op_name is not in `defs.get_all_schemas`.
+        """
         cls = TestShapeInferenceHelper
         if len(cls.all_op_versions) == 0:
             # Init the version map for all Ops.
             for schema in defs.get_all_schemas():
-                cls.all_op_versions[schema.name] = cls._get_available_versions(schema)
+                cls.all_op_versions[schema.name] = cls.get_available_versions(schema)
 
         def decorate(test_function):
             if op_name not in cls.all_op_versions:
@@ -271,7 +263,7 @@ class TestShapeInference(TestShapeInferenceHelper):
             graph, [make_tensor_value_info("y", TensorProto.FLOAT, (30, 4, 5))]
         )
 
-    @single_op_tester("Transpose")
+    @single_op_all_version_tester("Transpose")
     def test_transpose(self, version) -> None:
         graph = self._make_graph(
             [("X", TensorProto.FLOAT, (2, 3, 4))],
@@ -284,7 +276,7 @@ class TestShapeInference(TestShapeInferenceHelper):
             opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
         )
 
-    @single_op_tester("Transpose")
+    @single_op_all_version_tester("Transpose")
     def test_transpose_preexisting(self, version) -> None:
         graph = self._make_graph(
             [("X", TensorProto.FLOAT, (2, 3, 4))],
@@ -297,7 +289,7 @@ class TestShapeInference(TestShapeInferenceHelper):
             opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
         )
 
-    @single_op_tester("Transpose")
+    @single_op_all_version_tester("Transpose")
     def test_transpose_scalar(self, version) -> None:
         graph = self._make_graph(
             [("X", TensorProto.FLOAT, ())],
@@ -311,7 +303,7 @@ class TestShapeInference(TestShapeInferenceHelper):
             opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
         )
 
-    @single_op_tester("Transpose")
+    @single_op_all_version_tester("Transpose")
     def test_transpose_partial(self, version) -> None:
         graph = self._make_graph(
             [("X", TensorProto.FLOAT, (2, 3, 4))],
@@ -324,7 +316,7 @@ class TestShapeInference(TestShapeInferenceHelper):
             opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
         )
 
-    @single_op_tester("Transpose")
+    @single_op_all_version_tester("Transpose")
     def test_transpose_preexisting_incorrect_shape(
         self, version  # pylint: disable=unused-argument
     ) -> None:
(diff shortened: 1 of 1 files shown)


guoyuhong:
@jcwchen Do you have time to take a look at this change?

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +175,46 @@ def _compare_value_infos(
...
+        ret = []
+        name_pattern = re.compile(f"^{op_name}_([0-9]+)$")
+        for name in op_names:
+            m = name_pattern.search(name)
+            if m is not None:
+                ret.append(m.group(1))
+        if op_name in op_names or len(ret) > 0:
+            return ret
+        else:
+            return None
```
jcwchen, line 197: It seems more direct to use `onnx.defs` to get all available versions for operators? For instance:
```
from onnx import defs
from onnx.defs import SchemaError
def get_available_versions(schema):
  versions = set()
  for version in range(schema.since_version, 0, -1):
   try:
     versions.add(defs.get_schema(schema.name, version).since_version)
   except SchemaError:
     break
  return versions
  
  
op_versions = {}
for schema in defs.get_all_schemas():
  if schema.domain == "":
    op_versions[schema.name] = get_available_versions(schema)
print(op_versions)
```

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +175,46 @@ def _compare_value_infos(
...
+        if len(versions) == 0:
+            # Only one version found for this op.
+            function_name = f"_internal_{test_function_middel_name}"
+            test_func = getattr(self, function_name)
+            assert test_func is not None, "Cannot find test function {function_name}"
+            test_func()
+        for version in versions:
+            # Many versions found for this op.
+            if int(version) <= 5:
+                # Not sure how to fix the Reshape error in self._make_graph.
```
jcwchen, line 211: I was wondering what kind of error you have bumped into?

jcwchen (commented):
Thank you for the PR! I like the idea, but IIRC (I tried to test it before), we will probably need to modify quite a few existing tests to accommodate old opset versions. Still, it's great to have a starting point now and we can decouple the modified work if needed.

onnx/test/shape_inference_test.py
guoyuhong: That looks much better. I will try it later~

guoyuhong: This function works fine, so I used it in this PR. Thanks!

guoyuhong:
@jcwchen I refined my code and applied this pattern to several Ops. For most Ops, the one test case fits all versions of this Op. Until now, I found there are Ops that have some incompatible tests. Here is a simple summary:
1. `Reshape`'s `allowzero` since Version 14, so `test_reshape_static_shape_allowzero` before Version 14 is skipped.
2. `Upsample` is deprecated from Version 10. I added special test case for Version 10.
3. `Upsample`'s test case for Version 7 and Version 9 are different. I copied test from the old code from Version 7.
4. `Squeeze`'s test case for Version 11 and Version 13 are different.  I copied test from the old code from Version 11.

onnx/test/shape_inference_test.py
```diff
@@ -37,6 +44,20 @@
 
 
 class TestShapeInferenceHelper(unittest.TestCase):
+    def setUp(self):
+        self.op_versions = {}
+        for schema in defs.get_all_schemas():
+            self.op_versions[schema.name] = self._get_available_versions(schema)
+
+    def _get_available_versions(self, schema):
```
justinchuby, line 52: Consider moving this function out to the global namespace since it does use self?

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +194,34 @@ def _compare_value_infos(
                 "Unrecognized value info type in _compare_value_infos: ", str(vi_type)
             )
 
+    def _test_an_op(self, op_name, test_function_name) -> None:
+        """Given an operator name and test_function_name, run tests for op versions > 5.
+        :param op_name: The operator name.
```
justinchuby, line 199: Prefer google style comments.

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +194,34 @@ def _compare_value_infos(
                 "Unrecognized value info type in _compare_value_infos: ", str(vi_type)
             )
 
+    def _test_an_op(self, op_name, test_function_name) -> None:
+        """Given an operator name and test_function_name, run tests for op versions > 5.
```
justinchuby, line 198: Add an explanation on why >5 for clarity?

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +194,34 @@ def _compare_value_infos(
...
+        :return: None
+        """
+        versions = list(self.op_versions[op_name])
+        assert versions is not None
+        if len(versions) == 0:
+            # The version starts from 0 to now.
+            function_name = f"_internal_{test_function_name}"
+            test_func = getattr(self, function_name, None)
+            if test_func is None:
+                raise NotImplementedError(f"Cannot find test function {function_name}")
```
justinchuby, line 211: I would provide a suggested action in the error message

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +194,34 @@ def _compare_value_infos(
...
+        :param op_name: The operator name.
+        :param test_function_name: The test function name for the op. It's used to form test functions.
+            The caller can use self.get_caller_function_name() to get the function name.
+        :return: None
+        """
+        versions = list(self.op_versions[op_name])
+        assert versions is not None
+        if len(versions) == 0:
+            # The version starts from 0 to now.
+            function_name = f"_internal_{test_function_name}"
```
justinchuby, line 208: This creates an assumption on how we name the tests. @jcwchen is there a way we make this more robust?

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +194,34 @@ def _compare_value_infos(
...
+            test_func = getattr(self, function_name, None)
+            if test_func is None:
+                raise NotImplementedError(f"Cannot find test function {function_name}")
+            test_func(op_name=op_name)
+            return
+        for version in versions:
```
justinchuby, line 214: Is there a way we can create parameterized tests, such that each version of an op can fail separately, in it own test method?

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +194,34 @@ def _compare_value_infos(
...
+            # Many versions found for this op.
+            if int(version) <= 5:
+                # Not sure how to fix the Reshape error in self._make_graph.
```
justinchuby, line 217: Create an issue to track this if it needs to be fixed, then link it here?

onnx/test/shape_inference_test.py
```diff
@@ -189,48 +238,94 @@ def _identity_prop(self, op: str, **kwargs: Any) -> None:
             graph, [make_tensor_value_info("y", TensorProto.FLOAT, (30, 4, 5))]
         )
 
+    def get_caller_function_name(self):
+        """Return the function name that calls this get_caller_function_name."""
```
justinchuby, line 242: I’m not quite following this doc

onnx/test/shape_inference_test.py
```diff
@@ -189,48 +238,94 @@ def _identity_prop(self, op: str, **kwargs: Any) -> None:
...
+        return inspect.stack()[1][3]
+
+    def test_no_op(self) -> None:
```
justinchuby, line 245: Prefer to include the expected behavior when possible:

test_xxx_raises_key_error

or something more accurate. Here an everywhere else where appropriate. 

————-

I would include the expected behavior as part of the test method name. To make what a test is testing clear: `test_<method or function name>_<expected behavior>_[when_<condition>]`.

e.g. `test_method_x_raises_error_when_dims_is_not_a_sequence`

onnx/test/shape_inference_test.py
```diff
@@ -173,6 +194,34 @@ def _compare_value_infos(
                 "Unrecognized value info type in _compare_value_infos: ", str(vi_type)
             )
 
+    def _test_an_op(self, op_name, test_function_name) -> None:
```
justinchuby, line 197: Naming: _test_op, _test_single_op etc. ?

onnx/test/shape_inference_test.py
gramalingam: I wonder if it would be worth turning this into a decorator on the tests, like
```py
@test_op("Transpose")
def test_transpose(self):
```
instead of having to define two methods, as below.

onnx/test/shape_inference_test.py
gramalingam: Something like `with self.subTest(...):` ?

onnx/test/shape_inference_test.py
guoyuhong: What a good idea! That decorator makes the code much simpler.

onnx/test/shape_inference_test.py
guoyuhong: `subTest` is a good feature. If I do not implement v11 test functions for `Squeeze`, I will get a much clearer error.
image

onnx/test/shape_inference_test.py
guoyuhong: Yes, I changed it into a `classmethod` and the name to `get_available_versions`.

onnx/test/shape_inference_test.py
guoyuhong: Done. Moved the comment to `single_op_all_version_tester`.

onnx/test/shape_inference_test.py
guoyuhong: Added following sentences:
```
        However, there is a creation issue for Reshape Operator under version 5, which
        probably caused by the mismatch creation arguments in _make_graph.
        Therefore, all test cases now only run when version > 5.
```

onnx/test/shape_inference_test.py
guoyuhong: No such error in the new decorator design.

onnx/test/shape_inference_test.py
guoyuhong: Done: https://github.com/onnx/onnx/issues/5289

onnx/test/shape_inference_test.py
guoyuhong: Forget that. This function is deleted in the new decorator design.

onnx/test/shape_inference_test.py
guoyuhong: Changed the function to a decorator named `single_op_all_version_tester`.

onnx/test/shape_inference_test.py
guoyuhong: Thanks for the detail explanation. I will try to following this format in the new test cases.
These two test cases are deleted.

onnx/test/shape_inference_test.py
```diff
@@ -174,6 +192,65 @@ def _compare_value_infos(
...
+    Issue reference: https://github.com/onnx/onnx/issues/5289.
+
+    Args:
+        op_name: The name of the Op, which should be in `defs.get_all_schemas`.
+
+    Returns:
+        The decorated function for unittest case.
+        For example, we can use:
+        ```
+        @single_op_tester("Transpose")
```
gramalingam, line 210: Nit: name needs to be changed to match actual name "single_op_all_version_tester"

guoyuhong: Good catch...

onnx/test/shape_inference_test.py
```diff
@@ -174,6 +192,65 @@ def _compare_value_infos(
...
+        if op_name not in cls.all_op_versions:
+            raise KeyError(f"Invalid Op name: {op_name}")
+        versions = cls.all_op_versions[op_name]
+        assert versions is not None
+
+        def wrapper(*args, **kwargs):
+            assert len(versions) > 0
+            for version in versions:
+                # Many versions found for this op.
+                if int(version) <= 5:
```
xadupre, line 234: This could lead to a disabled test if all possible version are <= 5.

onnx/test/shape_inference_test.py
```diff
@@ -189,64 +266,85 @@ def _identity_prop(self, op: str, **kwargs: Any) -> None:
...
-            graph, [make_tensor_value_info("Y", TensorProto.FLOAT, (3, 2, 4))]
+            graph,
+            [make_tensor_value_info("Y", TensorProto.FLOAT, (3, 2, 4))],
+            opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
         )
 
-    def test_transpose_preexisting_incorrect_shape(self) -> None:
+    @single_op_all_version_tester("Transpose")
+    def test_transpose_preexisting_incorrect_shape(
+        self, version  # pylint: disable=unused-argument
```
xadupre, line 324: Why version is not used here?

onnx/test/shape_inference_test.py
```diff
@@ -1180,7 +1596,21 @@ def test_squeeze(self) -> None:
             initializer=[make_tensor("axes", TensorProto.INT64, (4,), (0, 2, 3, 5))],
         )
         self._assert_inferred(
-            graph, [make_tensor_value_info("y", TensorProto.FLOAT, (3, 2))]
+            graph,
+            [make_tensor_value_info("y", TensorProto.FLOAT, (3, 2))],
+            opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
+        )
+
+    def _test_squeeze_v11(self, version) -> None:
```
xadupre, line 1604: Unused?

onnx/test/shape_inference_test.py
```diff
@@ -591,10 +765,29 @@ def test_upsample(self) -> None:
         self._assert_inferred(
             graph,
             [make_tensor_value_info("y", TensorProto.INT32, (2, 4, 3, 9))],
-            opset_imports=[helper.make_opsetid(ONNX_DOMAIN, 9)],
+            opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
         )
 
-    def test_upsample_raw_data(self) -> None:
+    def _test_upsample_v7(self, version) -> None:
```
xadupre, line 771: Unused?

onnx/test/shape_inference_test.py
guoyuhong: Yes, I added an issue to track this. https://github.com/onnx/onnx/issues/5289
Currently, shape_inference_test.py only test recent op version by default. This PR helps to enable tests for more op versions > 5.

onnx/test/shape_inference_test.py
```diff
@@ -615,10 +808,10 @@ def test_upsample_raw_data(self) -> None:
         self._assert_inferred(
             graph,
             [make_tensor_value_info("y", TensorProto.INT32, (2, 4, 3, 9))],
-            opset_imports=[helper.make_opsetid(ONNX_DOMAIN, 9)],
+            opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
         )
 
-    def test_upsample_raw_data_v7(self) -> None:
+    def _test_upsample_raw_data_v7(self, version) -> None:
```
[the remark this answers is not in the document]
guoyuhong: For the above similar 3 questions, `_test_${case_name}_v${op_version}` is not disabled test cases. It is called by `single_op_all_version_tester`. 
It looks like this way is misleading. Maybe I need to put it into the test function directly with some `if version == X` code paths, which is clearer.

onnx/test/shape_inference_test.py
guoyuhong: This test case expects some exceptions thrown before the version is actually used. However, `single_op_all_version_tester` requires the test function to have this argument.

onnx/test/shape_inference_test.py
guoyuhong: Thanks for the clarification. Changed.

guoyuhong:
@jcwchen how about the decorator way?

onnx/test/shape_inference_test.py
```diff
@@ -35,6 +41,68 @@
...
+        ```
+        @single_op_all_version_tester("Transpose")
+        def test_transpose(self, version):
+            ...
+        ```
+
+    Raises:
+        KeyError: If the op_name is not in `defs.get_all_schemas`.
+    """
+    if len(all_test_op_versions) == 0:
```
justinchuby, line 80: Do we need this map or are a few api calls in  this function sufficient?

onnx/test/shape_inference_test.py
```diff
@@ -35,6 +41,68 @@
...
+            all_test_op_versions[schema.name] = get_available_versions(schema)
+
+    def decorator(test_function):
+        if op_name not in all_test_op_versions:
+            raise KeyError(f"Invalid Op name: {op_name}")
+        versions = all_test_op_versions[op_name]
+        assert versions is not None
+
+        def wrapper(*args, **kwargs):
+            assert len(versions) > 0
```
justinchuby, line 92:
```suggestion
            assert versions
```

onnx/test/shape_inference_test.py
```diff
@@ -35,6 +41,68 @@
...
+        for schema in defs.get_all_schemas():
+            all_test_op_versions[schema.name] = get_available_versions(schema)
+
+    def decorator(test_function):
+        if op_name not in all_test_op_versions:
+            raise KeyError(f"Invalid Op name: {op_name}")
+        versions = all_test_op_versions[op_name]
+        assert versions is not None
+
+        def wrapper(*args, **kwargs):
```
justinchuby, line 91:
```suggestion
        def wrapper(test_instance, *args, **kwargs):
```

onnx/test/shape_inference_test.py
```diff
@@ -35,6 +41,68 @@
...
+            assert len(versions) > 0
+            for version in versions:
+                # Many versions found for this op.
+                if int(version) <= 5:
+                    # Not sure how to fix the Reshape error in self._make_graph.
+                    # Issue reference: https://github.com/onnx/onnx/issues/5289.
+                    continue
+                with args[0].subTest(op_version=version):
```
justinchuby, line 99:
```suggestion
                with test_instance.subTest(op_version=version):
```

onnx/test/shape_inference_test.py
```diff
@@ -35,6 +41,68 @@
...
+                    test_function(*args, version=version, **kwargs)
```
justinchuby, line 100:
```suggestion
                    test_function(test_instance, *args, version=version, **kwargs)
```

justinchuby (commented):
I suggest leveraging parameterized which is better than subtests. Let me create an example in another comment

jcwchen (commented):
> @jcwchen how about the decorator way?

I like this idea. Will take another look at the whole PR again.

justinchuby:
Consider the following:

```python
import parameterized

def all_versions_for(op_name: str) -> list[int]:
    # FIXME(#5289): Reshape errors in self._make_graph when version <= 5
    return [(version,) for version in all_test_op_versions[op_name] if version > 5]
```

then in the test class

```python
    @parameterized.expand(all_versions_for("Transpose"))
    def test_transpose(self, version) -> None:
        ...
```

This way each version can show up as a separate test.

onnx/test/shape_inference_test.py
guoyuhong: The decorator is removed.

onnx/test/shape_inference_test.py
guoyuhong: The decorator is removed.

onnx/test/shape_inference_test.py
guoyuhong: The decorator is removed.

onnx/test/shape_inference_test.py
guoyuhong: Use this in `all_versions_for`.

onnx/test/shape_inference_test.py
guoyuhong: The decorator is also removed.

onnx/test/shape_inference_test.py
guoyuhong: The decorator is removed.

guoyuhong:
@justinchuby If using `parameterized` module, the case name is not related to the version number. Instead, the case name suffix is like `0`,`1`, which are index numbers. When the case failed, we need to find out which version it fails.
image

I add an extra unused argument for test cases. It will also show in the case name:
image

I also use `pytest.skip` for some cases. It will also generate better information.
image

justinchuby (approved):
Looking good, thanks!

onnx/test/shape_inference_test.py
```diff
@@ -36,6 +44,38 @@
 from onnx.parser import parse_graph
 
 
+def get_available_versions(schema):
```
jcwchen, line 47: Please add type annotations for this function. Thanks!

onnx/test/shape_inference_test.py
```diff
@@ -189,64 +233,79 @@ def _identity_prop(self, op: str, **kwargs: Any) -> None:
...
         graph = self._make_graph(
             [("X", TensorProto.FLOAT, (2, 3, 4))],
             [make_node("Transpose", ["X"], ["Y"], perm=[1, 0, 2])],
             [],
         )
         self._assert_inferred(
-            graph, [make_tensor_value_info("Y", TensorProto.FLOAT, (3, 2, 4))]
+            graph,
+            [make_tensor_value_info("Y", TensorProto.FLOAT, (3, 2, 4))],
+            opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)],
```
jcwchen, line 246: nit: would it be cleaner to make `_assert_inferred` handle `opset_imports=[helper.make_opsetid(ONNX_DOMAIN, version)]` and then we can just pass version to it?

jcwchen (commented):
Overall looks good to me. Some nit comments. Thank you for the effort!

onnx/test/shape_inference_test.py
guoyuhong: It would be difficult to do so, because `opset_imports` does not only handler versions, it also handles domains. For example, in `test_category_mapper`, the `opset_imports` is fed with different version as well as different domains.
```python
        opset_imports=[
                make_opsetid(ONNX_ML_DOMAIN, 1),
                make_opsetid(ONNX_DOMAIN, 11),
            ],
```

guoyuhong:
@jcwchen I will apply this test scheme to all other test functions in a later PR and try to fix the Reshape issue when op verison <= 5 after that.

jcwchen (approved):

gramalingam (approved):

adityagoel4512:
When making updates to shape inference logic I assume we don't change the older implementations (e.g. https://github.com/onnx/onnx/blob/925840b04819b70d519b2f8cd60c7dbed8e6784d/onnx/defs/tensor/old.cc#L5032)?

guoyuhong:
@jus

> When making updates to shape inference logic, should we reflect these changes in older shape inference implementations (e.g.
> 
> https://github.com/onnx/onnx/blob/925840b04819b70d519b2f8cd60c7dbed8e6784d/onnx/defs/tensor/old.cc#L5032
> 
> )?

@adityagoel4512 I'm not sure. @jcwchen what do you think about this question?

jcwchen:
> When making updates to shape inference logic, should we reflect these changes in older shape inference implementations (e.g.

Good question... It's really case by case. I would say if there is a critical bug in old shape inference, we should. A common scenario is all opset versions do not behave correctly in some corner cases and ONNX should fix them consistently among all versions. Another common scenario is if some situations only happen in the latest opset version (e.g., new attribute or new documentation) rather than older ones, we don't need to update/add such a logic in older opest versions. Also we need to be very careful about some shape inference changes might break runtime backward compatibility.

adityagoel4512:
That makes sense to me, I suppose it'll require careful reviews then.

gramalingam:
Agree with Jacky. If the fix makes sense for older op versions, ideally we should make the change to older version too.

PR #5283, opened 2023-06-04
infer old squeeze on known input shape and empty axes

daquexian:
the shape inference of squeeze 13 can handle the case that the input shape is known and the axes is unspecified, but that of older version squeezes not. This PR fixes it.


onnx/defs/tensor/old.cc
```diff
@@ -4128,17 +4135,24 @@ ONNX_OPERATOR_SET_SCHEMA(
...
-            return;
-          }
-
           if (!ctx.getInputType(0)->tensor_type().has_shape()) {
             return;
           }
 
           ctx.getOutputType(0)->mutable_tensor_type()->mutable_shape();
           const auto& input_shape = ctx.getInputType(0)->tensor_type().shape();
+          const auto input_ndim = input_shape.dim_size();
```
xadupre, line 4144: That seems to be the same code than above, could it be moved in one separate function?

onnx/test/shape_inference_test.py
```diff
@@ -1183,6 +1183,21 @@ def test_squeeze(self) -> None:
...
+                ("x", TensorProto.FLOAT, (1, 3, 1, 1, 2, 1)),
+            ],
+            [make_node("Squeeze", ["x"], "y")],
+            [],
+        )
+        operatorsetid = OperatorSetIdProto()
+        operatorsetid.domain = ""
+        operatorsetid.version = 11
+        self._assert_inferred(
+            graph, [make_tensor_value_info("y", TensorProto.FLOAT, (3, 2))]
```
gramalingam, line 1198: I guess that the opset needs to be specified, eg., something like `opset_imports=[helper.make_opsetid(ONNX_DOMAIN, 9)]`

justinchuby: Good catch. This PR has since been fully superseded by #5967 and the dynamic-shape follow-up #6314. After merging current `main`, there is no remaining PR diff, so I’m closing the duplicate rather than carrying this stale test forward.

onnx/defs/tensor/old.cc
justinchuby: Agreed. This implementation has since been fully superseded by #5967, which merged the equivalent Squeeze-1/11 inference logic into `main`. After merging current `main`, this PR has no remaining diff, so I’m closing it rather than retaining a duplicate.

justinchuby:
Closing as superseded. PR #5967 merged the equivalent Squeeze-1/11 shape-inference behavior, and PR #6314 subsequently covered the dynamic-input case. I merged current main into this branch non-destructively; origin/main...HEAD now has no changed files, so retaining this PR would be an empty duplicate.

